PR #1, opened 2016-03-22
Modify problem in new project

shun-iwasawa:
Modified a following problem;
OpenToonz aborts when creating a new project with project folder path including "$scenepath" argument.


RodneyBaker:
Just now noting that every time a post includes a number reference such as #1 or # 1 the original report gets a reference.

Issue #34, opened 2016-03-26
Linux compilation


blurymind:
First of all thank you for releasing this to the world. This is a historic contribution to both the animation and the open source community. 
I would like to ask a few questions.

- Can this be compiled to run on linux? There doesn't seem to be any dependency on libraries that are not available. The open source community would greatly appreciate it and I would personally love to make nightly builds for testing. We can even setup a cooker repository where such builds can be automated and done on a server.

- Will there be a translation (english) made available - for the documentation, if there already isnt one? If not, how can we get started on localizing it to different languages?

hdunderscore:
Seems to me it should have some level of english support, even japanese seems to be posed as a translation rather than primary language: https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/translations/japanese/colorfx.ts

blurymind:
sorry, i just got a build from the new website - yes it does have english language. :)

Now onto figuring out how to compile this on linux!

razvanc-r:
I'm also highly interested in the Linux compilation!

AmEv7Fam:
While I know little Japanese, I do know how to manipulate PDF files, and do a "starting" translation with online translators, and manually fix the English nuances.

Szalacinski:
Doesn't look like it would be too hard to port this to linux.  Some files would have to be rewritten specifically to use Linux systems like ALSA instead of OSX systems like AudioToolbox.
お休みなさい
- CS

razvanc-r:
I'll have a quick look today about Linux port, see if I manage to spot something

justinclift:
Looking at the main CMakeLists.txt, it has platform dependent compiling pieces for windows and OSX.  Shouldn't be too hard to add a Linux clause, probably initial modelling on the OSX one, and take it from there. :)

* https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/CMakeLists.txt#L15
* https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/CMakeLists.txt#L24

Note, the APPLE clause has a hard coded path in it to a manually installed Qt5 version in the (OSX) users home directory.  It seems to be ok on OSX (so far) when changing that to a system wide provided version of Qt5.

* https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/CMakeLists.txt#L31

razvanc-r:
[...]

PR #37, opened 2016-03-26
Allow TIFF, LZO, and SuperLU libraries to be found when compiling for…

Szalacinski:
… Linux

I'm going to work on this some more tomorrow, but it's past midnight here in Tennessee, and I figured I might as well share the results of me digging through the code for a little bit.  For full Linux support, the main cmake file is going to have to have another option for Linux, boost libraries will have to be found, and some C++ files will have to be rewritten specifically for   #34

### DIFF
--- a/toonz/cmake/FindLZO.cmake
+++ b/toonz/cmake/FindLZO.cmake
@@ -1,5 +1,5 @@
-find_path(LZO_INCLUDE_DIR NAMES lzoconf.h HINTS ${THIRDPARTY_LIBS_HINTS} PATH_SUFFIXES lzo/2.09/include/lzo lzo/2.03/include/lzo)
-find_library(LZO_LIBRARY NAMES liblzo2.a lzo2_64.lib HINTS ${THIRDPARTY_LIBS_HINTS} PATH_SUFFIXES lzo/2.09/lib lzo/2.03/lib/LZO_lib)
+find_path(LZO_INCLUDE_DIR NAMES lzoconf.h HINTS ${THIRDPARTY_LIBS_HINTS} /usr/include/lzo/ PATH_SUFFIXES lzo/2.09/include/lzo lzo/2.03/include/lzo)
+find_library(LZO_LIBRARY NAMES liblzo2.a lzo2_64.lib HINTS ${THIRDPARTY_LIBS_HINTS} /usr/lib/i386-linux-gnu/ /usr/lib/x86_64-linux-gnu/ PATH_SUFFIXES lzo/2.09/lib lzo/2.03/lib/LZO_lib)
 
 message("***** LZO Header path:" ${LZO_INCLUDE_DIR})
 message("***** LZO Libarary path:" ${LZO_LIBRARY})
--- a/toonz/cmake/FindTIFF.cmake
+++ b/toonz/cmake/FindTIFF.cmake
@@ -1,6 +1,6 @@
 # looks for libtiff(4.0.3 modified)
-find_path(TIFF_INCLUDE_DIR NAMES tiffio.h HINTS ${SDKROOT} PATH_SUFFIXES tiff-4.0.3/libtiff/)
-find_library(TIFF_LIBRARY NAMES libtiff.a HINTS ${SDKROOT} PATH_SUFFIXES tiff-4.0.3/libtiff/.libs NO_DEFAULT_PATH)
+find_path(TIFF_INCLUDE_DIR NAMES tiffio.h HINTS ${SDKROOT} /usr/include/i386-linux-gnu/ /usr/include/x86_64-linux-gnu/ PATH_SUFFIXES tiff-4.0.3/libtiff/)
+find_library(TIFF_LIBRARY NAMES libtiff.a HINTS ${SDKROOT}  /usr/include/i386-linux-gnu/ /usr/lib/x86_64-linux-gnu/  PATH_SUFFIXES tiff-4.0.3/libtiff/.libs NO_DEFAULT_PATH)
 
 message("***** libtiff Header path:" ${TIFF_INCLUDE_DIR})
 message("***** libtiff Libarary path:" ${TIFF_LIBRARY})
--- a/toonz/cmake/FindSuperLU.cmake
+++ b/toonz/cmake/FindSuperLU.cmake
@@ -1,6 +1,6 @@
 # preferred homebrew's directories.
-find_path(SUPERLU_INCLUDE_DIR NAMES slu_Cnames.h HINTS ${THIRDPARTY_LIBS_HINTS} PATH_SUFFIXES superlu43/4.3_1/include/superlu superlu/SuperLU_4.1/include)
-find_library(SUPERLU_LIBRARY NAMES libsuperlu.a libsuperlu_4.1.a HINTS ${THIRDPARTY_LIBS_HINTS} PATH_SUFFIXES superlu43/4.3_1/lib superlu)
+find_path(SUPERLU_INCLUDE_DIR NAMES slu_Cnames.h HINTS ${THIRDPARTY_LIBS_HINTS} /usr/include/superlu/ PATH_SUFFIXES superlu43/4.3_1/include/superlu superlu/SuperLU_4.1/include)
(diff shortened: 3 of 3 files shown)


toonz/cmake/FindTIFF.cmake
```diff
@@ -1,6 +1,6 @@
 # looks for libtiff(4.0.3 modified)
-find_path(TIFF_INCLUDE_DIR NAMES tiffio.h HINTS ${SDKROOT} PATH_SUFFIXES tiff-4.0.3/libtiff/)
-find_library(TIFF_LIBRARY NAMES libtiff.a HINTS ${SDKROOT} PATH_SUFFIXES tiff-4.0.3/libtiff/.libs NO_DEFAULT_PATH)
+find_path(TIFF_INCLUDE_DIR NAMES tiffio.h HINTS ${SDKROOT} /usr/include/i386-linux-gnu/ /usr/include/x86_64-linux-gnu/ PATH_SUFFIXES tiff-4.0.3/libtiff/)
```
justinclift: As a concept, using the shared libraries already on the system is probably a decent idea.  It might be better to use pkgconfig to retrieve the locations/flags though, as that's already a dependency + should be more cross platform compatible.

As an example on OSX:

```
$ pkg-config --libs libtiff-4
-L/usr/local/Cellar/libtiff/4.0.6/lib -ltiff
$ pkg-config --cflags libtiff-4
-I/usr/local/Cellar/libtiff/4.0.6/include
```

Those values should be usable to pass into appropriate values.

justinclift:
As a side note, after getting everything to compile ok on OSX, I attempted to switch from the bundled tiff 4.0.3 to the latest upstream tiff (4.0.6).  That failed with a variety of errors, so it looks (for now) like the dependency on 4.0.3 is not easily changed. :wink:

No idea if the bundled tiff-4.0.3 has been customised.  I don't have the inclination to dig into it just now.  Going to tear myself away from the computer for a bit instead. :smile:

roentgen:
I tried get tiff to out before I know the original toonz version had used tiff-modified-4.0.3. So I give it up.
But I guess that does not prevent for linux-build while you use opentoonz/thirdparty/tiff-4.0.3.

justinclift:
Yeah.  The changes in tiff 4.0.3 will probably need to be ported to tiff 4.0.6, and (perhaps) submitted to the upstream tiff project, depending on what they are.

Szalacinski:
Pull request #51 looks a lot more thorough than this starter patch.  A lot more people hopped on this project than I initially thought they would.  FOSS is wonderful.
If you feel it would be better, feel free to close this pull request and merge his instead.
With how quickly the work is going along, I wouldn't be surprised if we have a working Linux build within a week.

justinclift:
I looked at the changes made to tiff 4.0.3, to see how much effort would be required to update them into tiff 4.0.6 (the latest available at this time):

&nbsp; &nbsp; https://github.com/justinclift/opentoonz/commits/tiff406

The only major difference is the addition of the file `libtiff/tif_getimage_64.c`, plus a few tweaks to the supporting files to ensure it is included.

This new tif_getimage_64.c file is copy of the existing tif_getimage.c file in the same directory, with substantial modifications made to it for 64 bit support.

Someone with an understanding of TIFF file formats + and good at C coding (not myself at the moment) would be able to update this to use the tiff 4.0.6 tif_getimage.c as it's base.  That should probably be done, since tiff-4.0.3 has known CVE's for it.

I don't think I'll do much more with this.  Maybe this investigation will help someone else get it done though. :smile:

Fordi:
I don't think this is the way to do it.  There's a bit in toonz/sources/CMakeLists.txt, line 8, the determines the content of THIRDPARTY_LIBS_HINTS - this is the search path for external libraries.

I found that if I glom this - eliminating the branches and just setting it to "/usr/include" and nothing else, it gets past liblzo, libusb, libsuperlu, and liblz4.  Adding "/usr/include" and flipping it (so it's if (WIN32) ... else() ... endif()) is probably fine, too.  But adding system-specific pathnames (such as /usr/include/i386-linux-gnu/ /usr/include/x86_64-linux-gnu) is usually a big no-no.

Here's the patch:

     diff --git a/toonz/sources/CMakeLists.txt b/toonz/sources/CMakeLists.txt
     index a1749d9..c0dbbab 100644
     --- a/toonz/sources/CMakeLists.txt
     +++ b/toonz/sources/CMakeLists.txt
     @@ -4,10 +4,10 @@ project(OpenToonz)
      get_filename_component(SDKROOT ../../thirdparty/ ABSOLUTE)
      message("SDK Root:" ${SDKROOT})
      set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "../cmake")
     -if(APPLE)
     -    set(THIRDPARTY_LIBS_HINTS "/usr/local/Cellar/" "/opt/include" ${SDKROOT})
     -elseif(WIN32)
     +if(WIN32)
          set(THIRDPARTY_LIBS_HINTS ${SDKROOT})
     +else()
     +    set(THIRDPARTY_LIBS_HINTS "/usr/local/Cellar/" "/opt/include" "/usr/include" ${SDKROOT})
      endif()
      
      message("Thirdpary Library Search path:" ${THIRDPARTY_LIBS_HINTS})

toonz/cmake/FindTIFF.cmake
hfiguiere: And to use within cmake: https://cmake.org/cmake/help/v3.0/module/FindPkgConfig.html

Szalacinski:
@Fordi's patch set looks a lot cleaner than mine.  It would probably be better to use his.  I'll keep this pull request open for the sake of conversation.

pwllm:
@Fordi : had to add `"/usr/include/qt"` to `THIRDPARTY_LIBS_HINTS` too (on Debian)

Fordi:
I've gotten pretty far on the Ubuntu build, but I had to be done at 3am.
I'll push what I've got to my fork later, when I get home.

A large number of changes are to do with case sensitivity.  One big change
I realized could have been done more easily is going from `defined(LINUX)`
to `__linux__` - GCC doesn't define LINUX, but it should be easy enough to
copy the constant over (rather than replacing instances as I've done).

Presently I'm stuck on the XImplementation of some GL stuff - which I'm not
entirely convinced is necessary - the OSX version uses Qt for its
implementation without recourse to Carbon/Cocoa, so I might be able to get
away with a similar trick there (but then, it was 3am, and I needed to be
done).  Anyway, I'll update with progress later tonight.

toonz/cmake/FindLZO.cmake
```diff
@@ -1,5 +1,5 @@
-find_path(LZO_INCLUDE_DIR NAMES lzoconf.h HINTS ${THIRDPARTY_LIBS_HINTS} PATH_SUFFIXES lzo/2.09/include/lzo lzo/2.03/include/lzo)
-find_library(LZO_LIBRARY NAMES liblzo2.a lzo2_64.lib HINTS ${THIRDPARTY_LIBS_HINTS} PATH_SUFFIXES lzo/2.09/lib lzo/2.03/lib/LZO_lib)
+find_path(LZO_INCLUDE_DIR NAMES lzoconf.h HINTS ${THIRDPARTY_LIBS_HINTS} /usr/include/lzo/ PATH_SUFFIXES lzo/2.09/include/lzo lzo/2.03/include/lzo)
+find_library(LZO_LIBRARY NAMES liblzo2.a lzo2_64.lib HINTS ${THIRDPARTY_LIBS_HINTS} /usr/lib/i386-linux-gnu/ /usr/lib/x86_64-linux-gnu/ PATH_SUFFIXES lzo/2.09/lib lzo/2.03/lib/LZO_lib)
```
iRi-E: find_library() shouldn't use either "i386-linux-gnu" nor "x86_64-linux-gnu" directly because it won't work for the other CPU architectures such as ARM. Instead, CMAKE_LIBRARY_ARCHITECTURE variable can be used here.

kogaki:

kogaki:

shun-iwasawa:

ideasman42:
Was just looking into linking issues - #866.

This PR is quite old, makes some suspicious changes (hard coding architectures in the path), and not even sure its needed - opentoonz now builds on latest Ubuntu and ArchLinux, so closing this PR.

If there are problems finding libraries, best report this as a bug, with details on your Linux version.

PR #45, opened 2016-03-26
Use Homebrew provided Qt5

justinclift:


mich181189:
Actually on second thoughts, this is a little odd since cmake has a variable QT5_PATH designed for just this purpose already. 

Still for consistency this is probably still a good idea.

justinclift:
Looks like PR's aren't really welcome.  Sorry to waste people's time. :wink:

Issue #50, opened 2016-03-26
Crashes Upon Initializing Toonz environment.


HyperGamer28:
I could use some help. I tried to run, even use Admin Privileges and Windows 8 Compatibility, but it ends up crashing. I am using the HDMI cable for my TV cause my Laptop Monitor Broke. Here are my Computer Specs:

Processor: Intel(R) Core(TM) i7-4510U CPU @ 2.00GHz
Video Card 1: NVIDIA GeForce GT 730M
Video Card 2: Intel(R) HD Graphics Family
Memory: 8.0 GB
Operating System: Microsoft Windows 8.1 (build 9600), 64-bit

rioforce:
This happens for me too, I have a Windows 10 laptop. My computer specs are:

Processor: AMD A6-4400M APU with Radeon HD Graphics 2.70 GHz
Ram: 4 GB
OS: Windows 10 Home, 64 bit
Monitor: 1366 x 768

I realize that my graphics card is not an Intel, it's an AMD, and my screen resolution is too small for the recommended requirements. I wonder, though, is it crashing because of this, or is it another reason? I'm going to test it on another computer later on and give my results.

AmEv7Fam:
Mentioned in #46: Do either of you happen to have a Logitech webcam?

rioforce:
I do, in fact, have two Logitech webcam, though neither is plugged in right now (and haven't been for over a month).

I did test the program on a desktop computer (I don't have the specs with me right now) that I know has never had a Logitech webcam attached, and it works, though, it has higher specs so that may be a completely different reason why it's working there and not here. Just thought I'd mention that.

odysseywestra:
Closing since this is a duplicate of #46.

Issue #46, opened 2016-03-26
Crashing on Windows 10 due to lvcod64.dll


falconmick:
as allot of people are reporting https://groups.google.com/forum/#!topic/opentoonz_en/H8CDe_Uko6E I too cannot get this app to run either via my compilation or the installer. I keep getting this: [link to exception message](https://lh3.googleusercontent.com/-AB3t9ccT_mM/VvYTybZ8MsI/AAAAAAAABOg/hb_P9NSeVQUSmtIkoHk2ORWw_i_oSn1jA/s1600/Capture.PNG) inside lvcod64.dll

When I attached my debugger I was able to see the line of code responsible but it really pissed me off (spent quite allot of time trying to get it running) so I deleted the source + qt + boost so I can't reproduce and connect debugger, unless you guys respond and ask me to re-do everything again (which I can)

I just don't know how responsive you will be so I don't want to spend another hr and have nobody reply to this bug.

EDIT:
I only have 1 screen plugged in

retsyo:
crash for my win7 64 bits too.

yondoloki:
Crash on Win 8.1 64 bit also as soon as I open the program

AmEv7Fam:
Ditto. Instantly crashes on my Win10 x64 machine.

Based on the responses, it seems to be 64-bit machines that are affected.

3Rton:
Same
Renaming lvcod64.dll immediately allows the software to run normally. Anyone got clues how to fix this?

AmEv7Fam:
OK. Apparently, lvcod64.dll is part of Logitech Webcam software.

Uhm, I've got a Logitech web right under my monitor. Oh...

Those of you who are experiencing crashes, who has Logitech webcams, or have at least had them hooked up at one time!

(It appears OpenToonz is trying to access the webcam as an image capture device.)

danielgett:
I'm having the same problem.  I can't even open the program.  Any advice?

rioforce:
Non tech-savy workaround for the time being:

1. Visit `C:/Windows/System32`
2. Find `lvcod64.dll`
3. Rename the DLL to something easy to find (for example `RENAMED-lvcod64.dll`). Remember that the Logitech software or Logitech camera might not work without the DLL.
4. Open OpenToonz (I ran it as admin if that helps)

On startup, it might show something like this (at least, it did for me, strangely. It didn't on my other computer where I didn't have Logitech issues though.):

[...]

PR #52, opened 2016-03-26
Fix crash in initializing

kusano:
In some enviroment, access violations occures in ICOpen.
Althogh this code seems to intended to handle SEH, structured exceptions are not catched in current configulation.

Thank you, Laura.
https://groups.google.com/d/msg/opentoonz_en/-3ajuxOjAGA/vbJhSvlADwAJ

#46, #50

### DIFF
--- a/toonz/sources/image/avi/tiio_avi.cpp
+++ b/toonz/sources/image/avi/tiio_avi.cpp
@@ -1061,11 +1061,19 @@ Tiio::AviWriterProperties::AviWriterProperties()
 				if (!rc)
 					break;
 				HIC hic = 0;
+#ifdef _MSC_VER
+				[&](){
+					__try {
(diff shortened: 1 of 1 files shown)


roentgen:
ok, but this actually means that only MSVC is eligible for Windows build.
Could we offer an option to omit AVI support if non-MSVC is present.
Or I think it's better that we express  #error in the #else block.

roentgen:
ok, I know this issue is CRITICAL. let's discuss about a better option another day.

kusano:
Thank you for merging.
`__try` and `__except` are MSVC specific, not Windows API.
I have no idea which is the best way, compile error, run time error or omitting AVI support.
At least this code will work well in non-MSVC if ICOpen does not crash.

PR #61, opened 2016-03-27
Linux: Frameserver system memory access

ideasman42:
Fix for building tfarmserver.cpp on linux

### DIFF
--- a/toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
+++ b/toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
@@ -50,7 +50,7 @@ using namespace std;
 #endif
 
 #ifdef LINUX
-#include <sys/sysctl.h> //For getting total memory
+#include <sys/sysctl.h>
 #endif
 
 // forward declaration
@@ -697,23 +697,19 @@ void FarmServer::queryHwInfo(HwInfo &hwInfo)
 	hwInfo.m_type = Irix;
 #else
 
-	TINT64 physMemSize;
-
 #ifdef MACOSX
 	int mib[2];
+	TINT64 physMemSize;
 	size_t len;
 
 	mib[0] = CTL_HW;
 	mib[1] = HW_MEMSIZE;
 	len = sizeof(physMemSize);
 	sysctl(mib, 2, &physMemSize, &len, NULL, 0);
+#endif
 
-#elif defined(LINUX)
-	physMemSize = (TINT64)sysconf(_SC_PHYS_PAGES) * (TINT64)sysconf(_SC_PAGE_SIZE);
-
-#else
-#warning "Unsupported platform!"
-	physMemSize = 500000000;
+#ifdef LINUX
+	TINT64 physMemSize = (TINT64)sysconf(_SC_PHYS_PAGES) * (TINT64)sysconf(_SC_PAGE_SIZE);
 #endif
 
 	hwInfo.m_cpuCount = TSystem::getProcessorCount();
--- a/toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
+++ b/toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
@@ -49,6 +49,10 @@ using namespace std;
 #include <sys/sysctl.h> //To retrieve MAC HW infos
 #endif
 
+#ifdef LINUX
+#include <sys/sysctl.h> //For getting total memory
+#endif
+
 // forward declaration
 class FarmServer;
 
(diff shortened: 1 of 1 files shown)


toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
```diff
@@ -49,6 +49,10 @@ using namespace std;
 #include <sys/sysctl.h> //To retrieve MAC HW infos
 #endif
 
+#ifdef LINUX
+#include <sys/sysctl.h> //For getting total memory
```
linkmauve: According to sysconf(3), and on my glibc/linux system, `sysconf` is defined in `unistd.h` and not `sys/sysctl.h`.

toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
```diff
@@ -702,6 +708,14 @@ void FarmServer::queryHwInfo(HwInfo &hwInfo)
 	len = sizeof(physMemSize);
 	sysctl(mib, 2, &physMemSize, &len, NULL, 0);
 
+#elif defined(LINUX)
+	physMemSize = (TINT64)sysconf(_SC_PHYS_PAGES) * (TINT64)sysconf(_SC_PAGE_SIZE);
```
linkmauve: You can also query `_SC_AVPHYS_PAGES` for the number of available pages, but like `_SC_PHYS_PAGES` it can be non-standard, so not available on every system (in particular, other non-Linux kernels that will take this path too).

For total virtual memory it’s a lot trickier, the total is dependent on the architecture, with e.g. 4 GiB on most 32-bit OS, 16 EiB theoretically on x86_64, but actually only 256 TiB on current processors.

On Linux you can query these values from `/proc/meminfo`, which can be a bit tricky to interpret but there is some good documentation online about it.  Also remember it’s Linux-only, and won’t be portable to other systems.

toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
```diff
@@ -693,8 +697,10 @@ void FarmServer::queryHwInfo(HwInfo &hwInfo)
 	hwInfo.m_type = Irix;
 #else
 
-	int mib[2];
 	TINT64 physMemSize;
+
+#ifdef MACOSX
```
linkmauve: Might be better to refactor all of the `#ifdef` of this function into a single one with multiple `#elif`.

toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
ideasman42: The purpose of this check is to find the physical memory on the system, not virtual memory.
So not sure why you would even consider virtual memory.

----

Changes to actual logic for testing memory may be fine, but outside the scope of porting code.

linkmauve: I was telling you about virtual memory because you currently just set everything to 500 MB (lines 723-725) and that’s quite wrong.

ideasman42: Setting to 500mb is done only for platforms besides Linux/OSX/Win32.
Currently these systems aren't supported anyway. We could do an `#error` instead too.

linkmauve: No, if you read the #if/#else#endif flow carefully you will see it is only set properly on win32, both OSX, Linux and other platforms fallback to the 500 MB for available physical memory and total/available virtual memory.

ideasman42: Committed fix.

toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
```diff
@@ -49,6 +49,10 @@ using namespace std;
 #include <sys/sysctl.h> //To retrieve MAC HW infos
 #endif
 
+#ifdef LINUX
+#include <sys/sysctl.h>
```
linkmauve: As I already told you earlier, `sysconf` and the two constants you use are declared in `unistd.h` (which is already included), not `sys/sysctl.h`.

toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
```diff
@@ -701,6 +706,11 @@ void FarmServer::queryHwInfo(HwInfo &hwInfo)
 	mib[1] = HW_MEMSIZE;
 	len = sizeof(physMemSize);
 	sysctl(mib, 2, &physMemSize, &len, NULL, 0);
+#endif
+
+#ifdef LINUX
+	TINT64 physMemSize = (TINT64)sysconf(_SC_PHYS_PAGES) * (TINT64)sysconf(_SC_PAGE_SIZE);
```
linkmauve: I can’t point to the lines below, but look at the lines 719-721, you can provide a correct value for those, as I explained in https://github.com/opentoonz/opentoonz/pull/61#discussion_r57526961 in which case you can put them back to the OSX case only.

toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
```diff
@@ -701,6 +706,11 @@ void FarmServer::queryHwInfo(HwInfo &hwInfo)
 	mib[1] = HW_MEMSIZE;
 	len = sizeof(physMemSize);
 	sysctl(mib, 2, &physMemSize, &len, NULL, 0);
+#endif
```
linkmauve: My suggestion here was to have
```
#if defined(WIN32)
…
#elif defined(__sgi)
…
#elif defined(MACOSX)
…
#elif defined(LINUX)
…
#endif
```
for the whole function, instead of the current three levels of imbrication.

skitaoka:
merged, thanks

toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
ideasman42: Agree with this, think it could be done in a separate PR.

PR #63, opened 2016-03-27
Linux: Basic portability edits

ideasman42:
Mostly pre-processor and define tweaks for Linux portability.

(Other changes are needed too, but these should be simple to review)


toonz/sources/common/tfx/ttzpimagefx.cpp
```diff
@@ -5,6 +5,10 @@
 
 #include "ttzpimagefx.h"
 
+#ifdef LINUX
+#define strtok_s strtok_r
+#endif
+
```
andreygursky: strtok_s is used only for Windows, thus maybe just replace strtok with strtok_r for both MacOS X and Linux?

toonz/sources/common/twain/twain.h
```diff
@@ -97,6 +97,11 @@
  * Platform Dependent Definitions and Typedefs                              *
  ****************************************************************************/
 
+
```
andreygursky: Unneeded new line here.

toonz/sources/tnzbase/tscanner/tscannerutil.cpp
```diff
@@ -2,6 +2,10 @@
 
 #include "tscannerutil.h"
 
+#ifdef LINUX
+#include <cstring>
+#endif
```
andreygursky: for memcpy() but why only for LINUX?

toonz/sources/common/tfx/ttzpimagefx.cpp
ideasman42: Good point, committed change so now Linux & OSX use same code-block.

toonz/sources/common/twain/twain.h
ideasman42: Corrected

toonz/sources/tnzbase/tscanner/tscannerutil.cpp
ideasman42: Removed Linux ifdef's,

Only added them because (I assume), its not needed on other platforms.
However I don't have a strong opinion here.

toonz/sources/common/tfx/ttzpimagefx.cpp
```diff
@@ -5,17 +5,13 @@
...
-#endif
-
 //**********************************************************************************************
 //    Global  functions
 //**********************************************************************************************
 
 void parseIndexes(string indexes, vector<string> &items)
 {
-#ifndef MACOSX
+#ifdef WIN32
```
KindDragon: better use _WIN32 instead http://stackoverflow.com/a/662543/61505

ideasman42: Currently `WIN32` is being checked for all over.

Changes to existing conventions better be handled outside small patches like this.

toonz/sources/include/tpixel.h
```diff
@@ -154,6 +154,8 @@ class DVAPI DV_ALIGNED(8) TPixelRGBM64
 	Channel b, g, r, m;
 #elif TNZ_MACHINE_CHANNEL_ORDER_MRGB
 	Channel m, r, g, b;
+#elif TNZ_MACHINE_CHANNEL_ORDER_RGBM
+	Channel r, g, b, m;
 #else
```
roentgen: It looks like good. But is this need for it?

toonz/sources/common/ttest/ttest.cpp
```diff
@@ -15,6 +15,10 @@
 
 #include <set>
 
+#ifdef LINUX
```
roentgen: I think that '#if defined(LINUX)' is better from a point of consistency.

ideasman42: done

toonz/sources/include/tpixel.h
ideasman42: Its needed to compile, Linux could use `TNZ_MACHINE_CHANNEL_ORDER_MRGB` too. But rather focus on fixing build errors first.

PR #64, opened 2016-03-27
Linux: Use lowercase for all resources (failed to build)

ideasman42:
Upper/Lower case mismatch caused the build to fail.
Use lowercase for all image extensions.


linkmauve:
LGTM.

PR #68, opened 2016-03-27
READMEとビルド方法の英訳

m-grimaud:
I've given a rough translation for English for the README and the build instructions for both Windows and MacOSX.


justinclift:
@m-grimaud As a thought, would you be ok to squash all of these commits into a single one?

It's generally good practise, and helps keep code history concise. :smile:

PR #85, opened 2016-03-28
Fix build when libtiff exists on system

mich181189:
On OSX at least, if you have libtiff in your system include path,
it will use that instead of the bundled modified version, despite
having found the correct version in FindTIFF.cmake. This commit
fixes that by making sure the libtiff include directory is before the system
include paths.


justinclift:
Ahhh cool.  I'll try this out later too.

mich181189:
There might be a better way of achieving this now I look at it - only the image folder references tiff includes.

roentgen:
we discussed around issues/#79. this fix is good.

Issue #91, opened 2016-03-28
Windows 10: creating New Toonz Raster Layer crashes


bigjko:
The program crashes if I try to create a Toonz Raster level, but Toonz Vector and Raster work fine.

On Windows 10.

odysseywestra:
Thank you for bug report. Could you provide us 15-30 video illustrating this problem?

bigjko:
Sure! It's v1.0.1, btw.

https://www.youtube.com/watch?v=28mJOuC1QVs

odysseywestra:
Okay I can confirm this on my Windows 10 VM. When I did this, it say `lzcompress.exe` has stopped working so it's probably related to that.

How I reproduced the steps:
* Open Command Prompt
* Type in `"C:\Program Files\OpenToonz 1.0\OpenToonz_1.0.exe"` and Hit Enter
* Right Click on a Cell on the Xsheet and Click on `New Level`
* Name Level and Change type to Toonz Raster Level
* OpenToonz should crash with `lzcompress.exe` error.

brambox:
Toonz Raster is broken and crashes on windows 8.1+  
I hope we can get fix sometime soon.
As that is the biggest thing in Toonz!

Goofybud16:
Toonz Raster and Raster layers also appear to be broken in the Linux build from PR #51.

If I select either and create the layer, the application segmentation faults (equivalent of "Stopped Responding" on Windows.)

Has anyone seen if this causes crashes on Mac?

brambox:
Creating normal raster and toonz raster works on mac os x but its still hard to use it as there are tons of weird issues.
For example not always when you draw it actually  doing it...  
Also the program closes on own pretty often.

yvanc:
I was testing things out and I noticed that this bug was not present in OpenToonz1.0.
This bug only came along in the new release OpenToonz1.0.1.

Gulwer:
Hi
OSX version has the problem with crashing when creating a toonz raster level. it is present in the latest version 1.0.1. I tried Toonz 1.0 also and it did not crash so the problem came with the update. 

[Toonz_rasterlevel.pdf](https://github.com/opentoonz/opentoonz/files/200370/Toonz_rasterlevel.pdf)

SafariMonkey:
So it's confirmed as a regression in 1.0.1. Thanks for the info!

CodyEakins:
Encountered the same problem; trying to create a Toonz Raster Level consistently crashes the program.
[...]

PR #95, opened 2016-03-28
Fix crash of lzocompres and lzodecompress

kusano:
Without stdlib.h malloc is assumed to return int, which is not compatible with void * in x64.
#91

### DIFF
--- a/thirdparty/lzo/driver/lzocompress.c
+++ b/thirdparty/lzo/driver/lzocompress.c
@@ -15,6 +15,10 @@
 
 
 #include <stdio.h>
+#include <stdlib.h>
+#if defined(_WIN32)
+#include <io.h>
+#endif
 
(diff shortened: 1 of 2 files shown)


odysseywestra:
I'm guessing this should address #91?

linkmauve:
Is this also reported upstream?

kusano:
No.
Files in `thirdparty/lzo/driver/` are not LZO original.

mich181189:
Ah isn't C's default into behaviour wonderful!

Luckilly it's a simple fix though! What's with the io.h include? Is that for a similar problem?

kusano:
@mich181189
Because `_setmode` is in io.h.
Although `_setmode` actually returns `int`, eliminating warnings is meaningful.

skitaoka:
LGTM

PR #120, opened 2016-03-31
Use #error directive instead of explicit syntax error

ideasman42:
We had a report (that I can't confirm), that this caused errors.
Best to use `error` here.

See report on the problem: https://github.com/ideasman42/opentoonz/pull/7


skitaoka:
LGTM

PR #129, opened 2016-04-02
OSX fixes

mich181189:
I don't necessarily expect these to be taken, but I thought I'd push up what worked for me - I got a full build on OSX out of this.

Unlike the fix-cmake-osx branch this does not needlessly override the find modules for libjpeg and libpng * I don't really see a problem if the MONO versions are used ;-) as long as the tiff headers come from the right place.

* I'm using the version of Boost I already had around (1.59) since as long as it's new enough it works fine.

* the version of the developer tools I have for OSX really doesn't like the use of abs


mich181189:
Oh, and for anyone following along, my cmake line was 

`CMAKE_LIBRARY_PATH=/opt/local/lib:/usr/lib CMAKE_INCLUDE_PATH=/opt/local/include:/usr/include /Applications/CMake.app/Contents/bin/cmake ../sources/ -DQt5_DIR:PATH=~/Qt5.5.1/5.5/clang_64/lib/cmake/Qt5`

notably causing the macports folder to be searched before the system libraries (libiconv gets mixed up otherwise) (yes - I used macports rather than homebrew this time)

toonz/sources/CMakeLists.txt
```diff
@@ -187,7 +187,7 @@ endif()
 include_directories(BEFORE ${TIFF_INCLUDE_DIR})
 
 find_path(BOOST_ROOT include/boost boost HINTS ${THIRDPARTY_LIBS_HINTS} PATH_SUFFIXES boost155/1.55.0_1 boost/boost_1_55_0/)
-find_package(Boost 1.55 EXACT REQUIRED)
+find_package(Boost 1.55 REQUIRED)
```
justinclift: Ahhh, interesting.  Was wondering if newer Boost would work, and haven't bothered trying yet.

kogaki:

kogaki:

skitaoka:
this issue was fixed by PR https://github.com/opentoonz/opentoonz/pull/212.
thanks.

PR #131, opened 2016-04-03
Fixes for abs()

janisozaur:
abs() only returns integer values, while arguments and results are real
in the cases fixed.

Also fixes some unexpected tokens after `#endif`.


andreygursky:
> Also fixes some unexpected tokens after #endif.

It'd be better not to mix this addition here. And fix them all in a separate PR, which will be hopefully merged promptly due it's triviality.

janisozaur:
ok, I'll separate these and hopefully it'll get merged.

skitaoka:
LGTM

PR #132, opened 2016-04-03
Encapsulate expressions in parentheses

janisozaur:
Some expressions appear to be confusing, adding parentheses removes
doubt in those cases.


skitaoka:
LGTM

PR #105, opened 2016-03-29
Revert formatting changes to tif_getimage_64.c

jcowgill:
This reverts all the formatting changes made to `tif_getimage_64.c` when compared to `tif_getimage.c`. This should make the code easier to get upstream at some point, or apply fixes to.

This change is refactoring only, it should have no functional effect.

PR #141, opened 2016-04-05
Unformat thirdparty/tiff-4.0.3/libtiff/tif_getimage_64.c

kusano:
to compare with original tif_getimage.c.
cf. #105


kogaki:
As discussed in #105 , this commit just revert an unexpected formatting and cause no functional effect.
LGTM

Issue #145, opened 2016-04-06
Default camera preset cannot be loaded.


shun-iwasawa:
Camera preset are listed in combo boxes both in the Output Settings and in the Cleanup Settings.
For now, selecting item from default camera preset does not work.
When you select some item, a warning dialog "Bad camera preset" opens.

This is because in Toonz Ghibli Edition we decided to save in preset string not only pixel size and aspect ratio but also inch size and offset. (Note that offset is saved only for cleanup camera.) And when applying such modification, we did not give priority to keep compatibility with Harlequin's preset data.

skitaoka:
This issue is fixed by https://github.com/opentoonz/opentoonz/pull/146.

PR #146, opened 2016-04-06
Fix loading camera preset (reslist.txt)

shun-iwasawa:
This fix is to solve the issue #145 ; Default camera preset cannot be loaded.

With this modification OpenToonz can load both type of the preset; the default preset in reslist.txt (with only pixel size and aspect ratio, which is the same as Toonz Harlequin) and the "OpenToonz version" preset (with pixel size, aspect ratio, inch size and offset, which will be added by users).

### DIFF
--- a/stuff/projects/cleanupreslist.txt
+++ b/stuff/projects/cleanupreslist.txt
@@ -76,16 +76,16 @@ Academy Projection1 4K, 3656x1976, 1.85
 
 Academy Projection2 4K, 4096x2214, 1.85
 
-Anamorphic Pre-squeezed, 1K 914x774, 1.18
+Anamorphic Pre-squeezed 1K, 914x774, 1.18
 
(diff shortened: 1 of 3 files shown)


CodyEakins:
I support this pull request.
Some users have taken the liberty of replacing their OpenToonz relist.txt file in order to fix this issue in the mean time. 
But this is an excellent solution, considering backwards compatibility.

shun-iwasawa:
>CodyEakins
Yes it does. As you pointed, the cleanup camera presets are saved in a different file (cleanupreslist.txt). This fix covers for loading it as well. The difference between two presets (reslist.txt for render and cleanupreslist.txt for cleanup) is that ns/ew offset values are saved only in cleanup camera preset.

skitaoka:
Jenkins, test this

PR #150, opened 2016-04-07
remove unnecessary backup files (.bak)

henrich:
It should be cleaned up before upload :)

And it also help me to debian package with git (Linux porting is not yet, but preparing it...)


kogaki:
Thanks. LGTM :)

Issue #79, opened 2016-03-28
Mono prevents a compilation of libimage on OSX


roentgen:
find_package() uses to look for libpng/libjpeg may look Mono.framework/Headers/ out.
Mono.framework/Headers has png.h and tiffio.h on the directory, this cause an ijection of priority path for tiffio.h since it needs dedicate libtiff for it.

Environment: OSX installed Mono framework

roentgen:
For the issue, I'm working in progress in this branch: https://github.com/opentoonz/opentoonz/tree/fix-cmake-osx

mich181189:
I've got a fix for the libtiff issue on #85 - make sure the tiff path is at the beginning of the include list.

roentgen:
I think that's easy way. But we'll have to take care of the order of include paths among (future) fixes.

2nd option: Offering custom FindJPEG.cmake and FindPNG.cmake can exclude unexpected path (such as Mono) 
I'm trying to this in https://github.com/opentoonz/opentoonz/tree/fix-cmake-osx

3rd option: We'll get thirdparty/tiff-4.0.3(customized) distinguishable much clearly from other libtiff.

Which option is better for us?

roentgen:
oh, Mono's libpng is old(1.2.47), and it's not proper version!

mich181189:
I don't think excluding "unexpected" paths in general is a good idea. Taking jpeg and tiff from mono is fine if it works and if not, cmake find modules usually allow you to hint at where you want to find it if you have a load of different versions around.

Your modification to FindTIFF.cmake is probably a good move though. 

In summary, I think this is only really a problem when you have a custom modified library like this. In summary I don't think we should be trying to get too magical with this.

mich181189:
@roentgen your PNG change, for example, stops it finding the system PNG on OSX (and probably on other platforms, e.g linux) which is maybe not what we want here.

roentgen:
@mich181189 could you let me see the error?

> cmake find modules usually allow you to hint at where you want to find it if you have a load of different versions around.

It looks that find_package()'s config mode does not work for me. 
Could you tell me how to give a hint to find_package() without PNGConfig.cmake?

In the module mode, I wrote the small test to share the case,

    project(findpng)
    find_package(PNG 1.6 REQUIRED)
    message("find_package(PNG 1.6 REQUIRED): " ${PNG_INCLUDE_DIRS})

[...]

PR #151, opened 2016-04-07
Fix cmake to be able to search custom libpng include path

kogaki:
This PR is similar approach as https://github.com/opentoonz/opentoonz/pull/85 .

I bumped into the problem discussed in https://github.com/opentoonz/opentoonz/issues/79 . With this fix and some cmake option, I could finally build OpenToonz successfully.

Here's my cmake options

```bash
cd toonz/build
CMAKE_PREFIX_PATH=~/Qt5.5.1/5.5/clang_64 cmake \
  -DTIFF_INCLUDE_DIR=../../thirdparty/tiff-4.0.3/libtiff/ \
  -DPNG_PNG_INCLUDE_DIR=../../thirdparty/libpng-1.6.21/  \
  ../sources
make
```

### DIFF
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -184,7 +184,7 @@ else()
     message("LZO:" ${LZO_INCLUDE_DIR})
 endif()
 
-include_directories(BEFORE ${TIFF_INCLUDE_DIR})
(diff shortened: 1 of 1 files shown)


kogaki:
Since -DPNG_PNG_INCLUDE_DIR looks weird, `CMakeLists.txt` needs more fix. 
This commit is just a workaround.

tomosu:
I confirmed it works on my environmnet.

PR #168, opened 2016-04-13
Fix memory allocation at doDirectionalBlur

skitaoka:
- merge https://github.com/opentoonz/opentoonz/pull/160
- use unique_ptr<>


kogaki:
LGTM

PR #170, opened 2016-04-13
Fix CMakeList.txt for nightly build

skitaoka:


skitaoka:
This PR is no longer necessary.

PR #172, opened 2016-04-13
Fix to include cmath

kogaki:
This fixes commit#6dca1d3a6d64f6b55d5545fd209acc0f81aab9ec
In OSX, we need to include `cmath` to use `std::abs(float)`


tomosu:
Build on my OSX succeeded and it works.
Thanks for your fix.

Issue #109, opened 2016-03-30
Crash related to multiple monitors


scardario:
Hi, I have a crash every time I try to open the program and I think it's because multiple monitors

The only way I can start opentoonz is by deleting all the configuration files created in some subfolders of "open toonz stuff" folder during the start of the application, that's about 27 files. I just make a search for all the files in that folder, organize the search by modification date and delete all of those that were created last.

The crash seems to happen when I have a window in other monitor and that configuration is saved in some file.

If I don't move a window to another monitor the software seems to start again without problems, if I move a window, like say, the combo window, to other monitor, the software will crash the next time I try to start it.

I use 3 monitors (2 monitors and a cintiq), changing a single window to any of the non-primary monitors will crash the application next time I start it.

I'm running win10 pro, no monitor managers installed. Please let me know if you need additional information.

tomByrer:
Tried with your 2 monitors & not your cintiq?

moonracer191:
I use 2 monitors and a cintiq also. I haven't had any trouble with moving windows but if I try to load opentoonz with the cintiq it crashes on startup. But if I load opentoonz and then plug in my cintiq I have no trouble moving any panels to it. idk what the problem is but for me at least as long as I unplug my cintiq when i load its perfectly fine so you could try that

artintime:
Hi,
I am also using a cintiq and two monitors and Toonz worked everytime as long as I keep it on the main monitor, but as soon as i drag it to my Cintiq to work the program remembers the location and crashes the next time i open it. However, it did work when I dragged it back to the main monitor. so maybe just bring the program back after you finish your work is simple as that before they fix the issue?
Thought to share this observation and hope it'll come handy to some.
Cheers.

3Rton:
I can confirm Artintime, this also occurs to me in Windows 10 64-bit OTz 1.0.1 version.
I didn't notice before because my work monitor is my main monitor, but if I start the program, drag it to secondary screen and close it, next time opening it gives the error.

tomByrer:
I wonder if this is pertaining to USB monitors, & not exclusive to cintiq?

Nyrdbrain:
scardario, can you list the files you had to delete to get it to start again?
I'm having a similar issue with dual monitors, but nothing I've deleted seems to allow it to restart.
If anyone knows where it saves window location data I'd like to know that.

Nyrdbrain:
Actually found where it is.

[...]

PR #174, opened 2016-04-13
Workaround for crash in multiple monitors

kusano:
Dirty hack, but crash in multiple monitors is a big issue.
#109


skitaoka:
LGTM

adamearle:
im happy to do some testing. Can you put together a "Testing feature list" that way we can subtract or add to it. I have also shared the nightly build like with this small community:

[]](https://plus.google.com/communities/117328766497043417332)

but it would be good to get more people on there for testing everytime somethings needs to be tested so it could be more thorough.

Adam

scardario:
Hi, I just tested the night build and it doesn't crash anymore, that's awesome, thank you very much.

When I start opentoonz again the windows from other monitors are left minimized in the main one. It would be cool to have the program to remember the position and size of the windows.

At least I can work now without having to unplug and plug some monitors. Thank you!!

PR #176, opened 2016-04-14
Reduce boost dependencies

skitaoka:
- replace smart pointers (use std::shared_ptr instead of boost::shared_ptr)
- remove BOOST_NO_RVALUE_REFERENCES (require C++11)
- use the smart pointer for m_imp


tomosu:
Thanks for your refactoring.
I've validated it works.

PR #177, opened 2016-04-14
Allow building with a newer version of Boost (up to 1.60.0).

ss23:
The instructions for both OSX and Windows still say to use 1.55.0, but cmake will allow up to 1.60.0.
Using a later version will also mean the additional patch for VS2013 is not required.


kogaki:

kogaki:

skitaoka:
merged, thanks

Issue #138, opened 2016-04-04
Crash On Start-up with Blackmagic Design installed


krinux:
I recently installed Opentoonz, and it keeps crashing on startup. I open the program, and it freezes on initializing toonz environment. After waiting for about a minute the program stops responding and it closes out. I do not have a logitech webcam, nor do I run on multiple monitors. I have both a C: Drive and a secondary D: drive and have tried installing it on both with no success.  My computer is up to specs, so I'm not entirely sure what the issue is. Any ideas/help?

Here are my system specs: 
OS: Windows 10 Home
Processor: Intel Core i7-5700 HQ CPU @ 2.70 GHz
2.70 GHz
RAM: 16.0 GB
GPU: NVIDIA GeForce GTX 970M

ohsnapitsjoel:
I get crashes as well, but not every time. It's predictable for me, and I can reproduce it.  I run a few monitors, including a tablet monitor.  Fresh install of OpenToonz: if I open the application, the window opens on my main display (an Acer monitor).  As long as I exit the application with the OpenToonz window on my main display, I can restart it without a problem and even move the window to my tablet monitor.  

If I close OpenToonz while the window is on my tablet monitor, I am not able to restart the application - it crashes while loading. 

If I delete the OpenToonz 1.0 stuff > profiles > layouts >OpenToonz.user folder, I can restart the application successfully.  The settings that were saved on exit (like which monitor the window was on) are now forgotten, obviously, and the application window again opens on my main display.  Repeat ad nauseum.  

Event Viewer says the faulting module is Qt5Gui.dll installed with OpenToonz.  

Windows 10 Home x64, AMD FX-8350, Geforce GTX 960, Acer monitor and a Yiynova MSP19U+ V5.  

@krinux - you're on a laptop (I assumed from the 970M)?  Can you plug in another monitor and see if you can reproduce this? Also, can you check Event Viewer and see if you're getting the same Qt5Gui.dll error?  Event Viewer > Windows Logs > Application

![2016-04-07 12_55_51-event viewer](https://cloud.githubusercontent.com/assets/7570180/14359653/89289a78-fcc0-11e5-8d34-110737fcbeca.png)

ohsnapitsjoel:
Whatever the underlying issue is, it's hidden in the OpenToonz stuff > profiles > layouts> OpenToonz.user > mainwindow.ini file.  If I reset it to its default values, it'll launch again, on my main display.

krinux:
[...]

Issue #178, opened 2016-04-14
Upgrade to qt5.6


ss23:
qt5.6 states it includes more Windows 10 support, which a few people are getting, as well as other fixes of course.
It would also help prevent issues like #109.
I've checked and it compiles fine (I haven't done extensive testing though).

If this is something that should be done, it's simply a change to the documentation and cmake which I'm happy to make.

ss23:
This has been completed now.

Issue #165, opened 2016-04-13
Error upon launch (Windows 10)


Tyrvde:
OpenToonz stopped working upon launch. It seems to be a common error, so I ran it through Visual Studio's debugger and this is what I've found. Hopefully it can shed a bit of light on the situation. 
![debug_opentoonz](https://cloud.githubusercontent.com/assets/16713353/14479514/1045155e-00e5-11e6-9b02-c6333bebdc71.PNG)

Tyrvde:
Also noted is that the error persists after running under Windows 7/8 compatability mode, both as an Administrator and without said privledges.

ss23:
@Tyrvde It's hard to tell exactly what the problem is just based on that, but since it's qt based, I've compiled a version of opentoonz with a newer version of qt.

Are you able to try downloading and running https://github.com/ss23/opentoonz/releases/download/qt56/OpenToonz-qt5.6.7z? It's a 7zip archive, but once extracted, you should be able to run it in place.

If it works, it may point to one of the issues that was fixed with Windows 10 on qt5.6.

kogaki:
Now we are building with Qt5.6 and this seems to be fixed

PR #179, opened 2016-04-14
Upgrade to Qt5.6

### LINKED ISSUE #109: Crash related to multiple monitors

ss23:
Fix #109 #138 #165 #178.
Fixes multiple issues on Windows 10.

I haven't tested this on anything other than Windows, nor have I done any extensive testing at all. Will require review/QA.


ss23:
Whoops, too many issues in there. Don't believe it will fix #138

kogaki:

toonz/sources/toonz/comboviewerpane.cpp
```diff
@@ -105,16 +105,6 @@ ComboViewerPanel::ComboViewerPanel(QWidget *parent, Qt::WFlags flags)
...
-#ifdef Q_OS_WIN
-	//  Workaround for QTBUG-48288
-	//  This code should be removed after updating Qt.
-	//  Qt may crash in handling WM_SIZE of m_sceneViewer in splash.finish(&w)
-	//  in main.cpp. To suppress sending WM_SIZE, set window position here.
-	//  WM_SIZE will not be sent if window poistion is not changed.
-	::SetWindowPos(reinterpret_cast<HWND>(m_sceneViewer->winId()),
-		HWND_TOP, 0, 0, 0, 0, SWP_NOMOVE | SWP_NOSIZE);
-#endif
-
```
kusano: :+1:

andreygursky: You could consider replacing like:
```diff
- #ifdef Q_OS_WIN
+ #if QT_VERSION >= 0x050000 && QT_VERSION < 0x050600
```
and let it remain here for a while instead of promptly remove for all Win users.

ss23: I don't think there's much point of leaving it there if the project is going to require Qt5.6, but happy to do so if it wouldn't be merged otherwise.

andreygursky: I mean that making Qt5.6 a default, recommended version is a way to go. But since a workaround is found for previous Qt versions is found, it's not worth to immediately drop it and force users to upgrade their Qt installation.

kogaki:

kogaki:
We finally merged it. Thanks! @ss23

PR #184, opened 2016-04-15
Remove license and edition defines and code

ss23:
Previously, old code to support various editions of Toonz, as well as license activation code, was cluttering the repository.

I'm not in a good position to determine whether this way of doing it is correct. While most of the code seems to *exclude* features based on the edition, there are a few places with different/new code relevant to only a specific version.
That being said, given these features are already disabled in the current opentoonz release, it won't be missed.

I've checked compilation and running on Windows only.


kogaki:

skitaoka:
Jenkins

kogaki:
Of course, activation is completely useless feature. Thank you for your work.
But I feel it is too early to remove some features of `Linetest`. 
They might be bundled into OpenToonz in the future. 
How about reverting some changes and keep codes for `Linetest` ?

blurymind:
I agree with @kogaki - disabled features that have value to the workflow should be kept.
Only software licensing features should be removed.

ss23:
I'll make the change to include `Linetest` this week :)

kogaki:

ss23:
@kogaki I've gone through and made sure the LINETEST support is there again. There are a few places where it seems LINETEST had features specifically removed, so I've left them enabled.
Is this okay like this, or did you want me to ensure 100% compatibility with how it used to be?

kogaki:
Thanks. I think it is OK if key features are preserved. I will test your PR.

kogaki:
Jenkins

kogaki:
This commit was merged with #291 . Thanks a lot, @ss23 !

PR #185, opened 2016-04-16
Thirdpirty binaries should be ignored by git

kogaki:
`lzo` and ``tiff  are ignored in the same manner as `glew`, `jpeg`


kogaki:
Jenkins

kogaki:
Jenkins

ideasman42:
Shouldn't these be absolute paths? ( see: http://stackoverflow.com/questions/6168736 ),
the existing ones as well as the new paths.

kogaki:
Thanks, I prettified them.

kogaki:
Jenkins

PR #186, opened 2016-04-17
Include memory, needed for auto_ptr references

ideasman42:
The `<memory>` include is needed for building on Linux.

Without this, there is an error with GCC5.3.

```
toonz/sources/common/expressions/tgrammar.cpp:147:7: error: 'auto_ptr' in namespace 'std' does not name a template type
  std::auto_ptr<CalculatorNode> m_a;
```

These could be ifdef'd for Linux, however this is a standard include for the C++ standard according to:
http://www.cplusplus.com/reference/memory/

This commit also adds some `<cstring>` includes which are needed for building on Linux.


kogaki:

skitaoka:
Jenkins

PR #189, opened 2016-04-17
Fix const enum which is an error with GCC5.3

ideasman42:
This gives an error with GCC, and has no useful meaning with other compilers, see: 
http://stackoverflow.com/questions/17805427


kogaki:

skitaoka:
Jenkins

Issue #59, opened 2016-03-27
On Brush Tool Crash


paulorobertocruz:
In the "Cleanup" window, when the brush tool is selected opentoonz crash on alt+left click
Windows 8 
Intel i5
Sorry for my bad english.

harrisonkuo:
I can confirm that this problem exists. I could reproduce it on a Windows 10 64-bit with an nVidia card. By going to any environment (Cleanup, PltEdit, InknPaint) and pressing alt (without even clicking or scrolling), OpenToonz crashes. This only occurs when using the Brush tool.

ZackCasey:
I've had a similar crash with the alt+ functions. Alt+scroll causes 1.0.1 to crash.

kogaki:
It sould be fixed in `nightly build`. Thanks.

adamearle:
i know this must  be really obvious. but where do I find the nightly build page or how do i navigate to there.

kogaki:
@adamearle You can find every versions here, https://github.com/opentoonz/opentoonz/releases :)

PR #193, opened 2016-04-18
Fix brush tool crash

shun-iwasawa:
This fix is for the issue #59.
This problem happens when the brush tool is selected and "Alt" key is pressed while Tool Option Bar is hidden. 
In this fix, an item "<custom>" is added to brush preset property at constructor in order to avoid crash at BrushTool::onPropertyChanged() because of failure to find any items in the preset property.


shun-iwasawa:
Jenkins, test this

shun-iwasawa:
Jenkins, test again

kogaki:
It also works in OSX. LGTM.

PR #201, opened 2016-04-19
Fix vector brush library

shun-iwasawa:
This fix is for issue #74
Two files in library/vector brushes folder seems to be corrupted.
I just translate slightly(0.01mm) to upper-right direction and overwrite them.
Modified files seems to work on both Windows and OS X in our environment.


skitaoka:
LGTM

PR #202, opened 2016-04-19
Fix crash when clicking on/off button in palette viewer

shun-iwasawa:
This fix is for the issue #75
It was mistake of usage of disconnect().


shun-iwasawa:
Jenkins, test this

tomosu:
It works on my OSX environment.
Thank you.

Issue #196, opened 2016-04-18
Crashes just once at clean install


kogaki:
With nightly build version, it crushes in very first run. It goes OK with second try. 

Following is a backtrace I faced at first run
```
  * frame #0: 0x0000000101379c3d libtnzcore.dylib`TOStream::TOStream(std::__1::shared_ptr<TOStream::Imp>) + 125
    frame #1: 0x0000000101379ec5 libtnzcore.dylib`TOStream::TOStream(std::__1::shared_ptr<TOStream::Imp>) + 21
    frame #2: 0x000000010137c4c2 libtnzcore.dylib`TOStream::child(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >) + 722
    frame #3: 0x0000000101a5480a libtoonzlib.dylib`TSceneProperties::saveData(TOStream&) const + 4074
    frame #4: 0x0000000101c0bfba libtoonzlib.dylib`TProject::save(TFilePath const&) + 5914
    frame #5: 0x0000000101c150b3 libtoonzlib.dylib`TProjectManager::createSandboxIfNeeded() + 163
    frame #6: 0x0000000100039857 OpenToonz_1.0`initToonzEnv() + 3319
    frame #7: 0x000000010003a2c1 OpenToonz_1.0`main + 2017
    frame #8: 0x00007fff8f7f25ad libdyld.dylib`start + 1
    frame #9: 0x00007fff8f7f25ad libdyld.dylib`start + 1
```

Rayek:
For me it crashes each time I try to start it.

t32bitsrv.exe stopped working.

Win 7 Pro 64bit, 48GB, Nvidia 590GTX

BrundleThwaite:
I will second that, it crashed at startup originally. But has given no problems after a I removed the older version.

kogaki:
Solved

PR #203, opened 2016-04-19
Fix crash at first run with clean install

skitaoka:
see https://github.com/opentoonz/opentoonz/issues/196

- add move constructors


skitaoka:
Jenkins, plz

kogaki:
Good. This prevent Toonz from crash in my environment.

Issue #162, opened 2016-04-12
OSX 10.11.3 – Crash on Save Scene / New Level


geoffkhli:
I've just downloaded OpenToonz (_1.0) for my OSX 10.11.3 (Capitan) and can't seem to get started – the program crashes and I get an error message when I try to save my scene or create a new Level.

Is anyone else having this problem?

isampera:
I have the exact same problem. Don't know what to do tho..

ullebulle:
I also have the same problem, but only on version 1.0.1, not in 1.0
I'm also on OS X 10.11.4

Harold19776:
ullebulle, where did you find version 1.0.1 for Mac? I see the update for Windows but I can't find the mac update...

kogaki:
This problem must be caused by absent of `lzodriver` in OSX. We fixed that in #205 . 
You can try fixed version from `nightly build`. If the problem still have been there, please reopen. Thanks.

ullebulle:
With the latest nightly build, this still happens, but only when creating a new "Toonz Raster Level"
OSX 10.11.5

kogaki:
Hmm.. it doesn't happen in my environment(10.11.3). Can you find `lzocompress` in `/Applications/OpenToonz_1.0.app/Contents/MacOS/`? We think the absent of this file is the cause of the problem.

ullebulle:
Yes, the lzocompress file is there
https://dl.dropboxusercontent.com/u/215022/OT/Sk%C3%A4rmavbild%202016-04-20%20kl.%2012.05.57.png

kogaki:
It looks OK, so there should be another bug... If you have installed xcode and lldb, printing backtrace can really help us understanding the problem. Anyway, thank you for reporting.

ullebulle:
Ok, I have Xcode installed but not lidb, tell me how to install, with mac port?, I gladly help
Skickat från min iPhone

kogaki:
Thanks! Just `xcode-select --install` in your terminal will do.  (see http://stackoverflow.com/questions/34680789/how-to-install-lldb-on-mac/34680858 )

And following is a step to get backtrace
1. run `lldb /Applications/OpenToonz_1.0.app/Contents/MacOS/OpenToonz_1.0` in terminal
2. `(lldb) run` in lldb
3. Let OpenToonz crash
4. `(lldb) bt` in lldb
5. backtrace will shown in lldb

If you share the backtrace, we can see which function went wrong.

[...]

PR #205, opened 2016-04-19
fix bug: lzodriver should be bundled

kogaki:
**lzo{compress/decompress}** are created but not bundled in Release. This PR fix CMAKE to include them in package. 

This bug must be the cause of #162 

from #15 , this broken if statement has been used. So builds `v1.0.1~` have this bug.


kogaki:
Jenkins

kogaki:
Windows is not affected, because `CMAKE_RUNTIME_OUTPUT_DIRECTORY` are set for WIN32.
https://github.com/opentoonz/opentoonz/blob/6c9449add3b5450ded88c2bd6ac07eea6fefd518/toonz/sources/CMakeLists.txt#L199-200

skitaoka:
LGTM

Issue #124, opened 2016-04-01
Bug: file > save level on Mesh levels does not work while the plastic tool is active


blurymind:
To reproduce this bug:
1. create a level
2. Go to the plastic tool and create a mesh
3. add some bones to the mesh
4. Select the newly created mesh level in the x-sheet, while you are in the Plastic tool
5. File>save level

Expected result: Saves the mesh level and the animations in it. 
Current result: 'Toonz can not save this level'
![meshlevels](https://cloud.githubusercontent.com/assets/6495061/14206009/db835e38-f807-11e5-8d68-24e8cc9f64e6.JPG)

Saving mesh levels while the plastic tool is active does not work.
If you select another tool however you can get the save file dialog, but when you actually do save, opentoonz crashes: https://github.com/opentoonz/opentoonz/issues/198

blurymind:
A user at the blender forum has reported that saving mesh levels does not work at all for him:
http://blenderartists.org/forum/showthread.php?394970-ToonzPremium-(2D-Animation-tool)-now-Open-Sourced&p=3037775&viewfull=1#post3037775

Rayek:
That would be me - I have tried saving the mesh, but to no avail. Blurymind, could you provide a step by step explanation how you save those mesh levels?

blurymind:
You are actually right. In this bug report you can at least get to the save dialog, but when you actually do try to save - opentoonz crashes altogether.

I filed a bug report - as it is much more serious.
https://github.com/opentoonz/opentoonz/issues/198

saving in opentoonz in general is broken at the moment.

blurymind:
This bug has been fixed by @shun-iwasawa

Issue #134, opened 2016-04-03
Bug: Viewport drawing - vector and raster levels attached to a mesh node get a black alpha


blurymind:
When you attach multiple levels to a mesh level, you get black color instead of alpha in the viewport:
![opentoonz-glithch](https://cloud.githubusercontent.com/assets/6495061/14235397/2835cdee-f9f5-11e5-8d11-f5dfe78f217b.JPG)

This applies to whether you attach a single level or many drawing levels to the mesh level. It is most noticeable when you actually get to see the alpha - on a drawing that is smaller than the mesh.

Fortunately this is only a viewport bug - when you render the scene the alpha renders properly and the black color is gone.

blurymind:
This is definitely a regression bug. The alpha worked on toonz harlequin:
![blackbgregressionmesh](https://cloud.githubusercontent.com/assets/6495061/14704927/124d2dd4-07af-11e6-8955-a9602e633c3c.JPG)

Rayek:
A workaround is to create your character in a sub-sheet. No issues in that case.

blurymind:
@rayek thank you for the workaround. I will try it out. Mighr also be a hint of whats causing the glitch

Rayek:
Unfortunately, in the latest build a subsheet with a mesh animation crashes TOonz instantly!

blurymind:
This was fixed in today's nightly build. The fix however introduced two new serious bugs:
https://github.com/opentoonz/opentoonz/issues/253
https://github.com/opentoonz/opentoonz/issues/254

PR #207, opened 2016-04-20
Fix saving mesh levels problem + Modify style sheet for mesh columns

shun-iwasawa:
This fix is for resolving the issue #124 
Also modifying style sheets suitable for displaying mesh columns in xsheet for each style.


shun-iwasawa:
Jenkins, test this

kogaki:
It works in my OSX. LGTM.

blurymind:
It still crashes on windows 7

kogaki:
@blurymind Sorry for confusing announce, but nightly build for windows was just updated just a few minutes ago. Could you try latest one? 

This is because build for windows takes longer than that for OSX.

blurymind:
@kogaki Ah I see. No worries. After reinstalling again I can confirm that mesh saving works on windows7 now!

kogaki:
Good!

blurymind:
@shun-iwasawa unfortunately this seems to introduce a new  bug:
Once you make a mesh, you can not exit the mesh level.
As a result opentoonz gets locked in the plastic tool completely.

to reproduce:
1. create a mesh level
2. click on any other non-mesh level
3. select the brush tool and try to draw on the canvas

blurymind:
To understand why this is going to be a problem, consider this approach:
https://github.com/opentoonz/opentoonz/issues/134
opentoonz-glithch
when you attach multiple vector levels to the same mesh level. With this regression now we wont be able to edit the shape of the eye level drawings and see the head at the same time in the camera stand view.

blurymind:
@shun-iwasawa appologies - this is not a regression - the previous versions of opentoonz had this issue too. It is not a bug , but a limitation of the software.

Rayek:
@Blurymind: the mesh-controlled level can still be edited when you click the thumbnail in the level strip.

blurymind:
@Rayek I found out about it a week ago, then totally forgot and thought this is a bug again.
It just feels counter-intuitive to no longer be able to access the current frame active drawing through level when there is a mesh applied. I guess that this limitation would stop some technical problems from occurring at least.

There are slight limitations:
- You cant draw directly on the deformed by a mesh level - while the deformation effect is visible
- if you have attached more than one levels to the same mesh, you can not see the others while you make changes to one of them.

I guess that when the level is warped by a mesh, it's coordinate space becomes warped, so if the user tries to edit it through the warp, the software has trouble translating where the cursor is

Rayek:
@blurymind As far as I can tell, once you apply a mesh, the drawing is converted to a texture (you can test this by moving the content: notice the black background surrounding the image). 

Here is a nice workflow:
- create your object/character which will be deformed later with a mesh. Use as many columns (layers) as you like.
- save all levels, and save the scene. For example, "character".
- then create a new scene, and select the frame you would like to import the character. Then choose :Load Level". 

This will load the entire character scene as a subsheet.in your x-sheet. It is also possible to just drag a scene from the file browser into your x-sheet frame(s).

- now apply a mesh, and start animating with mesh deformation.

- if you wish to make a colour change, or edit the character content, just right-mouse click on the column header title, and open the sub sheet. When you have finished with edits, close the sub sheet again.

This is a very powerful workflow. You can really tell that OT has "studio production" written all over it.

blurymind:
@Rayek the black background surrounding the mesh is a nasty regression bug. Toonz harlequin didnt have that problem. 
It's a bug - not a feature or design. 
I explained how it can hinder workflow in my bug report #134 
blackbgregressionmesh

Issue #126, opened 2016-04-02
Adding New Style Causes Crash


Martank:
I'm trying to learn Open Toonz and so far its been quite an adventure but to day I tried to add a new style without drawing anything and boom the app crashed. I'm loving the app and hopefully this helps the developers fix this issue.

odysseywestra:
Thanks for the bug report. Could you tells us which OS you were using when this crash happen? Plus could provide a 15-30 video illustrating the crash.

mkalte666:
I have encounterd a similar Problem on windows. 
Creating a new style works fine as long as there is already a Level Palette. I think it should not be possible to create a style without an active level.

**Reproduce**:
 * Open OpenToonz with new Scene
 * Create a level (Col1) and draw something 
 * Switch to InknPaint/PltEdit or bring up Palette
 * Rightclick there -> new style -- no crash

'til here its fine. now
 * go back and select an empty level
 * back to the Palette, right click -> new style
 * Profit (crashes with "stopped working" etc)

The exception thats thrown is a "null"-read. (... while reading at 0x0000000000000080).

**My System**: 
 * Windows 10
 * i7 2700k @ 3.5ghz
 * 8gb ram
 * nvidia gforce gtx 980ti
 * 4 monitors

OriginalAdric:
Verified that mkalte666's repro works on Windows 10. Can't dig into it more atm, but I've put [the callstack for the crash here](https://gist.github.com/OriginalAdric/fec376cd237757b599d01bb0497948b7) for investigation.

shun-iwasawa:
Fix merged in v1.0.2

PR #219, opened 2016-04-22
Fix crash with New Style command on empty palette

shun-iwasawa:
This fix is for the issue #126 
I prevent the "New Style" and the "New Page" commands from being displayed when there is no current palette.


shun-iwasawa:
Jenkins, 一応

PR #51, opened 2016-03-26
Edits to build in Linux

ideasman42:
Edits needed for compiling on linux.

- Audio using SDL.
- Quicktime is disabled (using proxy only).

PR #226, opened 2016-04-24
CMake: initial support for Linux

ideasman42:
Here are changes to CMake from  #51,

This could be further split, (swapping `.dylib` with `${CMAKE_SHARED_LIBRARY_SUFFIX}` could be a separate commit for eg).
However not sure its really needed.

The one drawback with this commit is it references files from #51 which are **not included in this PR**, not sure thats acceptable. They could be removed.
On the other hand, this is a migration step towards supporting Linux, which doesn't currently work. So some small PR's that eventually add support may be acceptable.

### DIFF
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -6,7 +6,7 @@ message("SDK Root:" ${SDKROOT})
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/../cmake")
 if(APPLE)
     set(THIRDPARTY_LIBS_HINTS "/usr/local/Cellar/" "/opt/include" ${SDKROOT})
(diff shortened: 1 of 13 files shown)


kogaki:

kogaki:

kogaki:
Jenkins, test this

kogaki:
Jenkins hit the following error while building in OSX.

```

kogaki:
Maybe `${CMAKE_SHARED_LIBRARY_SUFFIX}` includes `.` :)

ideasman42:
Corrected

kogaki:
Jenkins

janisozaur:
@ideasman42 are you still working on that?

ideasman42:
Will close this and instead split it up into smaller changes.

Issue #56, opened 2016-03-26
usability bug: Default Keyboard shortcuts missing - load keymap files?


blurymind:
I have noticed that there are no default keyboard shortcuts with the installation (windows). The user can assign keys, but no keys have been assigned by default. So in essence opentoonz does not yet have a default keymap!

It also does not seem to have a way of saving and loading keymap files - at least not exposed to the gui. We need that in place in order for the community to create such keymaps for everyone who is migrating from different traditional animation software.

Is that something that is hard to add? We need to decide on some sort of a default key map - so others can learn the software and take advantage of keyboard shortcuts without having to set each one up.

My biggest missing keys (I am coming from tvpaint and toonboom experience):
- A button modifier combination to rotate the light table (rotate while holding alt+spacebar for example, release keys to get back to your brush/other tool)
- 'Delete' key used on the drawing canvas does not do anything - so I can not delete control points of lines, I can not delete lines. In fact there is no way to manipulate control points with the control point editor tool. No way of merging them. The right click menu only lets you set it to be linear or non linear.
Is that a missing feature or there just isn't any key assigned to delete them? It might need to be split out as another issue if so.

odysseywestra:
You could create a spread sheet on Google Drive with all the available shortcuts and open it up for comment to the community. I've done with MyPaint and its going smooth so far.

As far implementing it, it may require a bit of work.

blurymind:
@odysseywestra  Good idea :)

Here is a link to an open spreadsheet document:
https://docs.google.com/spreadsheets/d/1-taUohoRrFXnaNUcey8fvu3LKaFmDToydxga-jqBJy0/edit?usp=sharing

I will start populating it, anyone can join in! :)

BillDStrong:
Would it be better to have several keymap sets available to choose from, based on other popular programs?

blurymind:
Kind of like blender does! Yes it will! :) 
But we need some sort of a menu to switch between them - and also to locate the file where they are saved.

odysseywestra:
I think for now we could have alternative keymaps available for download via the Wiki. At least until it has the nessesay options to switch between keyboard shortcut maps. 

Dose openToonz saves the custom keyboard shortcut map as a config file?

blurymind:
I have started making a custom keyboard layout ini file for the toonzStuff folder here:
https://github.com/opentoonz/opentoonz/issues/80

odysseywestra:
[...]

PR #236, opened 2016-04-25
Adding default keyboard shortcuts + Enable to delete default shortcuts

shun-iwasawa:
This fix is for the issue #56 
This modification consists of two changes as follows:
- Adding default keyboard shortcuts. It is based on Toonz Harlequin's one but also including blurymind's proposed key bindings.
- Bug fix: OpenToonz had a bug that the default shortcut key cannot be removed unless another key is assigned to such command. This commit fixes such problem.


shun-iwasawa:
Jenkins, test this

PR #243, opened 2016-04-26
Rewrite update check

ss23:
Before release, the update URL will need to be changed to an official one.

There are some other TODO's in the code, but at this point a review of the general direction would be appreciated.

Also, to make things simpler, there is no support for Qt4, only Qt5. Is this going to be a problem?

### DIFF
--- a/toonz/sources/include/toonzqt/updatechecker.h
+++ b/toonz/sources/include/toonzqt/updatechecker.h
@@ -1,5 +1,3 @@
-
-
 #ifndef UPDATE_CHECKER_H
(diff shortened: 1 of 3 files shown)


kogaki:
It is OK to remove codes for Qt4 compatibility. Your suggestion about version check is simple enough. We will confirm this suggestion. Thanks.

ss23:
Since it's static, the update check URL can just be the Github site too (opentoonz.github.io).

kogaki:
This commit has been merged with #290 . Thank you for your work.

Issue #221, opened 2016-04-22
function editor bug: typing an expression crashes opentoonz


blurymind:
on windows 7 -64 -

to reproduce:
1. create a vector toonz level
2. in the function editor, for interpolation select 'Expression'
3. type anything for the expression

opentoonz will crash:
![expressioncrashbug](https://cloud.githubusercontent.com/assets/6495061/14748741/6793436e-08b3-11e6-870f-ac0d77d0628e.PNG)

I am also attaching some files that might help diagnose the problem:
[bugreport-OT-expressionEditor.zip](https://github.com/opentoonz/opentoonz/files/232240/bugreport-OT-expressionEditor.zip)

shun-iwasawa:
Fix merged in v1.0.2

PR #246, opened 2016-04-26
Fix crash when typing an expression without selecting curve in function editor 

shun-iwasawa:
This fix is for the issue #221 
I disabled the curve type combo box when no curve is selected in order to avoid expression box being touched.


shun-iwasawa:
Jenkins, test this

kogaki:
Jenkins

skitaoka:
Jenkins

kogaki:
Jenkins

Issue #237, opened 2016-04-25
Using expressions on a rotation causes scene files to load Incorrectly 


BrundleThwaite:
I am trying to link a column to part of a plastic rig with an expression. The link works and but when you try to open the save file it says the file is corrupted.

![untitled-2](https://cloud.githubusercontent.com/assets/18170954/14792757/145c20b0-0b1c-11e6-919f-58d971c01770.png)
![untitled-3](https://cloud.githubusercontent.com/assets/18170954/14792758/1577d822-0b1c-11e6-93dd-f483c7ea286f.png)
![untitled-4](https://cloud.githubusercontent.com/assets/18170954/14792760/1709fcb0-0b1c-11e6-980b-66bcf3cb6513.png)
![untitled-5](https://cloud.githubusercontent.com/assets/18170954/14792766/19764198-0b1c-11e6-9fa5-2b4fa32dd39e.png)

blurymind:
wow, it wiped out all three levels.

Could it be related to https://github.com/opentoonz/opentoonz/issues/221 somehow

BrundleThwaite:
No I have expressions working in other scenes with no problems but they are on effects and being driven by pegs. 

This wipes out everything including column names and effects.

BrundleThwaite:
Also #221 is a crash. this is scene file on after it is reloaded.

BrundleThwaite:
When I comment out  this section of the ,tnz file the scene loads.

```
 <!--      <keyframes>
            <Ex>
              0 "vertex(2,\"Vertex __3\").angle"° 
            </Ex>
            <L>
              15 0 
            </L>
            </keyframes> -->

```

BrundleThwaite:
I have checked a little more and any expression on the rotation parameter will cause the scene to load incorrectly.

BrundleThwaite:
I am going to close this thread as this is now fixed.

PR #258, opened 2016-04-27
Fix saving/loading an angle unit (degree sign) used in expression

shun-iwasawa:
This fix is for the issue #237 
With this modification, the degree sign will be saved as "\u00b0" instead of  "°" in the scene file.
This fix will keep the compatibility between Toonz Harlequin's scene file.
Please note that this fix will not enable to load the scene file saved with the previous version of OpenToonz.


shun-iwasawa:
Jenkins

tomosu:
It works correctly.

Issue #222, opened 2016-04-23
Setting "Level Subsampling" to 0 causes a crash.


lewislong:
Settings > Level Settings > Subsampling.

Windows 10 64-bit, i7-3820 3.60GHz, 16gb ram, Radeon 7850. Running on a single monitor.

Rayek:
Confirmed!

Windows 7 64bit, 48gb, Nvidia 590GTX

shun-iwasawa:
The modification was merged and will be included in the next version.

PR #281, opened 2016-05-10
Fix crash when setting "Level Subsampling" to 0 

shun-iwasawa:
This fix is for the issue #222 
In this fix the minimum value of the "Level Subsampling" input box is set to 1.


shun-iwasawa:
Jenkins

skitaoka:
LGTM

PR #303, opened 2016-05-15
Add `#pragma once` to headers

skitaoka:
Reduce compile times for msvc.

cf. https://msdn.microsoft.com/en-us/library/4141z1cx(v=vs.120).aspx


skitaoka:
Jenkins

kogaki:
It mustn't cause side-effects. I merged it.

Issue #276, opened 2016-05-05
bug: exploding a sub-xsheet that is connected to a mesh node crashes open toonz


blurymind:
to reproduce the bug:

1.  create a scene with two levels and save it
2. Create a new scene and load the scene from step 1 by using the 'load level' command in the x-sheet
3. Using the plastic tool, create a mesh on the sub-xsheet 

4.in the schematic view - right click on the loaded subxsheet node and select 'explode' - i used 'bring only columns option'
![explodebug](https://cloud.githubusercontent.com/assets/6495061/15044734/76035edc-12ce-11e6-8912-e521eb128c8e.PNG)

OT will crash

Here are the log fifes created by the crash:
[subX-crash.zip](https://github.com/opentoonz/opentoonz/files/250455/subX-crash.zip)

This is perhaps related to a regression from harlequin - #264 which also causes a problem when we try this #218

kogaki:
This might be fixed

PR #305, opened 2016-05-16
Fix crash on exploding sub-xsheet

shun-iwasawa:
This fix is for the issue #276 
I also modified a bug regarding Open Group command in this PR.


shun-iwasawa:
Jenkins

kogaki:
We have confirmed that issue can be reproduced and this fixed it. LGTM.

Issue #69, opened 2016-03-27
Consider avoiding to store binaries in the source code repository


andreygursky:
Please consider removing all the prebuilt libraries/binaries/object files (*.exe, *.dll, *.lib, *.a, *.obj, *.o, ...) and avoid adding them in the future. Additionally I would suggest to rewrite the history in order to purge them from the repository history to ensure the git repository remains small and quick.

*.exe:

> thirdparty/lzo/2.03/dict.exe
> thirdparty/lzo/2.03/precomp.exe
> thirdparty/lzo/2.03/simple.exe
> thirdparty/lzo/2.03/tools/lzocompress.exe
> thirdparty/lzo/2.03/tools/lzodecompress.exe
> thirdparty/lzo/2.03/lzotest.exe
> thirdparty/lzo/2.03/testmini.exe
> thirdparty/lzo/2.03/lzopack.exe
> thirdparty/lzo/2.03/precomp2.exe
> thirdparty/Lz4/Lz4_131/vs2010/CMakeFiles/2.8.12.2/CompilerIdC/CompilerIdC.exe

*.dll:

> thirdparty/lzo/2.03/lzo2.dll
> thirdparty/glut/3.7.6/lib/glut64.dll
> thirdparty/glut/3.7.6/lib/glut32.dll
> thirdparty/glew/glew-1.9.0/bin/32bit/glew32mx.dll
> thirdparty/glew/glew-1.9.0/bin/32bit/glew32.dll
> thirdparty/glew/glew-1.9.0/bin/64bit/glew32mx.dll
> thirdparty/glew/glew-1.9.0/bin/64bit/glew32.dll

*.lib:

> thirdparty/superlu/SuperLU_32.d.lib
> thirdparty/superlu/SuperLU_4.1/lib/libsuperlu_4.0.lib
> thirdparty/superlu/SuperLU_64.d.lib
> thirdparty/superlu/SuperLU_64.lib
> thirdparty/superlu/SuperLU_32.lib
> thirdparty/tiff-4.0.3/lib/LibTIFF-4.0.3.lib
> thirdparty/tiff-4.0.3/lib/LibTIFF-4.0.3_64.lib
> thirdparty/tiff-4.0.3/lib/LibTIFF-4.0.3d.lib
> thirdparty/tiff-4.0.3/lib/LibTIFF-4.0.3_64d.lib
> thirdparty/lzo/2.03/LZO_lib/lzo2_64.lib
[...]

PR #307, opened 2016-05-16
Port to vs2015

JosefMeixner:
I have ported opentoonz to VS 2015. It will probably not make sense at this time to move to it, but there are two commits, the first are only source changes which should be portable and the second one are all the binaries. I only built the 64bit variants of the files (release and debug).

The source changes are rather small, I had to remove `cprintf` in openblas, VS 2015 for some reason didn't link with it defined, so now `printf` is used instead. The other change was in `toonz/sources/toonz/cleanuppopup.cpp`, Microsoft for some reason decided to add a second set of `count` functions into containers so bind didn't resolve it. I substituted it with a small lambda.

Building QT is rather straight forward, only the OpenGL-setting has to be changed to `desktop`, I had no luck with the default setting of `dynamic`, it caused collisions between GLEW and QtGL. And you need ICU (forget the web pages saying differently), UIC doesn't compile without.


janisozaur:
How about https://github.com/opentoonz/opentoonz/issues/69 ?

JosefMeixner:
I am absolutely for it, it takes ages to upload the built binaries and having them everywhere is terrible. But I don't think I am able to do this currently. I would prefer to wait until it is fixed here and then also fix my branch. I will comment in #69.

turtletooth:
Josef- 
This is somewhat off topic- but do you know how to get a debug setup working for OpenToonz in VS?  I have to build a release and try and crash it.  I can't seem to get the debugger set up right in VS for OpenToonz.

JosefMeixner:
I use the target `RelWithDebInfo`, that is a release target with debugging info. I always start opentoonz from VS and have no problem this way to use it or debug it.

turtletooth:
> I use the target RelWithDebInfo, that is a release target with debugging info. I always start opentoonz from VS and have no problem this way to use it or debug it.

Is that a CMAKE change I need to make or a flag I set in VS?  I'm still learning.

skitaoka:
This PR was merged by https://github.com/opentoonz/opentoonz/pull/544.

Issue #300, opened 2016-05-13
Regression from harlequin: Edit tool - Global Key option is missing - causing problems with the interpolation handles


blurymind:
Here is the curves editor in Toonz harlequin:
![handlevisibilityregression](https://cloud.githubusercontent.com/assets/6495061/15260924/632d311e-1951-11e6-9e09-be40617bd7e6.JPG)

Please note in the screenshot that all channels are toggled on, and the interpolation handles (ease in/out) are all visible!

And here it is in Toonz Harlequin:
![otnohandlesissue](https://cloud.githubusercontent.com/assets/6495061/15261011/dcc86d9a-1951-11e6-81aa-6b2104bd4781.JPG)
 
when all the channels are visible, the handles in the curves editor disappear. If you however expose only the channel with the keyframes, you get them back - in the curves editor. They are however missing in the x-sheet:
![handlesstillmissing](https://cloud.githubusercontent.com/assets/6495061/15261058/26a664a8-1952-11e6-953e-9de1b06a96c0.JPG)

Handles in the x-sheet never disappear in harlequin, and they are for some reason constantly vanishing in opentoonz. This might be related to: https://github.com/opentoonz/opentoonz/issues/298

blurymind:
Ok I understand why this happens now - When you use the Edit tool in harlequin to pose a level, It creates a new keyframe on all channels. It uses that via 'Global Key' - this is an option that is missing in opentoonz:
![globalkey](https://cloud.githubusercontent.com/assets/6495061/15261492/c75255c2-1954-11e6-957a-9533723a43bb.JPG)

blurymind:
@shun-iwasawa  please bring back 'global key' to the edit tool. It is convenient to cut-out animation workflow.
Global key was by default always enabled in toonz harlequin and is now a removed feature from open toonz. Without it the handles will constantly get lost.

I already filed another related bug report for the same tool:
#157 
Please revert changes to the tool, as it is causing all sorts of trouble in the keyframing department.

blurymind:
Thank you @Banbury

kogaki:
Thanks to @Banbury , this issue must have been fixed.

blurymind:
thank you for merging it :)

PR #308, opened 2016-05-16
Added checkbox 'Global Key' to edit tool. (#300)

Banbury:
The code for the global key option was still in place. The only thing missing
was the checkbox in the tool options.

It took me about 2 hours to find this. This would be so much easier, if we could get access to the original sources, before all those changes were made.

### DIFF
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -238,6 +238,8 @@ class ArrowToolOptionsBox : public ToolOptionsBox
 	ToolOptionCheckbox *m_lockEWCenterCheckbox;
 	ToolOptionCheckbox *m_lockNSCenterCheckbox;
 
+	ToolOptionCheckbox *m_globalKey;
+
 public:
 	ArrowToolOptionsBox(QWidget *parent, TTool *tool, TPropertyGroup *pg,
 						TFrameHandle *frameHandle, TObjectHandle *objHandle, TXsheetHandle *xshHandle,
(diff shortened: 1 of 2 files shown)


Rayek:
Many thanks for your efforts, Banbury. Much appreciated.

kogaki:
This will fix #300

blurymind:
looking forward to trying it in a nightly build?
 :)

kogaki:
Jenkins, test this

kogaki:
Jenkins, confirm this

kogaki:
ss0620

Thanks, it works fine!
But, in my opinion, it's a little bit confusing that `Global Key` and `Pick` looks connected. How about put `Global Key` after `Pick` ?

Banbury:
That's the position the checkbox has in the original Toonz (according to the screenshot). So, I'd leave it there, to make it easier for users to switch to Open Toonz.

blurymind:
I agree with Keisuke it will look better after pick. :) original toonz
users wont have trouble finding it a few inches to the right

kogaki:
There may be another option that we can put `Pick` at leftmost of the toolbar (as original Toonz). Either will be better than now.

Banbury:
I could make the space bigger between the two controls.

kogaki:
@Banbury Of course, it can help. I really don't stick to the way of the solution. I just hope less confusing UI. Thanks again for your contribution.

Banbury:
I added the space between Global Key and Pick. I think it's now obvious, that they are two separate options.

kogaki:
Jenkins, confirm this

kogaki:
It looks better now. Thank you.

PR #310, opened 2016-05-16
Quiet GCC -Wparentheses warning with assignment

ideasman42:
These warnings are helpful to avoid accidentally writing `==` as `=`.

A mistake which looks like is made in `dvdialog.cpp`, `DVGui::MsgBox`.
so I made this an alternate PR (now merged, see #311).


skitaoka:
Jenkins, test this.

skitaoka:
LGTM

PR #313, opened 2016-05-16
Cleanup opens settings not pane

JosefMeixner:
My scanner doesn't have a 64bit twain-driver so I restored the original cleanup dialog instead of the cleanup pane (which i broke by that change). Before I clean up I would like to know, how to go along in this regard. If something like this is added, should there be a configuration option or is there already?

Sorry for the messy branch, started from my working branch. I can delete and recreate the branch if necessary.


skitaoka:
Jenkins, test this.

kogaki:
Jenkins, test again. Sorry to be late

kogaki:
@JosefMeixner I feel it is unclear that which option you need to restore. All options in `Cleanup settings` looks irrelevant to any scan process. Please reopen this with more description. Thanks.

PR #314, opened 2016-05-16
Handle more then 3 peg holes

JosefMeixner:
I added support to use pegbars with more than three holes. Due to the poor availability of paper for traditional pegbars I use a puncher for a filing system with six round holes and a custom built pegbar.

The code assumed the pegbar specification to have exactly three holes. The auto center code didn't work when a peg hole was at the top border of the paper (I scanned a bit poorly) and when there are more then three holes the original code centered the page at the middle hole of the three best matching holes.

I think I could also get the code to work with only two peg holes, but have not tried yet. It might help people try out toonz (a pegbar can be easily created with a 3d-printer), but it would be also quite hard to draw and keep the pages aligned, so don't know if it is worth the try.

### DIFF
--- a/toonz/sources/toonzlib/autopos.cpp
+++ b/toonz/sources/toonzlib/autopos.cpp
@@ -628,10 +628,14 @@ visit_bw(int i, int j, int x, int y, int bit, UCHAR *byte),
 #endif
 	visit_gr8(int i, int j, int x, int y, UCHAR *pix),
 	visit_rgb(int i, int j, int x, int y, TPixel32 *pix),
-	stampa_dot(DOT *dot);
+	stampa_dot(DOT const*dot);
 
+//! \brief Find the best matching pegs
+//!
+//! The found pegs are in array dots. The function checks, which of those best fits the reference in
+//! reference. The three best matching dots are returned in parameters i, j, k.
 static int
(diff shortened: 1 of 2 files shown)


kogaki:
Jenkins, confirm this

toonz/sources/toonzlib/autopos.cpp
```diff
@@ -1576,53 +1599,49 @@ int get_image_rotation_and_center(const TRasterP &img, int strip_width,
...
-static int compare_dots(DOT dots[], int *ndots,
-						DOT reference[], int ref_dot)
+static int compare_dots(DOT const dots[], int ndots, DOT reference[], int ref_dot, int& i_ok, int& j_ok, int& k_ok)
 {
 	int found;
-	int toll, i_ok = 0, j_ok = 0, k_ok = 0;
+	int toll;
 	float tolld;
 	int i, j, k;
 	short *dot_ok = 0;
```
kogaki: It looks dot_ok should be `bool`

JosefMeixner:
Thank you, I missed that.

kogaki:
Jenkins, confirm this

kogaki:
OK. LGTM.

Issue #302, opened 2016-05-15
Nightly build 13/5 OSX crash


Gulwer:
Hi
The nightly build from 13/5 crashes when I choose the brush tool and try to paint. Here is the crash report. 

[Opentoonz_crash_15_5.docx](https://github.com/opentoonz/opentoonz/files/264900/Opentoonz_crash_15_5.docx)

edit: the latest nightly build16/5 just hangs at the splash screen.

kogaki:
Thanks, I will check right now.

Gulwer:
Hi
Thanks for the fix. Now the program opens but it still crashes

kogaki:
Yes, and now I have created fix #317.

Gulwer:
Hi again
The fix works fine, thanks Kogaki

PR #286, opened 2016-05-11
Remove macros: CASE, __OR, and DEFAULT

skitaoka:

PR #317, opened 2016-05-17
Fix creating empty level fails

kogaki:
## What's happening?
As mentioned in #302 , drawing something on startup goes crash.

## Why
That reaches `assert(false)` <https://github.com/opentoonz/opentoonz/pull/286/files#diff-803a7eb5b528b30962c55bf071b8be06R1892> .
It's because absent of `break` following last `case`.


kogaki:
Jenkins

kogaki:
Jenkins

PR #333, opened 2016-05-19
Fix to stop creating zero-size image

kogaki:
`round(inch x dpi)` can result in zero-size image. Image size should greater than `1x1`.
This can be partial problem of #162


kogaki:
Jenkins

toonz/sources/image/tzl/tiio_tzl.cpp
```diff
@@ -2116,7 +2116,6 @@ TImageP TImageReaderTzl::load14()
 			throw TException("Loading tlv: bad icon size.");
 		fread(&actualBuffSize, sizeof(TINT32), 1, chan);
 
-		assert(actualBuffSize > 0 && actualBuffSize <= (int)(iconLx * iconLx * sizeof(TPixelCM32)));
```
kogaki: This is duplication of line below

skitaoka:
LGTM

PR #315, opened 2016-05-17
Use QOpenGLWidget instead of QGLWidget

skitaoka:
* ImageViewer
* ~~Viewer~~ (unused)
* HexagonalColorWheel

QGLWidget is obsolete. cf. http://doc.qt.io/qt-5/qglwidget.html

PR #350, opened 2016-05-23
Revert "Use QOpenGLWidget instead of QGLWidget"

kogaki:
Reverts opentoonz/opentoonz#315.

In OSX environment, #315 caused fatal error (`EXC_BAD_ACCESS`) at

```
inline void QOpenGLFunctions_3_0::glClearColor(GLfloat red, GLfloat green, GLfloat blue, GLfloat alpha)
{
    d_1_0_Core->ClearColor(red, green, blue, alpha);
```

So it must be reverted and we will wait merging it until the error will be fixed.


kogaki:
Jenkins

Issue #309, opened 2016-05-16
Failed to load PSD layer if it has Japanese name


shun-iwasawa:
When loading PSD file, if a layer name includes Japanese letter (such as "背景"), it fails to be loaded with error message "Layer ID not exists".

kogaki:
Does it happen for all not `ascii` characters ?

ullebulle:
This has happened to me too, several times. But in my case it has nothing to do with non ascII characters. OT can open psd files saved from photoshop but if the psd has been edited and saved in another graphical editor, I'll get the same "Layer ID not exist" - message when trying to open it in OT, maybe it has something to do with that?

kogaki:
OK, we now fixed for `Japanese Photoshop` version of psd file. So this issue will closed.
But there remain another problem for other languages or other text-encodings.

PR #352, opened 2016-05-23
Fix loading PSD layer with non-ascii name

shun-iwasawa:
This PR is for the issue #309 
Please note that I could modify only for Windows. For OS X the sources are left unchanged.


shun-iwasawa:
Jenkins

toonz/sources/image/psd/tiio_psd.cpp
```diff
@@ -40,7 +40,11 @@ TLevelReaderPsd::TLevelReaderPsd(const TFilePath &path)
 #ifdef REF_LAYER_BY_NAME
 		if (layerStr != "frames") {
 			TPSDParser psdparser(m_path);
+#ifdef _WIN32
+			m_layerId = psdparser.getLevelIdByName(layerStr.toLocal8Bit().constData());
```
kogaki: Why do you use `constData` instead of `toStdString` ?

kogaki:
@shun-iwasawa  This problem is not related to `local system` encoding but `psd file` encoding. I created alternative PR #355 . Plz review it. Thanks.

shun-iwasawa:
@kogaki did the better resolution in PR #355 
Thanks!

PR #354, opened 2016-05-23
Do not attempt to clone a palette when it is NULL.

JosefMeixner:
When a scanned strip was cleaned up and then without storing the scene cleaned up again and the query to overwrite the old files was answered with overwrite all, the program crashed.


skitaoka:
Jenkins, test this.

skitaoka:
LGTM

skitaoka:
merged, thanks.

PR #355, opened 2016-05-24
Fix to load .psd layers encoded in various codings

kogaki:
This PR is fix for #309 and alternation for #352


kogaki:
I haven't decided where we inject endings from.

kogaki:
Jenkins

shun-iwasawa:
LGTM. Clever solution!

Issue #334, opened 2016-05-19
gui bug: level strip width  is wrong when you start open toonz


blurymind:
Every time you open open toonz, if the starting room has a level strip - the level strip width will be set to a ridiculously wide value - not the one you had when you closed it last time.

To reproduce : 

1. Go to a room where the level strip is visible like this:
![levelstrip-1](https://cloud.githubusercontent.com/assets/6495061/15403384/a4466d52-1df0-11e6-9ca9-6fa6da5d0997.PNG)
2. quit open toonz
3. start open toonz again. When it loads at the same room, the level strip width will be doubled like this:
![levelstrip2](https://cloud.githubusercontent.com/assets/6495061/15403443/dd0f3196-1df0-11e6-93cc-3cda74c392db.PNG)
This is a slightly annoying bug - because it makes the user change the width.

I reported it at #80 as a comment, but I should have made a dedicated bug report for it - so as more developers notice it.

Rayek:
Yes, confirmed. Every time I start OT, it bugs me, and I kept thinking "I ought to report this as a bug", but I kept forgetting. Thanks Blurymind.

turtletooth:
Confirmed here too.

kogaki:
Thanks to @Banbury , this issue has been fixed

PR #357, opened 2016-05-24
Fix: level strip width is wrong when you start open toonz (#334)

Banbury:
On the creation the docking panel of the level strip was assigned a minimum
width. This has no effect, but making the docking panel too big, because the
size of the underlying controls hasn't been calculated at this point.
Removing the line fixes the issue with the level strip size, without affecting
the docking window otherwise.

### DIFF
--- a/toonz/sources/toonz/tpanels.cpp
+++ b/toonz/sources/toonz/tpanels.cpp
@@ -1106,7 +1106,6 @@ public:
 	{
 		Filmstrip *filmstrip = new Filmstrip(panel);
 		panel->setWidget(filmstrip);
-		panel->setMinimumWidth(filmstrip->width() + 2 * panel->getFloatingMargin());
 		panel->setIsMaximizable(false);
 	}
 } filmStripFactory;


kogaki:
How was it looks wrong ? Please attach the screenshot or tell us the way to reproduce the issue so that we can understand the issue well.

blurymind:
I filed a bug report here
https://github.com/opentoonz/opentoonz/issues/334

Banbury:
It says so right in the title :),

blurymind:
Yeah it does :b thank you for the fix! Hopefully it gets merged

kogaki:
Jenkins, confirm this

kogaki:
In my environment, OSX, I can't reproduce reported bug. And `MinimumWidth` works well without this commit. I will check the behavior in other environment. Please wait a while. Thanks.

Banbury:
On Windows the behaviour does exist. As it is, I prefer not having a minimum with, instead of the window resizing itself. I could make it compile for Windows only, though.

shun-iwasawa:
I still fail to reproduce the Issue #334 under my Windows environment.
I think I need more information about this problem.

@Banbury , do you mean that the incorrect (large) value is set at setMinimumWidth() on the creation of the Level Strip panel? If yes, is it possible for you to investigate the value of "filmstrip->width()" at setMinimumWidth()?

On my test here, the value of filmstrip->width() is 191 with the icon width of the Level Strip being set to 160 (= default value). 191 is exactly the same as the width of this widget - which is sum of the icon width, slider width and margin of FilmStripFrames.

Another one thing I would like to point is that the setMinimumWidth() seems to work properly as @blurymind mentioned in the issue #334 as follows:
 
>This is a slightly annoying bug - because it makes the user change the width.

Which means that user enable at least to resize the panel to the proper width afterward.
If the minimum width is set to incorrect value, this operation should be impossible.

Banbury:
After some further tests, I strongly believe this is an obscure bug in Qt. It's some kind of unwanted interaction between setMinimumSize() and docking the panel. It doesn't happen, when the panel is undocked.
All the calculations seem fine, including the minimum size. And FilmStrip is the only docking panel to use the minimum size.
My suggestion is to either remove the minimum size or add a maximum size, too. Both solve the problem.

turtletooth:
Confirmed here.  Windows 7 Pro.

blurymind:
Toonz harlequin does not have this bug, so comparing the source code might
help

blurymind:
In my test i am not resizing it to be wider and that is the point i made - it does that on its own accord when you start the software. The annoying part comes from the fact that you need to resize the width back to how it used to be before the unwanted change took place

kogaki:
OK. Now I understand that this can help some users. LGTM.

blurymind:
thank you ! :)

Issue #282, opened 2016-05-10
Feature request: Ability to see keyframes for Pinned Center


psychoanima:
When you animate with Inverse Kinematics Pinned Center keyframes for those pins are stored somewhere in the memory, outside XSheet. That means if you re-time something by moving keyframes, pinned center is still staying at the frame where it was originally created, there is no option to to see it/move it/delete it. There is only an option to reset all pins. It's frustrating.

blurymind:
That is a good one. Toonboom has it and allows to animate pivot

psychoanima:
Yes. And I really don't understand why guys from DigitalVideo hide keyframes for pinned center? It's a real mess to work with Inverse Kinematics without option to see on which keyframe pins are stored.

Is it hard to make this option available, anyone? :)

psychoanima:
Here's one visual idea how to show pinned center keyframes:

To make it clear, OpenToonz allows only one Pin Center per Column. With that in mind it's not a good idea to show Pin icon/keyframe in xsheet columns - If model is built from too many columns it can be frustrating to locate where is the keyframe for Pinned Center.

Why not then to show Pin gadget next to the onion skin?

Here's the video mockup I made: https://onedrive.live.com/redir?resid=F734AEC72AD52FD3!5508&authkey=!AOtBKAR4sUKfFws&ithint=video%2cmp4


Deal breaker is to show this gadget only when Inverse Kinematics menu is selected. That way it won't disturb ppl who are not using this technique.

And as I said, because only one pinned center can be activated per column it's a good place to locate it fast and to manipulate with it. Similar like with the onion skin.

kogaki:
#369 has fixed this issue. Thank you all for clarifying the idea.

PR #369, opened 2016-05-26
Show pinned center keys on Xsheet row area 

shun-iwasawa:
This PR is for the issue #282 
On the frames where the pinned center of the current skeleton is changed from the previous frame (= pinned center keyframes), blue square are displayed in the Xsheet row area.
Look and feel of this feature is based on @psychoanima 's proposal in the issue #282 but I made some changes as follows:
- This UI is displayed whenever the skeleton tool is selected, regardless of the current mode. 
- Index of the column which contains the pinned center object is displayed in the squares.
- Tooltip is set for this UI.

Since the row area is limited, this UI may obscure the frame numbers if the display mode of frame numbers is set to other than "Frame". This inconvenience might be another issue.

pinned_center_keys

### DIFF
--- a/toonz/sources/toonz/xshrowviewer.cpp
+++ b/toonz/sources/toonz/xshrowviewer.cpp
@@ -21,6 +21,11 @@
 #include "toutputproperties.h"
 #include "toonz/preferences.h"
 
+#include "tools/toolhandle.h"
+#include "tools/toolcommandids.h"
+#include "toonz/tstageobject.h"
+#include "toonz/tpinnedrangeset.h"
+
 #include <QPainter>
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
Jenkins

psychoanima:
Wooooooooow!!! <3

@shun-iwasawa 
Is it possible to move them up&down as a regular keyframes or for now they are just indicators?

shun-iwasawa:
@psychoanima 
For now they are just indicators and have no feature to adjust position of them.

kogaki:
OK, it works in my OSX. LGTM.

Issue #346, opened 2016-05-21
Bug: Character color on Memos turns white


Juls3000:
I found "Memos" very useful , between other things, as a help to syncronize sound with animation.

Now it has a bug:

(03) Case of you delete the first character:

![cats00](https://cloud.githubusercontent.com/assets/19145844/15447913/50107df6-1f51-11e6-9fde-3c8849c87baf.jpg)

(04) The characters color turn white:

![cats01](https://cloud.githubusercontent.com/assets/19145844/15447921/64858682-1f51-11e6-82e1-7532eca955b3.jpg)

BrundleThwaite:
I agree this is annoying and needs to be fixed.

There is a small workaround ( if you do not already know it :) ).
-open the style editor.
-then select the memo text and you should get arrow at the side:

![memo](https://cloud.githubusercontent.com/assets/18170954/15453383/249db378-2014-11e6-941a-2c350934e28c.png)

this will give you the text properties which will allow you to reapply the the colour with the style editor.

Juls3000:
Thank you !

shun-iwasawa:
Modification merged.

Juls3000:
Great thank you !

PR #371, opened 2016-05-26
Fix memo's text color

shun-iwasawa:
This PR is for the issue #346 
This was a problem of style sheets with white UI text color (i.e. gray_048 and gray_072 styles).


shun-iwasawa:
Jenkins

kogaki:
Yes, it works. 

FYI: To try this change, we need to remove old `qss` before installing this build (e.g. `rm -rf /Applications/OpenToonz/OpenToonz_1.0_stuff/config/qss/gray_*` in OSX).

Issue #375, opened 2016-05-27
Improve File Browser to refresh the current folder items automatically


shun-iwasawa:
For now, changes of the items (i.e. create/rename/delete of files) in the current folder are not reflected automatically to an item viewer panel of the file browser. You need to exit and re-enter the current folder in order to refresh the items, which is quite troublesome.

CodyEakins:
I'd almost rather have it so that OpenToonz utilizes an OS's native file explorer rather than relying on its current alternative... That way, people are used to it, and keyboard shortcuts & customization are available.

shun-iwasawa:
I understand that the users are familiar with OS's native file browser, but the File Browser panel has some advantages specialized for this software such as;
- uniting the sequential raster image files and display as one level,
- collecting the link to project folders under each project, etc.  
So, IMO, it would be better to keep using it and enhance the consistency with OS's native ones.

shun-iwasawa:
merged

PR #376, opened 2016-05-27
Refresh the current folder items of the File Browser automatically

shun-iwasawa:
This PR is for the issue #375 
I made mainly two modifications as follows:
- Monitor the current folder path with QFileSystemWatcher and refresh the item viewer panel  whenever the folder contents is modified or removed.
- Add a button for "Refresh Folder Tree" command on the buttons bar of the File Browser.


shun-iwasawa:
Jenkins

kogaki:
Wow! It works OK, and I feel it's very convenient. 👍

kogaki:
But.. `Refresh Folder Tree` button is not intuitive to me. Is it cost to watch `tree` folders and tell new children are created or not? At least, we can't understand it's function by only word `Refresh`.

shun-iwasawa:
I modified the points as follows:
- Making the folder tree to be monitored as well. Please note that only the expanded folders in the tree will be monitored in order to minimize the consumption of system resources. I tried to monitor 300 directories at the same time and it seemed not causing notable delay in my environment.
- Removing the "Refresh" button which I added in the previous commit. "Refresh Folder Tree" command may become unnecessary as the folder tree is now made to be updated automatically.

shun-iwasawa:
Jenkins

kogaki:
@shun-iwasawa Those changes make this UI more better. Thanks. It works nice in my OSX.

Issue #379, opened 2016-05-31
Edit mesh operation causes crash after save level 


tomosu:
1, Write raster image with brush and create mesh in plastictool.
2, Save level
3, Edit mesh vertex in plastictool
4, Crash

turtletooth:
Is this still an issue?  I can't recreate this.

tomosu:
This issue was solved in #380.
Thank you for reminding me about this.

PR #380, opened 2016-05-31
Keep the consistency of mesh image addr, after save operation

tomosu:
Change the mesh image to be isImageCompatible true, when it is already loaded. 
This PR is related to issue https://github.com/opentoonz/opentoonz/issues/379


tomosu:
Jenkins

kogaki:
LGTM

Issue #330, opened 2016-05-19
Default Project Settings


turtletooth:
Is there any interest in having the default project settings being standard high definition?  Right now the  default output settings of 768 x 576 don't seem very relevant.  Something like 1920 x 1080 seems like it would better reflect the needs of most users.

kogaki:
Thank you for your fix.

Issue #327, opened 2016-05-18
Enhancement - Add DCI_2K and Add DCI_4K to camera resolution presets


BrundleThwaite:
DCI_2K - 2048 - 1080 1:90:1
DCI_2K - 4086 - 2160 1:90:1

These formats should be added to the standard video presets. They are the standard for digital film cameras and DCI_2K is the minimum resolution for digital film in the admission requirements of the  Academy of Motion Picture Arts and Sciences.

http://www.oscars.org/sites/oscars/files/89aa_rule19_short_guidelines.pdf

turtletooth:
I'll do it.
Give me a day or two.

BrundleThwaite:
Thanks I appreciate it. Some day I will have to set up visual studio again and do these things myself. :)

turtletooth:
Pull request submitted.

kogaki:
Fixed by @turtletooth. Thanks.

BrundleThwaite:
Thanks

PR #385, opened 2016-06-01
Changed default resolution to 1920 x 1080

turtletooth:
The current default resolution doesn't reflect modern video.  1920 x 1080 is a much more standard resolution.  I also added the DCI 2K and DCI 4K resolutions to the resolution profiles.


kogaki:
Jenkins, confirm this.

turtletooth:
This should take care of #327 and #330

kogaki:
Jenkins, again

kogaki:
Confirmed. 
There's no reason to stick to `768 x 576`. Thanks!

PR #388, opened 2016-06-02
Change Scene BG to White

turtletooth:
While the default transparent scene background is good for advanced users.  Beginners to the program are often confused that the exports (on non-transparent types such as AVI or jpg) are black.  It is not apparent to most users why their exports don't work and many walk away thinking the program is broken.  
Advanced users should have no problem changing the setting back to transparent.


blurymind:
I am not a big fan of this default :)

1. White color backgrounds strain the eyes a lot while working. They emit much more light than grey. Even a novice will have the common sense of looking for a way to replace the white background for something that wont give them eye cancer after several hours of work

2. Most other animation or illustration software uses transparent background as default. There is a reason for that - in most cases people dont want to have a white bg in the final render. They either have a custom made background (level) or they render with transparency (sprite for game). So even a novice will most probably be looking for a way to replace the default white background you put here

blurymind:
Actually I am not sure if this change affects workflow or just the export result.
If it  only affects exporting - then I think it should be a good default.

AyameLee:
I see your point, but as far as I can tell it will adversely effect professionals. I'd rather it be taught to beginners how to set a scene up properly for export, perhaps in the way that Toon Boom is able.

It's incredibly easy using the default layout. Go to **XSheet** > **Scene Settings** or open the **Schematic Window** and toggle into **FX Schematic**, double click **XSheet** and change **Camera BG Color**, specifically the **M** (Matte?) value.

That's the only way I can see this being done, that would be a royal pain in the bum for me having to change it back to transparent every scene.

Perhaps instead of showing a default white BG in the **Preview Window**, change it to **transparent**?

t6

That way, people can at least have a eureka moment.

blurymind:
I think a better solution would be to have opentoonz render any non transparent type formats with a white background where the alpha is. That is what photoshop and most other graphic applications do anyway. Or select the color to render instead of alpha in cases where the format does not support transparency.

turtletooth:
> I think a better solution would be to have opentoonz render any non transparent type formats with a white background where the alpha is. That is what photoshop and most other graphic applications do anyway. Or select the color to render instead of alpha in cases where the format does not support transparency.

I think that is a much better solution.  I just don't know (yet) how to fix the black on export, but I will start digging..

kogaki:
@turtletooth FYI: I haven't tried ye, but you need to hook `MovieRenderer::Imp::saveFrame` (@see https://github.com/opentoonz/opentoonz/blob/12c444ee0dd7333f63166c1bfc8e173f40708cb9/toonz/sources/toonzlib/movierenderer.cpp#L381) to change BG color on rendering. But... `LevelUpdater` yet don't know the output format support alpha channels. So you need to change codes of `LevelUpdater` of all formats.

This change must be large one while BG color can be changed with UI. So I think we should tackle other critical problems.

walkerka:
AVI is actually not an opaque format. If you use "0000" fourCC, you can pass a 32 bit BMP as video format. The result is BGRA frames which contains transparency info. This is not good for final output but it's very useful if you want to do more processing (for example, import to After Effects).

RodneyBaker:
A post mortem follow up to this:

> a better solution would be to have opentoonz render any non transparent type formats with a white background where the alpha is

This approach was implemented in v1.3.
The Matte Color (via Preferences > Saving) defaults to white but can be changed by the user.
In this way image formats without an alpha channel aren't stuck with the common black background signaled via alpha channel.

As for why images tend to get black backgrounds when converting from RGBA to RGB (i.e. without alpha channel)... the value of black is 0,0,0 which is basically what 'no value' reverts to.

PR #389, opened 2016-06-02
Onion skin color

turtletooth:
Most animation programs use green and red for onion skin colors.  This helps the animator have a better visual cue what is happening.  

There is definitely room for debate, but I thought it might make a good default.


blurymind:
One default that bothers me in open toonz is the toolbar. When you first open OT, the toolbar always hides half of the items by default:
extraclick
Can you change it so it doesnt hide them when you start the software.
This forces the user to always have to make an extra click to get to half the tools on every restart of the software. It is a bad default to have imo

turtletooth:
I broke it- fixing now.

PR #394, opened 2016-06-02
Project settings drop down now working and added default folders for new project.

turtletooth:
Previously the project settings project chooser drop down did not work.  This enables you to choose different projects through the drop down now.

I also added default folder names for the project folders to save people time getting their new projects up and running.  Before, if a person just made a new project without filling in folder names, a palettes folder would not be made and therefore the project palettes option was missing from the Studio Palettes window.  

If a user wants to get custom names for their folder, it is no more work than before- just tab through the window.

I think Project Palettes are very useful but they aren't an option if the folder doesn't exist.

### DIFF
--- a/toonz/sources/toonz/projectpopup.cpp
+++ b/toonz/sources/toonz/projectpopup.cpp
@@ -347,7 +347,7 @@ ProjectPopup::ProjectPopup(bool isModal)
 			FileField *ff = new FileField(0, qName);
 			m_folderFlds.append(qMakePair(name, ff));
 			upperLayout->addWidget(new QLabel("+" + qName, this), i + 2, 0, Qt::AlignRight | Qt::AlignVCenter);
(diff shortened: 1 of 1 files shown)


kogaki:
Jenkins, confirm this

kogaki:
I will examine later, but this might cause the error on startup (even after clean install).
﻿

kogaki:
OK, it works in my environment. LGTM

turtletooth:
All of the changes I made only happen on the shop event for Project Settings or New Project.  I didn't actually make folders- I just pre-populated the text fields with folder names to save time and help make sure the palettes folder was made.

If it makes issues, I can fix it.

kogaki:
Ah yes, I faced the problem on startup with CI-builded package. But it works good with my own build. So I decided to test again with nightly build after merge :)
Anyway, thanks for adding convenient feature 👍

Issue #397, opened 2016-06-03
Randomly crashes after saving mesh on Mac OSX


tomosu:
Create and save meshes of a certain size, causes a crash randomly.

PR #398, opened 2016-06-03
Fix random crashes after saving meshes

tomosu:
This PR is related to https://github.com/opentoonz/opentoonz/issues/397
Legacy strstream library might have caused crashes after saving meshes.
Replaced strstream with stringstream in tstream.cpp.

### DIFF
--- a/toonz/sources/common/tstream/tstream.cpp
+++ b/toonz/sources/common/tstream/tstream.cpp
@@ -222,7 +222,7 @@ public:
 	ostream *m_os;
 	bool m_chanOwner;
 	bool m_compressed;
-	ostrstream m_ostrstream;
+        ostringstream m_ostringstream;
 
 	vector<std::string> m_tagStack;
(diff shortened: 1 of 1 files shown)


tomosu:
Jenkins

kogaki:
Could you explain more exact reason? What is the improvement from strstream to stringstream

tomosu:
After writing the large data to ostrstream and destruct it, OpenToonz goes unstable and crashes for various errors, like BAD_ACCESS or BAD_INSTRUCTION.
These errors seem to include invalid free operations of strstreambuf in ~ostrstream, but this cannot be always observed.
I tried replacing strstream with stringstream, OpenToonz no longer crashes when saving meshes, so I think ostrstream might have something wrong.

kogaki:
Hmm.. That seems strange behavior. Anyway, there's no doubt this improves code quality.  LGTM.

PR #403, opened 2016-06-04
Update toonzqt.ts

SaierMe:


skitaoka:
These commits ware merged by https://github.com/opentoonz/opentoonz/pull/461.

Issue #372, opened 2016-05-26
Option to retain original styles when loading color model's palette


shun-iwasawa:
When loading a color model, if the color model palette is different from the destination palette, 
for now there are two options as follows:
- Overwrite the destination palette with color model's palette.
- Keep the destination palette and apply it to the color model.

In many cases the color model’s palette is loaded to unpainted level with selecting the former option.
For the former option, the color model palette is expected to have the styles for lines with the same style indices as the ones in destination palette.

On the other hand, OpenToonz has a feature for loading raster image as a color model.
With this feature, color model palette is automatically created by picking the colors from the image.
Picking of the color is executed in scan-line order, which means that the arrangement of the styles may be all mixed up. Therefore, if you load raster image as a color model, the styles for lines in the destination palette may be replaced by some color used in area of color model.

In order to prevent such unwanted behavior, another option for loading the color model is needed which is;
- Add color model's palette to the destination palette

PR #407, opened 2016-06-06
New option to retain original styles when loading color model's palette

shun-iwasawa:
This PR is for the issue #372 
Three modifications were done as follows:
- A new option is added when loading the raster image as color model: "Add color model's palette to the destination palette" is for merging the two palettes instead of picking one of them. Note that this option will appear only when loading raster images. When loading levels with palette (i.e. TLV and PLI) as color model, this option will not appear.
- "Adding default palette" is now optional when converting the raster images into TLV.
- Japanese translations are added to some commands.


shun-iwasawa:
Jenkins

toonz/sources/toonz/convertpopup.cpp
```diff
@@ -988,9 +995,10 @@ void ConvertPopup::apply()
 	ConvertPopupBgColorG = (int)bgCol.g;
 	ConvertPopupBgColorB = (int)bgCol.b;
 	ConvertPopupBgColorA = (int)bgCol.m;
-	ConvertPopupSkipExisting = (int)m_skip->isChecked();
-	ConvertPopupRemoveDot = (int)m_removeDotBeforeFrameNumber->isChecked();
-	ConvertPopupSaveToNopaint = (int)m_saveBackupToNopaint->isChecked();
+	ConvertPopupSkipExisting = m_skip->isChecked() ? 1 : 0;
```
kogaki: 👍

kogaki:
I confirmed `Adding default palette` crate new tab `color model`. Looks OK

PR #421, opened 2016-06-08
Enable debug build

shun-iwasawa:
This PR is for enabling debug build with the current master branch.


shun-iwasawa:
Jenkins

skitaoka:
LGTM

turtletooth:
How do I use this?  When I choose debug I get a cannot find the file specified error on \opentoonz\build\x64\debug\ALL_BUILD

I would love to be able to debug but right now only can build, test, build, test . . .

turtletooth:
Apologies in advance for my ignorance on setting up debugging.

turtletooth:
Ok, I feel like an idiot because I did not have OpenToonz set as the startup project in VS.  But now I get an error that it can't find glut32.dll.  Any thoughts?

PR #431, opened 2016-06-10
Use stringstream instead of strstream

tomosu:
Replace deprecated std::strstream with std::stringstream.


tomosu:
Jenkins

kogaki:
LGTM. I hope this will reduce sudden crashes :)

Issue #81, opened 2016-03-28
Brush Tool: Stroke Stabilizer


blurymind:
Toonz brush tool currently has no way of stabilizing strokes. A brush stabilizer is a value that slows down the tracking of the stylus in order to make it easier for the user to create broad smooth strokes as he/she draws them.
![tumblr_inline_o1y6kedf9x1so46vz_540](https://cloud.githubusercontent.com/assets/6495061/14077148/b4b62936-f4e0-11e5-9786-6c922c2e81cd.png)

To show some examples of this, here is tvpaint's line smoothing:
http://www.tvpaint.com/doc/tvp11/index.php?id=lesson-first-steps-drawing-tools-line-smoothing
![scraps3_by_omni_science](https://cloud.githubusercontent.com/assets/6495061/14077215/53e7bdc6-f4e1-11e5-9499-7b924fccda47.gif)

Here is lazy nezumi - an application dedicated to stabilizing strokes in other applications that lack the feature:
http://lazynezumi.com/
![lnppressurespeedscaling](https://cloud.githubusercontent.com/assets/6495061/14077176/fd8bcac6-f4e0-11e5-8504-63db063dd653.gif)

Example of how line stabilization works:
https://youtu.be/3Hc1bexva70

Some open source examples of this - so you can potentially take code from them:

Inkscape - MASS value
https://inkscape.org/en/doc/tutorials/calligraphy/tutorial-calligraphy.en.html
> In physics, mass is what causes inertia; the higher the mass of the Inkscape calligraphy tool, the more it lags behind your mouse pointer and the more it smoothes out sharp turns and quick jerks in your stroke. By default this value is quite small (2) so that the tool is fast and responsive, but you can increase mass to get slower and smoother pen.
https://fossies.org/dox/inkscape-0.91/calligraphic-tool_8cpp.html

Krita - brush stabilizer - it has three types of stabilizing. Here is an example:
https://userbase.kde.org/Krita/Manual/Tools/FreehandBrushTool#Smoothing
https://www.youtube.com/watch?v=zRxyyFcAbRY
http://www.davidrevoy.com/article261/line-art-tips-with-krita
http://osdir.com/ml/kde-commits/2014-06/msg02993.html

walkerka:
The basic stabilizing is easy to implement but it's really hard to make it perfect. I have implement a simple smooth algorithm by averaging N samples of the stroke. It works but as the stroke points increase, the performance becomes worse. 

[...]

PR #432, opened 2016-06-10
[Done] Brush stroke stabilizer.

walkerka:
This is a WIP brush stroke stabilizer. Original from my GPU based anime editor.
See discussion in:[#81](https://github.com/opentoonz/opentoonz/issues/81)

Demo:

demo
[https://www.youtube.com/watch?v=yDV8ketAzCI&feature=youtu.be](https://www.youtube.com/watch?v=yDV8ketAzCI&feature=youtu.be)

Here's the original full realtime one. Users have better control but it relies on fast stroke rendering backend. I did some optimization to make it work better in OT.
[https://youtu.be/u1iAOpjN84E](https://youtu.be/u1iAOpjN84E)

The stabilizer logic is almost done but there're some issues:

1. I use "accuracy" value in tool option bar as smooth range for now. I also removed the extra smoothing step when stroke is done (by using a hardcoded accuracy value 100). I need to add a new control for the smooth range and leave accuracy logic unchanged.

2. It sometimes crashes when I finish the stroke. The cause is assert failure
In
[https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/common/tvectorimage/tstroke.cpp#L3578](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/common/tvectorimage/tstroke.cpp#L3578)

The assert fails because "cs_sign" is less than 0. 
  assert(!(cs_sign < 0) || acos(-sqrt(cs2)) > 10 * TConsts::pi_180);

It looks OT converts strokes to quadratic splines before rendering. I don't quite understand the algorithm of this function so if anyone is familar with it. Please tell me how to avoid the wrong input. 

BTW, after I removed this assert, OT still works well. I added log for this and it happened 4 times when I was drawing. Is this assert really important?

When it crashes, input parameters are:

cubic

    m_p0 {x=-96.292728088119091 y=-139.08880919302925 } TPointT
    m_p1 {x=-95.989899503066681 y=-137.57465707825921 } TPointT
    m_p2 {x=-95.687068583438844 y=-136.06050543039282 } TPointT
    m_p3 {x=-95.384237031318179 y=-134.54635390902470 } TPointT

error

2.0000000000000000 double


tomosu:
Jenkins

walkerka:
Added UI for smooth value control.
UI

I also reverted the changes to "Accuracy". Now it works together with the stabilizer.

walkerka:
Added supports for toonz raster levels. It doesn't include the "Raster Level" without "Toonz". Actually I don't know why there're two types of raster and they even don't share the same brush tool :(

Please note that even for toonz raster levels, brush tool has two different modes:
  1. When you set "Hardness" to 100 or check the "Pencil" option, it's in the "raster track" mode. 
  2. When "Hardness" is less than 100, it's in "blurred brush" mode. 

It looks the performance of second mode is a little poorer than first mode. So please test them separately.
demo

blurymind:
@walkerka this looks great! I suggested using a drop down list like paint tool sai's stabilizer as it takes less horizontal space - but this approach has a finer grain control and it is more consistent with the other sliders.

Have you managed to solve the crash issue?

walkerka:
@blurymind I prefer ClipStudio's UI more. It defines several smooth levels and user just need to click once to change the value. SAI's smooth level list is a little too long, not very efficient. 
OT's UI is dynamically generated from something like a config. I use the slider because it's easy to implement (mostly copied from "Accuracy" code).
ClipStudio

The crash has not been fixed yet. I just disabled the assert in my test. AFAIK, this assert is not really important. At least I didn't see any visual errors when the assert fails. So I suggest remove it. Can somebody verify this? @kogaki 

I also tried to add translations of smooth UI for Chinese & Japanese. Please help to complete French/German/Italian/Spanish (I know nothing about these ones).

blurymind:
Thank you for clarifying. This feature is going to be extremely useful. I know some animators who cant get good results without it on tvpaint. Especially traditional animators that start using a wacom tablet. The iron tool will have less use once this gets in.

blurymind:
@walkerka do you think the smooth range could be reduced to 0-50 instead of 0-100?
That way the slider will be easier to use with a tablet stylus - especially when switching between 10 and 20.

blurymind:
From the video demonstration- anything above 50 looks unusable anyways. We can get a twice more sensitive slider if the range is 0-50

walkerka:
@blurymind It makes sense. Actually I never use smooth value more than 50. The line quality is good enough between 10 to 20 while still keeps fine control. It's also a bad habit to rely on stabilizer as an artist. 50 is mainly for very long lines like long hairs.

kogaki:
Jenkins, confirm this

walkerka:
I still don't fully understand the computeQuadraticsFromCubic() function. Here're some articles about cubic to quadratic spline conversion algorithm, may be related.
[http://caffeineowl.com/graphics/2d/vectorial/cubic2quad01.html](http://caffeineowl.com/graphics/2d/vectorial/cubic2quad01.html)
[http://timotheegroleau.com/Flash/articles/cubic_bezier_in_flash.htm](http://timotheegroleau.com/Flash/articles/cubic_bezier_in_flash.htm)

blurymind:
@walkerka says "This branch has conflicts that must be resolved" 

@kogaki what needs to be done to get this merged?

kogaki:
We've started to employ clang-format from *v1.0.3*. So plz apply clang-format (with toonz/sources/.clang-format). We had focused on preparing for this new stable release this week. So we will verify this PR soon. Thanks for your continuous improvement.

walkerka:
@kogaki Clang-format applied.

kogaki:
Jenkins, confirm this.

kogaki:
OK, I confirmed this works on OSX, and very comprehensive code.

walkerka:
@kogaki Thanks for reviewing. Did you reproduce the crash? Should I disable the assert?

kogaki:
@walkerka Ah sorry, I missed that point because I coudn't reproduce the crash. But with looking at the code, I feel it must be a redundant assertion. Because following if statement will do almost the same and even more assertion in the if statement. And more, if statement seems works more correctly. That means split cubic bezier into more number of quadratic bezier until it reaches good estimation while assertion let program crashes too early.

Let's remove the stressful assertion :)

walkerka:
@kogaki Thanks. I removed the assertion. The feature is completed now.

BTW, do you know why there're two types of raster levels? I'm not going to implement stroke smoothing for "Raster Level" because the it looks "Toonz Raster Level" does the same job.

blurymind:
the color swapping technology in toonz raster comes with some limitations.
For that reason there is also a normal raster layer where there is no color
swapping, but the limitations with it are not present.

Thats my theory at least :) @rayek might know more

toonz/sources/translations/japanese/tnztools.ts
```diff
@@ -138,6 +138,10 @@
         <source>Pressure</source>
         <translation>圧力感知</translation>
     </message>
+    <message>
+        <source>Smooth:</source>
+        <translation>補正：</translation>
```
kogaki: `平滑化：` is better :)

walkerka:
@blurymind Ok. If users still use "Toonz Level", maybe it should be supported too.

blurymind:
in most cases I think users will do their inking on toonz vector layers - which you have covered already.
 :)

anyways, I really look forward to trying this. Toonz is getting better and better for actual drawing work. I think with this and some of the other incoming patches it will surpass harlequin

kogaki:
Thanks @blurymind , I didn't even know the actual difference of two raster levels...

kogaki:
Jenkns, let's try this

blurymind:
@kogaki - they both have pros and cons.

On toonz raster you can change the color palette of the image-each stroke or fill you put down can be later changed to another color effortlessly- in the same way you can on a vector layer . But it is a bitmap layer. This is an impressive feature that no other animation software that i know of has. It is very useful if you want to make retro style game sprites for games.
But that comes at a cost - you do not have opacity for the brush tool. Opacity is very important if you are doing a rough sketch. That is why i think there is also a normal raster layer.

On normal raster layer - changing a color will only change the color that you draw with. It will not update the color palette of your image. But you get a brush tool that has opacity! Pressure sensitive opacity. That makes this level type a much better choice for sketching a rough pass.

kogaki:
Jenkins, let's try this

nittolo:
@blurymind, I think Raster levels are mainly supposed to be used to hold images, because a .png or .tif image will load in as a Raster level. You can do sketching in Raster, but I don’t believe that was the original intention, as you can’t use fill, paint, finger, or tape tool on those levels like Toonz Raster levels. I think drawing in Raster levels is mostly supposed to be for filling gaps during the cleanup of scanned drawings. At least, that's my guess on the subject.

konero:
DigiCel FlipBook, Celsys Retas and TVPaint all have swappable color palettes with their bitmap layers the same way the Toonz Raster Layer has them, it's been around forever. The idea is if you needed to change a color for whatever reason, you wouldn't need to replace it on however many frames your scene is, just change it in the palette and you're done.

I also know a lot of colorists who use high saturation (often a color completely different from the final output color) just so they can be sure the area is filled without zooming or squinting, like a light grey or light brown you might use a vivid green, then simply swap it before render and it will update all frames automatically.

Toonz Raster also has the bonus of being able to use Color Trace Lines (Autopaint Lines) which is used extensively in the Japanim industry which vector or raster cannot use. This is the feature where you can draw your shadow areas in blue line, and have Toonz auto trace the line with a shade color.

e030a90fe8fe75249cd2183be76e6a2ee924ed4d_hq

If you got really good eyesight, you could use a zero size line for vector like in Harmony, but it's just not the same and you of course can't use that traditionally, Toonz and GTS (Retas TraceMan) can clean up and separate the lines automatically..

Unfortunately Toonz UI is terrible for this feature (see #433), so we currently still use PaintMan here. I'm aware Ghibli also used PaintMan instead of Toonz. IMHO it needs an overhaul else bitmap coloring is financially redundant in Toonz.

shun-iwasawa:
@konero
In Studio Ghibli we used Toonz for ink and paint work. But we did not used the Color Trace Lines (Autopaint Lines) for some reasons as follows:
- Including color information in the scanned image increases file size enormously.
- In the previous version, cleanup process sometimes fails to recognize colored line, especially at the intersection of red and blue pencils. Checking such error was troublesome.

So we decided to scan all drawings in black-and-white color and changed ink color at the Color Trace Lines by hand afterwards. Each scanned drawing was more than 5K x 3K pixels in size but file size could be minimized to 2-50KB in TIF format, CCITT Group4 Fax Encoding.
Regarding the issue #433, it is partially regression from Toonz Harlequin so I will tackle it.

Anyway, for this PR, LGTM

shun-iwasawa:
@walkera
Thank you for making such a great advance in the drawing feature of the software.
We'll be happy if you could contribute OpenToonz continuously and would like to assign you as a committer.
Let's improve OpenToonz together :-)

walkerka:
@shun-iwasawa Thanks you. Hope this feature makes tablet drawing easier and faster :)

blurymind:
This is fantastic. Thank you for adding this feature. I can't wait to give it a spin in the next nightly release! :tada:  :D

blurymind:
@shun-iwasawa  thank you for the insight on how ghibli's pipeline is set up.
This is incredibly interesting information. I did not know that in black and white images with such resolution can be reduced to such small size.

I took a look at the gts scan tool and it reminded me a lot of Firka - a scan software that my ex-boss and animation mentor used to capture frames from paper with.
Nowadays he has completely moved to digital and has told me that he sold his light tables, his capture camera and even the press he used to create peg holes on sheets of paper.
all of that is now replaced by a digital drawing tablet.

He still uses tvpaint though. I showed him opentoonz when it was released- while it had interface problems and that put him off it for a while as he couldnt do anything. But I might be able to get him to revisit it. I think that open toonz will greatly democratize the traditional animation pipeline. It is important that we support it's development and use. Contributions such as this one will bring it closer to the toolbox of many more animators!

One workflow that he uses a lot, and also I used on a production made with TvPaint - is described here:
https://github.com/mypaint/libmypaint/issues/17
The developer of Drawpile added this feature and it is extremely useful. Opentoonz has the underlying technology to pull it off already, but it is not taking advantage of it.

PR #437, opened 2016-06-11
Level Create Popup now sets default name and updates paths correctly

### LINKED ISSUE #391: feature request- creating a new level with no name should automatically give the level an automatic name instead of failing to create the level
This is something I notice that I am not the only one do. Staring opentoonz I create a new level, but forget to give it a name. Opentoonz fails to create the level, it closes the create level window and an annoying pop up tells me that I should name it. I have to start over.
If you look at people using OT at youtube - you will notice that they do the same thing when in hurry.

Here is a proposition. Instead of failing with a prompt, why not create the level and automatically name it to 'untitled_x' or 'Col_3' or whatever the column it was made on is. If the automatic name already exists, opentoonz could append and extra number to it's name automatically "Col_3_1".

newlevelcreate

This is similar to how a basic file manager works, or even chrome web browser when you save a file to a directory that already has the file with the same name.

turtletooth:
The level create popup now adds a starter name to the level and selects that box for quick editing.  The paths didn't always update consistently also.  Finally I added a setfocus to the name in the save dialog to save the user from having to click the name field first.

Demo:
https://youtu.be/MWFawEhj58A

Fixes #391


tomosu:
Jekins

skitaoka:
LGTM

PR #452, opened 2016-06-15
CMake: changes needed to build on Linux

ideasman42:
Note that these are the main changes from the linux port.
Not a fully working linux port.

Merging this will make updating from master less hassle.

### DIFF
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -6,7 +6,7 @@ message("SDK Root:" ${SDKROOT})
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/../cmake")
 if(APPLE)
     set(THIRDPARTY_LIBS_HINTS "/usr/local/Cellar/" "/opt/include" ${SDKROOT})
-elseif(WIN32)
+elseif(WIN32 OR UNIX)
     set(THIRDPARTY_LIBS_HINTS ${SDKROOT})
 endif()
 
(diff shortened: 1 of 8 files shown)


toonz/sources/CMakeLists.txt
```diff
@@ -198,13 +219,65 @@ else()
...
+    find_library(GL_LIB GL)
+    find_library(GLU_LIB GLU)
+    find_library(GLUT_LIB glut)
+    find_library(Z_LIB z)
+
+    find_package(TIFF REQUIRED)
+    set(TIFF_LIB ${TIFF_LIBRARY})
+    # the libraries have not .pc nor preferred Find*.cmake. use custom modules.
+    # find_package(PNG REQUIRED)
+    set(PNG_LIB ${SDKROOT}/libpng-1.6.21/.libs/libpng16.so)
```
janisozaur: Many distros ship current versions of libpng, can this be changed into a fallback only when system libpng is not found?

J5lx: If anything, there should be a normal option for it IMO. Autodetecting stuff like that and building a software differently based on the result is bad for packaging that software reproducibly (i.e. the packaging process will result in different packages depending on whether libpng is installed on the build system or not. So if some package maintainer is unaware of the autodetection he might create a package without the libpng dependency and someone with system libpng available will end up with a package with incomplete dependency information).

ideasman42: Updated it to use systems libpng.

@J5lx, CMake lists libraries used, package maintainers can pass them explicitly to CMake if they don't want to use auto-detection.

J5lx: Even so I believe it's better to be explicit about dependencies. But I guess it's a somewhat controversial topic after all (though far less controversial than code style or editors :wink:) so let's not make mountains out of molehills. I'll just try to find that list of libraries in the future.

skitaoka:
Jenkins, test this.

skitaoka:
LGTM

toonz/sources/CMakeLists.txt
ideasman42: Being explicit about dependencies within the CMake files can make bisecting older versions of the software difficult. (unless your project keeps an image of the platform used to build each version - some do).
In my experience its best to store exact configurations in a separate CMake file or build script.

PR #457, opened 2016-06-16
Enable Boost 1.61.0 by default

skitaoka:
- add path `boost/boost_1_61_0/` to CMakeLists.txt


skitaoka:
Jenkins

kogaki:
Anyway, it won't break existing build process. LGTM

Issue #449, opened 2016-06-15
Rename Text Color Bug


AyameLee:
Not really a bug, but the text in the rename field is unreadable because it is blue.

![tz_rename_bug](https://cloud.githubusercontent.com/assets/18118611/16085385/cee924ea-3313-11e6-8d0b-8864df0f4347.PNG)

Version 1.0.3 Stable

skitaoka:
This bug is fixed by https://github.com/opentoonz/opentoonz/pull/458.

Issue #442, opened 2016-06-13
Bug - After running a script or command in the script console the text goes black


BrundleThwaite:
If you run a script or a command the in the script console the input text then becomes black and it is very hard to make out what you are typing.

![script](https://cloud.githubusercontent.com/assets/18170954/16010371/33924852-3181-11e6-81d0-f401bbdca240.png)

BrundleThwaite:
Thanks

PR #458, opened 2016-06-16
Fix text color in the Script console and the Rename popup 

shun-iwasawa:
This PR is for the issues #442 and #449 .
I applied style sheet to both UI. For the Script console, I set the background color to light-gray regardless of the current style, and also set monospace font (Courier).


shun-iwasawa:
Jenkins

skitaoka:
LGTM

PR #444, opened 2016-06-14
Store the toolbar state (expanded / shrinked) in user's env file

jcome:
this PR is for #392

Issue #392, opened 2016-06-02
Show all tools by default


blurymind:
One default that bothers me in open toonz is the toolbar. When you first open OT, the toolbar always hides half of the items by default:
![extraclick](https://cloud.githubusercontent.com/assets/6495061/15740598/68df6ad0-28ad-11e6-8c1e-60d04f6273e1.PNG)
Can you change it so it doesnt hide them when you start the software.
This forces the user to always have to make an extra click to get to half the tools on every restart of the software. It is a bad default to have imo

skitaoka:
This feature is merged by https://github.com/opentoonz/opentoonz/pull/460.

PR #460, opened 2016-06-17
Update: Store the toolbar state (expanded / shrinked) in user's env file by jcome

shun-iwasawa:
This PR is for #392 and based on @jcome 's PR #444.

I just did a small change in order to simplify.
Thanks @jcome, it works fine.


shun-iwasawa:
Jenkins

skitaoka:
LGTM

Issue #370, opened 2016-05-26
Type Tool don't work in OSX


kogaki:
I can't type any characters with `Type Tool`. `Font` in tool option bar also don't work. There might be something wrong in loading fonts.

![2016-05-26 15 45 34](https://cloud.githubusercontent.com/assets/17926501/15565923/fe4581a4-2358-11e6-9e73-3a9d75aa3418.png)

kogaki:
OK, this is because `type tool` is programmed as 32bit feature. In windows we bundled 32bit while we don't in OSX.

artisteacher:
Is there going to be any resolution for this issue? I'm having my high school students use OpenToonz and we are using Macs. It would be better to remove the font tool from the menu if the functionality will not be added.

artisteacher:
Still an issue in 1.1.2. It would be really great to be ale to use text at some point.

artisteacher:
While the type tool can be selected, it is still unusable in OS X. The drop downs next to font and size do not open. Nothing happens when clicking & etc. @kogaki seemed to think that #463 would help...

PR #463, opened 2016-06-17
Linux support for tsound (SDL2) and tfont (Qt), from pr #51

ideasman42:
Linux support for tsound (SDL2) and tfont (Qt), from pr #51 

Note that these files don't have license headers, but neither do OSX and Windows files here.

Credit goes to François Revol for this work.

### DIFF
--- a/toonz/sources/common/tsound/tsound_sdl.cpp
+++ b/toonz/sources/common/tsound/tsound_sdl.cpp
@@ -1,0 +1,563 @@
+
+#include "tsound_t.h"
+#include "texception.h"
+#include "tthread.h"
+#include "tthreadmessage.h"
+
+#include <errno.h>
+#include <unistd.h>
+#include <queue>
+#include <set>
+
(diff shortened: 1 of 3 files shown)


kogaki:
Wow. This must solve another problem! In windows and OSX, we use depricated software for sound and font. You might solved them altogether. Really really thanks.

ideasman42:
Yes, both are portable, so once they're in, nothing against checking on using them from other platforms :)

kogaki:
#370 will be solved

kogaki:
@ideasman42 This is so great PR. If you don't mind, please divide this into two PRs, for `tsound` and for `tfont`. If you can do, we can merge this in less time.

ideasman42:
Done: #466 #467

artisteacher:
Will this functionality be added for OS X? :)

ideasman42:
What functionality exactly? Can't OSX play sounds?

PR #464, opened 2016-06-17
Simplify atomic code to rely on C++11-provided std::atomic

janisozaur:
This greatly simplifies atomic code by using `std::atomic` offered by C++11.

For reasons unknown to me, OpenToonz uses a post-increment operator on atomic values as pre-increment one (same for decrement) and does `typedef std::atomic<int> atomic_t`, while `TAtomicVar` tries to be `long`.

I changed the latter to use `std::atomic<long>` instead. Keep in mind `long` does not have to be of the same size across all platforms.

### DIFF
--- a/toonz/sources/include/tatomicvar.h
+++ b/toonz/sources/include/tatomicvar.h
@@ -15,7 +15,7 @@
 
 #include <atomic>
 
-typedef std::atomic<int> atomic_t;
+typedef std::atomic<long> atomic_t;
 
 static __inline__ void atomic_set(atomic_t *v, const int value) {
   v->store(value);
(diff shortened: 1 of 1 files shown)


janisozaur:
Please see this little snippet verifying how it works: https://ideone.com/fAFJ1J

skitaoka:
Jenkins, test this.

janisozaur:
I see the windows build has failed, but with no logs I'm not sure what is it about nor how to fix it/

skitaoka:
Thank you for your contributions. I cannot provide error messages now; because Jenkins, which is working in local environments, outputs those messages in Japanese. We will improve that by setting public CI environments to provide the log messages in English.

PR #474, opened 2016-06-19
Save settings updated - Updated and Fixed

turtletooth:
- Added a save all button to the close dialog (you can still just save the scene).
- Added a save all button to the changes dialog
- Autosave now saves all, not just the scene
- The save all menu command is now higher than the save scene command: it should be the first save that anyone sees.
- The save all command now uses Ctrl S, instead of save scene.

99% of the time a user is going to want to save all, not just save a scene, so this should be the easiest option for users to select.

saveall

savechanges


nittolo:
It's important to note that some people may actually want autosaving to only save the scene. The preference between Save Scene and Save All was discussed here: https://groups.google.com/forum/#!searchin/opentoonz_en/save$20all/opentoonz_en/Fwmuj_q5fX8/QKxVDDQrEQAJ. So I think automatic saving should have the option to Save All or Save Scene.

AyameLee:
I agree with @nittolo.

It would be better if the options were:

> **Autosave (enabled)**
> - [x] Scene
> - [ ] Levels
> - [ ] Palettes

That way users can choose which files get saved automatically.

Standard OS UI has the focus on the primary dialog box which is on the left most side, in this case I think it should be:

> The scene **sc_01** has been modified.
> Do you want to save your changes?
>
> **Yes** | No | Cancel

The current choices are a little unclear. TVPaint had this problem a couple years back which they rectified. The second dialog box should read:

> The following file(s) have been modified.
>
> A.tlv
> A.tpl
>
> Do you want to save your changes?
>
> **Yes** | No | Cancel

There is no need for a Quit button as it is implied by No. The Save All command will be implied by the fact Toonz is listing all files so it is obvious what Yes means. This is more clear and direct and inline with other professional software. I strongly feel Scene (the composite file) and all other files (assets) should be kept separately in a situation that invoked Close or Quit.

blurymind:
This is a very good addition. Thank you!

blurymind:
I agree with ayame

turtletooth:
Thanks all for the feedback.  I agree @nittolo that the option to autosave the scene only should be there.  @AyameLee I will see if I can also make an option to autosave the non-scene files only- that is not already built so that may take some work. 

@AyameLee I also agree that the save dialog boxes right now have confusing and inconsistent language, I will fix that.  I do **strongly** feel, however, that the very first option a user should be given when closing an unsaved project is to save all.  I can still make saving the scene and other files separate things, but I have watched a lot of youtube videos of people using OT and read alot of blogs for the purpose of seeing where people get most confused.  And saving ranks way up there.  It is not always intuituve right now.  I think most people will want to save all and that should be their first option.  

One idea that I had was- when a user clicks to close:
The current project has unsaved changes:
- Save all
- Choose what to save
- Quit
- Cancel

If the user clicks choose what to save, combine the current two dialogs into one:

Please choose what you want to save:
- Save scene only
- Save non-scene files only
- Save nothing
- Cancel

I am flexible on the wording but there needs to be a one click "Save all" option for most users.  I also think that being given four options may be confusing, so you can all help here.  However, the fact that a user can currently click yes to save the scene and then be presented with another box saying there are unsaved changes is confusing to many users who haven't quite wrapped their head around levels and scenes.  Please help me fix the wording and make it clear to users, but a one click option needs to be there for most users.

@blurymind I hear you about the file list.  It's a hard call.  Many new users don't even know what their tpl or tlv files are anyway, but more advanced users may like being able to see what exactly is unsaved.

BTW- if there are not any unsaved changes, it currently asks if you are sure you want to quit.  I think this is unnecessary and most other software doesn't have this prompt.  What do you think?

blurymind:
@turtletooth I never ran into the problem myself - so i think that having them shown is better than not having them shown. It's actually a very useful feature that it has. Warning what files will be lost is important indeed.

If there are no files going to be lost, perhaps the prompt is redundant.

turtletooth:
I have added options to the preferences, added a save all layers menu option and updated the dialog boxes:

autosaveprefs

savealllevels

saveexit

savelevelsquit

turtletooth:
also fixed #501 - Probably should have made that it's own PR.  But it was a small change.

blurymind:
I just lost some levels because the currently nightly build doesnt have this feature merged yet :)
Thank you for adding it.
Opentoonz really needs it at the moment - as it is extremely unstable. I get a crash very often and lose unsaved work - drawings. It discourages me from working in opentoonz.
With this PR autosave will also save any progress done on drawings.

blurymind:
As a related note- this is not as critical but would make the feature a little bit nicer:
https://github.com/opentoonz/opentoonz/issues/512

turtletooth:
I moved the changes for #501 to a new PR.

kogaki:
Jenkins

turtletooth:
Updated to master.

turtletooth:
Windows user can test this here: https://github.com/turtletooth/OpenToonzPortable/releases/tag/1.0.3-FeatureTest

turtletooth:
@kogaki or @shun-iwasawa Would either of you be willing to look at this?  I know that making saving more clear would benefit many folks- especially us teachers who are starting school again in about a month.

shun-iwasawa:
OK. I'll out of office this afternoon but I'll check this tomorrow.

toonz/sources/toonz/tapp.cpp
```diff
@@ -675,14 +675,25 @@ void TApp::autosave() {
 
   if (scene->isUntitled()) {
```
shun-iwasawa: If you would like to automatically save only levels, scene can be kept untitled and not needed to be saved.

toonz/sources/toonz/iocommand.cpp
```diff
@@ -1628,6 +1634,36 @@ bool IoCmd::saveAll() {
...
+//---------------------------------------------------------------------------
+
+bool IoCmd::saveNonSceneFiles() {
+	// try to save non scene files
+	// if anything is wrong, return false
+	bool result = true;
+
+	TApp *app = TApp::instance();
+	ToonzScene *scene = app->getCurrentScene()->getScene();
+	if (scene->isUntitled()) {
```
shun-iwasawa: .. So I think you don't need to check whether the scene is saved here.

turtletooth:
@shun-iwasawa I made the changes you suggested- good catch.  

In the saveNonSceneFiles I also found out that it was setting an isUntitled == true to isUntitled == false.  This caused the next autosave to save the scene in the temp directory.  So after saving the resources, if the scene was untitled to start, I set it back to untitled.  

There was also a failing assert in TApp::OnStopAutoSave() when you changed the minutes of autosave.  It would check to make sure that autosave was disabled, and it wouldn't be disabled if you only changed the minutes without disabling autosave.  I just commented out the assert to see if there was a solution you wanted there instead.  But the asset doesn't make sense since autosave can be stopped by disabling it in preferences and by changing the minutes between saves (which stops and then starts it again).

turtletooth:
@shun-iwasawa What are your thoughts on this feature?

shun-iwasawa:

turtletooth:
@shun-iwasawa Would you mind looking at this one next when you have the chance?

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -926,8 +951,11 @@ PreferencesPopup::PreferencesPopup()
...
   CheckBox *minimizeRasterMemoryCB =
       new CheckBox(tr("Minimize Raster Memory Fragmentation *"), this);
-  CheckBox *autoSaveCB = new CheckBox(tr("Save Automatically Every Minutes"));
-  m_minuteFld          = new DVGui::IntLineEdit(this, 15, 1, 60);
+  m_autoSaveCB      = new CheckBox(tr("Save Automatically Every Minutes"));
+  m_autoSaveSceneCB = new CheckBox(tr("Automatically Save the Scene File"));
+  m_autoSaveOtherFilesCB =
+      new CheckBox(tr("Automatically Save Non-Scene Files"));
+  m_minuteFld = new DVGui::IntLineEdit(this, 15, 1, 60);
   CheckBox *replaceAfterSaveLevelAsCB =
```
shun-iwasawa: Just like another modification done in the preferences popup ( #664 ), 
how about to use QGroupBox for this option like this?:

toonz/sources/toonzlib/preferences.cpp
```diff
@@ -255,6 +255,8 @@ Preferences::Preferences()
     , m_fitToFlipbookEnabled(false)
     , m_previewAlwaysOpenNewFlipEnabled(false)
     , m_autosaveEnabled(false)
+    , m_autosaveSceneEnabled(true)
+    , m_autosaveOtherFilesEnabled(true)
```
shun-iwasawa: IMO the initial value of `m_autosaveOtherFilesEnabled` should be `false` since this feature originally did not save non-scene files. Setting the initial value of this option to "enabled" may cause unexpected behavior for some users.

shun-iwasawa:
Jenkins

toonz/sources/toonzlib/preferences.cpp
turtletooth: I can see your point.  I enabled it by default because I would think that most people would want to autosave everything.  I think the number of folks would would only want to save the scene is significantly smaller than those who want to save everything.  I will make the change, but I think there is room for discussion around this.

turtletooth:
I added the group box with a slight change:  I think having the checkbox in the middle of the group box title is less intuitive than having it in the box.  I moved it inside the box:

autosavegroupbox

I also changed the default setting for non-scene files.  I still think it should be on- many new users have complained about saving their work and losing it because they only saved their scene.

Thanks for the suggestion about the group box- I definitely think it looks better.

What are your thoughts on autosave being enabled by default eventually?

konero:
Actually the checkbox in the QGroupBox label area would be the right location for it, I'm sure it has a pen name but I can't think of it right now.

**Read here:** http://doc.qt.io/qt-5/qtwidgets-widgets-groupbox-example.html

Basically if all options are going to be disabled when it is unchecked, then it should go in the label area. 👍 

groupbox-example

turtletooth:
@konero @shun-iwasawa Makes sense.  My bad.

turtletooth:
Fixed.

toonz/sources/toonzlib/preferences.cpp
```diff
@@ -255,6 +255,8 @@ Preferences::Preferences()
     , m_fitToFlipbookEnabled(false)
     , m_previewAlwaysOpenNewFlipEnabled(false)
     , m_autosaveEnabled(false)
+    , m_autosaveSceneEnabled(true)
+    , m_autosaveOtherFilesEnabled(false)
```
shun-iwasawa: It is quite understandable that most of users would prefer to save everything rather than to save only scene.
Sorry to trouble you but it is OK to set this initial value to `true`. I will add a note of caution to the next version's release notes regarding this change.

turtletooth: Updated

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. @turtletooth thank you for the improvement!

Issue #451, opened 2016-06-15
Shaders crash OT 1.0.3


Rayek:
Any of the effects found in the Shaders menu of the FX nodes crash OpenTOonz the instant the preview is activated.

blurymind:
Yeah i reported this earlier :)

blurymind:
https://github.com/opentoonz/opentoonz/issues/413

PR #495, opened 2016-06-22
(incompletely) Fix crash on using Shader Fx 

shun-iwasawa:
This PR is for the issue #451, but it is still incomplete.

In OpenToonz we introduced a new version of Qt (v.5.5 and 5.6) and it seems that the usage of OpenGL context becomes severe. In this PR I reluctantly revert some code to the old one - which uses obsolete class, but enables us to render the shader fx somehow... but only for single frame.

Please note this quick-fix is provisional and incomplete. The shader fx is still very unstable and it crashes with only rendering multi frames. It still needs to be overhauled.

### DIFF
--- a/toonz/sources/include/toonzqt/planeviewer.h
+++ b/toonz/sources/include/toonzqt/planeviewer.h
@@ -9,7 +9,7 @@
 #include "timage.h"
 
 // Qt includes
-#include <QOpenGLWidget>
+#include <QGLWidget>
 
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
Jenkins

skitaoka:
LGTM

skitaoka:
@shun-iwasawa Please make a new issue for completely fixing this.

PR #507, opened 2016-06-23
Fixes

janisozaur:


janisozaur:
`std::max_element`: http://ideone.com/SQV86b

skitaoka:
Jenkins, test this.

ideasman42:
These could be separate PR's, also would be good to use less vague title.

skitaoka:
I'm merging this, because there are no problems to implement functions; however I also prefer that single PR contains modifications for single issue.

PR #544, opened 2016-06-29
Support Visual Studio 2015

skitaoka:
https://github.com/opentoonz/opentoonz/pull/307

- fix conflicts
- support Visual Studio 2013 for CI
- use GIT-LFS for *.lib and *.dll


skitaoka:
Jenkins

janisozaur:
Before this gets merged, can you please *please* at least squash the changes? This will somewhat reduce amount of binaries needed to be stored in history, if you really have to store them in this repository (see https://github.com/opentoonz/opentoonz/issues/69). They are changed multiple times in this PR and each such binary has to be stored in git. This PR alone adds 22MB to the repository.

skitaoka:
Jenkins

ideasman42:
This PR looks to have changes in application logic (whats hole filling for example), that should probably be separate PR's.

Basically any fixes or improvements _not_ related to MSVC2015 support, should be applied separately.

ideasman42:
Regarding binary files, think it would be good if these used GIT-LFS or at least make `thirdparty` an external repository.
Long terms this will make the git repository large (everyone will store every binary library version for every platform).

skitaoka:
Jenkins, again.

kogaki:
Jenkins

kogaki:
using Git-LFS looks good idea.

turtletooth:
Thank you for this.

PR #548, opened 2016-06-30
WIP - help?  H264 encode via ffmpeg

turtletooth:
I am using ffmpeg to export h264 but I'm stuck.  
Right now, it looks like it exports, but the video file is almost empty- no video frames, just a few hundred bytes.  I think it is writing the header and the trailer, but the frames are empty.  Anyone have experience with ffmpeg or willing to take a look?  

Right now it is Windows only since I can't build ffmpeg for mac or linux.  The ffmpeg build uses openh264 instead of x264.  

You may need to rerun CMake to get VS to see the new files and move the windows/bin files to your OT root folder.

I've been working on this for a while, but the documentation for ffmpeg isn't great for getting started.

Thanks


turtletooth:
Once I figure out what I'm missing, getting mp4 import should be pretty straight forward.  That would be great for roto work.

turtletooth:
@kogaki Do you have the ability to compile ffmpeg on mac with h264 and webm?
@ideasman42  or @janisozaur - could you possibly do it for linux?  I will pull the binaries out of the commit.

kogaki:
@turtletooth I've tried this process (https://trac.ffmpeg.org/wiki/CompilationGuide/MacOSX), and no bloblem compiling. And more, you can compile ffmpeg with homebrew (`brew install ffmpeg --with-openh264`).

I think it's not good manner to include ffmpeg binary into this repository. Developers should use their own ffmpeg. For users, we can bundle them with OpenToonz while packaging.

turtletooth:
@kogaki I agree, I will pull the binaries out, but perhaps we could include links to binaries we have built to save other developers time.

kogaki:
I feel ffmpeg is a fundamental software (not only for Opentoonz), so it's not a extra burden for developers to prepare it. Especially on Linux or OSX, it's really easy to install them with package managers.

turtletooth:
I might add include a link for Windows folks only then.  Building wasn't straight forward (at least for me).

turtletooth:
Removing binaries and creating new PR

PR #552, opened 2016-06-30
Allow dynamic linking to SuperLU

### LINKED ISSUE #551: SuperLU should be used as a dynamic library instead of static
SuperLU, at least on Linux, is used as a static library, can it rather be used as a dynamic one?

janisozaur:
Fixes #551


ideasman42:
Committed, without the .4 suffix, closing.

PR #559, opened 2016-06-30
Fix checking the left-node before operating on the right

ideasman42:
This block would never run (`newLeftNode->m_concave` is checked directly before this check).


skitaoka:
Jenkins, test this.

skitaoka:
LGTM

PR #554, opened 2016-06-30
Add support for libpng<1.5

janisozaur:
original code taken from https://github.com/opentoonz/opentoonz/issues/140

PR #553, opened 2016-06-30
isnan fixes

janisozaur:
`isnan` was used without specifying namespace, yet it was coming from `<cmath>` header.

PR #563, opened 2016-07-01
Travis Linux job

janisozaur:
Adds Linux travis job.

Travis uses Ubuntu 12.04 by default, which does not have all the required packages, so I made it use a slightly newer version, 14.04. It still has improperly packaged `liblz4`, so I used a package from current LTS version, 16.04.

I also had to use external PPA, as Ubuntu 14.04 only has Qt 5.2 available.

There are two caveats with regards to `libtiff`:
* it will pick up `liblzma`, which is already installed on Ubuntu and link to it, but `libimage` is not aware of that, so I added optional dependency on `liblzma` and if found, it will be used to link `libimage`,
* similarly, `libjbig` will also get picked up by `libtiff`, but since it offers no `pkg-config` module, I simply disabled it for travis builds.

The only source changes are already contained in #553 and #554, so I think there's no need to test in on jenkins, I will rebase this PR once the other two are merged.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,20 +1,14 @@
 language: cpp
 sudo: required
 dist: trusty
-before_install:
-    - sudo add-apt-repository --yes ppa:beineri/opt-qt551-trusty
-    - sudo apt-get update
 install:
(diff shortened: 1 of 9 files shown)


skitaoka:
LGTM

Issue #562, opened 2016-07-01
New feature: Pencil Test


shun-iwasawa:
**What is Pencil Test?**
[Wikipedia - Traditional Animation](https://en.wikipedia.org/wiki/Traditional_animation#Pencil_test)

> While working on a scene, a key animator will usually prepare a **pencil test** of the scene. A **pencil test** is a preliminary version of the final animated scene; the pencil drawings are quickly photographed or scanned and synced with the necessary soundtracks. This allows the animation to be reviewed and improved upon before passing the work on to his assistant animators, who will go add details and some of the missing frames in the scene.

In paper-based animation workflow, the pencil testing is essential for animators during their work. In Japanese animation industry, they had used NAC's Quick Action Recorder (a.k.a QAR) machine released in 1980's, then transitioned to the software Quick Checker by Celsys (maybe released in 1999).

The pencil testing feature is quite simple - capturing animation drawings via camera and place them into xsheet as raster level (after some image adjustment). This feature will also be useful for creating clay animation with OpenToonz.

shun-iwasawa:
The feature prototype is almost done - like this:
<img src="https://cloud.githubusercontent.com/assets/17974955/16516901/a60a09a0-3fb5-11e6-906b-1d969691b0e9.png" width=600>

psychoanima:
Awesine feature! It will support only web cameras or dslr's as well? (like DragonFrame)

blurymind:
Could also be useful to stop motion animators.

turtletooth:
This is awesome!  I do stop motion with my students using DragonFrame but they always ask if there is a free software they can use at home.  Depending on how this works, I might use it instead of DragonFrame to keep all their 2D animation in one program.

turtletooth:
@psychoanima If this is only webcam for now, we could probably eventually add DSLR support in using libgphoto2.

Rayek:
This is awesome news! I had been investigating options to hook up a web camera, but so far I've found only indirect ways to do this. Really looking forward to this. I am hoping DSLR cameras will be supported as well at some point in the future.

turtletooth:
Merged

RodneyBaker:
Want you to know other folks are paying attention and testing these features/fixes as they are published.  I've successfully ran the new Pencil Test feature and... Looking good!

[...]

PR #575, opened 2016-07-04
Pencil test feature

shun-iwasawa:
This PR is for the issue #562
New Pencil Test feature is added. Please note you need to add a "Pencil Test..." command to your menu bar in advance. The command is in "Menu Commands">"Scan&Cleanup" category.

Regarding supported camera, for now I'm using Qt library (QCamera class) for detecting devices. I tested with Logicool HD Pro Webcam C910 and Elecom UCAM-DLM130H. They worked fine. For Microsoft LifeCam VX-7000, OT failed to detect it directly so that I needed to go through virtual webcam software such as [manycam](https://manycam.com/). I have not tested any DSLR camera.

### Choose camera and resolution
- Select camera from the "Camera" combo box.
- Select Resolution from the "Resolution" combo box.
- Then the finder image will be shown in the preview area.

### Setting file & level name
- Specify level name in "Name" input box by typing directly or click the right-arrow button which will input next available single-letter name automatically.
- Specify frame number in "Frame" input box. If the file already exists, overwrite confirmation dialog will open.
- Specify file type and save in folder path. Please note that the file will not be saved when you capture the images. You have to save the level afterwards.

### Adjust image
- Threshold value is used only for "black&white" color type.
- "BG reduction" is for reducing the unevenness of lighting (just like white balance). You need to capture white paper with "capture white BG" button before use this parameter.

### Onion skin
- Onion skin is available for stop motion animation. It displays the previously-captured frame with specified opacity.

### Capture
- Press "Capture" button, then the captured image will be placed in the current xsheet as raster level.
- If you specify the same level name as the current cell content, then the captured frame will be set in the same column. If the level is different, the new column will be inserted for it.

### Checking the frame containing more than one levels
- Use Preferences>Interface>"Show Raster Images Darken Blended in Camstand View" option so that you can see more than one non-transparent raster levels darken-blended both in the viewer and in the rendering result.


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins, check this again

shun-iwasawa:
Jenkins one more

shun-iwasawa:
Jenkins

shun-iwasawa:
@turtletooth 
I know you are super busy on implementing other features, but could you please kindly review briefly this PR?
As written in my previous comment, I used QCamera for handling cameras. I already noticed that it has a problem: Capturing to buffer is not currently supported on Windows (according to [this](http://lists.qt-project.org/pipermail/interest/2015-April/016302.html)). So I had no choice but to save the captured image to file (in $TOONZCACHEROOT) for every frame, which slows the frame rate awfully. I'm thinking of transferring to another library (such as libgphoto2 as you suggested) for the future improvement considering the speed issue and DSLR support. Thanks!

turtletooth:
I tried it today and it is great.  Truly wonderful.  It worked very well and I am so excited for this feature.  I would say that this is ready for pulling.  There are a few things I suggest at least for the future.

Keep in mind that I love this feature, and these suggestions are small compared to how awesome this is to have in the software.  I am also approaching this from a stop motion background.  I have never done a pencil test, but I do stop motion a lot.

Is there a way to add a maximize button to the window?  I found resizing the window to be something I had to do often. 

Can Capture be linked to a shortcut?  In stop motion, being able to capture with the enter key (DragonFrame) is very useful - perhaps a shortcut only for when the Pencil Test window is open.

Can the resolution be set to its highest value by default instead of its lowest?  My camera started with 160 x 120 which isn't a very practical resolution.

One thing that DragonFrame does to keep quality high but playback smooth is capture two versions of the file.  For playback it uses a jpg and records a tif for the final product.  These are recorded to disk when the frame is captured.  This is nice in case the program crashes, the images aren't lost.

Do you think the name Pencil Test might be changed to Camera Capture?  There is a large stop motion community that would flock to this feature, and Pencil Test is a term that may not be as familiar.  

The ability to playback at least the captured images in the window would be very useful but probably difficult to make happen.

Overall I love it!  Thank you so much for this feature.  I have wanted an open source stop motion program for a long time.

Do you want the current version merged as it is right now or is there anything you think needs changed before merging?

turtletooth:
Regarding libgphoto2, it doesn't look like it currently supports webcams (http://www.gphoto.org/proj/libgphoto2/support.php).  DSLR support would be awesome, but I would hate to lose webcam support.  

Right now, the libraries that I know that support webcams are QCamera, GStreamer, and FFMPEG.  The current implementation doesn't see too slow here, but I don't know how it runs on slower hardware.

turtletooth:
I did get a crash just now when I had multiple captured levels and did a Ctrl R preview.  I will see if I can duplicate it.

turtletooth:
Saving and closing is slow.  Is there a way to either show the progress of saving (or even just a window that says "Saving, please wait") so that users don't think the program has stopped responding?  

Another solution would be to save the images as they are captured.

turtletooth:
This feature is really fun- thanks for implementing it.
penciltest

shun-iwasawa:
@turtletooth 
Thank you so much for your detailed feedback on this feature.
Your suggestions are very helpful and I will tackle on them.

If the source codes seem to have no problem in your eyes, could you please merge this PR?
Then I will post your suggestions here as a new issue for discussion. Thanks!

turtletooth:
lgtm

blurymind:
changing the resolution crashes opentoonz
[Temp.zip](https://github.com/opentoonz/opentoonz/files/355563/Temp.zip)

blurymind:
Another nice to have feature would be to have the ability to shoot a reference video with the webcamera (in realtime) straight from opentoonz.

charliemartinez:
Hi! I'm very interested in using opentoonz for animations with stop motion technic.

But the software does not recognize any camera: neither the one that is incorporated in my laptop nor a new one I just bought, Logitech C922. Both cameras work without problems in other applications such as OBS and official opentoonz (from the AUR repositories of Manjaro).

But now that I'm using Debian, I can only install Morevna.

I am using the OpenToonz-1.2.0.3-morevna-18.06.17-linux64-b8b54 version. The previous one version presented the same problem.
Did it happen to someone else? Is there a solution?

Thank you very much from Buenos Aires, Argentina and congratulations for this excellent work.

shun-iwasawa:
The camera capture feature is using Qt's [QCamera](http://doc.qt.io/qt-5/qcamera.html) class.
I'm not so familiar with Linux, but according to [this](https://bugreports.qt.io/browse/QTBUG-39798), someone advised as follows:

> It looks like some GStreamer plugins are missing. Are all of them installed on your machine? (For GStreamer 0.10) Make sure gst-plugins-good, gst-plugins-ugly and gst-plugins-bad are installed.

Also in the [Qt Manual](http://doc.qt.io/qt-5/linux-requirements.html#multimedia-dependencies) it says;

> To run applications that use Qt Multimedia, you might also need to install the following GStreamer plugins: 'good', 'ugly', 'bad', ffmpeg (0.10), and libav (1.x). These additional plugins contain various codecs for audio and video decoding, **as well as the necessary components for using the camera APIs**.

PR #578, opened 2016-07-05
Quiet warnings

ideasman42:
Trivial edits, quiet warnings with GCC

### DIFF
--- a/toonz/sources/common/tsystem/cpuextensions.cpp
+++ b/toonz/sources/common/tsystem/cpuextensions.cpp
@@ -52,8 +52,8 @@ long __declspec(naked) CPUCheckForExtensions() {
 		push	esi
 		push	ebx
 
-		xor		ebp,ebp			;cpu flags - if we don't have CPUID, we probably
-								;won't want to try FPU optimizations.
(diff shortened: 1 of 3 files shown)


toonz/sources/common/tsystem/cpuextensions.cpp
```diff
@@ -52,8 +52,8 @@ long __declspec(naked) CPUCheckForExtensions() {
 		push	esi
 		push	ebx
 
-		xor		ebp,ebp			;cpu flags - if we don't have CPUID, we probably
-								;won't want to try FPU optimizations.
+		xor		ebp,ebp			;cpu flags - if we do not have CPUID, we probably
+								;will not want to try FPU optimizations.
```
janisozaur: So on one hand opentoonz does not support 32 bit builds, but on the other it checks for cpuid and CPU presence?

ideasman42: Theres a lot of code that could be evaluated for removal, search for `__sgi`  and `SGI` for example.

While this can be done, its outside the scope of this patch.

skitaoka:
Jenkins, test this.

skitaoka:
LGTM

Issue #543, opened 2016-06-29
Regression Bug : Double clicking on a style thumbnail in the level palette window should open the style editor to change the color


blurymind:
Basically if you double click on a style now - a window pops up to change that style's name:
![stylename](https://cloud.githubusercontent.com/assets/6495061/16438333/c212a366-3da6-11e6-9c03-2700ccb671de.PNG)
Right now the level palette window does not even show the names of the different styles by default. Yet the default double click action is to change their names.

It would be much more useful if double clicking brought the style editing window.
That way I dont have to keep the style editing window open all the time. Instead I can get it to pop out whenever a color needs to be changed.

The style editing window is big and it takes a lot of screen estate.
It doesnt need to be visible all the time. I like to keep my layout in a way that gives more space for the drawing area.

@shun-iwasawa it makes much more sense than what we have now as the default double click command.

How about this for a design proposal:
![doubleclickproposal](https://cloud.githubusercontent.com/assets/6495061/16438459/d78d6fd6-3da7-11e6-9306-c7b94fc336f9.PNG)

- Double clicking on the style thumbnail (1) brings up the Style editor window (right now we have to right click > style editor) - thus double clicking on a color lets you change said color.
- Double clicking on the Style name (2) (when visible) brings up the rename style window - thus double clicking on a name lets you change said name.

That would be much more intuitive and at the same time useful than the way it is currently. Easier to change a color without having the style editor visible all the time that way.
Since this is a usability feature, perhaps @turtletooth might be interested? :)

turtletooth:
It seems like it could definitely be useful, but right now I'm up to my ears in h264.  ffmpeg is hard stuff.

blurymind:
@shun-iwasawa This is actually a regression bug from Toonz Harlequin.

I just checked - in Harlequin the behavior is the same as the design that I am proposing! :)
Clicking on a color thumbnail makes the style editor pop up to change the color. Clicking on a name of a color (when visible in list view mode) - lets you change the name of the color - without any extra pop ups.
![paletteregression](https://cloud.githubusercontent.com/assets/6495061/16438889/1b54bc62-3dab-11e6-9452-9c58a65bded0.PNG)

Why it is the way it is right now is beyond me. Seems like another broken feature.

blurymind:
@turtletooth It might be possible to get some help from other devs that use ffmpeg if you get stuck. I think a lot of non linear video editors use it for export.

blurymind:
[...]

Issue #428, opened 2016-06-09
Interface feature request - context sensitive Style editor


blurymind:
This is a feature that would make open toonz brush tool much less confusing to new users. It is similar to this one:
https://github.com/opentoonz/opentoonz/issues/343

@turtletooth might like it :)

Basically the style editor shows different tabs. Some do something only for the brush tool when used on a vector level, but nothing when used on a raster level. Other apply for both type of levels.
![hidenonapplicable](https://cloud.githubusercontent.com/assets/6495061/15944461/975ae4d8-2e85-11e6-908f-3dde149cbdf7.PNG)

My suggestion is to make it context sensitive. Show only tabs that have use for the current active level/brush tool mode. If a tab has no use and no effect on the brush in the current mode (vector or raster) simply do not show it!

blurymind:
Another note - 'Auto Apply' should be switched ON by default. This would take away an extra step between cause and effect for anyone first using the software. It used to be on by default in toonz harlequin and that made it more userfriendly to new users. I dont know why they made it OFF in OT.

turtletooth:
I totally agree on both points.  I have been bothered by having options that aren't valid for the current layer style and also think that auto apply should be on by default.  99% of the time, if a user is changing a color, they actually wan't it applied.

blurymind:
Another thing to note:
'Special', 'Custom' and 'Vector' brushes should all be in one tab i think.
I filed another bug report on that to keep discussion organised: 
https://github.com/opentoonz/opentoonz/issues/429

turtletooth:
@blurymind I was thinking about this when I was digging around with linked styles.  What would you do about palettes pulled in from the studio section?  If they have patterns or vector brushes but are pulled in to a raster type layer, the only option would be to delete them if they can't access the tabs to change the change the style.  I agree that having options visible on a layer type that doesn't support the feature isn't very user friendly, but I'm not sure the best approach for the rare case I mentioned.

turtletooth:
Based on the discussion in #1207 feature requests are going to be closed here on GitHub unless there is a developer actively working on the feature. Feature requests can be discussed at the OpenToonz Google Group: https://groups.google.com/forum/#!forum/opentoonz_en

turtletooth:
Done in #1397

blurymind:
@turtletooth while the changes are great, they dont seem to include making
it context sensitive. But they are changes in the right direction and
definitely declutter the interface!

By context sensitive I mean this:
[...]

PR #580, opened 2016-07-05
Revert behavior when double clicking on styles to open the style editor (and some fixes)

shun-iwasawa:
This PR is for the issue #543
- Now the style editor will open when double clicking the style chip other than the style name region.
- The name editor popup is still available, you can open it from the context menu of the palette.
- Text alignment of the rename field is changed to leftward when the "List" view mode.

--- and also contains some fixes as follows:
- Modified frame color of the style chip of the palette in the "List" view mode.
- The "Auto Apply" toggle state will be saved in user's env file so that it can reproduce the previous state on launching the software. Note that the initial state of this toggle is set to "ON" (= Auto Apply is activated) reflecting the comments by @blurymind and @turtletooth in the issue #428

### DIFF
--- a/toonz/sources/toonzlib/palettecontroller.cpp
+++ b/toonz/sources/toonzlib/palettecontroller.cpp
@@ -7,7 +7,12 @@
 #include "toonz/txshlevelhandle.h"
 #include "toonz/txshlevel.h"
 
+// TnzBase includes
+#include "tenv.h"
+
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
Jenkins

blurymind:
Thank you @shun-iwasawa   :+1:

turtletooth:
LGTM - Thank you @shun-iwasawa

PR #404, opened 2016-06-05
Maximize Pane where mouse is shortcut

turtletooth:
I added the ability to set a mouse shortcut to maximize the pane where the mouse is.  For example, if the mouse is in the xsheet, clicking ` (back tick) will make the xsheet maximized.  The same works for the scene viewer and the palettes. . .
This should take care of #335 

Edit-
I added the ability to maximize an entire room with Ctrl + `
This is #336 

These shortcuts can be changed of course

Demos:
https://youtu.be/80EF-2QG-gE

https://youtu.be/3ykJ-88BftI

Issue #336, opened 2016-05-19
feature request - full screen mode that does not change the room layout 


blurymind:
Full screen mode is something present in most drawing software nowadays.
What it does is when enabled, it full screens the application - so it uses the entire screen of the monitor. Any panels and window decorations that come with your operating system become invincible and no longer take screen estate.

This will in my opinion further improve the painting tools in open toonz. :)

Rayek:
Full screen mode is already possible: right-mouse click in the view, and choose "Full Screen". I assigned the F12 shortcut.

Or double-click on the title bar of the view, and the other panels will be hidden. Double-click again to switch back.

blurymind:
Ah awesome. Sorry i missed it. Closing

blurymind:
Ok it does full screen, but it hides all the panels. This is now what I meant. When going full screen - i want the software to preserve it's interface, just use all the screen estate of the monitor

turtletooth:
Demo: 
https://youtu.be/3ykJ-88BftI

blurymind:
This is perfect. Much better than what we currently have! Looking forward to getting it in the nightly builds :)

adamearle:
Yes Yes Yes is the hotkey by any chance the Tilda Key? Well done on
improving workflow on such a horrible interface IMO.

Personally I think why bring it up to the same standard as Harlequin when
Harlequin clearly needs a massive overhaul done its workflow. I get jumping
through the hoops and getting it one par.

But I am pretty sure if you asked the community what kind of timelines and
libraries they are wanting to deal with in terms of workflow; the  first
thing that spring to mind would not be Open Toonz  IMO

Maybe we can ask some questions to the world and hold a Kick starter and
get what we really want and some people can get paid.

Adam Earle

blurymind:
Well we need to first establish a developer or a developer team with proven knowledge of the code base, a list of goals and more than anything else- the time and will to do it. For now we can clearly see a number of active contributors who improve the current features immensely, while also getting to know the code base. Its not clear if dwango's devs work on it full time. There is no roadmap either. The best course of action really is to file bug reports and suggestions, make tutorials and videos - so as to place the software in more people's workflow. Stabilize it, iron out the rough edges and get the low hanging fruits. I already posted a couple of bigger feature requests, but they are not enough for a kickstarter campaign

blurymind:
[...]

Issue #335, opened 2016-05-19
Feature request - hit tab key to toggle only canvas mode


blurymind:
This is something present in most drawing applications nowadays.
The idea is basically when you hit the tab key, all other windows dissappear and you only see the canvas.

so for example when you work on this and draw something:
![levelstrip-1](https://cloud.githubusercontent.com/assets/6495061/15403641/ebb21744-1df1-11e6-9224-17b06bed4127.PNG)

hitting tab makes the entire window show only the canvas or the window that your mouse hovered over when you hit the button:
![capture](https://cloud.githubusercontent.com/assets/6495061/15403701/3001b774-1df2-11e6-848a-e35b7884d9b3.PNG)

Hitting tab again brings you back to the normal view.

That way the artist has more screen estate to work on the illustration and distracting GUI disappears!
It's especially useful for people who know all the keyboard shortcuts.

shun-iwasawa:
Double-clicking the title bar of the panel will enlarge it to entire window size.
(Double-clicking again will return to the normal view.)
Does it meet your demand, or you mean you would like to assign shortcut to such feature?

Rayek:
It would indeed be handy if we could assign a shortcut to this function.

blurymind:
A shortcut would be quicker and indeed nice to have. When using a stylus ,
double clicking doesnt feel natural

blurymind:
Is it possible to assign one? If so, i will close this

tofarley:
There is a shortcut setting for this. Under `Customize->Configure Shortcuts` in the `Visualization` section, `Show/Hide Full Screen`. The GUI doesn't accept tab as a shortcut (it tries to tab to another field) but you can set it to Ctrl+Tab or something similar.

blurymind:
The problem is that it also full screens it. Full screen and show only
canvas should be two separate functions-like it is in most other software

turtletooth:
Done.  Pull request 404

blurymind:
Thank you :) 
Now we only need a full screen mode without pane maximizing

turtletooth:
Demo:
https://youtu.be/80EF-2QG-gE

blurymind:
Please merge this. Its so much more comfortable to use this way!

turtletooth:
Merged #581

PR #581, opened 2016-07-05
Maximize Pane where mouse is shortcut by turtletooth (modified)

shun-iwasawa:
This PR is based on the PR #404 by @turtletooth . It remain basically unchanged from the original PR,  except small modifications as follows:
- Regarding the function MainWindow::maximizePanel(); changed to get cursor position (_QPoint p_) even when the panel is maximized in case of the mouse cursor being outside of the main window. Accordingly I removed the flag _panelMaximized_ which seems to be unused anymore.
- Regarding the function MainWindow::fullScreenWindow(); changed to call functions directly from MainWindow itself.
-  Changed the command name tentatively from "**Toggle Full Screen**" to "**Toggle Main Window's Full Screen Mode**" in order to distinguish from "Show/Hide Full Screen" command which is already available in the viewer and the flipbook. Please let me know if anyone have a good idea for the command name.
- Applied clang format.

The original comment for the PR #404 by @turtletooth is as follows:

> I added the ability to set a mouse shortcut to maximize the pane where the mouse is. For example, if the mouse is in the xsheet, clicking ` (back tick) will make the xsheet maximized. The same works for the scene viewer and the palettes. . .
> This should take care of #335
> 
> Edit-
> I added the ability to maximize an entire room with Ctrl + `
> This is #336
> 
> These shortcuts can be changed of course
> 
> Demos:
> https://youtu.be/80EF-2QG-gE
> 
> https://youtu.be/3ykJ-88BftI
> 

@turtletooth , thank you very much for the improvement!


shun-iwasawa:
Jenkins

shun-iwasawa:
@turtletooth 
Could you please review my modification whether it follows your original intent? Thanks.

turtletooth:
This looks good. "Toggle Main Window's Full Screen Mode" seems a little long.  What do you think about Toggle Program Full Screen or Toggle Full Screen Program.  
Another thing that could be considered (that I didn't think of until now) - should there be a way for users to discover this functionality other than browsing the keyboard shortcuts?  I realize that the menu structure is already pretty full, but it's worth considering.

blurymind:
~ seems like a good default keyboard shortcut. Its usually the TAB key in other software

turtletooth:
Right now, the default shortcut is the ~ key (same as the `).  I took this from Premiere which uses the same shortcut.  It looks like tab actually is used to tab around the interface.

blurymind:
` is a good default too! It is close to tab and understand if tab is reserved already.

I dont think in this case we need a menu entry for this command, but if there is one - where would you put it?  The mouse needs to be over the window that is going to be maximized

turtletooth:
Good point.  I'm not sure where a command would go.  Maybe someday once shortcuts are finalized, a shortcut reference can be made available.  

@blurymind Do you have any thoughts on the name of the command?  I think keeping them brief if possible is best, but still clear.

blurymind:
@turtletooth  How about:
 "Toggle maximize viewport"
"Maximize viewport"

"Toggle fullscreen area" (thats how Blender calls it)

toonz/sources/toonz/mainwindow.cpp
```diff
@@ -1878,12 +1895,15 @@ void MainWindow::defineActions() {
 
   createMenuWindowsAction(MI_ResetRoomLayout, tr("&Reset to Default Rooms"),
                           "");
-
+  createMenuWindowsAction(MI_MaximizePanel, tr("Toggle Maximize Panel"), "`");
+  createMenuWindowsAction(MI_FullScreenWindow,
+                          tr("Toggle Main Window's Full Screen Mode"),
```
turtletooth: What about changing the existing full screen to "Toggle Full Screen Canvas" so this can be "Toggle Full Screen"?

turtletooth:
Since the only thing that might change is the name of the command, but the implementation is good, I think it is good to merge.  If the name should be changed, it can happen later.
Thanks @shun-iwasawa for the PR.

blurymind:
good thinking. It will be easier to decide that way I think.

Issue #583, opened 2016-07-06
Toolbar and tool options bar too small in default rooms


turtletooth:
The basics room needs more space for the tool bar and the tool option bar- especially on smaller monitors.

turtletooth:
PR #584

PR #584, opened 2016-07-06
Updated rooms for consistency

turtletooth:
This provides more room for the tool bar and tool options bar.  Fix for #583


blurymind:
Can you make a screenshot :)

turtletooth:
Of course.

new_basics

And smaller: 
new_basics_small

Before, the tool option bar didn't go all the way across, which works fine for large displays, but on smaller displays, you lose visibility of some options.

In fact, I just checked the other rooms, and it's inconsistent.  I think I will update the other rooms for consistency.

blurymind:
It would be amazing if we could get more space for those sliders in the
options bar. Why are the drop down lists before "miter" eating up so much
horizontal space. This is madness xD

blurymind:
wastedhspace
should I file this as a bug report? :) @shun-iwasawa

blurymind:
https://github.com/opentoonz/opentoonz/issues/593
https://github.com/opentoonz/opentoonz/issues/594

This seems to be a persistent problem with the options toolbar

shun-iwasawa:
@blurymind 

> should I file this as a bug report? :) @shun-iwasawa 
You can almost fit another slider in the wasted (yellow space)

Please check if the style sheets in $TOONZCONFIG/qss are updated. Note that they will NOT be overwritten if they already exist (since they might be customized by users).

blurymind:
@shun-iwasawa thank you - that fixed it! :)

turtletooth:
- Made the tool option bar across the entire top.
- Made the tool bar fill the side
- Changed Pltedit to Palette
- Added Pencil Test to Scan & Cleanup menu

@shun-iwasawa The first time the program is loaded, the Pencil Test doesn't show up in the menu, but it shows up after that.  Why is that?

turtletooth:
> The first time the program is loaded, the Pencil Test doesn't show up in the menu, but it shows up after that. Why is that?

Figured it out for a different PR I'm working on.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

Issue #573, opened 2016-07-03
Crash on Save Level As for imported Level


turtletooth:
From Google Groups (Marcellis):

> Didn't see a report on this crash, so I thought I'd write it up.
> 
> Issue: Crash on Save Level As.
> Version: 1.0.3
> Repro rate: 100% (4/4, using two different Projects)
> Repro Steps:
> 1. Set Project
> 2. File/New Scene
> 3. File/Load Level, select numbered frame range
> 4. Play through twice (once to load into RAM, once to watch it play, just for kicks)
> 5. File/Save Scene As
> 6. File/Save Level As
> 
> Crashes as soon as I click save, after typing in the new Level name.
> 
> (Thanks to everyone who has made this open source software possible!)

I was able to duplicate the error and it caught on txshlevel.h line 81.  
`int getType() const { return m_type; }`

shun-iwasawa:
Confirmed on my Win7 environment.
I found that this crash can be reproduced more easily. Just select the empty cell in the xsheet and use "Save Level As" command.
Thank you for reporting the bug.

turtletooth:
Merged

PR #587, opened 2016-07-06
Fix crash on Save Level As when the current level is not selected

shun-iwasawa:
This PR is for the issue #573 
For Save Level As command, it is needed to check if the current level is  selected (just like Save Level command) BEFORE opening the popup. So I decided  to use an original MenuItemHandler rather than OpenPopupCommandHandler.


shun-iwasawa:
Jenkins

turtletooth:
LGTM

PR #589, opened 2016-07-06
WIP - H264 Export and Import - Working but buggy

turtletooth:
Here is what I have so far for mp4 export.  I think the header and trailer of the file is being written but not the frames.  I know that the swscale from rgb32 to yuv420p is working - I wrote the outframes to ppm files and read them in Photoshop.  Somewhere in the conversion to h264 and writing the frames, things aren't working. 

I may just have to forego this route and use QProcess to run the executables (trigger an OT export to png or something and use QProcess to convert the png sequence to mp4.  That would definitely be easier but not as built in as I had hoped to create.

If anyone wants to help tackle this, I would appreciate the help.  I'm pretty sure I can do the QProcess route easily but it wouldn't be the same process as the other export formats.


turtletooth:
@shun-iwasawa @kogaki I have the basic export working using QProcess, but I get a Stack Overflow on the progress bar if the animation is over a certain number of frames (I tried 50).  Short animations work, but my code needs looked at by someone who has been doing c++ longer than me.

kogaki:
OK, I will check it. 
At first, ffmpeg source might be removed from OpenToonz repo. It makes hard this PR to be reviewed.

turtletooth:
Removed ffmpeg sources.  The current implementation uses ffmpeg internally to convert from rgba to rbg and then uses ffmpeg through qprocess to convert the temp images to a video.

janisozaur:
@turtletooth by introducing a whole bunch of sources and then removing them *in the very same PR* you're just bloating the repository. Can you please rebase your changes so that unnecessary bits are not introduced in the first place?

turtletooth:
@janisozaur If ffmpeg implementation finds its way into OpenToonz, won't the bits will be necessary (eventually) for the developers?  I don't mind resubmitting a clean PR, but won't the ffmpeg headers be necessary for developers eventually?  There were not any binaries in my PR.  I'm not trying to be stubborn here, and still have plenty to learn.

janisozaur:
1. There's no need to re-submit PR, just rebase
2. *If* ffmpeg headers are needed t any point, they can be introduced then. I hope it does not ever happen (why would it? just your system-provided ffmpeg, it comes with headers already), but if it does, the headers you provided probably won't be reused (from git's point of view)

kogaki:
I also think we shouldn't use ffmpeg headers. We can just use ffmpeg as external process.

turtletooth:
@janisozaur @kogaki I'm pretty sure I can use only the external process for exporting image formats by using temp images, but how would you see using it for importing files?  If a user wants to import a mp4 or gif, the only way I can think of without using the headers and libraries would be to use QProcess to call ffmpeg to convert the file to an image format that OT can read.  This would create many temp files potentially cluttering the project folder.  

It's not a big deal to use temp files for export because you can clean them up right after ffmpeg converts to the desired output, but I don't know if it would work on import.

What are your thoughts?

turtletooth:
Rebased without headers.

kogaki:
Sorry if I misunderstand the problem. But I feel there's no big difference between the process importing and exporting. We can import images from video with temporal image file and clean them up after converting them into RasterLevel.

Why I strongly recommend you to use just a binary is related to patent and license. There're so many patents related to codecs and encoding techniques. If we need just a ffmpeg binary, users can get them from outside of OpenToonz (e.g. https://ffmpeg.org/download.html). I will be glad if you understand the difficulty of distributing such a sensitive software.

walkerka:
Well ffmpeg is quite difficult to use. There's little resource about encoding. It's much easier to use ffmpeg command line tool. If you decide to do that, you don't need to compile your special version. The official build (GPL) already includes x264, which has no problem in exporting mp4. 

Here's a working sample of h264 encoder. The fps looks wrong, but at least I can see the frames in the output file.
https://github.com/leixiaohua1020/simplest_ffmpeg_video_encoder/tree/master/simplest_ffmpeg_video_encoder

You can also export/import a patent free format (ogg/webm) as the intermediate format. Then do the conversion using ffmpeg for other formats.
https://en.wikipedia.org/wiki/Comparison_of_video_container_formats

turtletooth:
I was able to get import and export working using just QProcess and command line, but export crashes if the animation is long at all. (Over a couple seconds).  Import works, but playback is really slow.  It doesn't look like it caches images, just loads them each time.  

I could use some help here optimizing the performance and figuring out why it is crashing.  

@walkerka I still think we need non-generic version of ffmpeg.  OpenH264 needs to be used instead of x264 for licensing issues.  x264 is GPL and even linking to it from other software is probably not compliant with the OT license.  As for webm, once this is optimized, it will be no problem to change a few lines and have webm export and import working too.  

@kogaki At the time I wrote about import, I hadn't yet looked into how OT does importing.  I was able to do importing with QProcess.  

As for licensing, I think the decision needs to be made about what formats Dwango wants to support.  If the Dwango team decides not to use h264 at all, we can still use the code to create webm videos.  All that needs to be changed is a few lines.  This would also work for gif too.  I think there needs to be a build it small file size video format build in to OT.  This may not seem important to professionals who would export as an image sequence, but there are a lot of amateurs and teachers looking at OT and this option would be very useful

I think that distributing ffmpeg with OT would make life much easier for many folks instead of making them go download it and figure out which folder to put it in. But if we do that, it needs to be LGPL compliant (not x264) and mention probably needs to be included in the software somewhere. 

I understand that the licensing issues to consider here are significant, and I appreciate the fact that great care needs to be given when including external libraries.  Thank you considering this feature.

turtletooth:
@shun-iwasawa Perhaps you could look at this?  Your understanding of how OpenToonz handles images and playback could really be helpful.

turtletooth:
It seems to work when choosing single cpu for export.  
It crashes when loading a previously exported scene and trying to render.  Something in the levelreader is catching it.

kogaki:
I tried exporting mp4, and it works (although some frames were upside-down). Thank you for implementing this essential feature.

And, here are reasons why we think ffmpeg should be distributed outside from OpenToonz.
1. There are so many patents related to codecs and encoding techniques. We cannot distribute software which potentially has a risk of patent violation. 
2. We don't need to care which codecs can be used with OpenToonz from the both view point of license and patent while users got it from outside of OpenToonz. So there also be benefit for users that they can use various codecs easily.
3. Exporting video is a optional feature for OpenToonz because we tend to export images and use other software for encoding in professional scene.

toonz/sources/image/mp4/tiio_mp4.cpp
```diff
@@ -0,0 +1,470 @@
...
+	args << "-b";
+	args << QString::number(finalBitrate) + "k";
+	args << "-c:v";
+	args << "libopenh264";
+	args << m_path.getQString();
+
+	std::string outPath = m_path.getQString().toStdString();
+
+	//get directory for ffmpeg (ffmpeg binaries also need to be in the folder)
+	QString ffmpegPath = QDir::currentPath();
```
kogaki: It's great if we can customize the path of ffmpeg binary

toonz/sources/image/mp4/tiio_mp4.cpp
```diff
@@ -0,0 +1,470 @@
...
+	//for debugging directory
+	
+	std::string ffmpegstrpath = ffmpegPath.toStdString();
+
+	//write the file
+	createMp4.start(ffmpegPath + "/ffmpeg", args);
+	createMp4.waitForFinished(-1);
```
kogaki: return code might be checked

turtletooth:
@kogaki Thanks for the feedback.  I think the upside down frames might happen when more than one cpu core is used.  Did you render with all cores or just 1.  

I will work on making the location of ffmpeg changed by the user tonight.  I will also look into checking the return code.  

I totally understand your reasons for not wanting to distribute ffmpeg with OT.  I think it makes sense.  I will make sure that my implementation only calls external processes and allows the user to provide their own ffmpeg.

Quick question- I know that there is debate about calling a GPL process from non-GPL software and whether that is right.  Right now I have it built to only use the openh264 codec which requires and ffmpeg built with openh264.  In ffmpeg, we can output a mp4 file without telling ffmpeg which codec to use- it will just choose the one it thinks it should use.  That would allow users to use whichever version of ffmpeg they want, even if they are using a gpl enabled version.  Do you (and the Dwango team) think it should be written to not specifically need the openh264 codec (which would allow the use of any codec) or do you want to force the use of openh264?

@ideasman42 You have done a lot of work on Blender (GPL), do you have any insights on allowing the end user to use whichever version of ffmpeg they want and how that would be viewed by the GPL community?

kogaki:
@turtletooth Thank you for understanding. 
It seems OK to use any codec which users has *outside* of OpenToonz. So it might be useful for users to list all codecs which their ffmpeg has ability to use (Of course it's better to implemented with another PR). 

I'm so sorry for let you be confused about `openh264`. As you mentioned, it become less reasonable to use `openh264`. At first, we also think ffmpeg can be bundled with OpenToonz. But after discussing with our legal department, we found there're unignorable risks.

turtletooth:
@kogaki Thanks for the clarification.  I understand that there are legal risks that have to be considered when dealing with these situations.  

If it is ok with you, I will take the openh264 specific code out of this implementation, which makes it more flexible.  I think it might make sense to also close this PR and merge my progress here with the webm PR since the implementation is the same.

kogaki:
@turtletooth 
So both webm and mp4 exporting are merged into the PR like "ffmpeg integration" ? Sounds good!

I think there's no problem putting it into the "General" tab.  Of course it's OK to create new tab like "load/save" or "saving".  But I don't recommend "Extras" because it's less explanatory name.

turtletooth:
Closing this and merging h264 with the webm pr.

PR #598, opened 2016-07-07
CMake: Conditionally add warnings when supported

ideasman42:
This adds utility functions to conditionally add compiler flags (when they're supported by the compiler),

allowing warnings from newer compiler versions to be used without
causing a lot of noise for users of older compilers.

Using `C/CXX_WARNINGS` also makes it easier to configure warnings based on different compilers instead of platforms, Clang/GCC on OSX and Linux can use the same warnings for example.


skitaoka:
Jenkins, test this.

skitaoka:
LGTM

PR #600, opened 2016-07-08
CMake: Test for endian instead of guessing based on the platform

ideasman42:
Setting endian was making an arbitrary guess, this sets the value correctly.

### DIFF
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -312,6 +312,16 @@ elseif(UNIX)
 endif()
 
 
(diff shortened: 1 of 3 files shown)


skitaoka:
Jenkins, test this.

ideasman42:
Any info on why this fails for OSX? This is basic stuff that should work fine on all platforms.

janisozaur:
And OS X works on travis, the jenkins tests are super confusing as they sometimes seem random.

kogaki:
So, it's OK to be merged.

PR #602, opened 2016-07-08
Fix TNZ_MACHINE_CHANNEL_ORDER_MRGB check

janisozaur:


skitaoka:
Jenkins, test this.

janisozaur:
It appears Travis' OS X builds are under heavy load right now, which is why there's still no green from travis: https://www.traviscistatus.com/

It was green before I rebased it though.

skitaoka:
OK. LGTM

PR #604, opened 2016-07-08
CMake: differentiate system includes

ideasman42:
Use system includes where supported.

With GCC and Clang this passes includes as `-isystem` to the compiler.
This suppresses some warnings in system headers which are maintained externally.

### DIFF
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -142,6 +142,7 @@ find_package(Qt5 REQUIRED
 
 if(WIN32)
     include_directories(
+        SYSTEM
         ${SDKROOT}/glut/3.7.6/include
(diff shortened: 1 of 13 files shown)


kogaki:
Jenkins

janisozaur:
@ideasman42 just a minor hint: travis schedules a build on every push to a PR *and* to linked repository, https://github.com/opentoonz/opentoonz in this case. If you make a PR from your own repo, rather than from this one, you save travis a build (currently two jobs). As OpenToonz's OS X job takes around half an hour on a constrained farm, this would free up quite a lot of resources to other projects, while doing no harm to OpenToonz.

ideasman42:
@ janisozaur, wasn't aware of this, will do PR's from my own fork in future.

kogaki:
LGTM

PR #608, opened 2016-07-09
CMake: initialize build type

ideasman42:
Avoid using an empty build type.

While supported, it means some settings aren't well defined.
better to default to Release.


ideasman42:
Accidentally created the PR on this repo.

PR #610, opened 2016-07-09
Make functions static, ensure declarations match headers

ideasman42:
This patch used `-Wmissing-declarations` to show functions and symbols that had no declarations, and either make them static or declare in headers.

This helps avoid possibility that declarations and functions get out of sync. _(which can cause hard to track down crashes which may only happen on some configurations)_
And ensures all source files reference headers correctly.

This ensures functions declared in `extern "C"`, do so in the header too.
An error found in `calligraph.h` while writing this patch.

It helps with refactoring code, when its obvious that functions are only ever used locally, theres no need to double check if they're called from other files, or forget to and cause bugs.

The compiler may also do a better job optimizing the code it knows isn't used elsewhere _(inline static calls, fold branches when called with constant arguments)_.

This has been applied to `toonzlib`, to avoid making very large global changes.
If accepted, other modules could have similar changes made and `-Wmissing-declarations` added to CMake.

### DIFF
--- a/toonz/sources/include/toonz/tcolumnfx.h
+++ b/toonz/sources/include/toonz/tcolumnfx.h
@@ -281,4 +281,11 @@ private:
   TOutputFx &operator=(const TOutputFx &);
 };
 
+//*******************************************************************************************
+//    TColumnFx functions
+//*******************************************************************************************
+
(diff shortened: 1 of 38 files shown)


skitaoka:
Jenkins, test this.

skitaoka:
LGTM

PR #612, opened 2016-07-09
Fix use-after free on render

ideasman42:
m_info was used after being freed, see: [asan.txt](https://github.com/opentoonz/opentoonz/files/355651/asan.txt) for details.


skitaoka:
Jenkins, test this.

skitaoka:
LGTM

PR #613, opened 2016-07-09
Fix uninitialized variable use

ideasman42:
Both these variables are read before being initialized.

```
./toonz/sources/toonz/previewer.cpp:315:  if (m_subcamera && subCameraRect.getLx() > 0 && subCameraRect.getLy() > 0) {

```

```
./toonz/sources/tnztools/tooloptions.cpp:581:  if (splined != m_splined) m_splined = splined;
```


ideasman42:
Seems that the build stalled on OSX, but didn't fail.

kogaki:
Build restarted

kogaki:
Good refactoring. Thanks.

PR #615, opened 2016-07-09
Fix crash loading levels

ideasman42:
Effects Linux, but may apply to other systems too.

`--level->end()` resolved to an invalid pointer and crashes.

Simple to redo, load new levels and keep clicking on the "..", would crash after 2 clicks.


skitaoka:
Jenkins, test this.

skitaoka:
LGTM

PR #616, opened 2016-07-10
Caseless filepath optimize

ideasman42:
Even though this code is more verbose, this comparison function may be called lot when browsing large directories.

- Compare the strings in-place instead of duplicating to make a lower-case comparison.
- Keep the case-sensitive comparison as a fallback instead of comparing again.


skitaoka:
Jenkins, test this.

skitaoka:
LGTM

Issue #617, opened 2016-07-10
[BUG] Shortcut Window: List Doesn't Update After Using Remove


konero:
**Location**

Open **Configure Shortcuts** window and assign a shortcut to something, then try to remove that shortcut, the list does not update to reflect the change, a restart is required.

![t1](https://cloud.githubusercontent.com/assets/19820721/16713041/f86a15e4-4692-11e6-9afd-745bcf27ede7.PNG)

**Version**

[nightly_534c9edc8878537acdca714f81dc04ca0b1a2533](https://github.com/opentoonz/opentoonz_nightlies/releases/tag/nightly_534c9edc8878537acdca714f81dc04ca0b1a2533)

turtletooth:
Confirmed.  Submitted PR #620

shun-iwasawa:
Merged. @konero and @turtletooth thank you very much for reporting and modifying the bug!

PR #620, opened 2016-07-10
Fixed remove shortcut not updating, added focus when item selected

turtletooth:
Fix for #617

### DIFF
--- a/toonz/sources/toonz/shortcutpopup.cpp
+++ b/toonz/sources/toonz/shortcutpopup.cpp
@@ -49,10 +49,7 @@ public:
     text.remove("&");
     setText(0, text);
     QString shortcut = m_action->shortcut().toString();
-	if (shortcut != "")
-		setText(1, shortcut);
-	else
-		setText(1, "");
+	setText(1, shortcut);
   }
   QAction *getAction() const { return m_action; }
 };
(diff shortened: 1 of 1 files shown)


toonz/sources/toonz/shortcutpopup.cpp
```diff
@@ -49,8 +49,10 @@ class ShortcutItem final : public QTreeWidgetItem {
     text.remove("&");
     setText(0, text);
     QString shortcut = m_action->shortcut().toString();
-    if (shortcut != "")
-      setText(1, shortcut);
+	if (shortcut != "")
+		setText(1, shortcut);
+	else
+		setText(1, "");
```
yamisama: Unless I'm missing something really subtle here, you can simply replace the entire `if … else` thing with just `setText(1, shortcut);` since in your `else` branch, `shortcut` would logically always be equal to "" anyway, i.e. you don't have to discriminate between the two cases.

turtletooth: Good catch.  I fixed it.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. @turtletooth thank you very much for modification!

Issue #622, opened 2016-07-10
[feature] option to use png, webp or even jpg instead of tif - for raster levels - for smaller projects 


blurymind:
Ordinary raster levels in opentoonz are useful for rough sketching - because they are the only ones that allow pressure sensitivity to affect the opacity of a brush stroke.
As such - I dont consider them to be an option for export as much as any of the other levels. They are not a level type I would use as a final render.

In terms of size on the hard drive however - they are ironically the ones that take the most space! They are also consuming more memory than other levels if i understand correctly. Each individual frame is between 2 and 6 mb. When you have hundreds of these frames - it really adds up!

So that got me thinking - wouldnt it be useful to have the option when you create an ordinary raster level - to select file type for them?
Something that is better compressed - takes the least amount of space possible for each frame.

So what do you say - do you think it will be possible to optimize this level type somehow? 
Why does it have to be tif to begin with?

Ideally it should be webp:
https://developers.google.com/speed/webp/

> WebP is a modern image format that provides superior lossless and lossy compression for images on the web. Using WebP, webmasters and web developers can create smaller, richer images that make the web faster.
> 
> WebP lossless images are 26% smaller in size compared to PNGs. WebP lossy images are 25-34% smaller than comparable JPEG images at equivalent SSIM quality index.
> 
> Lossless WebP supports transparency (also known as alpha channel) at a cost of just 22% additional bytes. For cases when lossy RGB compression is acceptable, lossy WebP also supports transparency, typically providing 3× smaller file sizes compared to PNG.

blurymind:
If opentoonz can already  read image sequences in other better compressed image formats - why does creating a raster level in it use tif? There are other losless formats that are much much much smaller:
http://flif.info/

blurymind:
Think about the cases where you backup your project on dropbox - where you have a maximum backup space.
If you try to backup an opentoonz project - keeping the raster level sketch pass adds hundreds of MB and even several GB some times to your projects. You have to backup only the vector levels or delete your sketch history if its on an ordinary raster level.

yamisama:
> They are also consuming more memory than other levels if i understand correctly.

[...]

Issue #365, opened 2016-05-25
So I installed the nightly's build but now im getting error when trying to run it


dumplekin:
![what](https://cloud.githubusercontent.com/assets/19562205/15551491/3e484d48-2284-11e6-82fb-72e81da3039f.png)

turtletooth:
This is the same as #364- can they be merged?

dumplekin:
merge?

dumplekin:
oooh! Yes, sorry

psychoanima:
Go to **C:\Program Files\OpenToonz 1.0\srv\**  and locate t32bitsrv.exe Delete that file and start OpenToonz. It should work until someone fix that in code.

dumplekin:
Oh okay, thank you it, works now, but man, after solving one problem there's another, so it still crash when I try to clean it up :/

kogaki:
This is duplicated issue.

PR #624, opened 2016-07-11
FFmpeg Integration - mp4, webm, gif

turtletooth:
Gif and Webm export are working when 1 cpu is used for export.  I modified outputsettingspopup to disable other cpus when gif or webm is selected.  Obviously it would be better if it could use all, but it is inconsistent right now.

I submitted these both together because the implementation is basically the same(as is the h264 PR).  I can split them if you prefer, but anything that doesn't work on one will probably not work on the other.  The only difference is ffmpeg options.

I realize that h264 support is being debated due to licensing issues, but these options shouldn't have licensing issues.  

The code works but I tagged it WIP because I'm sure there are things I missed.  If you could look at it, I would appreciate it.


blurymind:
You are awesome! Thank you so much :D

blurymind:
gif import would be amazing too - but its probably a lot of work :) 

Still great progress on the ffmpeg front!! We finally have internet friendly output - out of the box

turtletooth:
Added an intermediate conversion to jpg to delete the raw image.  Without the jpg coversion, a 5 minute 24fps animation would take 57gb in the process.  It would be deleted after the final output, but a failed conversion would eat hard drive space.  With the jpeg conversion, it drops to around 1.4gb.  That is still freed after the final output, but is a much better figure to be dealing with.

turtletooth:
@blurymind If I made a portable version with this feature, would you test it out and try and break it?
Note: if you change the settings, the sliders go down to their lowest value if you don't change them.  I need to figure that out.  Right now, dragging it down and back up takes care of it.

blurymind:
@turtletooth of course i will test it. :) if it helps getting it merged

ideasman42:
Jpeg will give compression artifacts, especially for solid color. Why not use PNG as an interchange format? (compresses solid color well).

blurymind:
@turtletooth its also worth noting that webm supports transparency:
https://developers.google.com/web/updates/2013/07/Alpha-transparency-in-Chrome-video?hl=en

blurymind:
There is no other video format that i know of that also supports having an alpha channel :b its pretty coil

blurymind:
Google claims that webp has much better quality and compression than png. That is another format worth investigating

ideasman42:
PNG is lossless, so not sure what better quality means in that context.

ideasman42:
Note that its possible to avoid writing files at all, which avoids many files on disk and time to compress as one format then re-encode into another.

See:

- https://en.wikipedia.org/wiki/Frameserver
- Encode from a frameserver: https://wiki.blender.org/index.php/Doc_talk:2.6/Manual/Render/Output/Frameserver
- Frameserver implementation in C: https://developer.blender.org/diffusion/B/browse/master/source/blender/blenkernel/intern/writeframeserver.c

blurymind:
Ah sorry - i totally overlooked that point. I guess only advantage might be file size.

Note that webp supports both lossless and lossy
https://developers.google.com/speed/webp/gallery2

blurymind:
@ideasman42 that does seem better. @turtletooth was earlier asking for help on the ffmpeg exporter. Blender does use ffmpeg as well and it works great there

blurymind:
Aligorith wrote some notes on export with it a couple of years ago

http://aligorith.blogspot.co.uk/2014/01/blender-ffmpeg-notes-on-getting-usable.html?m=1

turtletooth:
@ideasman42 I chose jpg for file size, but I did choose the highest quality.  You bring up a good point about compression artifacts.  I will try it with png and compare file sizes and see how it turns out.  

As for a frameserver, that is probably the best long term solution.  However, it looks like it involves creating a network or socket between one program and the other.  I have never done anything with networking before and don't feel equipped to take that on just now.  It seems like it would be very doable as OT serves up raw frames and I am just saving those.  It would be a much more efficient solution.  Right now mine saves the raw frame, converts it to jpg (or png), deletes the raw frame to save disk space, converts all jpg (or png) frames to video or gif, and then deletes the jpg files.  There are definitely intermediate steps that would be great to get rid of.  

Do you have any experience writing a frameserver?  Is that something you might be able to help out with?

ideasman42:
@blurymind, Blender's ffmpeg is built-in, so it wont help with integration in this case (the frameserver can be used with ffmpeg, but its not ffmpeg spesific).

@turtletooth, I didn't write Blender's frame-server, but reading over the code its quite simple, I've tested using it, it works quite well.

Regrettably I wont be able to help with this, task, just showing how its been done elsewhere.

turtletooth:
Thanks for the feedback.  It sounds like a good goal to add in the future.  In the meantime, this solution seems to work.  It is just a little slow.

yamisama:
As far as WebP goes, it seems that at least the Mac version of OpenToonz is deployed with the WebP QT plugin dll, so it may be possible to add support through `QImage` or `QMovie` without relying on external tools like ffmpeg, but I don't know enough about QT to say for sure. There is also a `libqgif.dylib` in the plugins folder, no idea if that also supports writing, plus libraries for a few other image file formats.

blurymind:
@yamisama well spotted. It might be advantageous to preserving space .
@turtletooth considering that it creates a lot of frames -using webp could potentially get the size to 50% less than using png

Which would be a huge win.
https://www.andrewmunsell.com/blog/png-vs-webp/

turtletooth:
@blurymind I will test it out.  It looks very promising and has the added benefit of making webp export already baked in.  OT already supports png out, so using webp as the intermediate adds webp as an export format too.

turtletooth:
@yamisama Thank you so much for letting me know about the built in webp.  I wasn't able to use it, because the rendering time was too slow (but the file sizes were amazing), but it made me look up saving a QImage which is built into Qt.  I'm still learning Qt, but this allowed me to bypass the step of saving the raw image, and then converting to an intermediate format.  I can save to PNG directly.  

@blurymind I will probably use this method for creating webp exports, but note that the process is slow, However, the file sizes were 1/3 of a jpg at 100% quality and half of a png. - Amazing and great quality.  I think slowness is the tradeoff for the filesize.

blurymind:
@turtletooth  @yamisama  could webp be used as a replacement fir tif - for normal raster levels?
Related: https://github.com/opentoonz/opentoonz/issues/622

turtletooth:
@kogaki If you want to play with it and see if it works- webm, gif and mp4 all should work now.  There is also no need now to set it to only render on one cpu core.  Mp4 and webm have sound (at least on Windows).  Mp4 can work with any version of ffmpeg and has nothing specific to openh264 or x264 in it.  

You can configure the path to ffmpeg in preferences.  I named the section "Import/Export."

There is still some refactoring and error checking to add, but it is close.

@blurymind I should have a portable version to check tomorrow.

toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include "trasterimage.h"
+#include "timageinfo.h"
+#include "qprocess.h"
+#include "qstringlist.h"
+#include "qdir.h"
+#include "tiio_ffmpeg.h"
+#include "qimage.h"
+#include "toonz/preferences.h"
+
+Ffmpeg::Ffmpeg() {
```
kogaki: "FFMpeg" might looks pretty. Isn't it?

kogaki:
Jenkins, test this.

toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include "tsystem.h"
+//#include "texception.h"
+#include "tfilepath.h"
+#include "tproperty.h"
+#include "tsound.h"
+//#include "tiio_webm.h"
+#include "tenv.h"
+#include "trasterimage.h"
+#include "timageinfo.h"
+#include "qprocess.h"
```
kogaki: plz use `#include <QProcess>` style include

turtletooth:
Changed, testing now.  Will repush once it builds.

turtletooth:
It doesn't build on mine this way, but I made the changes to see if it will build with Travis.   I think I need to add some linker items to the CMakeLists.txt for the Image project, but I don't have much experience with that.  What would I need to add to make it build?  Mine can find the qprocess.h, but not <QProcess>.

turtletooth:
Any ideas why its failing now?

kogaki:
@turtletooth

turtletooth:
I'm not at my computer now, but I can change that later.  Was that on ffmpeg.cpp? 
I'm glad you could compile it on Mac. Please let me know what you think.  There is still work to do but it's much closer.

kogaki:
No, you just need to change the codes in this PR (tiio_ffmpeg.cpp or so).
Change `#include <QImage>` -> `#include <QtGui/QImage>`, and add `#include <QProcess>` .

And you also need to link `Qt5::Gui` in CMakeLists.txt.
`target_link_libraries(image Qt5::Gui Qt5::Core Qt5::Network ${Z_LIB} ${GLUT_LIB} ${GL_LIB} ${JPEG_LIB} ${TIFF_LIB} ${PNG_LIB} ${EXTRA_LIBS})`
 -> `target_link_libraries(image Qt5::Gui Qt5::Core Qt5::Network ${Z_LIB} ${GLUT_LIB} ${GL_LIB} ${JPEG_LIB} ${TIFF_LIB} ${PNG_LIB} ${EXTRA_LIBS})`

kogaki:
Here's the change I made. https://github.com/turtletooth/opentoonz/compare/webm...kogaki:pull/624/merge

turtletooth:
Windows users can test this here: https://github.com/turtletooth/OpenToonzPortable/releases/tag/1.0.3-FeatureTest

Rayek:
I attempted to export a 48 frames long scene to Gif, Webm, and mp4. Also scaled the result in the export options. All works, except for mp4: it drops the final 6~8 frames.

And for some reason the setting for webm kept resetting back to 1 for both the quality and the scale.

turtletooth:
Import works now also.  Gif export does not support transparency- I haven't found a ffmpeg command that will take a png with alpha and convert it to a gif with transparency.  It sounds simple, but I haven't been able to get it yet.

Exporting with an imported gif, webm or mp4 is very slow.  I don't know what is causing the really long delay.  For now it would work for roto work if you don't need the original in the export.  I will look into what is taking so long - it just waits on Precomputing - but the cpu utilization is minimal.

@kogaki I had to make a change in tcolumnfx.cpp at line 1356 or else the assert at line 756 failed.  It looks like the wrong width and height are being read otherwise.  This affects trying to export with an imported gif, webm or mp4.  If there is a better way, please let me know.  

Imported files are cached in a ffmpeg folder in the cache folder.  This folder is deleted when OT closes.  It is also checked to delete when OT opens in case OT closed on a crash previously.  

I still need to do a bunch of error checking, but the basic functionality is there.

@Rayek Thanks for testing it out.  I will look into the mp4 issue.  I think it might be because the mp4 export is actually the fastest and it tries to run before all the files have been written.  I will see if I can put a check in there to make sure that all files have been written first.  I'm aware of the sliders not remembering their value.  I am looking into it.  Right now, if you open the settings, the sliders need to be moved to remember their value- even if it is to just move it right back.  This is actually an issue for not just webm.

toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
```diff
@@ -0,0 +1,338 @@
...
+	//QString tempPath= m_path.getQString();
+	QString number = QString("%1").arg(frameIndex, 4, 10, QChar('0'));
+	QString tempName = "In" + number + ".png";
+	tempName = tempPath + tempName;
+
+	//for debugging	
+	std::string strPath = tempName.toStdString();
+	if (TSystem::doesExistFileOrLevel(TFilePath(tempName))) {
+		QImage *temp = new QImage(tempName, "PNG");
+		if (temp){
```
yamisama: In C++, the normal `new` operator will never return `nullptr`, it will throw an exception of type `std::bad_alloc` if it fails. That means there is no need to and no point in testing the return value. 

If the intention was to be sure that the `getImage` function absolutely won't throw no matter what happens, you can use the `std::nothrow` version of the `new` operator instead, but since the function is marked neither `noexcept` nor `throw()`, this is probably not necessary.

toonz/sources/image/ffmpeg/tiio_gif.cpp
```diff
@@ -0,0 +1,261 @@
...
+//
+//  TLevelWriterGif;
+//
+//===========================================================
+
+TLevelWriterGif::TLevelWriterGif(const TFilePath &path, TPropertyGroup *winfo)
+	: TLevelWriter(path, winfo) {
+	if (!m_properties) m_properties = new Tiio::GifWriterProperties();
+	std::string scale = m_properties->getProperty("Scale")->getValueAsString();
+	sscanf(scale.c_str(), "%d", &m_scale);
```
yamisama: You can just use `std::atoi()` or `std::strtol()` from `<cstdlib>` instead when you want to do string to int conversion. Using a format string and `std::sscanf()` is slightly overkill.

Not trying to criticize you at all, just thought I'd point it out since you said you were new to C++.

toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
turtletooth: @yamisama Thanks- I will probably take out the if there and look at other error checking.

turtletooth:
@kogaki I fixed the slow export by not using precomputing on scenes where videos have been imported through ffmpeg.  Something in the precomputing was slowing it down horribly.  It works fine without the precompute.  I also disabled the flipbook popup after export because of the slowness.  

I had to go back to using one cpu core for export because there were occasional errors.  I might put a setting in preferences that lets users override this and use all cores at their own risk.  It often works, but it fails sometimes.

Right now, here is the status:
- Import for mp4, webm, and gif works. (Gif import supports transparency)
- Export for mp4, webm, and gif works. (Gif export does NOT support transparency)
- Audio is exported on mp4 and webm formats.
- Gif export framerate is not quite accurate.
- There is still some error checking to put in there

@kogaki @yamisama @shun-iwasawa If any of you could at least direct me toward what error checking needs to be put in there, I would appreciate it.  I apologize for my lack of knowledge in several c++ areas.  Thank you for any help.

@kogaki I really appreciate all the work and feedback you have given and for communicating the legal and licensing needs so that this might be able to be a part of OpenToonz.

kogaki:
Great! I'll check it out.

kogaki:
Jenkins, test this

turtletooth:
New testing version: https://github.com/turtletooth/OpenToonzPortable/releases/tag/1.0.3.071916

turtletooth:
@RodneyBaker helped me track down occasional frame flipping.  I also made looping the default for gif.

Rayek:
The animation that I tested for export to mp4 still is missing the last 8 or so frames.

Rayek:
Aside from the mp4 issue, I am elated to have the gif export. I've already begun to use your OT version to convert image sequences to smaller scale GIF versions to include them on forums, and so on. I prefer it over other gif tools now!

Example:

ballies

turtletooth:
@Rayek That's awesome to hear about the gif!  

Could you help me duplicate the issue with the missing frames?   What settings are you using?  What type of levels?  Any info would be appreciated.

Rayek:
You can download the project file here: http://www.estructor.altervista.org/ot/project.7z
The mp4 files are in the render folder.

After doing more tests, I discovered that the mp4 files do not play properly in VLC player. It cuts off before the end of the animation. Weird. They play fine in the viewer, media player, and browser.

turtletooth:
@rayek I did not have the same issue you did with VLC.  I brought your file into premiere and it showed up as 48 frames.  This was the last frame in Premiere:
rayek_premiere

This was the last frame in VLC:
rayek_vlc

I'm sorry I wasn't able to duplicate the issue.  But it was playing correctly elsewhere?  Are there any big holes you are seeing in the functionality?  

Thanks

kogaki:
Jenkins, test this

Rayek:
@turtletooth As far as functionality goes, my only wish is to have much more direct control over the quality and size settings. I am used to the level control as seen in professional NLEs, and the dropdowns are somewhat limiting my choice of export settings.  An "advanced" button that exposes all of ffmpeg's parameters would be nice to have.

But otherwise this is a fine addition - even Toonboom Harmony is unable to write anything other than MOV format. With Webm and GIF formats OT ought to gain extra attention by amateur animators.

kogaki:
@turtletooth Jenkins says he cannot build t32bitsrv.  It might because lack of linking "toonzlib" (` _find_toonz_library(TNZLIBS "tnzcore;tnzbase")` ) could you fix it?

turtletooth:
Let's try that.

kogaki:
Thanks!

kogaki:
Jenkins, try this on

turtletooth:
It really doesn't like windows huh?

kogaki:
Jenkins

kogaki:
Jenkins

kogaki:
Jenkins

kogaki:
Jenkins

kogaki:
We tried it. Here's a result

OSX
 * With ffmpeg installed with homebrew, exporting works great! Thanks.
   * `brew install --with-libvpx ffmpeg` enabled me to export webm
 * Crash while importing.
   * I will try debugging

Windows
  * The app cannot start. It says that there are error on `t32bitsrv.exe`.
    * Do you have windows environment and can you try working on `t32bitsrv`?

turtletooth:
@kogaki 
Glad export worked on Mac.  If you could let me know what's crashing on import, I can look into it.  I can't build on Mac so I can't check it.

I don't think the windows error is because of my code.  (I could be wrong) t32bitsrv.exe is related to QuickTime and I didn't touch any files related to it.  

There are other reports of it causing errors: 
#365
https://www.reddit.com/r/OpenToonz/comments/4c62tl/t32bitsrvexe_error_fix/

Again, I could be wrong.

turtletooth:
I added format checks to make sure the installed version of ffmpeg supports the format before enabling it in the output options.

kogaki:
So, can you build with `t32bitsrv` ? Does `typetool` works ? (It's depend on t32bitsrv).

turtletooth:
Just tested:
type

turtletooth:
The portable version includes ffmpeg and t32bitsrv - you can test it if you want.

turtletooth:
Does the homebrew install of ffmpeg include ffprobe?  Import requires ffprobe - I added a check for it today.  
Just a thought on why import might have crashed for you.

kogaki:
Yes, ffprobe is bundled with ffmpeg installed with homebrew. Anyway, I will try latest version. Thanks.

kogaki:
Jenkins, test this

toonz/sources/image/tiio.cpp
```diff
@@ -159,6 +162,33 @@ void initImageIo(bool lightVersion) {
   TFileType::declare("rgb", TFileType::RASTER_IMAGE);
   Tiio::defineWriterProperties("rgb", new Tiio::SgiWriterProperties());
 
+  // ffmpeg
+  if (Ffmpeg::checkFfmpeg()) {
```
shun-iwasawa: Please add a following line before the line 166:
`#if (defined(x64))`
and after the line 189, please add `#endif`

In Ffmpeg::checkFfmpeg() you are checking the value in Preferences, and Preferences tries to see TOONZPROFILES path from registry (in its ctor). This behavior seems to abort t32bitsrv.exe since registry paths are different between 32bit and 64bit binaries.
Adding `#if (defined(x64))` is to avoid this problem by skipping Ffmpeg::checkFfmpeg() on running t32bitsrv.exe. This check must be only needed for OpenToonz_1.0.exe.

turtletooth:
I added the check above.  

I also think I might have figured out the Mac crash on import.  I was parsing the ffprobe output and splitting the string using \r as a delimiter.  I don't think Mac uses \r - only \n.  So the new line is:
`tiio_ffpmeg.cpp: 329: QStringList split   = sizeResults.split("\n");`

turtletooth:
>As far as functionality goes, my only wish is to have much more direct control over the quality and size settings. I am used to the level control as seen in professional NLEs, and the dropdowns are somewhat limiting my choice of export settings. An "advanced" button that exposes all of ffmpeg's parameters would be nice to have.

@Rayek I was able to get the sliders working again and actually keep the values when the settings were reopened.  So now there is much more granular control over the scale and quality.  Once the initial commit is merged I will probably work on an advanced input dialog where users can input their own settings for ffmpeg to run.  This would take knowledge of how ffmpeg takes it's commands, but would give the user more control.

kogaki:
Jenkins, build this

turtletooth:
Hoping this works for you.  After you test this, I have added the ability to set how long ffmpeg can run before timing out.  I will wait to push it until you get a chance to test this.

kogaki:
@turtletooth I think you don't need to push the changes. You can just revert the change if it doesn't work!

kogaki:
Oh sorry, I mean "you don't need to *wait pushing* the changes."

turtletooth:
> I think you don't need to hesitate to push the changes. You can just revert the change if it doesn't work!

That works too.

kogaki:
Today's Check

OSX
  * Importing / Exporting works collectly. 🎉 

Windows
  * Importing works
  * Exporting almost works, but some frames are dropped with agif format.
     * @shun-iwasawa is investigating now.

toonz/sources/image/ffmpeg/tiio_gif.cpp
```diff
@@ -0,0 +1,246 @@
...
+  // set scaling
+  outLx = m_lx * m_scale / 100;
+  outLy = m_ly * m_scale / 100;
+  //ffmpeg doesn't like resolutions that aren't divisible by 2.
+  if (outLx % 2 != 0) outLx++;
+  if (outLy % 2 != 0) outLy++;
+
+  QString palette;
+  QString filters = "fps=" + QString::number(m_frameRate) + ",scale=" +
+                    QString::number(outLx) + ":-1:flags=lanczos";
```
shun-iwasawa: I think you should remove the "fps" filter here and use "-r" input option instead.

To be more precise, please modify the lines 74-75 to be as follows;
`QString filters = "scale=" + QString::number(outLx) + ":-1:flags=lanczos";`
...and insert following lines to the line 93;
```
preIArgs << "-r";
preIArgs << QString::number(m_frameRate);
```


When I checked your PR, I found some problem: If you set a scene fps to other than 25, some frames are dropped in the output GIF. For example, when you render a scene with 20 frames, the result GIF becomes 19 frames under 24fps (= the default value of OpenToonz) and 11 frames under 12.5fps.
According to [the documentation of ffmpeg filters](https://ffmpeg.org/ffmpeg-filters.html#fps-1), it says;

> ## fps
>Convert the video to specified constant frame rate **by duplicating or dropping frames** as necessary.

I'm new to ffmpeg so I'm not 100% convinced, but this option may be the cause of this problem.
The "-r" option is documented [here](https://ffmpeg.org/ffmpeg.html#Video-Options) as follows:

> ### -r[:stream_specifier] _fps (input/output,per-stream)_
> Set frame rate (Hz value, fraction or abbreviation).
> 
> As an input option, ignore any timestamps stored in the file and instead generate timestamps assuming constant frame rate fps.

Here is a comparison of the rendered results. Both results are rendered from the same scene which has 20 frames in length and 8 fps frame rate. Please note that the some frames are dropped from the former result. 

- With "fps" filter (current implementation):
using_fps_filter_8fps
 
- With "-r" input option (my suggestion):
using_-r_option_8fps

turtletooth:
@kogaki @shun-iwasawa Thank you for the help.  I added the change and caught an error where scaling wouldn't happen if a palette wasn't generated.  Gif fps now works as expected.  Great suggestion.

kogaki:
Jenkins

kogaki:
@turtletooth I think it's time to be merged because basic I/O feature works. Do you think this is still "WIP" PR ?

turtletooth:
I totally agree.  I'm sure things will come up that will need to get fixed, but it's all basically there.  I will squash bugs as they come up.  Thank you so much for all your help.

turtletooth:
Let's do this!

kogaki:
Merged! Thank you for your great contribution and continuous improvements !

turtletooth:
Thank you! Thank you!

blurymind:
Thank you @turtletooth ! This is an epic addition

PR #628, opened 2016-07-12
WIP - Linked styles enabled and style right click menu organized

turtletooth:
- Re-enabled linked styles - "Toggle Link to Studio Palette"
- Changed "Remove Links" to "Remove Reference to Studio Palette"
    - This sounds wordy, but it was to avoid confusion about toggling the link
- Re-enabled "Delete Unused Styles"
- Made toggle link and remove connection able to have shortcuts (Default for toggle link is \\)

To update linked colors, save the Studio Palette and now the option to update linked styles works.

stylelink

### DIFF
--- a/toonz/sources/include/toonzqt/paletteviewergui.h
+++ b/toonz/sources/include/toonzqt/paletteviewergui.h
@@ -200,9 +200,6 @@ protected:
   void zoomOutChip();
 
 protected slots:
-
-  void toggleLink();
-  void eraseToggleLink();
   void removeLink();
 
 private:
--- a/toonz/sources/include/toonzqt/selectioncommandids.h
+++ b/toonz/sources/include/toonzqt/selectioncommandids.h
@@ -12,6 +12,8 @@
 #define MI_PasteColors "MI_PasteColors"
 #define MI_PasteNames "MI_PasteNames"
 #define MI_GetColorFromStudioPalette "MI_GetColorFromStudioPalette"
+#define MI_ToggleLinkToStudioPalette "MI_ToggleLinkToStudioPalette"
+#define MI_RemoveReferenceToStudioPalette "MI_RemoveReferenceToStudioPalette"
 #define MI_Clear "MI_Clear"
 #define MI_SelectAll "MI_SelectAll"
 #define MI_InvertSelection "MI_InvertSelection"
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -1611,7 +1611,7 @@ void MainWindow::defineActions() {
   createRightClickMenuAction(MI_ToggleLinkToStudioPalette,
                              tr("Toggle Link to Studio Palette"), "\\");
   createRightClickMenuAction(MI_RemoveReferenceToStudioPalette,
-                             tr("Remove Connection to Studio Palette"), "");
+                             tr("Remove Reference to Studio Palette"), "");
   createMenuEditAction(MI_Clear, tr("&Delete"), "Delete");
   createMenuEditAction(MI_Insert, tr("&Insert"), "Ins");
   createMenuEditAction(MI_Group, tr("&Group"), "Ctrl+G");
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -1608,6 +1608,10 @@ void MainWindow::defineActions() {
   createRightClickMenuAction(MI_PasteNames, tr("Paste Name"), "");
   createRightClickMenuAction(MI_GetColorFromStudioPalette,
                              tr("Get Color from Studio Palette"), "");
+  createRightClickMenuAction(MI_ToggleLinkToStudioPalette,
+                             tr("Toggle Link to Studio Palette"), "\\");
+  createRightClickMenuAction(MI_RemoveReferenceToStudioPalette,
+                             tr("Remove Connection to Studio Palette"), "");
   createMenuEditAction(MI_Clear, tr("&Delete"), "Delete");
   createMenuEditAction(MI_Insert, tr("&Insert"), "Ins");
   createMenuEditAction(MI_Group, tr("&Group"), "Ctrl+G");
--- a/toonz/sources/toonzqt/paletteviewergui.cpp
+++ b/toonz/sources/toonzqt/paletteviewergui.cpp
@@ -1023,28 +1023,29 @@ void PageViewer::contextMenuEvent(QContextMenuEvent *event) {
   bool isLocked = m_page ? m_page->getPalette()->isLocked() : false;
 
   // remove links from studio palette
-  if (m_viewType == LEVEL_PALETTE && m_styleSelection &&
-      !m_styleSelection->isEmpty() && !isLocked) {
-    menu.addSeparator();
-    QAction *removeLinkAct = menu.addAction(tr("Remove Links"));
-    connect(removeLinkAct, SIGNAL(triggered()), this, SLOT(removeLink()));
-  }
+  //if (m_viewType == LEVEL_PALETTE && m_styleSelection &&
+  //    !m_styleSelection->isEmpty() && !isLocked) {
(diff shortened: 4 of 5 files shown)


konero:
Nice.

Would recommenced changing **Remove _Connection_ to Studio Palette** to **Remove _Link_ to Studio Palette** for consistency since it is referred to as linking both in OpenToonz and Toonz Premium. :+1:

turtletooth:
@konero - I understand your point there.  The reason I chose the word connection was to prevent users from mistaking it from simply disabling the link temporarily.  Once you remove the connection, there is no way to get it back ( I think ), so I wanted it to be clear that all interaction between the level style and the studio style would be severed with that option.  The original code that I uncommented out used the term "Remove Reference to Studio Palette".  I am not too attached to the term, so if it seems that people prefer "Remove link" I don't mind changing it.

konero:
I was wrong, it's not link in TP. I had a look back through the Harlequin manual, its called "Remove Reference to Studio Palette" and they call that little white square the "studio palette reference icon". So if remove is used it would remove the white square, when toggle is used it toggles the icon between blank and an arrow. In that case reference would make more sense, considering its what you'll find in documentation and Toonz Premium.

> Remove Reference to Studio Palette

turtletooth:
@konero I think it is a valid point that keeping the documentation between the two in sync could be useful.  However, the point could be made that each new feature that is added to OT makes the old documentation less accurate anyway.  I think that the wording should be the least confusing for end users.  I personally think that the term "connection" makes more immediate sense than "reference," but again that's only my opinion.  I would like to hear from others regarding the phrasing.  I will put whatever seems to be the consensus.

blurymind:
I agree with @turtletooth  that connection sounds a little bit more obvious :)

blurymind:
I think you should review and merge it and then get feedback from the entire user community - as they will then try it in the nightly

AyameLee:
If you're going to change it to **Connection** you might as well use a shorter synonym like **Link**, since that menu looks wide.

How does this feature work exactly? Since I haven't used it, then it is difficult to gauge. The way I would personally utilize this is by using the Studio Palette to pick and hold all colors which Levels "refer" to, meaning they can't overwrite the Studio Palette. In that case, Reference is a better term, like a color model.

RodneyBaker:
You could always be bold and use 'Ref'.  :)

Reference does seem a bit more technically correct.
This especially in light of other things that use the same term.

blurymind:
i think shortening the names too much will make them vague

turtletooth:
I'm going to change it back to reference.  I don't think link is best because I think there should be a clear distinction between temporarily turning on and off the connection to the Studio Palette (Toggle Link) and permanently removing the connection - (Remove Reference).  @AyameLee You are right that it is a one way street- the layer palette cannot push to the studio palette.

turtletooth:
BTW- I am going to put this in tonight's portable version for anyone who wants to try it on Windows.

blurymind:
Cant wait to try it :)

blurymind:
@turtletooth can we make it a two way street? Having the level palette push back changes to the studio palette when link is enabled would be more intuitive.

Then saving the changes to the studio palette would still be all that it takes to propagate them to the other levels

blurymind:
This coul be optional :) but its an idea

blurymind:
@AyameLee the goal of the linking feature is to setup global colors that can easily be shared by multiple levels. Without it dealing with color consistency across levels is a laborious on opentoonz

turtletooth:
@blurymind That would be nice.  Could you make an issue for it?  It should probably be its own PR and I won't be able to get to it super soon.

turtletooth:
Connection has been changed to Reference

turtletooth:
Windows users can test this here: https://github.com/turtletooth/OpenToonzPortable/releases/tag/1.0.3-FeatureTest

Rayek:
Just tested it - it works like a charm. Thank you, it indeed simplifies palette management tremendously.

AyameLee:
@blurymind It doesn't make sense to push changes to the Studio Palette, why would you use it this way on something that has such a global effect? It's haphazard.

blurymind:
@AyameLee You dont even use the feature , so how can you understand its value :D 

You still would need to have palette editing unlocked.
You still would need to have enabled linking.
You still would need to save the studio palette in order to apply the change.
How many more safeguards do you want?

blurymind:
That said - perhaps there should be a * symbol or some kind of a hint on styles on the studio palette that have been changed since they were saved - this would warn the user that hitting the save button would write a new value to the style - but wasnt there already a warning on changed styles?

blurymind:
@Rayek  how did you get the linking to work? I seem to have trouble enabling it. When I enable it no arrow indicator appears - and the functionality is not there

blurymind:
Now it started working - the first time it didnt. I will find out why

blurymind:
@turtletooth It doesnt initially work if you 'load into current palette'  

Ok here are the steps to reproduce the problem:
1. create a new level by drawing with the brush on the empty canvas
2. select a studio palette and 'load into current palette'
3. right click on a level style and  'toggle link to studio palette'

nothing will happen

4. save studio palette's change and try again - nothing

5. right click on studio palette and this time do 'merge to current palette' - a new page will apear with the styles

toggling link still doesnt work on them either

blurymind:
I dont understand why its not working - removing reference doesnt seem to do anything either.

Yet I did something to get it to work. What am I missing

blurymind:
I am seeing more weird behavior - if I load a studio palette to be the level palette - the fill colors on it are different from the ones that the bucket fill applies.

blurymind:
I need to figure out whats causing this:
palettewoes

blurymind:
Linking doesnt work - its still very broken.

blurymind:
this color offset problem is something i have encountered before on opentoonz btw - so its not a regression on this build. Previous builds had it too. I will file it as a new issue to the tracker later

blurymind:
The color offset is for another issue - but it gets triggered if you 'adjust current level to this palette'

blurymind:
I will write down if i manage to figure out how I got linking to work once.

In harlequin you just 
load a studio palette to the current level palette (load into current palette) and enable it (toggle link..). Thats it.

here it doesnt work at all

blurymind:
Here it is still Broken in Opentoonz:
brokenlinking

and here it is functioning in Toonz Harlequin:
linkingworking

turtletooth:
I will look into it. For now does it work if you drag the palette into the
level palette?

blurymind:
@turtletooth  When you drag and drop it - linking does work.

So something must be broken/missing the right click commands to:
load into current palette and merge to current palette - when using them - it is not possible to toggle the link

blurymind:
Another problem I found:
stripfailstoupdatethumb

When using a linked style - the strip thumbnail of the frame fails to update to changes made on it.

turtletooth:
@blurymind Thank you for finding the other bugs.  It looks like this is a much bigger rabbit hole than I expected.

turtletooth:
I'm closing this until I can submit a more complete PR.  I didn't realize the variety of ways there were to break linked styles.

blurymind:
Thats understandable. I hope you dont give up on it. The feature is definitely a must have imo.
It's hard to imagine why dwango would remove it in the first place. How removing it would cause all of these regressions.

turtletooth:
I haven't given up on it.  I just didn't realize the depth of the issue when I submitted.  Right now, there are a few more pressing PR I want to make before the school year starts so my kids have the best interface.

blurymind:
without this - coloring would be difficult - especially for cut out style

turtletooth:
I haven't forgotten about this. There are just a few other things I'm going
to try and get to before my students start using it this year.  Right now
the palette file gets really dirty when importing more than one studio
palette or deleting a palette and reimporting it.  Would you mind using a
toonz raster level and linking and unlinking a bunch of studio palettes and
sharing the file or telling me what happens?

The current setup gets multiple copies of the same palette on the file and
i think OT gets confused.

Without knowing the actual desired behavior, it's hard to implement it.

PR #629, opened 2016-07-12
Correct delete[] use

ideasman42:


yamisama:
Thanks for your work, it's greatly appreciated!

One thing though: I think it might be better to replace these with appropriate STL smart pointers instead of explicit `delete[]` operators. That way, the probability of introducing similar memory leaks in the future would be reduced (like for example when someone decides to use another `return` statement somewhere earlier in the function without realizing there is cleanup to do), and it would also make the code exception safe.

skitaoka:
Jenkins, test this.

skitaoka:
I also prefer to use shared pointers; however I merge this PR because fixing bugs is top priority.

PR #632, opened 2016-07-13
Cleanup: declare local functions static

ideasman42:
Applies to toonz, toonzfarm, toonzqt.
Follow on from (#610).

Note that there was one function left as an `extern`, because of type declarations.
`getSafeAreaSizeList`.

Also added header `sceneviewerevents.h`.


skitaoka:
Jenkins, test this.

skitaoka:
LGTM

Issue #381, opened 2016-06-01
OS X crash on vector animation (nightly build 5/27)


amitgollapudi:
Steps to recreate ... 

1) Create a simple vector shape
2) Animate a simple move of shape using sheets over a few frames
3) Hitting play in console within application is functional
4) Going to preview or render causes an immediate crash

I've verified using released version (1.0.2) and nightly build.  I've confirmed on El Capitan and Yosemite.

amitgollapudi:
Some details from a sample run ... crashes are consistent.

```
Process:               OpenToonz_1.0 [27676]
Path:                  /Applications/OpenToonz_1.0.app/Contents/MacOS/OpenToonz_1.0
Identifier:            ???
Version:               ??? (???)
Code Type:             X86-64 (Native)
Parent Process:        ??? [1]
Responsible:           OpenToonz_1.0 [27676]
User ID:               1322684350

Date/Time:             2016-06-11 09:39:52.689 -0400
OS Version:            Mac OS X 10.10.5 (14F1808)
Report Version:        11
Anonymous UUID:        BFED478C-1414-CE77-7D31-7EE1456D0915

Sleep/Wake UUID:       872C24ED-8BE5-4FCE-9053-43D2F6C8F3C2

Time Awake Since Boot: 19000 seconds
Time Since Wake:       600 seconds

Crashed Thread:        12  QThread

Exception Type:        EXC_CRASH (SIGABRT)
Exception Codes:       0x0000000000000000, 0x0000000000000000
```

Stack from thread 12 -

```
Thread 12 Crashed:: QThread
0   libsystem_kernel.dylib        	0x00007fff94c30286 __pthread_kill + 10
1   libsystem_c.dylib             	0x00007fff87eda9ab abort + 129
2   org.qt-project.QtCore         	0x000000010ed78ab9 0x10ed53000 + 154297
3   org.qt-project.QtCore         	0x000000010ed7a487 QMessageLogger::fatal(char const*, ...) const + 231
4   org.qt-project.QtGui          	0x000000010e83fa20 QOpenGLContext::makeCurrent(QSurface*) + 656
5   libtnzcore.dylib              	0x000000010be51d21 QtOfflineGL::makeCurrent() + 209
6   libtnzcore.dylib              	0x000000010be73ffc TOfflineGL::makeCurrent() + 76
[...]

PR #636, opened 2016-07-14
Fix OSX crash on vector animation rendering

tomosu:
This PR is related to #381.
In qtofflinegl.cpp, call moveToThread(0) to abandone the ownership of the GLContext before doneCurrent(), and add doneCurrent() at the end of some stdfx functions.


tomosu:
Jenkins

skitaoka:
LGTM

PR #637, opened 2016-07-14
Fix crash reading level (TLZ) file

ideasman42:
- double fclose call on the same file/
  (readVersion closed the file, however this is already closed when the class is freed)
  The error was that setting the file pointer to NULL only changed the local value.
  It could be passed as a reference or pointer to pointer, however since its closed when the class is freed, not sure its worth extra hassles.

- buffer overrun reading file magic which isn't `\0` terminated.
  use memcmp instead.


ideasman42:
Accidentally pushed the branch to this repo.
Seems to have confused travis-ci.

skitaoka:
Jenkins, test this.

skitaoka:
OK. LGTM

Issue #626, opened 2016-07-11
Improve the Pencil Test feature


shun-iwasawa:
This issue is for summarizing the feedback on the first release of Pencil Test feature (currently available only in the nightly build):

- [ ] Add a maximize button to the window.

- [ ] Enable Capture with shortcut key (enter key) - a temporary shortcut only for when the Pencil Test window is open.

- [ ] Set the resolution to its highest value by default instead of its lowest.
       
- [ ] Add option to save the images as they are captured.

- [ ] Change the command name from Pencil Test to Camera Capture, in consideration of usage for stop motion animation.

- [ ] The ability to playback at least the captured images in the window.

- [ ] Multi frame onion skin.

- [ ] The ability to shoot a reference video with camera in real-time for acting shots.

- [ ] BUG: Changing the resolution sometimes causes crash. 

- [ ] Support DSLR camera.

Rayek:
Love all of these suggestions.

RodneyBaker:
Thanks from here also.
I'm really enjoying this new feature.

If I can be so bold as to suggest another enhancement it would be to allow
a timer for capture.  In this way the users hands would be freed to fully
concentrate on placing the next drawing (or in case of stopmotion... adjust
props/characters) in front of the camera.   I'd say the timespan might be a
10 count (or 10 seconds) between captures but this would optimally be a
number the user can set.  When the user is set to capture they hit
Capture/Start and off they go only pausing to wait for the image to be
captured.  Of course, some visual clue and/or sound would be nice to signal
the completion of the capture.  Then the user knows it's time to set the
next shot in anticipation of the next capture.

turtletooth:
@RodneyBaker added your request to the list at top

shun-iwasawa:
Updated the task list.

turtletooth:
@shun-iwasawa Thank you so much for the improvements.  I added one more item to the list:
The ability to load a reference image for lining up images between stop-motion sessions.
This will help stop-motion artists continue where they left off from day to day.

RodneyBaker:
Nice update!
Looking good on Win10.

I hope to purchase an external webcam as early as tomorrow and will test that also.
It works great with my internal webcam.

shun-iwasawa:
Updated the task list. @RodneyBaker , thank you for reporting the bug.

>  Support the preferences option "Show ABC Appendix to the Frame Number in Xsheet Cell"

I think this task needs explanation. This demand arose from some Japanese animation studio.
[...]

PR #652, opened 2016-07-19
Camera Capture (former Pencil Test) improvement

shun-iwasawa:
This PR is for some items listed the issue #626

1. Added the maximize button.
1. Enter/Return key is now available as shortcut key for capture.
1. Default resolution is changed to the camera's highest available value. 
1. Added checkbox to save images as they are captured. Note that the default state of this check box is ON (= do save images).
1. Command name and popup title are changed to "Camera Capture"
1. "Interval timer" option is added. It enables to automatically capture frames every specified seconds. Note that if you set "Interval(sec)" value to 0, OpenToonz try to capture frames as frequent as possible (=real time).
1. When the interval timer is activated, countdown is shown on the bottom-right corner of the finder area.
1. Default file type is changed from PNG to JPEG in order to reduce file size.


shun-iwasawa:
Jenkins

turtletooth:
LGTM - Thank you for the great changes.  It works great and the added functionality is wonderful.

Rayek:
素晴らしい !

RodneyBaker:
I need to test again but it appears to me that when the checkbox "Upside Down" in the Image Adjust section is turned on, the Countdown in the bottom right corner also flips and then displays in the upper left corner with numbers inverted upside down.  I'm sure this side effect wasn't intended and is a bit hard to read.  

Regardless, you know I'm a fan of these enhancements.  Many thanks!

shun-iwasawa:
@RodneyBaker Thank you for pointing the "upside down" bug! I'll fix it right away.

PR #653, opened 2016-07-20
Fixed crash caused by lvcod64.dll on AVI settings popup

turtletooth:
lvcod64.dll was causing a crash on using the AVI settings popup.  This is the same problem from #52.  I used the same solution as in #52.


turtletooth:
What needs to be changed for this to pass?  Works on Windows 10 64 bit.

toonz/sources/toonzlib/avicodecrestrictions.cpp
```diff
@@ -220,20 +220,34 @@ QMap<std::wstring, bool> AviCodecRestrictions::getUsableCodecs(
...
+          hic = ICOpen(icinfo.fccType, icinfo.fccHandler, ICMODE_COMPRESS);
+        } __except (EXCEPTION_EXECUTE_HANDLER) {
+        }
+      }();
+#else
+      try {
+        hic = ICOpen(icinfo.fccType, icinfo.fccHandler, ICMODE_QUERY);
+        ICMODE_COMPRESS
+        catch (...) {
+        }
```
shun-iwasawa: Please fix here in order to enable to build when _MSC_VER is not defined.

shun-iwasawa:
I'm not actually understand the issue modified here as I couldn't reproduce the crash on using the AVI settings popup. Is there any condition to reproduce the problem?

turtletooth:
I don't know if it is a Windows 10 issue, but the crash is consistent for me. 
avi_crash

> Please fix here in order to enable to build when _MSC_VER is not defined.

I will look into that.

shun-iwasawa:

RodneyBaker:
REF also:  https://github.com/opentoonz/opentoonz/issues/46

Issue #576, opened 2016-07-04
Bug: OpenToonz doesn't recognize .jpeg file extension (1.0.3)


cyaoeu:
Tiny issue but OpenToonz doesn't recognize the .jpeg file extension in Load Level. It's basically .jpg with a non-standard (?) file extension so it should work. Renaming the file in explorer or whatever makes OpenToonz recognize the file again.

animking:
it's alright here win 8.1. Just rename(may be your file include complected name with lots of "_") it and load.

cyaoeu:
The point is that you shouldn't have to rename the file for it to work. Here's a screenshot:

![opentoonzbug](https://cloud.githubusercontent.com/assets/3804162/16582599/1db6317a-42b0-11e6-8cce-31a51e609ef0.jpg)

Both files should show but only the .jpg one is showing.

turtletooth:
#656

PR #656, opened 2016-07-20
Support for jpeg import

turtletooth:
This allows jpeg images to be imported.  #576

### DIFF
--- a/toonz/sources/common/tiio/tiio_std.cpp
+++ b/toonz/sources/common/tiio/tiio_std.cpp
@@ -15,6 +15,9 @@ void Tiio::defineStd() {
   declare("jpg", TFileType::RASTER_IMAGE);
   defineWriterProperties("jpg", new JpgWriterProperties());
 
+  defineReaderMaker("jpeg", Tiio::makeJpgReader);
+  declare("jpeg", TFileType::RASTER_IMAGE);
+
   defineReaderMaker("bmp", Tiio::makeBmpReader);
   defineWriterMaker("bmp", Tiio::makeBmpWriter, true);
   declare("bmp", TFileType::RASTER_IMAGE);


turtletooth:
On a side note:  is it important to have tif and tiff both as export options?  It seems redundant.  Tiff import could be supported without tiff export being needed.  Are there programs that require a tiff instead of a tif?

blurymind:
I too fail to understand the advantage of tiff over other file formats - in a pipeline. Opentoonz seems to heavily rely on it for storing frames in raster levels. This results in them taking a lot of hard disk space:
https://github.com/opentoonz/opentoonz/issues/622

perhaps in a workflow with other software (paintman? retas?) - or something to do with storing colors? Perhaps somebody could explain?

RodneyBaker:
Tiff files are on par with OpenEXR in that they lean forward into
production ready image formats.
If one can imagine how OpenEXR with its various channels and such might be
advantageous that might shine some light on similar aspects with Tiff.
If we are talking older Tiff functionality then perhaps Tiff is not as
important but I suspect OT might use some of the more modern aspects of the
file format.

As far as TIF vs TIF... I regard that in much the same way as JPG vs JPEG
and a similar equivalent in text files would be HTM vs HTML
There is no difference between the files other than their
naming/extension.  The three letter extension of course being a carry over
from DOS/Windows early restriction to a three letter file extension. **
That appears to have more to do with file organization than the format
itself.


**Many years ago I tried to standardize htm vs html for my personal
projects with HTM meaning global/external/online and HTML meaning
local/internal.
I thought it was a decent idea at the time but I didn't stick with it
because I was the only one doing it that way and I wasn't doing it
consistently.  ;)
I believe I was trying to develop a methodology for automatically and
instantly knowing where files came from and/or resided.

turtletooth:
Regarding the tif vs tiff- I was simply wondering if they both needed to be available as export formats.  I think they should both be available for import.  Internally, OT handles them the same way.  I think that having tif and tiff in the export dropdown clutters up the list.

RodneyBaker:
Several exporters/renderers whose workflow I like have just the one file
extension but... they have a slightly different approach that can get you
any file extension quickly.
The idea is that since the user is typing anyway that *if* they type the
extension the program automatically resolves the target file format.
For instance, if the user types image0000.tga the program sees the desired
extension and matches that.
If output.mov is typed, .mov is assumed and automatically selected.
If "out.tiff" is typed the exporter will assume a .tiff extension.
If "out.tif" is typed the exporter will assume a .tif extension.

What this does is all line entries to read things like "tif, tiff" and
"mov, mp4" etc. and the final determination is up to the user.

If on the other hand, the user doesn't specify the extension but instead
selects "tif, tiff" the extension ".tif" is assumed and appended as the
extension.

I realize there is a little more to this than in those programs because OT
has regex (regular expression) formatting that ties into the mix.
I must assume other programs may too but hide that from the user.

One of the example programs I'm thinking of is easily investigated because
of its free version and that is Blackmagic Design's Fusion, an excellent
node-based compositor that works in very similar fashion to OT's Schematic.
(see https://www.blackmagicdesign.com/products/fusion)

These programs try their best to make file types and extension selection as
trivial as possible and as such the user can opt to select a file type from
a dropdown but don't have to if they type the extension.

It's probably just muscle reflex but I'm in the habit of typing things like
image0000.jpg in other programs and pressing render.
Appending the zeros simply signals the number of digits for the count after
the filename and .jpg is the supplied extension.
If I type "image0000.tiff" Fusion and Animation:Master render that as a
sequence starting with image0000.tif

This filenaming and numbering quickly gets complicated because of frames
versus cels etc. (starting with a frame 0 vs frame 1) but... yeah.

shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa Thanks!

shun-iwasawa:
LGTM

Issue #159, opened 2016-04-11
Function curves editor in open toonz broken - can't move keyframes or their bezier handles


blurymind:
Would it be possible to revert changes done to the opentoonz Function editor? It's quite broken in it's current state.
here is the function editor in toonz:
![functeditor-toonz](https://cloud.githubusercontent.com/assets/6495061/14440057/1e016b76-0026-11e6-83de-a0f410c635cb.JPG)

and here it is in opentoonz:
![functioneditorotde](https://cloud.githubusercontent.com/assets/6495061/14439998/dabf6368-0025-11e6-811d-d216a0d70777.JPG)

- You can no longer move keyframes in the graph view! This is the big one.
- You can no longer adjust spacing - the bezier curves can not be moved either. You can select them, but grabbing them does nothing.

All this makes the function editor's graph view incapable of doing anything but display the curves.

jay3d:
You can hold CTRL while dragging around keys and handles to adjust them

blurymind:
@jay3d ah I see. Thank you for telling me about this.

This behavior is strange to me. Why does one have to hold control key now? 
At first I simply thought that it doesnt work. Is this something that can be re-configured?

blurymind:
I would suggest removing the need to hold the ctrl key - as it makes the software much less welcoming to new users. I dont think the hold to do anything in the curves editor design atm is justified. What do we gain by having it? Why was it introduced in the first place?

blurymind:
@jay3d  is there a way to disable the having to hold ctrl behavior in order to edit keyframes?

I see this as a regression from toonz harlequin - where holding control was required only when you need to add keyframes.

turtletooth:
I just read this-  I thought for the longest time that the graph was just broken.  Having to hit Ctrl does seem like a strange thing.  I am glad to know that the function graph actually works.  It just needs to be more user friendly.

blurymind:
Yes indeed. This wasnt the case in toonz harlequin. It should be an easy to revert change - low hanging fruit. I asked why it was introduced to see if @shun-iwasawa would have anything against reverting it and also if keeping the option has any merrits

turtletooth:
#658 and #657

shun-iwasawa:
Merged option 1.

blurymind:
Thank you! :)

PR #657, opened 2016-07-20
Function Curves Window Usability - Option 1

turtletooth:
This is my preferred option- remove the need to press the control key for modifying positions/values.
The current setup is almost impossible to discover.
#159


Rayek:
Agreed - I know of no other graph editor in either 2d or 3d software that requires a control key to modify the curves.

Definitely prefer this option.

shun-iwasawa:
I added the "Ctrl key condition" to Ghibli Edition reflecting demands from the studio in order to avoid changing values by mistake. I agree that it is no longer a necessary feature, in consideration of usability for first-time users.
I prefer the option 1 too.

shun-iwasawa:
Jenkins

blurymind:
@turtletooth  thank you!! :)

shun-iwasawa:
LGTM

PR #658, opened 2016-07-20
Function Curves Window Usability - Option 2

turtletooth:
This option changes the functions curves window title to "Function Curves - Use Control Key to Modify Positions".  I don't think this is as user friendly as option 1, many users don't read the title bars of windows.  (Does the title bar for this window show up on Mac?)  

But at the very least this might give users a chance to find out how to change the values.
#159


shun-iwasawa:
Option 1 was merged. @turtletooth , thank you!

PR #662, opened 2016-07-21
Change default icon size to 16 by 9 (160 x 90)

turtletooth:
With default camera size being 1920 x 1080, I thought it might make sense for the icon to match.  I realize that raster levels don't always match the size of the final output, but Toonz Raster and normal raster have wasted space on a 160 x 120 icon.  This will allow more frames to fit in the column.


shun-iwasawa:
LGTM

Issue #373, opened 2016-05-26
Feature request - The ability to add a custom project root through the interface


BrundleThwaite:
In order to add a custom project location you have to manually change either files (Mac) or modify the registry (Windows). I do not think that it is a good idea for people to be changing the registry.

I would probably be better to be able to do this directly through the interface, probably as a setting in preferences,

![registry](https://cloud.githubusercontent.com/assets/18170954/15573678/f3f92b20-2348-11e6-8d60-83817bb32057.png)

Rayek:
Agreed, this would simplify life a lot in classroom situations, for example.

turtletooth:
I agree.  I am a teacher and have multiple classes and not every class should save to the same location.  I added the ability to save to the desktop earlier as a workaround, but that only works for levels and scenes.

blurymind:
Autodesk maya uses this approach where it can creaye a new project tree, or set current prpject root to a custom directory. Blender uses a single file to store everything- but internally its just the same thing- a project folder

turtletooth:
#664 takes care of most of this.  (Not multiple locations at the same time, but customizable locations)

BrundleThwaite:
I did a video in which I did explain how to make multiple project locations.

https://youtu.be/THY3C2jj-_M

Allowing people to set custom locations from the Interface is a good start though.

turtletooth:
I added a second choice in the PR with checkboxes for each location.  I think I prefer it better:
![project2_pref2](https://cloud.githubusercontent.com/assets/4576381/17012915/2c76b3c2-4ed8-11e6-85ec-c8ad38744a08.jpg)

However, I still want to resolve having an individual sandbox for each user in a multi-user environment.

BrundleThwaite:
If you do not want sandbox to not be the default consider the following.

a) The user is going to have to specify the default folder when the application is first launched. But every new install (Especially on windows) is going to require them to set any custom project locations manually.

b) The default is going to have to be set as is standard per application so for example in windows (mac and linux would be different) you would have data stored in.
 C:\Users\NAME\AppData\Roaming\Opentoonz 1.0 stuff 
and documents would be stored in
C:\Users\NAME\Documents\Opentoonz 1.0


[...]

PR #664, opened 2016-07-21
Additional Project Root Location

turtletooth:
This option allows users too add additional project roots to OpenToonz. 
- The checkboxes for documents and desktop are for a quick and easy way to add those locations.  
- The custom box allows for multiple file locations separated by ** (no spaces).  
    - This allows for easily adding multiple paths without using regedit or editing files.  
- By default the documents checkbox is checked to help in multi-user environments so that not everyone's work has to go in the sandbox or default project root.

This does not change the default project root folder or the sandbox.  If previous multiple locations were added through regedit or file editing, they will still show up (but not in the custom box).

Here are screenshots:
proj3_pref

proj3_custom

proj3_new

proj3_brows


turtletooth:
This addresses #373

turtletooth:
Another option would be to allow checkboxes for each item:
project2_prefs
This allows for this:
project2_new
And this:
project2_browser

But raises the question about the sandbox and multiple users.  I don't think a global sandbox should be the default - especially in a multi-user environment.

turtletooth:
Making changes based on feedback.

turtletooth:
@shun-iwasawa Do you have any further thoughts on this?  If the feature was disabled by default and moved from the front page of the preferences would you consider it?  It doesn't change any existing workflow and will help in certain situations.

shun-iwasawa:
Reopen for #736

turtletooth:
I'm making some small changes to this right now.  Will push soon.

turtletooth:
@shun-iwasawa Here is the current set of options:
currentprojectroots

I have hidden the ability to turn off the stuff folder.  (This would break the resolution presets - which I suggested moving to a different location, and potentially break existing projects.)  By default, no other location is activated (although I love having a OpenToonz folder in My Documents for easy access).  This shouldn't break any existing functionality.  Thanks for considering this.

shun-iwasawa:
Using QStandardPaths looks nice! I didn't know it.

FYI, failing OSX compilation in the previous code was because of the reason that a link declaration to cocoa library was missing in CMakeLists.txt. Exploiting QStandardPaths is smarter anyway.  :-)

turtletooth:
@shun-iwasawa Thanks.  I figured the failing part was a missing CMake issue, but I don't yet know enough about CMake to know for sure what to add without breaking something, so I looked for an alternative and found QStandardPaths.

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -911,6 +940,15 @@ PreferencesPopup::PreferencesPopup()
...
 
+  m_projectRootDocuments = new CheckBox(tr("My Documents/OpenToonz*"), this);
+  m_projectRootDesktop   = new CheckBox(tr("Desktop/OpenToonz*"), this);
+  m_projectRootStuff     = new CheckBox(tr("Stuff Folder*"), this);
+  m_projectRootCustom    = new CheckBox(tr("Custom*"), this);
+  m_customProjectRootFileField = new DVGui::FileField(this, QString(""));
+  m_customProjectRootLabel     = new QLabel(tr("Custom Project Path(s): "));
+  m_projectRootDirections      = new QLabel(
+      tr("Advanced: Multiple paths can be separated by ** (No Spaces)"));
+
```
shun-iwasawa: How about to use ";"(a semicolon) as separator?
In the original TOONZPROJECTS registry, a new projectroot path should be separated by semicolon and IMO it is quite common as a path separator.

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -1369,6 +1424,25 @@ PreferencesPopup::PreferencesPopup()
...
+        projectRootLay->addWidget(m_projectRootStuff, 2, 0);
+        projectRootLay->addWidget(m_projectRootDocuments, 3, 0);
+        projectRootLay->addWidget(m_projectRootDesktop, 4, 0);
+        projectRootLay->addWidget(m_projectRootCustom, 5, 0);
+        projectRootLay->addWidget(m_customProjectRootLabel, 6, 0,
+                                  Qt::AlignRight | Qt::AlignVCenter);
+        projectRootLay->addWidget(m_customProjectRootFileField, 6, 1, 1, 3);
+        projectRootLay->addWidget(m_projectRootDirections, 7, 0, 1, 4);
+      }
+      generalFrameLay->addLayout(projectRootLay, 0);
```
shun-iwasawa: It might be a matter of taste, but how about to put these checkboxes in QGroupBox?
Like this:

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -278,6 +278,35 @@ void PreferencesPopup::onPixelsOnlyChanged(int index) {
...
+void PreferencesPopup::onProjectRootChanged() {
+  int documents = m_projectRootDocuments->isChecked() ? 1 : 0;
+  int desktop   = m_projectRootDesktop->isChecked() ? 1 : 0;
+  int stuff     = m_projectRootStuff->isChecked() ? 1 : 0;
+  int custom    = m_projectRootCustom->isChecked() ? 1 : 0;
+
+  QString projects = QString::number(documents) + QString::number(desktop) +
+                     QString::number(stuff) + QString::number(custom);
+  m_pref->setProjectRoot(projects.toInt());
+  if (custom) {
```
shun-iwasawa: You are using decimal (base-10) flag. I think it would be faster to use bit flags instead.

toonz/sources/toonz/preferencespopup.cpp
turtletooth: I chose the ** because a semicolon is a valid character for a Windows path.  I don't think ** is valid in any os.  Thoughts?

toonz/sources/toonz/preferencespopup.cpp
turtletooth: I like it.

toonz/sources/toonz/preferencespopup.cpp
turtletooth: I've never done bitwise operations before, but I will look into it.

turtletooth:
Changes submitted.  I didn't change the separator right now because semicolon is a valid character in a Windows path (https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx)

If you would prefer a different character, I don't mind changing it.

toonz/sources/toonz/preferencespopup.cpp
shun-iwasawa: Understood.

shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

turtletooth:
Thank you!!

gabivallu:
why isn't this option available ate the 1.1.1 version? and if it came later, when will it be available? this surely makes a BIG difference! 

I'm sorry I am not a programmer nor understand much of how things must be done at the backstage, but from what I already tested opentoonz so far, the biggest issue is how complicated it is to create/save projects and files. the idea of project folders and organization is nice, but the saving process must be simplified, and also if possible, to enable saving in other hard drives/partitions or folders, as the user prefers.

I work with after effects , and maybe the way it works with multiple linked files and the freedom of project organization could help you have ideas of how to tackle this issue...? 

great software so far, keep it evolving please!

turtletooth:
The change is in the latest nightly builds: https://github.com/opentoonz/opentoonz/releases
There are also a bunch of other great features in there.  I don't know when a new stable release will be released, but I only use the nightly and it is at least as stable as 1.1.1 in my experience.

gabivallu:
yey! nice! totally forgot to check these builds! thank you.

shun-iwasawa:
@gabivallu @turtletooth 
We are preparing next stable release (v1.1.2). It'll be available next week at the latest.

Issue #670, opened 2016-07-23
Pixel Unit Measurement Feedback


konero:
Just trying out the Pixel Unit Measurement (thanks @turtletooth).

I think it would be better if the **px** acronym was used instead of **pixel** especially in the window settings like cleanup or output. Actually, I think this of all the measurement unit types.

```
cm
mm
in
fld
px
```

The measurement type doesn't always update in Output Settings or Cleanup Settings, a restart is required. Also noticed both pixels and Pixels used.

![output_tz](https://cloud.githubusercontent.com/assets/19820721/17075922/c1f74f3c-509b-11e6-9e9a-3c31ac14cd78.PNG)

That lock icon could also do with a fix. :+1:

turtletooth:
Good catch - I will look into that.  Anyone else have any preference about px vs pixel?  I can go either way.  
I think I've also seen some places where decimals are included (1920.000001) - so a little clean up might be needed.  

I agree about the lock icon. :smile:

konero:
Also noticed the decimals. This one is under the create New Level window (vector).

![tx_px6](https://cloud.githubusercontent.com/assets/19820721/17076013/b5b544c4-509e-11e6-996a-4d9d7889ed28.PNG)

This one under Cleanup Settings.

![tx_px5](https://cloud.githubusercontent.com/assets/19820721/17076019/d94e04c0-509e-11e6-9cc3-e3949118eeff.PNG)

Also, using the resolution presets under Cleanup Settings is inaccurate.

turtletooth:
Thanks again.  There were a bunch of places to try and catch.  I apologize for missing some.

Rayek:
This is an excellent addition - for any animator wishing to rely on the exact physical pixel scale, such as pixel artists. In cases like that DPI would only confuse things.

turtletooth:
I think I took care of it all in #674.  Please let me know if I missed anything.

PR #674, opened 2016-07-24
Pixels Only Mode: Fixed preset dropdown, dpi lock, labels, decimals

turtletooth:
Misc fixes for pixels only mode:
- Labels now update without a restart
- Fixed resolution presets not working as expected
- Fixed issue where dpi wasn't locked at Stage::inch in camera settings
- Pixel values now don't show a decimal in all camera settings, new level, and level settings
- Restored the resolution label to level settings to make it clear that changing the width and height only scales the image- not actually resizes it. 

Level Settings screenshot:
level_settings_pixels

This addresses #670


turtletooth:
@shun-iwasawa @kogaki Before releasing an official 1.1 version, could you please look at this.  This fixes some issues with pixels only mode.

kogaki:
Jenkins, check this

turtletooth:
Thanks

PR #675, opened 2016-07-24
Updated field guide for new default resolution

turtletooth:
The field guide was too small for the new default resolution- updated it to fit the new default resolution (1920 x 1080).

### DIFF
--- a/toonz/sources/toonzlib/sceneproperties.cpp
+++ b/toonz/sources/toonzlib/sceneproperties.cpp
@@ -35,8 +35,8 @@ TSceneProperties::TSceneProperties()
     , m_markerOffset(0)
     , m_fullcolorSubsampling(1)
     , m_tlvSubsampling(1)
-    , m_fieldGuideSize(12)
-    , m_fieldGuideAspectRatio(1.38) {
+    , m_fieldGuideSize(36)
+	, m_fieldGuideAspectRatio(1.77778) {
   // Default color
   m_notesColor.push_back(TPixel32(255, 235, 140));
   m_notesColor.push_back(TPixel32(255, 160, 120));


shun-iwasawa:
Same as the #676, the current default value of field guide is set based on the actual size used in animation industry. They usually uses 12-16 field sized frame for production. 
I think changing the default camera dpi is more reasonable.

shun-iwasawa:
IMO both the modification in #675 and #676 are for making the look of default scene nicely, but please remember that the default value of camera dpi (53.33333) is for now very unnatural because of Stage::inch. And by following such unnatural dpi, the modifications may lose one of big advantages of this software - availability to work on virtual table and virtual camera-stand by showing images with actual sizes which is commonly used in animation industry.
Just changing the default camera dpi will solve the situation only for the case using non-pixel unit. We have to think of altering Stage::inch in order to make the UI reasonable with any unit being selected.

turtletooth:
> They usually uses 12-16 field sized frame for production. 
I think changing the default camera dpi (Stage::inch) is more reasonable.

Thanks for the info.  I did not know that.  I agree that changing Stage::inch makes more sense.

I was playing with different Stage::inch values and field guides- here are some shots:

Default field guide at Stage::inch = 160
160 default field guide

Widescreen AR field guide at Stage::inch = 160
160 12 wide field guide

Default field guide at Stage::inch = 120
120 default field guide

Size 16 field guide at Stage::inch = 120
120 16 field guide

Size 16 Widescreen AR field guide at Stage::inch = 120
120 16 wide field guide

What are your thoughts on Stage::inch values?  You have more experience in the animation industry- do they prefer size 12 or 16?  Does the AR need to stay the current value or can it be widescreen?

turtletooth:
I'm going to close this until more progress has been made on Stage::inch changes

PR #676, opened 2016-07-24
Updated table/disc size for new default resolution

turtletooth:
The drawing table/disc was totally covered by the new default resolution.  I updated it to fit.  
Here it is with default settings (1920 x 1080)
table

Here it is for 1280 x 720 at the same dpi.
table720

### DIFF
--- a/toonz/sources/toonz/viewerdraw.cpp
+++ b/toonz/sources/toonz/viewerdraw.cpp
@@ -765,7 +765,7 @@ unsigned int ViewerDraw::createDiskDisplayList() {
       sinCosTable[i].y = sin(ang);
     }
   }
-  double r = 10 * Stage::inch;
+  double r = 20 * Stage::inch;
   glNewList(id, GL_COMPILE);
   glColor3d(.6, .65, .7);
   glBegin(GL_POLYGON);


Rayek:
I solved this by changing the PPI value in the camera output setting to 120.

But I can see this is an issue when working with pixels only as a unit.

turtletooth:
@Rayek I also wanted to make it visible with default settings.

shun-iwasawa:
Is there any further suggestion for revising Stage::inch ?
IMO resizing the table is OK as intermediate workaround if you must... But it should be reverted once Stage::inch will be altered to more reasonable value. Because the camera stand view is simulating actual size and the animation disc with diameter of 40 inches is quite unnatural.

turtletooth:
> the animation disc with diameter of 40 inches is quite unnatural.

That is true unless you are a giant. :smile: 

> Is there any further suggestion for revising Stage::inch ?

I think that either 120 or 160 would work fine.  I prefer 120 because it makes the size an even 16 x 9 and at 160 the animation disc feels a little big - especially for projects smaller than 1920 x 1080.

discbig

I haven't implemented this myself because I don't know enough about how Stage::inch affects vector drawings.  I don't mind starting it if you can direct me to where and how Stage::inch relates to vector rendering.

turtletooth:
Closing this - changing Stage::inch is better

Issue #393, opened 2016-06-02
next and previous drawing commands (next/prev frame without holds) hard to discover


blurymind:
Right now when you use the next/previous frame commands, OT moves the timeline to the next and previous frame (jumps marked In red)
![nextprevframe](https://cloud.githubusercontent.com/assets/6495061/15745395/8a53f62e-28c9-11e6-9ea9-b210e4d29305.PNG)

If you have a frame hold for more than one frame, you have to hit the next/prev command a few times to get to the next/previous key drawing.

It would  be useful to have a next drawing and previous drawing command that jumps between drawings in the xsheet without the holds (jumps marked in yellow).
It should take the currently active level for reference of course.

blurymind:
Ah sorry, OT already has this command, but it is not available as a button in the viewer or any of the menus. The only way to discover it is via the keyboard shortcuts.

blurymind:
I remember setting them up to shift+z and shift+x actually. I suppose the point to make here is that they are not easy to discover by new users.

turtletooth:
Do you think having an option in preferences to enable a next and previous drawing button in the scene viewer would help?

blurymind:
@turtletooth it might help :) to be honest that is already overcrowded at the moment

blurymind:
Perhaps it would be more useful to have the option to set the default behavior for next/prev frame be set to next/prev drawing instead. I certainly use it much more often than next previous frame commands. If you look at other animation software's playback buttons you will see they do too

konero:
No, those buttons would be for composite checks, you would most likely have a dozen levels that don't share the same exposure, the functions you're talking about only apply to a single level, so those buttons would skip passed other levels.

Why not just use the standard used by everyone else? F/G keys (Flash and Harmony). I've noticed right-handed people use F/G and left-handed people use J/K because of the key bumps and so the rest of the keys remain on the inside of their hand rather than the outside; which is the natural way to use a keyboard. Since 90% of people or so are right-handed the standard would be F/G.

Whether it's a setup you or I would use is irreverent when considering the bigger picture. For anyone wanting alternate shortcuts then they can do so, but standards are used for preventing confusion and for making migration from other software a breeze, Toon Boom understood that well. :+1:

PR #677, opened 2016-07-25
Add Hidden Commands to Menus

turtletooth:
#393 
- Added some hidden commands to xsheet and cells menu bar.  This aids in discoverability and shortcut learning.
XSheet (It's the commands at the end):
xsheet
Cells (It's the drawing substitution commands):
cellsmenu

- Added these items to the menu bar customization window.  (I think anything you can remove, you should be able to add back)
customize


toonz/sources/toonz/mainwindow.cpp
```diff
@@ -1726,6 +1726,14 @@ void MainWindow::defineActions() {
...
   createMenuXsheetAction(MI_RemoveGlobalKeyframe, tr("Remove Multiple Keys"),
                          "");
+  createMenuXsheetAction(MI_FirstFrame, tr("First Frame"), "");
+  createMenuXsheetAction(MI_LastFrame, tr("Last Frame"), "");
+  createMenuXsheetAction(MI_PrevFrame, tr("Previous Frame"), "Shift+A");
+  createMenuXsheetAction(MI_NextFrame, tr("Next Frame"), "Shift+S");
+  createMenuXsheetAction(MI_NextDrawing, tr("Next Drawing"), "Shift+X");
+  createMenuXsheetAction(MI_PrevDrawing, tr("Prev Drawing"), "Shift+Z");
+  createMenuXsheetAction(MI_NextStep, tr("Next Step"), "");
+  createMenuXsheetAction(MI_PrevStep, tr("Prev Step"), "");
```
shun-iwasawa: Here you changed types of these commands from PlaybackCommandType to MenuXsheetCommandType. 
Is it only because you would like to include these commands into the menu bar customization popup?
If so, how about to make the menu bar customization popup to browse the playback commands additionally,  instead of changing command types? Because we already have too many Xsheet commands...

turtletooth: Good point.  I will make the change.

turtletooth:
Good call on the change @shun-iwasawa .  Makes much more sense.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

PR #680, opened 2016-07-25
Camera capture improvement - supporting "Show ABC Appendix to the Frame Number in Xsheet Cell" option

shun-iwasawa:
This PR is updates of the camera capture feature listed in #626.

- Fixed the bug that the countdown number flips when the "Upside Down" check box is active.

- Supported the preferences option "Show ABC Appendix to the Frame Number in Xsheet Cell".

With this option OpenToonz accepts an alphabet modifier after the frame number as follows:

Please note that it just changes the display mode for frame number. For example, the actual frame numbers stored in the first column in the above image are; 10, 20, 21, 30, 31, 32, 33, 40 and 50. Please see my comment in #626 for details.

### DIFF
--- a/toonz/sources/toonz/filmstrip.cpp
+++ b/toonz/sources/toonz/filmstrip.cpp
@@ -34,6 +34,7 @@
 #include "toonz/tscenehandle.h"
 #include "toonz/toonzscene.h"
 #include "toonz/levelset.h"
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
Jenkins

turtletooth:
LGTM

PR #683, opened 2016-07-26
WIP: Update Stage::inch to more practical value

turtletooth:
This makes Stage::inch = 120.  (I'm fine with 160 too).
The new default level size is 16 x 9 inches (still 1920 x 1080)

@shun-iwasawa I know that Stage::inch also affects vector sizes, and should be set to Stage::vectorDpi, but I can't find where that is happening.  I have looked a long time.  It's probably something very simple, but I don't see it.

### DIFF
--- a/toonz/sources/common/tunit/tunit.cpp
+++ b/toonz/sources/common/tunit/tunit.cpp
@@ -347,7 +347,7 @@ TMeasureManager::TMeasureManager() {
   m->add(cameraYFld.clone());
   add(m);
 
-  const double stage_inch = 53.33333;  // Consider changing to 120 or 160
+  const double stage_inch = 120;  // Consider changing to 120 or 160 (old value 53.33333)
 
   TUnit fxLength(L"fxLength"),
       fxInch(L"in", new TSimpleUnitConverter(1 / stage_inch)),
--- a/toonz/sources/include/toonz/tcamera.h
+++ b/toonz/sources/include/toonz/tcamera.h
@@ -39,9 +39,9 @@ class DVAPI TCamera {
 public:
   /*!
 Constructs TCamera with default value, size (12,9) and resolution (768,576).
-  Constructs TCamera with default value, size (36, 20.25) and resolution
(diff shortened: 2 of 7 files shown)


shun-iwasawa:
Thank you for starting this PR. I will look into it.

shun-iwasawa:
This is a quick note to let you know - I agree with you that 120 dpi would be nice for the alternative.

FYI, The Association of Japanese Animations (AJA) released a standard frame size for animation production ([AJA Fr01](http://aja.gr.jp/jigyou/chousa/aja_layout#)) which is using 192dpi, 10 x 5.625 inches. They set 10 inch in width in order to make the dpi conversion calculation easier and to make the frame fit in A4-size paper. A4-size paper is very common in Japan and using it can reduce costs.
I know it is too small to be default camera for OpenToonz, so I will put this size in camera presets afterwards.

konero:
We also use Aja Fr01 here. Celsys Clip Studio Paint's 1080p preset is set to this specification too.

csp2

shun-iwasawa:
I'm still investigating the side effects to be appeared by the change of Stage::inch as it should not cause any changes to the scene created with the previous version.
I found the following issues so far:
#### Interface becomes reduced in size

- Camera name text displayed top-left corner of the camera frame (e.g. "Camera1" or "Cleanup Camera")
- Numbers displayed in the field guide
- Handles of the Ruler tool
- Vector brush size

#### Data becomes reduced in size (= affects rendered result)

- Vector levels
- Mesh levels
- Motion paths
- Size/length/scale parameters for Fxs

I'm trying to fix the above inconsistencies. 
Some features can be fixed easily (by just enlarging 120 / 53.33333 = 2.25 times), but some parts (such as the motion paths) are not so simple as I expected.

turtletooth:
Thanks for looking into this.  I saw many of the same issues, but
especially for the vector size issues, I couldn't find where to fix it.

turtletooth:
Closing. Thanks @shun-iwasawa for the new PR.

PR #685, opened 2016-07-27
Brush Size Shortcuts Change

turtletooth:
This adds the ability to use Ctrl + Alt to change min and max brush sizes individually.  Holding Ctrl + Alt and mouse up/down changes the min brush size.  Left/right changes the max brush size.  

This also fixes the incremental shortcut keys for brush size change not working on raster levels and some UI inconsistencies.


nittolo:
@turtletooth removing the Alt shortcut is probably a good idea, as it seems to annoy a fair amount of users. And I think that the values should change more slowly. From what I see in the .gif, it seems too sensitive.

konero:
Corel Painter and Clip Studio Paint use Ctrl+Alt+LeftClick&Drag. In CSP it can be mapped to any modifier key or combination.

tz_mod

turtletooth:
@nittolo I'm not sure the gif is the best example of the speed.  It currently changes the same way the Alt shortcut does as far as speed goes.  If you think you have enough control with the Al shortcut, this should be fine.  I personally would like it slower, but I'm flexible.  It's an easy change.

toonz/sources/tnztools/brushtool.cpp
```diff
@@ -1616,28 +1616,40 @@ void BrushTool::mouseMove(const TPointD &pos, const TMouseEvent &e) {
...
+  //} else
+  if (e.isCtrlPressed() && e.isAltPressed()) {
+    const TPointD &diff = pos - m_mousePos;
+    double max          = diff.x / 2;
+    double min          = diff.y / 2;
+
+    locals.addMinMaxSeparate(
+        TToonzImageP(getImage(false, 1)) ? m_rasThickness : m_thickness, min,
+        max);
+  } else {
```
shun-iwasawa: I know it is the problem included in the original source and not of your PR, but could you please replace lines 1648-1649 by the following?:

`(m_targetType & TTool::ToonzImage) ? m_rasThickness : m_thickness, min, max);`

Because when the empty cell is selected, `getImage(false, 1)` returns 0 so `addMinMaxSeparate` will try to change `m_thickness` (= for vector brush) . But the current brush tool for creating new level is set according to the Preferences > Drawing > Default Level Type and it can be the raster brush.
So I think it would be better to use the target type for identifing the brush type (just like the same way done in line 1657 of this source).

turtletooth:
Made the requested changes.

shun-iwasawa:
@turtletooth I'm so sorry to trouble you but could you please apply clang formatting again since the beautification.sh is now fixed with #745 . Travis CI said that there is some unformatted part as follows:

> --- a/toonz/sources/tnztools/brushtool.cpp
> +++ b/toonz/sources/tnztools/brushtool.cpp
> @@ -1644,8 +1644,9 @@ void BrushTool::mouseMove(const TPointD &pos, const TMouseEvent &e) {
>      double max          = diff.x / 2;
>      double min          = diff.y / 2;
>  
> \-    locals.addMinMaxSeparate((m_targetType & TTool::ToonzImage) ? m_rasThickness : 
> \-           m_thickness, min, max);
> \+    locals.addMinMaxSeparate(
> \+        (m_targetType & TTool::ToonzImage) ? m_rasThickness : m_thickness, min,
> \+        max);
>    } else {
>      m_brushPos = pos;
>    }

Thanks in advance.

shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins again

shun-iwasawa:
Jenkins again

shun-iwasawa:

turtletooth:
@shun-iwasawa What are your thoughts on this modification?

shun-iwasawa:
Jenkins again

shun-iwasawa:
It works fine. Thank you @turtletooth and sorry for my late response! LGTM

PR #687, opened 2016-07-27
Allow Enter in Save File Field to Save

turtletooth:
This is a user experience PR.  Right now, to save, you type the name and then have to click save.  Pressing enter or return does nothing - unlike most other programs.

This allows you to hit enter to save instead of having to click Save.


turtletooth:
I changed the signal emitted in the first commit in case there was a reason the standard qt returnPressed signal was being bypassed.

RodneyBaker:
> This allows you to hit enter to save instead of having to click Save.

Thanks for this!  I had noted not being able to use Enter to launch the Save button and this will be a useful time saver and enhance workflow.  Every unnecessary move back and forth from mouse to keyboard should (reasonably) be avoided.  

I was considering putting a request in for Enter key functionality but stopped short adding a formal report... still wondering if there might be a reason for not having that keystroke work.  How might an animator accidentally hit the enter key and save over the top of an earlier work, etc.  It's hard to imagine a reason not to have the Enter key work in this fashion but I do try to think about what might happen out of the ordinary in the course of workflow.

shun-iwasawa:
As @RodneyBaker pointed, I disabled return key under the request from studio staffs in order to prevent them from overwriting file by mistake.
But it is true that many users are familiar with the way using return key in place of clicking return key. So l totally agree on this feature. We can disable it again (as a preference option) if users demand later.
Thanks @turtletooth !

toonz/sources/toonz/filebrowserpopup.cpp
```diff
@@ -407,7 +407,9 @@ TFilePath GenericLoadFilePopup::getPath() {
 //***********************************************************************************
 
 GenericSaveFilePopup::GenericSaveFilePopup(const QString &title)
-    : FileBrowserPopup(title) {}
+    : FileBrowserPopup(title) {
+  connect(m_nameField, SIGNAL(returnPressedNow()), this, SLOT(onOkPressed()));
```
shun-iwasawa: It is just a matter of taste, but how about to use `QAbstractButton::animateClick` slot like this ?

`connect(m_nameField, SIGNAL(returnPressedNow()), m_okButton, SLOT(animateClick()));`

Which shows the button-clicking animation before executing the saving process so that users can easily understand what process is invoked by their action.

turtletooth: I like it.  Nice touch.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. @turtletooth thanks again for improvement.

PR #691, opened 2016-07-29
Shortcuts Selection Choice and Proposed New Shortcuts

turtletooth:
This adds the ability to set the shortcuts to Adobe Animate or Toon Boom Harmony shortcuts as well as reset them to OpenToonz.

This also adds proposed new shortcuts for OpenToonz.  I tried to keep basically everything the same, but there were a few that seemed to make sense to change, either because they seemed to be the standard, or they seemed to make more sense.

You can see the proposed new shortcuts here:
https://docs.google.com/spreadsheets/d/1OgILbe6eOvHnBRSW_iV_jIf9JDeKVWagwSa93TpB3D4/edit#gid=0
I don't think these necessarily the best- just my thought on them.  Please share your feedback.

The notable shortcuts that change from the current ones:
Ctrl + S is now Save All
P is now Play instead of Pinch
Previous Frame is now , (comma)
Next Frame is now .
Previous Drawing is <
Next Drawing is >
First Drawing is Ctrl + ,
Last Drawing is Ctrl + .

Using comma, period, and < and > for navigating frames seems to be the standard between Toon Boom and Adobe.

presetbox

presetoptions

A warning does pop up that tells you that this will change all custom shortcuts:
overwrite

I am not absolutely sure my defaults for Toon Boom and Adobe are mapped to the best feature fit in OpenToonz.  If I am off or missed something, please let me know.

This also fixed a bug where removing a default shortcut wasn't permanent if more than one default shortcut was removed.  (Only one removed shortcut was saved to the shortcuts.ini file)


blurymind:
That is quite nice ! Thank you for implementing. Would it be possible for the user to create their own preset, save it, back it up, share it?

That would make it solid imo

turtletooth:
That's not too hard at all. Let me look into it.

turtletooth:
@blurymind It's there now:
exportshortcuts

loadshortcuts

turtletooth:
Note- I haven't actually added the default shortcuts to the mainwindows.cpp file.  I will do that after any discussion about shortcuts.

konero:
> Previous Drawing is <
> Next Drawing is >
> First Drawing is Ctrl + ,
> Last Drawing is Ctrl + .

This is a bad fit IMO, traditional animators will use Next Drawing and Previous Drawing the most when animating so they shouldn't be mapped to a modifier key combination. Perhaps J+K or K+L?

Since @shun-iwasawa is looking to make OpenToonz more compatible with PaintMan users, how about using the PaintMan mapping?

```
Previous Drawing is "," (comma)
Next Drawing is "." (period)
First Frame is Alt + <
Last Frame is Alt + >
```

Even when I do marionette/switch/3DCG animation I haven't seen people using Next and Previous Frame very often, these could be mapped to Shift+</>? Others will need to chime in.

I like the Shortcut Mapping chooser.

I would change "Use Preset" to just "Presets" and bring some of the actions to the front.

tz_shrtc_w

I'd make a remove all warning.

"**This will remove all shortcuts to the currently selected preset. Are you sure you want to continue?**".

Does the preset change need a warning? In CSP it overwrites the actual mappings so a warning makes sense because you can lose everything, but I think your implementation keeps them in separate files so they're quite safe I would have thought. CSP too can keep them in separate files, but you need to make sure they are saved first in a workspace.

Good PR though. :+1:

turtletooth:
> Does the preset change need a warning?

My system has the presets in a file, but any custom changes to shortcuts are not remembered when choosing a preset.  If I set play to be Ctrl+Alt+Shift+PgDown, it will be lost if I choose a preset.  (Please don't ever make that the shortcut for play.)  So I think a warning is necessary.  

I agree with you on the shortcuts, especially considering the PaintMan focus.  Actually, I actually use next and last frame often, but I don't actually use the shortcuts, since the arrow keys navigate the xsheet.  So, it doesn't affect me either way.

I like the idea of making remove all shortcuts a separate button, but I'm leaning toward keeping "Load from File" in the dropdown.  My implementation doesn't actually add a preset to the dropdown, but simply imports the presets stored in an external file and sets them as the current shortcuts.  If it actually added a preset to the dropdown, then I could see import being a separate button, but then the presets could get cluttered and you would need a way to delete presets - which makes for a increasingly complex UI.

@konero - I don't have access to the PaintMan shortcuts and all the documentation I find is in Japanese.  Would you be willing to either tell me the mappings or set your shortcuts to the PaintMan shortcuts and send me your shortcuts.ini file?  I would love to add PaintMan as a preset.

konero:
@turtletooth Sure, I can make a PaintMan file, I'll post it here soon.

Just for reference, here is what the shortcut settings window looks like in Adobe Animate.

tz_an

The two icons at the top are...

```
1. Save the current shortcuts set under a new name.
2. Delete the current shortcuts set.
```

There is also a "Default Set" which says it is "read-only".

If I set a new shortcut to something and try to switch presets, I get this warning...

tz_wi

The shortcuts sets are saved to separate xml files.

turtletooth:
Alright- here's what's new:
I made the shortcut changes @konero suggested regarding next and previous drawing/frame and first/last frame.

I added the ability to save the current shortcuts as a new preset.
I added the ability to delete user made presets (can't delete built in presets)
Presets aren't actually made active until the apply button is pressed (This allows to select a preset for deleting without applying it first)
Exporting shortcuts still works
Clearing all shortcuts is its own button.

newpresets

newpresetsdropdown

addedpreset

There is still a warning before applying a preset in case a user has customized any shortcuts after choosing the preset.

turtletooth:
Still not done: 
- RETAS Shortcuts
- Setting default shortcuts in mainwindow.cpp (waiting for feedback)

turtletooth:
You can try these out in the latest portable.

blurymind:
@turtletooh Thank you!! :)

shun-iwasawa:

turtletooth:
@shun-iwasawa Would you have the time to look at this one?  The list of suggested default shortcuts is at: https://docs.google.com/spreadsheets/d/1OgILbe6eOvHnBRSW_iV_jIf9JDeKVWagwSa93TpB3D4/edit?usp=sharing

turtletooth:
I took RETAS out of the dropdown for now until the shortcuts are finished for it.

toonz/sources/toonz/shortcutpopup.cpp
```diff
@@ -415,4 +471,320 @@ void ShortcutPopup::onSearchTextChanged(const QString &text) {
...
+    return;
+  }
+
+  if (!showConfirmDialog()) return;
+  showDialog("Setting Shortcuts");
+  if (preset == "OpenToonz") {
+    clearAllShortcuts(false);
+    TFilePath fp = TEnv::getConfigDir() + TFilePath("defopentoonz.ini");
+    setPresetShortcuts(fp);
+    return;
```
shun-iwasawa: For now, default shortcut preset files (defopentoonz.ini, otharmony.ini, etc.) are stored under `$TOONZCONFIG` directly.

How about moving them to some folder under `$TOONZPROFILES `? Because the user's shortcuts data are saved in `$TOONZPROFILES/layouts/settings.username/shotcuts.ini` and it would be clear to save them in the same folder.

I think `$TOONZPROFILES/layouts/defaultshortcutpresets` would be nice as the default room presets are also saved in `$TOONZPROFILES/layouts/rooms`. Maybe folder name can be shorter, but I have no idea...

toonz/sources/toonz/shortcutpopup.cpp
```diff
@@ -415,4 +471,320 @@ void ShortcutPopup::onSearchTextChanged(const QString &text) {
...
+          CommandManager::instance()->getShortcutFromAction(action);
+      shortcutString = shortcutString + QString::fromStdString(id) + "=" +
+                       QString::fromStdString(shortcut) + "\n";
+    }
+  }
+  QFile file(fp.getQString());
+  file.open(QIODevice::WriteOnly | QIODevice::Text);
+  QTextStream out(&file);
+  out << shortcutString;
+  file.close();
```
shun-iwasawa: I think using `QSettings` would be easier to maintain.
Please refer to the function `CommandManager::setShortcut()` at the line 97 of `menubarcommand.cpp`.

Furthermore, how about to skip the commands with no shortcut and save the commands with shortcuts only?

toonz/sources/toonz/shortcutpopup.cpp
```diff
@@ -415,4 +471,320 @@ void ShortcutPopup::onSearchTextChanged(const QString &text) {
...
+  QAction *action;
+  for (QString id : allIds) {
+    QByteArray ba      = id.toLatin1();
+    const char *charId = ba.data();
+    action = CommandManager::instance()->getAction((CommandId)charId);
+    CommandManager::instance()->setShortcut(
+        action, preset.value(id).toString().toStdString(), false);
+  }
+  preset.endGroup();
+  m_sViewer->shortcutChanged();
```
shun-iwasawa: Please add `emit` before `m_sViewer->shortcutChanged();` in order to make it easier to understand that this line calls signal... Or you can call `m_list->onShortcutChanged()` directly since for now this signal is only for updating ShortcutTree.

toonz/sources/toonz/shortcutpopup.cpp
turtletooth: I think that makes sense.  I will move them.

turtletooth: What do you think about simply calling the folder shortcuts, since the default rooms are simply called rooms?

turtletooth: I think your version is definitely more clear, but I wanted to throw out a suggestion based on what is currently in there.

toonz/sources/toonz/shortcutpopup.cpp
shun-iwasawa: Using text stream is fine if QSettings was slow. Thank you for testing!

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. @turtletooth Thank you for adding this! It's very nice.

Issue #486, opened 2016-06-21
bug: onion skinning should be disabled during playback


blurymind:
With the latest version of opentoonz (today's nightly) When you hit the play button onion skinning is still active.

It should switch off automatically during playback and switch back on when playback is stopped.

turtletooth:
I think it has always done this, but I agree that it is not super useful during playback.  I can think of a few times when it has been nice to see.  There is a quick toggle in keyboard shortcuts but it isn't set by default.

blurymind:
perhaps this could be a setting in the preferences (onion skin tab) window?
"Disable during playback"

I dont like having an extra step every time i start and stop playback and would like to have it disabled during 99% of the time when the animation is running. Having it on is distracting.

turtletooth:
I agree with you that it should be generally disabled during playback.  If it is an option I think it should be disabled by default during playback.

blurymind:
i agree - "disable during playback" should be ticked on default!

turtletooth:
#708 should take care of this.

blurymind:
AWESOME! :o

PR #708, opened 2016-08-05
Disable onionskin during playback

turtletooth:
This disables onion skin during playback by default.

It can be allowed during playback via preferences, but will still honor the current state (on/off).  

onionskinplayback


turtletooth:
#486

blurymind:
@turtletooth you are a damn hero! This has bugged me for so long. :D

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

PR #709, opened 2016-08-05
Switched sceneviewer with comboviewer to sync playback

### LINKED ISSUE #592: Inverted colors in raster level. Linux.
Adding a raster level, and use the brush, some colors are inverted: 

https://dl.dropboxusercontent.com/u/69554201/OpentoonzLinuxWrongColorsBug.mp4

### LINKED ISSUE #692: Set start/stop markers in 1 room, visible but ignored in others
I animated something over a number of frames, say 20 frames, and set start and stop markers  at 4 and 12, for example. 

In the room I set it in, it works normally when playing in the viewer and only runs between 4-12.  The First frame button brings me to 4 and Last frame button to 12. 

If I switch to another room, I can see the start/stop markers at 4 and 12 where I set them.  However, when I play in the viewer in that room, it completely ignores them.  First frame button takes me to 1 and Last frame button to 20.

Additionally, if I reset one of the markers to where it already is, then it recognizes both 4 and 12, but only in that room.

If I change one of the markers in the 2nd room (changed 4 to 2), when I go back to the 1st room where I set the markers originally, it still only recognizes 4-12 but the start marker is clearly set to 2.

I don't recall it working that way in prior versions and only noticed it on the Nightly build I installed.

Using Windows version 1.0.3 (built Jul 30 2016 11:03:49)

turtletooth:
This fixes #692 

The sceneviewer wasn't honoring the markers, but the combo viewer does.  This switches each room to use a comboviewer and hides the parts that are already in the room.


toonz/sources/toonz/mainwindow.cpp
```diff
@@ -2344,9 +2378,9 @@ void RecentFiles::saveRecentFiles() {
 
 QList<QString> RecentFiles::getFilesNameList(FileType fileType) {
   QList<QString> files =
-      (fileType == Scene)
-          ? m_recentScenes
-          : (fileType == Level) ? m_recentLevels : m_recentFlipbookImages;
+      (fileType == Scene) ? m_recentScenes : (fileType == Level)
```
turtletooth: This is an example of #715 - These are the results of running clang-format using the beautification scripts.

shun-iwasawa:
Sorry for my too late response on your modification.
It is OK to use ComboViewer instead of Viewer.
But since this issue #692  is caused by a bug in Viewer, IMO it would better to resolve the bug itself if you don't need toolbar or tool options bar contained in ComboViewer.

Looking into the difference between ComboViewer and Viewer, only the ComboViewer calls `onSceneChanged()` in `showEvent()`. The function `m_flipConsole->setMarkers()` is called in `onSceneChanged()` so that the preview range is updated.

So, how about to modify viewerpane.cpp as follows?:

> @@ -228,7 +228,7 @@ void SceneViewerPanel::showEvent(QShowEvent *) {
>    TObjectHandle *objectHandle  = app->getCurrentObject();
>    TXsheetHandle *xshHandle     = app->getCurrentXsheet();
>  
> -updateFrameRange();
> +onSceneChanged();
>  
>    bool ret = true;

Thanks.

turtletooth:
@shun-iwasawa Thanks.   I didn't know if the scene viewer behavior was intentional, so I didn't touch it.  I will update the change.

turtletooth:
Closing - updated for #778

PR #711, opened 2016-08-05
Fix appvayer to work for v1.1.0

kogaki:
This will fix the appvayer CI, which might have been broken from #704 .


skitaoka:
LGTM

PR #718, opened 2016-08-08
v1.1.1

shun-iwasawa:
#This is an emergency release for updating SystemVar.ini used only in OSX


kogaki:
Looks OK. Let's go on the release.

PR #723, opened 2016-08-10
Add format string literal

ckorn:
Prevents security issues when the string contains
format specifiers.
Fixes: warning: format not a string literal and
no format arguments [-Wformat-security]


ideasman42:
LGTM, note that we could use -Wall so these are printed by default:
https://gist.github.com/ideasman42/c0771332819494e0c6e9475d129cd32d

PR #725, opened 2016-08-12
Quick Export Feature

turtletooth:
This adds a feature to export a mp4 video w/ one click if ffmpeg is installed.  No need to set anything in output settings.

I am building this for personal use, but thought others might like it.


ideasman42:
Quick export could be an OK feature, but why limit to the desktop?

Exact output location could be a preference (default to the desktop when left blank for eg).

turtletooth:
@ideasman42 I like the way you think.  I'll add an option to choose the location.

turtletooth:
I added an option to choose a location.  It defaults to desktop if nothing is chosen.

RodneyBaker:
Very nice.  :)

shun-iwasawa:

turtletooth:
@shun-iwasawa After sound playback, would you mind looking at this one next.  Thanks for all you do.

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -1117,8 +1124,9 @@ PreferencesPopup::PreferencesPopup()
 
   //--- Import/Export ------------------------------
   categoryList->addItem(tr("Import/Export"));
-  m_ffmpegPathFileFld = new DVGui::FileField(this, QString(""));
-  m_ffmpegTimeout     = new DVGui::IntLineEdit(this, 30, 1);
+  m_ffmpegPathFileFld       = new DVGui::FileField(this, QString(""));
+  m_fastRenderPathFileField = new DVGui::FileField(this, QString(""));
```
shun-iwasawa: Clicking "..." button (Open browser button) of this field takes extra time before opening the browser popup, since it tries to find non-existent path "desktop" as initial folder position.
In PR #853, I modified `DVGui::FileField`  constructor to be available to set `doNotBrowseInitialPath` flag for avoiding this behavior.

How about changing line 1128 like as follows?
`m_fastRenderPathFileField = new DVGui::FileField(this, QString("desktop"), false, true);`

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -1755,6 +1765,14 @@ PreferencesPopup::PreferencesPopup()
         ioGridLay->addWidget(new QLabel(tr("FFmpeg Timeout:")), 4, 0,
                              Qt::AlignRight);
         ioGridLay->addWidget(m_ffmpegTimeout, 4, 1, 1, 3);
+        ioGridLay->addWidget(new QLabel(" "), 5, 0);
+        ioGridLay->addWidget(
+            new QLabel(tr("Please indicate where you would like "
+                          "exports from Fast Render(MP4) to go.")),
+            6, 0, 1, 4);
+        ioGridLay->addWidget(new QLabel(tr("Quick Render Path: ")), 7, 0,
```
shun-iwasawa: IMO the term associated with this feature should be standardized to "Fast Render" or "Quick Render".

toonz/sources/toonz/mainwindow.cpp
```diff
@@ -1601,6 +1601,7 @@ void MainWindow::defineActions() {
   createMenuFileAction(MI_OutputSettings, tr("&Output Settings..."), "Ctrl+O");
   createMenuFileAction(MI_PreviewSettings, tr("&Preview Settings..."), "");
   createMenuFileAction(MI_Render, tr("&Render"), "Ctrl+Shift+R");
+  createMenuFileAction(MI_FastRender, tr("&Fast Render"), "Alt+R");
```
shun-iwasawa: I'm a little concerned that the term "Fast Render" may give the false impression to users that this command will render the image in a short computation time.
How about to add "MP4" to the command name (like "Fast Render to MP4" or "Quick Render to MP4") in order to clarify the role of this command?

toonz/sources/toonz/preferencespopup.cpp
turtletooth: Great catch

toonz/sources/toonz/preferencespopup.cpp
turtletooth: agreed

toonz/sources/toonz/mainwindow.cpp
turtletooth: agreed

turtletooth:
Changes submitted.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

PR #733, opened 2016-08-14
Remove Are You Sure? when closing without unsaved content.

turtletooth:
If there is a good reason to ask "Are you sure?" when closing, I don't mind closing this, but I can't think of any other software that asks this when there is nothing to save.


shun-iwasawa:

turtletooth:
@shun-iwasawa What are your thoughts on this one?

shun-iwasawa:
I remember that I added this dialog to the Ghibli Edition because Toonz 4.6 has the same behavior. I agree it is unnecessary any more. LGTM

PR #734, opened 2016-08-14
Fit scene to camera on New Scene and Load Scene

turtletooth:
When a different scene is loaded (or new scene) this fits it to the camera view.


turtletooth:
This is probably unnecessary if the new Stage::inch pr is merged.

turtletooth:
Not needed anymore.  Closing

PR #735, opened 2016-08-15
WIP: Update Stage::inch to more practical value (original PR by tutletooth)

shun-iwasawa:
This PR is continued from PR #683 by @turtletooth 

In order to change Stage::inch, I took care of the following aspects to be kept compatibility with older versions:

#### Interface becomes reduced in size

- Camera name text displayed top-left corner of the camera frame (e.g. "Camera1" or "Cleanup Camera")
- Numbers displayed in the field guide
- Handles of the Ruler tool
- Vector tool sizes
- 3D camera view

#### Data becomes reduced in size (= affects rendered result)

- Vector levels
- Mesh levels
- Motion paths
- Size/length/scale parameters for Fxs

Please note that I updated the version of scene file from 71.0 to 71.1 in order to keep backward compatibility of the motion paths and the fx parameters by multiplying values on loading scenes. Which means that after merging this PR the scene saved with the latest version should not be loaded with the older versions or it will get wrong result.

@turtletooth I still marked this PR as "WIP" since I may overlook something. Could you please check if this PR works properly? Especially I 'm not really sure about the plastic tools.


turtletooth:
I'll look at this tonight (tomorrow for you) and see how it works.  My suggestion if it seems to be ok is to merge it soon.  This will give plenty of time to find hidden bugs between now and the next minor release.  At first glance, it seems very thorough.  

My plan tonight is to make a project in the older version that uses all the things that you said were modified by this PR.  I will then open that project in this PR's version and see if it seems to work right.  If it seems ok to me, do you mind if I merge it and then we can take care of bugs as they come up?  This would give more people a chance to try it and point out any areas that might have been missed.

Two totally unrelated questions:
What are you using to compile less to qss?  I have been playing with a different theme but I have been doing straight qss since I can't find a compiler that makes clean qss.

Do you know what the timeline is for review on my other PRs?  I have them merged in my own version, but I would much rather be able to work on just one code base.

turtletooth:
@shun-iwasawa 
Here are my results so far.
It looks like it is mostly ready.  However, loading a mesh with bones from a previous version, the bones were not the right size.

Here is the original:
old_version_mesh

Here is it loaded in the new version:
new_mesh

At first it also appeared the ruler was giving the wrong lengths.  For example on a scene at 1920x1080:
ruler_old

I went to check the camera settings to make sure the scene was actually 1920x1080, and the camera's box jumped down in size when I opened camera settings.  
after_camera_settings

It looks like things for for new levels created.  I did get a crash on revert scene, but I have seen that before (can't duplicate it consistently) and I don't think it is related.

shun-iwasawa:
Thank you very much for pointing the problem. Last night I thought about another idea to realize the comfortable UI with pixel unit without changing the Stage::inch. I will try to implement it right away.

> What are you using to compile less to qss? I have been playing with a different theme but I have been doing straight qss since I can't find a compiler that makes clean qss.

I'm using Crunch 2 ( https://getcrunch.co/ ) for the less compiler.

> Do you know what the timeline is for review on my other PRs? I have them merged in my own version, but I would much rather be able to work on just one code base.

Sorry for keep you waiting for your PRs to be merged. I'll use several hours for reviewing (hopefully) 2-3 PRs per day. Thank you for your patience.
If there are any PRs which you would like to put priority, please let me know.

shun-iwasawa:
Close this PR as we take the alternative resolution #739

PR #739, opened 2016-08-16
Introducing Stage::standardDpi instead of updating Stage::inch

shun-iwasawa:
This PR is an alternative idea for making UI comfortable when using "pixel" unit. (replacement of #683 and #735 )

@turtletooth  I think this would be simpler resolution than altering Stage::inch. Sorry for trouble you again but could you please check if I don't miss anything?
Please note in advance that IMO the camera dpi with "pixel" unit must be changed from 53.33333 to 120 anyway, as it is an original intent of this modification. So the scene created by old versions with "pixel" unit environment would lose compatibility. For now I do not take any measures for such problem...


shun-iwasawa:
Jenkins

turtletooth:
I love this idea.  I think it is much easier than trying to find every area affected by Stage::inch.  I will look at it tonight.

turtletooth:
This seems to be a much smoother (and safer) change.

There is still a weird glitch when you open a scene saved from a previous version- it opens with the old dpi, and if you go to Camera Settings, it jumps down to the new dpi.  

I realize that the new change doesn't guarantee compatibility between versions, but this makes a strange change in the middle of an opened scene without the user actually changing anything.

toonz/sources/toonzlib/stage.cpp
```diff
@@ -69,8 +69,8 @@ typedef std::vector<Player> PlayerSet;
    thickness of
                 images .pli.
 */
-const double Stage::inch      = 53.33333;  // consider changing to 120 or 160
-const double Stage::vectorDpi = 53.3333;
+const double Stage::inch        = 53.33333;  // consider changing to 120 or 160
```
turtletooth: I would take out the "consider changing comments" - this would confuse future developers

turtletooth:
The glitch I mentioned only appears when pixels only mode is on.  

Do you think there would be a way to show a dialog:

This scene dpi is incompatible with pixels only mode.  What would you like to do?
Turn off pixels only mode
Change the scene dpi

toonz/sources/toonzlib/stage.cpp
shun-iwasawa: Absolutely. I'll remove the comment.

shun-iwasawa:
> Do you think there would be a way to show a dialog:
> 
> This scene dpi is incompatible with pixels only mode. What would you like to do?
> Turn off pixels only mode
> Change the scene dpi

OK. I'll try to implement it.

shun-iwasawa:
I added a dpi incompatibility warning on loading scene.

For now selecting "Change the camera dpi" option will change the dpi of current camera to 120.
Please note that it is not enough to reproduce the scene saved with the previous version with pixels only mode.
You need to adjust some values manually such as dpi of raster levels and mesh levels, scale of vector levels, sizes of motion paths and skeletons, fx parameters etc. Would you like them to be changed automatically as well?

Another option would be to change the pixel size of the current camera according to the dpi difference.
For example, if you have a scene with 1920x1080 pixels size under the previous version, loading such scene will make the camera size to 4320x2430  (120/53.33333＝2.25 times enlarged). I think it is very easy to implement and can keep  compatibility of rendering result except that it is enlarged.
Please let me know your thought.

turtletooth:
> Another option would be to change the pixel size of the current camera according to the dpi difference.  

It sounds like this is the safer option.  The first option sounds like it would have the same potential for unexpected side effects like we saw when trying to change Stage::inch.  Simply resizing the camera seems much easier and safer for the maintaining the scene.  

With that, should the question be changed to something like:
This scene was created in an earlier version of OpenToonz and is incompatible with pixels only mode. What would you like to do?
What would you like to do? 
Turn off pixels only mode
Keep pixels only mode on and resize the scene

(I know the second option is long, but I don't know how to make it shorter)

The new question phrasing removes mention of dpi, which might confuse people who don't understand that even in pixels only mode, there is an underlying dpi.

I can take a look at this tonight.  I really like this solution over changing Stage::inch

shun-iwasawa:
OK, I will implement the second option.

> This scene was created in an earlier version of OpenToonz and is incompatible with pixels only mode. What would you like to do?
> - Turn off pixels only mode
> - Keep pixels only mode on and resize the scene

The popup will appear when you load scene with non-120 dpi.  It can occur not only when the scene is created in an earlier version, because you can create scene with any dpi if you are not in pixels only mode.
So I'll change the question like as follows:

> This scene is incompatible with pixels only mode of the current OpenToonz version. What would you like to do?

shun-iwasawa:
Jenkins

turtletooth:
Is this ready to be checked?  It looks like travis had some sort of issue.

turtletooth:
LGTM

PR #745, opened 2016-08-19
Fix beautification.sh

shun-iwasawa:
This is to fix "no such file or directory" error.


turtletooth:
Ran into this error tonight.

skitaoka:
LGTM

PR #748, opened 2016-08-21
Spritesheet Export

turtletooth:
Features: 
- Crops transparent area from images (if other images will use a space, it is left) - See ball example below.
- Set output format (Grid, Vertical, Horizontal)
- Scale output
- Set padding for top, bottom, left, right
- Export text file along with spritesheet with info about image sizes, padding etc. . .

Note- this is not an image atlas where every layer is exported into its own area of an image.  This is a traditional spritesheet.  Tested with Game Maker Studio.

Settings box:
sprite_settings

Examples of output:
example_spritesheet

Text file contents:
sprite_text_output

Example of trimming but keeping space if used by other images:
untitled48

### DIFF
--- a/toonz/sources/image/CMakeLists.txt
+++ b/toonz/sources/image/CMakeLists.txt
@@ -24,6 +24,7 @@ set(HEADERS
     ffmpeg/tiio_webm.h
     ffmpeg/tiio_mp4.h
     ffmpeg/tiio_ffmpeg.h
+    sprite/tiio_sprite.h
     mesh/tiio_mesh.h
 )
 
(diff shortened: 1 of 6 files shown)


blurymind:
this is incredibly useful to have! Thank you @turtletooth

turtletooth:
Branch created in wrong repository- closing

RodneyBaker:
Spritesheets... that's a great addition!  Thanks Jeremy!

blurymind:
@turtletooth  It is indeed very useful. I do game art as freelance sometimes and this feature saves me the hassle to pack the frames. It also makes opentoonz more useful on gamejam events with 48 hr deadline - where saving time is important

PR #751, opened 2016-08-23
Add a load image button to camera capture for reference image

turtletooth:
This feature allows the loading of a reference image into the onion skin for camera capture.  This is useful in stop-motion work for lining up your camera with a previous image.


turtletooth:
@shun-iwasawa For playback of camera capture, do you think it would be better to load each image from the level or to keep a vector of images as they are captured?  I know memory would be a concern so I am trying to find a balance between memory and performance.

shun-iwasawa:
@turtletooth  
Regarding the multi-frame onion skin and the playback features, I'm thinking of as follows:

- Keeping all previously-captured images in the camera capture window instance is not appropriate as it is memory consuming.
- Instead, it would better to use TImageCache and leave the memory management up to it.
- In normal use, captured images will line up in the current column. Which means that the multi-frame onion skin or playback can be realized by retrieving the contents of the current column and show them in the camera capture window.
- If you would like to continue shooting your stop-motion work from where you left off yesterday, just load the scene and set the current column/frame to the next frame of the cell containing the latest-captured image. Which enables to load previous images into the camera capture window, just like the normal onion skin in the scene viewer.
- The previous images will be loaded from the file and store in the cache only when the user demanded to show them.

How do you think about this idea?

turtletooth:
Sounds good- I have no idea how to do it, so if you could- that would be great.  But the idea makes sense.

shun-iwasawa:

turtletooth:
@shun-iwasawa I realize this has conflicts to be resolved, but what do you think about using this fix until a more complete solution can be implemented?

toonz/sources/toonz/penciltestpopup.cpp
```diff
@@ -1288,6 +1293,71 @@ void PencilTestPopup::onOnionCBToggled(bool on) {
...
+    void* buffer = malloc(totalBytes);
+    memcpy(buffer, buffin, totalBytes);
+
+    image->getRaster()->unlock();
+
+    QImage qi = QImage((uint8_t*)buffer, m_lx, m_ly, QImage::Format_ARGB32);
+    QImage qi2(qi.size(), QImage::Format_ARGB32);
+    qi2.fill(QColor(Qt::white).rgb());
+    QPainter painter(&qi2);
+    painter.drawImage(0, 0, qi);
```
shun-iwasawa: For now, if you are capturing images with the "Upside down" option, this button loads the previous image in the opposite direction.
How about to rotate the image if the "Upside down" option is ON, by adding lines like as follows?:

`    if (m_upsideDownCB->isChecked()){`
`      painter.translate(m_lx/2, m_ly/2);`
`      painter.rotate(180);`
`      painter.translate(-m_lx / 2, -m_ly / 2);`
`    }`

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. Thank you, @turtletooth  !

Issue #173, opened 2016-04-13
Idea - Open toonz json export - for games engines cut out animation


blurymind:
Right now the open source community is looking for an open standard json file format that stores animation data in the form of keyframes.

We have an addon for blender that does that already:
https://github.com/ndee85/coa_tools
It is compatible with Godot too!

**What are the technical advantages of having game animation stored as a json file+ an image atlas?**
- Your game uses much less memory - as only one frame is used and the animation (movement of body parts) is being driven by the game engine's runtime. You get better framerate!
- Your animation can be procedurally altered during runtime - since the runtime is driving it, the programmer can layer ontop or override animations. For example you can have your player character's head turn in the direction of a specific object on the screen.
- Your game/app uses much less HD space - smaller file size makes it much better for mobile development where file size is important.
- You can swap body parts of your character, as the animation is running. This makes it very easy to implement an equipment system that actually affects the way your character looks or reuse the same animations an all the enemies you want in the game - while having them look different.

**What other software can use this technique and export animation data to a json file?**
- Spine 2d
- Spriter
- Creature
- anime studio
- toonboom harmony

**Popular games that use it:**
- rayman origins, rayman legends,  dragon's crown, etc etc

**What open source software can do this already?**
- Godot and Blender (via coa tools)

**The problem with the json file format, the game runtime and the lack of one open standard:**
The three major competing applications for cut out game animation do basically the same things. However as a means to lock their users to their product, the structure of the json files they output is different. That forces game developers to have to rewrite a new run time for each of the three - that parses the data accordingly. What is worse - they put a license on their runtimes that prohibits anyone else  - who does not own the license to their software to use their runtimes.

We need a json file specification that is open - one that can work on all game engines without the need to purchase a license from someone.

If opentoonz gains the ability to export cut out animation data to a json file - it will be good if the effort is coordinated with godot and coa_tools developers on their respected trackers.

hako:
I think this is a good idea, OpenToonz does come with a scripting engine and JSON is available there. That could be a good place to start. :) :+1:

blurymind:
I am trying to coordinate this between godot, coa tools and opentoonz developers.

Here is the post at coa_tools. Ndee has already started making such a file btw. He has an importer and exporter, but his json file implementation still lacks ability to store mesh data and thus have done deformation.
https://github.com/ndee85/coa_tools/issues/26

[...]

PR #752, opened 2016-08-24
Spritesheet Export

turtletooth:
Features: 
- Crops transparent area from images (if other images will use a space, it is left) - See ball example below.
- Set output format (Grid, Vertical, Horizontal)
- Scale output
- Set padding for top, bottom, left, right
- Export text file along with spritesheet with info about image sizes, padding etc. . .

Note- this is not an image atlas where every layer is exported into its own area of an image.  This is a traditional spritesheet.  Tested with Game Maker Studio.

Settings box:
sprite_settings

Examples of output:
example_spritesheet

Text file contents:
sprite_text_output


blurymind:
Would it be possible to make the trimming optional in this case? It works
for atlases but for spritesheets it may cause issues if different frames
are a different size and centering each frame.

On 24 Aug 2016 04:22, "Jeremy Bullock" <notifications@github.com> wrote:
>
> Features:
>
> Crops transparent area from images (if other images will use a space, it
is left) - See ball example below.
> Set output format (Grid, Vertical, Horizontal)
> Scale output
> Set padding for top, bottom, left, right
> Export text file along with spritesheet with info about image sizes,
padding etc. . .
>
> Note- this is not an image atlas where every layer is exported into its
own area of an image. This is a traditional spritesheet. Tested with Game
Maker Studio.

turtletooth:
>Would it be possible to make the trimming optional in this case? It works
for atlases but for spritesheets it may cause issues if different frames
are a different size and centering each frame.

I can make the trim optional, but it already accounts for different sized frames and equalizes the frame size based on the most extreme drawing in each direction.  It calculates the area that is used by the entire animation and makes the image size match.  Each drawing is in its relative space within that area.

blurymind:
Ah i see..sounds like it will work on. Need to try it out

turtletooth:
You can try it here: https://github.com/motionmonster/motionmonster/releases/tag/v1.1.1.1

turtletooth:
Here is a better example- here is a partial spritesheet from this gif:
animated-parrot-gif-2

untitled70

RodneyBaker:
Sweet.   Very nice.

blurymind:
very nice indeed! :) Looking forward to seeing it merged. Thank you @turtletooth

shun-iwasawa:

RodneyBaker:
This spritesheet option for output still hasn't been added to OpenToonz yet but would certainly be a welcome addition.

Here's a quick sequence of frames/sketches rendered with it:

bunnyspritesheet_sm

turtletooth:
@RodneyBaker good point.

blurymind:
@turtletooth Now if we only had dragonbones json animation data file export :)
https://github.com/opentoonz/opentoonz/issues/173
That would be amazing!

Btw ndee has had success with writing an exporter for blender in python
https://github.com/ndee85/coa_tools/issues/26

turtletooth:
@blurymind COA Tools is so powerful - I'm not sure I see the benefit of duplicating the features here.  What do you see as something that OpenToonz has the potential to offer that COA Tools doesn't support?

turtletooth:
@shun-iwasawa Would you mind looking at this next when you have the time?  

On a side note, there is an Android game that was created using this feature:
https://groups.google.com/forum/#!category-topic/opentoonz_en/share-your-work/Y3Xoaa2dSik

turtletooth:
@shun-iwasawa I made a copy of the image in the same function the buffer was created in.  QImage only has the memory problem on images created with a buffer, so the copy is easy to delete later.  Instead of figuring out a way of passing the pointer to the buffer to the destructor, this seemed like an easier fix.  I tested the memory usage and memory usage drops back to the right place after spritesheet export.

shun-iwasawa:
The memory problem seems to be fixed.
However, from this version the transparent region in the rendered image seems to be replaced by semi-transparent light gray color. Please see the attached image:
dwanko_ss_test

Please note that I added 20 pixels padding which are still transparent in this result.

turtletooth:
Thanks for catching that.  I'm sorry I didn't catch that in my fix.  I am testing a fix right now.

shun-iwasawa:
I confirmed that the problem was fixed. Thank you!
Jenkins

shun-iwasawa:
Now I agree with @RodneyBaker that this feature would be helpful not only for game creators, but for animation studios. Lining up the animation drawings is so beneficial to overview whole actions.
@turtletooth thank you for adding the nice feature. LGTM

PR #759, opened 2016-08-26
WIP - New Slate Theme

turtletooth:
There are still some items not available in stylesheets, but this takes advantage of what is available.  The main goals were neutral gray, flatten out the ui a bit, new icons for many items available to qss.

I think the level colors in the schematic viewer are hard coded- this prevents changing the level colors in the xsheet for consistency.

Less files compiled with EasyLess plugin in Visual Studio Code- works great!

slate2

Try it here:
[slate.zip](https://github.com/opentoonz/opentoonz/files/438500/slate.zip)


blurymind:
it looks very sleek! :)

adamearle:
Woahhhh that look so much better than the standard version.  What would you do if you could make it sexier.

adamearle:
Maybe change up the scrollbar so they are softer on the eyes. I think that would be pretty nice. To me the beveling on the scrollbars feels to strong.

This is so much better than the original though.

turtletooth:
@adamearle I agree about the scroll bars.  Unfortunately, they are not all reachable via qss yet.  I haven't yet learned how to bring things out to qss, but I plan on changing the scrollbars too.

turtletooth:
Changed to WIP - I need to look at what needs to happen for the mac side of things.

shun-iwasawa:

turtletooth:
Closing for now.

PR #769, opened 2016-08-28
Gif Trimming and Transparency

turtletooth:
This enables trimming of transparent areas on gif images and exporting with transparency.
The exporting with transparency uses ImageMagick which needs to be configured in preferences the same way ffmpeg is. 

preferencesgif

I also added the ability to add extra padding to trimmed images in case someone didn't want the exported image at the very edge of the screen.

gifsettings

For images that have a lot of semi-transparency, I added a image dithering option.  This is not desirable for most images, but makes a difference sometimes- see below.

Non-transparent gif:
trimmed-non-transparent

Transparent gif:
trimmed-transparent

Transparent gif with dithering:
trimmed-transparent-dithered

Note that the dithering was only needed because of the semi-transparent area.  I realize that the final image isn't that great, but gif doesn't support semi-transparency.  It is either on or off, so a dithering pattern is the best that ImageMagick seems to offer at the moment.


shun-iwasawa:

RodneyBaker:
I ran across this reference and it has me wondering if rgb24 (versus pal8) format might be related to the issue with gif transparency via ffmpeg...

https://trac.ffmpeg.org/ticket/715

The pertinent part:
> This works now for ffmpeg -i 11.png -f image2 out.gif
> There are two gif encoders, only one supports pal8 output, the other only supports **RGB24** which **by definition does not support transparency**.
> 

Perhaps this won't make any difference but including it here just in case.
Assuming the use of pal8 encoder gains transparency it might negate any benefits due to loss of RGB24. 

At any rate, it would appear that if ffmpeg uses the RGB24 encoder then transparency will not be an option but it might be if using the pal8 encoder.  The issue with using pal8 would likely be that a sorting of color palette might need to be executed first because pal8 only supports 256 colors and therefore some optimization would be required in order to gain the best possible image with the limited palette.

turtletooth:
I'm going to close this as I'm not sure the need for transparent gifs is high enough to make the users jump through the hoops needed to get this working.

RodneyBaker:
Interestingly, this feature appears to be implemented at least partially due to implementation of #2418.

spinner


I did get a report that the transparency was not appearing on mobile phones...

PR #770, opened 2016-08-28
Fix for gif reader not getting all frames

turtletooth:
Ffprobe was not returning the correct value for gif number of frames.  This counts the number of frames made from the conversion to png and sets that as the number of frames.

Found this error when I was playing with spritesheet generation.

### DIFF
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -399,6 +399,26 @@ void Ffmpeg::getFramesFromMovie(int frame) {
   }
 }
 
+int Ffmpeg::getGifFrameCount() {
+  int frame               = 1;
+  QString ffmpegCachePath = getFfmpegCache().getQString();
+  QString tempPath        = ffmpegCachePath + "//" +
+                     QString::fromStdString(m_path.getName()) +
+                     QString::fromStdString(m_path.getType());
+  std::string tmpPath = tempPath.toStdString();
+  QString tempName    = "In%04d." + m_intermediateFormat;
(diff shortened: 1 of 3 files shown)


shun-iwasawa:

shun-iwasawa:
I've tested your modification with the GIF animation of parrot which you posted in #752 .

It seems that the original GIF contains 60 different images and has 260 frame length.
And ffprobe returns the number of images (=60) and failed to load whole footage. Is it correct?
If so, is it possible somehow to load 60 images and put them in xsheet with appropriate steps so that it can reproduce original timing and frame length, rather than load it as 260 different images?

shun-iwasawa:
One more point related to frame length: It seems that the GIF images are converted to ”intermediate" PNG images once you browse them in File Browser. If so, is it possible to show frame length of gif and mp4 files in File Browser?
( I think I should post this as another feature request. )

turtletooth:
I have no idea how to get the frame timing out of ffmpeg.  It would definitely be more efficient, but with school starting again, I don't have the time to dig into ffmpeg again.

turtletooth:
@shun-iwasawa I did a quick search.  I'm not seeing much in the way of getting the delay information out of ffmpeg.  It's probably there, but not easily found.

shun-iwasawa:
@turtletooth thank you for the quick search! So let's leave it as feature issue and I'll merge this PR since this definitely fixes the bug.
BTW, I found some other issues regarding loading of GIF files. Please take a look; #812 , #813 

Jenkins

turtletooth:
@shun-iwasawa I saw the other issues and confirmed them.  Thanks for catching them.

shun-iwasawa:
LGTM. Thank you for fixing the bug!

PR #773, opened 2016-08-30
Allow to reset rooms without restarting

turtletooth:
This rebuild the default rooms without restarting when a user selects reset rooms in menu.

This enables a user to get their rooms back in order without restarting if they mangle their rooms.
Note: the browser and farm room are not rebuilt without restarting.  Something about rebuilding them in the middle of running crashes OT.


turtletooth:
This was built because my students were occasionally accidentally messing up their rooms and then needing to get things back to default.

turtletooth:
Closing for now- I think this can be done in a cleaner way.

Issue #774, opened 2016-08-30
Save All Issue


turtletooth:
Save all sets a new scene path to the cache folder if cancel is pressed in the save scene dialog.  The scene untitled is also set to false, so hitting save scene later will save the scene to the cache folder.

To duplicate:
1. New Scene
2. Draw something
3. Hit Save All and cancel the save scene dialog
4. Do Save Scene As
5. Notice the file path.

PR #775, opened 2016-08-30
Fix save all issue setting wrong path on cancel save scene

turtletooth:
Fix for #774


shun-iwasawa:
Jenkins

shun-iwasawa:

shun-iwasawa:
LGTM. Thank you for the fix!

PR #786, opened 2016-09-04
Audio Recording Popup

turtletooth:
This feature allows recording of audio through the system input devices.
- You can choose to have the animation play as you record and playback
- Audio meter to indicate input levels
- Add recorded audio into the scene

recorded
recording


Rayek:
Wow, that is actually quite a helpful addition indeed. I wonder why other animation application developers haven't thought of this before :-)

Will you be supporting export to WAV as well?

turtletooth:
> Will you be supporting export to WAV as well? edit Ah, I just noticed the audio export thread you worked on. Thanks so much!

Just to clarify- I saw that you saw the wav export PR too, but this also saves the sound to a wav file before importing it to a level.

2dvision:
Maybe need to add "pause" button too.

blurymind:
That would be super useful too! :)

blurymind:
Adobe flash lets you record.multiple takes in one sound clip and then
select the play range that the audio clip will use in order to pick the
right take

shun-iwasawa:

turtletooth:
Here is the latest:
newaudio

turtletooth:
Found a few things that need to be tweaked.  Changed name to WIP

turtletooth:
I think this is ready- removed the WIP.

shun-iwasawa:
It seems to work fine. Jenkins

shun-iwasawa:
@turtletooth I think the fix #928 would help this feature, which enables to display the Level Settings and the Info Viewer for sound levels.

turtletooth:
Thanks for #928

shun-iwasawa:
LGTM

Issue #787, opened 2016-09-04
Typo: lenght


mcccs:
Where? [Here, here, here...](https://github.com/opentoonz/opentoonz/search?q=lenght&type=Code)

So you want to change them with what? [These](https://github.com/opentoonz/opentoonz/search?q=length&type=Code)

If you confirm, I can open a Pull Request.

RodneyBaker:
A few more typos...

Looking at @mcccs search results I see another typo with 14 occurrences.
The typo'd word is 'simmetric' which should be 'symmetric'.
It is often preceded with with another word such as 'Not_' or appended with 'index'.

There are 174 instances of the correctly spelled 'symmetric'.

https://github.com/opentoonz/opentoonz/search?p=2&q=simmetric&type=Code&utf8=%E2%9C%93

PR #790, opened 2016-09-05
Typos

mcccs:
Resolves: #787 

I bulk-replaced all `simmetric`s and `lenght`

I'll fix the wrong ones and squash it later.

**UPDATE**: Done.

**UPDATE 2 & 2.5**: Attempted to fix compile error. Waiting for CI.

### DIFF
--- a/toonz/sources/common/tapptools/tcolorutils.cpp
+++ b/toonz/sources/common/tapptools/tcolorutils.cpp
@@ -183,7 +183,7 @@ void chooseLeafToClusterize(ClusterContainer::iterator &itRet,
     assert(!ISNAN(clusterCovariance[0]));
 
     // calcola gli autovalori della matrice di covarianza della statistica
-    // del cluster (siccome la matrice e' simmetrica gli autovalori
+    // del cluster (siccome la matrice e' symmetrica gli autovalori
     // sono tutti reali)
     KEYER_FLOAT eigenValues[3];
     tmpMulteplicity = calcCovarianceEigenValues(clusterCovariance, eigenValues);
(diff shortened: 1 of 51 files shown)


RodneyBaker:
I'm posting this here as it is a Typo topic that is still open.

I"m not sure if this one is a problem or not but I happened upon the word 'zerary' which is the toonz word for 'geometric' and my curiosity got the best of me so I dug deeper.  I saw that notes in the code state that word isn't used often because no one understands it's meaning.  But that word isn't the (potential) problem word I want to bring to attention.  Another word, 'zeray' (note the missing second 'r') got my attention mostly because it appears to be used in only one instance of code (although four times in that code).

It can be found here:
https://github.com/opentoonz/opentoonz/blob/1131f1c4e12011edf025cd82e3e1f2bb249e2f45/toonz/sources/toonzqt/fxdata.cpp

As it appears to be internally consistent it may not currently be producing a bug but if it is meant to be 'zerary' instead of 'zeray' then I assume it should be flagged as a typo.

I suspect this is a potential bug producer because unlike 'zerayfx' there are many instances of 'zeraryfx'.

Perhaps the word zeray is correct.
Perhaps not.

shun-iwasawa:
@RodneyBaker 
Thank you for pointing the other typo. The term "zerary fx" is a quite unique term in Toonz's source. I think it is used as one of "[arity](https://en.wikipedia.org/wiki/Arity)" and it means "fx with no input port" (i.e. zero + ary = zerary) , under the following rule:

- Fx with two input ports is called "binary fx".
- Fx with one input port is called "unary fx".
- And fx with no input port is called "**zerary fx**".
(Usually the term "**nullary**" seems to be employed for such case... But zerary is OK for me.)

So, IMO "zerary" is correct and "zeray" would be typo.

RodneyBaker:
Thanks @shun-iwasawa.  That's great information.
I had a feeling that term was specifying some deeper information.
I was equating the first part with zero but... couldn't grasp that latter part.
I'm learning a ton of information from you guys/gals!

My favorite word (as of this moment) is 'zerary'.  :)

PR #793, opened 2016-09-07
Startup popup when OT starts

turtletooth:
newstartup

This allows for getting to recent scenes faster and setting basic settings easier for new scenes.  The option to use pixels is also available since setting scenes using a default of inches may be confusing for some.  This would also allow autosave to be enabled by default since scenes would be created with a name and a path.  This would help avoid lost work - a common user complaint.

You can disable this on the startup page or in settings.  

Closing this window takes you to the normal untitled scene without anything changed.


Rayek:
I kinda like this, but how is this going to work when the user(s) work on multiple concurrent projects? How do we choose a particular project to work on, and create a new scene in that project? And what if the user wants to create BOTH a new project and new scene from this dialog?

I feel this dialog is perhaps oversimplifying things a bit too much.

blurymind:
The tick box to disable it is clearly seen on the screenshot too

Rayek:
Yes, I know - but still: I would like this dialog to include project-related start-up options as well. One of the basic causes for beginner mistakes is that they fail to create a project. This dialog seems to promote that behaviour.

RodneyBaker:
@Rayek 
I'm following you!  :)
I'm not sure we will see the disambiguation of what a Project file is/is not any time soon.
Perhaps this is an opportunity to sort that out?

It is my understanding that the Project is the directory the files reside in (providing another path is not specified in the New Project Dialog).  So in the case of the screen capture above the Project Name is 'sandbox'.  Or perhaps better put... that particular Project is currently sandboxed/untitled?

A question I might ask would be whether what is entered in that first field is the Project Name or Scene Name.  I suspect it is the Scene Name and the path in the next field is supplying the Project folder (again... the Project Name and the Folder the Scene files reside in are the same.  Unless changed... which is something Jeremy is working on... Scenes must be stored in the root of the Project directory. (REF:  related comments by @shun-iwasawa)   

So from one perspective... there is no such thing as a 'Project _**file**_'.  There is only a link to the directory where the scene files reside and an appropriation of that folders title.

Questions that remain unanswered until we can test this new dialog would be:

- By entering a name in the top field are we creating a Project folder or Scene File?
I suspect this is for the naming of a Scene file; the one created when the button labeled "Create New Scene" is pressed.
Either way it might be best to give that a full title 'Project Name' or 'Scene Name'.

 -On the right might it be more appropriate to have it titled "Open Recent Scene" as all of these are Scene files?

The 'Project Name' might be provided at the very top via an uneditable field that links the name from the current Project location.  If a new location/path is selected that field could then update with the new Project Name.  In that way eveyone could readily see what the current Project is.

I think some of my confusion would disappear if it could be confirmed that with OpenToonz there is no such thing as a single file that qualifies as a Project file.  I think this is where a lot of confusion resides.  If this is the case the terminology might be changed to emphasize the term "Project Folder" at every opportunity.

turtletooth:
How's this:
newstartup

I think I took care of most of the issues mentioned.

blurymind:
very nice !! Thank you @turtletooth :)

RodneyBaker:
Looking good.  :)

You could drop 'Name' in 'Scene Name' to add a little more marginal space on the left and/or allow shifting of everything to the left .  It would still maintain the current tapered look as the fields descend.  I suppose 'Scene Name' might help users understand that a scene doesn't just have to be a number.  So that works too.

As it is it looks quite useful to me.  
Nice work.  :)

turtletooth:
I'm not a huge user of non-pixel measurements, but for those who are, I added a resolution label to know how the current measurements compare with pixels.  The resolution label goes away when pixels only mode is selected but it is nice to have in other modes.
addedresolution

turtletooth:
Latest version:
* Changed don't show to "Show at startup"
* Moved show at startup to bottom
* Removed path from recent files for space saving

lateststartup

RodneyBaker:
Nice changes!

I have a few workflow questions but those should wait until we get to play with this in action.
The  primary question would be... Does the dialog automatically close upon selecting 'Create New Scene'? and (of course) provide instructions to the user on what to do if they haven't supplied the new Scene Name yet.

turtletooth:
You can try it here:
https://github.com/motionmonster/motionmonster/releases/tag/v1.1.1.2

2dvision:
I'm not UI designer, But i change a few things in OpenToonz Welcome Screen:

welcom-screen


- Change "Name" field position.
- Add "Camera Size" field.
- Add "DPI" field (I usually use 72 DPI and pixel unit, I never use inch because this is confusing for video resolution)
- Change "New Project" button position and rename to "Create New Project".
- Change "Load Other Scene" button  position.
- Separate "Open Recent Scene" with a gray boarder.
- Remove "Close" button (Because we have close button in top)

What do you think?

turtletooth:
@2dvision I agree with almost everything.

I don't think the name should be moved- it needs to have a prominent place and is hard to find in your mockup.

I also don't think the new project should be moved.  In your mockup new project and new scene both look like a finalizing click.  Creating a new project is just a part (sometimes) of creating a new scene.  

I do like shortening the fps field and the way you added dpi without making the list longer.  

Removing the close button makes sense

I was going to add a separator between the recent scenes and the rest- just haven't figured out how I want to do it yet.

The camera size field is a good idea.  I need to figure out how best to implement that.

RodneyBaker:
DPI (Dots Per Inch) is primarily a setting for printers and printing so while it's okay to have it here it doesn't serve much of a purpose (other than add to the confusion of DPI).  What you are likely looking for is PPI (Pixels Per Inch).  A lot of software gets this wrong or have just given in and use DPI to mean PPI and press on.

RodneyBaker:
> You can try it here:
> https://github.com/motionmonster/motionmonster/releases/tag/v1.1.1.2

I'm still running it through it's paces and trying every combination of options but so for so good!

There is one oddity that perhaps could be addressed in that the Load Other Scene button at lower right probably should start it's search for Scenes from the location of the Project file specified in the dropdown menu (upper left). 

 In addition to enhancing the usefulness of identifying a Project folder the user desires to work in, starting the search for other Scenes from that specified Project location would address another issue:  it is not immediately clear how to select a Scene from a Project selected in the Project dropdown menu that isn't listed over on the right (in the Recent Scenes listing).  

It seem that launching a search via Load Other Scene might help us locate it but it takes us to the Project root of all projects (which might actually be appropriate we just have to navigate to the current Project to get at that unlisted Scene).  

On the plus side, it is very clear how to create a new Scene in the specified Project location by entering a name for the Scene and hitting the Create New Scene button.   I like that and it works quite well!

RodneyBaker:
This splash screen is working quite well for me.
I like that it puts an emphasis on the Project we are currently working in as well as how it reminds us from the moment of launching the program that identifying where we are (our project location) is important.   This should also go a long way toward resolving issues related to misplaced files for tose of us that lack organizational skill, especially as the splash presents lots of options for where our files are most likely to be located.

An added plus (going on behind the scenes of this feature) is that of freeing up Scenes so they can reside outside of the primary Project root folder.  That's a big gain.

RodneyBaker:
I've run into something of a fringe problem...

Is there are way to get this dialogue back (without restarting) after it has been dispatched?
It's useful enough that I'd like to retrieve it because I know it contains information I want to readily access.
Perhaps it can be added to the Help menu but I will guess this gets into the realm of exposing it for inclusion in customization of menus/rooms.

turtletooth:
> It seem that launching a search via Load Other Scene might help us locate it but it takes us to the Project root of all projects (which might actually be appropriate we just have to navigate to the current Project to get at that unlisted Scene).

This is actually the load scene behavior.  Clicking Load Other Scene simply triggers load scene as you would from the menu.  I think the behavior was designed to make it easier to access scenes from other projects.

RodneyBaker:
> This is actually the load scene behavior. Clicking Load Other Scene simply triggers load scene as you would from the menu. I think the behavior was designed to make it easier to access scenes from other projects.

That makes sense and works for me.

turtletooth:
Here is the current version- I added back the DPI because one of the things that separates it from other animation software is how well it works for traditional workflows, but I am torn because it introduces the strange balance of managing pixel dimensions, other dimensions and dpi.  This can sometimes be a frustrating thing- for example should changing dpi change the inch width and height or change the resolution?

Allowing the dpi to be changed also makes it easier to create scenes that are incompatible with pixels only mode.  Pixels only mode locks the dpi in at 120, and scenes created with another dpi but loaded into OT with pixels only mode enabled give a warning.

Thoughts?

So here is the current version:
currentstartup

Note: I left the close button for the sake of being obvious.  Is a user unchecks "Show this as startup" it may not be obvious to them that they need to close using the X.

blurymind:
You could have it be like most other apps and select the metric system with a drop down menu, then depending on what the option is the input boxes are different in terms of values and what's shown
http://pe-images.s3.amazonaws.com/photo-effects/studio-background/new-photoshop-document.gif

:)

Rayek:
I like this version much better - bravo!

I do agree with Rodney that DPI is used for paper, and PPI ought to be used when referring to the relative screen resolution<->bitmap resolution. DPI is just plain wrong as it is used in OpenTOonz.

But alas, OT is not the only piece of software to get this wrong. When printing or scanning DPI is used. When referring to pixels and inches, PPI is used.

shun-iwasawa:

konero:
I would strongly recommended some accenting, grouping of the different options. This is done quite well if you look at the output or cleanup settings, different options are grouped with white frames. Here's a quick mock I did, sorry it doesn't look like the original UI, I don't have Photoshop on this PC, only Paint.Net so just bashed it together with the pencil tool.

welcome_ui
([view full-size](https://cloud.githubusercontent.com/assets/19820721/18395773/0d718f3a-76b8-11e6-8a44-acf45c694190.png))

I'd remove the close button, if the user can't find the OS close button then they won't get far using Toonz TBH, but your call.

I would hope there was some way to only display the camera size pixel dimensions.

Perhaps if you set Toonz to use The Association for Japanese Animations standard layout paper size which @shun-iwasawa mentioned awhile back, but I can't remember why it wasn't used, of 10in x 5.625in (254mm x 142.875mm) you could take the first three digits from the width box to set the DPI/PPI value behind the scenes automatically.

10in x 128dpi = 1280x720
10in x 192dpi = 1920x1080
10in x 384dpi = 3840x2160

However I don't understand the pixel unit feature and whether that would work.

turtletooth:
@konero- that looks great.  I'm not a UI wizard in Qt yet, but I will do what I can.  

> I would hope there was some way to only display the camera size pixel dimensions instead of the analog unit, rename "Preset" to Camera Size to make it clear what the drop box is for.

Currently the pixels checkbox sets the units to pixels and removes the dpi section, but based on @blurymind suggestion, I am going to do a units drop down - like in preferences.  This will remove the checkbox for pixels but still give the option to use various units (which will include pixels).  If pixels is selected, it will turn on pixels only mode as it also does in preferences.  

The way pixels only mode is built right now- it locks the dpi in at 120 and the pixels resolution changes the size of the scene.  I want to only change the dpi if a user specifically changes that box.  That way scenes are compatible with pixels only mode unless the user specifically changes the dpi box.  

If pixels only mode is selected and a scene with a different dpi is opened, a warning pops up with the option to turn off pixels only mode or potentially mess up your scene.

The easiest solution to this is to remove the dpi box altogether (which is fine with me - since I prefer pixels), but I think that traditional animators might prefer to have the option (I don't know though, since I have never done traditional paper based).

2dvision:
@turtletooth I install Motion Monster v1.1.1.2, Every time I minimized and then maximize it, Startup popup be displayed.

turtletooth:
@2dvision Good catch.  That will be fixed in the new version (tonight?)

turtletooth:
Here is the most recent version- probably not going to change much after this except play with the stylesheet.

Here it is with the default inches selected as the unit:
finalish

And here it is with pixels selected as the unit.  Note that some fields go away as they are not needed:
finalpixelsonly

konero:
From a UX perspective I'd still push for less ambiguity here with the use of **QGroupBox**.

**Read here:** http://doc.qt.io/qt-5/qtwidgets-index.html

parent-child-widgets

This is because the process of properly setting up Toonz is sequential with _Choose Project > Create New Scene_.

You can remove the period from "Show this at startup".

For buttons that lead to other windows with user options such as browsing for a file, you should include ellipses, which are the three dots at the end.

**Wrong:**

Open

**Correct:**

Open...

**Read here:** https://msdn.microsoft.com/en-us/library/dn742392.aspx#usingellipses

> **Using ellipses**
> While menu commands are used for immediate actions, more information might be needed to perform the action. Indicate a command that needs additional information (including a confirmation) by adding an ellipsis at the end of the label.

> Proper use of ellipses is important to indicate that users can make further choices before performing the action, or even cancel the action entirely. The visual cue offered by an ellipsis allows users to explore your software without fear.

turtletooth:
Thanks @konero - I didn't know about QGroupBox before.  That will work well.

2dvision:
@turtletooth that looks good. Are you can add centimeter (or millimeter) to units? in many countries people do not use inch.

turtletooth:
@2dvision they are already there.

turtletooth:
Thanks @konero for showing me QGroupBox.  Here is the final? version:
final

turtletooth:
@2dvision new version available now.

RodneyBaker:
Very nice.  This popup has to be the most practically useful I've seen in any program.  Interestingly, it's not intrusive like others (which I usually shut off if there is an option to do that).  My first experience with this latest version of the popup felt like the menu/popup has always been there and I find myself missing it if I open a release without it.  The styling, layout and seamless fit into the UI... the additions of QGroupBox... all of these really add a nice final polish so thanks @konero!

I have only one nitpick and that is how to retrieve the popup once it has closed without restarting the program.  I hope there is a way to add a link to open the popup via the Help menu.  This will be useful also if users opt to shut off the popup but then want it back.  They won't have to restart OpenToonz to get it back.

turtletooth:
> I have only one nitpick and that is how to retrieve the popup once it has closed without restarting the program. 

@RodneyBaker You can configure it to be a shortcut in the shortcuts popup.  Search for startup and you should see it.

RodneyBaker:
> @RodneyBaker You can configure it to be a shortcut in the shortcuts popup. Search for startup and you should see it.

Awesome.  Thanks much for that. I found it and have added it it to the Help menu.  I'll have to check to see if I should add it to specific rooms... I love that we can customize our menus in OpenToonz.

Aside:  Those listings (and some other menu listings... like plugins) should probably be alphabetized as it's hard to pick things out especially if we don't know the name/title we are looking for.  

I'm officially ran out of nitpicks.  :)

2dvision:
@turtletooth Very beautiful. 
If we have Startup Popup in Help menu by default, it will be useful. 
Scene name in "Open Scene" zone separate with a long space (It is not line after line).
startup popup
(motion monster v1.1.1.3)

turtletooth:
@2dvision Thanks- fixed:
fixedspacing

2dvision:
@turtletooth ﻿I find new problem, 
1. Create new scene by OpenToonz Startup 
2. save that
3. open OpenToonz Startup
New scene do not add to "Open Scene".

suggestion:
- Rename "Open Scene" to "Open Recent Scene"
- Rename "Open Another Scene..." to "Open Scene..."

turtletooth:
@2dvision This was built to only show once, which is why the list isn't updating.  But I do see the value in using it beyond the first time, so I will make it refresh the list when it shows again.

turtletooth:
@shun-iwasawa Thank you so much for merging PRs last week.  When you have the change would you mind looking at this one next?

toonz/sources/toonz/startuppopup.h
```diff
@@ -0,0 +1,108 @@
...
+  QList<TFilePath> m_projectPaths;
+  QCheckBox *m_showAtStartCB;
+  QComboBox *m_projectsCB;
+  QComboBox *m_unitsCB;
+  QPushButton *m_loadOtherSceneButton;
+  QPushButton *m_newProjectButton;
+  QComboBox *m_presetCombo;
+  QPushButton *m_addPresetBtn, *m_removePresetBtn;
+  CameraSettingsWidget *m_cameraSettingsWidget;
+  int m_dpi, m_xRes, m_yRes;
```
shun-iwasawa: m_dpi should be double as dpi value can be decimal number.

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -0,0 +1,889 @@
...
+  if (Preferences::instance()->getPixelsOnly()) {
+    if (m_dpiFld->getValue() != Stage::standardDpi) {
+      m_dpiFld->setValue(Stage::standardDpi);
+    }
+    m_widthFld->setValue(m_xRes / Stage::standardDpi);
+    m_heightFld->setValue(m_yRes / Stage::standardDpi);
+  } else {
+    m_widthFld->setValue(m_xRes / m_dpi);
+    m_heightFld->setValue(m_yRes / m_dpi);
+  }
```
shun-iwasawa: For now, inputting pixel size field causes unexpected change in width/height fields - it looses fractional portion.
Lines 865-869 should be ;
`    m_widthFld->setValue((double)m_xRes / Stage::standardDpi);`
`    m_heightFld->setValue((double)m_yRes / Stage::standardDpi);`
`  } else {`
`    m_widthFld->setValue((double)m_xRes / (double)m_dpi);`
`    m_heightFld->setValue((double)m_yRes / (double)m_dpi);`

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -0,0 +1,889 @@
...
+  m_fpsFld->setValue(fps);
+  m_unitsCB->setCurrentText(Preferences::instance()->getCameraUnits());
+  m_dpi  = cameraRes.lx / cameraSize.lx;
+  m_xRes = cameraRes.lx;
+  m_yRes = cameraRes.ly;
+  m_resXFld->setValue(m_xRes);
+  m_resYFld->setValue(m_yRes);
+  m_resXFld->setDecimals(0);
+  m_resYFld->setDecimals(0);
+
```
shun-iwasawa: Here, `m_dpiFld` should also be initialized with `m_dpi`.
Please insert `m_dpiFld->setValue(m_dpi);`

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -0,0 +1,889 @@
...
+  m_fpsLabel                = new QLabel(tr("Frame Rate:"), this);
+  m_fpsFld                  = new DoubleLineEdit(0, 24.0);
+  m_cameraSettingsWidget    = new CameraSettingsWidget(false);
+  m_presetCombo             = new QComboBox(this);
+  m_unitsCB                 = new QComboBox(this);
+  m_addPresetBtn            = new QPushButton(tr("Add"), this);
+  m_removePresetBtn         = new QPushButton(tr("Remove"), this);
+  m_showAtStartCB           = new QCheckBox(tr("Show this at startup"), this);
+  QPushButton *createButton = new QPushButton(tr("Create Scene"), this);
+  QPushButton *newProjectButton = new QPushButton(tr("New Project..."), this);
```
shun-iwasawa: It is not actually a problem, but IMO it looks better if you can modify them in a consistent way by;
- setting `this` instead of `0` for the parent widget
- removing `DVGui::` namespace qualifiers

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -0,0 +1,889 @@
...
+  m_resYFld->setRange(0.1, (std::numeric_limits<double>::max)());
+  m_showAtStartCB->setChecked(Preferences::instance()->isStartupPopupEnabled());
+  m_showAtStartCB->setStyleSheet("QCheckBox{ background-color: none; }");
+  m_addPresetBtn->setStyleSheet(
+      "QPushButton { padding-left: 4px; padding-right: 4px;}");
+  m_removePresetBtn->setStyleSheet(
+      "QPushButton { padding-left: 4px; padding-right: 4px;}");
+  QLabel *label = new QLabel();
+  label->setPixmap(QPixmap(":Resources/startup.png"));
+  label->setObjectName("StartupLogo");
```
shun-iwasawa: Is this `StartupLogo` object name used ?

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -0,0 +1,889 @@
...
+  m_projectBox->setObjectName("StartupProjectBox");
+  m_sceneBox->setObjectName("StartupNewSceneBox");
+  m_recentBox->setObjectName("StartupRecentSceneBox");
```
shun-iwasawa: If these three boxes can be the same design, I think it would be clear to give them the same object name.
Just sharing `SolidLineFrame` would be fine as it is already defined in the style sheet.

toonz/sources/toonz/startuppopup.cpp
turtletooth: agree, and fixed.

toonz/sources/toonz/startuppopup.cpp
turtletooth: It's not used right now.  I was going to try and make this available in the qss originally.  I'm removing the line.

toonz/sources/toonz/startuppopup.cpp
turtletooth: Great catch.  Thank you.

turtletooth:
@shun-iwasawa Thank you for the PR review.  You do a great job of catching little things that I miss in the 1000 lines of code.  I appreciate your thoroughness and learn from your attention to detail. 

> I think it is a good choice since the startup popup should be easy for beginners
I'm glad you agree.  I thought that non-square pixel dpi on the startup would be confusing.

I think I fixed the errors you noted and cleaned the qss since only SolidLineFrame is needed.

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -0,0 +1,889 @@
...
+  if (m_sceneNames.count() <= 0) {
+    m_recentSceneLay->addWidget(new QLabel(tr("No Recent Scenes"), this), 1,
+                                Qt::AlignTop);
+  } else {
+    int i = 0;
+    for (QString name : m_sceneNames) {
+      if (i > 9) break;  // box can hold 10 scenes
+      QString justName = QString::fromStdString(TFilePath(name).getName());
+      m_recentNamesLabels[i] = new StartupLabel(justName, this, i);
+      m_recentSceneLay->addWidget(m_recentNamesLabels[i], i, Qt::AlignTop);
```
shun-iwasawa: How about to changing background color of `StartupLabel` on mouse hover, in order to make it clear for users that these "Open Recent Scene" items are clickable ?



(As it is not a bug, I can post this mock-up as a separate PR after this PR is merged.)

turtletooth: That sounds great.

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -0,0 +1,889 @@
...
+
+  TProjectManager::instance()->setCurrentProjectPath(projectFp);
+
+  IoCmd::newScene();
+  m_pathFld->setPath(TApp::instance()
+                         ->getCurrentScene()
+                         ->getScene()
+                         ->getProject()
+                         ->getScenesPath()
+                         .getQString());
```
turtletooth: I think I fixed that.  I also added enter in the name field animating the click for create scene

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. @turtletooth thank you for adding this feature, it'll be helpful especially for beginners!

Issue #690, opened 2016-07-29
InknPaint workflow compatible with Japanese animation industry


shun-iwasawa:
In order to make OpenToonz to be introduced into Japanese Animation industry, some new features are necessary to be implemented. 

As a first step, I will focus on InknPaint work. In Japan, almost all animation studios are using Retas PaintMan for their InknPaint work and staffs have been familiar with it over the years. So it is very important to modify OpenToonz and gain consistency with PaintMan in usability.
With the cooperation of some Japanese animation studio, we made an InknPaint workflow plan using OpenToonz as follows:

<img src="https://cloud.githubusercontent.com/assets/17974955/17247814/d08f81ba-55d0-11e6-8958-350f41eaa8c5.png" width=400>
[InknPaint workflow using OpenToonz.pdf](https://github.com/opentoonz/opentoonz/files/390476/InknPaint.workflow.using.OpenToonz.pdf)

Please note that we made the input and output data to be the same as the current workflow done with PaintMan so that they can transfer their workflow comfortably. 
To realize the above, some features are still missing as follows (marked with red boxes in the diagram):

- [ ] When converting images to TLV, make “Autopaint for Lines” option to be activated by default for the initial color styles for self-color lines.

- [ ] Add "Organize palette" option to the style picker tool. With this option, the picked style is automatically moved to the first page of the palette.

- [ ] Make it enable to switch style using numpad keys. Hit Tab key to shift a scope of 10 styles to be selected with numpad keys.

- [ ] Highlight a part of the color model image temporarily when the style switched in order to show which part of the character should be filled with the current style. 

- [ ] Open the Locator sub-window from the Viewer for showing close-up image at mouse position.

- [ ] Add "autoclose" option to the Fill tool.  With this option, line gaps will be automatically closed on filling somehow. (I have no idea yet how to realize this.)

- [ ] Memorize picked position in the style in order to maintain link between styles and the raster color model. 

- [ ] Add  “Update colors by using picked position” command. Colors are picked from the color model by using the picked position memorized in each style.

blurymind:
The bucket tool autoclose would be a godsent!! :)
I really hope that gets implemented for both vector and bitmap levels

blurymind:
The best system I have seen at the gap autoclosing is lazybrush.
Fundamentally it is very different from the way the current fill works - but TVpaint uses it and is excellent. Currently it has an open source implementation that the krita team are working to integrate as a layer type. 

I filed more info here : #420 

[...]

PR #804, opened 2016-09-12
Enable switching the current style using numpad keys

shun-iwasawa:
This PR implements one of the items listed in the issue #690.

> Make it enable to switch style using numpad keys. Hit Tab key to shift a scope of 10 styles to be selected with numpad keys.
- The shortcut is available in the Viewer and ComboViewer.
- This feature becomes available when the Preferences option "Use Numpad and Tab keys for Switching Styles" (in Drawing category) is set to ON. Default value is OFF.
- If any numpad keys are assigned to some commands' shortcuts, activating this option will release these shortcuts. A confirmation dialog will open in advance.
- This feature is available only in the first page of the palette. Hitting Tab key when the scope is at the bottom of the first page will make the scope move back to the top of the first page.

**UPDATE:** Enabled Shift+Tab for moving the scope backward.
**UPDATE:** Disabled numpad keys to be assigned as shortcut of another command when this feature is activated.


shun-iwasawa:
Jenkins

2dvision:
Dear @shun-iwasawa , Is it possible use Shift+Tab for take backward?

shun-iwasawa:
@2dvision My original intention of this PR is to gain consistency with RETAS PaintMan in usability. I noticed that the scope move backward by Shift+Tab in PaintMan as you exactly pointed.
Thanks! I'll include it right away.

shun-iwasawa:
Done. Jenkins again.

shun-iwasawa:
Fixed conflicts. Jenkins

turtletooth:
I really like this.  Tested and it works.  Here are some thoughts on it.  Is it possible to use 0 as the first style instead of 1?  By default style numbers begin with 0, so there is a confusion about clicking 1 to select style 0.  

This is useful enough that I think it could be on by default.  The downside is that a few shortcuts are lost with this mode.  I use 9 and 0 a lot for changing the view.  Do you have any ideas for other shortcut keys that could replace them?

Thanks @shun-iwasawa so much for this feature.  I really like it.

turtletooth:
One more thought- is it possible to add the shortcut functionality when the mouse is over the level palette as well?  Sometimes I move my mouse to the palette as I am thinking about color and try to tab to a different color set.  It's a little thing, so it's not crucial.  But it might help people from thinking the feature only works sometimes - not realizing that the mouse needs to be over the viewer to use it.

turtletooth:
>  Is it possible to use 0 as the first style instead of 1?

Another idea might be to overlay the shortcut number over the palette chip somewhere.  Even with my earlier suggestion- if you have numbers 0-23 and delete number 6, the the numbering gets thrown off anyway.  And starting with 0 doesn't really help.  It would be great to somehow have a visual cue as to what keyboard shortcut will trigger what color.

shun-iwasawa:
@turtletooth Thank you very much for the feedback!
 
I set the shortcuts starting from 1 for two reasons: First, numerical keys (not numpad) are aligned starting from 1 in keyboard so IMO it would be more intuitive to set the shortcut in the same order. Second, I think it would better (at least for Japanese users) to keep consistency between Retas PaintMan in usability. 

> One more thought- is it possible to add the shortcut functionality when the mouse is over the level palette as well? 

That's make sense. I'll try to do so.

> Another idea might be to overlay the shortcut number over the palette chip somewhere.

Nice idea. I'll think how to show them on the chips.

turtletooth:
> I set the shortcuts starting from 1 for two reasons: First, numerical keys (not numpad) are aligned starting from 1 in keyboard so IMO it would be more intuitive to set the shortcut in the same order. 

I thought about the number key order after I posted my comment and agree that the current implementation makes sense.

turtletooth:
@shun-iwasawa What do you think about making this the default and using Alt+Number to replace the shortcuts that already use numbers?  Since the shortcuts have already been changed with my shortcut overhaul, it would make sense to change them here so they don't get changed after the next release - and confuse people who are getting used to the new shortcuts.

shun-iwasawa:
@turtletooth 
I briefly checked the number key usage in the current shortcut presets.

- **defopentoonz** uses 1,9 and 0 for OpacityCheck, ZoomFit and ZoomReset.

- **otharmony** uses 1,2 and 0 for ZoomOut, ZoomIn and Collapse.

I totally agree that changing default shortcuts of these commands to `Alt`+_Number_ or Non-number keys would be helpful for users since almost all of these commands (except "Collapse", I think) will be frequently used in Ink&Paint work as well.  

On the other hand, if you meant in the last comment that you suggested to register new commands, say, "Switch the current style to N-th in the scope" commands to `CommandManager` and assign 0-9 number keys for the default shortcuts, I'm quite negative to do so.

turtletooth:
> if you meant in the last comment that you suggested to register new commands, say, "Switch the current style to the N-th in the scope" commands to CommandManager and assign 0-9 number keys for the default shortcuts

I totally agree with you.  I don't think it should be moved to the command manager at all.  I was simply saying that I think the feature could be enabled by default in preferences.  It is very useful and makes a painting workflow so much faster.  I wasn't suggesting to allow the individual positions to have their own custom keys.  I think 1-0 works perfect.  

To summarize, my suggestions are
* Enable this by default in the preferences
* Change the shortcuts that are already mapped to numbers to Alt+Number
* Enable the switching with the mouse in the Level Palette
* Show a visual indicator in the chip to help users know what number to press

I love this feature and think it will speed up the workflow for many people.

shun-iwasawa:
@turtletooth Modifications done.
I displayed shortcut indicator on top of each chip.
Note that the indicators are only shown with large and medium chip view.


Also shortcut is now available when mouse is on the palette,  enabled this feature by default in the preferences and changed the shortcut presets from numbers to Alt+Number.

shun-iwasawa:
Jenkins

turtletooth:
I love this new feature.  Thanks for doing this.

I don't know what's up with Travis.  

Also on mainwindow.cpp lines 2050, 2051

```
createViewerAction(V_ZoomReset, tr("Reset View"), "0");
createViewerAction(V_ZoomFit, tr("Fit to Window"), "9");
```

should be 

```
createViewerAction(V_ZoomReset, tr("Reset View"), "Alt+0");
createViewerAction(V_ZoomFit, tr("Fit to Window"), "Alt+9");
```

shun-iwasawa:
Modified. @turtletooth thank you for the good catch!

Jenkins

turtletooth:
@shun-iwasawa Any idea what is going on with Travis?

shun-iwasawa:
Fixed conflicts. Jenkins

turtletooth:
@shun-iwasawa Tested again, and now I can't get the numpad keys to change styles.  The number row keys work, but when I try numpad keys it doesn't change.  Am I doing something wrong.

shun-iwasawa:
@turtletooth  Oh, now I noticed that `Qt::KeypadModifier` is set for the numpad keys.
Thank you for pointing this! I'll fix it right away.

shun-iwasawa:
Jenkins again

turtletooth:
LGTM.  Thanks @shun-iwasawa

gab3d:
@shun-iwasawa 
I've just discovered this feature, and it is indeed wonderful!
Thanks for implementing it! (In fact I've read the whole proposal for Ink and Paint and it looks really great!) :)

What motivated me to write is the issue of the unmatching numbers for the keys vs the colors selected.
I've already read the justification for that choice (the main keyboard numbers starting at 1, and compatibility with PaintMan), but honestly it seems still unnecessarily unfriendly to me...
<starts Teacher thinking here> ;)
...because there's no doubt, the fact that there are certain numbers (the keys) not matching with other suspiciously similar numbers (the color indexes), quickly raises valid questions from any new user: 
- Why 1 key takes me to color 0? 
- So, is there another 'hidden index' for colors in which the first color is really 1 and not 0 as displayed?

...and suddenly the program starts looking as if it had a level of depth and complexity in this area that it really doesn't (because that's just something that's there for compatibility sake with another software).
<Teacher mode off>

I know taking care for certain animation houses is a good thing for OT, but perhaps there could be a way to make the program look as easy as it could also to new users. (Perhaps via a Preference option?)

Well, just to add some more thinking to that matter.
And thanks again for making OT progress as it is!

PR #805, opened 2016-09-13
Update version numbers for tcleanupper, tcomposer and tconverter

shun-iwasawa:
For now the batch render or batch cleanup are unavailable, with the following error popup appears:

This PR fixes such problem by updating version numbers for each batch executable - tcleanupper, tcomposer and tconverter.

This also fixes another bug as follows:

> When executing tcleanupper with .cln file as input argument, if some project folder alias (such as "+drawings") is used in "Save in" path, output file is saved in wrong place.


shun-iwasawa:
Jenkins

turtletooth:
LGTM

PR #814, opened 2016-09-14
Fix for FFmpeg renders wrong range if previous exported range was bigger 

turtletooth:
To duplicate:
1. Create a scene with 10 frames and render it using Ffmpeg
2. Render the scene again only choosing frames 1-5
3. The export uses the previous export to get its image info from and renders the previous range.


ideasman42:
Shouldn't this be moved into a method?
*(easy to miss some extension if this is copy-pasted all over)*.

`bool usesFfmpeg = type == "gif" || type == "mp4" || type == "web"`

Also, please squish corrective commits.

turtletooth:
> Shouldn't this be moved into a method?
Probably.  Please don't merge this until #826 is merged and I will use the method I created there for this.

turtletooth:
Moved the change into a function.

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed the fix. @turtletooth thank you!
LGTM

Issue #520, opened 2016-06-23
bug regression (harlequin)- toggle link to studio palette is missing, there is no way to link a style to the studio palette


blurymind:
It looks like every level in opentoonz has its own palette. It cant be instanced.

The only way I found to share a color palette between levels is to use the studio palette window to set my colors, then load my palette manually to every level that I want to use it.

Edit: I seem to have found a regression bug again.

Toonz harlequin allows you to link a style to the studio palette. This command has been removed in opentoonz.

blurymind:
@shun-iwasawa  "Do you want to update all linked styles in current scene"
Does not seem to do anything in today's nightly build.

I am assuming that this is the feature that i am looking for

blurymind:
Ok I actually found that there is a regression from toonz harlequin!
In harlequin you can right click on a style on the level palette and "toggle link to studio palette"

It gets a tiny link indicator in the white box like this:
![regressionpalette](https://cloud.githubusercontent.com/assets/6495061/16323884/604f7b8a-39a6-11e6-9dfe-6aebff4f1879.PNG)

When link is enabled every time you change the studio palette color and save the studio palette, opentoonz updates it on every level where the style is linked!

This feature seems to have been removed from opentoonz. In the right click menu there is only an option to "remove links" which actually is the equivalent of "remove refference to studio palette" command in toonz harlequin.

When you save the studio palette, it asks you if you want to update it on the levels, but saying yes to that does absolutely nothing as none of them are actually linked.

@shun-iwasawa this is on the latest nightly build

blurymind:
"Get color from studio palette" also does absolutely nothing.
@shun-iwasawa 
The functionality to link/update a level palette style to a studio palette style is completely broken/missing.

In toonz harlequin all you need to share a style between levels is have them all linked to studio palette. Then in order to update a color on two levels, all you need is to change it in the studio palette and save the palette. In a single click it updates everywhere.
OT can not seem to be able to do that.

turtletooth:
> "Get color from studio palette" also does absolutely nothing.

I was just playing with it.  It works if the studio palette has been saved after updating.  It is still a bunch of steps, but it does work.

blurymind:
[...]

PR #817, opened 2016-09-15
Modified: Linked styles enabled and style right click menu organized by turtletooth

shun-iwasawa:
This PR is for the issue #520 and based on the PR #628 by @turtletooth 

@turtletooth did almost everything so I just modified minor parts as follows:

- Use TStyleSelection::removeLink() instead of TStyleSelection::eraseToggleLink(). The former function takes care of removal of "original name" which is newly added from Ghibli Edition.
- Show these commands without right-clicking the linked style. Which makes easier to apply these commands to all styles in the palette.
- Removed default shortcut for "Toggle Link to Studio Palette" as it can be added by users afterwards.
- Removed confirmation dialog on save studio palette, asking whether to update linked styles.

### DIFF
--- a/toonz/sources/include/toonzqt/paletteviewergui.h
+++ b/toonz/sources/include/toonzqt/paletteviewergui.h
@@ -199,9 +199,6 @@ protected:
   void zoomInChip();
   void zoomOutChip();
 
-protected slots:
-  void removeLink();
-
 private:
   DVGui::LineEdit *m_renameTextField;
   QPoint m_dragStartPosition;
--- a/toonz/sources/include/toonzqt/styleselection.h
+++ b/toonz/sources/include/toonzqt/styleselection.h
@@ -98,6 +98,8 @@ public:
   void removeLink();
   // get back the style from the studio palette (if linked)
   void getBackOriginalStyle();
+  // return true if there is at least one linked style in the selection
+  bool hasLinkedStyle();
 };
 
 #endif  // STYLESELECTION_INCLUDED
--- a/toonz/sources/include/toonzqt/styleselection.h
+++ b/toonz/sources/include/toonzqt/styleselection.h
@@ -95,7 +95,7 @@ public:
   void toggleKeyframe(int frame);
 
   // remove link from the studio palette (if linked)
-  bool removeLink();
+  void removeLink();
   // get back the style from the studio palette (if linked)
   void getBackOriginalStyle();
 };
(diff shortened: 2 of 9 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
@blurymind 
Regarding this problem you reported in #628 :


Sorry for my late response, but IMO this was not a bug. Please check whether the styles "test" and "color_2" in the studio palette were created with "New Style" command or copy&pasted from another studio palette.
If the styles are created with "New Style" command, it becomes a "parent" of the link. If the color is copy&pasted from another studio palette, it becomes a "child" of the copied one even if they are pasted in the studio palette. And the update of linked color is only available from "parent" to "child".
To identify "parent" or "child" of the link, change the palette display mode to "Large Thumbnails View" and "Both Names". The style name of the parent (="original name") will appear on the top-left corner only in the "child" styles.
A diagonal line on the style chip means that it is the style with reference to studio palette, but its color is changed from the "parent" style in studio palette.

turtletooth:
Thanks for looking at this @shun-iwasawa I didn't quite understand all that was going with the links and got stuck.  Thanks for taking care of the rest.

turtletooth:
@shun-iwasawa When I right click a project palette and select "Load into Current Palette" the new colors in the level palette don't update when I choose "Toggle Link to Studio Palette" or "Get Color from Studio Palette" - I think this is the issue that @blurymind was talking about above.  My styles are original to the project palette (not copied from elsewhere). 
Also, what do you think about hiding the option to update the link, get color, and remove reference for styles that are not linked to a color?

turtletooth:
Ok, I can make it easy to duplicate:
1. Have a Project Palette with some colors
2. Create a new level
3. Load Project Palette into Studio Palette
At this point- toggling the link works and getting the color works.
4.  Make a change to the project palette (change a color to something else)
5.  Save the project palette
6.  Linking and gettting color don't work anymore.

shun-iwasawa:
@turtletooth Thank you for the reviews, I'll fix it right away.

shun-iwasawa:
- Hide commands from context menu if there is no linked style in the selected styles.
- Fixed problem that the style link becomes unavailable after using "Load into Current Palette"

shun-iwasawa:
Jenkins

blurymind:
@shun-iwasawa  this is big to me. Thank you for the fixes!
Will test it asap 
:+1: 
@turtletooth  thank you for explaining the problems with linking

turtletooth:
That seems to have fixed it.  Thank you so much @shun-iwasawa - It is very useful.
Looks like the mac build on Travis stalled.  It is ready to merge though.

turtletooth:
LGTM.  Merged.  Thanks so much for fixing this @shun-iwasawa

blurymind:
AWESOME! Thank you guys :)

Issue #813, opened 2016-09-14
Fails to load more than one GIF files with the same file name


shun-iwasawa:
When loading GIF image, the "intermediate-cache" images are saved in the same folder ($TOONZCACHEROOT/ffmpeg) and are  identified by file names. 
After you load some GIF image (e.g. +extras/folder1/A.gif), you cannot load another GIF with the same name (e.g. +extras/folder2/A.gif) properly since the cache of the first one is used.

PR #827, opened 2016-09-19
Fix for Multiple FFmpeg files with same name #813

turtletooth:
Allows importing multiple ffmpeg type files with the same name but different locations.
#813


shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

Issue #779, opened 2016-08-31
ffmpeg export to mp4 issue


tjoen:
I noticed that if I render to mp4 and I DON'T start the render at keyframe 1 to 4, the finalizing progressbar does NOT appear, and the file just gets deleted, It does make the tempfile and the png's while rendering, but at the end, the png's and the mp4 get deleted,

It seems to happens with several ffmpeg builds I tried.
Rendering the whole movie from one of the first four keyframes does work.

I'm on windows 10 with OT 1.1.1

turtletooth:
I'll look into it.  Thanks for pointing that out.

turtletooth:
Fixed in #828.  Thanks @tjoen for pointing it out.

PR #828, opened 2016-09-19
Fix for ffmpeg export range error

turtletooth:
#779 
This allows the export range to be specified and correctly exported in ffmpeg formats


Rayek:
Thank you!

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed that the bug is fixed. LGTM

PR #844, opened 2016-09-28
Fix/Workaround bug #591

### LINKED ISSUE #591: Crash adding a toonz raster level. Linux
Opentoonz Crash just adding a new toonz raster level: 

https://dl.dropboxusercontent.com/u/69554201/OpentoonzLinuxRasterCrashBug.mp4

ideasman42:
We could check for the extensions but these OpenGL options are never used,
so no need to disable them.

Resolves #591


shun-iwasawa:
I confirmed that your modification works with no problem in my Windows environment.
Sorry to trouble you, but can you please apply clang-format? Thanks!

ideasman42:
Updated patch, ran clang-format.

Issue #843, opened 2016-09-27
Clone Levels/drawings not saving/working correctly


RodneyBaker:
I'm getting some strange results with cloned columns.

Working in latest nightly release.  Steps to Reproduce:
Create New Project
Draw something
Select cell in xsheet that has drawing
Clone that cell/column via Menu>Cells>Clone
Save the Scene/Save All/Save Level... (I'm still exploring differences in results here)*
Create New Scene or Project (to flush old project)
Open the Old Project and note the cloned drawing is missing.

*In the case of saving a level OT will notify me that a previous level already exists (if I've previously attempted to specifically save a cloned level) although I don't see the file anywhere on the harddrive.

shun-iwasawa:
Confirmed in my Windows7 environment and fixed. Thank you for report the bug!

RodneyBaker:
I confirm clone levels are now being saved correctly.
Nice sleuthing and thanks for the quick turn around!  :)

This issue can be closed.

PR #847, opened 2016-09-29
Fix clone levels not saving correctly

shun-iwasawa:
This PR is for the issue #843.

Save Level command will work on only levels with dirty flags being ON.
In this PR I modified the clone level command so that the cloned levels are set to "dirty".

Please note that with this modification will only turn on the dirty flag. You still need to Save Level / Save All in order to save them into files. (This behavior is same as Toonz Harlequin)


shun-iwasawa:
Jenkins

turtletooth:
LGTM

Issue #852, opened 2016-10-05
Improve convert command for color adjustment work in Japanese animation production


shun-iwasawa:
This feature is requested from some Japanese animation studio.

They are making an animated feature and would like to improve quality in character colors as well.
Usually (in TV series), characters are painted with specific color charts and are identical through sequence. But they decided to add extra work; they would like to adjust character colors **for each scene** in order to fit the background image.

Actually it is the same approach as Studio Ghibli. Ghibli also adjusted character colors for each scene.
Adjusting colors can be easily done in Toonz since Toonz can handle drawing data and palette data separately. However, they differ from Ghibli in that their ink&paint work was done with PaintMan and composition work will be done with AfterEffects.

Based on the situation above, we made a new workflow plan as follows:
[Color adjustment workflow using OpenToonz.pdf](https://github.com/opentoonz/opentoonz/files/510305/Color.adjustment.workflow.using.OpenToonz.pdf)
<img src="https://cloud.githubusercontent.com/assets/17974955/19104973/969134ca-8b1a-11e6-8d2f-afcd34efb1a6.png" width=300>

In order to realize this workflow, some features are needed to implement:
- Enable to specify input palette when convert raster images to TLV with "Painted tlv from non AA source" mode.
- Add option to remove unused styles from input palette.

turtletooth:
I think this is complete.

PR #853, opened 2016-10-05
Improved convert command from non AA raster images to painted tlv

shun-iwasawa:
This PR is for the issue #852 and is improving "Convert" command which can be triggered from the right-click menu of File Browser.

- Enabled to input palette when "Painted tlv from non AA source" option.
- Added an option to remove unused styles from input palette.

### DIFF
--- a/toonz/sources/include/toonz/Naa2TlvConverter.h
+++ b/toonz/sources/include/toonz/Naa2TlvConverter.h
@@ -146,7 +146,8 @@ public:
       return -1;
   }
 
-  TToonzImageP makeTlv(bool transparentSyntheticInks);
+  TToonzImageP makeTlv(bool transparentSyntheticInks,
+                       bool removeUnusedStyles = false);
 
   TVectorImageP vectorize(const TToonzImageP &ti);
   TVectorImageP vectorize(const TRaster32P &ras);
--- a/toonz/sources/include/toonzqt/filefield.h
+++ b/toonz/sources/include/toonzqt/filefield.h
@@ -53,6 +53,8 @@ class DVAPI FileField : public QWidget {
   QStringList m_filters;
   QFileDialog::FileMode m_fileMode;
   QString m_windowTitle;
+  QString m_descriptionText;  // if the initial text is not path, set the string
+                              // here and prevent browsing
 
 protected:  // used in the child class for CleanupSettings
   QPushButton *m_fileBrowseButton;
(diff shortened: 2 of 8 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
@turtletooth 
Could you please review this?
Since this is very specific, here I uploading the sample files in order to try this feature:
[convert_sample.zip](https://github.com/opentoonz/opentoonz/files/550044/convert_sample.zip)

Please do the following:

1. There are three files in the package "convert_sample.zip":
     - colormodel_sample.tlv
     - colormodel_sample.tpl
     - level_sample.tga

1. The level "colormodel_sample" was converted from PNG to TLV and then organized palette. "level_sample.tga" is a sample of drawing data painted with PaintMan (Non-AA, TGA format). 

1.  In File Browser, right click "level_sample.tga" > click "Convert..." command.

1. In the Convert popup, do as follows;

     - Select "File Format" to "tlv"
     - Set "Mode" to "Painted tlv from non AA source"
     - Click "..." button next to the "Palette" field and specify the path to "colormodel_sample.tpl"
     - Ensure that "Remove Unused Styles from Input Palette" option is ON
     - Click "Convert" button.

1. After a while, "level_sample.tlv" will be generated.

1. Load "level_sample.tlv" and see the contents of the level palette.

Please note that the level palette are already organized, named according to the input palette.
Also please note that the unused styles in the level (in this example, green colors for clothes) are removed and a color found only in the level is added to the first page of the palette, as an "additional color".

convert

turtletooth:
@shun-iwasawa I'm sorry I haven't looked at this yet.  I will try and get to this tonight or tomorrow.

turtletooth:
@shun-iwasawa It looks like it works as expected.  I am totally unfamiliar with the workflow, so thank you for the detailed instructions on verifying it working.  

One think I noticed- when I clicked on the ... for the palette location, it stalled and finally popped up with the path set to my project/Create new palette.  It looks like a bug similar to the load level bug.  I think it can be a different PR, but it does seem to be a bug.

Do you want me to wait on merging this until the osx build is passing?

shun-iwasawa:
@turtletooth Thank you for checking the feature.
I modified `FileField` not to browse the initial path on first launch of the popup if the description text (=non file path, like "Create new palette") is input.
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa

PR #862, opened 2016-10-13
Locator sub-window for showing close-up image at mouse position

shun-iwasawa:
This PR is implementing the locator sub-window, which is requested in the issue #690 .

In Japanese animation industry, almost all companies are using Retas PaintMan for Ink&Paint work.
Since they handle enormous amount of drawings every day, they need to raise the efficiency as much as they can.
Under those circumstances, a color locator sub window in PaintMan is exploited. It shows close-up image at mouse position so that they don't need to take time to zoom up the main viewer even when painting small areas.

The locator sub-window in OpenToonz can be launched from ”magnifying glass” button in the viewer's console, next to the histogram button.

locator_window


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

turtletooth:
This looks awesome!  I'll try to look at this tonight.

turtletooth:
The code looks good as far as I can tell.  I also played with the feature.  For complex frames, it did seem to slow the performance down, but I think that is expected.  I also once had it follow a section of the canvas where my mouse wasn't- as if the coordinates it was following were a few hundred pixels up and to the left of my mouse.  I wasn't able to duplicate it and it only happened once in my attempts.  I think it can be merged because it works well and I think the one glitch I found is a very rare behavior.  

What are your thoughts?

shun-iwasawa:
@turtletooth Thank you for checking the feature.
I noticed that the displayed position in the locator becomes not correct if you are selecting scene frame and the current column has NS/EW translation. I'll try to solve this problem.

turtletooth:
> I noticed that the displayed position in the locator becomes not correct if you are selecting scene frame and the current column has NS/EW translation. 

That makes sense.  I think this is what happened when I found the glitch.  Thanks for noticing this.

shun-iwasawa:
Jenkins again

shun-iwasawa:
@turtletooth I think I could fix the problem. Could you please check again? Thanks!

shun-iwasawa:
Jenkins

turtletooth:
LGTM thanks @shun-iwasawa  for the useful feature.

RodneyBaker:
I'm enjoying this feature.  Thanks!

One thing I discovered is that if I'm drawing on the canvas there seems to be considerable lag and feels a bit detached BUT... and this is an important 'but'... if I'm drawing in the sub-window the drawing is smooth and feels great.

I was going to ask a few question but in the process of typing I discovered the answer!

I discovered that drawing on the canvas was a little choppy with a little lag BUT if I am drawing in the sub-window it works great.  The question I was going to ask related to how I might change the zoom of the sub-window or move it to another location on the canvas and I found the shortcut keys work great.  For instance, using Space allows us to move around the canvas while in the sub-window.  Shift Space allows for zooming in and out.  A quick hit of the B shortcut key has me back drawing again.

This is refining my workflow a lot with regard to closing gaps in lines.

I guess this is a long way of saying that this feature may have been implemented primarily for other reasons but it's a great addition for drawing/sketching digitally too.

Thanks!

Issue #867, opened 2016-10-18
Load scene popup fails to initialize if the "+scenes" project folder path is relative 


shun-iwasawa:
When opening the Load Scene popup for the first time, the initial folder position should be set to "+scenes" project folder*.  But it fails to locate a correct position if relative path (such as "scenes") is set to the project folder path. 

*For some specific rooms in "StudioGhibli" room set, the initial position changes in order to separate working data for each workflow.

turtletooth:
merged

PR #868, opened 2016-10-18
Fix Load Scene popup to initialize correctly

shun-iwasawa:
This PR is for the issue #867 
Now the Load Scene popup can locate correct initial position even if the "+scenes" project folder path is relative.

### DIFF
--- a/toonz/sources/include/toonz/tproject.h
+++ b/toonz/sources/include/toonz/tproject.h
@@ -57,7 +57,7 @@ public:
   bool isConstantFolder(int index) const;
 
   TFilePath getFolder(int index) const;
-  TFilePath getFolder(std::string name) const;
+  TFilePath getFolder(std::string name, bool absolute = false) const;
 
   TFilePath getScenesPath() const;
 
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
Jenkins

turtletooth:
I tried testing this, but am unsure what steps to follow to duplicate the bug.  I set the scenes path to the project path but didn't see any difference in the load scene popup behavior.  I'm sure I'm doing it wrong.

shun-iwasawa:
@turtletooth Thank you for testing this.
This problem only appears on the first launch of the Load Scene popup. Please do the following in order to reproduce the problem:

1. Make sure that the "+scenes" project folder of the current project is set to relative path. (such as "+scenes=scenes")
1. Trigger "Load Scene" command.

In my test, OpenToonz fails to set the initial folder position since it tries to find the relative path ("scenes") from the file system WITHOUT converting the path to absolute. Furthermore, it takes extra time for searching the path before opening the popup.

shun-iwasawa:
Jenkins again. (We are now transferring Jenkins environment for OSX to new machine)

turtletooth:
LGTM.  Thanks @shun-iwasawa

PR #870, opened 2016-10-18
Startup Popup: Highlight "Open Recent Scene" items on mouse hover

shun-iwasawa:
This PR modifies Startup Popup, changing background color of "Open Recent Scene" items on mouse hover in order to make it clear for users that they are clickable. Also added a tool tip showing the full scene path.


turtletooth:
Tested and works well.  All this needs is a Jenkins.  Thanks @shun-iwasawa

shun-iwasawa:
Jenkins

turtletooth:
LGTM- thanks @shun-iwasawa

artisteacher:
In Mac OS X El Capitan, the tooltip works but the changing background color on hover is not working.

Issue #766, opened 2016-08-26
New Default: Rounded Ends for Vector


turtletooth:
The flat ends for vectors are often quite harsh and I find I rarely use them.   I would prefer if the rounded vector ends were the default. If I am alone here, I don't mind closing this, but I find the rounded ends to be much easier to work with.

PR #872, opened 2016-10-20
Change brush default to rounded

turtletooth:
The default flat end brush cap is sometimes harsh.  I think using rounded as the default would be nicer and easier to work with.  

#766


shun-iwasawa:
LGTM. 
I always use the round cap too. It make sense since the raster brush is always looks like "rounded-capped", so changing the default value of vector brush to rounded cap is good to enhance consistency between vector and raster brushes.

PR #873, opened 2016-10-20
Make view and overlay buttons on viewer and combo viewer the same

turtletooth:
The viewer currently does not have the buttons for grid overlay and safe area overlay.  It also has the freeze button in a different place than the combo viewer.  This change takes the layout of those buttons in the combo viewer and makes it the same in the viewer.  

This should help the UX so users don't get confused about why buttons are sometimes there in the viewer and sometimes not.  

It will also help with theme programming since the code is now identical between combo and viewer.

### DIFF
--- a/toonz/sources/toonz/viewerpane.cpp
+++ b/toonz/sources/toonz/viewerpane.cpp
@@ -326,25 +326,49 @@ void SceneViewerPanel::initializeTitleBar(TPanelTitleBar *titleBar) {
 
   TPanelTitleBarButtonSet *viewModeButtonSet;
   m_referenceModeBs = viewModeButtonSet = new TPanelTitleBarButtonSet();
-  int x                                 = -188;
+  int x                                 = -232;
   int iconWidth                         = 17;
   TPanelTitleBarButton *button;
-  button = new TPanelTitleBarButton(titleBar, ":Resources/freeze.png",
-                                    ":Resources/freeze_over.png",
-                                    ":Resources/freeze_on.png");
-  button->setToolTip(tr("Freeze"));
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins.
@turtletooth Thank you for this modification. Actually I stopped updating the Viewer after we started using the ComboViewer in Studio Ghibli.

turtletooth:
Thanks @shun-iwasawa - Besides the toolbar and tool options bar, are there are any other modifications you made to the combo viewer that might be useful in the viewer?  I only prefer the viewer because I like a vertical toolbar.

shun-iwasawa:
Well, I can't recall properly what I did only in the ComboViewer... I briefly took a look and found two differences between the Viewer and the ComboViewer as follows:

1. With no current column selected, I removed zoom ratio information from the ComboViewer's title bar. Because there is no reference dpi to be used for computing the zoom ratio.
(...Oh, now I notice that the zoom ratio is needed to be displayed when the pixel unit is used!)

1. Flip Console height of the ComboViewer is smaller than the Viewer.

shun-iwasawa:
Anyway, LGTM for this PR. Thanks @turtletooth !

shun-iwasawa:
@turtletooth Sorry for my late notice but I remember one more thing regarding the difference between the Viewer and the ComboViewer:
For the ComboViewer I reimplemeted `TPanel::widgetFocusOnEnter()`, `TPanel::widgetClearFocusOnLeave()` and `TPanel::widgetInThisPanelIsFocused()` in order to move focus properly to `SceneViewer`, switch the current flip console and highlight the title bar on mouse enter.

Issue #754, opened 2016-08-24
[Qt] Fix Scrollbar BG, TabPage Underline, Palette Border (List Mode) and Move TPanel CloseButton to Stylesheet


konero:
_@shun-iwasawa | In reference to #696 and #719._

We've been looking at reducing 'glare' on bright objects that take the focus away from the content. Since we use professionally calibrated displays users cannot alter the color for their preference.

The palette window contains all the issues, so we'll use this to example. Here is how it currently looks...

![elem](https://cloud.githubusercontent.com/assets/19820721/17928046/ccfa3a20-69f0-11e6-8d38-dc2caa806d29.png)

And this is what we're trying to do...

![elem_f](https://cloud.githubusercontent.com/assets/19820721/17928069/e38d160e-69f0-11e6-9c5d-67965c2d833b.png)

For our case, we're simply trying to push the UI to the back and make it as invisible as possible.

### What Are The Issues?

- [x]  **Scrollbar BG**
When there is no content to scroll or when there is it's only displayed on half the bar, the scrollbar displays a check board pattern, on our monitors which display a very sharp image, its a little awkward to look at and when you move your eyes across the screen it generates a '[moire](https://en.wikipedia.org/wiki/Moir%C3%A9_pattern)' effect in vision.

![2](https://cloud.githubusercontent.com/assets/19820721/17928648/9e4f12d8-69f3-11e6-80a3-89ac11218439.PNG)

- [ ] **TPanel CloseButton**
It seems everyone is using floating windows a lot of the time, so there are bright blue close buttons all over the screen, is it a remnant of the old Harlequin titlebar?

![1](https://cloud.githubusercontent.com/assets/19820721/17928622/7ee4048a-69f3-11e6-97ef-4401b5887fdc.PNG)

- [ ] **TabPage Underline**
This is a strange one. We're able to style right/bottom/left sides of this element but not the top, could it be an alignment issue with another element? It's the 'underline' you see under tabs. We don't know whether it's to do with the body element under it (the color swatches) or the TabPage container.

![3](https://cloud.githubusercontent.com/assets/19820721/17928752/2ec14926-69f4-11e6-9083-87378a4ce7c2.PNG)

- [ ] **Palette (List View) Selected Swatch Border**
It's more obvious what's selected when the entire row BG is highlighted. Currently there is a strange outline, look at it close up...

[...]

PR #874, opened 2016-10-20
Unify all scrollbars to same style and fix background

turtletooth:
The scrollbar in the xsheet and the rest of the UI were not consistent.  This unifies the scrollbars and takes care of the checkered background on the scrollbars mentioned in #754

scrollbars

Note: There is still discussion to be had regarding theme colors, flattening, etc, but I think that should be its own PR/issue.

### DIFF
--- a/stuff/config/qss/gray_048/gray_048.less
+++ b/stuff/config/qss/gray_048/gray_048.less
@@ -1,4 +1,4 @@
-
+// out: gray_048.qss
 /* LESS Definitions */
 
 /*Image URL*/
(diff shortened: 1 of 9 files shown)


blurymind:
can we make the scrollbars thinner? They are wide - it will give us some
extra space for the canvas :)

shun-iwasawa:
@turtletooth Thank you for this modification.
Actually, in Toonz Ghibli Edition I set `QScrollBar`'s width to 16 pixels and set 2pixel-larger size (18pixels) for `XsheetScrollBar` in order to keep UI consistency with the older version.
In consideration of @blurymind 's suggestion, I think it would be fine to unify them to 16 pixels for now.
Please let me modify width of scrollbar based on your PR. Thanks!

turtletooth:
Closed for @shun-iwasawa improvement.  Thanks.

Issue #795, opened 2016-09-08
UI Enhancement:  Add Render Button to Output Settings Dialog


RodneyBaker:
This enhancement would save a few mouse clicks or keystrokes so it's more of a workflow enhancement than anything else.  It would be a redundant means of rendering but one that streamlines workflow and (especially for new users) provides self evident instructions all via one dialog box, namely, 'Fill in the desired output settings and then hit the render button'.  As it stands now users must remember to first open Output Settings panel via File Menu and then go back to almost the same place again and choose another option (Render) to launch the render.  For some users it may not be self evident that they need to open the Output Settings first before launching a render and in the process of learning they may render over the top of other files.

Currently there are two primary Render options (if we don't count the Render Farm aspect of OpenToonz.  The general method of rendering an image or sequence is as follows:
1) Open the Output Settings dialog and enter appropriate options (filename, file type, resolution, fps, etc.) These settings make this panel a critical stop if for no other reason than to check to make sure all settings are as desired.
2) Select Render from the File Menu (or alternatively use shortcut keys (if assigned*) such as "R")
Aside:  There is an implied hazard in rendering in OpenToonz that suggests rendering is going to immediately execute and overwrite any files that match what is specified in the output settings.  This is by design and as it should be because users may often want to update rendered imagery without resorting to hitting 'OK' buttons to confirm a render or overwrite a file.  Adding a button to the Output Settings dialog would strengthen the idea that the files designated in that dialog box will be overwritten (if present) or otherwise created for the first time.  

Adding a 'Render' button to the Output Settings panel would save a trip to the File Menu to launch a rendering or a trip to the keyboard (when already using the mouse).

I note that when expecting to render more than one outputted file I often leave the Output Settings panel open, change whatever is required and then render again with the new settings.  Adding a Render button would not interrupt this workflow but would save repetitious movement back and forth to the File menu.

Ideally, the Render button would execute on click or when the Enter key is pressed.  This would enhance workflow for those filling in data via keyboard or adjusting settings via mouse.

The ideal location for the button would likely be at bottom center of the dialog box but before the area of 'Other Settings'. This to ensure it remains available for render launch with or without the Other Settings being presented in the dialog.  Having it at this positions would also ensure that it doesn't get initial focus which could lead to inadvertently launch of a render before the user has had a chance to confirm desired output.

[...]

PR #875, opened 2016-10-20
Render Button added to Output Settings

turtletooth:
#795

Added a render button to the bottom of output settings.  This makes it easier to set the settings and render from the same place.

renderinoutputsettings


shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

Issue #857, opened 2016-10-09
[UX] Missing 'VerticalLineColor' QPROPERTY on XSheet Spreadsheet


konero:
Currently unable to style the vertical row line in the XSheet Spreadsheet viewer. The white is too strong IMO on the dark gray themes, it steals focus.

![xs1](https://cloud.githubusercontent.com/assets/19820721/19220579/667a1958-8e28-11e6-96d6-7b52215a2aca.png)

The Function Editor Spreadsheet viewer on the other hand has a QPROPERTY for it.

`qproperty-VerticalLineColor: rgb(120,120,120);`

![func](https://cloud.githubusercontent.com/assets/19820721/19220598/e5eb751a-8e28-11e6-81fd-b8aa5036332a.PNG)

PR #876, opened 2016-10-20
Fix white xsheet vertical line

turtletooth:
This is a fix for #857 

The vertical line in the xsheet was actually two lines- a black and a white line.  The white line was an accent while the black line actually was in the correct place for a cell divider.  This removes the white line and moves the black line to qss for styling.  I set the value of the black line to the color of the vertical lines in the function viewer for each theme.

Two new properties are available in the qss (example from gray_048):
qproperty-VerticalLineColor: rgb(120,120,120);
qproperty-VerticalLineHeadColor: rgb(160,160,160);

The first one is the color of the vertical line in the xsheet itself.  The second one is the color of the vertical line in the column header.  Since column headers are not the same color as the main xsheet, I found that some adjustment was needed to maintain contrast.

xsheetvertical

### DIFF
--- a/stuff/config/qss/gray_048/gray_048.less
+++ b/stuff/config/qss/gray_048/gray_048.less
@@ -1,4 +1,4 @@
-
+// out: gray_048.qss
 /* LESS Definitions */
 
 /*Image URL*/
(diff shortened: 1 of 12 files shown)


konero:
Fantastic job!

Rayek:
Nice. Looks tons better - much easier on the eyes. Thanks!

shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa Is this able to be merged? Thanks

shun-iwasawa:
Sorry to keep you waiting. It looks good. LGTM

Issue #881, opened 2016-10-21
UI Bugs


konero:
- [x] **Palette Window White Accent**

Non-stylable, a big annoyance.

This is also present in the **Style Editor Window**.

![fix16](https://cloud.githubusercontent.com/assets/19820721/19606660/c5c6315c-97bb-11e6-94a1-dd3056f35de2.png)

---
- [x] **Autocenter Alignment**
1. GroupBox doesn't line up with the other boxes.
2. CheckBox isn't being styled.

![fix1](https://cloud.githubusercontent.com/assets/19820721/19606300/41cc2962-97b9-11e6-8d38-160e054f08ff.png)

---
- [x] **Autosave Checkbox**

Missing style.

![fix2](https://cloud.githubusercontent.com/assets/19820721/19606379/b8e33cac-97b9-11e6-9544-2f47b5176d71.png)

---
- [ ] **Startup Window Check Label**

Changing the style or removing it shows the label is too narrow.

![fix4](https://cloud.githubusercontent.com/assets/19820721/19606425/193c0a34-97ba-11e6-874c-3b3d23ceb49b.png)

---
- [ ] **Cleanup Settings Window: Despeckling QLineEdit Gets Cropped**

![fix5](https://cloud.githubusercontent.com/assets/19820721/19606442/3bc1ac62-97ba-11e6-9ce5-d32a1508acb7.png)

---
- [ ] **Miter QLineEdit Gets Cropped**

Found under...
1. Brush Options
2. Selection Tool Options
3. Geometric Tool Options

![fix6](https://cloud.githubusercontent.com/assets/19820721/19606471/7606097c-97ba-11e6-828c-57fd8f0bbc98.png)

---
- [ ] **Geometric Tool Polygon Sides QLineEdit Gets Cropped**

![fix8](https://cloud.githubusercontent.com/assets/19820721/19606529/d21e6aa6-97ba-11e6-9804-5d016cd4efe0.png)

---
- [ ] **Pump Tool Accuracy QLineEdit Gets Cropped**

![fix10](https://cloud.githubusercontent.com/assets/19820721/19606564/1cb9e202-97bb-11e6-8c45-0619be98d2a0.png)

---
- [x] **Double QSlider Has Hard-Coded Styling**

AKA: #IntPairField / #DoublePairField.

![fix11](https://cloud.githubusercontent.com/assets/19820721/19606583/37f593a4-97bb-11e6-8b4d-3e43df6f2055.png)

---
[...]

PR #883, opened 2016-10-22
Tidy GroupBox Styles

konero:
#881 Adjusts alignment and adds styling to the indicator.

qgroupbox_fix

I don't know why left and right margin was added to the box, let me know if there was a good reason for that.

Why is a source map added to the themes? On my end that line breaks some styles, removing it reveals or unblocks them. Since I used WinLess it removed that line.

### DIFF
--- a/stuff/config/qss/gray_048/gray_048.less
+++ b/stuff/config/qss/gray_048/gray_048.less
@@ -347,7 +347,7 @@ QGroupBox {
 		subcontrol-origin: margin;
 		left: 15px;
 		margin-top: -2px;
-		.set_padding(5px, 0px);
+		.set_padding(3px, 0px);
 	}
 	&::indicator {
 		&:extend(QCheckBox::indicator all);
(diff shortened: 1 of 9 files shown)


turtletooth:
> Why is a source map added to the themes? On my end that line breaks some styles, removing it reveals or unblocks them. Since I used WinLess it removed that line.

I use Visual Studio Code with the Easy LESS extension for compiling.  It removes the last line on compile.  @shun-iwasawa Is that line needed?  Also, what is the difference between the normal qss and the mac qss?

shun-iwasawa:
@konero Fixed conflicts and posted. I'm terribly sorry to keep you waiting.

@turtletooth I was using Crunch for LESS editor-compiler. From this I started to use Visual Studio Code with the Easy LESS based on your comment. It is so good!

> Also, what is the difference between the normal qss and the mac qss?

For now there is no difference between the normal qss and the mac qss. I agree that it can be unified as you implied.

turtletooth:
Glad you like VS Code for this. I think it is awesome!

turtletooth:
Closing - conflict resolving PR (#1003) merged

Issue #887, opened 2016-10-26
Dragging a Studio Palette's tree item may delete palette


shun-iwasawa:
**PLEASE NOTE** : the following operation will remove palette file and THERE IS NO WAY TO RECOVER. Please make sure to keep backup before you try to reproduce this problem.
 
How to reproduce:

1. On Studio Palette, left-click any item in the tree.

1. Hold mouse button and drag a few pixels. Release the button as soon as the mouse cursor changes to <img src="https://cloud.githubusercontent.com/assets/17974955/19717505/e3764564-9b9b-11e6-975e-96a85dcf8777.png">(cursor with small arrow).

With this operation, the selected item unexpectedly disappears. A correspondent palette file is deleted and cannot be restore.

turtletooth:
Confirmed.  Will check the fix tonight or tomorrow.

turtletooth:
merged

PR #888, opened 2016-10-26
Fix Studio Palette not to delete palette file unexpectedly

shun-iwasawa:
This PR is for the issue #887

 This problem occurs when the `m_dropItem` is NULL because the mouse released without certain length of mouse move and `dragMoveEvent()` was not executed before `dropEvent()`.
In such case, `newPath` (= drop destination) becomes empty which causes unexpected deletion of the selected palette item.


shun-iwasawa:
Jenkins

turtletooth:
LGTM @shun-iwasawa Thanks for the fix.

Issue #866, opened 2016-10-17
[Ubuntu 16.10 amd64] Build error liblzo2, (workaround available, waiting on review)


ckorn:
The build fails on Ubuntu 16.10 amd64.

```
[ 74%] Building C object lzodriver/CMakeFiles/lzocompress.dir/lzocompress.c.o
/Â«PKGBUILDDIRÂ»/thirdparty/lzo/driver/lzocompress.c: In function 'main':
/Â«PKGBUILDDIRÂ»/thirdparty/lzo/driver/lzocompress.c:51:3: warning: ignoring return value of 'fread', declared with attribute warn_unused_result [-Wunused-result]
   fread(src, 1, src_len, stdin);
   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[ 74%] Linking C executable ../bin/lzocompress
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/6/../../../x86_64-linux-gnu/liblzo2.a(lzo1x_1.o): relocation R_X86_64_PC32 against symbol `memset@@GLIBC_2.2.5' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Bad value
```

This the entire build log:
http://zulu309.server4you.de/build/logs/opentoonz_1.1.1-1~getdeb2~yakkety-amd64-20161017-0303

//Just for my reference: https://paste.ubuntuusers.de/423188/

ideasman42:
Could you enable `-fPIC` as suggested?

ideasman42:
No reply in over a week. closing.

caryoscelus:
I have a similar issue on debian sid.

>Could you enable -fPIC as suggested?

Enable where? Error line suggests that fPIC should be enabled on system library if i read that correctly. It doesn't make sense to rebuild system lib. Compile guide mentions only tiff building. Do we actually need build all third-party packages?

```
/usr/bin/ld: /usr/lib/gcc/x86_64-linux-gnu/6/../../../x86_64-linux-gnu/liblzo2.a(lzo_init.o): relocation R_X86_64_PC32 against symbol `__stack_chk_fail@@GLIBC_2.4' can not be used when making a shared object; recompile with -fPIC
```

ideasman42:
Try add `-fPIC` to `CMAKE_C_FLAGS` and `CMAKE_CXX_FLAGS` in `CMakeCache.txt`.

caryoscelus:
[...]

PR #899, opened 2016-11-02
Fix/Workaround for #866

ideasman42:
Resolves linking issue on Linux, see: #866 
Note, we could re-order so it only impacts Linux, though I doubt it causes problems on OSX.

### DIFF
--- a/toonz/cmake/FindLZO.cmake
+++ b/toonz/cmake/FindLZO.cmake
@@ -9,12 +9,14 @@ find_path(
         lzo/2.03/include/lzo
 )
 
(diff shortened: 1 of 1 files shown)


ideasman42:
Note, we could re-order so it only impacts Linux, though I doubt it causes problems on OSX.

lordadamson:
this is really good and makes my life easier. I hope it gets merged soon.

shun-iwasawa:
Confirmed that this PR has no problem with building on both Win & OSX. Thank you @ideasman42 for the fix. LGTM

PR #908, opened 2016-11-09
Fix style removal problem on Convert to Painted TLV from non AA source

shun-iwasawa:
This PR fixes the problem on Covert to Painted TLV from raster images.

For now, the "Remove Unused Styles from Input Palette" option has a problem - the style removal operation executed for each frame. As a result, only styles used in every frames are left, but styles appears in the part of frames will be removed.

I modified this problem by changing the removal operation to be done once in the end of the conversion of whole frames.

Also, I changed the conversion rule, to make the closed regions (= regions not touching image border) can be background if they are painted in white.

### DIFF
--- a/toonz/sources/include/toonz/Naa2TlvConverter.h
+++ b/toonz/sources/include/toonz/Naa2TlvConverter.h
@@ -146,12 +146,13 @@ public:
       return -1;
   }
 
-  TToonzImageP makeTlv(bool transparentSyntheticInks,
-                       bool removeUnusedStyles = false);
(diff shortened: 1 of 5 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
@masafumi-inoue , could you please check this PR, since this bug-fix is quite specific to Japanese animation workflow?

masafumi-inoue:
LGTM @shun-iwasawa

Issue #910, opened 2016-11-10
Easy input feature for the Style Name Editor


shun-iwasawa:
In animation production, it is important to set appropriate name to styles in order to manage color design, but inputting style name is one of the burdensome tasks.
At least in Studio Ghibli's production, there were frequently-used keywords for style name; every human characters have colors for "eye", "skin", "hair" and "mouth" etc. In addition, every styles for shade parts have "_K" suffix. ("K" stands for "KAGE(影)", which means "shadow" or "shade" in Japanese)
In consideration of the above, I think it will improve efficiency if there is a way to input frequently-used keywords by just clicking the buttons.

pculli:
Very cool!

turtletooth:
Merged. Closing

PR #911, opened 2016-11-10
Introducing easy input feature for the Style Name Editor

shun-iwasawa:
This PR is for the issue #910 
The Name Editor (which opens via right-click menu of the style) is now improved as follows:

- Users can input frequently-used keyword to the style name by just clicking buttons under the input box.

- A new keyword can be added by users by clicking a plus ("+") button.
 
- Keyword button can be removed from the right-click menu of each button.  

- Keyword are arranged in three panels, which are originally intended to be used for "character name", "parts name" and "suffix" respectively - but it is left to users.

- "Apply" button is now changed to "Apply and Next" button, which will move the current style to the next one.


shun-iwasawa:
Jenkins

konero:
This is good!

I noticed in your image you have **medium sized** color chips but I can't seem to find a menu item to switch to that view. Am I just not seeing it? I only have **Large/Small/List**.

pal_l

shun-iwasawa:
@konero Missing of the "Medium Thumbnails View" option is a bug. Thank you for reporting the problem. I'll fix it and post another PR right away.
For now, you can Zoom In/Out shortcut in the Palette panel to get medium-sized chips.

shun-iwasawa:
Jenkins

turtletooth:
I love this idea.

The saving doesn't seem to be working for me.  What is the intended behavior?  Also there are some workflow things to possibly tweak regarding focus of the name box.  I'll post a more detailed review later today.

turtletooth:
@shun-iwasawa On a slightly related note- what do you think of making medium the default palette size.  This way the shortcut to the numpad change you created could be seen by default.

turtletooth:
First of all - this is a great feature.  There are a few tweaks that would make the workflow easier. 
When the name editor is opened it would be nice if the existing name was selected and the name lineedit had focus consistently.  

I have also found that when I click on a "easy input" name to insert it, it only replaces selected text if it was selected by hitting "apply and next".  Text that I selected and attempt to replace by inserting a word isn't removed, but the inserted word is simply placed wherever the mouse cursor is.  

I think the workflow should replace any selected text,

I also found that when hitting "apply and next" the focus isn't set to the name window.  When I click "apply and next" and try and type, I trigger keyboard shortcuts unless I click in the name window.  I think that both when opening the name editor and clicking apply and next, all the current text should be selected and the focus in the window.

The names also don't appear to be saved.  When I create easy input names and then save and reopen OT, the names are not there.  Is there a workflow for saving the names that I am missing.

I also think it might be handy in the name editor box to give a name to each of the 3 boxes to give a hint as to what they might be used for.  For example the first box might be labeled object or character, the second could be part, and the last could be suffix.  Even if a user chooses to use a different scheme, this would help new users understand what the boxes could be used for.  

Just a few thoughts.  I'm excited to use this feature.  Thanks for creating it @shun-iwasawa

shun-iwasawa:
@turtletooth Thank you very much for the detailed feedback! I'll look into them.

shun-iwasawa:
@turtletooth  Fixed as follows.

- Ensured the name editor box to be focused & selected-all when opening the popup and when pressing "Apply and Next" button

- Clarified the behavior when the word button is clicked as follows: 

 1. If there is a selected text in the name editor box, it will be replaced by the clicked word. 
 1. If the name editor box has no selected text but is focused, the clicked word will be inserted into the current cursor position.
 1.  OR, the clicked word will be added at the end of the existing text.

- Fixed in order to save the word list properly (on adding the new word).

- Added category name ("Name", "Part" and "Suffix") to the "+" button in each panel

- Inherits `DVGui::Dialog` in order to reproduce the window geometry on launch

- Changed the timing of construction of the `StyleNameEditor` - from ctor of the `PageViewer` to the moment of the first click of "Name Editor" menu. 

Jenkins

shun-iwasawa:
> @shun-iwasawa On a slightly related note- what do you think of making medium the default palette size? This way the shortcut to the numpad change you created could be seen by default.

@turtletooth Thank you for the suggestion.
Hmm, IMO the large chip would be fine as the default value (for beginners), since it is easy to recognize the information of each style. Once you set the medium chip, the chip size will be saved in `mysettings.ini` and reproduced on the next launch. Is it not enough?

turtletooth:
> Once you set the medium chip, the chip size will be saved in mysettings.ini and reproduced on the next launch. Is it not enough?

This is my mistake- I have been using the small chip for so long that I thought it was the default.  I think the large chip is fine as a default.  I think I personally prefer the medium as a nice balance of size and information, but I think the large is fine.

turtletooth:
@shun-iwasawa The focus and saving issues are resolved.  Thank you for adding the text to the buttons- I think it will help new users understand the purpose of the columns.  However, the text is cut off and resizing doesn't resize the columns to fix it.  One option would be to move the text to the top of the column as a column header either where Easy Inputs is now or under Easy Inputs.  Then each button could simply say new.

On a side note- what do you think of adding a little more default height to the window.  Right now adding 3 items to the suffix column brings up scrollbars.

Thanks again.

shun-iwasawa:
@turtletooth Thank you for your feedback again!

>  However, the text is cut off and resizing doesn't resize the columns to fix it.

Please update the style sheet. I made a new style for word buttons of which the labels can be displayed with no cut-off.
Plus, I modified the layout so that the buttons expands according to the width of popup. Now we can get wider button, so I changed the first panel name from "Name" to "Character", which is longer but appropriate.

> On a side note- what do you think of adding a little more default height to the window. Right now adding 3 items to the suffix column brings up scrollbars.

The Name Editor popup is now inheriting `DVGui::Dialog` which can reproduce its geometry on the next launch. As a side effect, for now there is no way to set initial size of the popup. I think it can be done by modifying the `dvdialog.cpp`, but IMO it is OK with the current situation, considering that some users may not need this feature and just would like to type style name.

shun-iwasawa:
Jenkins

turtletooth:
> Please update the style sheet

That makes sense.  Thanks.

turtletooth:
It looks like the r of character is cut off.
cut_off

shun-iwasawa:
Yes, I noticed it once and changed the label to shorter one ("Name").
For now, you can stretch the size of buttons according to the popup. So I decided to bring back "Character", which is cut-off if you shrink the popup to the minimum.
Do you have any good idea about this?

shun-iwasawa:
I'll try to move the panel names on top as you suggested in the previous comment.

turtletooth:
> I'll try to move the panel names on top as you suggested in the previous comment.

Either that or a font size of 10 in the stylesheet.  Whatever you think looks better/cleaner

shun-iwasawa:
How about this?



Jenkins again

turtletooth:
LGTM  thanks @shun-iwasawa

RodneyBaker:
I'm missing something here (with latest nightly build).
I am unable to get this new menu.
I still get the old menu with "Apply".
I'll double check to see if something persisted from the last install.


Looking forward to using it.  :)

shun-iwasawa:
@RodneyBaker 
Please check the build date of the software. It seems that this feature is included after Nov. 17 built.
Since Jenkins needs some time to create the package (from around 20 minutes to more than 10 hours - depends on network condition), there may be a case that the nightly build built from old head is released with a tag of the latest one. In such case, the features written in the description of download page will not be actually included in the software.

RodneyBaker:
@shun-iwasawa 
Thanks!  That was indeed the case and I am checking out the feature now.
In the future I will keep that delay with Jenkins in mind.

A very useful feature.

PR #912, opened 2016-11-11
Display "Medium Thumbnails View" option in the Palette's tool bar

shun-iwasawa:
This PR fixes the problem notified by @konero in the comment for #911 .


shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa  I would actually be in favor of making medium the default size since it can show the shortcut for fast changing via numpad.

Issue #915, opened 2016-11-14
Camera Capture popup doesn't open


shun-iwasawa:
This issue was reported from some Japanese animation studio.

This problem occurs if the popup position (position of the top-left corner) was near the bottom edge of the screen on the previous launch of the software. 
All the dialogs (inheriting `DVGui::Dialog`) remember its geometry and reproduce it on the next launch. But the popup with `Qt::WindowMaximizeButtonHint` such as Camera Capture popup seems to offset to bottom-right direction on being shown (for reasons unknown to me). 
Therefore, the popup get out of the screen and seems not to open.

**EDITED**: They also demanded the Camera Capture popup to be enable to put behind the main window.
Since this popup is space-consuming, they would like to switch back and forth between the popup and the main window by `Alt+Tab` (switching the active window shortcut).

turtletooth:
Merged

PR #916, opened 2016-11-14
Prevent the Camera Capture popup from being at outside of the screen

shun-iwasawa:
This PR is for the issue #915 
Although it is not clear why the flag `Qt::WindowMaximizeButtonHint` makes the popup to move to right-bottom direction, I added a "safeguard" in order to prevent the popup not to go out of the screen.
This fix will also take care when you are temporary using screen with low resolution (such as XGA projector), you will not lose the popup outside of the screen.

Also I changed the `parent` of `PencilTestPopup` from `TApp::instance()->getMainWindow()` to `0`, in order to enable the popup to be behind the main window.


shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa

Issue #918, opened 2016-11-15
Move ffmpeg temp export files to the ffmpeg temp directory


turtletooth:
Right now when rendering using ffmpeg, temp files are written to the destination directory.  OpenToonz has occasionally crashed during export, so the temp files don't get deleted leaving a clutter of files in that folder.  If the temp files were in the ffmpeg temp folder in cache, this wouldn't be a problem since that folder is deleted both upon opening and closing of OT.

turtletooth:
@shun-iwasawa If I can add this tomorrow- do you think this could make it into the next release?  OT crashes during export sometimes and my students end up with messy output folders.

shun-iwasawa:
@turtletooth OK. Also please let me know if you have any other specific PR you would like to be reviewed & merged before the next release. I'll try to do so.

turtletooth:
@shun-iwasawa Thanks.  There are three other PRs that I use regularly that I would love to have looked at- load image in camera capture #751, spritesheet export #752, and audio recording #786.  I realize that three PRs may be too many for you to look at since they are somewhat longer PRs.  Also the load image one isn't the solution that you suggested that I change it to, but I don't know how to implement your suggestion on that one.

What are your thoughts?

PR #921, opened 2016-11-16
Moved ffmpeg temp files to cache folder

turtletooth:
#918 
This saves the temp export files to the cache folder so they don't clutter up output folders if OT crashes during export before the cleanup command runs.


shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. @turtletooth thank you for the improvement!

Issue #858, opened 2016-10-09
Floating point exception in Control Point Editor Tool.


jabarrera:
Hi.

When I use the **control point editor tool** I always get a crash (**floating point exception**). I'm trying to reproduce an operation that always throw the error, but I can not. Anyway, I just make a simple draw with vector brush, just start clean Opentoonz, select the control point editor tool, make some operations (to move vertex or spline controls) and I get the error randomly, aparently.

I'm using an appimage compilation (thanks to Morevnaproject) for 32 bits in a GNU/Linux Debian. Opentoonz version 1.1.1. I think it is nigthly builds compilation, but I'm not sure.

Thanks.  :)

Scheherazade:
I've been having the same problem on my Mac for a long time now. It works fine on Windows 10 for some reason.

Xenobus:
I can reproduce it consistently. Looks to be related to control point initialization and only occurs when trying to manipulate a curve by clicking on it or dragging it directly. Using the knobs and moving the control points themselves doesn't seem to ever crash, but it's very easy to misclick on the curve itself.

If you have a completely straight-lined shape, like something made with the geometric tool, you can click or drag the line without crashing. It will spawn the little curve-editing handles upon the first click and remain non-crashing, even after saving, restarting and reloading.
Curves made in other fashion seem to always crash when clicked, including those to which additional control points have been added by any method.

Xenobus:
I thought I'd build it from the source and try to pinpoint the issue, but I could not get this bug to occur in my own build.
Seems to only be an issue specific to Morevna Project's AppImage releases, which I was also using before. Currently persists to their latest build - at the time of writing, 2016.11.13-5bb68, both 32 and 64-bit releases.

Probably something worth taking up with them instead, then.

ideasman42:
Possibly FPE's are handled differently depending on the build options, I don't get crashes here but there are quite a few divide by zeros when using painting tool (just dragging the cursor)

- toonz/sources/toonzqt/icongenerator.cpp:221:45: runtime error: division by zero
- toonz/sources/common/tvectorimage/cornerdetector.cpp:134:61: runtime error: division by zero
- toonz/sources/common/tvectorimage/cornerdetector.cpp:136:61: runtime error: division by zero

Would be good to resolve these.

PR #922, opened 2016-11-17
Fix crash on editing curve with control point tool

Xenobus:
OpenToonz crashes when directly manipulating curves via the lines
themselves, unless the curve was initialized as a straight line.

Depending on the build environment, a calculation interprets doubles
as ints, resulting in a divison by zero.

 -Explicitly specify the use of floating point abs() - fabs()

Resolves: #858, #919

### DIFF
--- a/toonz/sources/tnztools/controlpointselection.cpp
+++ b/toonz/sources/tnztools/controlpointselection.cpp
@@ -873,7 +873,7 @@ void ControlPointEditorStroke::moveSegment(int beforeIndex, int nextIndex,
       m_controlPoints[beforeIndex].m_isCusp = true;
   } else if (!isSpeedOutLinear(beforeIndex) && !isSpeedInLinear(beforeIndex) &&
              !isCusp(beforeIndex)) {
-    t = 1 - abs(w - w0) / abs(w4 - w0);
+    t = 1 - fabs(w - w0) / fabs(w4 - w0);
     moveSingleControlPoint(beforeIndex, t * delta);
     t = 1 - t;
   }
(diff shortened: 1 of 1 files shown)


ideasman42:
Thanks for the fix, note that there are other suspicious uses of `abs` where its likely `fabs` should be used, worth looking into though best ensure each one is *really* a mistake.

`float xRatio = 1.0f - abs(pos.x - p0.x)`, all inputs are floats.

https://github.com/opentoonz/opentoonz/blob/792dc2b6b402d622292efe815bfaa2635c43eeb9/toonz/sources/stdfx/iwa_motionblurfx.cpp#L317

This is almost certainly a mistake.

`if (abs(cosf(part->flap_phi * 3.14159f / 180.0f)) < 0.03f) {`

https://github.com/opentoonz/opentoonz/blob/792dc2b6b402d622292efe815bfaa2635c43eeb9/toonz/sources/stdfx/iwa_particlesengine.cpp#L881

Xenobus:
So I noticed.
I grepped through looking for instances, and there doesn't seem to be all that many.
Thinking on it a bit and testing some more, it seems to me that problem stems from abs() being resolved to C's implementation, which can't handle floating points. The C++/std implementation just overloads to whatever is needed.

There's no real consistency in whether or not fabs() or std::abs() is used right now, though. I kind of like the idea of being explicit, but specifying std everywhere would handle both cases, including the issue fixed here. I went with fabs because it was used elsewhere in the same file. Having the same solution across the board would be a good idea, regardless of choice.

Would there be any particular reason not to go through the lot and just replace all instances of fabs and abs with std::abs?

ideasman42:
@Xenobus for C++ code, agree it makes sense to go with `std::abs()`,

Best do this as a PR that makes no functionality changes, so it doesn't accidentally introduce regressions.

Xenobus:
@ideasman42 Guess I'll get to it in the near future.

Should instances of `abs()` in the `std` namespace also get prefixed?
Use of that namespace in general is a bit sketchy, as it's just another detail of which to be mindful. I'm not sure if there are any other ambiguities at present, but those would be unknown unknowns.

ideasman42:
@Xenobus I'm not responsible for defining code standards here, though it would be good to settle on a convention.
Could other devs comment on this?

PR #924, opened 2016-11-17
Fixed gif import

turtletooth:
I broke gif import on #827 

This fixes it.


ideasman42:
Since there are 4x `QRegExp` 's, wouldnt it be better to make a static function? `filepath_clean_for_ffmpeg` or something like this?

turtletooth:
@ideasman42 There you go again making sense.

turtletooth:
@shun-iwasawa This is a pretty simple PR.  Would you mind doing a quick look.  I would hate for the next release to have broken gif import.  (Sorry I broke it in the first place)

shun-iwasawa:
Confirmed that GIF files can be imported properly now. Jenkins

shun-iwasawa:
LGTM. @turtletooth thank you for the fix!

Issue #923, opened 2016-11-17
Suspicious/incorrect use of `abs`


ideasman42:
There are uses of `abs` in opentoonz where it seems  `fabs` was intended (without converting to int).

See: bea8195c08ccfe23284105f144c07f3ca75d0ece for an example of the problem.

Other examples:

[`float xRatio = 1.0f - abs(pos.x - p0.x)`, all inputs are floats.](https://github.com/opentoonz/opentoonz/blob/792dc2b6b402d622292efe815bfaa2635c43eeb9/toonz/sources/stdfx/iwa_motionblurfx.cpp#L317)

This is almost certainly a mistake.

[`if (abs(cosf(part->flap_phi * 3.14159f / 180.0f)) < 0.03f) {`](https://github.com/opentoonz/opentoonz/blob/792dc2b6b402d622292efe815bfaa2635c43eeb9/toonz/sources/stdfx/iwa_particlesengine.cpp#L881)

[`illuminant = abs(normVec.x * lightVec.x + normVec.y * lightVec.y + normVec.z * lightVec.z);`](https://github.com/opentoonz/opentoonz/blob/792dc2b6b402d622292efe815bfaa2635c43eeb9/toonz/sources/stdfx/iwa_particlesengine.cpp#L928)

shun-iwasawa:
Great catch. @ideasman42 and @Xenobus thank you so much for pointing out the problem!

RodneyBaker:
@shun-iwasawa  I have taken the liberty of unassigning you from this issue so that others will feel free to give it a look and take it on.  Please feel free to self assign again as you see fit.  I'd say you are still the most likely person to take it on.

RodneyBaker:
Some of this has been addressed via #3226 but other parts still remain...

RodneyBaker:
Closing as likely resolved by #3226.
Please feel free to open a new report on suspicious/incorrect usage that might have been missed.

PR #925, opened 2016-11-18
[WIP] Using 'fabs' instead of 'abs'

shun-iwasawa:
This PR is for #923
@ideasman42 and @Xenobus  thanks again for pointing this.

### DIFF
--- a/toonz/sources/stdfx/iwa_adjustexposurefx.cpp
+++ b/toonz/sources/stdfx/iwa_adjustexposurefx.cpp
@@ -121,8 +121,9 @@ void Iwa_AdjustExposureFx::doCompute_CPU(TTile &tile, double frame,
   float scale    = (float)m_scale->getValue(frame);
   float offset   = (float)m_offset->getValue(frame);
 
-  float exposureOffset = (powf(10.0f, (float)(abs(offset) / hardness)) - 1.0f) *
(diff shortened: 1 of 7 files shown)


shun-iwasawa:
Jenkins

ideasman42:
Mightn't it be better to use `std::abs` instead of `abs/fabs/fabsf`... etc.

For example, in this patch `(float)(fabs(offset)` in this patch should really use `fabsf` (since its operating on floats, doing float -> double -> float cast is redundant, although the compiler may optimize it out its still confusing).

Using `std::abs` may be better overall to avoid unintended type coercing.

shun-iwasawa:
That's make sense. I'll work on this after releasing the next version.

jpturcotte:
Any update on that?

RodneyBaker:
Took a quick look at this report...

As std::abs is overloaded for both signed integer and floating point types it does appear appropriate to replace both abs and fabs with std::abs in the source.

Aside:  The use of abs instead of fabs (or vice versa)... integer versus float... may directly relate to several bugs, crashes and abnormal calculations where data appears to be rounded, truncated or simply wrong.  As a wild speculation it might also result in additional/unnecessary processing.

RodneyBaker:
Because of #3226 this PR may no longer be needed.

RodneyBaker:
Closing as this should be implemented by #3226

PR #928, opened 2016-11-21
Level settings and info for sound levels 

shun-iwasawa:
For now, "Info" command triggered from right-click menu of the sound levels does nothing (for bug), so there is no easy way to know the file path and other information of them.
This PR fixes such problem, allowing to display the Level Settings and the Info Viewer for sound levels.


shun-iwasawa:
Jenkins

turtletooth:
LGTM Thanks @shun-iwasawa for the update.

PR #930, opened 2016-11-21
Fix/playaudio linux #863

### LINKED ISSUE #863: Load WAV level -> Get creepy sound.
Hi!

When I load a **WAV sound**, I always get a continuous creepy very loud sound (for example, when I play it from the level in the x-sheet).

I suppose that it must be related with the **frequency** or something related (maybe only happens in GNU/Linux), but I don't get one good WAV (and I think that must be accepted all of them).

I am using Opentoonz and ffmpeg in GNU/Linux.

One I tried was this: https://dl.dropboxusercontent.com/u/58974111/11SecClub_May_Competition.wav

Thanks and greetings.

jabarrera:
Hi.

This resolve #863 issue.

### DIFF
--- a/toonz/sources/common/tsound/tsound_sdl.cpp
+++ b/toonz/sources/common/tsound/tsound_sdl.cpp
@@ -89,11 +89,11 @@ public:
     if (m_data->imp) {
       if (m_data->m_doNotify == false) return;
       m_data->m_doNotify = false;
-      if (m_data->imp->m_isPlaying) m_data->imp->doStopDevice();
       std::set<TSoundOutputDeviceListener *>::iterator it =
           m_data->imp->m_listeners.begin();
       for (; it != m_data->imp->m_listeners.end(); ++it)
         (*it)->onPlayCompleted();
+      if (m_data->imp->m_isPlaying) m_data->imp->doStopDevice();
     }
   }
 };
@@ -134,6 +134,8 @@ static void sdl_fill_audio(void *udata, Uint8 *stream, int len) {
 
   /* Mix as much data as possible */
   len = (len > audio_len ? audio_len : len);
+
+  SDL_memset(stream, 0, len);  // Mix with silence
   SDL_MixAudio(stream, (Uint8 *)myData->entireFileBuffer + myData->byteOffset,
                len, _this->m_volume);
   myData->byteOffset += len;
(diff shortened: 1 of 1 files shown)


ideasman42:
Note, it looks like removing braces around `TThread::ScopedLock sl(MutexOut);` would change its behavior when un-commented (probably it should be removed entirely, or left in with braces).

Would prefer fixes be completely separated from superficial changes, so they can be reviewed more easily.

@mmuman would you be able to check this patch?

mmuman:
Indeed, since it's constructed on stack its scope is that of the enclosing braces, so if you remove the braces the lock will stay until the end of the method.
I didn't really look into it though.

jabarrera:
@ideasman42, you are absolutely right about the braces. I prefer, too, to recover them and let the fix isolated.

I revert the deleted braces/block in the last commit (https://github.com/opentoonz/opentoonz/pull/930/commits/15bf4faa8488a44172205eed0c6a54137bb31544).

Thanks to @mmuman, too, for the point.

mmuman:
For such small changes you can squash your commits and force push the branch btw.

jabarrera:
It seems the **OSX Travis** check failed with a _timeout_.

Could someone run again that check?

Thanks!

PS: I know that it is not a real problem, but I hate see the red cross. I want my green mark! : ))

ideasman42:
Merged with minor edits, let `#if defined(i386)` block in, better handle this separately.

jabarrera:
I'm sorry, @ideasman42 , but if you let the define(i386) block the sound is not fixed for 32 bits.

I'm using 32bits compilation, and in my tests the operation maden for 16 bits sounds in 32 bits platform is not need (make noisy). So the same operation for 32 and 64 bits are enough.

Is not it better goes in the same fix?


Thanks.

ideasman42:
@jabarrera, ah, thought this was part of some cleanup,
thanks for the heads up - I can't i386 easily, so taking your word for it,
committed e8ad7771865125fcdd5def005c5742cd7fc98c2a

In the future it would be good to give a bit of info on each change, just a few lines explaining what a change is needed - if its not totally obvious.

jabarrera:
@ideasman42 , you are absolutely right. My PR wasn't too much (in fact, anything at all  :)) well explained. My fail.

Thank you very much for this and for the work you made with the Linux compilation.

PR #927, opened 2016-11-18
Use English for 'iwa_motionblurfx' comments

ideasman42:

PR #932, opened 2016-11-22
Use English for 'iwa_motionblurfx' comments by ideasman42 [modified]

shun-iwasawa:
This PR is based on #927 by @ideasman42 , modified some translations in order to clarify.

@ideasman42 It would be helpful for other developers. Thanks!

### DIFF
--- a/toonz/sources/stdfx/iwa_motionblurfx.cpp
+++ b/toonz/sources/stdfx/iwa_motionblurfx.cpp
@@ -1,7 +1,7 @@
 /*------------------------------------
 Iwa_MotionBlurCompFx
-Amount of light/Motion blur considering the trajectory of the object
-Enable synthesis of exposure value with background.
+Motion blur fx considering the exposure of light and trajectory of the object.
+Available to composite exposure value with background.
 //------------------------------------*/
 
 #include "iwa_motionblurfx.h"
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

ideasman42:
Seems fine, any reason for odd word wrapping?

I ran the patch through clang-format so thought it wouldnt need any edits in that respect, eg:

picky, but would avoid odd-line splitting.

```
        /* Calculate the inner product of 'p0' -> sampling point and 'p0' ->
         * 'p1' */
```

```
      /* Do depremultiply on sources that are premultiplied, such as regular
      * Level.
      * It is not done for 'digital overlay' (image with alpha mask added by
      * using
      * Photoshop, as known as 'DigiBook' in Japanese animation industry) etc.
      */
```

yamisama:
Very small point: I believe the common terminology for the opposite of "to premultiply" is "to unpremultiply", not "to depremultiply". I'm not sure what term the original Alvy Ray Smith paper uses do describe that operation, but in most software today "to unpremultiply" seems to be sort of a consensus.

shun-iwasawa:
Fixed odd word wraps caused by clang-formatting. Thanks @ideasman42 for pointing it out.
Also changed the term 'depremultiply' and 'demultiply' to 'unpremultiply' - it is true that 'unpremul-' is more common as @yamisama pointed. I learned that for the first time.
Jenkins

ideasman42:
LGTM, note: would squish all commmits before merging.

shun-iwasawa:
Squashed.
Jenkins

ideasman42:
Any reason to hold off putting this in master?

PR #945, opened 2016-12-02
Change the default size of field guide

shun-iwasawa:
This PR is to change the default size of the field guide in order to fit the default camera size and aspect ratio.


shun-iwasawa:
Jenkins

turtletooth:
LGTM - Thanks @shun-iwasawa

PR #956, opened 2016-12-09
Fix typo

shun-iwasawa:


shun-iwasawa:
I'll merge this since it is an obvious typo.

Issue #854, opened 2016-10-06
Can not build against SuperLU_5.2.1 on Linux


jcome:
Hello, I am trying to build OpenToonz.flatpak, [the new framework for desktop applications on Linux](http://flatpak.org). When compiling tlin_superlu_wrap.cpp, I got the following error:

> ...
> /run/build/opentoonz/toonz/sources/tnzext/tlin/tlin_superlu_wrap.cpp:312:36: error: cannot convert ‘SuperLUStat_t_’ to ‘GlobalLU_t_’ for argument ‘12’ to ‘void dgstrf(superlu_options_t_, SuperMatrix_, int, int, int_, void_, int, int_, int_, SuperMatrix_, SuperMatrix_, GlobalLU_t_, SuperLUStat_t_, int*)’
>           F->L, F->U, &stat, &result);
>                                     ^
> 
> tnzext/CMakeFiles/tnzext.dir/build.make:734: recipe for target 'tnzext/CMakeFiles/tnzext.dir/tlin/tlin_superlu_wrap.cpp.o' failed
> ...

more details: 
[build-log.txt](https://github.com/opentoonz/opentoonz/files/514252/build-log.txt)

the flatpak Manifest and patch used
[OpenToonz Flatpak.zip](https://github.com/opentoonz/opentoonz/files/514263/OpenToonz.Flatpak.zip)

Thanks for your help.

ideasman42:
Resolved in master

PR #958, opened 2016-12-11
[FIX] Plastictool: SuperLU non thirdparty issue on Linux.

### LINKED ISSUE #623: Plastic Tool, Erratic behavior and crash. Linux.
Plastic Tool creates bad mesh, a box mesh #406, and crash in animate mode. Here is the video:

https://dl.dropboxusercontent.com/u/69554201/OpentoonzPlasticToolCrash.mp4

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/35905413-plastic-tool-erratic-behavior-and-crash-linux?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jabarrera:
Hi.

This PR fixes the problem with using the non thirdparty SuperLU (the upstream). So, it close #623 issue.

Opentoonz are using a modified version of SuperLU(v.4.1) because multi threading issues with static variables (see https://github.com/opentoonz/opentoonz/blob/master/thirdparty/superlu/README.txt) . SuperLU >=5 becomes thread-safe, so it is safe use it instead of the thirdparty.

Anyway, SuperLU 5+ change the interface of the method used for factorizing, so it is not enough linking with the new library.

**What happens**:

Without this fix, in Linux box if the upstream SuperLU is used, you get a "Segmentation fault" (get it when select "Animate" before create the mesh and the bones) because the function has a mixing of parameters.

There is a mixing of the headers file, too. Always gets the headers of the thirdparty directory.

**What do the fix**:

_In build_:
	Change CMake configuration to get the correct header when you are using the thirdparty or the official upstream of SuperLU.
	Added a validation for correct version of SuperLU library (>=5) in casef you use non-thirdparty.

_In code_:

If using the non-thirdparty, it uses the new interface of the function of factorizing (dgstrf) with the new parameter.

If thirdparty used, do nothing. I've tried to make it the more transparent for non-Linux platforms, in case the thirdparty SuperLU wanted to be kept.

As said, the fix is only for Linux. Microsoft Windows and OSX are using still the thirdparty; although I think that it could be interesting use the official upstream SuperLU for all platforms ( @shun-iwasawa ). As it works in Linux, it must works fine in Windows and OSX, and it does the code more unified and easy to mantain. Anyway, as I can only test on Linux, I prefer that someone related with Microsoft Windows and OSX checks this.

This is my first time with CMake, so it is possible that are something missing or incorrect. Just comment it.  :)

Thanks.

### DIFF
--- a/toonz/cmake/FindSuperLU.cmake
+++ b/toonz/cmake/FindSuperLU.cmake
@@ -1,15 +1,4 @@
 # preferred homebrew's directories.
-find_path(
-    SUPERLU_INCLUDE_DIR
-    NAMES
-        slu_Cnames.h
-    HINTS
-        ${THIRDPARTY_LIBS_HINTS}
-    PATH_SUFFIXES
-        superlu43/4.3_1/include/superlu
-        superlu/SuperLU_4.1/include
-)
-
 find_library(
     SUPERLU_LIBRARY
     NAMES
@@ -22,6 +11,36 @@ find_library(
         superlu43/4.3_1/lib
         superlu
 )
+
+string(TOLOWER ${SUPERLU_LIBRARY} SUPERLU_LIB_CASE_INSENSITIVE)
+
+if (${SUPERLU_LIB_CASE_INSENSITIVE} MATCHES "thirdparty")
+    find_path(
+        SUPERLU_INCLUDE_DIR
+        NAMES
+            slu_Cnames.h
+        HINTS
+            ${THIRDPARTY_LIBS_HINTS}
+        PATH_SUFFIXES
+            superlu43/4.3_1/include/superlu
+            superlu/SuperLU_4.1/include
+    )
+else()
+    set(SUPERLU_NO_THIRDPARTY "true")
+    find_path(
+        SUPERLU_INCLUDE_DIR
+        NAMES
+            slu_Cnames.h
+        PATH_SUFFIXES
+            superlu
+    )
+    #Get the version of Superlu. We need >= 5.0.0
+    if (SUPERLU_INCLUDE_DIR)
+        file(STRINGS "${SUPERLU_INCLUDE_DIR}/slu_util.h" SUPERLU_VERSION REGEX "#define SUPERLU_MAJOR_VERSION.+[0-9]+")
+        string(REGEX MATCH "[0-9]+" SUPERLU_VERSION ${SUPERLU_VERSION})
+    endif()
+endif()
+
 
 message("***** SuperLU Header path:" ${SUPERLU_INCLUDE_DIR})
 message("***** SuperLU Library path:" ${SUPERLU_LIBRARY})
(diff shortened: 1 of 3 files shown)


jabarrera:
It seems that Travis Ubuntu has an outdated version (<5) of SuperLU. Could someone confirm that?

ideasman42:
Thanks for investigating a fix.

Some comments...

In general, checking for `thirdpart` seems wrong - really we want to know the superlu version (thirdparty may be updated - or the user may have an old superlu installed - though Im not so much worried about that case).

Wouldn't it be better just to check the version in the code and not add `SUPERLU_NO_THIRDPARTY` ?

----

Im using superlu5.2 from https://aur.archlinux.org/packages/superlu/
And ` SUPERLU_MAJOR_VERSION` isnt defined (upper or lowercase).

----

Don't think the `ifdef` that switches between angle bracket and quote includes is needed, better use angle brackets for all cases since its a 3rd party library, even when bundled.

jabarrera:
Hi, @ideasman42 .

**About checking the version of SuperLU**:

It's strange that thing with **Arch Linux**. That macro are defined by the upstream (you can see it http://crd-legacy.lbl.gov/~xiaoye/SuperLU/superlu_5.2.1.tar.gz ). I can't believe that Arch maintainer of the package change that. Are you sure you are looking in the right include file? The macro is in **slu_util.h** in case you think that it must be in the same file used in the _file_path_ function. (I could had use slu_util.h to search, but I prefered to maintain the same search of thirdparty).

I think that know the version in build time gives to the user a more comprensive message in case that the user have an old version in build time. And as I need to get the correct includes (with the old search, always got the includes of thirdparty) I prefered to define our own macros.

**About includes**:

You are right. But when I thought about what to do, I tried to maintain the same code of non-Linux platforms. As Windows and OSX are using the thirdparty, yet ( : ) ), I let that ugly duplication. :)

I not consider thirdparty as external library in this case but... I understand your point.


Please, confirm that you prefer unify the includes of SuperLU in angle form to make the change.

jabarrera:
This PR should close #854 , too.

ideasman42:
Applied fix based on yours 3ebaf33693caa2d9faf3cfad864f84c638e2cabe

Now its possible to build working plastic tool. `./toonz/cmake/FindSuperLU.cmake` still defaults to using the thirdparty include-path. For now the `SUPERLU_INCLUDE_DIR` can be manually set. But we will want a way to switch between thirdparty and system libs.

Rather handle this as a separate PR.

ideasman42:
Applied fix based on yours 3ebaf33693caa2d9faf3cfad864f84c638e2cabe

Now its possible to build working plastic tool. `./toonz/cmake/FindSuperLU.cmake` still defaults to using the thirdparty include-path. For now the `SUPERLU_INCLUDE_DIR` can be manually set. But we will want a way to switch between thirdparty and system libs.

Rather handle this as a separate PR.

ideasman42:
Added issue #962 to keep track of the issue.

PR #963, opened 2016-12-13
Fix crash on applying shortcut preset

shun-iwasawa:
To reproduce crash:

1. Open "Configure Shortcuts" popup.

1. In Preset combobox, select "OpenToonz" and press "Apply" button.

The software crashes while setting the shortcuts.

For now, in OpenToonz shortcut preset there is a setting with ID "MI_ReloadStyle", corresponding to ”Reload qss” command which is only available in debug build. 
In the release build the software crashes because no action is obtained with such ID.

### DIFF
--- a/toonz/sources/toonz/shortcutpopup.cpp
+++ b/toonz/sources/toonz/shortcutpopup.cpp
@@ -527,6 +527,7 @@ void ShortcutPopup::setPresetShortcuts(TFilePath fp) {
     QByteArray ba      = id.toLatin1();
     const char *charId = ba.data();
     action = CommandManager::instance()->getAction((CommandId)charId);
+    if (!action) continue;
     CommandManager::instance()->setShortcut(
         action, preset.value(id).toString().toStdString(), false);
   }


shun-iwasawa:
Jenkins

ideasman42:
Is it really necessary to Jenkins test NULL checks like this?

If its crashing and you can redo the error - running test suite over the code seems more of a ritual then actually helping the development process.
Does Jenkins even cover this code beyond checking it compiles?

Of course in some cases a simple NULL check might be implicated in more complex scenarios, but this case looks quite trivial.

Tests are fine but they should help with development - not stall trivial fixes from being applied IMHO.

shun-iwasawa:
@ideasman42 thank you for your suggestion.
A few months ago we modified Jenkins not to delete the intermediate files before&after each test. It may lose accuracy but reduces the build time very much. For now, it usually takes less than 30 minutes which is IMO acceptable to test any PRs even if they are just small changes.
On the other hand, your suggestion indeed makes sense that some simple & obvious fix like this PR does not need to do Jenkins test. I agree with you that waiting for the test before merge is not must-do in such case.

turtletooth:
@shun-iwasawa Thanks for catching this.  I apologize that I didn't catch the reload qss command.

turtletooth:
LGTM - Thanks

Issue #469, opened 2016-06-18
Filenaming:  Crash with range of drawings starting with 0000


Rayek:
I use Krita to export animations as a PNG numbered file sequence: name.XXXX.png. Krita saves the first frame as name.0000.png.

When I import these frames into my project's drawing folder, the instant I use the file browser in OpenTOonz to open the drawings folder, OT crashes. If the file browser happens to be pointing to the drawings folder while I place such a range in the drawings folder, OT crashes.

To work around this issue, I have to either manually delete the first frame (file), or rename the entire range. This is a bit frustrating.

OT cannot handle a numbered range of images starting with 0000. 

It would be great if OT would accept both options: image ranges beginning with either 0000 or 0001 would be accepted as a valid animation sequence.

At the very least it ought not to crash instantly when encountering a sequence of images starting with 0000.

MisterABK:
In addition to the problem with frame sequences beginning 0000, Opentoonz also ignores frame sequences which have too many leading zeroes (eg. "imgseq_00001.tif") and treats them as separate images. It also doesn't seem to accept sequences where the number is not separated by an underscore (eg. "imgseq0001.tif").

artisteacher:
OT does recognize sequences with a period (i.e. imgseq.0001.tif). This causes a problem with some of the my paint brush icons not being recognized #1382. My camera automatically saves images with the name.0001.jpg format, so that ends up working really well. 

Just like how the load/import preference can be set, it would be helpful to have a preference to choose between loading a selection of images as an image sequence or as separate levels.

RodneyBaker:
A problem related to this is that a way some programs know to render out to an image sequence and to a specific format is to specify the output in the form "image000.png" where 'image' is the filename, the zeros related the padding for digits desired and ".png" relates the desired format.  This is a great timesaver because it saves the user from having to navigate anywhere to specify the number of trailing digits or the filename.  

The problem of course is that OT sees that filename (even with added periods generated from the other program) as malformed.

Ideally OpenToonz would recognize any of these common filenaming conventions as legitimate sequences:

image0.png

thisonealso00.tga

andthistoo.tif

The filename and the extension aren't problematic in any way but the numerical sequencing is even before any attempt to bring the file into OT.  The example of this can be seen occassionally where OT encounters what it deems as a malformed file (or more likely a sequence of files) and throws an error or crashes.  

[...]

PR #966, opened 2016-12-15
Fix Opentoonz Filebrowser crash when a sequence has a frame with a  malformed named.

jabarrera:
Hi.

This PR close partially #469 issue: only fix de crash.

**What happen**:

Opentoonz file browse crash when enter in a directory with a file name that start a sequence with 0 (file.0000.png, for example) or have letters (except a final "a") file.0wre.png.

That's because for Opentoonz a sequence can not start in 0 or contain letters (except "a". I suppose that this a feature for alpha images... but I really don't know).

This is really bad because an user can have a file named that way and it just crash Opentoonz.

**How reproduce this**:

Create a image file named file.0000.png. Navigate with file browser to the file created. You'll get a crash.

**What do this fix**:

This fix throws and capture exception more gentle avoiding the crash. If it find a file with a malformed name, it skips that frame, showing warning messages in the message center.

### DIFF
--- a/toonz/sources/common/timage_io/tlevel_io.cpp
+++ b/toonz/sources/common/timage_io/tlevel_io.cpp
@@ -5,6 +5,7 @@
 #include "tiio.h"
 #include "tcontenthistory.h"
 #include "tconvert.h"
+#include "tmsgcore.h"
 
 // STD includes
 #include <map>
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
Jenkins
I understand that accepting a sequence starting with 0 (WIP in #967) would be better, but this indeed fixes the crash and worth merging as an quick resolution. Thanks @jabarrera !

PR #967, opened 2016-12-15
(WIP) Filenaming:  Sequence can start in 0 not only 1

### LINKED ISSUE #469: Filenaming:  Crash with range of drawings starting with 0000

jabarrera:
Hi.

This PR close #469. (See #966, too)

**What happen**:

Opentoonz not accept sequence starting with 0. This is explicitaly prohibed in code. One reason is that "0" frame is used to delete a cell.

I suppose that starting with 1 is a industry standard, but I think that it must be possible for the user start with 0 (if he wants it).

**What do this fix**:

Activate the use of 0 frame to start a sequence.

**What problem with enter this fix**:

As said, setting a cell with 0 frame is used to delete. I suppose that this feature is used (as is programmed), but I really don't know.

If deleting is not used, this fix could be considered as is.

if deleting is used and, of course, is prefered continuing with this, I could make an options in preference configuration that let to the user select if he wants start with 0 or continue as now (this will be the default option).

I accept other suggestion (including forgot this  :)).

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -1049,6 +1049,16 @@ void CellArea::drawLevelCell(QPainter &p, int row, int col, bool isReference) {
   TXshCell cell = xsh->getCell(row, col);
   TXshCell prevCell;
 
+  // If the user change the option to start a sequence with 0 to 1 in-live, we
+  // delete all cell and frame 0 of the actual Xsheet
+  if (!Preferences::instance()->isSequenceCanStartWith0() &&
+      cell.getFrameId().getNumber() == 0) {
+    cell.getSimpleLevel()->eraseFrame(TFrameId(0));
+    xsh->clearCells(row, col);
+    cell.getSimpleLevel()->setDirtyFlag(true);
+    return;
+  }
+
   TCellSelection *cellSelection     = m_viewer->getCellSelection();
   TColumnSelection *columnSelection = m_viewer->getColumnSelection();
   bool isSelected                   = cellSelection->isCellSelected(row, col) ||
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -528,7 +528,13 @@ void RenameCellField::showInRowCol(int row, int col) {
                         m_viewer->getFrameNumberWithLetters(fid.getNumber()));
     else {
       std::string frameNumber("");
-      if (fid.getNumber() >= 0) frameNumber = std::to_string(fid.getNumber());
+      // Check if the user wants a sequence starting with 0
+      if ((Preferences::instance()->isSequenceCanStartWith0() &&
+           fid.getNumber() >= 0) ||
+          (!Preferences::instance()->isSequenceCanStartWith0() &&
+           fid.getNumber() > 0)) {
+        frameNumber = std::to_string(fid.getNumber());
+      }
       if (fid.getLetter() != 0) frameNumber.append(1, fid.getLetter());
       setText((frameNumber.empty())
                   ? QString::fromStdWString(levelName)
@@ -637,7 +643,14 @@ void RenameCellField::renameCell() {
 
   TXshCell cell(xl, fid);
 
-  cellSelection->renameCells(cell);
+  if (!Preferences::instance()->isSequenceCanStartWith0() &&
+      (fid.getNumber() == 0)) {
+    TCellSelection::Range range = cellSelection->getSelectedCells();
+    cellSelection->deleteCells();
+    // revert cell selection
+    cellSelection->selectCells(range.m_r0, range.m_c0, range.m_r1, range.m_c1);
+  } else
+    cellSelection->renameCells(cell);
 }
 
 //-----------------------------------------------------------------------------
@@ -1169,7 +1182,13 @@ void CellArea::drawLevelCell(QPainter &p, int row, int col, bool isReference) {
     else {
       std::string frameNumber("");
       // set number
-      if (fid.getNumber() >= 0) frameNumber = std::to_string(fid.getNumber());
+      // Check if the user wants a sequence starting with 0
+      if ((Preferences::instance()->isSequenceCanStartWith0() &&
+           fid.getNumber() >= 0) ||
+          (!Preferences::instance()->isSequenceCanStartWith0() &&
+           fid.getNumber() > 0)) {
+        frameNumber = std::to_string(fid.getNumber());
+      }
       // add letter
       if (fid.getLetter() != 0) frameNumber.append(1, fid.getLetter());
       p.drawText(nameRect, Qt::AlignRight, QString::fromStdString(frameNumber));
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -528,7 +528,7 @@ void RenameCellField::showInRowCol(int row, int col) {
                         m_viewer->getFrameNumberWithLetters(fid.getNumber()));
     else {
       std::string frameNumber("");
-      if (fid.getNumber() > 0) frameNumber = std::to_string(fid.getNumber());
+      if (fid.getNumber() >= 0) frameNumber = std::to_string(fid.getNumber());
       if (fid.getLetter() != 0) frameNumber.append(1, fid.getLetter());
       setText((frameNumber.empty())
                   ? QString::fromStdWString(levelName)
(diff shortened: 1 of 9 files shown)


shun-iwasawa:
@jabarrera thank you very much for the modification!
"Get empty cell by typing 0" feature is useful for inputting Xsheet only with numpad keys. Especially in Japanese animation industry, they are familiar with such operation since it is the same as QuickChecker by Celsys. So IMO it would be better to make your fix an option in preference as you kindly suggested, in terms of consistency with other applications.

jabarrera:
Hi.

I have encapsulate this functionally in am user preference options. It is located in **XSheet** item

preferences_sequence_0

I have used the concept of "sequence", but I don't know if it is fine. Feel free to change it or request a change with the new suggestion.

If the user checks on the option (he wants start levels with 0), he can load sequences with 0 (file.0000.png) and when he draw a new level, it starts with 0 frameid.

automatic_creation_level


Options related with frames (start and end) when creating a new level change to 0 by default.

new_level_popup


If the user changes again the option and there is a Xsheet active (with levels), the frames and cells with 0 are deleted. Please, checks this. I think that it is right, but maybe I missed something in this process.

I don't like too much this deleting. Maybe making a renumbering of all cells and frames could be a cleanest process, but I suggest it to a future enhacement, as this, probably, require a deeper factoring.


There is an "obscure" process that I prefered not change. In **Scene Setting**, you can set the **Start Frame** (even in 0). But looking to the code, it doesn't make what I expected it did. It just only updates the position of the markers (I expected that the Xsheet first row was setting to 0). So, I let it as was.



Please, feel free to add whatever suggestion or missed.

jabarrera:
Hi.

I'm confused..

It seems that there is a problem linking the project (maybe a circular dependency with the entering of Preferences in tfilepath) but in my local box works fine (compilation and linking), so, I don't know what to fix.

Could anyone help with this?

Thanks.

shun-iwasawa:
Hi @jabarrera , sorry for replying you late & a happy new year!

You cannot use `Preferences` class directly in tfilepath.cpp, since `TFilePath` belongs to `tnzcore` and `Preferences` belongs to `toonzlib `which depends on `tnzcore`.

How about to adding a static value to TFilePath class which monitors `Preferences::instance()->isSequenceCanStartWith0()` value? 
Please refer to the usage of `bool TFilePath::m_underscoreFormatAllowed` as it seems to be made for the same purpose. For now, `m_underscoreFormatAllowed` is always set to `true` but you can find a static method (`static void setUnderscoreFormatAllowed(bool)`)to control it from child projects.

jabarrera:
Don't worry for the lateness, @shun-iwasawa , and happy new year for you, too.  :)

Thanks for the point. I thought that it could be fixed in build script, but I think this way is right, too. In my local box works good, but the last before works fine, too; so... Let's see!

shun-iwasawa:
Jenkins

toonz/sources/toonz/xshcellviewer.cpp
```diff
@@ -1042,6 +1049,16 @@ void CellArea::drawLevelCell(QPainter &p, int row, int col, bool isReference) {
   TXshCell cell = xsh->getCell(row, col);
   TXshCell prevCell;
 
+  // If the user change the option to start a sequence with 0 to 1 in-live, we
+  // delete all cell and frame 0 of the actual Xsheet
+  if (!Preferences::instance()->isSequenceCanStartWith0() &&
+      cell.getFrameId().getNumber() == 0) {
+    cell.getSimpleLevel()->eraseFrame(TFrameId(0));
```
shun-iwasawa: I agree that we have to think how to handle the 0-start levels in the environment WITHOUT the 0-start option. 

You are erasing the frame data here only in the drawing function of Xsheet.
But if user works without opening the Xsheet panel, this operation will not be called.
I think "controlling data" operations like this should not be in "viewing data" function.

How about to making this option to demand users to restart the software, 
and moving this operation to the moment when when loading levels?
By restarting the software, you don't need to take care of the existing scene.

In addition, IMO it would be better to open some confirmation dialog before erasing the frame 0 data, 
which enables users to choose whether to erase frame 0 or to activate the preference option.

jabarrera: Good point about the X-sheet non active. I really don't think that case. :D 

I'll check out make this a reboot setting as you suggest. But, instead of reset Opentoonz, do you see accesible only closing and opening the actual scene programatically?

Agree with the confirmation popup. I din't like the deleting, too. I thought in a re-numbering instead of the deleting. I think that this functionality is made... but I don't know if it is interesting do it that way or not.

I'm going to re-think this a bit. I'm setting this again to WIP status, because I don't know when I'll get the time to put on this.

shun-iwasawa: > do you see accesible only closing and opening the actual scene programatically?

Yes, I agree that just closing and start a new scene would work. I think you can do it by triggering "New Scene" command like as follows:
`CommandManager::instance()->getAction("MI_NewScene")->trigger();`

I realized that allowing frame 0 will have influence over more factors in the software than expected. Thank you very much for tackling this and your contribution!

jabarrera: Thanks for the correct point, @shun-iwasawa . I'll try to check it out as soon as possible.

You are right. This software is bit huge, and make yet little change it can become really complex. :)

turtletooth:
Should this be closed?  Or will this be worked on?

RodneyBaker:
I regularly use programs that have a frame zero.  Is there a way that imported image sequences could be accepted but just be bumped up one frame so that there is no longer a frame zero?  In that way the internals for everything else would not have to change.  I realize this might be ideal because one frame difference for every frame can throw a sequence out of sync with audio and more.  

I agree that in general there should not be a frame zero so there are several approaches that would be acceptable to me personally.  One might be to simply remove that frame.  Another would be to bump it up one frame.  The initially important thing is that the sequence be accepted as such; to be recognized as a sequence.

At a minimum an image encountered with that trailing zero needs to be caught so that an error message can be thrown to advise the user of the problem.  At present there is no indication of what was encountered and why it failed.

RodneyBaker:
Filenaming issues are still very much a work in progress and some aspects of the problem will no doubt move into the 2019 developments cycle.

The primary workaround for these issues is to rename the sequence into optimal syntax and numbering.  

This PR specifically deals with zeros in the sequence but assistance with identifying specific filenaming issues that need to be addressed will also be appreciated.  Most of the identified issues are titled "Filenaming:" followed by the specific focus of the issue.

RodneyBaker:
Not sure how similar this PR is to #2607 but both purport to resolve the sequencing issue where Opentoonz fails to process sequences starting with a trailing zero.

cc:  @manongjohn

manongjohn:
Something else fixed the issue.  My PR just removes the message that was added because it blocks loading the file.

I would recommend closing this as it is no longer needed.

RodneyBaker:
THanks @manongjohn 
Closing.  Pending any identified reason to reopen.

RodneyBaker:
Reopening this as Trailing zeros on filenames is still a problem.

manongjohn:
Not sure why you are reopening this? The changes proposed here are not needed as far as I can tell since, after removing the malformed message per my PR, files with sequence number 0 do appear to load successfully.

RodneyBaker:
Reopened because I have no knowledge of this being fixed or not.
I only know the underlying issue is still present.
I'll gladly close this once the other PR is merged and there is no longer an issue.

Feel free to close but we likely should have let @jabarrera close the PR last time assuming he's still watching.

RodneyBaker:
Closing this as unnecessary now that the other PR is merged and no signs of trouble experienced.  
See previous discussion for additional information.

PR #971, opened 2016-12-17
Fix crash saving an AIFF level missed

### LINKED ISSUE #957: Missing level causes crash when saving
On Mac OS X 10.11.6 & 10.11.4, a missing level causes a crash every time levels are saved (this includes "Save All" and "Save All Levels"). I do know that with the proper work flow (i.e. removing unused levels  after moving/deleting a level), this would not be an issue but for new learners, it would be helpful to have this resolved. 
screen shot 2016-12-09 at 10 22 13 am
Perhaps when there is a missing level, saving could prompt a warning message asking to remove unused levels before saving.
[crash3.txt](https://github.com/opentoonz/opentoonz/files/642515/crash3.txt)

jabarrera:
Hi. This PR closes #957 issue.

**How to reproduce**:

1) Create a scene
2) Add an AIFF level sound.
3) Save all.
4) Change the name, move or delete the file name of the AIFF sound.
5) If you didn't close the scene before step 4, close now or change to a new scene.
6) Open the scene created in step 1.
7) Save all (or save all levels).
8) You'll get a crash.

**What does the fix**:

It checks that the AIFF data sound is valid, throwing a message to the user (to Message Center) if the save cannot be possible.

**Comment to this fix**:

This probably must be controled in an upper stage (at resource stage), but that should require a deeper refactor. So, as a quick fix, this should works.

### DIFF
--- a/toonz/sources/sound/aiff/tsio_aiff.cpp
+++ b/toonz/sources/sound/aiff/tsio_aiff.cpp
@@ -457,7 +457,9 @@ TSoundTrackWriterAiff::TSoundTrackWriterAiff(const TFilePath &fp)
 //------------------------------------------------------------------------------
 
 bool TSoundTrackWriterAiff::save(const TSoundTrackP &st) {
-  assert(st);
+  if (!st)
+    throw TException(L"Unable to save the soundtrack: " +
+                     m_path.getWideString());
 
   TSoundTrackP sndtrack;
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. @jabarrera Thank you for the fix!

PR #972, opened 2016-12-17
[Enhacement] Add "aif" extension to AIFF sound list

jabarrera:
Hi.

This PR adds the Microsoft old-style (three letters) file extension **_aif_** to load and save AIFF sound files.

As there are a lot of sound files that have the short AIFF extension (and it is an standard extension, too) I think it is good to have listed, too.


turtletooth:
The change is pretty straight forward.  I don't use aif files, but I don't see why this shouldn't be added.  Once jenkins is called on this.  LGTM.

shun-iwasawa:
@jabarrera Thank you for the improvement and @turtletooth thank you for reviewing this.
Jenkins

PR #974, opened 2016-12-17
[Cosmetic] Cleaning for-nothing methods

jabarrera:
Hi.

This push out extra _for-nothing_ methods maden, it seems so, by a programmer in a bad day (or not very confident with his skills).  :)


shun-iwasawa:
Oh, you found an Easter Egg. :) Thank you for pointing this out!  LGTM

PR #975, opened 2016-12-17
[Fix] Crash in History Panel when a level is unlinked.

### LINKED ISSUE #965: Occasional Crash When Accessing History
On OS X 10.11.6, OpenToonz (both 1.1.1 and 1.1.2) occasionally crashes when switching to a History room.

[crash5.txt](https://github.com/opentoonz/opentoonz/files/655625/crash5.txt)

jabarrera:
Hi. This PR closes #965 issue.

**How to reproduce**:

1) Select brush and draw something to get a new level.
2) Open History panel.
3) Open Schematic panel.
4) Toggle to FX schematic.
5) Select the link between the level (drawed in step 1) node and the XSheet node and delete it (press _supr_).
6) You'll get a crash.

**What happens**:

It is used an un-initialized iterator pointer for calling a method. Probably somebody was doing some test in code and forgot get back to a previous stage.

**What does the fix**:

Call the method from the right iterator.

### DIFF
--- a/toonz/sources/toonzlib/fxcommand.cpp
+++ b/toonz/sources/toonzlib/fxcommand.cpp
@@ -2022,12 +2022,11 @@ QString DeleteLinksUndo::getHistoryString() {
 
   if (!m_terminalFxs.empty()) {
     str += QString("  :  ");
-    std::list<TFxP>::const_iterator it1;
     std::list<TFx *>::const_iterator ft;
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
@jabarrera Confirmed crash and your fix. I'll merge this as soon as Jenkins check is done.
Thank you!

shun-iwasawa:
LGTM. @jabarrera Thanks again for the modification.

PR #982, opened 2017-01-08
[Feature] Set auto markers

### LINKED ISSUE #572: feature: X-sheet - Auto Set markers
This is a feature suggestion for the x-sheet right click menu - for a command to autoset start and stop markers based on gaps between cells on the active level.

Here is a mockup gif to explain how it could work:
autosetmarkers

This would be useful for those who like to organize their animation into separate clips/takes and quickly want to cycle between clips that loop.

jabarrera:
Hi.

This adding closes #572 issue request.

**What the new feature adds?**

It adds a new item (_Set Auto Markers_)  in the popup of the frame number column.  When the item is selected (if there is a cell non-empty in the current column and row) the start and finish markers are positioned in the first and last continuous cell.

The current row is where the user right click to get the popup. The current column is the selected column.

Here is a video with simples testings: https://youtu.be/X9ATeUelUao

### DIFF
--- a/toonz/sources/toonz/xshrowviewer.cpp
+++ b/toonz/sources/toonz/xshrowviewer.cpp
@@ -688,6 +688,11 @@ void RowArea::contextMenuEvent(QContextMenuEvent *event) {
   connect(setStartMarker, SIGNAL(triggered()), SLOT(onSetStartMarker()));
   QAction *setStopMarker = menu->addAction(tr("Set Stop Marker"));
   connect(setStopMarker, SIGNAL(triggered()), SLOT(onSetStopMarker()));
+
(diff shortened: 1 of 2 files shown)


turtletooth:
I like it.  Thanks for adding this.

blurymind:
This is ace! Thank you so much for adding it. It will make the workflow
much smoother- especially when dealing with animation for games :D

I hope it gets merged soon

shun-iwasawa:
Jenkins

shun-iwasawa:
It works fine. Thank you @jabarrera ! LGTM

blurymind:
@shun-iwasawa  thank you for reviewing!

PR #983, opened 2017-01-11
Color filtering feature for columns in camstand view

shun-iwasawa:
This feature is to apply color filtering to each column in the viewer with camstand view.

**How to use this feature**
Color filter combo box is added under the transparency slider in a popup which appears when you click and hold the camstand visible button (an orange rectangle) in the column head.

**Main purpose of this feature**
When pencil testing, this feature is intended to distinguish characters by applying different color filters. 
In the following example, every drawings were captured in gray scale:

Associated with this feature, I modified the composition calculation of color-scaled images in the viewer.
Please note that it will also affect appearance of onion skin as well. The change becomes noticeable if you set some neutral color (i.e. not 255-red or 255-green) for the onion skin colors:

### DIFF
--- a/toonz/sources/toonz/xshcolumnviewer.cpp
+++ b/toonz/sources/toonz/xshcolumnviewer.cpp
@@ -58,6 +58,7 @@
 #include <QToolTip>
 #include <QTimer>
 #include <QLabel>
+#include <QComboBox>
 //=============================================================================
 
 namespace {
@@ -101,6 +102,14 @@ bool containsRasterLevel(TColumnSelection *selection) {
     }
   }
   return false;
+}
+
+const QIcon getColorChipIcon(const int id) {
+  static QList<QColor> colors = {Qt::red,        Qt::green,    Qt::blue,
+                                 Qt::darkYellow, Qt::darkCyan, Qt::darkMagenta};
+  QPixmap pixmap(12, 12);
+  pixmap.fill(colors.at(id - 1));
+  return QIcon(pixmap);
 }
 }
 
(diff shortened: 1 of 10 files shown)


shun-iwasawa:
Jenkins

toonz/sources/toonz/xshcolumnviewer.cpp
```diff
@@ -1315,13 +1334,38 @@ m_value->setFixedWidth(30);
...
+    hlayout->addWidget(m_slider);
+    hlayout->addWidget(m_value);
+    hlayout->addWidget(new QLabel("%"));
+    mainLayout->addLayout(hlayout, 0);
+
+    QHBoxLayout *filterColorLay = new QHBoxLayout();
+    filterColorLay->setMargin(0);
+    filterColorLay->setSpacing(2);
+    {
+      filterColorLay->addWidget(new QLabel(tr("Filter:"), this), 0);
```
turtletooth: @shun-iwasawa 
What do you think of this modification.  Adding a tooltip might help folks confused about why it doesn't affect vector layers:

      QLabel *filterLabel = new QLabel(tr("Filter:"), this);
      filterLabel->setToolTip(tr("Filter doesn't affect vector layers"));
      filterColorLay->addWidget(filterLabel, 0);
      m_filterColorCombo->setToolTip(tr("Filter doesn't affect vector layers"));
      filterColorLay->addWidget(m_filterColorCombo, 1);

shun-iwasawa: Thank you very much for catching this. I forgot to give thought to the case of vector images.
I'll modify as you suggested and will cover vector levels in the future, as another PR.

shun-iwasawa:
Tooltip added and rebased. Jenkins

turtletooth:
LGTM.  Thanks

Issue #894, opened 2016-10-31
Feature Request: New Level Buttons in XSheet Corner


turtletooth:
I think one of the great features of the software is to have a variety of level types.  Currently it is a handful of clicks to create a new level and change the type to a format other than the default format.  Buttons in the xsheet could speed up the workflow.  It currently looks like this:
![new_level](https://cloud.githubusercontent.com/assets/4576381/19864706/061780b2-9f5f-11e6-9f44-462c418f93c9.PNG)

What do you think about something like this:
![add_new_level](https://cloud.githubusercontent.com/assets/4576381/19864714/155d8652-9f5f-11e6-8c4c-131c418d81a7.png)

One click for the various level types.

artisteacher:
I think it is a great idea! Would clicking one of the buttons pull up the new level menu or would it create a level automatically?

turtletooth:
@artisteacher I think it would be better to bring up the new level popup so you can name your levels.  Many of my students get lazy and use levels named a, b, c . . and then get confused when their scenes get complex.

pculli:
I was thinking of a similar idea lately.

Maybe you could set it to automatic or "name" mode.
In automatic, it would create a level and name it. In "name" mode, it could bring up a name popup only, just to give it a name and apply it by pressing Enter. That would be pretty fast and useful.

shun-iwasawa:
In the traditional production workflow, most of levels are scanned from papers and the "New Level" command is rarely used. I understand that these buttons are somewhat convenient for paperless workflow or in educational use, but IMO it should be optional ( as an item in Preferences ) in order to keep the interface simple.

2dvision:
Is this possible add a toolbar on top xsheet window, like toon boom? user can add/remove buttons and show/hide toolbar.

![har11_custom_marks](https://cloud.githubusercontent.com/assets/16509592/19932546/c23e4fba-a125-11e6-804e-dc811496d5a4.png)

similar feature requests 325

artisteacher:
Name mode would be amazing!

turtletooth:
I'm combining this with #906.  Any ideas on what an icon would look like for differentiating new level types?  Also any other thoughts on buttons you would like in the toolbar?

artisteacher:
These would need plenty of adjustment but could be a direction to go in. Not sure what to do about toonz raster. I do think that a traditional layer icon is a good base. 
[...]

PR #989, opened 2017-01-15
XSheet Toolbar

turtletooth:
This is for #894 

This adds optional buttons in the corner for creating new levels:
buttons

This can be enabled in preferences in the drawing section:
preferences_buttons

Note: This option is off by default.


yamisama:
Thank you, that looks like quite a useful addition.

However, I don't think the UI is very intuitive in the current form. How about instead of squeezing these three buttons into the corner of the XSheet, we add a toolbar on top of the XSheet window that can be shown and hidden to taste? 

That way, it would be consistent with conventions that users are familiar with other applications as well as OpenToonz' own viewer window. Other common XSheet commands could also be placed there for quick and easy access. This would also eliminate the need for the quirky setting in preferences – most of the UI customization settings currently live in the main UI (mostly in the menus and popup menus). Moreover, this design would allow for using icons that are easy to distinguish instead of non-intuitive one-letter text labels. 

And lastly, a toolbar would allow users using pen tablets to quickly reach all relevant commands without any right-clicks, which are always a bit clumsy to perform with pens, or keyboard shortcuts, which are inconvenient if the user has to activate an on-screen software keyboard first.

Users concerned about screen space could simply hide the toolbar to be back to the current state of OpenToonz.

turtletooth:
@yamisama I understand you concerns.  I placed it where it is to preserve screen space, but I understand that adding a toolbar at the top is more conventional and allows for more tools.  

I will take a look at moving it to the top.  You mentioned icons instead of text.  I have no idea what an icon looks like that differentiates between vector, toonz raster and raster.  The current text shows a tooltip if you hover over it to let you know what the button does.  

As far as screen space goes, I want to try and make it around the height of one row in the xsheet.

turtletooth:
Added an XSheet Toolbar.  
xsheettoolbar


This is available through right click in the xsheet header and in the preferences.
preferences_xsheet

The other option in preferences allows the height of the function editor header to also be extended so that the xsheet toolbar can be used and the xsheet and function editor will line up if side by side.

Note:  I tried to squash my commits, but had errors squashing all of them.

2dvision:
Awesome! Is it possible add "Toolbar Manager window" for organize toolbar like ToonBoom? ([Toolbar Manager section](http://docs.toonboom.com/help/animate/Content/HAR/Stage/003_Interface/069_H1_Managing_the_Toolbars.html))

turtletooth:
> Is it possible add "Toolbar Manager window" for organize toolbar like ToonBoom? 

Right now, that functionality is beyond the scope of what I have time to bite off right now. Are there any other buttons that you would like to see added?

2dvision:
I personally use a lot these too:

- Load level...
- Replace Level...
- Autorenumber
- Reverse
- Open Sub-xsheet
- Close Sub-xsheet

turtletooth:
Here is the current set:
xshtoolbar2-9

turtletooth:
Changed to WIP- will replace other icons tomorrow.

turtletooth:
Here it is with the new icons.  
xshtoolbar

Should we have a slightly darker shade in the icons?  It seems a touch too bright.

adamearle:
yeah darker would be nicer i think


adam

konero:
If it helps, Toonz icon colors are `#e6e6e6` for the base and `#404040` for the outline.

shun-iwasawa:
Jenkins

toonz/sources/toonz/xshtoolbar.cpp
```diff
@@ -0,0 +1,265 @@
...
+//-----------------------------------------------------------------------------
+
+void Toolbar::onRepeatButtonPressed() {
+  CommandManager::instance()->execute("MI_Dup");
+}
+
+//-----------------------------------------------------------------------------
+
+void Toolbar::onCollapseSubButtonPressed() {
+  CommandManager::instance()->execute("MI_Collapse");
```
shun-iwasawa: In most of the slots you are just invoking a correspondent action from `CommandManager` .
How about to make the toolbar buttons by `QToolBar->addAction(QAction*)` instead of `addWidget` for such case?

You can set icon and tooltip for these actions in `mainwindow.cpp` (Please refer my modification [here](https://github.com/opentoonz/opentoonz/pull/1060/files#diff-c4e8720ba98e1879d0179b7851697c42).)
Buttons can be styled like as follows: 
```
#XSheetToolbar {
	margin: 0;
	padding: 0;
	border: 0;
		&::separator:horizontal {
		margin-right: 7;
		margin-left: 5;
	}
	& QToolButton {
		.base_outset;
		border-width: 1px;
		border-radius: 4px;
		margin: 2;
		padding: 3;
		&:pressed {
			border-image: none;
			.base_inset(light, 10%);
		}
		/*lighten lilttle bit when hover*/
		&:hover {
			border-image: none;
			.base_outset(light, 10%);
			&:pressed {
				.base_inset(light, 10%);
			}
		}
		/*lighten lilttle bit when pressed*/
		&:disabled{
			.base_outset(light, 5%);
			color: rgb(80,80,80);
		}
	}
}
```
IMO It will improve UI consistency in two points: First, the same icons will be added to the main menu bar as well. Secondly, the actions' state (enable/disable) will be reflected to the tool bar buttons.

turtletooth:
@shun-iwasawa I got it updated and working the way you suggested and figured out @konero themes I think.  It's a pretty slick way to do it.

shun-iwasawa:
Jenkins

toonz/sources/toonz/xsheetviewer.cpp
```diff
@@ -911,6 +940,7 @@ void XsheetViewer::onSceneSwitched() {
 //-----------------------------------------------------------------------------
 
 void XsheetViewer::onXsheetChanged() {
+  updatePanelsSizes();
```
shun-iwasawa: For now you are notifying the toggle of Xsheet toolbar with the `TXsheetHandle::xsheetChanged` signal.
I think it would be better to use `TSceneHandle::preferenceChanged` signal instead, since the `xsheetChanged` will be emitted so often throughout common operations on the xsheet.
Please see my implementations for the "blank frame" setting in [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/preferencespopup.cpp#L372-L376) and [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonzqt/flipconsole.cpp#L1855-L1932).

turtletooth: Good idea.   I didn't know that signal existed.

turtletooth:
Changes made.

turtletooth:
@shun-iwasawa Once this is merged, there are going to be conflicts in #1244 - I'll resolve those as soon as this is merged.  #1246 is a dead simple UI fix that probably doesn't even need a Jenkins on it.  It resolves an issue in the new themes that @konero pointed out.

shun-iwasawa:
LGTM. Thank you @turtletooth .

shun-iwasawa:
I'll merge it after Jenkins, just in case.

Agnyy:
@turtletooth  I think so the icons will be more intuitive.

Agnyy:
Has a little finished. To a small size it was better to see.

2dvision:
@turtletooth I have a suggestion in #1291 for add "XSheet Toolbar" to Windows menu.

artisteacher:
This works beautifully and is a really helpful addition. At some point it would be great to have a load level button added - for importing audio, .psds or etc.

RodneyBaker:
I hesitate to ask this question because the Xsheet Toolbar has been merged and is working great.

I am wondering if a hint of color could be added to the icons used to create Toonz Vector, Toonz Raster and Raster levels.  That color hint would match/echo that of the color of columns in the xsheet (Yellow/Brown (Redish?) for Toonz Vector, Green for Toonz Raster and Blue for Raster).  

I ask for several reasons but mostly because while having color for no reason isn't very helpful, color that is specifically designed with a purpose can assist users as they navigate through OT.

While I like the current icons alot they are small enough that I have mainly just memorized which is which and therefore am fairly sure what Level type I will get when I click on an icon.  A subtle hint of color that matches the current color coding of Levels that already exists in the xsheet will reinforce good habits; the user clicks the (slightly) bluish icon and it creates a (slightly) bluish Level which with time they know to be a raster level.  So, an artist with a desire to use MyPaint brushes will learn to seek the bluish icon to create the correct Level.

Hope this makes sense and isn't too far afield from the ongoing effort to lock down the Xsheet Toolbar.

The color need not be dramatic or overpowering.
Assuming it can be perceived it might simply be a color at the bottom of the icon (see attachment).


new level icons w splash of color
.

artisteacher:
The tricky thing is that the different themes don't all have the same colors set. So instead of adding the color directly in the image, it should really be added to the css.

adamearle:
why does this have to be a folded bit of paper?

RodneyBaker:
@artisteacher said:

> The tricky thing is that the different themes don't all have the same colors set. So instead of adding the color directly in the image, it should really be added to the css.

Agreed.  I looked in the css but didn't find anything that suggested that such customization is exposed for editing.  It might be a trivial thing to add a thin colored line at the bottom of the icons via CSS (or border... or tint to the entire icon, etc. ).

@adamearle said:

> why does this have to be a folded bit of paper?
Excellent question.  I will guess that the folded corner represents 'new' so that the user knows even before they click the icon they will be creating something new.  (I'm currently having some issues with icons showing in the OTX release so can't confirm)

artisteacher:
The folded paper is a standard icon for making new layers in graphics programs. While levels aren't quite the same as layers in OT, they are pretty close. For new users familiar with Photoshop, Flash, Affinity & etc., the folded paper icon will be familiar and easily understood. That said, it wouldn't hurt to simplify the icons further. I love @Agnyy's proposed icons - especially the vector one,
 but am concerned they are too detailed for such a small scale.

The grey areas on the following icons are completely transparent, so they should recolor through the css easily. This returns to including the letter for each level type.
screen shot 2017-08-06 at 1 43 43 pm


[svg files](https://github.com/opentoonz/opentoonz/files/1203285/maskicons.zip)

yamisama:
I think the "+" icon popularized by Apple over the last few years could be used instead. I think it's a bit more logical and obvious for novices than the Microsoft-style blank sheet of paper, especially since a Level in Toonz usually consists of multiple frames and therefore multiple sheets of cel or paper in traditional animation. Moreover, the piece of paper iconography is already used in Toonz to mean "Memo" and could thus be slightly confusing.

Therefore, if we were to redesign these icons, my suggestions would be to use a completely different icon for each level type with a "+" in the top right corner. For example:

Rectangular stylized landscape "picture" icon with a "+" in the top right corner to indicate "New Raster Level", as raster levels are mostly used for painted backgrounds.
Some sort of extremely blocky pixelated filled shape like a circle with a "+" in the top right corner for "New Toonz Raster Level", as these levels are usually used for filled foreground objects.
The same shape in sharp with Bézier handles or anchor points and a "+" in the top right corner to indicate "New Vector Level"
Speech bubble with that same "+" for the "Notes" level type
The lack of piece-of-paper frame would maximize the space that's left for the element that's actually differentiating the icons, and the "+" would link them visually and make it clear that they all add something to the XSheet.

PR #990, opened 2017-01-16
Remove can't save warning on untitled autosave

### LINKED ISSUE #777: Workflow- Autosave on bring up save scene dialog if new scene
Right now if a scene has not been saved at all and autosave is on, it brings up a dialog that says it is not possible to save an untitled scene.  (tapp.cpp line 676-580)

Why not just bring up the save scene box?  For the current box, you have to click ok and to cancel the save scene box is also one click.  This produces no more clicks, but also gives the option of saving right away.

Removing the lines mentioned makes it work.  #775 needs to be merged first before this change.

turtletooth:
This fixes #777.  This removes the can't save warning if a scene is untitled during autosave and just brings up the save dialog.

### DIFF
--- a/toonz/sources/toonz/tapp.cpp
+++ b/toonz/sources/toonz/tapp.cpp
@@ -673,13 +673,6 @@ void TApp::autosave() {
   } else
     m_autosaveSuspended = false;
 
-  if (scene->isUntitled() &&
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
It works fine. Thank you @turtletooth ! LGTM

Issue #695, opened 2016-08-01
Scrub timeline in view


Rayek:
In Maya holding down k+middlemouse and dragging in the viewport to scrub is really nice, and I would like to add this as a request.

It can be taxing on the arm, and an additional option to use the a shortcut key (k?) + scroll-wheel ought to be implemented as well.

Similar to Mscrub for 3dMax:

mScrub allows you to easily scrub the timeline and navigate thru keyframes with the mousewheel.

Only for 3dsmax 2013+
Press and hold SPACEBAR

Wheel – Moves back and forth 1 frame

CTRL + Wheel – Moves 10 frames forward or backwards

SHIFT + Wheel – Moves to next or previous keyframe.

Release SPACEBAR and you’re done.

turtletooth:
#994

PR #994, opened 2017-01-16
Added new scrubbing options

turtletooth:
This is for #695 
This adds the ability to hold shift + middle mouse to scrub through the timeline
and adds shift + scroll wheel to go forward and backward as well, ctrl + scroll to go prev/next drawing and ctrl + shift + scroll to go prev/next step (default is 10).

I didn't add next keyframe since step is a little more global to every animation.  Not every scene might use keyframes.


shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. @turtletooth thank you for the nice feature.

Issue #997, opened 2017-01-18
Deleted scenes show in startup pop-up


artisteacher:
If I delete a scene (either in the browser room or directly through the stuff/projects folder) and re-open OpenToonz, the list of recent scenes in the startup pop-up still displays the names of any deleted scenes. Selecting that deleted scene opens the default, unsaved and empty scene. Then the next time I access the startup pop-up, the deleted scene's name is at the top of the list of scenes. I'm not sure if this is an issue in other OS but I've noticed this issue with OS X 10.9.5, 10.11.6, and 10.12.2.

turtletooth:
It's going to be an issue on all OSes - I'll see if I can make a check for the scene existing before the list populates.

ideasman42:
Would be careful attempting to fix this. IIRC Blender had this reported too & resolved too,
then people complain that the application starts slow.

Turns out if you have a bunch of files on a network drive - it can slow down startup considerably just to check they exist.

turtletooth:
Any suggestions?  Would you leave it and let people realize - oh yeah, I deleted that?  Or something else?

turtletooth:
I think I'm only going to check for the file once it has been clicked and if it doesn't exist, give a popup warning and refresh the list then.  What do you think?

ideasman42:
Yeah, some way to handle the case where the file doesnt exist is nice, "This file is missing, remove from recent file list..." or similar.

turtletooth:
Fixed in #998

turtletooth:
Thanks @ideasman42 for the advice.

artisteacher:
Thank you!

ideasman42:
Any reason to keep the report open?

PR #998, opened 2017-01-18
Fixed removed scene from startup

turtletooth:
Fix for #997 

Before attempting to open a recent scene it checks to see if it exists.  If it doesn't it gives a warning and the popup stays open with the scene list updated.

Note:  didn't check to see if the scenes in the list exist when first populating based on the warning from @ideasman42 in #997

### DIFF
--- a/toonz/sources/toonz/startuppopup.cpp
+++ b/toonz/sources/toonz/startuppopup.cpp
@@ -52,6 +52,20 @@ namespace {
 // the first value in the preset list
 const QString custom = QObject::tr("<custom>");
 
+void removeAll(QLayout *layout) {
+  QLayoutItem *child;
+  while (layout->count() != 0) {
+    child = layout->takeAt(0);
+    if (child->layout() != 0) {
+      removeAll(child->layout());
+    } else if (child->widget() != 0) {
+      delete child->widget();
+    }
+
+    delete child;
+  }
+}
+
 QString removeZeros(QString srcStr) {
   if (!srcStr.contains('.')) return srcStr;
 
@@ -346,16 +360,20 @@ void StartupPopup::showEvent(QShowEvent *) {
 
   // update recent scenes
   // clear items if they exist first
+  refreshRecentScenes();
+  // center window
+  this->move(QApplication::desktop()->screen()->rect().center() -
+             this->rect().center());
+}
 
-  if (m_recentSceneLay->count() > 0) {
-    QLayoutItem *child;
-    while (m_recentSceneLay->count() != 0) {
-      child = m_recentSceneLay->takeAt(0);
-      delete child;
-    }
-  }
+//-----------------------------------------------------------------------------
 
+void StartupPopup::refreshRecentScenes() {
+  removeAll(m_recentSceneLay);
+
+  m_sceneNames.clear();
   m_sceneNames = RecentFiles::instance()->getFilesNameList(RecentFiles::Scene);
+  m_recentNamesLabels.clear();
   m_recentNamesLabels = QVector<StartupLabel *>(m_sceneNames.count());
 
   if (m_sceneNames.count() <= 0) {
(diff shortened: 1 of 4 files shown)


shun-iwasawa:
Jenkins

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -346,16 +360,21 @@ void StartupPopup::showEvent(QShowEvent *) {
 
   // update recent scenes
   // clear items if they exist first
+  refreshRecentScenes();
+  m_recentSceneLay->addStretch(100);
```
shun-iwasawa: Adding the stretch should be moved at the end of `refreshRecentScenes()` .
If you just would like to put the items aligned at the top, it is enough to add stretch with the stretch value 1 and set the stretch value of the "scene name" items to 0 like as follows:

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -372,18 +391,13 @@ void StartupPopup::showEvent(QShowEvent *) {
       m_recentSceneLay->addWidget(m_recentNamesLabels[i], i, Qt::AlignTop);
```
shun-iwasawa: The second argument (stretch value) should be `0` instead of `i` .

turtletooth:
Thanks @shun-iwasawa I think I fixed the issues.

toonz/sources/toonz/startuppopup.cpp
turtletooth: Done.  Thanks

shun-iwasawa:
Jenkins

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -369,21 +387,17 @@ void StartupPopup::showEvent(QShowEvent *) {
...
           name.remove(0, name.indexOf(" ") + 1));  // remove "#. " prefix
-      m_recentSceneLay->addWidget(m_recentNamesLabels[i], i, Qt::AlignTop);
+      m_recentSceneLay->addWidget(m_recentNamesLabels[i], 0, Qt::AlignTop);
       i++;
     }
-    m_recentSceneLay->addStretch(100);
   }
-
   bool ret = true;
   for (int i = 0; i < m_recentNamesLabels.count() && i < 7; i++) {
```
shun-iwasawa: Sorry for the late notice.
For now, 8th, 9th and 10th item in the list cannot be opened on clicked since the connection is limited to 1-7th items here.

turtletooth: Good catch.  I'm guessing I put that there before I actually knew how many recent scenes would fit there.

turtletooth: Fixed.

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -804,10 +818,21 @@ void StartupPopup::onRecentSceneClicked(int index) {
...
   QString path =
       RecentFiles::instance()->getFilePath(index, RecentFiles::Scene);
-  IoCmd::loadScene(TFilePath(path.toStdWString()), false);
-  RecentFiles::instance()->moveFilePath(index, 0, RecentFiles::Scene);
-  RecentFiles::instance()->refreshRecentFilesMenu(RecentFiles::Scene);
-  hide();
+  if (!TSystem::doesExistFileOrLevel(TFilePath(path.toStdWString()))) {
+    RecentFiles::instance()->removeFilePath(index, RecentFiles::Scene);
+    RecentFiles::instance()->refreshRecentFilesMenu(RecentFiles::Scene);
+    for (int i = 0; i < m_recentNamesLabels.count() && i < 7; i++) {
```
shun-iwasawa: Oh, please don't forget to fix here as well! 
IMO it would better to define some const value instead of using magic number of 10.

turtletooth: Let's try this.

shun-iwasawa:
Jenkins again, just in case.

shun-iwasawa:
LGTM

Issue #996, opened 2017-01-17
Utility additions for the pencil test feature


shun-iwasawa:
This request was asked by some Japanese animation studio. They are using OpenToonz for pencil testing.
Since they need to increase efficiency, they would like to have options which allow them to work on xsheet without touching mouse. Moving right hand back and forth between mouse and numpad is troublesome for them.

So, the following two options would be nice to have. Both of them should be optional in Preferences.
- [x] Typing arrow key on Xsheet shifts cell selection. (For now, typing up and down arrow keys moves current frame but clears cell selection.) Also, shift + arrow key increases/decreases cell selection range. 
- [x] Enable to input number of drawing into selected cell without double click.

(UPDATE: implemented in #999)

They usually share one PC by multiple animators. So they demands following feature (again, as an option in Preferences), in order to prevent overwriting by mistake captured images taken by another user:

- [x] On launching Camera Capture window, open a dialog to specify the folder location for saving captured images.  

Actually, these features are modeled after QuickChecker (now bundled in Retas Studio) by Celsys. Staff in the studio is familiar with QuickChecker and would like OpenToonz to enhance consistency with it.

jabarrera:
About the first, what do you think about use **ctrl** + up/down to the shifts?

With that map key, the user can work only with the right hand and the actual behavior of the arrow keys can be maintained. In fact, I think that the two options could alive together.


I like the second. It will be really useful. Now is a bit "annoying" the double-click (and the unselection of the text) to introduce just the number. It will be interesting see how to solve this. :)

turtletooth:
> Enable to input number of drawing into selected cell without double click.
What are your thoughts on adding this to the function editor also?

shun-iwasawa:
@jabarrera Thank you for the comments.
The main objective for this option is to make the operations of pencil test as simple as possible. Especially, the pencil test work are done by animators working on papers, and they tend to be unwilling to do complicated operations. 
Your suggestion to use `Ctrl` for the shift does have valid point, but I think it would better to make it with no modifier for the above reason. And if users find that this feature is useful, it would be nice to enable it in any case,  but only with some modifier pressed if the preference option is OFF. (Maybe in the future, as a separate PR.)

BTW I found that `Shift+Up` and `Shift+Down` keys are already reserved for ”skip holding frame”, so I will use `Ctrl` modifier for the increase/decrease of cell selection range.

@turtletooth Thank you for your suggestion. It makes sense from the perspective of consistency.

[...]

PR #999, opened 2017-01-19
Typing arrow key shifts cell selection & Input cells without double click

shun-iwasawa:
This PR will add two options mentioned in the former part of the issue #996.

- Both are added to Preferences, in Xsheet category.

- With the option "Use Arrow Key to Shift Cell Selection" being activated, typing arrow key on Xsheet will shift cell selection. Also, **ctrl** + arrow key will increase/decrease cell selection range.

- With the option "Enable to Input Cells without Double Clicking" being activated, the cell input field will always appear when cells are selected.

In association with this feature, I changed the specification of the cell input field as follows:

- Appearance of the field is changed. Font family and position are modified to be the same as texts in other cells. Background color becomes semi-transparent.
input_cell_ui_change

- Even when the field is shown, the shortcut keys of the commands for cell selection are still available. Conversely, you cannot type level name with key assigned to such commands. For example, if you assign `C` (not `Ctrl+C`) for the "Copy" command, you cannot type `C` as level name in the field since the Copy command will be triggered. 

- When you type number with selecting more than one cell width, you can edit only drawing number with keeping level name in each column. 

- Now, moving the field with arrow keys will only shift the field and will NOT actually input the value. (Typing Enter key is needed to confirm the input.) This change is open to discussing.

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -340,6 +340,16 @@ public:
     return m_showKeyframesOnXsheetCellArea;
   }
 
+  void enableUseArrowKeyToShiftCellSelection(bool on);
+  bool isUseArrowKeyToShiftCellSelectionEnabled() const {
+    return m_useArrowKeyToShiftCellSelection;
+  }
+
+  void enableInputCellsWithoutDoubleClicking(bool on);
+  bool isInputCellsWithoutDoubleClickingEnabled() const {
+    return m_inputCellsWithoutDoubleClickingEnabled;
+  }
+
   // Animation  tab
 
   void setKeyframeType(int s);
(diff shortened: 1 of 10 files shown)


shun-iwasawa:
Jenkins

turtletooth:
I tested this tonight and it does seem like a nice quick way to build an xsheet column.  I'm not totally familiar with the workflow of the studios looking for this feature, but the commands seem to be working the way you indicated.

I am glad that you separated the function editor sheet from this- I could see wanting it in one but not the other.

@shun-iwasawa Are you ready for this to be merged?

shun-iwasawa:
@turtletooth Thank you very much for trying this! Yes, I think it is ready to be merged.

turtletooth:
LGTM.

PR #1004, opened 2017-01-26
Add lzo to cmake include path suffixes.

boombatower:
Building against openSUSE `lzo-devel` does not work without `lzo` path suffix in cmake.

### DIFF
--- a/toonz/cmake/FindLZO.cmake
+++ b/toonz/cmake/FindLZO.cmake
@@ -7,6 +7,7 @@ find_path(
     PATH_SUFFIXES
         lzo/2.09/include/lzo
         lzo/2.03/include/lzo
+        lzo
 )
 
 find_library(


boombatower:
```
/usr/include/lzo
```

ideasman42:
This doesn't prevent thirdparty headers being found, it even seems likely they will be preferred since they come first in the list, submitted #1021 which resolves this.

boombatower:
Yeah, I should have been more clear. I wanted to build it against system headers so I drop all of third party except the driver and tiff. This makes sense to do and was required for legal review to get [opentoonz](https://build.opensuse.org/package/show/openSUSE:Factory/opentoonz) accepted into openSUSE Factory (main distro repo).

```bash
# Remove all thirdparty except tiff which is patched.
find thirdparty/* -maxdepth 0 ! -name "tiff-*" ! -name "lzo" -type d -exec rm -r "{}" \;
# Keep thirdparty/lzo/driver, but remove library.
rm -r thirdparty/lzo/2.*
```

ideasman42:
@boombatower, dont think its reasonable to have developers remove parts of the git repo, though I can see why you'd to it if your only building (not making changes or committing).

Could you test #1021 and confirm if this works for you?

boombatower:
@ideasman42 it has the exact same goal as your PR to build against system versions. It's is rather none standard to include all dependencies in the source repo since they are not actually part of the project being compiled. I'll take a look at the PR.

ideasman42:
@boombatower, its not typical for *nix only packages,
for cross platform software its not so unusual (Audacity, Blender, BRL-CAD do this).

Irrespective of whats common, Linux developers need to be able to work on the code-base without removing parts of it, which can cause problems when committing.

boombatower:
@ideasman42 The OSS projects I have come across with deps in main repo is typically doneso for lack of using either a language or OS based package manager rather than an actual need which seems to be the same case here. It's a choice for sure, but your PR is essentially "ignoring" the code that I "removed" to create the same effect without having to patch the build system to add a switch for all the libraries. Having the switch definitely seems preferred if the libs are to stay in the repo, but is clearly more work hence why I avoided it.

ideasman42:
@boombatower right - moving third-party into its own repo is preferable.<br>OTOH - Linux isn't (yet) a primary development platform for open-toons *(from what I can tell)*. Removing directories is fine for one-off builds, not development. It's important to co-exist with Win32/OSX.

boombatower:
Yeah, and as such your PR with a flag to switch what it is compiled against definitely makes sense.

ideasman42:
getting off topic now, but these paths are no longer included in Blender :)

```
'extern/libredcode'
'extern/ode'
'extern/x264'
'extern/xvidcore'
'extern/qhull'
'extern/make'
'extern/verse'
```

PR #1008, opened 2017-01-27
Subfolder Popup for Camera Capture

shun-iwasawa:
This PR is for the last item in the issue #996 .
In contrast with other workflows, a machine for pencil testing is usually shared by multiple animators.
Since animators are focused on drawing their animation better & faster, some of them may not care of the location to save captured images and overwrite by mistake.
This PR introduces a popup to create and to specify the subfolder for captured images, in order to prevent such tragedy.

- I took "Save In" field out from the right area and put it on top of the Camera Capture window, in order to obtain wider space for file path.
- Added "Subfolder" button on the right of the field, which will open the popup.
- Usability of the popup references QuickChecker by Celsys in order to gain consistency.
- Checking "Show This on Launch of the Camera Capture" checkbox will open the popup every time you trigger the "Camera Capture" command. I decided to save it to user's env file rather than preferences.
- Japanese translation updated.


shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa Thanks for doing this.  Does the folder default to the same folder as usual?  Is there any checking for a level existing when setting the level name to avoid overwriting?  I haven't had a chance to look at it yet, but those are the questions that popped into my head when I saw this.  

I'm looking forward to trying it out.

shun-iwasawa:
@turtletooth Thank you very much for the comment.
Well, the Camera Capture already had warning dialog on overwriting the image. The problem was that if you forget to create a new folder before capturing (and leave the "Save In" field to the default value), OT suggests to skip existing level name and start capturing from subsequent level name. Which may make confusion for animators because they usually would like to start with level A.
So, this popup is to remember them to create a new subfolder before capturing.

And for now if you set existing folder name and press "OK" button in the popup, the warning message appears saying that you should input different name.

Jenkins again

turtletooth:
Tested.  Seems to work fine. I think in the future we should add a minHeight and minWidth to the Pencil Test Window.

LGTM.  Thanks @shun-iwasawa

PR #1010, opened 2017-01-29
Made rounded join the default

turtletooth:
The rounded brush is the default vector brush.  This makes the join also default to rounded for consistency.


shun-iwasawa:
That makes sense. LGTM

Issue #1014, opened 2017-01-31
Cell input feature problems 


shun-iwasawa:
This issue was reported by some Japanese animation studio, using OpenToonz as the pencil test tool by activating  `Use Arrow Key to Shift Cell Selection` and `Enable to Input Cells without Double Clicking` preference options.
When using such options, there are some problems as follows:
- "Delete cell" command cannot be used when the input field is shown.
- If a single cell is selected, user cannot drag&move such cell since the side bar of the cell is covered with the input field.

turtletooth:
#1015 merged

PR #1015, opened 2017-01-31
Cell Input Feature Fixes

shun-iwasawa:
This PR fixes the issue #1014.

- Changed the default shortcut string for `MI_Clear` command from "Delete" to "Del", in order to keep consistency with `QKeySequence::toString(SequenceFormat format)`  which is used with the default argument value of  _format_ = PortableText and the "Delete" is handled as "Del" in many parts of the sources. By this modification, `CommandManager::getActionFromShortcut()` can properly find the `MI_Clear` command from "Del" shortcut key.

- The rename cell field is decreased in width so that the field may not cover over the side bar of the cell.

- Enabled `Undo` and `Redo` commands even when the field is shown.

- Now `Cut` and `Delete` commands will not clear cell selection, if the `Use Arrow Key to Shift Cell Selection` option is ON.

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -1591,7 +1591,12 @@ void TCellSelection::deleteCells() {
       new DeleteCellsUndo(new TCellSelection(m_range), data);
 
   deleteCellsWithoutUndo(r0, c0, r1, c1);
-  selectNone();
+  // emit selectionChanged() signal so that the rename field will update
+  // accordingly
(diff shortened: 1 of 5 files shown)


shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa Thanks for the recent PRs.  I will try and take a look at them tomorrow.

turtletooth:
LGTM - Thanks for the fix. 

 This could be done later, but what do you think of making the box that highlights the cell 1px wider both left and right.  With my PR that removed the two lines between each column it changed how the selected frame looks.  If it were wider, the move cell handle wouldn't have a line going through it.

cell

PR #1019, opened 2017-02-01
Horizontal timeline option in Xsheet (issue#503)

jaros3:
# Overview

This feature adds a "toggle orientation" button to the Xsheet, letting you switch between default Xsheet layout (with vertical timeline) and compact horizontal timeline:

vertical

horizontal2

I've tested as many features as I could, though I'm no veteran in OpenToonz. I've got the basics of selecting, resizing, moving of columns and cells covered. I could not reproduce several more obscure column types; these will behave correctly in old layout and will need tweaks for the new orientation.

# Implementation

OpenToonz code is strongly tied to vertical timeline; it maps X to column number and Y to row number. The link is pervasive in the code. In order to untie them, I introduced several concepts:

- I recognize that there are Frame Axis and Layer Axis.
- Frame Axis corresponds to Y, and Layer Axis corresponds to X in classic layout. They swap roles in the other layout.
- Which axis goes where is a matter of View presenting the data. The data is stored exactly as it was.
- Code makes excessive use of terms "row" and "column". I could not replace it everywhere; instead, I introduced a rule that "row" should be understood to mean Frame Axis and "column" to mean Layer Axis (always!)
- A cell is identified by the pair of (frame, layer). This is called CellPosition.
- Rather than map x <-> layer and y <-> frame separately, I map pairs (x, y) <-> (frame, layer).
- The Column classes keep their name, and should be understood to mean Layer (synonymous: Level).
- There is a class responsible for mapping CellPosition to screen XY or vice versa. The class is Orientation.
- Orientation is responsible for everything layout related.
- Xsheet remembers its orientation and can switch it.

Orientation operates on slightly higher level abstractions than the old code. Wherever possible, I used QRects to mark rectangular areas with meaningful role.

Old code
```
  int x      = m_viewer->columnToX(col);
  int y      = m_viewer->rowToY(row);
  QRect rect = QRect(x + 1, y + 1, ColumnWidth - 1, RowHeight - 1);
```

New code
```
  QPoint xy      = m_viewer->positionToXY(CellPosition(row, col));
  int x          = xy.x();
  int y          = xy.y();
  QRect cellRect = o->rect(PredefinedRect::CELL).translated(xy);
  QRect rect     = cellRect.adjusted(1, 1, -1, -1);
```
I keep `x` and `y` for compatibility; however, whenever possible, I prefer higher level abstraction of `QPoint xy`. `QRect` arithmetics rewards this approach. Sometimes QLine is more appropriate; and sometimes I fall back to X and Y calculations, termed as Layer Axis and Frame Axis. Orientation helps me translate them to screen XY in the end.

Orientation provides quite a few enumerated QRects, QLines, ints and QPainterPaths to satisfy any needs in marking up the screen. These can be retrieved via methods similar to Orientation::rect ():
```
enum class PredefinedRect {
  CELL,                     //! size of a cell
  DRAG_HANDLE_CORNER,       //! area for dragging a cell
  KEY_ICON,                 //! position of key icon
  CELL_NAME,                //! cell name box
  CELL_NAME_WITH_KEYFRAME,  //! cell name box when keyframe is displayed
  END_EXTENDER,             //! bottom / right extender
  BEGIN_EXTENDER,           //! top / left extender
```
These are almost always relative to the top left corner of corresponding cell. These primitives are used for both displaying the items and responding to mouse events:

Old code:
```
    if (column && column->getSoundColumn()) {
      // sound column
      if (x > 20 && 3 * RowHeight + 5 <= y && y < 3 * RowHeight + 33)
        m_tooltip = tr("Click to play the soundtrack back");
      else if (x >= 10 && x <= 20 && RowHeight + RowHeight / 2 < y &&
               y < 8 * RowHeight - RowHeight / 2)
        m_tooltip = tr("Set the volume of the soundtrack");
    }
```
New code:
```
    if (column && column->getSoundColumn()) {
      // sound column
      if (o->rect(PredefinedRect::SOUND_ICON).contains(mouseInCell))
        m_tooltip = tr("Click to play the soundtrack back");
      else if (o->rect(PredefinedRect::VOLUME_AREA).contains(mouseInCell))
        m_tooltip = tr("Set the volume of the soundtrack");
    }
```
which, among other things, reduces chances for inconsistencies.


turtletooth:
Thank you so much for this.  I am going to try it out tonight.  This was a wonderful surprise to see come in.  I'm excited to try it.  

Please be patient as it will take a decent amount of testing and checking to confirm.  There will probably be things that need tweaked here and there, but I think this feature will be greatly appreciated by the users.

Thanks again for the work and I look forward to having you around here.

turtletooth:
Thanks again for doing this.  I was able to play with it and have a few thoughts.

I was able to make it crash by just saving a simple scene.

timelinecrash

linecrash

That is probably easily fixed.  

Also, the column header feels wider than it needs to be.  Is there a way to make it resize-able or remove some of the empty space?

What are your thoughts on making the column (frame-axis) width changeable?  Many timelines have a zoom capability to zoom in or out.  At a certain zoom level, some cell elements would need to be removed such as drawing number.  

In timeline view, the onion skin controls take most of the top, leaving a small area for selecting a frame or scrubbing.  Do you have any thoughts on making the target for scrubbing a bit larger?

Do you think it would be possible to move the cell moving handle back to the left of the cell.  I found several times where I went to select a frame or frame range that I actually began moving the cells.  Moving it to the left may not be the best option, but the current moving target seems a little to easy to trigger.

Thanks again for this amazing feature.  I really enjoy being able to see more columns in a condensed view, and a timeline view will feel more comfortable for many users coming from other software.

I'm really excited for this.

jaros3:
Fixed the crash.

The header is indeed wider than needed, it will be used for some changes I have incoming. Sure can reduce it for now.

Making frame axis scalable is certainly a possibility (since we have all geometry gathered in one place now), but will be probably better off as a separate patch on top of this. It's a big one.

The onion controls only occupy middle of the frame header vertically; anything above or below them would select the frame as normal. This mirrors how it worked in old layout.

Having the cell drag handle on the left would be a little odd, as it repeats on every cell in the sequence. It would appear as broken geometry (and suggest that lines are continued vertically):
left handle

I can make the handle slightly thinner — I was worried it would be too hard to target.

jaros3:
What is this error? I have no idea:
```
-- Apple System
PLATFORM:64
CMAKE_PREFIX_PATH:/usr/local/Cellar/qt@5.5/5.5.1/lib/cmake/
CMake Error at CMakeLists.txt:137 (find_package):
  By not providing "FindQt5.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "Qt5", but
  CMake did not find one.
  Could not find a package configuration file provided by "Qt5" with any of
  the following names:
    Qt5Config.cmake
    qt5-config.cmake
  Add the installation prefix of "Qt5" to CMAKE_PREFIX_PATH or set "Qt5_DIR"
  to a directory containing one of the above files.  If "Qt5" provides a
  separate development package or SDK, be sure it has been installed.
```

turtletooth:
I think the mac build system is having issues.  I got the same error on a PR that passed previously.

turtletooth:
I just tested it again.  Thank you for the changes.

I noticed a few other things:
* Closing and reopening doesn't remember the state of the xsheet.  It always starts horizontal for me.  Ideally it should remember the orientation on a per room basis and default to the vertical as that is what most rooms are configured for right now.
* When the mouse is in the xsheet, the arrow keys navigate the frames the way you would expect, but when the mouse is in the viewer, down and up go to the previous and next frames (which works for vertical) but when the xsheet is horizontal I feel like the arrow keys need to do the same thing no matter where the mouse is.
* The onion skin controls are below the start/stop markers which breaks up the scrubbing area.  In vertical xsheet, the onion skin controls are to the left of the scrubbing area and start/stop markers.

@shun-iwasawa Have you had a chance to look at any of this yet? It's pretty exciting, but you've got the most experience with this program and I'd love to get your feedback.

jaros3:
Saving orientation in the room file now.

There can be several XsheetViewers, each switched independently. This makes it hard to tell how arrow keys should work globally.

turtletooth:
> This makes it hard to tell how arrow keys should work globally.

What about adding the left and right keys in sceneviewerevents.cpp around line 905 in addition to the up and down keys.  That way it wouldn't matter which key was hit- they would both work.

Left and right don't seem mapped to anything else right now.

turtletooth:
BTW- You are really fast with your changes.  Thanks for all your work.

jaros3:
Could the error be due to use of Qt5.6, while CMAKE_PREFIX_PATH points to 5.5?

Error in log file:
```
-- Apple System
PLATFORM:64
CMAKE_PREFIX_PATH:/usr/local/Cellar/qt@5.5/5.5.1/lib/cmake/
CMake Error at CMakeLists.txt:137 (find_package):
```

sources\CMakeLists.txt
```
elseif(APPLE)
    message(STATUS "Apple System")

    if(NOT PLATFORM)
        set(PLATFORM 64)
    endif()
    message("PLATFORM:" ${PLATFORM})
    if(PLATFORM EQUAL 64)
        set(QT_PATH "~/Qt/5.6/clang_64/lib" CACHE PATH "Qt instlattion directory")
```

turtletooth:
I made a portable version of this feature for those who want to test it:
https://github.com/turtletooth/OpenToonzPortable/releases/tag/v.1.1.2-timeline

2dvision:
@jaros3 @turtletooth thanks very much!

I have a few suggestions:

- Change color top toggle switch to gray and when click on it:
  - **On/Off All** for Preview Visibility Toggle
  - **Show/Hide All** for Camera Stand Visibility Toggle
  - **Lock/Unlock All** for Lock Toggle
toggle switch

- Frame handle is overlap on frame border and it is difficult to reframe.

frame handle
frame handle

- Zoom In/Out horizontal frame size.

half size frame

- Ability to duplicate (change) vertical size Sound Level, Because it is difficult to see the sound wave.
sound wave

manongjohn:
Wow! I didn't expect this to be worked on any time soon and suddenly here it is! It was truly a wonderful surprise and so far it all looks great.

I've been briefly playing with @turtletooth's portable version (Windows) and so far have noticed a few things:

1) In the timesheet view only, whenever you set a keyframe in a cell, the level name and # (i.e A 2) disappears for everything in that column (er...row).

2) When you initially scroll vertically, the header row shifts upwards.  When you initially scroll horitzontally the left pane shifts left.

3) If for some reason you have 2 Xsheets windows open, 1 in Timeline and 1 in Xsheet, and vertically scroll in either window, the scroll is done on both windows.

This can be confusing because in Timeline, you are scrolling through columns but in Xsheet mode you are scrolling through frames.

Horizontal scrolling is independent between the windows.  Vertical scrolling should be too or maybe vertical scrolling in timeline causes horizontal scrolling in xsheet and vice versa.

Btw, I second @2dvision's suggestion for the Toggle All buttons.  Would be a really nice feature of the timeline view :)

My suggestion for zooming out: compress to the width of the onion-skin marker and replace level name/# with just a dot, like so:

extreme-zoomout

Hovering over a dot would bring up a hint with the actual level name/#.  Double clicking to edit would allow them to see and edit it as well.


Thanks for all the work and effort put in so far!

jaros3:
There's a reason behind positioning extender handle in the upper half: it's the loop toggle icon. Once a keyframe is set, there will be the loop button at the end of keyframe sequence. It would be overlapping with the extender handle. Not sure which one is worse, anyway there's just not enough space for everything.

Both views subscribe to current frame switched event, which scrolls them into position.

I think name disappearing behavior is the same from original vertical timeline code. But maybe I can improve on it by keeping the name where it fits.

turtletooth:
@jaros3 Thanks again for your work.  Is there a way to make the frame handle not trigger the move area?  I agree that we can't overlap the loop icon, just trying to think of ways to make it work.

> My suggestion for zooming out: compress to the width of the onion-skin marker and replace level name/# with just a dot. 

I like that suggestion.  I think that ToonBoom Studio did something like that.

jaros3:
Sure. We'll check for the extender handle first and the mover second.

turtletooth:
I'm glad to have you around.  Have you been using OpenToonz long?

jaros3:
Not really. My customer is OpenToonz enthusiast. He has some plans for improving OpenToonz and making the changes available to the public.

turtletooth:
That's interesting.  Are you able to share any of the other changes he is hoping for?

jaros3:
In Timeline mode, vector layers will be expandable, similar to AfterEffects, listing Contents (the strokes) and Transforms (offset, rotation, scale). The key feature here is controlling keyframes of individual items. Also, I'm currently working on making vector image strokes (either from brush, primitive or text) animatable.

turtletooth:
Wow!  That would be amazing.  Have you started on any of that yet?

jaros3:
Kinda... I'm wrapping my head around the classes representing curves, keyframes and such — this is complex.

turtletooth:
Good luck,  If you need any testing, I'm glad to help.  It sounds amazing what you are taking on.

manongjohn:
@jaros3 Thanks for fixing the keyframe/missing frame number issue.  Looks better.

I found another issue with the vertical and horizontal scroll bars.  When I load an existing scene with a lot of frames/columns, the scroll bars on both Timeline and Xsheet windows do not appear to account for the last frame/column in the scene, only what is visible in the pane. You can't use either scroll bar.  I got the scroll bars to recalculate by moving the onion skin marker past the last visible frame in the window.  This forced the pane to scroll causing the the scroll bars to update.

The released version of OT does not have this problem.

Oh just to clarify the vertical scrolling issue i mentioned earlier...when you scroll down in Xsheet to frame 50, the Timeline scrolls down to Col50.  I would expect it to horizontally scroll to frame 50 in the timeline.  Conversely, I would expect horizontal scrolling in timeline to vertically scroll in Xsheet, keeping the top most visible frame in Xsheet and left most visible frame in Timeline the same.

This also happens with the Function Editor and the Timeline, a more likely combination to have open in 1 room.

In the released version, vertical scrolling is synced on Xsheets and Function Editors, keeping the top most visible frame the same in all windows, but horizontal scrolling is independent.

jaros3:
## Scrolling along frame axis
Got it. XsheetViewers and FunctionViewer formed an FrameScroller pool where each notified each about vertical scrollbar changes. Clearly, this was not working with orientation.

I had to make a refactor: scrolling along the frame axis causes FrameScroller to map the change to (frame, layer) coordinates and send it to all other FrameScrollers. They map the logical coordinates back to screen ones. By the way, I made FrameScroller a QObject and excluded it out of inheritance chain; it's a field now.

This seems working now.

2dvision:
I think it's better that add 3px (or 5px) to top margin levels:
top_margin_level

another problem in timeline window is order Levels, in professional software like adobe animate, toon boom or TVPaint, Upper layers are front and lower layers are back. But this is different:
back_front_level

blurymind:
Back to front verical order should be reversed- so it is consistent with
photoshop

RodneyBaker:
>
> Upper layers are front and lower layers are back.


It might help to think of the flow in terms of water and gravity.
Hierarchy trees in most programs, file formats, etc. tend to follow this
convention*.
As such, Upper is back and Right is forward. (or should be)

I need to crack open Photoshop and give the flow there a relook before I
can comment further on that aspect.


*I believe an exception is that of data processing where 'stacks' are
 'pushed' and 'popped' into place before we can gain access but this
methodology is generally used in the background, hidden from the user.
This is a constraint of 'real world' processes where we need to work from
top to bottom and right to left because its hard to get at that 'real'
pancake or plate at the bottom of a stack or to travel through time
backward**.
In virtual worlds we can do that.


**We don't have much trouble traveling forward in time.  That happens quite
naturally.
As we get older, most of us would like an exception to that.  :)

artisteacher:
While in every graphics programs I've used, the front layer is also the top layer, I think that under the circumstances, it may be best to keep the current order, in order to maintain consistency with the xsheet. Since the rows are numbered, I do think it's logical that the newer rows with the lower numbers overlap the older rows above. 

Reversing the order would call for a very different workflow. If the front, most recent level is up top, you can't right-click or click and draw on the next row to add a new level. The structure of the timeline would need to change so that there would be no empty rows. 

Also, @jaros3 I am really looking forward to testing this out. The timeline looks to be s great addition.

jaros3:
Reversing the order of layers is hard, just because they expand from top to bottom endlessly; this would need them growing from bottom to top, and I doubt I could make scrolling compatible with that.

manongjohn:
Found an issue with how folded columns are displayed in timesheet mode.  Here is a vertical and horizontal comparison.  The circled part isn't flipped to the side 

folding

I really like working with the timeline. I've adapted to it quite quickly. Hope to see this in the next release or even nightly builds so I can play with it more with some of the fixes you've already made. :)

AnimatorGF:
Any news ?

Number0001:
@shun-iwasawa  Have you looked at all? @konero

shun-iwasawa:
@jaros3 First of all, thank you very much for this great improvement to the software!
I will (hopefully) look into this next week. Thank you for your patience.

morevnaproject:
We plan to include this feature into our Linux builds of Open Toonz soon (https://morevnaproject.org/2016/09/26/opentoonz-package-ubuntu-debian-fedora-linux/). ^__^

turtletooth:
@shun-iwasawa Have you seen this?  It is where I think he is trying to take this feature long-term.  Pretty exciting.
https://www.youtube.com/watch?v=ZNJgmsilyt4

Agnyy:
@turtletooth @shun-iwasawa  Sorry but I do not speak English well. The video has a more timeless timeline. We post it after correcting a number of errors. In the future, many new functions will appear. The timeline will resemble the After Effects 
https://pp.userapi.com/c636727/v636727133/3ca23/K9rNaq0EEPw.jpg

morevnaproject:
@Agnyy I would clarify your post a bit more if you don't mind. ^__^ The perspective of future development for Horizontal Timeline feature is the ability to make shape animation, within one keyframe (in the same way as it is done in After Effects, Synfig or Moho).

morevnaproject:
Just to clear things up: @Agnyy is the guy who provides funding and mentoring for the feature developed by @jaros3.

manongjohn:
Thanks @Agnyy! Exciting to see where this timeline view is heading. Looks awesome. How close is that to being at a point to play with?

As for just the basic timeline, I've been using a local build from a clone of jaros3's branch which have whatever fixes have been noted in this thread. Aside from 2 issues I previously mentioned, I haven't come across any other issues with it.

For an initial offering of the basic timeline view, these should be fixed:
         1) Display of folded rows
         2) Scroll bars not usable after loading scene until you cause the timeline to shift somehow.

A zoomed out view and Toggle All buttons (instead of right-click to get that option) would be nice to haves. :)

toonz/sources/toonz/xsheetviewer.cpp
```diff
@@ -694,18 +835,12 @@ void XsheetViewer::paintEvent(QPaintEvent*)
 //-----------------------------------------------------------------------------
 
 void XsheetViewer::resizeEvent(QResizeEvent *event) {
-  int w              = width();
-  int h              = height();
-  int scrollBarWidth = 16;
-  m_noteScrollArea->setGeometry(3, 1, m_x0 - 4, m_y0 - 3);
-  m_cellScrollArea->setGeometry(m_x0, m_y0, w - m_x0, h - m_y0);
-  m_columnScrollArea->setGeometry(m_x0, 1, w - m_x0 - scrollBarWidth, m_y0 - 3);
```
shun-iwasawa: As you can see here, there were 2 pixel spacing between the column area and the cell area in the previous version. IMO it would better to retain such spacing in order to clarify the appearance.

toonz/sources/toonz/xshcolumnviewer.cpp
```diff
@@ -1474,12 +1411,13 @@ void ColumnArea::openTransparencyPopup() {
...
 
-void ColumnArea::startTransparencyPopupTimer(QMouseEvent *e) {
+void ColumnArea::startTransparencyPopupTimer(QMouseEvent *e) {  // AREA
   if (!m_columnTransparencyPopup)
     m_columnTransparencyPopup = new ColumnTransparencyPopup(
         this);  // Qt::ToolTip|Qt::MSWindowsFixedSizeDialogHint);//Qt::MSWindowsFixedSizeDialogHint|Qt::Tool);
 
-  int x = e->pos().x() - m_tabBox.left() - m_viewer->columnToX(m_col);
+  int x = e->pos().x() - m_tabBox.left() -
+          m_viewer->columnToLayerAxis(m_col);  // what is happening here?
```
shun-iwasawa: `ColumnTransparencyPopup` appears when clicking & holding the camstand button (orange button) in the column. It is for controlling opacity of the column and applying a color filter to the raster level column.
For now the popup seems to appears incorrect position in the Timeline mode.

toonz/sources/include/saveloadqsettings.h
```diff
@@ -0,0 +1,18 @@
...
+#ifndef SAVE_LOAD_QSETTINGS_INCLUDED
+#define SAVE_LOAD_QSETTINGS_INCLUDED
+
+#include <QSettings>
+
+class QSettings;
+
+//! An interface that claims: this object wants to save / load something
+//! into / from provided qsettings
+class SaveLoadQSettings {
```
shun-iwasawa: **[THIS IS NOT THE CHANGE REQUEST!]**
This implementation is NICE ! It's better to use this class for some existing parameters instead of using user .env file. Thanks!  :+1:

shun-iwasawa:
Three more points:

- Regarding `XsheetViewer::wheelEvent`: when using the mouse wheels, the Xsheet scrolls down by frame length equals to the marker interval. IMO this distance should always be 1 regardless of the marker distance value if it is in the Timeline mode.

- When double-clicking the cell area, `RenameCellField` appears. Press `Ctrl+arrow` key while displaying the field will resize the cell selection range. For now, such resizing behavior is incorrect in the Timeline mode. Please apply the same modification done in the `XsheetViewer::keyPressEvent` to the 'RenameCellField::keyPressEvent' as well, regarding the response to arrow keys.

- The last one is I think the most critical & important issue: With this PR, the viewer playback becomes too slow to ignore. I tested with a common scene which has 13 columns x 62 frames. I loop-played the scene in the viewer with the camstand mode while showing the xsheet and measured the frame rate. The previous version could played in 21fps, but with this PR the frame rate depressed to 11fps under the same condition.  I still could not yet find what is actually a bottleneck, but it is indispensable to fix before merging this PR.

turtletooth:
@jaros3 @Agnyy Any news on this?

Agnyy:
@turtletooth I do not speak English well. Yaroslav (jaros3) does not want to take part in the project. And I'm looking for a new programmer. If anyone knows Russian then here's a news group on our developments https://vk.com/opentoonzupgrade.

manongjohn:
Ok. So what do we do to get this PR completed? I have already cloned and built jaros3's project. My C++ is rusty and I'm new to git, but I think i can do most of these change request.

Agnyy:
@manongjohn It's very good if you can help. Here is my post there I'll tell you everything Agniyya@gmail.com. Welcome to the team!

Agnyy:
@manongjohn I'll be waiting impatiently when you write to me. Since I will need to send you documents with information.

manongjohn:
@agnyy I just want to get this PR done which will provide just the basic timeline view. The advanced features you want to add can be done in a new PR? I'll send you email shortly regardless.

Agnyy:
1) The first dimension. Zoom in timeline 
timeline18vj5xxe
wog7b
zoom-timeline

Agnyy:
2. Deploy the scale of sound.

manongjohn:
So, I've fixed the 1st 2 issues so far. Dumb question, how do I update the code for this PR? :)

jaros3:
> So, I've fixed the 1st 2 issues so far. Dumb question, how do I update the code for this PR? :)
I've granted you push access to jaros3/opentoonz. You'll need to set up it as a remote.
The procedure is simple:
• Commit the changes locally
• Push them to feature/horiz branch of jaros3/opentoonz
GitHub will pick them up and show in this pull request

toonz/sources/toonz/xshrowviewer.cpp
```diff
@@ -149,16 +158,14 @@ void RowArea::drawRows(QPainter &p, int r0, int r1) {
         str = QString("%1\"").arg(QString::number(koma).rightJustified(2, '0'));
       }
 
-      p.drawText(QRect(width() / 2 - 15, y + 1, width() / 2 + 7, RowHeight - 2),
-                 Qt::AlignRight | Qt::AlignBottom, str);
+      p.drawText(labelRect, align, str);
```
manongjohn: Fixed. Modified settings in orientation.cpp.  In timeline mode, frame label is centered in column and in xsheet mode, margins restored

toonz/sources/toonz/xshcellviewer.cpp
```diff
@@ -543,9 +544,10 @@ void RenameCellField::showInRowCol(int row, int col, bool multiColumnSelected) {
   setStyleSheet(styleSheetStr);
 
   TXshCell cell = xsh->getCell(row, col);
+  QPoint xy     = m_viewer->positionToXY(CellPosition(row, col)) - QPoint(1, 2);
   if (!cell.isEmpty()) {
     setFixedSize(XsheetGUI::ColumnWidth - 5, XsheetGUI::RowHeight + 4);
-    move(QPoint(m_viewer->columnToX(col) + 7, m_viewer->rowToY(row) - 2));
+    move(xy);
```
manongjohn: Fixed.

Also fixed notes in cell area to resize to the same size as a single cell.

toonz/sources/toonz/xshcellviewer.cpp
```diff
@@ -1211,51 +1312,41 @@ void CellArea::drawLevelCell(QPainter &p, int row, int col, bool isReference) {
 
   // paint cell
   p.fillRect(rect, QBrush(cellColor));
-  // paint side bar
-  p.fillRect(QRect(x, y, 7, RowHeight), QBrush(sideColor));
 
-  if (yetToCleanupCell)
+  drawDragHandle(p, xy, sideColor);
+
+  if (yetToCleanupCell)  // ORIENTATION: what's this?
```
manongjohn: Fixed. Modified the top-left of bounding box that makes this effect to move along x or y axis depending on orientation.

toonz/sources/toonz/xshcellviewer.cpp
```diff
@@ -1407,10 +1502,12 @@ void CellArea::drawPaletteCell(QPainter &p, int row, int col,
...
   if (cell.isEmpty() && prevCell.isEmpty()) return;
-  int x      = m_viewer->columnToX(col);
-  int y      = m_viewer->rowToY(row);
-  QRect rect = QRect(x + 1, y + 1, ColumnWidth - 1, RowHeight - 1);
-  if (cell.isEmpty()) {  // vuol dire che la precedente non e' vuota
+  QPoint xy      = m_viewer->positionToXY(CellPosition(row, col));
+  int x          = xy.x();
+  int y          = xy.y();
+  QRect cellRect = o->rect(PredefinedRect::CELL).translated(xy);
+  QRect rect     = cellRect.adjusted(1, 1, -1, -1);
```
manongjohn: Fixed as suggested

toonz/sources/toonz/xshcolumnviewer.cpp
manongjohn: Fixed. Added logic to calculate position for timeline mode.

toonz/sources/toonz/xsheetviewer.cpp
manongjohn: Fixed. Went through and compared the look from the current version and made sure this version looked the same or similar.  NOTE: I think a merge from another PR caused a number of font size changes.

turtletooth:
@manongjohn Thanks for your work on this.  How's it going?

manongjohn:
Found additional bugs with how stuff was being displayed which I have fixed but not yet commited.  I'm working on fixing the scrolling issue between xsheet/timeline and the function editor, the last of the change requests.

Still need to see if I can emulate the performance issue shun-iwasawa noticed.  My basic 13 column 66 frame one must have been too simple to cause the issue.

turtletooth:
Perhaps @shun-iwasawa could share a test scene with you.

shun-iwasawa:
Sorry but I've tested with a scene which is not allowed to go public. I'll try to make another test scene for reproducing the performance issue (hopefully tomorrow, after releasing v1.1.3) .

shun-iwasawa:
@manongjohn I checked with the latest PR and found that the performance issue doesn't appear. Maybe I mistook something in the previous review. I'm sorry to have caused you trouble!

manongjohn:
@shun-iwasawa No problem. Who knows, maybe something I did fixed it. :)

Quick update. I finally fixed the scrolling issue last night. However, while testing, found a few bugs that need to be fixed, one of which causes a crash.

Once I fix them I'll commit the changes.

manongjohn:
The good news: Just commited all the changes to fix the scrolling issues and other bugs I found in testing. Only 2 very minor scrolling issues I have to fix.

The bad news: I ran an animation I was working on and encountered that performance issue @shun-iwasawa described. :( It was set for 24 fps but i was getting anywhere between 17-21. In 1.1.3, it ran fine.

I decided to turn off every xsheet and function editor in every room (just in case) to see if there was an issue due to display/scrolling issues when an xsheet was open. It still ran badly so I don't think it's that.

I'm not sure if this issue is related to changes from this PR or something that got merged into this branch. I will continue testing different scenarios (i.e. 1.1.3 w/ this PR) to see if I can figure out the cause or rule out this PR.

toonz/sources/toonzqt/spreadsheetviewer.cpp
```diff
@@ -21,60 +22,94 @@
...
-  }
+  if (frameScrollers.contains(this)) frameScrollers.removeAll(this);
 }
 
-void FrameScroller::prepareToScroll(int dy) {
-  if (dy == 0) return;
+void FrameScroller::prepareToScrollOthers(const QPoint &offset) {
+  CellPositionRatio ratio = orientation()->xyToPositionRatio(offset);
   for (int i = 0; i < frameScrollers.size(); i++)
-    if (frameScrollers[i] != this) frameScrollers[i]->onPrepareToScroll(dy);
```
manongjohn: Fixed. Since the timeline columns are narrower, coordinates have to be adjusted accordingly on other scrollbars depending on which scroll bar triggered the event.  Had to correct logic to get the other panes to sync properly without causing more scrolling events.

manongjohn:
Ok...some performance tests looping my animation set for 24fps.

After machine reboot becauase my machine was acting poor in general :

1) Using released 1.1.3: 24 fps constant
2) Freshly cloned OT repository and compiled for Release: 23-25fps changing rapidly while looping.
3) Compiled OT w/ timeline: 23-25fps changing rapidly while looping.

I'm surprised by #2. I would have expected it to match #1. #3 is telling me its not this PR. Maybe my dev enviroment is slightly different than the one that makes the official one?

Thoughts? Suggestions? Not sure what to do from here.

I'm building on a windows machine. That's all I have access to. Can someone build and test this on other platforms?

manongjohn:
@shun-iwasawa I've pulled prior AppVeyor builds from this PR trying to see if I can find out if/when something added in this PR caused this odd performance issue.  The 1st build has the performance issue I was seeing.  However, when I went to other open PR builds (older and newer) randomly, I see the same thing in all of them, even recent ones.  I'm not sure why the official 1.1.3 build is different.  For now I will stop looking into this issue.

Aside from that.  All the change requests you have brought up have been addressed plus other issues I found along the way if you want to do another review.

shun-iwasawa:
@manongjohn 
Thank you for the modifications and the tests so far!
I'll check if the performance issue can be reproduced with the current PR.

turtletooth:
@Agnyy @jaros3 After this change is implemented, do you plan on sharing the other changes you made that allow individual stroke animation?

Agnyy:
@turtletooth Many more changes. Here are some of them. 
https://docs.google.com/document/d/1CyVbJ6dlhFC-IjEIcG52RoUt45PQ6dvDvEO7eL5GQcw/edit?usp=sharing

turtletooth:
@Agnyy I saw the video on YouTube showing some of them already implemented.  Is that code available?

Agnyy:
@turtletooth The programmer had to lay out. Here you can look it up. If you do not find it then let me tell the programmer to send the information. https://github.com/jaros3/opentoonz/tree/feature/timeline-func

manongjohn:
@shun-iwasawa 
I finally figured out why I was seeing a performance issue. This is due to my computer.

I forgot that I had to force the official OT releases in the normal installation folder to use GPU in order to run decently on my system.  For testing purposes, I was running right out of the Release/RelWithDebInfo folders after building which wasn't set up to use GPU and therefore ran badly.

After setting the builds in my dev environments to use GPU, I see no issue with performance.

turtletooth:
@manongjohn Thank you for all your work with this.  You are awesome!

I am playing with getting text levels working again in OpenToonz.  (They started out only for Magpie Pro files, but I am working at expanding their functionality.)

Could you load this scene and see how it looks in the horizontal timeline?
[text.zip](https://github.com/opentoonz/opentoonz/files/1132778/text.zip)

It looks like this in vertical:
vista

manongjohn:
@turtletooth  I saw there was a column type of "SoundText" that appeared to be sourced from Magpie files.  I didn't have one to test and I couldn't make one so I gave up on it.  It was the only column type I couldn't check/test.  I had hoped the layout was going to be ok as is. Apparently, it isn't.

After loading your sample file, the column header in both vertical and horizontal layouts need a lot of work.  It seems the layout is still from Toonz Harmony?  I'm guessing the SoundText layout was never updated to be in line with the other columns in OpenToonz.  I also have to fix how it loads into the cell area in horizontal mode.

Now that I have an example of one, I will work on fixing it to match the rest of OpenToonz. :)

turtletooth:
Doh!  I'm sorry for the extra work.  I had just seen a bunch of requests related to text columns and thought I would look into it.

manongjohn:
@turtletooh No problem. Felt rather incomplete no being able to test it all. Now I can. :) Shouldn't take me very long.

2dvision:
Is it possible:
* fill gap by shortcut? Toon boom harmony and Adobe animate have "F5" shortcut for fill gap.
fill gap
* add camera to timeline, like a level? At the moment, for animate camera need only use function editor.
toon boom camera

tGuMedia:
@2dvision i don't think adding timeline exclusive functionality makes alot of sense, these features probably need their own issue.

2dvision:
@tGuMedia Ok, I thought these could be a few optimizations for the timeline, not feature requests.

2dvision:
@artisteacher Thank you. I found a trick for [animate camera in Xsheet](https://www.youtube.com/watch?v=aGmUKoTCkdo) too.

morevnaproject:
This feature is now included into our builds ^__^ - https://morevnaproject.org/2017/07/17/horizontal-timeline-now-available-opentoonz/

manongjohn:
@morevnaproject  I just submitted a commit to this PR.  If you can, you might want to pull the latest as it includes some Timeline SoundText columns and tool tips fixes that needed to be made. I've also added a preference setting to show/hide Column number in Timeline mode as well as Ctrl-Click of buttons in Xsheet mode will do a Toggle All of that button.

@shun-iwasawa I've decided to freeze the code for this PR pending any major bugs found or additional changes you might require of it.  I do not plan on modifying this PR any more beyond that. The latest commit does **not** include the UI changes we've been discussing.  I will do that as a separate PR since I don't know when that will ever be finalized.

Having said that, please review when you have time.  Hopefully it can be merged sometime soon.

Once this is merged my plans are as follows:
1) Sync these changes to the @jaros3's feature/timeline-func branch
    This branch is quite far behind and I will need to merge all my changes into it. It's going to be a pain to do but must be done in order for continued work on it..
2) Reorder timeline columns to be bottom-up, in line with other software
    The feature/timeline-func branch has some code that I think I can use to help reorder the columns which is why I want to merge everything as soon as I can. I'd like to have it reversed by the next official release.
3) Review the feature/timeline-func functionality and see what I can do to further it along.

shun-iwasawa:
Jenkins

2dvision:
What do you think about replace icons with checkbox in timeline?

timeline checkbox

beeheemooth:
Dots

morevnaproject:
@manongjohn Thank you! We will include it into the next build. ^__^

shun-iwasawa:
I confirmed that it works great for now.
Thank you very much @manongjohn and @jaros3 for this big improvement!

I found a tiny problem that the xsheet and the function editor do not scroll synchronously. I'll post the separate PR for it.

LGTM

morevnaproject:
Great! Also, much thanks to @Agnyy for coordination and funding the work of @jaros3!

shun-iwasawa:
Yes, of course! Thank you @Agnyy for your help!

turtletooth:
Thank you all.  @manongjohn Thank you for picking this up and running with it.  You have done a great job!

PR #1021, opened 2017-02-03
CMake: add WITH_SYSTEM_LZO option

ideasman42:
Allows to easily switch between using thirdparty libs, this patch:

- Adds the option `WITH_SYSTEM_LZO`
- Sets `WITH_SYSTEM_LZO` **off** for Apple and Win32, **on** for Linux.
- Defines reusable utility functions for defining initial options based on platforms.
- This fixes an error using thirdparty headers on Linux, see below.

---

There is a problem currently where building on Linux will default to using the `thirdparty` headers with the system libraries.

This is dangerous since there may be differences (the same issue caused crashes with SuperLU).
This is a complete solution for the change proposed in #1004

----

If accepted, we can add `WITH_SYSTEM_SUPERLU` and eventually `WITH_SYSTEM_TIFF`,

Regarding option initialization - it may seem odd for a single option, see Blender's [CMakeLists.txt](https://developer.blender.org/diffusion/B/browse/master/CMakeLists.txt;ddf99214dc0677542d6ca300c3ee49549e6514bc$128-537) to see how this can be useful.

### DIFF
--- a/toonz/cmake/FindLZO.cmake
+++ b/toonz/cmake/FindLZO.cmake
@@ -1,12 +1,36 @@
+
+if(WITH_SYSTEM_LZO)
+    # depend on CMake's defaults
+    set(_header_hints)
+    set(_header_suffixes
+        lzo
+    )
+    set(_lib_suffixes)
+else()
+    set(_header_hints
+        ${THIRDPARTY_LIBS_HINTS}
+    )
+    set(_header_suffixes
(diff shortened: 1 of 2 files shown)


ideasman42:
Note that the reason for Travis CI failing seems to be unrelated to the changes made.

ideasman42:
A week with no response, merging it since its a simple change that prevents correct builds on Linux.

PR #1022, opened 2017-02-04
Allow no trim on spritesheet

turtletooth:
This allows spritesheets to not have their empty space trimmed off so the user can set a predetermined width and height.

This also changes the startup popup to allow widths less than 120 pixels (one inch).


shun-iwasawa:
Jenkins

shun-iwasawa:
It works fine. LGTM

PR #1023, opened 2017-02-04
Fix for pixels only mode conflict too easily.

turtletooth:
Right now, it is too easy to trigger the warning about scene incompatibility with pixels only mode.  I have made several scenes in pixels only mode and am later told they are incompatible.

This eases the comparison so that any dpi within .1 of 120 is allowed to work in pixels only mode.


turtletooth:
@shun-iwasawa Would you mind taking a look at this and #1022 ?  They are pretty straight forward and help in using OpenToonz as a sprite creator for video games.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

PR #1025, opened 2017-02-05
Fix playback stuck bug

### LINKED ISSUE #1024: Using loop shortcut can get stuck playing
To repeat:
1.  Draw a simple level
2.  Use the L shortcut to start looping playback before you ever hit play in the viewer
3.  Hit the loop button in the viewer
4.  Good luck stopping it

turtletooth:
Fixes #1024 

This sets the active flip console to the visible flip console before triggering the button.

I also set hitting play while playback was active to trigger pause.  This allows playing and pausing using the buttons to be much faster without having to move the mouse.

### DIFF
--- a/toonz/sources/toonzqt/flipconsole.cpp
+++ b/toonz/sources/toonzqt/flipconsole.cpp
@@ -1355,26 +1355,31 @@ void FlipConsole::pressLinkedConsoleButton(UINT button, FlipConsole *parent) {
 //-----------------------------------------------------------------------------
 
 void FlipConsole::onButtonPressed(int button) {
-  // if(!isVisible()) return;
   makeCurrent();
-  doButtonPressed(button);
+  if (m_playbackExecutor.isRunning() &&
+      (button == FlipConsole::ePlay || button == FlipConsole::eLoop)) {
+    pressButton(ePause);
+  } else
+    doButtonPressed(button);
 
   if (m_areLinked) pressLinkedConsoleButton(button, this);
 }
 
 //-----------------------------------------------------------------------------
 void FlipConsole::pressButton(EGadget buttonId) {
-  if (m_buttons.contains(buttonId)) {
-    // If you press "play" or "loop play" button while playing, which means
-    // "pause"
-    if (m_playbackExecutor.isRunning() &&
-        (buttonId == ePlay || buttonId == eLoop)) {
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
@turtletooth Thank you for this PR, but I couldn't reproduce (or I may not yet understand) the issue #1024 . I followed the procedure written in #1024 and confirmed that hitting the loop button while loop-playing causes nothing, but it doesn't seem to get stuck playing (i.e. you can stop the playback by pressing the Pause button afterwards). Could you please advice if there is any other condition to reproduce it?

>I also set hitting play while playback was active to trigger pause. This allows playing and pausing using the buttons to be much faster without having to move the mouse.

I confirmed this enhancement. Thanks!

turtletooth:
@shun-iwasawa The funny thing is that I just tried it and couldn't reproduce it right now either, so I'm not sure exactly what triggers it.  But here's what was happening that caused me to write the PR (my students have also had this happen)  I was using the non-Ghibli rooms and it seems the shortcuts for playback were triggering the last created viewer to play (the x-sheet room).  That was considered the active viewer, even though it wasn't in the active room.  Hitting the control buttons in the basics room didn't trigger the other viewer to stop.  Only going to the xsheet room and hitting pause worked.

So the point of the PR is to make sure that the visible viewer is the one the playback shortcuts trigger.  

I don't quite know why I could trigger it so consistently when I needed this PR but can't right now, but my students have also had  this happen to them.

turtletooth:
@shun-iwasawa BTW- thanks for looking at this and the other PRs you can looked at.  My family is moving houses right now, but I will try and look at your other PRs soon.

turtletooth:
@shun-iwasawa I was trying to duplicate this again tonight and currently can't, but it has happened to both me and some students.  I reviewed the code here and don't think it slows the program - the change only runs if the active console isn't visible.

What are your thoughts?

artisteacher:
I definitely have had several students run into this issue. A fix would be great!

turtletooth:
I managed to recreate the bug:
https://youtu.be/W5NHrX0xguI

shun-iwasawa:
Sorry for my late response on this! Finally I managed to reproduce the problem.
Jenkins

shun-iwasawa:
Confirmed that the problem was fixed. Thank you @turtletooth !

Issue #1026, opened 2017-02-06
Crash when using down arrow with mouse in scene viewer.


turtletooth:
To duplicate

1. Draw a cell in the xsheet
2. With camera/xsheet view still active, move the mouse over the level strip.
3.  Hit down twice.
4. Crash

Looks like the level strip is trying to access frames that don't exist.

turtletooth:
Fixed in #1027

PR #1027, opened 2017-02-06
Fix crash on level strip arrow key presses

turtletooth:
#1026 
This checks to see if a level is null before trying to select


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed the crash fixed. Thank you @turtletooth ! LGTM

PR #1028, opened 2017-02-07
Allow down arrow to create new frame in level strip.

turtletooth:
With the xsheet selected you can already prepare for a new frame for a level using the down arrow.  This allows the same functionality when level view is active.  

It is sometimes beneficial to create levels on their own and not directly in the xsheet for items such as mouths, eyes etc. . . and then later animate them using drawing substitution.


turtletooth:
@shun-iwasawa What are your thoughts on this one?  I created it for making levels intended for drawing substitution.

adamearle:
This would be so soooo handy

At the studio, we thought we might be able to use two open copies of OT. One as a library and the other open copy for copying and pasting the animations into.

one of the biggest problems we have with cutout/ replacement animation is having too many vector drawings in one column will make OT crash alot if you start using the specials tab (should be call "utiities") OT cashes even more.

In comparison to Adobe Flash this is where Adobe Flash really kicks ass. In Flash you can put Multiple sub nested graphics all in one on layer (OT's collapsed columns) but with Flash you can access them from the top level and switch them out.

in OT as far as we know once columns have been collapsed the in way switch drawing out from the "level strip editors timeline is to go into the collapse columns switch drawing out using the "Q" & "W" hot keys.

This is very slow and clumsy ideally working this way The current version of Adobe Flash deals with this really well now but it is still missing a very important aspect that ToonBooom does really well with its library system and that OT only just skims the surface on.

This simply the ability to create library catororgeries of animation this goes right back to Hanna-barbera days and how they did it with the "Smurfs" accept Toonboom does this digitally really well.

This is the kinda thing that is really hard to write about and is better in conversation format accompanied with a demonstration.

Maybe Discord?  [](https://discord.gg/yuuEk78)

Also here is a link to the studios' notes on OT that seem to be never-ending [](https://docs.google.com/document/d/148iZT4U11A3p8K50IF_dVbYtyDk3MBKNYuv96kYi-BU/edit?usp=sharing)

Feel free to chat with me more about this. I hope this doesn't sound like a rant.  If Blender and OT could do this easily then honestly IMO this would be a massive help to artist.

Also this tech that is similar to Blender the Maya version would be really handy for artist as well because it really is a pimped out version of the Blender version: [](http://lesterbanks.com/2017/06/cosmos-command-launcher-maya-finally/)

For us at the studio, it really is about workflow so that artist can make the tools disappear and just create.

Thanks guys

Adam Earle

jpturcotte:
Sorry, @adamearle, but there are no links in your post.

adamearle:
https://docs.google.com/document/d/148iZT4U11A3p8K50IF_dVbYtyDk3MBKNYuv96kYi-BU/edit#heading=h.3ru1wggy9ggm

adamearle:
Discord OT: https://discord.gg/yuuEk78

shun-iwasawa:
IMO this feature is needed to be optional considering ink&paint workflow in Japanese animation industry.
In ink&paint work they usually work on the level strip frames, because they work on each drawings rather than on xsheet frames. However, they seldom create a new frame because all the frames are already drawn by animators in the previous workflow. 
In such case this feature should be disabled in order to navigate easily only inside the existing frame range and to minimize a risk of inserting a new frame by mistake.

turtletooth:
@shun-iwasawa I understand your concern.  Would you place the preference in the drawing section of preferences?

shun-iwasawa:
@turtletooth 

>  Should I place the preference in the drawing section of preferences?

Thank you for your kind understanding. Yes, in the drawing category would be appropriate.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @turtletooth !

Issue #955, opened 2016-12-08
Issues with long animation + video layer


artisteacher:
Changing the output folder causes a crash. This is on a Mac OS X 10.11.6 - with a file that uses a video layer. This happened with a student's scene file in both with 1.1.1 and 1.1.2. A duplicate of the file was saved and has the same issue. Also, the rendered mp4 video will not play. I'm wondering if the scene or video could be corrupted?

Anyways, you are all doing an amazing job. Thanks for all of your hard work!
[crash2.txt](https://github.com/opentoonz/opentoonz/files/642778/crash2.txt)

artisteacher:
So it looks like the problem stems from the length of the video layer.  There is an issue rendering a long mp4 video from a scene with a video layer. The resulting mp4 cannot be opened. Afterwards, a second render with the same name and to the same output location causes a crash.

Steps to duplicate:
- Create and save a scene
- Import an mp4 video to the Extras folder
- Load the mp4 video as a level (resulting level has 201+ frames)
- Save all
- Fast Render to mp4 
- Try to open video - will not open
- Delete frames and save
- Fast Render to mp4 causes crash

Deleting the old mp4 file does allow re-rendering without a crash.

Crash report from 2nd render:
[crash4.txt](https://github.com/opentoonz/opentoonz/files/642665/crash4.txt)

artisteacher:
So I just realized we needed to increase the ffmpeg timeout in the preferences. However, I think that a fix would still be helpful. Ideally, user error shouldn't cause a crash.

turtletooth:
Do you think the default ffmpeg timeout should be extended?  I made it 30 seconds to keep the user from thinking it stalled out, but I have noticed a few folks have had issues with longer animations and not realizing they needed to extend the timeout.

artisteacher:
Extending the default timeout is definitely a good idea. I'm not sure how feasible it would be, but having a warning message pop-up when the timeout length is insufficent would be helpful.

turtletooth:
Unfortunately there is not really a good way to estimate how long it will take ffmpeg to render a video.

artisteacher:
Too bad! Is there a way to automatically delete the incomplete video file if FFMPEG times out before it is complete? That would probably need to trigger an error message, saying timeout was too short.

turtletooth:
I'll try and duplicate the error.  The program already tries to delete the file if it already exists.  I'll try and see what the hangup is.

turtletooth:
[...]

PR #1029, opened 2017-02-07
Extend ffmpeg timeout to 60 seconds

turtletooth:
#955 This doesn't fully solve the problem, but does extend the timeout for ffmpeg.


shun-iwasawa:
LGTM

PR #1034, opened 2017-02-12
Pair field style #881

turtletooth:
This moves the style of pair field sliders to qss.

The standard sliders for the darker two themes have a softer border, while the lightest theme has sharp edges.  This allows for either option via qss.  

lines


turtletooth:
Note- this is a custom UI element and can't be styled as easily as standard Qt elements, but I tried to add what flexibility I could.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

PR #1035, opened 2017-02-12
Move level palette lines to qss

turtletooth:
#883

Lines below the tabs on the style editor and palettes was not accessible via qss.  This fixes that.


shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

Issue #1017, opened 2017-01-31
Loading external files does not truly import


artisteacher:
According to the Harlequin manual, loading an external file (audio, video or etc) as a level should give the option to import or to load from the existing location (creating an absolute path). Using OS X, I do not have that option and all external files are only linked. This is inconvenient for students that want to work at home or for anyone working on multiple computers. I know people can move the files to the correct location manually, but it seems like an unnecessary, additional step. Collecting assets does place the external files into the extras folder but the paths do not update in the level settings (which must be a bug). 

Anyways, it would  be super helpful to have a preference that could be enabled, such as "import external files into the extras folder". Or perhaps when right-clicking in the x-sheet to load a level, there could be an additional option to "import level". 

Thanks for everyone's hard work! OpenToonz is really coming along!

turtletooth:
@shun-iwasawa Is this something that was changed at Studio Ghibli?  It sounds like a useful feature to have.

RodneyBaker:
A few thought related to this...

I know first hand the usefulness of 'Embedding' files in a Project and use
it quite often in other software.  With embedding, the files that cannot be
embedded are linked to their original location  This sounds a lot like what
OpenToonz attempts to do in current configuration.  A separate process
called "Consolidation" can be used to bring everything together into one
file (a compressed zip file that maintains the original file structure so
that when unzipped everything can be found in the same place).  This
'Collection' of external files into the Extras folder you identify sounds
like a great intermediary between ( partial) Embedding and (complete)
Consolidating and thus a very useful option.

If this is indeed something that can be un-regressed that would indeed be
very useful.

shun-iwasawa:
Yes, as @turtletooth pointed,  I changed the software not to ask whether import or load the external resources, which was demanded by Studio Ghibli staffs.
In Studio Ghibli they kept every data at designated folder organized by resource types and scene numbers. Allowing to import data may disturb such arrangement by mistake and may increase file size.

So, they demanded to skip to open the dialog asking "import or load", in order to increase work efficiency.
In particular, I changed the default value of `importPolicy` of a `struct LoadResourceArguments` from `ASK_USER` to `LOAD` in `iocommand.h`.

I think it would be OK to revert such dialog, but I think it should be optional (in preferences) since other Japanese animation studios are taking somewhat of a similar strategy as Ghibli regarding file management.

mcjadski:
Maybe the dialog could have a "Remember my decision" or "Always do this" checkbox.

[...]

PR #1037, opened 2017-02-13
Allow importing files to project dialog box

turtletooth:
#1017 

When loading an external file, this pops up dialog allowing you to choose whether to load the file from its current location or import the file to the current project folder.  There is a checkbox to remember the choice.  This is also in Preferences->Loading.


toonz/sources/toonzqt/dvdialog.cpp
```diff
@@ -757,6 +758,29 @@ void Dialog::addButtonBarWidget(QWidget *first, QWidget *second, QWidget *third,
...
+                                                   bool hasButton,
+                                                   bool hasFixedSize,
+                                                   const QString &name)
+    : Dialog(parent, hasButton, hasFixedSize, name) {}
+
+//=============================================================================
+
+void MessageAndCheckboxDialog::onButtonPressed(int id) {
+  if (m_checked > 0) {
+    done(id + 5);
```
shun-iwasawa: You are returning the state of checkbox by adding 5 to the return value.
It's not a problem for the moment, but if the dialog is created with more than 5 buttons, return value will be overlapped. 
How about replacing this magic number with a larger & named constant (say, `const static int  MessageAndCheckboxDialog::CHECKED = 1000` ) in order to surely prevent the overlap? Or, how about adding a separate function which will return `m_checked` value?

turtletooth:
I made a new function to communicate the checked state.
Thanks @shun-iwasawa for the review.

shun-iwasawa:
Jenkins

shun-iwasawa:
It works fine. Thank you @turtletooth ! LGTM

Issue #1046, opened 2017-02-16
endless png errors - remove bad profiles


keijukainen:
After compiling on Linux the application displays hundreds of errors in the console each time it's run.  They are all related to bad png profiles.

Error is either:
iCCP: known incorrect sRGB profile
OR
iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50

To fix:
mogrify file.png
OR
pngcrush -ow -rem allb -reduce file.png

[png_errors.txt](https://github.com/opentoonz/opentoonz/files/780943/png_errors.txt)

keijukainen:
attached bash script to fix png errors.
`mv fixpng.txt fixpng.sh`
`chmod +x fixpng.sh`
`./fixpng.sh`
[fixpng.txt](https://github.com/opentoonz/opentoonz/files/781166/fixpng.txt)

blackwarthog:
I still have tons of png warnigs:
```
...
libpng warning: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50
libpng warning: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50
libpng warning: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50
libpng warning: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50
libpng warning: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50
libpng warning: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50
libpng warning: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50
libpng warning: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50
libpng warning: iCCP: profile 'Photoshop ICC profile': 0h: PCS illuminant is not D50
...
```

PR #1047, opened 2017-02-17
fixed bad png profiles

keijukainen:
Fixes errors reported below.

https://github.com/opentoonz/opentoonz/issues/1046


ideasman42:
Thanks for the PR, pushed alternate change since it conflicts

turtletooth:
@ideasman42 Your new push undid the work that @konero did to fix the viewer icons.

keijukainen:
Konero's changes should be preserved in this commit. The only changes made were removing unnecessary profiles. But the transparency profiles shouldn't be affected. I suppose I could use "-rem iCCP" instead of "-rem allb" to strip the profiles causing errors. Let me know if you need a new pull request.

PR #1050, opened 2017-02-18
New Viewer Icons

turtletooth:
These are new icons for the titlebar of the viewer and comboviewer.

They are based on @konero icons from earlier.

They are yellow on hover.  The freeze icon is red when activated.  All others are green.

viewer_icons_capture


turtletooth:
I'm going to try and make the color of the icons configurable in qss.  Changes to come.

artisteacher:
That would be exciting!

turtletooth:
Here is a screenshot of the new setup- color of active icons is configurable in qss.  

new_icons

konero:
Quick review...

Did you use the alpha/stencil method (where you make the icon with inverted alpha)?

* Won't the icons be surrounded by a dark/light box if the title bar changes color this way?
* The letters on the 3D icon are not discernible on my 4K display or a Cintiq 13" HD, that was one of many reasons why I was deliberating on that one.
* I can't put my finger on it but they seem to be 'activated' at first glance, probably to do with values. If we take the color out of your preview you can see what I mean and how close the activated icons are to the title bars value.
crop000

If you want to wait I can come back to this and we can deliberate some more and involve more people. I didn't mean to abandon ship, it just got hectic and work piled up over December. ✌️

turtletooth:
@konero I would love your help here.  I did the stencil method which does leave a box around the icon for different themes, but it didn't look that bad.  I'm open to suggestions though.  I couldn't figure out how to bring the whole icon to qss.  I think that would be the best option- I tried, but I just couldn't figure it out.  I don't know if @shun-iwasawa has any ideas on this.

I would love it if a new theme and updated UI were part of the next release, so I don't know if using this for now and tweaking as we go makes sense.  Thoughts?

shun-iwasawa:
@turtletooth @konero Thank you very much for tackling UI improvement. I have not looked into this in detail yet, but as a quick impression - if you use a padlock icon for the freeze button, I expect that it looks like "unlocked (opened)" shape when it is in unfreezed state.

konero:
@turtletooth I made a PR #1103 for now that cleans up the original icons slightly, which should improve compatibility on both yours and my theme.

I'll get my theme up ASAP, I'll comb it through first though, could be anytime - either today or by the weekend.

PR #1054, opened 2017-02-19
Allow border on docks- use QFrame

turtletooth:
This changes the dock widgets from QWidget to QFrame.  Since QDockWidgets weren't used, styling the border of the widgets was unavailable.  

Thanks @konero for the suggestion.

### DIFF
--- a/toonz/sources/toonzqt/docklayout.h
+++ b/toonz/sources/toonzqt/docklayout.h
@@ -163,7 +163,7 @@ private:
 
   \sa DockLayout and DockPlaceholder classes.
 */
-class DVAPI DockWidget : public QWidget {
+class DVAPI DockWidget : public QFrame {
   friend class DockLayout;  // DockLayout is granted access to placeholders'
                             // privates
   friend class DockPlaceholder;  // As above.
--- a/toonz/sources/toonzqt/dockwidget.cpp
+++ b/toonz/sources/toonzqt/dockwidget.cpp
@@ -105,7 +105,7 @@ inline void DockLayout::update() {
 //! (i.e. not docked
 //! into the layout).
 DockWidget::DockWidget(QWidget *parent, Qt::WindowFlags flags)
-    : QWidget(parent, flags)
+    : QFrame(parent, flags)
     , m_dragging(false)
     , m_resizing(false)
     , m_floating(true)


konero:
> @turtletooth the toolbar has a weird issue that the right border only shows up with a border of 2px

I noticed that too. I need to find time to try it out, but I'm going to just throw this assumption out there anyway.

Take this example, if we dock a widget to the left will that widget display a left-border? It would be ideal if it didn't.

If it does, perhaps a new QProperty or conditional to set a negative margin of the mainwindow dock space for themes that would use this, does that sound right? Doubt it would work though, but I hope you get the idea. 👍 

```
qproperty-mainwindow-margin-left: -1px;
qproperty-mainwindow-margin-bottom: -1px;
qproperty-mainwindow-margin-right: -1px;
```

turtletooth:
@shun-iwasawa I don't think this should have any impact on the other functionality.  QFrame inherits from QWidget but simply allows frames.  I wouldn't be surprised if I am overlooking something, however.

turtletooth:
@konero That makes sense and might work.  I just started playing with qproperties and don't much about them yet.  That would definitely add a nice finishing touch to the widgets, but I may have to push it back until I finish with icons and a few theme tweaks.

Let me run an idea by you- your jsfiddle inspired this.  I am going to play with making the green area of the viewer icons in  #1050 actually transparent on their active state and paint the colored part in using qpainter and a qproperty.  That way, the colors could be tweaked for different themes without having different icons for each theme.  The gray area of the icon wouldn't have that flexibility, but at least the color could.

konero:
Wouldn't you need to create multiple icon sets anyway for each theme color for use as the stencils?

This is something else I've not done in a long time so things may have changed, but you can change the color of an SVG file by user input, basically you just overwrite the XML data in the file. I can't see why it won't work in Toonz.

So if there was an option in preferences where you can choose an icon color, it just overwrites the color string in all relevant files, once the app is rebooted it renders a new bitmap.

turtletooth:
@konero The stencils I made do look best on the new theme, but they also work fine with the other themes.  I tried moving the icons themselves to qss but couldn't get it to work.

I think the current setup with the color in qss is fine for now.  If someone else wants to extend it in the future, that would be awesome.

shun-iwasawa:
Jenkins

shun-iwasawa:
I think there is no apparent problem in changing the dock widgets to QFrame. LGTM.

> This is something else I've not done in a long time so things may have changed, but you can change the color of an SVG file by user input, basically you just overwrite the XML data in the file.

@konero Thank you for the suggestion. I agree that replacing images from bitmap to SVG will be applicable not only for availability of changing color but also for adaptation to high-res displays. It would be a future issue.

PR #1060, opened 2017-02-24
Make the QFileSystemWatcher optional in File Browser

shun-iwasawa:
This PR will modify the QFileSystemWatcher behavior in File Browser in two points as follows:

- Enabled to choose whether to watch the file system. The option was added in the General category in Preferences and the default value is ON. If you deactivate this option, you have to refresh the browser manually. A button and menu for such command will appear in the button bar and right-click menu in the File Browser.  

- Modified the File Browser to share one QFileSystemWatcher instance in order to reduce the system resource consumption.

This option was demanded by some Japanese animation studio. They stores all data in a server, separating them in folders for every scenes. OpenToonz hang-up so often when working with the File Browser, due to the overload in monitoring of a large amount of the network folders.


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #232, opened 2016-04-25
Fix "library" and "fxs" folder locations

shun-iwasawa:
This fix is for the issue #110 
With this modification, data for TOONZLIBRARY and TOONZFXPRESETS will be moved as follows:
[Windows] 
- $TOONZROOT/projects/library -> $TOONZROOT/library
- $TOONZROOT/projects/fxs -> $TOONZROOT/fxs

[OS X]
- /Applications/OpenToonz/OpenToonz_1.0_stuff/projects/library   ->  /Applications/OpenToonz/OpenToonz_1.0_stuff/library
- /Applications/OpenToonz/OpenToonz_1.0_stuff/projects/fxs  ->  /Applications/OpenToonz/OpenToonz_1.0_stuff/fxs

Please note that this modification will be reflected in the installer. Re-installing OpenToonz after this modification will newly create "library" and "fxs" folders and it will NOT delete the existing "projects/library" and "projects/fxs" folders. Please delete them manually if needed.

PR #1062, opened 2017-02-25
fixes warnings when library and fxs folders don't exist

keijukainen:
If $HOME/.config/OpenToonz/stuff/projects/library or $HOME/.config/OpenToonz/stuff/projects/fxs folders don't exist then a warning will be displayed in the console each time opentoonz is launched on linux.


inkapnea:
Just for curiosity, I'm relatively new to OpenToonz and trying to figure out the workings...
Why are those two directories still needed under "projects" on Linux while they were moved a while ago to another location with #232 on Win and Mac platforms?

Also, I'm generally confused about the purpose of the "projects" folder. With the v1.1.2 stable release it also contained 3 TXT and INI files that were moved to the "config" directory in subsequent nightly builds. On recent windows builds this directory is empty!

Issue #1063, opened 2017-02-27
Scene path becomes invalid when import scene from sandbox to another project


shun-iwasawa:
**To reproduce:**

1. Save a scene anywhere **outside** of the "+scenes" folder.

1. Set the current project to other than "sandbox".

1. Load the scene saved in (1). A dialog appears asking "The Scene _SCENE_PATH_ belongs to project ’sandbox'. What do you want to do?".

1. Select "Import Scene".

1. Try to save the scene.

After the above operations a warning popup appears saying that the scene path is invalid.

PR #1064, opened 2017-02-27
Fix invalid scene path when import scene from sandbox to another project

shun-iwasawa:
This fixes the issue #1063.
When importing the scene, if the scene is in outside of the "+scenes" folder, it will be imported under the "+scenes" folder of the current project.


shun-iwasawa:
Jenkins

turtletooth:
LGTM  Thanks @shun-iwasawa

PR #1071, opened 2017-02-28
Fix the Flipbook's console to be updated when the "Blank Frame" option is changed in the Preferences

shun-iwasawa:
This PR fixes the Flipbook's console bug as follows.

**TO REPRODUCE:**

1. Open both the Viewer and the Flipbook.

1. Change the "Blank Frames" or "Blank Frames Color" in  Preferences > Preview category.

The Viewer will update the "blank button" at the right side of the frame slider, but the Flipbook won't.


shun-iwasawa:
Jenkins

turtletooth:
LGTM  Thanks again.

PR #1073, opened 2017-02-28
Fix preview fx not to fail updating render result

### LINKED ISSUE #1072: Preview fx fails to update render result
This problem occurs when previewing relatively "heavy" (time-consuming) fx such as the Soap Bubble Iwa Fx.

**TO REPRODUCE:**

1. Create the simple scene using the Soap Bubble Iwa Fx like this: 

1. Trigger "Preview" command. (Flipbook opens.)

1. Select Control Point Editor Tool and change the shape of the level A.

The rendered result in the Flipbook will be updated if you change the shape steadily. But if you change the shape twice before the rendering is finished, the Flipbook will turn black and show nothing.

previewfx_bug

shun-iwasawa:
This PR fixes #1072 .

I toggled `PreviewFxInstance::m_renderFailed` to ON on abort rendering in order to prevent updating `m_overallRenderedRegion` at `PreviewFxInstance::onRenderFinished()` which causes this problem.

### DIFF
--- a/toonz/sources/toonz/previewfxmanager.cpp
+++ b/toonz/sources/toonz/previewfxmanager.cpp
@@ -994,8 +994,13 @@ void PreviewFxRenderPort::onRenderRasterStarted(
 
 void PreviewFxRenderPort::onRenderRasterCompleted(
     const RenderData &renderData) {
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1076, opened 2017-03-01
SoapBubbleIwaFx Update

shun-iwasawa:
This PR updates the SoapBubbleIwaFx as follows:

- The result becomes being masked with the Shape image.

- Added "Multiple Bubbles in Shape Image" option. With this option, the Shape image will be divided into each opaque region and computed separately so that it can render more than one bubble with various size.

- Added "Mask Center of the Bubble" option. With this option, the center of the bubble will be masked so that it can make the "wrinkles" of noise pattern less noticeable.

- Inserted a function which will check whether the current rendering process is cancelled. (For example, when fx parameter changed twice rapidly, the rendering process invoked by the first change will be cancelled.) When cancelled, it will immediately release the raster memory and return in mid-flow of the computation.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

Issue #1007, opened 2017-01-27
Some closed shapes unfillable before switching to Fill Tool


mcjadski:
Right now the coloring functionality is very easily broken on Toonz Vector Levels.  Here is a very simple example to illustrate:  

![deformed_circle](https://cloud.githubusercontent.com/assets/25377429/22357568/a86f4708-e3fe-11e6-8c36-612066119d89.jpg)

This is a blob named Gorilla.  He's a closed shape:  

![gorilla_connect](https://cloud.githubusercontent.com/assets/25377429/22357647/37ee10bc-e3ff-11e6-9a68-b043d8112da9.jpg)

But Fill Check is turned on, and he is unfillable.  However, if you draw anything to the right of him (on the same level), he becomes fillable, whether the dot is near him or far away:  

![gorilla_fillable](https://cloud.githubusercontent.com/assets/25377429/22357718/9d42a37e-e3ff-11e6-9aa5-be5a6fb520e0.jpg)

If you draw to the left of him, above him, or below him, he remains unfillable (unless you draw something to the right of him):  

![gorilla_not_fillable](https://cloud.githubusercontent.com/assets/25377429/22357756/e7b6ba76-e3ff-11e6-955b-0dc5dc61f7ea.jpg)

As you pull one of the end points downward (always keeping it a closed shape), at some seemingly arbitrary point, he becomes fillable:  

![gorilla_almost_fillable](https://cloud.githubusercontent.com/assets/25377429/22357796/36e97520-e400-11e6-8fde-47f3a510187d.jpg)

![gorilla_suddenly_fillable](https://cloud.githubusercontent.com/assets/25377429/22357817/531e9af4-e400-11e6-9fa9-48902dabbc2f.jpg)

Here are the .tnz and .pli file:  
[gorilla_test.zip](https://github.com/opentoonz/opentoonz/files/734154/gorilla_test.zip)

I tested this file on Win 7 and OSX 10.11 and got the exact same results.  

If such an incredibly simple picture as this can't be filled correctly, imagine how nightmarish it would be to make fillable a complicated drawing...!  

OpenToonz is on the whole an excellent program, and I have HUGE respect for the team;  but this is something of an "Achilles' heel" right now, as it renders the vector functionality all but unusable.

turtletooth:
[...]

PR #1079, opened 2017-03-03
(Partially) fix vector shapes unfillable

shun-iwasawa:
This PR fixes issue reported in #1007 , ~~and hopefully fixes other "unfillable vector" problems.~~ ( Unfortunately, the third problem still remains. )

As far as I noticed, there were ~~two~~ three (I found one more, now) separate problems related to the "unfillable vector" as follows:

1. Fill Check information are not properly updated in some shape. Confirmed with the sample  [gorilla_test.zip](https://github.com/opentoonz/opentoonz/files/734154/gorilla_test.zip) provided by @mcjadski . Also he reported that the Fill Check fails to detect the last stroke. (Thank you @mcjadski for the "random fish" test!)

1. Very small region could not be filled. For now, an area less than approx. 0.0001 square inch (= 1.26 square pixel in 120dpi scene) could not be filled.

1. Other unfillable shapes like this. [unfillable.zip](https://github.com/opentoonz/opentoonz/files/816428/unfillable.zip)

The cause of the problem 1 is related to the timing of calling the function `void roundStroke(TStroke *s)` at the [line 50 of tcomputeregions.cpp](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/common/tvectorimage/tcomputeregions.cpp#L50). In this function, all points of vector strokes are slightly moved to discrete positions. If the Fill Check is processed before calling this function, the check will be failed because of as follows: The Fill Check does rectangular fill with the bounding box of whole strokes. Without calling `roundStroke()`, it will be called  "in the middle" of Fill Check procedure and the strokes will be moved and may run off the box. 
In this PR I added `TVectorImage::findRegions()` on loading PLI file and on adding a stroke with brush tool. `TVectorImage::findRegions()` will be followed by `roundStroke()`. 

The cause of the problem 2 was not a bug, there was some threshold of area size [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/common/tvectorimage/tcomputeregions.cpp#L2585). 
For now the value is 0.5 (unit is a square of 1/53.33333 inch). Which means that the rectangular with less than approx. 1.125 * 1.125 pixel size will be unfillable in 120dpi scene. I reduced this threshold to 0.1 in order to fill small region comparable to a single pixel.

Actually I just enabled to reproduce the problem 3 now. I'll investigate this.

### DIFF
--- a/toonz/sources/common/tvectorimage/tcomputeregions.cpp
+++ b/toonz/sources/common/tvectorimage/tcomputeregions.cpp
@@ -2582,7 +2582,7 @@ public:
     m_quasiArea += (p2.y + p1.y) * (p1.x - p2.x);
   }
 
-  bool isClockwise() { return m_quasiArea > 0.5; }
+  bool isClockwise() { return m_quasiArea > 0.1; }
 };
 
 //----------------------------------------------------------------------------------------------
--- a/toonz/sources/image/pli/tiio_pli.cpp
+++ b/toonz/sources/image/pli/tiio_pli.cpp
@@ -420,6 +420,8 @@ TImageP TImageReaderPli::doLoad() {
   outVectImage->checkIntersections();
 #endif
 
+  outVectImage->findRegions();
+
   return TImageP(outVectImage);
 }
 
--- a/toonz/sources/tnztools/brushtool.cpp
+++ b/toonz/sources/tnztools/brushtool.cpp
@@ -374,6 +374,12 @@ static void addStroke(TTool::Application *application, const TVectorImageP &vi,
                                                 frameCreated, levelCreated));
   }
 
+  // Update regions. It will call roundStroke() in
+  // TVectorImage::Imp::findIntersections().
+  // roundStroke() will slightly modify all the stroke positions.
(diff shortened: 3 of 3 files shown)


shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa Thank you so much for looking into this.  The fill issues on vectors are a common issue and are enough to drive someone away from OpenToonz.  Thank you!!

turtletooth:
LGTM.  Thanks @shun-iwasawa for the work on this.

Do you know if problem 3 existed before the changes around Stage::inch and StandardDpi were implemented?  Just trying to throw out ideas. . .

shun-iwasawa:
@turtletooth Thank you for reviewing this. I tried to look in depth regarding the problem3 and found that the cause of this problem is failure of auto-closing with self-crossing stroke. I couldn't solve the problem yet but I'll write my finding in #1093 by this weekend.

shun-iwasawa:
@turtletooth  Commented. Please see #1093 !

PR #1086, opened 2017-03-07
Enable the "Convert File" command from main menu bar 

shun-iwasawa:
For now, the "Convert File" command triggered from the main menu bar is not available since the "Convert" button in the convert popup is grayed out. This PR fixes such malfunction, by displaying the "File to convert" field which already existed in the source but has been long-ignored.
(This bug seems to be inserted when I made Toonz Ghibli Edition from Toonz Harlequin...!)


shun-iwasawa:
Jenkins

turtletooth:
It works.  Thanks @shun-iwasawa

PR #1090, opened 2017-03-09
Fix path to textures, effects, brushes and templates

blackwarthog:
Fix paths in "~/.config/OpenToonz/SystemVars.ini" for Linux


morevnaproject:
Without this fix there are no presets for vector brushes and textures in Linux version of OpenToons.

blackwarthog:
I've splitted first commit and also rename BUILD_ENV_~~GCC~~ variable to BUILD_ENV_UNIXLIKE to make it meaning more clear.

inkapnea:
Are those two related and one of them could be redundant?
#1062

morevnaproject:
@inkapnea The problem is that "fxs" and "library" dirs are exist (with a lot of templates), but ini file points to wrong directory, so OpenToonz looks in the wrong place.

inkapnea:
@morevnaproject Ok, thank you! I can see that the dirs you referenced are populated with files.
I suppose I still have to understand the logic and the purpose of multiple directories and subdirectories with the same name under Stuff like:
/stuff/projects/fxs/
and
/stuff/fxs/
and such...

shun-iwasawa:
LGTM. @blackwarthog thank you for the fix!

Issue #417, opened 2016-06-07
Feature: libmypaint for brush tool in raster levels


blurymind:
Libmypaint is a popular brush library used by mypaint and now also available for gimp. Pencil2d devs did some work to make it easier to integrate in qt apps. I thought that it might be useful to have a link to it here:

https://github.com/feeef/QTMyPaint

This is a big task, but it will make the brush tool extremely powerful.
We can take the best from the open source world :)

Btw i will post some demo videos of mypaint in action later.

Here is libmypaint's original github for the latest version:
https://github.com/mypaint/libmypaint

blurymind:
Libmypaint is a very powerful open source brush engine. It had a huge set of brush presets created for it. Its now a part of gimp and mypaint, but it is also ported to blender, krita and pencil2d among others. Here are some demo videos:

https://youtu.be/MLMqOyxUPvg

https://youtu.be/T-_gofxpgNQ

http://www.davidrevoy.com/article142/ressource-mypaint-brushes

http://www.davidrevoy.com/article55/mypaint-v4-brushkit

https://youtu.be/hzUwPF1_6H8

blurymind:
Here is something relevant you should consider:
https://plus.google.com/+gimp/posts/agohJvtj4Sm

During +Libre Graphics Meeting in Madrid we met awesome folks from +mypaint, Krita, and Tupi teams and discussed how we could simplify sharing resources like brushes. So we came up with a proposal for a simple ZIP-based file format for distribution. +Boudewijn Rempt of Krita fame posted the first version of the proposal for public discussion last night.

Later on we could also use this file format for distribution of other types of 3rd party resources (gradients, palettes etc.).

We are also considering adding support for MyPaint brushes in either 2.10 or 3.0, depending on the workload. This is now easy, since MyPaint team recently abstracted their brush engine into a nice simple library.﻿
https://lists.freedesktop.org/archives/create/2013-April/004616.html

They already added libmypaint to gimp btw.

turtletooth:
It is GPL licensed.  I love the idea, but the licenses are incompatible.

Banbury:
QTMyPaint is GPL. libmypaint isn't. 

So writing your own wrapper is still an option. I've had a short look at QTMyPaint and it doesn't look like brain surgery to me.

turtletooth:
Awesome!  Good to know!

blurymind:
qtmypaint is just a qt implementation for the interface, right?
[...]

PR #1092, opened 2017-03-09
Allow to cross compile for Windows with Linux/Mingw

blackwarthog:
Allow to compile OpenToonz for Windows with MinGW/GCC. I've checked native build for Linux and crosscompilation from linux (mingw-w64) to Windows.

Build in MacOS and MSVC should works fine, but i've not tested.


toonz/sources/tnzbase/CMakeLists.txt
```diff
@@ -207,17 +207,17 @@ elseif(APPLE)
...
     _find_toonz_library(EXTRA_LIBS "tnzcore")
-    set(EXTRA_LIBS ${EXTRA_LIBS}
-        ${SDL_LIB_LIBRARIES}
-    )
-
-    include_directories(
-        SYSTEM
-        ${SDL_LIB_INCLUDE_DIRS}
-    )
-    target_link_libraries(tnzbase Qt5::Core Qt5::Gui)
```
ideasman42: Did you intend to remove this line?

toonz/sources/toonzfarm/tfarm/CMakeLists.txt
```diff
@@ -38,19 +38,22 @@ include_directories(
...
     set(EXTRA_LIBS
         ${CMAKE_CURRENT_BINARY_DIR}/../../tnzcore/libtnzcore.dylib
         ${CMAKE_CURRENT_BINARY_DIR}/../../tnzbase/libtnzbase.dylib
         ${CMAKE_CURRENT_BINARY_DIR}/../../toonzlib/libtoonzlib.dylib
     )
-elseif(UNIX)
+elseif(BUILD_ENV_GCC)
     _find_toonz_library(EXTRA_LIBS "tnzcore;tnzbase;toonzlib;tnzext")
+    if(BUILD_TARGET_WIN)
+        set(EXTRA_LIBS ${EXTRA_LIBS} -lws2_32)
```
ideasman42: This will use `-lws2_32` on unix too

ideasman42:
What advantages does MinGW have as a target for OpenToonz?

----

One concern with this PR, it mixes `UNIX` & `MINGW` branches where its not appropriate (especially where libraries are concerned).

I don't think its a good idea to replace `if(UNIX)` with `if(BUILD_ENV_GCC)`. Since there are non-gcc compilers on Unix (Intel's compiler for example).

Checks for platform and compiler shouldn't be mixed up.

Before going into details its probably best to figure out if its worth supporting MinGW.

toonz/sources/tnzbase/CMakeLists.txt
blackwarthog: You right, this line got here by mistake. I wanted to ask you about two lines [209](/opentoonz/opentoonz/blob/master/toonz/sources/tnzbase/CMakeLists.txt#L209) and [220](/opentoonz/opentoonz/blob/master/toonz/sources/tnzbase/CMakeLists.txt#L220)(this line) in this file which seems unnecessary, because command in final line replaces them:

```
if(WIN32)
    set(EXTRA_LIBS
        tnzcore
    )
elseif(APPLE)
    find_library(TWAIN_LIB TWAIN)
    find_library(IOKIT_LIB IOKit)
    find_library(COCOA_LIB Cocoa)

    set(EXTRA_LIBS
        ${CMAKE_CURRENT_BINARY_DIR}/../tnzcore/libtnzcore.dylib
        ${USB_LIB}
        ${TWAIN_LIB}
        ${IOKIT_LIB}
        ${COCOA_LIB}
    )

    target_link_libraries(tnzbase Qt5::Core Qt5::Gui) # <------------------ this
elseif(UNIX)
    _find_toonz_library(EXTRA_LIBS "tnzcore")
    set(EXTRA_LIBS ${EXTRA_LIBS}
        ${SDL_LIB_LIBRARIES}
    )

    include_directories(
        SYSTEM
        ${SDL_LIB_INCLUDE_DIRS}
    )
    target_link_libraries(tnzbase Qt5::Core Qt5::Gui) # <------------------ and this
endif()

target_link_libraries(tnzbase Qt5::Core Qt5::Gui ${EXTRA_LIBS}) # <-------- and finally
```

I'll fix this commit. Will it right to remove both these lines, or them have some meaning?

blackwarthog:
MinGW is not only simulation of Linux-environment in Windows. It is also cross-compilation environment that can to build windows-applications from Linux OS. So we don't need to have separate computer, VirtualBox or QEMU for build for Windows.

We have six build targets for now: Win32, Win64, Mac32, Mac64, Unix32, Unix64. With this PR four from them we may build from the same environment at the same computer. Also the way to search dependecies in MinGW is identical to Unix - i mean pkgconfig and directories /usr/lib, /usr/local/lib, /usr/share/cmake, etc. So when developer will edit dependencies for Unix, he/she will not worry about build for Windows-MinGW, and most of linux developers can verify them code contributions at both systems Linux and Windows (with wine).

You right about variable BUILD_ENV_GCC, this name is not correct and makes confusion. Actually it means unix-like environment for search dependencies, and i actually don't check the compiler (gcc or not). I'll better to rename it into BUILD_ENV_UNIXLIKE.

Because in this PR i introduce ability to cross-compilations i think will be better to represent build method in two independent aspects - build environment and build target (target OS). Because we may to build from unix-like environment to windows target.

As like you I already thought about the simple method with four build targets: MSVC, Apple, Unix and MinGW. But while implementation i seen that this method produce a lot of code duplication in Unix and MinGW parts. Actually MinGW configuration just adds couple of libraries to the Unix configuration. So i came to solution to considers system specifications in two aspects: build environment and target OS.

In common way definition of configuration differences will looks like this:
```
# set options specified for target OS
if (BUILD_TARGET_WIN)
    ...
endif()

# set options specified for build environment
if(BUILD_ENV_MSVC)
    ... for example take deps from thirdparty directory
elseif(BUILD_ENV_APPLE)
    ...
elseif(BUILD_ENV_UNIXLIKE)
    ... libraries configuration for unix-line environment
    if(BUILD_TARGET_WIN)
        ... some options special for Windows-MinGW (freeglut for ex.)
    else()
        ... some options special for unix targets only (for ex.: sdl, may be JACK)
    endif()
endif()
```

This way seems proper for projects with cross-compilation feature.

ideasman42:
Would be good to get maintainers feedback on this topic before spending too much time.

Suggest splitting out all trivial changes into separate PR (`WIN32` -> `_MSVC`) checks for example.

morevnaproject:
> The benefits you mention are only really advantages if active developers are going to make use of them.

Hello @ideasman42.  Let me clarify some aspects. @blackwarthog is working on integrating MyPaint brushes engine into OpenToonz (see #417). Actually, he already got first  version working (https://github.com/blackwarthog/opentoonz/commits/brush). 

Our team is already delivering AppImage  builds of OpenToonz for Linux users.  We are going to provide testing Linux builds with MyPaint brushes integration, so our users will be able to do early testing of this feature to eliminate possible issues and bugs. We also would like to deliver Win32/64 builds, to allow  Windows users do early testing as well.  We create windows binaries on Linux environment, using mingw crosscompilation. This allows to deliver builds for both platforms in the shortest time. Also, this allows for us to utilize our Linux servers for automated builds. I hope you will consider those arguments as _"use by active developers"_. ^__^

It would be really nice to get feedback from maintainers about this pull request, before it gets outdated.

toonz/sources/tnzbase/CMakeLists.txt
blackwarthog: i made separate commit for it

blackwarthog:
I've splitted first commit and also rename BUILD_ENV_~~GCC~~ variable to BUILD_ENV_UNIXLIKE to make it meaning more clear.

ideasman42:
@morevnaproject it sounds like this change is mainly for cross compilation then? In that case is there anything preventing free windows10 and MSVC being used?

Even if its some work to setup, it avoids burden of maintaining a separate build system (where changes to Linux might easily break MinGW) which is only used for one purpose.

morevnaproject:
> it sounds like this change is mainly for cross compilation then? 

Exactly.

> In that case is there anything preventing free windows10 and MSVC being used?

Windows 10 is not free. ^__^

We can not afford (read: "not willing") to pay for extra licenses within our development infrastructure.

Also, when crosscompiling on Linux I can be 100% sure that we deliver builds without any virus/malware. With WIndows I cannot be sure about that.

> changes to Linux might easily break MinGW

Our team can take a burden to fix them (if any), since we are actually using this toolchain.

> Has this patch been tested building on MS-Windows / MinGW?

The purpose of this patch is crosscompilation on Linux host. I think MS-Windows / MinGW compatibility should be the next (independent) step (if anyone is interested in that).

ideasman42:
> Windows 10 is not free. ^__^
>
> We can not afford (read: "not willing") to pay for extra licenses within our development infrastructure.

There is a free version I've installed on a VM with MSVC and used to build a C/C++ application. It was quite locked down, but worked for compiling. I don't know Windows 10 well, maybe they change policy since? ... but quite sure free version remains.

morevnaproject:
> There is a free version I've installed on a VM ...

Yes, there is free VM of Windows. But compilation on VM is slow. My other points regarding malware/viruses remain actual.

toonz/sources/tnzbase/CMakeLists.txt
ideasman42: Committed to OpenToonz c6e5093e87f96eaff8526cd619da5c96fca264ee

toonz/sources/toonz/main.cpp
```diff
@@ -385,7 +385,7 @@ int main(int argc, char *argv[]) {
   fmt.setStencil(true);
   QGLFormat::setDefaultFormat(fmt);
 
-#ifdef LINUX
+#if defined(LINUX) || (defined(_WIN32) && defined(__GNUC__))
```
ideasman42: While this is a  small example, why would function call on MINGW not be called with MSVC ?

Having MSVC & MINGW behave differently WRT function calls seems like its going to cause mismatch on behavior, possibly bugs on one system but not others.

ideasman42:
Note that I'm not going to the person who signs off on this PR, but am not convinced of its overall benefits.

For Blender we had MinGW for some years and kept running into bugs, even recently until it just wasn't worth maintaining.

Even besides the bugs there are minor differences between Linux/Mingw/MSVC making Mingw maintenance more hassle then you might expect. Some examples: 

- https://developer.blender.org/T26330
- https://www.blender.org/forum/viewtopic.php?f=5&t=10088
- https://developer.blender.org/rBS308f813f3a1e7269e33212d07438d312bc8fcd58
- https://developer.blender.org/rBd227471b72f6f20b35c8cfa5fc5cf7d264060838
- https://developer.blender.org/rBa6ebc9a813fab5e3d56b347b763ce4e99704e69e
- https://developer.blender.org/rB77f357728f708dd7a19a62110f34fa9afa5e9213

toonz/sources/toonz/main.cpp
blackwarthog: I don't know. Official documentation at opengl.org tell that GLUT should be initialized, but MSVC-part of OpenToonz does not call glutInit, i don't know why. I thought that this bug discovered while porting to linux, and the fix was added for LINUX only to avoid potential risks to broke compatibility with windows and mac.

https://www.opengl.org/resources/libraries/glut/spec3/node9.html

Me too don't wanted to broke MSVC- and Mac-code and kept it unchanged. But i think this function should be called at all systems, will be cool if anyone can check this at Mac and/or MSVC.

You right i was should to comment this.

morevnaproject:
> I'm not convinced of its overall benefits of this PR.

> I find it a little strange that you are pushing to run your own nightly build

You can do nightly builds only when changes are merged to main OT repository, right?
We are developing some features for OpenToonz, like this one - https://morevnaproject.org/2017/03/16/opentoonz-mypaint-brushes-available-download/
Merging such changes (through PR) usually takes a long time (and that's perfectly understandable why). We would like to provide for our users a way to do early testing of our feature to deliver you a better polished pull request. Also, early testing allows to get meaningful feedback from real users (not geeky ones only, who can build source themselves) to understand if some feature really makes sense. This is an important part of research.

> Mingw maintenance more hassle then you might expect

We have experience with maintaining MinGW for Synfig for years. I repeat: we are willing to take burden on maintaining it. If it will get broken then we will fix. This not harms your build process.

> MSVC/MINGW C++ ABI incompatibility also complicates including new libraries

We have this solved within our build system.

> keep UNIX/MSVC/LINUX checks, add just add MINGW as a 4th option before LINUX should work

This would make maintaining more complicated. And things are more likely to break at every change to build system.

ideasman42:
> > I'm not convinced of its overall benefits of this PR.

> > I find it a little strange that you are pushing to run your own nightly build

> You can do nightly builds only when changes are merged to main OT repository, right?
> We are developing some features for OpenToonz, like this one - https://morevnaproject.org/2017/03/16/opentoonz-mypaint-brushes-available-download/
> Merging such changes (through PR) usually takes a long time (and that's perfectly understandable why). We would like to provide for our users a way to do early testing of our feature to deliver you a better polished pull request. Also, early testing allows to get meaningful feedback from real users (not geeky ones only, who can build source themselves) to understand if some feature really makes sense. This is an important part of research.

Did you ask if its possible to have branches included in regular builds?

Further you speak as if using Mingw the only way to do your own nightly builds, which isn't the case.

> > Mingw maintenance more hassle then you might expect

> We have experience with maintaining MinGW for Synfig for years. I repeat: we are willing to take burden on maintaining it. If it will get broken then we will fix. This not harms your build process.

Supporting different environments adds overhead for developers and users if they run into Mingw spesific bugs.

> > MSVC/MINGW C++ ABI incompatibility also complicates including new libraries

> We have this solved within our build system.

How?

> > keep UNIX/MSVC/LINUX checks, add just add MINGW as a 4th option before LINUX should work

> This would make maintaining more complicated. And things are more likely to break at every change to build system.

Not convinced of this, especially if Mingw on WINDOWS is to be supported,

blackwarthog:
I'll describe some technical details. 

>>> MSVC/MINGW C++ ABI incompatibility also complicates including new libraries

>> We have this solved within our build system.

> How?

Build workflow in MinGW is the same to Linux build workflow. So before build OpenToonz we need install some dependencies as described [here](https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_build_linux.md#required-software). And whole this documentation file (https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_build_linux.md) gives good explanations how to build OpenToonz for MinGW (and Linux of course), but Windows-build documentation not.

Fedora provides many packages for MinGW, also some MinGW packages avaiable in Debian. MinGW for windows have own repositiry. So we can install some deps from repo. And, of course, we can build any package from sources as we can do in Linux. Also packages built with MinGW provides configs for pkg-config and cmake. In our build system we build almost all dependencies from the sources. It's our decision to use latest versions of libraries. 

This is the main reason why i suggest to insert MinGW options inside of UNIX-branches of CMake files. Any additinal options in UNIX (UNIXLIKE) branch with very high chance will work fine for MinGW too. But if MinGW will have separate branch in CMake then nobody will add copy of UNIX-options into MinGW branch. And we will need to replicate any modifications commited into UNIX-branch (of CMake-conditions) manually after every such commit.

ideasman42:
>>>> MSVC/MINGW C++ ABI incompatibility also complicates including new libraries

>>> We have this solved within our build system.

>> How?

> Build workflow in MinGW is the same to Linux build workflow. So before build OpenToonz we need install some dependencies as described [here](https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_build_linux.md#required-software). 

This does not solve the problem then. It means every library OpenToonz uses must be able to be built with Mingw too.

I can only assume this would cause binary incompatibility with plugins too *(although I'm not sure how important this is)*.

turtletooth:
First of all, @morevnaproject and @blackwarthog Thank you both a ton for the big features you are working on.  I'm excited to see them become part of OpenToonz.

I know basically nothing about build systems but I hope a solution can be found.

@morevnaproject Is the only thing preventing you from using MSVC the desire to build it all from a linux environment?  Are the builds you are hoping to release mostly for testing purposes until the changes are merged with OT master?

I maintain the portable version of OT and for this version, I have a branch that I merge with master when I need a new portable version.  I don't need these changes to be part of the main repo and just add them when I need them.  Would that be an option here- keep your changes as a separate branch and as part of the build process merge that branch with your release?  

@shun-iwasawa Would you mind weighing in on this?  Do you have any thoughts on a solution that would meet the needs of all involved?

morevnaproject:
> > > > MSVC/MINGW C++ ABI incompatibility also complicates including new libraries

> > > We have this solved within our build system.

> > How?

> > Build workflow in MinGW is the same to Linux build workflow. So before build OpenToonz we need install some dependencies as described here.

> It means every library OpenToonz uses must be able to be built with Mingw too 

I repeat: we have a build system (a library of build "reciepes") that automatically builds  all dependencies for given application - https://github.com/morevnaproject/morevna-builds/tree/win. 

Here is a list of "receipes" - https://github.com/morevnaproject/morevna-builds/tree/win/docker-builder-data/build/script/packet

This system is already proven for Synfig (which have 3-4 times more dependencies than OT) and Papagayo.

The approach is similar to MacPorts, Homebrew, Nix, etc.

> every library OpenToonz uses must be able to be built with Mingw too (this isn't guaranteed although it happens to be the case right now).

Cross-compiling with MinGW is very common approach for many applications.  So, the coverage of libraries is very good - almost everything can compile with MinGW. For example, check the list of libraries in MXE - http://mxe.cc/

morevnaproject:
> Did you ask if its possible to have branches included in nightly builds?

We still need to build on our own systems to make a proper development.

> You just choose not to setup a windows build environment, which is quite a reasonable thing to do if you want to automate windows builds, and seeing as all tools to do so are free (gratis).

- Yes, with MinGW it's easy to automate builds.
- Yes, we have automated builds on VPS server. 
- Yes, the tools are free. 
- Yes, cross-compiling on MinGW guarantees that the binary is virus/malware free. 
- Our builds do not require Visual C++ 2015Redistributable, so there are less hassles for users running Win7 and older.
- With MinGW we plan to deliver 32-bit builds as well, since I saw several user requests here and there about them.

> Supporting different environments adds overhead for developers and users...

Again, I repeat: we are willing to take a burden of maintaining MinGW-specific part. All the rest remains the same.

> ...if they run into Mingw spesific bugs.

The bugs you outlined below is not a fault  of MinGW. In many cases it's a code quality flaws (standard violations, hacks, GOTOs, etc). For example, just check cstring usage bug, which was discovered yesterday - https://github.com/opentoonz/opentoonz/pull/1118
So, MinGW helps to make your code better.

morevnaproject:
>>> keep UNIX/MSVC/LINUX checks, add just add MINGW as a 4th option before LINUX should work

>> This would make maintaining more complicated. And things are more likely to break at every change to build system.

> Not convinced of this, especially if Mingw on Windows is to be supported.

See reply form @blackwarthog:

> This is the main reason why I suggest to insert MinGW options inside of UNIX-branches of CMake files. **Any additinal options in UNIX (UNIXLIKE) branch with very high chance will work fine for MinGW** too. But if MinGW will have separate branch in CMake then **nobody will add copy of UNIX-options into MinGW branch**.

Imagine, someone adds an option for UNIX. There are high (really high) chances that the same option will be required by MinGW. So, if MinGW in the same "case", then it is automatically "in business" and no additional actions required. But if MINGW is separate "case", then there is a high probability that a person who added an option will forget to add it to MINGW "case". So, the things are more likely to break and harder to maintain.

morevnaproject:
> Wouldn't this mean plugins built for MSVC will fail with an Mingw build? (although I'm not sure how important this is, OpenToonz does have a plugins system so it should be considered).

The things are worse than you think. Even without  MinGW you will have ABI incompatibility with different versions of MSVC. More than that, MSVC uses different ABI for "debug builds" and "release builds" **within the same version**.

What would you do if someone (externally) ships plugin to you (as binary) and you want to debug it? Then for plugin should be also delivered a "debug binary", compiled with the same MSVC version as your version of OT. It's easy if plugin's source is available, but what if it isn't?

So, it's not a problem of MinGW. @blackwarthog sasy it is probably possible to write ABI converter to allow connect plugins compiled with different ABIs. But it's not 100% clear, needs additional investigation.

Just to mention: AFAIK the ABI of MinGW is unchanged since 2004.

morevnaproject:
@turtletooth Thank you for the kind words!

> @morevnaproject Is the only thing preventing you from using MSVC the desire to build it all from a linux environment?

Yes building from Linux environment have several advantages:
- It's faster.
- Resulting builds do not require Visual C++ 2015Redistributable (work out-of-box)
- Builds are automated
- They are guaranteed to be free from virus/malware (Which might be the case with MSVC if your host WIndows system gets infected. We have several virus alarms during development of Synfig - some antivirus software claimed that we delivered virus within our builds. In this case I had convenience to contact antivirus developer and claim to fix false-alarm.)
- We can easily deliver 32bit builds for windows.

> Are the builds you are hoping to release mostly for testing purposes until the changes are merged with OT master?

Currently - yes. This is to allow users to do early testing. But we also plan to deliver a stable builds for 32bit versions of Windows (since I've heard there is no interest from OT team to do that).

> I maintain the portable version of OT and for this version, I have a branch that I merge with master when I need a new portable version. I don't need these changes to be part of the main repo

This doesn't work here. You can examine the list of changes for this pull - they are huge. Everytime we do merge/rebase (happened several times during this discussion), we've got a bunch of conflicts. It takes time to resolve all them every time. This is a manual work and it makes impossible to make automated builds.

blackwarthog:
> So, it's not a problem of MinGW. @blackwarthog sasy it is probably possible to write ABI converter to allow connect plugins compiled with different ABIs.

Some note. I've done some investigation. Seems it's impossible in our case. Parent and child class have different ABI, it's very serious collision.

Anyway i cannot assume situation when we cannot rebuild plugins. And MSVC changes ABI every major release.

ideasman42:
> More than that, MSVC uses different ABI for "debug builds" and "release builds" within the same version.

While true this is a strange point to be making, regular users aren't using debug builds.

> @blackwarthog sasy it is probably possible to write ABI converter to allow connect plugins compiled with different ABIs. 

This is highly speculative, I would assume if this were practical it would have been done already since MSVC ABI incompatibility is a very well known pain-point for Mingw.

morevnaproject:
> > @blackwarthog says it is probably possible to write ABI converter to allow connect plugins compiled with different ABIs.

> This is highly speculative, I would assume if this were practical it would have been done already since MSVC ABI incompatibility is a very well known pain-point for Mingw.

Yes, as @blackwarthog mentioned in previous comment (https://github.com/opentoonz/opentoonz/pull/1092#issuecomment-288951239) this wouldn't work. He put corrections on that, as you can see. ^__^

blackwarthog:
I thought about unify format of plugin file, but it' seems impossible because OpenToonz does not use any plugin API. A plugin code just becomes a part of OpenToonz code and have full access to all functionality. Am i right? It's good but, clearly, plugin should be compiled with the same tools and options as main OpenToonz files. GCC (and MinGW) solved this broblem for C++, MSVC not. MS recommends to use C-style API for compatibility. Anyway we have other problem. In such plugin system a plugin should be compiled with the same code in **header files**, changes in any header will break a plugin that used it.

When we provide for plugin full access to main OpenToonz functionality, we should to have ability to rebuild plugins in case of any change in that functionality. That means all the time of the development process. So we can rebuild plugins for MinGW too when it's will required.

Mentioned python written at C, isn't it? Plugin system with stable strong and well-documented API, or just C-style API, may be compatible or convertible to other ABI. Also some c++ applications uses script-language for their plugins. So "just c++ apps with plugins" is a bad example.

We should consider only cases with plugin system like used in OpenToonz. I don't know such which using independent thirdparty providers of precompiled plugins.

I think reasoning about abstract plugins already went too far away from topic of this pull-request. This PR keeps in work all functionality and plugin system too, and all agree with it.

ideasman42:
@blackwarthog opentoonz has a plugin system which is written using C++ from what I can tell:
https://github.com/opentoonz/dwango_opentoonz_plugins

If a new platform (Windows+Mingw) is to be added, its quite reasonable to check if using different ABI's is likely to cause real-world problems for users & developers who need to support them.

----

As for the Python example, I ran into ABI issues with MSVC & Python before, so its not just limited to C++,  (although not sure this example helps, just to show it exists, see this post: https://mail.python.org/pipermail/python-dev/2014-October/136626.html ).

The point is Python used a single MSVC version on windows so modules could be easily used on Windows and managed to get ABI compatibility.
Also noteworthy is CPython isn't compatible with mingw [(unless you count 3rd party patch sets for older Python versions)](http://mingwpy.github.io/motivation.html)

blackwarthog:
I agree. I've fixed my last commit including your notices. But comonly your passed by main sence of my message. Also I remember how i've imported functions from thirdparty dll in Delphi. Because Delphi has amazing ABI compatible with every MSVC dlls. Or incompatibility of ABI has the limits, which should by clarified.

DLL ABI incompatibility affects:
- call methods;
- name mangling;
- packing data in structs;
- formats of C++ classes (vfat etc).

First two points are solvable in all cases. Point three is solvable in most cases. Only point four is true problem.

morevnaproject:
> we have a build system (a library of build "recipes") that automatically builds all dependencies for given application - https://github.com/morevnaproject/morevna-builds/tree/win.

Note: we have merged "win" branch to "master, so the correct URL for build system is the following now - https://github.com/morevnaproject/morevna-builds

morevnaproject:
@ideasman42 please consider this part of @blackwarthog 's message:

> In such plugin system a plugin should be compiled with the same code in header files, changes in any header will break a plugin that used it.

If I understood correctly, any change to headers might break plugin compatibility (within the same toolchain). So, when header is changed all plugins should be re-compiled against the new version. Am I right?

ideasman42:
> @ideasman42 please consider this part of @blackwarthog 's message:
> 
>> In such plugin system a plugin should be compiled with the same code in header files, changes in any header will break a plugin that used it.
>
> If I understood correctly, any change to headers might break plugin compatibility (within the same toolchain). So, when header is changed all plugins should be re-compiled against the new version. Am I right?

Yes, but again this is a strange point to be making.

In practice when you maintain a plug-in interface, this is typically a small API made with the express purpose of being stable. While breakages may happen, it's normally managed between releases, with notice for plug-in developers.

This is common practice for scripting languages and applications where ABI breakages are generally only accepted between major point releases.

shun-iwasawa:
First of all, @blackwarthog thank you so much for your big contribution done so far regarding introduction of LibMyPaint into OpenToonz. Also thank you @ideasman42 and @morevnaproject for deeply discussing this matter.
Please let me recap the situation, and then I would like to comment for your PRs (both #1092 and #1124). I'm unfamiliar with MinGW, so please point if I overlook something.

1. As it is now, LibMyPaint library is dependent on GNOME library (such as Glib2 and GEGL).

1. Thus, it is necessary to use MinGW for introducing LibMyPaint to the Windows version of OpenToonz.  #1124 cannot be built with MSVC.

1. #1092 will enable us to build OpenToonz with MinGW. 
Advantages of introducing this PR would be as follows:
   - It enables us to introduce #1124 to the Windows version.
    - It gains consistency between the build environments of Linux and Windows versions. It will make development clearer.
OTOH, disadvantages of introducing this PR would be as follows:
    - Introducing #1124 as-is will require completely replacing MSVC with MinGW. (i.e. Not just adding MinGW as an option.)
    - There may be risks attached to MinGW, as @ideasman42 remarked in this conversation.

Based on the situations above, my opinion is as follows:

I would like to maintain the MSVC build environment, in order to keep development efficiency of contributors who work on Windows (like @turtletooth and me).
IMO Introducing MinGW is fine *ONLY IF IT IS OPTIONAL*. It shuld not be a risk for other development environments.
For the above reason, I can't agree with merging #1092 or #1124 to the master branch as it now stands.

I believe that introducing MyPaint brushes to OpenToonz will be remarkable improvement. (Again, Thank you so much!) I think the best way for introducing #1124 would be; modify LibMyPaint to be compatible with MSVC build, by removing dependency to GNOME libraries (and replacing with other libraries, maybe with Qt)... At least, it should be still compilable with MSVC by excluding MyPaint feature when building with MSVC. Capability for OSX build should be taken into account too.

Thank you for your kind understanding and cooperation.

Meanwhile, I will ask my acquaintance working at Japanese animation studios to try & review OpenToonz+LibMyPaint. Again&again, Thank you very much for the great feature!

morevnaproject:
@shun-iwasawa Thank you for taking your time on reviewing our requests.

First of all, let me clarify about #1092: We have no intention to push MINGW as replacement to MSVC build toolchain. This pull request introduces support for MINGW toolchain exactly as an **optional** way to build OpenToonz for Windows.

> LibMyPaint library is dependent on GNOME libraries (such as Glib2 and GEGL)

> I think the best way for introducing #1124 would be; modify LibMyPaint to be compatible with MSVC build, by removing dependency to GNOME libraries (and replacing with other libraries, maybe with Qt)

Actually, GEGL is an optional dependency of libmypaint. We do not use it in our build of OpenToonz.
Also libmypaint can be compiled without Glib and GObject-Introspection (see - https://github.com/mypaint/libmypaint#dependencies). But we didn't tested that yet.
The only dependency remaining is json-c.

So, I am proposing the following:

1. Merge this pull request (#1092) to add optional MINGW support to OpenToonz. This will allow us to  continue developing other features that currently in progress.
2. We can ship a MSVC-compatible version of libmypaint.
3. We will add necessary code for MSVC to #1124.

> Capability for OSX build should be taken into account too.

As I can see from https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_build_macosx.md, the building on OSX is pretty much straightforward. It uses HomeBrew to install dependent libraries, which I am familiar with. The easiest way would be to introduce a formula for libmypaint here. Still, our access to OSX platform is limited, so it would be nice if we get feedback/testing from someone who maintains OSX pacakge of OpenToonz.

turtletooth:
> So, I am proposing the following:
> 
> Merge this pull request (#1092) to add optional MINGW support to OpenToonz. This will allow us to continue developing other features that currently in progress.
> We can ship a MSVC-compatible version of libmypaint.
> We will add necessary code for MSVC to #1124.

I agree with @shun-iwasawa that MSVC compatibility should not be lost and I am glad to hear your proposal.  I also look forward to a MSVC compatible version of libmypaint as I don't want to lose mypaint functionality if I don't build with mingw.  I believe that would also create confusion - if some versions of OpenToonz support MyPaint and some don't.

shun-iwasawa:
@morevnaproject Thank you for your proposal! Then I'll review in detail this PR next week. Thank you for your patience.

morevnaproject:
> I will ask my acquaintance working at Japanese animation studios to try & review OpenToonz+LibMyPaint

@shun-iwasawa You might want to forward this information regarding our latest changes to MyPaint functionality in OpenToonz - https://morevnaproject.org/2017/04/09/open-toonz-builds-updated/

blackwarthog:
I've fixed bugs and checked build for MSVC, Linux and MinGW (linux).
Also i've splitted this PR to three new:

### #1148 preprocessor
Split _WIN32 preprocessor branches to _MCS_VER (code specific for MSVC) and _WIN32 (code specific for Windows)

### #1149 bugfixes
Fix bugs which MinGW sensitive to.

### #1150 cmake
CMake rules to allow build in MinGW.

RNavega:
Sorry for the noob question.
Will this PR allow people to build OpenToonz on Windows, using MinGW or MinGW-w64 instead of VC++?

morevnaproject:
@RNavega Exactly.

shun-iwasawa:
Merged split PRs. Closing.

morevnaproject:
@shun-iwasawa Can you please take a look at this one PR also? - #1118 
Without it the MinGW version of OT will not save files correctly.

PR #1103, opened 2017-03-15
Cleaned up titlebar icons

konero:
This adds an alpha-channel that was missing to some of the icons and I cropped off their corners for improved theme compatibility (see below).

Sorry, I know it's a tedious thing!

fix_def

crop1

crop2


turtletooth:
LGTM.  Thanks @konero for the attention to the little details.

PR #1106, opened 2017-03-16
Fix tconverter crash

shun-iwasawa:
TO REPRODUCE THE CRASH:  

1. Prepare some source image ( say `srcimage.tif` )

1. Run `tconverter.exe srcimage.tif dstimage.png`.

This PR fixes the crash by adding null pointer check.
The `tconverter.exe` is still not well-functional, but it's better to make it available.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1109, opened 2017-03-18
Astral Theme (orphaned)

konero:
OK, I'm closing this one. I spoke with a buddy who uses Linux/Mac and had him look through the source, seems the way Toonz handles fonts is a bit wonky and there is a typo in main.cpp "Hervetica".

For this and other reasons (I got a ton of feedback, thanks!) I'm refactoring and changing the entire cascade. So I'll open a new PR for that (apologies for orphan) because this would be really bad to review otherwise.


RodneyBaker:
Very nice.  Works quite well (on a PC with Win10 here). 

Thanks!  :)

artisteacher:
After a quick look (on a Mac), the theme works well so overall but there are some details that could be improved in terms of size, spacing and/or clipping.

The start-up pop-up has clipped text and like most of the theme, the text margins look off center vertically.
screen shot 2017-03-18 at 3 29 51 pm
At the bottom of the style editor, the level information text could use some padding on the top.
screen shot 2017-03-18 at 4 18 18 pm
If there should be checkboxes here, they are missing.
screen shot 2017-03-18 at 3 41 23 pm
I really like the grip handles - though on my screen the contrast could be increased slightly.
screen shot 2017-03-18 at 3 51 37 pm
All of the lock icon's are minuscule.
screen shot 2017-03-18 at 4 06 31 pm
screen shot 2017-03-18 at 3 35 21 pm
screen shot 2017-03-18 at 4 11 45 pm
The interpolation area of the function editor needs a min-height set to the text (this is an existing issue in the other themes).
screen shot 2017-03-18 at 4 22 22 pm

If I run into anything else, I'll let you know.

konero:
@artisteacher Thanks for the feedback! These will be tricky to solve since I don't have a Mac at hand, but I'll give it a go nonetheless. I'd be most grateful for your continued trial and feedback.

**Minuscule Icons & Style Editor Padding**

Not sure what is happening with those, I used the container method by setting a container a width and height and assumed the SVG would scale the space but it isn't happening on Mac it seems. I'll swap the SVG files over to PNG which should solve it. Could you give this version a try? It should hopefully solve the issue with Style Editor padding too.

[Astral_FIX-Icons.zip](https://github.com/opentoonz/opentoonz/files/852846/Astral_FIX-Icons.zip)

**Global Font Clipping**

TBH I knew there would be an issue with the font, there always is! This is shooting blind but we can try a few things here. Below are various versions of the theme but with different font settings, could you be so kind as to try them all and send more screenshots back this way? There are no icons with these, this is purely for the text issue.

[MacFontFixes.zip](https://github.com/opentoonz/opentoonz/files/852841/MacFontFixes.zip)

**Function Editor Segment Viewer**

Unfortunately that isn't something that can be solved with QSS. I tried but the treeview will overlap the segment box meaning the buttons won't be accessible. I'll keep an eye on it though. I did spot a bug though in your image with the scrollbars (aesthetic).

**Context Menu Checkbox**

Apologies, this applies only to Windows (and Linux?) since on Mac the native menu bar is used.

astral_qmenu_checkbox

adamearle:
WOW that looks totally class. Big step in the right direction.
    Really really well done. I have shown this to people in the studio
    and people are nuts about it.  The only thing that i heard was "it
    would be be good if you could change colours like blender or zbrush"

    Awesome work

    Adam Earle

konero:
@adamearle Thanks!

I've closed this PR and made a new PR (Minimalist Theme) since I had to do a line-by-line refactor for better Linux support, sorry about that.

In regards to the ZBrush suggestion this could be possible with a simple string replacement and variable. We'll see where the theme goes and I'll submit an issue for it. 👍

PR #1118, opened 2017-03-23
Fix toUpper/toLower functions

blackwarthog:
Standard forbids modifying of data by string::c_str() pointer, in some
case it's may affect other strings.


turtletooth:
LGTM.  Thank you @blackwarthog

Issue #1127, opened 2017-04-02
Mac Version of tfarm components path hard coded version 1.0


damies13:
Mac Version of tfarm components references a path hard coded version 1.0

./OpenToonz_1.0.app/Contents/Resources/configfarmroot.txt

for version 1.1 should be:
./OpenToonz_1.1.app/Contents/Resources/configfarmroot.txt

damies13:
Not sure how to change Assignees? please assign to me

damies13:
Pull Request #1140 provides a fix for this issue;

The basic logic (psudo code) is now:

if windows:
	set registrypath = "SOFTWARE\<AppName>\”<AppName>\<AppVersion>\FARMROOT"
	read regisrty
else:
	if MacOSX:
		set filepath = “./<AppName>_<AppVersion>.app/Contents/Resources/configfarmroot.txt"
	else:
	      set filepath = "/etc/<AppName>/opentoonz.conf"
	end if
	read file
end if

where <AppName> is currently OpenToonz and <AppVersion> is currently 1.1

Issue #1126, opened 2017-04-02
Windows Version of tfarm components Registry path hard coded version 1.0


damies13:
Windows Version of tfarm components uses Registry path that is hard coded version 1.0 however the registry location no longer exists as version 1.1 of OpenToonz installs the registry keys corectly into version 1.1 registry path.

SOFTWARE\OpenToonz\OpenToonz\1.0\FARMROOT 

SOFTWARE\OpenToonz\OpenToonz\1.1\FARMROOT

damies13:
Not sure how to change Assignees? please assign to me

damies13:
Pull Request #1140 provides a fix for this issue;

The basic logic (psudo code) is now:

if windows:
	set registrypath = "SOFTWARE\<AppName>\”<AppName>\<AppVersion>\FARMROOT"
	read regisrty
else:
	if MacOSX:
		set filepath = “./<AppName>_<AppVersion>.app/Contents/Resources/configfarmroot.txt"
	else:
	      set filepath = "/etc/<AppName>/opentoonz.conf"
	end if
	read file
end if

where <AppName> is currently OpenToonz and <AppVersion> is currently 1.1

PR #1128, opened 2017-04-02
Version in FARMROOT path Issue #1126 & Issue #1127

damies13:
Windows Version of tfarm components Registry path hard coded version 1.0
#1126 opened 11 minutes ago by damies13


damies13:
I intended for
- Issue #1126 Windows registry path hard coded version to 1.0  …			23fd327
to be in this pull request, but I didn't intend for 
- Issue #1127 Mac path hard coded version 1.0  …			e4eed9f
- Use QString for Issue #1126 & #1127  …			3a5845f
to be in this pull request.

I'm kinda new at this, can I get the 2nd and 3rd changes removed from this pull request? Also can someone tell me what I did wrong, so I don't make this mistake again?

yamisama:
I don't have much experience with GIT/GitHub either, but from what I understand, you directly submitted the master branch of your forked repository as a pull request. Normally, you'd create a branch for each feature you develop in your forked repository and then submit that branch as a pull request.

Also, from my understanding, once you submit a pull request, any commits you make to that branch will automatically become part of that pull request, hence your later commits to your master branch automatically became part of the pull request as well. Instead, you should probably have created separate new branches for each of the other issues and then made those into separate pull requests later.

As for how to fix it, I should probably leave that to somebody more competent with GIT than me to answer.

toonz/sources/toonzfarm/tfarmserver/tfarmserver.cpp
```diff
@@ -60,18 +59,24 @@ class FarmServer;
...
+const char *applicationVersion  = "1.1";
+const char *applicationRevision = "2";
+
 //--------------------------------------------------------------------
 TFilePath getGlobalRoot() {
   TFilePath rootDir;
 
 #ifdef _WIN32
-  TFilePath name(L"SOFTWARE\\OpenToonz\\OpenToonz\\1.0\\FARMROOT");
+  QString regpath = L"SOFTWARE\\OpenToonz\\OpenToonz\\" + applicationVersion + "\\FARMROOT";
```
yamisama: You are adding multiple values of type _const char*_ together here. In C++, this does not concatenate the actual strings, it adds the values of the pointers together, giving you a nonsense result that may crash the application when accessed. I'm really surprised that this code even works on your system.

One of the easiest ways to avoid this is by explicitly constructing a QString from your literal, at least the first one in the chain. As long as one operand is a QString or std::(w)string, the + operator will work since it would then call the corresponding overloaded operator on the QString/std::(w)string class. But if both operands are pointers, you're just adding up the address in memory.

damies13: @yamisama I just figured that out, with a little google help. But thanks for replying. It builds on my machine now, hopefully it will here too, we'll see in an hour.

damies13:
This error doesn't seem to be related to my changes:


Build started
git clone -q --depth=1 https://github.com/opentoonz/opentoonz.git C:\projects\opentoonz
Downloading thirdparty/LibJPEG/jpeg-9/lib/LibJPEG-9_2013.lib (2.38 MB)
Error downloading object: thirdparty/LibJPEG/jpeg-9/lib/LibJPEG-9_2013.lib (98a2e2d1ebb2656551cfa940f0c6c7370b2dab3660ba9d305c0d2ac2b21638f5)
Errors logged to C:\projects\opentoonz\.git\lfs\objects\logs\20170404T120312.7568573.log
Use `git lfs logs last` to view the log.
error: external filter 'git-lfs filter-process' failed
fatal: thirdparty/LibJPEG/jpeg-9/lib/LibJPEG-9_2013.lib: smudge filter lfs failed
warning: Clone succeeded, but checkout failed.
You can inspect what was checked out with 'git status'
and retry the checkout with 'git checkout -f HEAD'
Command exited with code 128

damies13:
After spending 2 days creating a Windows 7 VM with all the pre requisites for compiling OpenToonz (windows is not my normal development environment) I got it to compile in windows and have checked in the the change, however the windows build server seems to have an issue with cloning the source, so I don't know how to proceed from here.

toonz/sources/toonzfarm/tfarmcontroller/tfarmcontroller.cpp
```diff
@@ -90,14 +97,16 @@ TFilePath getLocalRoot() {
   TFilePath lroot;
 
 #ifdef _WIN32
-  TFilePath name(L"SOFTWARE\\OpenToonz\\OpenToonz\\1.0\\TOONZROOT");
+  QString regpath = QString::fromStdString("SOFTWARE\\OpenToonz\\OpenToonz\\" + std::string(applicationVersion) + "\\FARMROOT");
```
yamisama: As far as I can tell, TFilePath already has a constructor that takes an STL std::string, so I don't think there is a need to convert to a QString.

This should be all you need:

```
const std::string regpath = "SOFTWARE\\OpenToonz\\OpenToonz\\" + std::string(applicationVersion) + "\\FARMROOT";
const TFilePath name = regpath;
```

Also, it might be useful to replace the hard-coded "OpenToonz" in the strings with the _applicationName_ variable, since having it declared above is kind of misleading if changing it would have no effect. In that case you could do this:

```
const std::string regpath = "SOFTWARE\\" + std::string(applicationName) + "\\" + std::string(applicationName) + "\\" + std::string(applicationVersion) + "\\FARMROOT";
const TFilePath name = regpath;
```

If you are inclined to learn about std::stringstream or format strings, you could go into that if you prefer, but simple concatenation like in my example should work just fine in our case.

damies13: @yamisama, thanks for the feed back and reminding me about the applicationName, I had intended to to that but for got while trying to get it to work.

The original string was prefixed with a 'L' which if I understand correctly is a std::wstring, but in this project it seems QString is preferred. I don't know why the original string was not a standard string and I didn't want to break anything which is why I used QString. It's probably safest to leave it this way unless we can confirm nothing else will be broken.

I'll add in the applicationName change now, any idea how to get the the build server to clone successfully?

yamisama: The L prefix just means it's a unicode string with 16 bit characters. It's also what modern Windows and TFilePath use internally. Since neither of these strings contain anything but ASCII/ANSI characters (i.e. no Japanese characters for instance) since the components you are building them from are _const char*_ literals without any UTF8 characters, I don't think there is a need to worry about Unicode here.

QString is what Qt uses internally and thus is used for strings that need to be used with the UI. TFilePath uses std::wstring internally. If you supply a QString, it will just convert it to std::wstring anyway. I think not depending on Qt classes where there is no real gain from them is a good idea since it keeps code more portable.

As for the build and version control system, I'm afraid I can't be of much help, sorry.

damies13: @yamisama (Peter?) just found your response,

Thanks for this I actually learned a lot from it, so it was quite helpful. I am new to c++ so still trying to figure things out, as I mostly code in ancient (ansii 99) C or python, though I have worked in c# and a bunch of other languages as well in the past, so mostly it's the C++ specific bits i need to get familiar with.

The build system sorted issue seems to be sorted and the version control stuff i'll figure out eventually even if just through trial and error.

Thanks for your help.

Dave.

damies13:
OK I think this is complete, I have nothing more to add to this pull request and last build was successful :)

I am closing it so that I can create a new pull request for my next fix.

Dave.

PR #1139, opened 2017-04-12
Fix level placement when the option "Show Appendix to the Frame Number" is activated

shun-iwasawa:
The preference option  "Show "ABC" Appendix to the Frame Number in Xsheet Cell" was introduced for inserting drawings when pencil testing using the camera capture feature.

If this option is activated, ones place digit of frame numbers are converted to "ABC" appendix (10 > 1,  11>1A, 12>1B and so on) and captured image will normally increment frame number at interval of 10.
  
When placing such level with "Load Level" command, with the option "Arrangement in Xsheet>step" is set to "Auto" in the popup, it will cause unwanted spacing between frames in Xsheet:

level_placement

This PR will fix such problem, by setting the step to "1" in such case.


shun-iwasawa:
Jenkins

turtletooth:
LGTM Thanks.

Issue #1129, opened 2017-04-07
ToonzFarm version number discrepancies


damies13:
I have noticed some discrepancies in the version number in the ToonzFarm components in release 1.1.2, as shown below:

`./tfarmcontroller -console
Fri Apr 7 23:55:40 2017 INF:
ToonzFarm Controller 1.0

Fri Apr 7 23:55:40 2017 INF:
Starting Controller on port 1

`

`./tfarmcontroller -release
Toonz 7.1
`

`./tfarmserver -console
Fri Apr 7 23:55:27 2017 INF:
ToonzFarm Server 1.0
`

`./tfarmserver -release
Toonz 7.1
`

damies13:
I will attempt to fix this....

damies13:
I have fix for this ready to merge in, when I figure out how to do a seperate merge request from th eone I have currently open.

Dave.

damies13:
Pull Request #1140 provides a fix for this issue. Example outputs below:

> ./tfarmcontroller -help
usage: ./tfarmcontroller -help
       ./tfarmcontroller -release
       ./tfarmcontroller -version
       ./tfarmcontroller [ -console ] [ -install name ] [ -remove ]
  -help
       Print this help page
  -release
       Print the current Toonz version
  -version
       Print the current Toonz version
  -console
       Run as console app
  -install name
       Install service as 'name'
  -remove
       Remove service

> /tfarmcontroller -release
OpenToonz  v1.1.2

> ./tfarmcontroller -version
OpenToonz  v1.1.2

> ./tfarmcontroller -console
Wed Apr 12 23:54:50 2017 INF:
OpenToonz Farm Controller v1.1.2
Wed Apr 12 23:54:51 2017 INF:
Starting Controller on port 1

> ./tfarmserver -help       
usage: ./tfarmserver -help
       ./tfarmserver -release
       ./tfarmserver -version
       ./tfarmserver [ -console ] [ -install name ] [ -remove ]
  -help
       Print this help page
  -release
       Print the current Toonz version
  -version
       Print the current Toonz version
  -console
       Run as console app
  -install name
       Install service as 'name'
  -remove
       Remove service

> ./tfarmserver -release
OpenToonz  v1.1.2

> ./tfarmserver -version
OpenToonz  v1.1.2

> ./tfarmserver -console
Wed Apr 12 23:55:31 2017 INF:
OpenToonz Farm Server v1.1.2

Issue #1125, opened 2017-04-02
Linux Version of tfarm components  uses MAC path for configfarmroot.txt


damies13:
Linux Version of tfarm components uses MAC path for configfarmroot.txt

damies13:
Not sure how to change Assignees? please assign to me

ideasman42:
Not sure if the macOS version even works since this points to OpenToonz `1.0` when the version is now 1.1.

I think this should be using one of the functions from `toonz/sources/toonzlib/toonzfolders.cpp`

damies13:
I have fix for this ready to merge in, when I figure out how to do a seperate merge request from th eone I have currently open.

Dave.

damies13:
@ideasman42, the MacOSX version is not currently being shipped in the OpenToonz_1.1.app file, Issues #1127 and #1130 will resolve that, I have fixes ready to merge in for both, so it's coming.

damies13:
Pull Request #1140 provides a fix for this issue;

The basic logic (psudo code) is now:

if windows:
	set registrypath = "SOFTWARE\<AppName>\”<AppName>\<AppVersion>\FARMROOT"
	read regisrty
else:
	if MacOSX:
		set filepath = “./<AppName>_<AppVersion>.app/Contents/Resources/configfarmroot.txt"
	else:
	      set filepath = "/etc/<AppName>/opentoonz.conf"
	end if
	read file
end if

where <AppName> is currently OpenToonz and <AppVersion> is currently 1.1

PR #1140, opened 2017-04-12
Various TFarm fixes

damies13:
This PR contains fixes for Issues #1125 #1126 #1127 & #1129


damies13:
So close .....

> [ 93%] [32mBuilding CXX object toonz/CMakeFiles/OpenToonz_1.1.dir/moc_previewfxmanager.cpp.o[0m

> [ 93%] [32mBuilding CXX object toonz/CMakeFiles/OpenToonz_1.1.dir/moc_projectpopup.cpp.o[0m

> [ 93%] [32mBuilding CXX object toonz/CMakeFiles/OpenToonz_1.1.dir/moc_psdsettingspopup.cpp.o[0m

> [ 93%] [32mBuilding CXX object toonz/CMakeFiles/OpenToonz_1.1.dir/moc_renumberpopup.cpp.o[0m


> The job exceeded the maximum time limit for jobs, and has been terminated.

shun-iwasawa:
@damies13 Thank you very much for the fix.
The class `TVER::ToonzVersion` seems to overlap considerably with `TEnv`. IMO it would better to unify.

damies13:
@shun-iwasawa Thanks for your comment, you actually helped me to understand how the version works in the GUI app, I was trying to understand this before but could only see it being defined in main.cpp.

I agree that the functions in "TVER" and "TEnv" serve the same purpose, but actually they are kind of doing the opposite things:

- "TEnv" set the app name and version number in main.cpp then use TEnv functions to share this information with the other cpp files.

- "TVER" set the app name and version number in a central location that can be used by "main.cpp" of all the command line utilities.

My concern with the "TEnv" approach is if I define the app name and version number (especially the version numbers) in each "main.cpp" of all the command line utilities then it will drift from the main app and get forgotten about, leaving us not much better off than the current situation, so my opinion is this should be centralised.

As you have the most commits on this project I gather you are one of the senior developers, so I am happy to defer to you and the other senior developers (I'm not sure about the hierarchy or decision making process of this project). I can adjust my code changes to define the app name and version number in each "main.cpp" and use TEnv as is done in the GUI if this is preferred, if you think its better to centralise it then I'm happy to work with you on how best to do this as well.

Dave.

shun-iwasawa:
@damies13 
I totally agree with you that the version number should be centralized and be shared among the command line utilities.
I just concern about that with your modification we will have three types of the way for setting the version number, I think it is quite troublesome to maintain;

- Set with `TEnv` (OpenToonz main GUI application)
- Set in `TVER` (`tfarmserver` and `tfarmcontroller`)
- Set in "main.cpp" (Other command line utilities; `tcomposer`, `tcleanupper`, and `tconverter`)

I think it would better to unify them (with the centralized way as you proposed) .
Could you make other utility softwares (`tcomposer`, `tcleanupper`, and `tconverter`) to refer the version number in the same way as `tfarm` ? Converting the version number control in OpenToonz main app to `TVER` might be a hard work, so it can be separate for the future work.

NOTE: Please make sure that your modifications are `clang-format` ted. (Please see [CONTRIBUTING.md](https://github.com/opentoonz/opentoonz/blob/master/CONTRIBUTING.md)  )

damies13:
@shun-iwasawa 

Thanks for the clarification,

Can I suggest to allow the Tfarm fixes go though we create a new issue to centralise the version number into one place? I am happy for you to assign this issue to me. If you would like I could define the version number within TEnv, and then use the functions defined in TEnv and only bring across functions from TVER if there is no equivalent already in TEnv, this way I can remove TVER and it should be reasonable simple to also update the OpenToonz main app to use the centralised version numbers. I think I can do this all at the same time, but for clarity I think it might be better to keep it seperate from Issues already addressed here.

Regarding the clang-format ing, thanks for pulling me up on this I admit I missed that, I will try not to repeat that mistake. I have not used clang-format before so I may need guidance if I make mistakes using it.

Dave.

shun-iwasawa:
>Can I suggest to allow the Tfarm fixes go though we create a new issue to centralise the version number into one place?

OK, so here are 3 points I would like to ask you before merging this:
- Please add `toonz\sources\include\tversion.h` to `toonz\sources\tnzcore\CMakeLists.txt` like as follows.
```
--------------------- toonz/sources/tnzcore/CMakeLists.txt ---------------------
index 8b605b7..76bf389 100644
@@ -120,6 +120,7 @@ set(HEADERS ${MOC_HEADERS}
     ../include/tenv.h
     ../include/tmeshimage.h
     ../include/tgldisplaylistsmanager.h
+    ../include/tversion.h
 ) 
```

- Please update the `const float applicationRevision` in `toonz\sources\include\tversion.h` to **`3`**. (For now we are releasing v1.1.3... So sorry for my late response on this PR!)

* Please apply clang-format.
  * Install clang-format. Then follow the instruction written in [CONTRIBUTING.md](https://github.com/opentoonz/opentoonz/blob/master/CONTRIBUTING.md)
  > * apply `clang-format` with `toonz/sources/.clang-format.`
  >   * `cd toonz/sources`
  >   * `./beautification.sh` or `beautification.bat`.

Thanks!

damies13:
@shun-iwasawa done as requested :)

Let me know if I didn't apply clang-format correctly and i'll try to fix it.

Dave.

damies13:
I opened issue #1368 to centralise the version number into one place.

Dave.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. Thank you @damies13 for the fix and opening the new issue!

Issue #1130, opened 2017-04-07
tfarmcontroller & tfarmserver on Mac OSX


damies13:
I found out what happened to tfarmcontroller & tfarmserver on Mac OSX, these files are compiled into folders in the toonzfarm folder of the build folder, but the core app is compiled into a OpenToonz_1.1.app folder in the toonz folder also in the build folder.

A solution to include these components into the OSX package will need to be determined and the build scripts updated to do this.

damies13:
I'm not sure how to do this at the moment, if anyone else knows please add comments here.

If no one else fixes this issue I will attempt it after I have fixed some of the other issues I have raised.

damies13:
Line 362 of: https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonzfarm/tfarm/tfarmtask.cpp 

Tells me the answer, they should be put into /Contents/MacOS/ of OpenToonz_1.1.app which is what I suspected the answer would be.

I also noticed that this is hard coded as Toonz 7.1.app which will also need to be fixed at the same time.

damies13:
Great news!, I have a fix for this, once I know how to create a new pull request i'll merge it.

Best bit was it was only a small change to 1 file.

artisteacher:
Thanks so much for working on this!

damies13:
Pull Request #1141 provides a fix for this issue.

After setting CMAKE_RUNTIME_OUTPUT_DIRECTORY to the Contents/MacOS/ folder in OpenToonz_1.1.app the following binaries are now included that were previously missing:

-rwxr-xr-x   1 dave  staff   120424 10 Apr 21:44 tcleanup
-rwxr-xr-x   1 dave  staff   126600 10 Apr 21:44 tcomposer
-rwxr-xr-x   1 dave  staff    98888 10 Apr 21:44 tconverter
-rwxr-xr-x   1 dave  staff   158068 10 Apr 21:44 tfarmcontroller
-rwxr-xr-x   1 dave  staff    99208 10 Apr 21:44 tfarmserver

It turned out to be a very simple fix once I learned a little about how CMake works, and as a bonus it was a good learning experience for me too.

PR #1141, opened 2017-04-12
Issue #1130 - tfarm binaries into .app folder

damies13:
set CMAKE_RUNTIME_OUTPUT_DIRECTORY path for OSX so that when
add_subdirectory() calls add_executable(), the binaries know where they
need to be written so they are bundled into the mac .app folder


damies13:
The job exceeded the maximum time limit for jobs, and has been terminated.

damies13:
Issue #1130 - tfarm binaries into .app folder

shun-iwasawa:
@damies13  Just a quick note that I failed to build this PR properly. Is this needed to include another fix in advance, such as #1140 ?

damies13:
@shun-iwasawa I don't believe this PR has any dependancy, It seemed to build successfully with the automated builds. The only thing this PR does is change the CMAKE_RUNTIME_OUTPUT_DIRECTORY variable in the makelist file so that the command line utilities (including the tfarm binaries) get created inside the .app contents folder for the mac build. without this the get created but because they are outside the .app folder the get missed in the packaging of the mac version.

I realise that the command line utilities could just be added to the package separately by scripts, but putting them inside the .app folder is more inline with how mac applications are usually packaged, so I think this is probably best.

Out of curiosity what did the build fail on?

Dave.

shun-iwasawa:
Jenkins

shun-iwasawa:
@damies13 
I'm so sorry for my too late response and thank you so much for tackling with tfarm+OSX issue.
I noticed that in both Jen.kins and my OSX environment the build fails just after the message like as follows: 

> \> Linking CXX executable OpenToonz_1.1.app/Contents/MacOS/OpenToonz_1.2.app/Contents/MacOS/OpenToonz_1.2

As you can see, a string `Contents/MacOS/OpenToonz_1.2` is doubled and tried to link the executable with wrong path.

Based on this PR I'll post another PR with small fix, which copies executables `tcomposer` and `tcleanup` into `OpenToonz_1.1.app/Contents/MacOS` after build, just like `lzocompress`.

shun-iwasawa:
#1993 is merged. Thank you @damies13 for this !

PR #1144, opened 2017-04-13
(Hiatus) Support for Haiku

korli:
Patch to enable the build on Haiku, mostly by reusing the LINUX case. 

Runs but is ATM not usable because of OpenGL related bugs in Qt Haiku plugin.
Please consider merging.


ideasman42:
While I'm not against this, Haiku is a *very* niche system. I think it only makes sense to apply the patch if its usable.

Note that platform checks are getting a bit messy, removed SGI checks in #1145 to avoid ifdef-soup :)

korli:
Sure, platform checks are hard to get right and clean. I'd like to replace (defined(LINUX) || defined(HAIKU)) with something more sensible where it matters, hence minimizing our own patch for Haiku, and possibly other systems.

ideasman42:
@korli for proproc blocks that deal with each platform we could check as follows:

```
#if defined(WIN32)
...
#elif defined(MACOSX)

#else /* Generic Unix */

#endif
```

This way Linux/BSD... etc can mostly share code which is fine since they are often similar where OpenToonz is concerned. And for exceptions fallback to what we have already (individual platform checks).

Before making changes like this, better maintainers approve.

RodneyBaker:
Initiated a review of open PRs to determine what is still active and desired.
This one is two years old as of yesterday with no activity since then.

Any updates?

korli:
Well, we can update this PR. About the platform checks, any thought on the GENERIC_UNIX idea?

jpturcotte:
@korli I know Haiku is now light-years beyond where it was two years ago, but what about those Qt and OpenGL issues? Is OT even usable without HW acceleration? What about graphic tablet support?

korli:
> @korli I know Haiku is now light-years beyond where it was two years ago, but what about those Qt and OpenGL issues? Is OT even usable without HW acceleration? What about graphic tablet support?
@jpturcotte Qt and OpenGL issues still stand, AFAIK these make OT not usable. Graphic tablets are supported OS-wise, I dunno in context of OT.

jpturcotte:
@korli If you are willing to maintain it, I would actually be pretty stoked, but only if it becomes a showcase app like Blender. I know you put a lot of effort through your stack to get some of those running, such as LibreOffice. However, seeing as both staff are already stretched thin, I do not see the benefit if the point is just adding another package to the HaikuDepot.

korli:
Don't understand your point. OT patches for Haiku should be minimal and no one invested time in it since at least a year.

jpturcotte:
Resolve conflicts, and see how it goes, but seeing how there is still no BSD support, and you admit to OT's lack of usability on Haiku right now, I am not optimistic. I hope you continue your great progress, though!

RodneyBaker:
Closing as an abandoned effort.

PR #1145, opened 2017-04-14
(Hiatus) Cleanup: remove SGI checks

ideasman42:
Submitting this PR since there are some proposals to introduce support for other environments, eg: #1144 
So I think it makes sense to remove SGI support to keep things maintainable.

----

Most instances simply remove, with a few exceptions.

- Keep reference code that isn't implemented elsewhere.
- Keep for mostly unmaintained tfarm (temp file usage).

Also replace some `__sgi` checks with `!TNZ_LITTLE_ENDIAN`.


RodneyBaker:
The conflicts are the only thing that appear to be holding this PR back from being merged.
Although tagged as 'Hiatus', the PR itself is not so much on Hiatus as just forgotten and therefore off of everyone's radar.

RodneyBaker:
Closing as abandoned.

PR #1149, opened 2017-04-15
 Allow to cross compile for Windows with Linux/Mingw (part 2 of 3, bugfixes)

blackwarthog:
see PR #1092

Fix bugs which MinGW sensitive to.

### DIFF
--- a/toonz/sources/include/tfx.h
+++ b/toonz/sources/include/tfx.h
@@ -210,16 +210,16 @@ public:
   double m_end;
 };
 
-inline DVAPI TFxTimeRegion operator+(const TFxTimeRegion &tr1,
-                                     const TFxTimeRegion &tr2) {
+inline TFxTimeRegion operator+(const TFxTimeRegion &tr1,
+                               const TFxTimeRegion &tr2) {
   return TFxTimeRegion(tr1) += tr2;
 }
 
-inline DVAPI TFxTimeRegion operator+(const TFxTimeRegion &tr1, double shift) {
+inline TFxTimeRegion operator+(const TFxTimeRegion &tr1, double shift) {
   return TFxTimeRegion(tr1) += shift;
 }
 
-inline DVAPI TFxTimeRegion operator-(const TFxTimeRegion &tr1, double shift) {
+inline TFxTimeRegion operator-(const TFxTimeRegion &tr1, double shift) {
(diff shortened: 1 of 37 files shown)


shun-iwasawa:
jenkins

toonz/sources/include/tfx.h
```diff
@@ -210,16 +210,16 @@ class DVAPI TFxTimeRegion {
...
 
-inline DVAPI TFxTimeRegion operator+(const TFxTimeRegion &tr1,
-                                     const TFxTimeRegion &tr2) {
+inline TFxTimeRegion operator+(const TFxTimeRegion &tr1,
+                               const TFxTimeRegion &tr2) {
   return TFxTimeRegion(tr1) += tr2;
 }
 
-inline DVAPI TFxTimeRegion operator+(const TFxTimeRegion &tr1, double shift) {
+inline TFxTimeRegion operator+(const TFxTimeRegion &tr1, double shift) {
```
shun-iwasawa: You are removing `DVAPI` from many parts. It seems that `DVAPI` is actually defined as `__declspec(dllexport)` in [tcommon.h](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/include/tcommon.h#L145) as follows:

> 145  #ifdef _WIN32
> 146  #define DV_EXPORT_API __declspec(dllexport)
> 147  #define DV_IMPORT_API __declspec(dllimport)
> 148  #define DV_EXPORT_VAR __declspec(dllexport)
> 149  #define DV_IMPORT_VAR __declspec(dllimport)
> 150  #else
> 151  #define DV_EXPORT_API
> 152  #define DV_IMPORT_API
> 153  #define DV_EXPORT_VAR
> 154  #define DV_IMPORT_VAR
> 155  #endif

Since `__declspec` is used only in MSVC, how about just replacing the line 145 above by `#ifdef _MSC_VER` instead of removing `DVAPI`s ?

blackwarthog: I've removed DVAPI for inline function only. There is no reason to use dllexport/dllimport for inline functions. Because they are inline.

shun-iwasawa: Oh, I got it!

shun-iwasawa:
LGTM

PR #1152, opened 2017-04-15
fix #1151 deleted canvas.h

### LINKED ISSUE #1151: toonz/sources/include/tw/canvas.h is completely useless, delete it?
and what if there are other files like it?
This is the kind of stuff that confuses me when I'm contributing to projects, when I see things and try to understand them and they turn out to be deprecated or was part of something that doesn't exist now.

Shouldn't we delete canvas.h and other files like it?

lordadamson:


shun-iwasawa:
Thanks @lordadamson for refactor, it seems that most sources in `sources/include/tw/` are legacy, used in Toonz Harlequin v5 which was not Qt-based.

lordadamson:
okay great. maybe I'll look more into it and see if I can clean more stuff up

PR #1154, opened 2017-04-16
OpenToonz Farm Documentation

damies13:
First attempt at documenting the usage of OpenToonz Farm,
hopefully I have covered everything, and hopefully this helps someone.


turtletooth:
@damies13 I was just looking at your PRs and was wondering if you would mind making the farm root configurable under preferences instead of needing to set a registry key.  That's what I did for #664 so people wouldn't have to open regedit.

damies13:
@turtletooth, thanks for the feedback, actually the Registry keys were not my call, that was existing in OpenToonz, probably brought over from Toonz? 

In my efforts to figure out how to use / get working the ToonzFarm, I identified that the farm services were looking in a different place to where the configuration was being set:
- Windows - I just noticed that when OT 1.1 package (via exe on website) installed the registry key to the 1.1 path but the farm executables look for the registry keys in the 1.0 path which in my case didn't exist. So I just made it use the registry path that was installed, but tried to do this in a way that will minimise the number of places to change in future. As far as I can tell the OT 1.1 gui uses the 1.1 registry paths already.
- Linux - I noticed that on Linux the config file was being looked for in a Mac OSX path, obviously this didn't work, so I changed it to look in the /etc directory (the place that most linux applications / admins would look).
- Mac OSX - Again they were looking for a path that was OpenToonz_1.0.app, but on Mac OSX the app is now OpenToonz_1.1.app and I expect it to change in the future (1.2, 1.3, 2.0, 2.1, etc) So I just made it use the correct OpenToonz_N.n.app name.

Sorry if it wasn't clear, all my PR's to date (except the documentation one) have been about fixing bugs to just get the ToonzFarm working on all platforms with the absolute minimum of changes. My intention to start with was to just get everything working first, then from a "everything working" starting point, start to try and improve things (this is where i'm at now).

I'm not very experienced in GUI programming (server side components are more my area), so I'm not sure how to go about making the farm root configurable under preferences, it may already be? I am still reading through the GUI code trying to understand how it all works, so while I would be happy to make the change you suggested, it might be quicker for someone familiar with the GUI to do It? I will keep this suggestion in mind as I familiarise myself with the GUI code, if I can see where this should be added I will.

I had a look at the GUI and found the following place the FARMROOT is configured in the GUI:
screen shot 2017-06-12 at 8 44 13 pm
Not sure if it writes to the registry though? I'm also not sure where to look either?

damies13:
The Other thing I forgot to mention, I can conceive in the future separating the ToonzFarm install from the GUI install so that people setting up a ToonzFarm don't need to install the full GUI on the farm servers, so in this case we would need to know how to change the settings without the GUI.

Not something for a 1.1.x release though.

turtletooth:
@damies13 Thank you for doing this.  This seems to me like it would be good to put this in the new OpenToonz documentation site: https://github.com/opentoonz/opentoonz_docs

turtletooth:
Closing this as documentation suggestions should now be posted at the new site.

PR #1158, opened 2017-04-17
fix #1156 which causes program to crash when expanding history

### LINKED ISSUE #1156: SEGABRT when trying to expand history when saving/loading anything (scene/level)
steps to reproduce:
1. file >> load scene
1. click on history
1. SEGABRT at `dvdirtreeview.cpp:1634`

OS: fedora 25

lordadamson:
fix #1156 which causes program to crash when expanding history


turtletooth:
Confirmed crash (only crashes if there is no history) and the fix works.

PR #1162, opened 2017-04-18
Change closest field value to 999

shun-iwasawa:
The closest field is one of the unused cleanup settings parameters. Here is the description from Toonz 7.1 Harlequin manual:

> The Closest Field defines the smallest field size you will zoom into the drawing with the camera when compositing the scene. This value is meaningful if it is smaller than the camera field size, as it increases the  final image resolution, preventing zoomed-in images from appearing jagged. For example if the cleanup camera is 768x576 pixels (PAL) with a 12 field size and the closest field is set to 6, the cleaned up drawings will have twice the camera resolution, that is 1536 by 1152 pixels.

Overlooking of this value sometimes caused an unwanted mismatch in image resolution between the cleanup setting and the result. So we decided to hide this parameter from the cleanup settings panel of Toonz Ghibli Edition and to set the value to 999 in order to certainly  skip the resolution adjustment.

In Ghibli Edition it was done by altering the closest field value in each project settings file. This PR will change the default value to 999 in the first place.


turtletooth:
LGTM.  Thanks @shun-iwasawa

PR #1175, opened 2017-04-26
SoapBubbleFxIwa Improvement

shun-iwasawa:
This PR will improve SoapBubbleIwaFx by adding several features as written below.
All of these features were requested by some Japanese animation studio.

Please note in advance that all example images here were rendered with the input images as follows:
soapbubble01

 ## 1. Fit Thickness to Each Bubble option
This option will automatically resize the thickness image in order to fit with the bounding box of each bubble. With this option you don't need to adjust the thickness image according to the size & movement of bubbles.
soapbubble02

## 2. Opacity of Bubble's Center is now adjustable
Added a new parameter "Opacity of Bubble's Center", which allows to have semi-transparent center of the bubble. It is replacement of "Mask Center of the Bubble" checkbox so I set the "Mask Center of the Bubble" parameter as OBSOLETE. If you load the scene saved with older version, the "Mask Center of the Bubble" value will automatically converted to "Opacity of Bubble's Center" parameter in order to get the same rendering result.
soapbubble03

## 3. Loop and Shift the spectrum
Although this feature is not physically correct, it can obtain somewhat nice result...
Added two new parameters, "Loop Spectrum Fade Width" and "Spectrum Shift".
Consider that the bubble spectrum pattern is "looped" outside of the range between "Thickness Min" and "Thickness Max". The "Loop Spectrum Fade Width" specifies the width of fading around the border:

And the "Spectrum Shift" can shift the looped spectrum pattern. By animating this parameter you can easily get the following result:
soapbubble05

## 4. "Thickness" and "Depth" render modes
These render modes will render thickness and depth pattern used for calculating bubble color. 
soapbubble06

Although the new option "Fit Thickness to Each Bubble" is useful, it has some disadvantage; if you render multiple bubbles, the appearance of bubbles becomes very identical since the same thickness images are used in all bubbles.
In order to vary them, you can do some trick as follows: 1. Render the bubble with "Thickness" render mode. 2. Composite it with some noise pattern. 3. Use it as "Thickness" image for ANOTHER SoapBubbleFx.  By doing this, you can obtain "similar, but not identical" appearance in each bubble.
soapbubble07

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1177,6 +1177,7 @@
   <item>"STD_iwa_PNPerspectiveFx.waveHeight"		"Wave Height"</item>
   
   <item>"STD_iwa_SoapBubbleFx"	"SoapBubble Iwa"	</item>
+  <item>"STD_iwa_SoapBubbleFx.renderMode"	"Render Mode"	</item>
   <item>"STD_iwa_SoapBubbleFx.intensity"	"Intensity"	</item>
   <item>"STD_iwa_SoapBubbleFx.refractiveIndex"	"Refractive Index"	</item>
   <item>"STD_iwa_SoapBubbleFx.thickMax"	"Thick Max"	</item>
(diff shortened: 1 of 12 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins again

masafumi-inoue:
LGTM

Issue #1161, opened 2017-04-18
Destructors are noexcept by default in gcc 6


otakuto:
This project use throw in destructor with implicit noexcept(false). However, implicit noexcept of destructor changed noexcept(true) to noexcept(false) at gcc 6.
It's reported from Gentoo's Bugzilla.
https://bugs.gentoo.org/show_bug.cgi?id=612938
https://gcc.gnu.org/gcc-6/porting_to.html

ideasman42:
Please give info on the actions needed to cause this error.

aliceinwire:
@ideasman42 compile with gcc6

ideasman42:
I'm building with GCC 6.3.1, is this only a problem with older GCC versions?

aliceinwire:
@ideasman42 you can find the error output here https://bugs.gentoo.org/show_bug.cgi?id=612938

ideasman42:
The cxxflags must be defining `-Werror=terminate`, opentoonz is a very big codebase with warnings that would be nice to resolve, but don't consider this a bug. Submit a PR to resolve warnings is generally welcome of course.

Closing.

PR #1182, opened 2017-05-07
Don't allow throw() in destructors

Peter-Levine:
Destructors are incapable of deducing whether they are being called in the context of normal destruction or stack unwinding.  Throwing in a destructor can lead to multiple exceptions being thrown, inadvertent process termination, or worse.  Furthermore, even if the destructor is adequately handled without process termination, its object's memory might not be properly deallocated.

For such reasons, in C++11/14/17, destructors default to `noexcept(true)` and all instances of `throw()` become calls to `std::terminate()`.  This however changes the original intent of the code which was to throw an exception to be handled outside of its immediate scope.

Forcing such pre-C++11 throwing behavior in such destructors (`TLevelWriter3gp::~TLevelWriter3gp()` and `TLevelWriterMov::~TLevelWriterMov()`) by explicitly marking them `noexcept(false)` wouldn't be enough since they are virtual and more base destructors aren't allowed more lax exception specification than exist in more derived classes.  Some even derive from QT classes (via multiple inheritance) and would render such an endeavor impossible.

In both offending instances, an exception is thrown based on the state of a stream object instantiated in the destructor.  Either, such a scenario is so serious as to warrant an explicit call to `std::exit()` or `std::terminate()`, or it should be benign enough to be ignored (at least in the context of an object's destruction).

This PR assumes the latter.
It also resolves issues of [compiling with `-Wno-error=terminate`](https://github.com/opentoonz/opentoonz/issues/1161).

### DIFF
--- a/toonz/sources/image/3gp/tiio_3gp_proxy.cpp
+++ b/toonz/sources/image/3gp/tiio_3gp_proxy.cpp
@@ -138,8 +138,6 @@ TLevelWriter3gp::~TLevelWriter3gp() {
   QString res;
 
   stream << (msg << QString("$closeLW3gp") << m_id);
-  if (tipc::readMessage(stream, msg) != "ok")
-    throw TException("Unable to write file");
 }
 
 //------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
LGTM. Thank you @Peter-Levine .

PR #1184, opened 2017-05-09
Set premultiply option and choose level name type on loading PSD layer levels

shun-iwasawa:
This feature was demanded by some Japanese animation studio.
Modified the PSD file loading feature in two points as follows:

1.   Each layer level will be set the premultiply option on loading in order to obtain the same appearance as the original.

1.  When loading PSD files as "Columns", now it is available to select the level name type from two options; `FileName#LayerName` (an existing naming rule) and `LayerName` (a new option). the second option will shorten the level name so that it can be shown in the xsheet cell.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1242,10 +1242,11 @@ bool IoCmd::saveSceneIfNeeded(QString msg) {
 
     //--- If both the level and scene is clean, then open the quit confirmation
     // dialog
-    //if (!isLevelOrSceneIsDirty && msg == "Quit") {
+    // if (!isLevelOrSceneIsDirty && msg == "Quit") {
     //  QString question("Are you sure ?");
     //  int ret =
-    //      DVGui::MsgBox(question, QObject::tr("OK"), QObject::tr("Cancel"), 0);
+    //      DVGui::MsgBox(question, QObject::tr("OK"), QObject::tr("Cancel"),
+    //      0);
     //  if (ret == 0 || ret == 2) {
     //    // cancel (or closed message box window)
     //    return false;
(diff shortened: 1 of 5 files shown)


shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa This is good.  How hard would it be to set premultiplied for png files too?  This has been a common complaint about imported png files.

shun-iwasawa:
@turtletooth Thank you for reviewing this.

For setting "premultiply" option for loaded PNG files, it can be done by using "Preferences > Loading > Level Settings by File Format". If you would like to set all the png file to be premultiplied, write `..*\.png` in the Regular Expression field will work fine.

level_settings

Jenkins again

shun-iwasawa:
Well... now I realized that you can do the same way for PSD files as well! :smile:

shun-iwasawa:
Regarding the premultiply setting, I'll change the approach by using the "Level Settings by File Format" settings instead of including them in the source.

shun-iwasawa:
Modified. Thanks @turtletooth for pointing it out!

Jenkins

turtletooth:
@shun-iwasawa Confirmed it works.  What does premultiplied do?  Never used it before.  Are you opposed to a PR doing the same for pngs?

LGTM.  Thanks.

shun-iwasawa:
"Premultiply" is one of the terms of digital compositing. [Here](https://en.wikipedia.org/wiki/Alpha_compositing#Alpha_blending) , [here](https://www.teamten.com/lawrence/graphics/premultiplication/)  and these videos ( [1](https://www.youtube.com/watch?v=-dK71TVrBLM), [2](https://www.youtube.com/watch?v=c0kPLEMF4xk) ) would be helpful to understand.

The "premultiply" option in the level settings will premultiply the raster level on showing in the viewer and on rendering, **without altering the original file**. Since OpenToonz handles levels with the assumption that they are premultiplied, you need to use this option for straight-alpha images such as PSD layers.

Regarding the PNG files, according to the [PNG Specification](https://www.w3.org/TR/PNG/#6AlphaRepresentation) , it says that all of them are straight-alpha. So I think it would be fine to introduce the same setting for PNG fromat as well.

turtletooth:
Thank you so much for the great explanation. PR submitted.

PR #1190, opened 2017-05-10
SoapBubbleFx improvement; curve the seam of looped spectrum 

shun-iwasawa:
This PR improves the SoapBubbleFx, especially for looping the spectrum by adjusting "Loop Spectrum Fade Width" and "Spectrum Shift" parameters.
In the old version, the seam of looped spectrum was straight (not varied in accordance with the depth) and caused irregular appearance since the color stripes of the spectrum are curved along the bubble shape.
This PR will curve the seam in order to make the appearance more natural like the attached image below:

stripe_curved

- The seam will be curved only if the "Loop Spectrum Fade Width" is higher than 0.
- Please note that the curvature of the seam cannot be fit completely with the stripes since they are not parallel.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1191, opened 2017-05-10
Create zerary fx node at right-clicked position in the fx schematic

shun-iwasawa:
This PR will modify the behavior when right-click on the fx schematic and create a zerary fx with "Add Fx" command.
Compared to non-zerary (normal) fxs, the created zerary fx node does not placed at the right-clicked postion and placed at next to the xsheet node. It is troublesome especially in the complicated scene with large schematic tree.
With this PR, zerary fx nodes will be placed at the right-clicked position as well as other fx nodes.

addfx


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1203, opened 2017-05-16
Fixes on convert feature from NAA raster images to TLV

shun-iwasawa:
This PR will modify the convert feature, especially for NAA (non-antialiased) raster images converting to TLV. This feature is mainly used for handling RETAS-compatible TGA drawings. 

I've introduced 3 modifications as follows:

- In the previous version, if you enable the "Add Antialiasing" option in the level settings, the unwanted gap appears in the converted TLV at boundaries between lines and filled areas. I modified this problem by setting the style indices at the line pixels properly.

- In the previous version, the one-pixel-width border lines of color regions are generated always at the "larger" region. I modified this behavior in order to generate the border line always at the "darker" regions. As it is more consistent with the rules of animation production. 

- In the previous version, the DPI of converted TLV is always set to 72. I added the option to set the dpi of the result, which can be selected from "Image DPI", "Camera DPI" and "Custom DPI". Please note that if the source image does not contain dpi information, the current camera DPI will be used.

naa_tga_to_tlv

### DIFF
--- a/toonz/sources/include/convert2tlv.h
+++ b/toonz/sources/include/convert2tlv.h
@@ -39,6 +39,8 @@ private:
   bool m_isUnpaintedFromNAA;
   bool m_appendDefaultPalette;
 
+  double m_dpi;
+
   void buildToonzRaster(TRasterCM32P &rout, const TRasterP &rin1,
                         const TRasterP &rin2);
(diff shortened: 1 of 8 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1206, opened 2017-05-19
Fix GUI glitches with high dpi monitors

### LINKED ISSUE #1201: Guide display issue
There is a significant display issue with the guides on my high resolution mac. They are offset and have a size issue when zooming. They are also not easy to see. I'm not sure if the issue stems from the resolution or from Sierra 10.12.4
img_3876

shun-iwasawa:
This fixes #1201 , modifying the display problems in high-dpi monitors as follows:

- Fixed the viewer guides to be displayed at correct positions.

- Fixed the flipbook to be resized to its content size on double-clicked on the title bar.

- Fixed some viewer gadgets ( including the fx gadgets and the anchor points for the skeleton tool ) to be displayed larger.

- Fixed swatch viewers in some popups such as the "Export Level" popup, the "Convert to Vectors" popup, the "Create Meshes" popup etc. in order to display the image properly. (i.e. Made `PlaneViewer` class to support high-dpi.)

### DIFF
--- a/toonz/sources/include/tools/toolutils.h
+++ b/toonz/sources/include/tools/toolutils.h
@@ -128,7 +128,7 @@ void drawBalloon(
     std::string text,       // balloon text
     const TPixel32 &color,  // ballon background color (text is black)
     TPoint delta,  // text position (pixels; pos is the origin; y grows upward)
-    bool isPicking = false,
+    double pixelSize, bool isPicking = false,
     std::vector<TRectD> *otherBalloons =
         0);  // avoid other balloons positions; add the new ballons positions
 
--- a/toonz/sources/include/toonzqt/planeviewer.h
+++ b/toonz/sources/include/toonzqt/planeviewer.h
@@ -8,6 +8,9 @@
 #include "traster.h"
 #include "timage.h"
 
+// TnzQt includes
+#include "toonzqt/glwidget_for_highdpi.h"
+
 // Qt includes
 #include <QGLWidget>
 
@@ -57,7 +60,7 @@ shader fx for some unknown reasons. So I will reluctantly keep using the
 obsolete class until the shader fx being overhauled. 2016/6/22 Shun
 */
 
-class DVAPI PlaneViewer : public QGLWidget {
+class DVAPI PlaneViewer : public GLWidgetForHighDpi {
 public:
   PlaneViewer(QWidget *parent);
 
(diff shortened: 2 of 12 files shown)


shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa Thanks for doing this.  I apologize for not catching these things in my review.

turtletooth:
I notice with the brush tool the red dot and black crosshairs are not in the same place.  This is not tested on a high dpi device but changing my 1920 x 1080 monitor to 150% scale.  - This makes drawing confusing.  

I don't know if this is related to a change from high dpi or when we changed pixels mode, but using "Actual Pixel Size" on a vector layer does not show actual size (I'm using the default of 1920 x 1080)- it says 100%, but it's not.

shun-iwasawa:
@turtletooth 

>I notice with the brush tool the red dot and black crosshairs are not in the same place. This is not tested on a high dpi device but changing my 1920 x 1080 monitor to 150% scale. - This makes drawing confusing.

Is this the same problem as #1212 or another problem happened only with the high dpi mode?

>I don't know if this is related to a change from high dpi or when we changed pixels mode, but using "Actual Pixel Size" on a vector layer does not show actual size (I'm using the default of 1920 x 1080)- it says 100%, but it's not.

Basically the vector layer has no dpi which is needed to compute the actual pixel size. If you set the current unit to "pixel", the "Actual Pixel Size" will zoom the viewer based on the camera size.

turtletooth:
> Is this the same problem as #1212 or another problem happened only with the high dpi mode?
I only have had that problem on high dpi mode.

> Basically the vector layer has no dpi which is needed to compute the actual pixel size. If you set the current unit to "pixel", the "Actual Pixel Size" will zoom the viewer based on the camera size.

That makes sense

shun-iwasawa:
>I only have had that problem on high dpi mode.

Could you let me know how to reproduce the problem? 
I used 1920 x 1080 monitor with 150% scale as well. But I couldn't see any notable difference with the positions between the red circle (representing brush tip) and the cross hair (mouse cursor) other than the small offset mentioned in #1212 .

turtletooth:
Here is a video of it.  I don't know how to reproduce it.

offset

1920 x 1080 at 150% - the only thing that might affect it is that I have three monitors (all set to 150% for the test).

It's probably a rare use case, so if anything it can be an issue for later.  I'll go ahead and merge this.

Thanks.

Issue #1208, opened 2017-05-19
It is impossible to close that window


beeheemooth:
I wrote about this here #1056
![ezgif-1-b2a45c8ced](https://cloud.githubusercontent.com/assets/25724133/26239349/5dd034fc-3c85-11e7-8fe2-f545993f6f45.gif)

turtletooth:
Fixed in #1214

PR #1214, opened 2017-05-23
Fix problem : the info viewer and the flipbook becomes impossible to close.

### LINKED ISSUE #1056: In "Run script" I can't close "File Info"
It's a bug?

shun-iwasawa:
This PR fixes #1056 and #1208 .
The problem can occur not only in the info viewer opened from "Run Script" browser, but also in any popup/panes opened from any file browser popup.

In the previous version, they tried to avoid this problem by setting the modal file browser to non-modal as soon as the info popup is shown ( it is the same resolution as Toonz Harlequin ). However for now it seems not to work properly for unknown reasons. I struggled to restore it but failed.
 
Finally I decided to set all the file browser popup to non-modal even if they are opened with `exec()` function. It will make the software less user-inductive, but I think it is better than freezing the interface...

### DIFF
--- a/toonz/sources/include/toonzqt/infoviewer.h
+++ b/toonz/sources/include/toonzqt/infoviewer.h
@@ -25,7 +25,6 @@ class InfoViewerImp;
 class DVAPI InfoViewer final : public DVGui::Dialog {
   Q_OBJECT
   std::unique_ptr<InfoViewerImp> m_imp;
-  QWidget *m_parent;
 
 public:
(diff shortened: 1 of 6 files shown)


shun-iwasawa:
Jenkins

turtletooth:
Thanks for doing this. At first glance it looks good. I'll try and test it tomorrow.

turtletooth:
Tested and works.  LGTM. Thanks for the fix.

Issue #794, opened 2016-09-07
[Enhancement] Flag "Premultiply" by default when importing 32bit transparent PNG (and TGA?) files


Rayek:
Currently semi-transparent pixels in 32bit PNG files are not premultiplied by default, and the user must open the level settings to turn this on. It is quite confusing to new users, and requires extra unnecessary clicks.

I would like to suggest that this level settings is automatically flagged (activated) when placing 32bit (semi) transparent PNG files (and perhaps for other file types such as TGA as well).

artisteacher:
I agree, having premultiply on by default for pngs, gifs, and psd layers would be really helpful. My students always think something is wrong with the image - or with OpenToonz when they see the white border around a drawing that they have loaded from Photoshop or etc.

turtletooth:
I don't fullly understand premultiplied, but thanks to a recent PR by @shun-iwasawa I know how to implement this.  Should this only be applied to 32bit pngs?  or all pngs?  psds are already implemented.

PR #1217, opened 2017-05-24
Premultiply by default on png

turtletooth:
#794

This turns on premultiply by default for png files.

### DIFF
--- a/toonz/sources/toonzlib/preferences.cpp
+++ b/toonz/sources/toonzlib/preferences.cpp
@@ -103,7 +103,7 @@ inline bool formatLess(const Preferences::LevelFormat &a,
 //=================================================================
 
 void getDefaultLevelFormats(LevelFormatVector &lfv) {
-  lfv.resize(2);
+  lfv.resize(3);
   {
     LevelFormat &lf = lfv[0];
 
@@ -116,6 +116,11 @@ void getDefaultLevelFormats(LevelFormatVector &lfv) {
     lfv[1].m_name                  = Preferences::tr("Adobe Photoshop");
     lfv[1].m_pathFormat            = QRegExp("..*\\.psd", Qt::CaseInsensitive);
     lfv[1].m_options.m_premultiply = true;
+
+    // for all PNG files, set premultiply by default
+    lfv[2].m_name = Preferences::tr("PNG");
+    lfv[2].m_pathFormat = QRegExp("..*\\.png", Qt::CaseInsensitive);
+    lfv[2].m_options.m_premultiply = true;
   }
 }
 


RodneyBaker:
I'm certainly glad this can be changed.

I understand the panic that sets in when people see images with the halo around it.
While the math behind the scenes is what is most important most people won't appreciate the difference but it can help to understand that "premultiplied" alpha is actually polluting the color channels.  With straight Alpha the channels are optimal (each individually) and the end result depends on the math and compositing processes applied to each. 

I find it interesting that Alpha Channels have increasingly become similar to Gamma workflow in that the source and target must be considered carefully.  This is likely the case where digital artists are using a wider range of programs than they did in the past; where they primarily stayed within one application. 

Perhaps the oddest thing is that using premultiplied images can force a user to remove the color matting from the color channels (i.e. unpremultiply them) to get back to the cleaner color channels (because the RGB channels have been inadvertently altered by the act of premultiplying their values with the alpha channel..  Whether they can do this will depend on the program they use and the options available to them.   

I've always found Alpha Channels to be fascinating (and powerful) but I had no idea just how fascinating they were.

shun-iwasawa:
LGTM Thank you @turtletooth !

@RodneyBaker In MotionBlurFxIwa, I implemented the procedure for judging whether the source image is premultiplied or straight alpha [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/stdfx/iwa_motionblurfx.cpp#L36-L42). 
For the premultiplied image, if channels are normalized to 0.0 - 1.0, each RGB channel value must be equal to or less than Alpha channel value for every pixels. So if you find at least one pixel of which R/G/B channel is greater than its alpha channel, the image can be guessed as straight alpha.

RodneyBaker:
@shun-iwasawa said:

> @RodneyBaker In MotionBlurFxIwa, I implemented the procedure for judging whether the source image is premultiplied or straight alpha here.  For the premultiplied image, if channels are normalized to 0.0 - 1.0, each RGB channel value must be equal to or less than Alpha channel value for every pixels. So if you find at least one pixel of which R/G/B channel is greater than its alpha channel, the image can be guessed as straight alpha.

Very interesting.I think I actually understood that!  
For speed of processing I find myself wondering  if your test is running a sample across the middle of the image first, starting from the top/left, or some other approach.  It is useful that only one pixel needs to be found where R/G/B is greater than it's alpha to identify the type of alpha.  Nice.

Thank you for the information and I look forward to trying out MotionBlurFXIwa!

RodneyBaker:
I'm going to go out on a limb and suggest something isn't quite right in the processing of premultiplied PNGs.

Assumption:  OpenToonz will render out images with premultiplied alphas.
I don't see an option to have OT render "straight" alphas although I assume there must be.

Try this test:
Render out a PNG image from OpenToonz that has an alpha channel.  For instance, use the Type Tool to create some text.
Import/Load that PNG into the xsheet and then Right Click on that frame/level in the xsheet to select Level Settings where we will toggle on/off the "premultiply" setting.
Test with premultiply on and render a preview in OT.
Note the dark edge around the text.
Test with premultiply off and render a preview in OT.
Note the edge is smooth (no dark edge).

My assumpion is that the alpha channel is being multiplied twice but I may be wrong in this.  Hopefully I'm right as this might be an esay fix.

Of note (as mentioned in my post above) I tested with a 'straight' alpha rendered in another program and  the image looks correct with "premultiply" on in Level Settings and wrong with it off (the image has the dreaded white halo.  So, OT is handing "straight" alphas with PNG images correctly by default.  So that much appears to be correct.

This is with the latest nightly build as  of 2017-05-26 08:10:40 UTC.

shun-iwasawa:
@RodneyBaker Thank you for pointing the problem out !

The problem is that OpenToonz renders the PNG image with premultiplied alpha.

According to the [PNG Specification](https://www.w3.org/TR/PNG/#6AlphaRepresentation) , it says that PNG should be straight-alpha in normal. 

shun-iwasawa:
I posted a new issue.

RodneyBaker:
Thanks for the quick response. 


For completeness sake... 
I tested with the TIF format as well and see the same result as with PNG.  

I also tested TGA.
The TGA format has a few other considerations but appears to be the same as PNG and TIF.   As the Alpha Channel is only supported in the 32bit flavor of TGA the abscense of an alpha channel in 8 and 24bit flavors might confuse some users BUT...  TGA isn't a commonly used format these days except by those who will likely know how to handle it.  After rall these years... TGA is still a great format.

shun-iwasawa:
PNG supports straight alpha only whereas TIFF and TGA also support premultiplied alpha.
Applying "premultiply" option to already-premultiplied image will cause dark edge due to multiplication multiplication.

In Japanese animation industry they are handling tons of TGA files as it is standard format for RETAS PaintMan.

Issue #1178, opened 2017-05-01
Bug: Camera scale 0 crashes OpenToonz


BrundleThwaite:
This is not the most important bug but as the title says when the camera scale is 0 OpenToonz crashes.

turtletooth:
fixed in #1219

PR #1219, opened 2017-05-24
Fix crash on camera scale 0

turtletooth:
Fix for #1178 

Makes the camera almost 0, but not quite.


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed the crash & the fix. Thank you @turtletooth .

PR #1220, opened 2017-05-24
trying to make the code more understandable `leftButtonUp` in brushtool

lordadamson:
Hi,
I'm trying to make the functions shorter and easier to read and understand. But I don't know if you guys think this is a good idea so I thought I'd open a pull request early to see if you'd like me to continue.

and if you do you probably have some comments on my code which I'll be happy to hear.

Thank you.

### DIFF
--- a/toonz/sources/tnztools/brushtool.cpp
+++ b/toonz/sources/tnztools/brushtool.cpp
@@ -1031,14 +1031,14 @@ void BrushTool::onDeactivate() {
    * ドラッグ中にツールが切り替わった場合に備え、onDeactivateにもMouseReleaseと同じ処理を行う
    * ---*/
   if (m_tileSaver && !m_isPath) {
-    bool isValid = m_enabled && m_active;
-    m_enabled    = false;
-    if (isValid) {
+	bool isValid = m_enabled && m_active;
+	m_enabled = false;
+	if (isValid) {
       TImageP image = getImage(true);
       if (TToonzImageP ti = image)
         finishRasterBrush(m_mousePos,
                           1); /*-- 最後のストロークの筆圧は1とする --*/
-    }
+	}
   }
   m_workRas   = TRaster32P();
   m_backupRas = TRasterCM32P();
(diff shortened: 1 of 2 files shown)


toonz/sources/tnztools/brushtool.cpp
```diff
@@ -1292,94 +1292,102 @@ void BrushTool::leftButtonDrag(const TPointD &pos, const TMouseEvent &e) {
...
-    TStroke *stroke = m_track.makeStroke(error);
-    int points      = stroke->getControlPointCount();
+	if (!isJustCreatedSpline(vi.getPointer())) {
+		m_isPrompting = true;
 
-    if (TVectorImageP vi = getImage(true)) {
-      struct Cleanup {
-        BrushTool *m_this;
-        ~Cleanup() { m_this->m_track.clear(), m_this->invalidate(); }
-      } cleanup = {this};
```
lordadamson: I'm not sure if I understand the point of this struct other than having its destructor called when out of scope. but then why not just call `m_track.clear()` and `invalidate()` at the end of the function?

shun-iwasawa: It seems to be aiming to make sure to call `m_track.clear()` and `invalidate()` in any case when out of `if` scope.
For instance, this will be called just after `if (ret == 2 || ret == 0) return;` at line 1322.

toonz/sources/tnztools/brushtool.cpp
```diff
@@ -1292,94 +1292,102 @@ void BrushTool::leftButtonDrag(const TPointD &pos, const TMouseEvent &e) {
...
+			DVGui::MsgBox(question, QObject::tr("Yes"), QObject::tr("No"), 0);
 
-      if (!isJustCreatedSpline(vi.getPointer())) {
-        m_isPrompting = true;
+		m_isPrompting = false;
 
-        QString question("Are you sure you want to replace the motion path?");
-        int ret =
-            DVGui::MsgBox(question, QObject::tr("Yes"), QObject::tr("No"), 0);
+		if (ret == 2 || ret == 0) return;
```
shun-iwasawa: (continued from the previous comment)
So, I think you need to call `m_track.clear()` and `invalidate()` here if you do not use `Cleanup` .

turtletooth:
I'm going to close this for now.  The brushtool.cpp file has changed quite a bit since this was introduced.

Issue #1209, opened 2017-05-19
Crash with Eraser Tool in Vector Level (gif)


beeheemooth:
![ezgif-1-411a57c1fa](https://cloud.githubusercontent.com/assets/25724133/26246753/69328c78-3ca3-11e7-8c53-dcf64733c9b3.gif)

PR #1221, opened 2017-05-24
Fix crash on arrows using eraser

turtletooth:
#1209 

This takes care of arrow presses changing the frame, but doesn't take care of shortcuts for next frame, next drawing. . .

@shun-iwasawa Any idea how to get those shortcuts recognized in the eraser tool to block them?


shun-iwasawa:
When the current frame is switched, `EraserTool::onImageChanged()` will be called.
So, how about performing the erase-finishing process in such function just like I implemented in `EraserTool::onDeactivate()` , instead of blocking the frame switching operation?

turtletooth:
It seems that by the time onImageChanged has been called, finishing is no longer available (the next image is the active image).  Thoughts?

shun-iwasawa:
Sorry, you're right!
I remember that I once fixed the same problem in the raster eraser.
Please see [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/tnztools/rastererasertool.cpp#L1175-L1176),  I made the tool to remember the current working FrameId during the mouse dragging from `leftButtonDown()` to `leftButtonUp()`. How about doing the similar approach, holding the FrameId or pointer to the current working image (`TVectorImageP`) in the `EraserTool` class?

turtletooth:
@shun-iwasawa What do you think of this?

shun-iwasawa:
Jenkins

shun-iwasawa:
Thank you @turtletooth it works fine. LGTM

Issue #1222, opened 2017-05-25
Flipbook crash (gif)


beeheemooth:
Here i use 2017-05-24 04:24:37 UTC opentoonz_nightlies
![ezgif-1-21b6540602](https://cloud.githubusercontent.com/assets/25724133/26458069/eb716e80-417a-11e7-8265-dc8da601732e.gif)

PR #1224, opened 2017-05-25
Fix Flipbook Crash

turtletooth:
Fix for #1222 
Not sure if this is the best way, but it works.


shun-iwasawa:
@turtletooth Thank you for the fix.
How about just inserting `if(!m_image) return;` at the beginning of `mousePress/Move/Release/DoubleClickEvent` in order to prevent any control before loading the image?

turtletooth:
@shun-iwasawa That makes sense.  Changes made.

shun-iwasawa:
Jenkins

shun-iwasawa:
@turtletooth  It works fine. LGTM

Issue #668, opened 2016-07-22
RGB Picker Tool  sometimes crashes on Pick Screen


turtletooth:
- Use the RGB Picker Tool
- Select Pick Screen
- Click UI Elements (Such as room names)
- Crash on assert - screenpicker.cpp
  ![assert_crash](https://cloud.githubusercontent.com/assets/4576381/17047205/98d90cd4-4f98-11e6-9f2f-60d926281b7b.jpg)
  ![call_stack](https://cloud.githubusercontent.com/assets/4576381/17047208/9af5487a-4f98-11e6-964a-a38af352090e.jpg)

turtletooth:
It seems to crash when the GUI is changing from a hover event - such as passing over a frame in a level to the black area and clicking.
![scenes_2](https://cloud.githubusercontent.com/assets/4576381/17047325/f04592d4-4f99-11e6-8067-57cdcb18b1f6.gif)

PR #1225, opened 2017-05-25
Fix for movement during screen pick

turtletooth:
Fix for #668


shun-iwasawa:
LGTM. Thank you @turtletooth for the fix!

Issue #860, opened 2016-10-09
Program crashing with eraser tool.


SandDance:
At the moment this can only be confirmed on raster mode but the program crashes if you quickly alternate between eraser and brush tools.
Example Using a wacom pen and you Alternate between eraser (back of the pen) and the point (front of the pen) quickly it will crash.

ideasman42:
This is not enough information: please give *exact steps* (what you do to setup the scene, how you access tools in a way that crashes)

TheGaluMoD:
I have also encountered this issue. here are the details:
Whenever you hold down whatever button you have the eraser tool set to, and then let go of the button before you are done erasing (so basically you are still touching the tablet with the pen) the program crashes. Obviously the expected behaviour is to continue erasing until you stop touching the tablet. The same thing also happens with the brush. So if you have the eraser tool selected for example, and then you hold down 'b' (or whatever you set the brush tool to), start drawing, and then let go of 'b' while still finishing a line, opentoonz crashes. I don't know if there are any other tools affected by this. The selection tool seems to not exhibit this behaviour, so it looks like it affects drawing related tools. Also it looks like it doesn't matter what level type you have selected (raster, or toonz raster/vector)

The best way to circumvent this issue atm is to not hold down a button when using a tool, but instead just tap it to switch to said tool, and then switch to the previous one when you're done. Of course you can also just keep holding down the button until your done erasing/drawing, however I don't trust myself to do that. I just instinctively let go of the button while erasing ^^

Also this is my first post here. I'm really thankful for this great software, and the amazing community surrounding it :) Great job guys! <3

ideasman42:
Can this crash be done without a tablet?

TheGaluMoD:
Yes, it's the same deal. You hold down the shortcut button, and then stop holding it down, while still pressing the left mouse button. For me those are all the steps required for the program to crash (without any type of error message, it just closes). I'm using Win10 x64.

ideasman42:
Managed to redo:

- Run OpenToonz with the default startup file.
- Press S-Key once,
- Press B-Key, while held LMB drag to paint.
- Without releasing the left mouse button, release the B-Key.
- Notice the brush stroke isn't executed properly *(doesn't draw at all or draws wrong color, for others it may crash)*.

I didn't manage to get a crash, but it does cause some strange half-finished brush state to be left.

Probably this isn't so hard to fix, just need to delay the key-release action once the stroke is finished.

[...]

PR #1228, opened 2017-05-26
Fix for Raster Level Tool Switch Crash

turtletooth:
#860 
I'm not sure this is the best fix, since the crash only happens on raster levels, but this seems to fix it.

I don't know if there will be residual effects since mouse up is not called on the tool before it is switched, but I don't know how to get the mouse position required in the function for leftbuttonup.

### DIFF
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -204,6 +204,7 @@ void FullColorBrushTool::onActivate() {
 //--------------------------------------------------------------------------------------------------
 
 void FullColorBrushTool::onDeactivate() {
+  if (m_mousePressed) leftButtonUp(m_mousePos, m_mouseEvent);
   m_workRaster = TRaster32P();
   m_backUpRas  = TRasterP();
 }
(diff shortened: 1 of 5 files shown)


shun-iwasawa:
The switching of tools can be detected with the function `onDeactivate()` and `onActivate()`.
So I think inserting a "finishing brush" procedure (the same procedure as `leftButtonUp()`) into `FullColorBrushTool::onDeactivate()` and `FullColorEraserTool::onDeactivate()` will work fine.
Please see [my modification done in the brush tool](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/tnztools/brushtool.cpp#L1029-L1045) for reference.

Meanwhile I'll try your modification next week. Thanks!

turtletooth:
How about something like this?

shun-iwasawa:
@turtletooth I'm so sorry for being not available these days as I was busy on writing some report to the Young Animator Training Project in Japan. 
I'll restart to review PRs tomorrow and will postpone the next version release for few days until the most of bug-fix PRs will be included.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM @turtletooth thank you for the fix.

Issue #1105, opened 2017-03-16
Change default level type


turtletooth:
I would like to propose changing the default level type to Toonz Raster - here's why:

1. There are some consistent issues with some shapes not being fillable in vector levels.
2. Vector levels can bog down the software if a user uses it for sketch type work with too many lines.

Several times this week I have had students cause OT to become unresponsive because they were drawing in a sketch style and I have experienced the frustration of shapes being unfillable on several occasions.

Thoughts?

RodneyBaker:
I'll add my 2 cents as someone who normally would prefer drawing in vector but recently changed to starting with Toonz Raster... and in particular with the Pencil option checked to be on.  

The fill issue has been such that I've been hard pressed to find an ideal and consistent workflow that leads to  shapes that will consistently fill in vector and I've found that starting in raster and then converting that to vector yields good results.  

The reason I've moved to using the Pencil option (in Toonz Raster) is that I get a consistent line whereas if I use brushes sometime I get what I want but sometimes not.

So, for my part, changing the default to Toonz Raster will be fine with me.  We can always adjust that setting in our preferences.

turtletooth:
Additionally- this is the level type where the eraser behaves the most like many would expect- you can erase part of a fill, and erasing part of a line that encloses a fill doesn't erase the fill.

artisteacher:
Both Toonz Vector & Toonz Raster are pretty unstable on Mac Os when using a tablet.

VektorHead:
You can change the default level type by going File | Preferences | Drawing and change the Default Level Type setting to Toonz Raster or Raster Level.

turtletooth:
> You can change the default level type by going File | Preferences | Drawing and change the Default Level Type setting to Toonz Raster or Raster Level.

I am aware of that, I am proposing changing the initial default for new installs.  This way new users might encounter less frustration than they are likely to have with vector..

turtletooth:
Closing - discussion can be found in the PR for this issue at #1233

PR #1233, opened 2017-05-26
Default to Toonz Raster Level

turtletooth:
This is probably a controversial PR.  And I don't mind closing it if the consensus opposes it.

This is for #1105 

Right now, vector fill is unreliable to the point of driving people crazy.  But vector is most everyone's first impression with painting in OT.  

I would prefer to keep vector the default since it gives greater flexibility for scale and camera movements, but I don't know anything about vector graphics and it doesn't seem that vector fill is being worked on.

I'm torn, but I thought I would throw this out there.

### DIFF
--- a/toonz/sources/toonzlib/preferences.cpp
+++ b/toonz/sources/toonzlib/preferences.cpp
@@ -313,7 +313,9 @@ Preferences::Preferences()
     , m_importPolicy(0)
     , m_watchFileSystem(true) {
   TCamera camera;
-  m_defLevelType   = PLI_XSHLEVEL;
+  // Change while vector level fill is still unreliable. 05/2017 JCB
+  // m_defLevelType   = PLI_XSHLEVEL;
+  m_defLevelType   = TZP_XSHLEVEL;
   m_defLevelWidth  = camera.getSize().lx;
   m_defLevelHeight = camera.getSize().ly;
   m_defLevelDpi    = camera.getDpi().x;


RodneyBaker:
This would be a new PR but...

As for the default, I believe I favor Toonz Raster for a number of reasons.  The primary one probably being that most images imported into OpenToonz will be raster.
The primary question that might be asked is, "What format do Japanese Studios generally default to and prefer."
Based on prior commentary by @shun-iwasawa I believe the answer might be (TGA format) which is raster.

But, here's a little tangent...
Perhaps at the end of installation (or even better... during installation) messages could let users know some of the defaults that are set in OpenToonz.  Kind of a 'Tips' documentation but one designed to educate users so they can know to change the defaults if they aren't optimal for their use.  The downside of this would be the long tail created for installation text that would require update should any of those defaults ever change.  Hmmm... okay, I"m talking myself out of that idea.  

While I'm a big fan of vector, if I had to cast a vote I believe it would be to set the default to Toonz Raster.
I'm trying to think of a downside of setting that as the default but none immediately comes to mind.

I will add...
Much of this may be a moot point if OpenToonz recognizes the image file formats being imported and automatically defaults to raster or vector based on that determination.   The consideration has more to do with someone opening OT and diving into drawing.  Some of this workflow might be resolved by adding two additional menu items to the File menu by default "New Scene (Raster)" and "New Scene (Vector)".  Yes, yet another (potential) PR.  ;)

And ah, since we are now talking about workflow and @turtletooth is the creator of the very useful Startup popup... that might be the better place to put the choices.  Rather than the current (Create Scene)
 button two buttons might be there "Create New Scene (Raster)" and "Create New Scene (Vector")

Apologies for the lengthy post but I believe considering the user's (optimal) workflow will minimize any difficulty experienced with the default scene being set to Toonz Raster.

This is the long way of me going on record as saying... Toonz Raster as default should be fine.

artisteacher:
Personally I like vector levels the best, but they really shouldn't be default until the fill and tablet issues are resolved. At least on OS X, toonz raster levels are only marginally more stable than vector levels - fills work but tablets are super glitchy. Raster levels are the only usable levels if you use a tablet.

Having a dropdown in the startup popup to set the default level type would be ideal. I wouldn't want to add more options to the file menu because it is already so long.

turtletooth:
I think this may become a moot point if I get the changes made to the xsheet toolbar.  That would give everyone instant access to whatever level they would like.  With MyPaint brushes coming soon, having fast access to all three level types would be useful.

One more reason why Toonz Raster may be a good default is for people who use more of a sketching style.  I have had students bring OT to its knees because they have had thousands of stroke lines on a single frame.  While Illustrator and ToonBoom seem to be able to handle greater numbers of strokes, OT isn't optimized for that (right now).

I keep coming back to how much vector makes for great scaling, zooming and camera movement, and I have seen great use of that, but I just don't think vector is there right now.

On a totally side note- I wish that the folks at Toonz would consider throwing a little help at some of the bigger issues-  I think they would actually help their brand if OT was more stable.  This is most people's first impression of anything Toonz and as crazy powerful as it is, we lose a lot of users due to bugs and glitches.

turtletooth:
I am going to close this since the xsheet toolbar gives faster access to all level types.

blurymind:
I personaly love the vector level for inking but must admit that coloring
in it can be dificult. It would be amazing if someone figures out a way to
improve the bucket tool there. :D

I will always prefer the vector level because they can do tweening.

Perhaps we can use vector levels for linework cleanup and inbetweens and
toonz raster levels for rough pass and color fills! Its the ideal scenario
atm - take the best of both worlds

Issue #1038, opened 2017-02-13
Inbetween button's clickable area is offset. 


artisteacher:
When using the levels panel to make an automatic tween with the in-betweener, the clickable area is offset to the right. Clicking on the left side of the button results in deselecting instead of tweening. Clicking on the right side of the button, and also the black border to the left of the button, brings up the tween pop-up to choose interpolation type as expected. 

I'm not sure if this is Mac specific or not. Thanks so much for all the progress you all are making!

turtletooth:
I've noticed that too.

turtletooth:
Fixed in #1241

PR #1241, opened 2017-05-31
Fix click area for inbetween

turtletooth:
Fix for #1038 

Ensured the entire graphic is clickable to bring up the inbetween dialog.


shun-iwasawa:
LGTM Thank you for the fix.

PR #1245, opened 2017-06-01
Fix of the file browser and popups hierarchy

### LINKED ISSUE #1231: Pop-up Hierarchy Incorrect
When there are multiple pop-up windows, the most recent does not always display on top. For example, in the last couple nightly builds that I've tried, when I choose a scene from the file browser (when using the start-up pop-up), the file browser displays underneath the startup pop-up. Same thing happens with the output settings pop-up. If I change the output location, the browser displays underneath the output settings pop-up. I've had similar issues before when working with FX, but these are new.
screen shot 2017-05-26 at 8 42 21 am

Also, sometimes the pop-ups display on top of other programs when they are active. For example, while rendering to video, if I check my email in a browser, the rendering pop-up stays in view on top of the browser (pre-existing issue at least in OS X)

shun-iwasawa:
This PR fixes #1231 .

I think I found the proper solution regarding the modal file browser and other popups (such as the info viewer and the flipbook) open from it.
I changed the strategy done in #1214, I introduced an utility function `FileBrowserPopup::setModalBrowserToParent()` instead. This function will set the modal file browser to parent of the popup, in order to allow it controllable with leaving the browser modal.

### DIFF
--- a/toonz/sources/toonz/filebrowserpopup.cpp
+++ b/toonz/sources/toonz/filebrowserpopup.cpp
@@ -61,6 +61,7 @@
 #include <QGroupBox>
 #include <QCoreApplication>
 #include <QMainWindow>
+#include <QApplication>
 
 //***********************************************************************************
 //    FileBrowserPopup  implementation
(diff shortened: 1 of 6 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

turtletooth:
I've tested it and the hierarchy seems to be working correctly.  I missed the issues with the last one- sorry about that.  Would you like me to merge?

shun-iwasawa:
@turtletooth Thank you for reviewing this. Yes, please merge it if it seems to be OK.

PR #1247, opened 2017-06-01
Use Segoe UI font on Windows

turtletooth:
Segoe UI is the standard UI font on Windows and in my opinion looks better than Arial.

### DIFF
--- a/toonz/sources/toonz/main.cpp
+++ b/toonz/sources/toonz/main.cpp
@@ -97,13 +97,14 @@ const char *dllRelativePath     = "./toonz6.app/Contents/Frameworks";
 #endif
 
 #ifdef _WIN32
-TEnv::StringVar EnvSoftwareCurrentFont("SoftwareCurrentFont", "Arial");
+TEnv::StringVar EnvSoftwareCurrentFont("SoftwareCurrentFont", "Segoe UI");
+
 #else
-TEnv::StringVar EnvSoftwareCurrentFont("SoftwareCurrentFont", "Hervetica");
+TEnv::StringVar EnvSoftwareCurrentFont("SoftwareCurrentFont", "Helvetica");
 #endif
 TEnv::IntVar EnvSoftwareCurrentFontSize("SoftwareCurrentFontSize", 12);
 TEnv::StringVar EnvSoftwareCurrentFontWeight("SoftwareCurrentFontWeightIsBold",
-                                             "Yes");
+                                             "No");
 
 const char *applicationFullName = "OpenToonz 1.1.2";
 const char *rootVarName         = "TOONZROOT";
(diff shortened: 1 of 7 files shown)


RNavega:
That's a lot of places with the explicit name of the font. Tough!

With the exception of [toonzqt/functionpanel.cpp](https://github.com/turtletooth/opentoonz/blob/d2912309ed48e1b635d74fb413fc061b4d37e41d/toonz/sources/toonzqt/functionpanel.cpp) and [toonzqt/spreadsheetviewer.cpp](https://github.com/turtletooth/opentoonz/blob/d2912309ed48e1b635d74fb413fc061b4d37e41d/toonz/sources/toonzqt/spreadsheetviewer.cpp), all the files seem to include [toonz/preferences.h](https://github.com/turtletooth/opentoonz/blob/master/toonz/sources/include/toonz/preferences.h).
The global font name ("Segoe UI") could be placed there so if there's the need to modify it, you could do it once and have all QFonts use that, and it also suggests that the ui font is a preference and could be made editable in the future, like with a "Pick UI font" widget in the preferences panel.

shun-iwasawa:
When I was developing Toonz Ghibli Edition, I did a comparison survey on UI font. I prepared screenshots of UIs with varions fonts (Arial, Verdana, Tahoma, Georgia, Lucida Fax, Lucida Sans, and Arial+Bold weight). At that time the Segoe UI did not exist.



Finally we decided to use Arial + Bold font as it had the best visibility among them. (As you may know, there were many ~~elderly~~ veterans worked in the studio!) 

It is somewhat reasonable to introduce the non-bold Segoe UI since it is standard, but I'm not sure if it will be useful for all users... As @RNavega suggested, if you would like to change the UI font, IMO it would better to make it available to choose font family and font weight in the preferences (I'm not sure whether altering the font size may break the UI layout or not). 
For now the font is set in the user .env file but I think it can be moved to the preferences, as the UI font will not be changed so often during normal use.

turtletooth:
Any ideas on how to list font choices so they are valid across platforms?

shun-iwasawa:
I'm quite unfamiliar with this, but it seems that [QFontDatabase](http://doc.qt.io/qt-5/qfontdatabase.html) class can obtain available font list.

RNavega:
The font in the .env file seems to tell the last one that was used with the Type tool, it's referenced here:
https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/tnztools/typetool.cpp#L84

If you're talking about the ui font then doing it through the style-sheet settings should be easier, like doing a
`setStyleSheet( "QWidget{ font-size:" + userPrefFontSize + "; font-family: \" + userPrefFontFamily + "\";" );`

or hardcoding the same, but in the [.qss files](https://github.com/opentoonz/opentoonz/blob/master/stuff/config/qss/Astral_072_Dark/Astral_072_Dark.qss#L407) (I tried doing that, it only changes for some dialogs).
Some widgets seem to override the stylesheet with a different font family, like in here:
https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonzqt/infoviewer.cpp#L227

shun-iwasawa:
@RNavega In the .env file, you can find the UI font settings as well, set in `SoftwareCurrentFont` value. It is referenced [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/main.cpp#L99-L106).

RNavega:
You're absolutely right. Thank you.

turtletooth:
Closed for #1386

Issue #1094, opened 2017-03-10
Geometric Tool:  End Polyline by pressing Enter


mcjadski:
Right now, when using the Geometric Tool's Polyline, you have to double-click to make the line stop.  This is somewhat inconvenient, as double clicking requires you to make the line longer.  

I think it would be more convenient if you could end the polyline by pressing the Enter key.

artisteacher:
Enter to end the polyline would definitely be an expected behavior for anyone familiar with graphics programs.

turtletooth:
Based on the discussion in #1207 feature requests are going to be closed here on GitHub unless there is a developer actively working on the feature. Feature requests can be discussed at the OpenToonz Google Group: https://groups.google.com/forum/#!forum/opentoonz_en

mcjadski:
I would personally classify this as a needed fix.  Without it, the Polyline is difficult to use, and it would be a very quick fix.

PR #1250, opened 2017-06-05
Allow enter to end polyline

turtletooth:
This is for #1094 

Allowing enter to end a polyline is somewhat expected behavior from other vector software.


shun-iwasawa:
Thank you @turtletooth it works fine!
LGTM

artisteacher:
The selection tool has a simplified polyline as well. It's probably not used frequently but in terms of consistency, should also use enter to complete the selection.

Issue #1115, opened 2017-03-21
Multiple studio palette paths


artisteacher:
In Mac OS X, the default global palettes are located in stuff/studiopalette/Global Palettes. However, the studio palette links to stuff/projects/studiopalette/Global Palettes (this is also where the browser room links directly. Is this just an oversight, or is there a reason to have multiple studio palette folders?

artisteacher:
It looks like the multiple paths are causing an issue with the cleanup palette.
![screen shot 2017-04-12 at 12 50 18 pm](https://cloud.githubusercontent.com/assets/21159116/24970696/4eecc082-1f83-11e7-8460-0f1c507ffc3e.png)

turtletooth:
PR Submitted, closing.

PR #1253, opened 2017-06-06
Fix for Studio Palette Path

turtletooth:
Fix for #1115 

I think I caught all the places this was happening.


shun-iwasawa:
Confirmed that it fixed the `studiopalette` path properly on OSX.
I can't check with Linux, but as far as I can see it should work fine as well.
Thanks @turtletooth ! LGTM

PR #1254, opened 2017-06-06
Fix to show result on mov rendering

shun-iwasawa:
This PR fixes the problem as follows (in windows version):

1. Set "Preferences>Interface>Open Flipbook after Rendering" to ON.

1. Set the output file format to "mov".

1. Render the scene.

In my environment, flipbook fails to show the rendered mov file and becomes blank.

It seems that `tipc::readMessage(stream, msg)` (removed in #1182) was needed in the `TLevelWriterMov` destructor. I reverted it without throwing exception.


masafumi-inoue:
LGTM

PR #1255, opened 2017-06-06
V1.1.3

shun-iwasawa:
- Applied clang-format  to whole sources
- Changed version number in `main.cpp`

I would like #1254 bugfix to be merged before this.


shun-iwasawa:
Jenkins

PR #1256, opened 2017-06-07
Allow New Levels to Default to Camera Size

turtletooth:
This PR adds a new item in the drawing section of preferences - to allow the new levels to be the same size as the current camera.  Right now, if you create a new Toonz Raster or raster level the default size is locked to what is set in preferences.  If you create a variety of scene sizes, such as game art, it can be tedious to resize the level each time you make a new level.

default_camera_size


artisteacher:
This will be really handy!

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -826,7 +826,8 @@ void PreferencesPopup::onDefLevelTypeChanged(int index) {
   if (0 <= index && index < m_defLevelType->count()) {
     int levelType = m_defLevelType->itemData(index).toInt();
     m_pref->setDefLevelType(levelType);
-    bool isRaster = levelType != PLI_XSHLEVEL;
+    bool isRaster =
+        levelType != PLI_XSHLEVEL && !m_newLevelToCameraSizeCB->isChecked();
     m_defLevelWidth->setEnabled(isRaster);
     m_defLevelHeight->setEnabled(isRaster);
     if (!m_pixelsOnlyCB->isChecked()) m_defLevelDpi->setEnabled(isRaster);
```
shun-iwasawa: If the `New Levels Default to the Current Camera Size` option in ON, the DPI field becomes disabled as well as the width and the height fields.
But for now the created level refers the value in the DPI field, not the current camera's dpi.

I think the level should be created with the current camera dpi, or, the DPI field should remain enabled even if this option is activated.
(IMO the former way would be simple and matches the purpose of this option.)

turtletooth: Agreed.

turtletooth:
Changes made.  Thanks for catching that @shun-iwasawa

turtletooth:
Conflicts resolved

shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa Any idea why the mac builds keep timing out?

shun-iwasawa:
@turtletooth Yesterday I tried to restart the job many-many times but always timed out. Next week I'll investigate which procedure becomes a bottleneck.
Meanwhile, I think we can ignore the OSX build result of Travis CI if it is verified by Jen.kins.

shun-iwasawa:
It works fine. Thanks @turtletooth for this useful option!
As mentioned above, for now I'll ignore the timed out Travis CI, since OSX build is completed in Jen.kins CI.

LGTM

Issue #1243, opened 2017-05-31
Bug: SpriteSheets not padded on the top and bottom


BrundleThwaite:
When I add padding to the top and bottom of a sprite the rendered sprite sheet does not include the padding.

SpriteSheet Dimensions: 213 X 474

Total Images: 7
Individual Image Width: 69
Individual Image Height: 158
Individual Image Width with Padding: 71
Individual Image Height with Padding: 160
Images Across: 3
Images Down : 3
Top Padding: 1
Bottom Padding: 1
Left Padding: 1
Right Padding: 1
Horizontal Space Between Images: 2
Vertical Space Between Images: 2

SpriteSheet Dimensions: 207 X 474

Total Images: 7
Individual Image Width: 69
Individual Image Height: 158
Individual Image Width with Padding: 69
Individual Image Height with Padding: 158
Images Across: 3
Images Down : 3
Top Padding: 0
Bottom Padding: 0
Left Padding: 0
Right Padding: 0
Horizontal Space Between Images: 0
Vertical Space Between Images: 0

I get the same result for much greater padding values like 8 and 24.

turtletooth:
Fixed in #1259

PR #1259, opened 2017-06-08
Fix vertical padding on spritesheets

turtletooth:
Fix for #1243 

The line for adding vertical padding to the total image size wasn't always being called.


shun-iwasawa:
Confirmed the fix. Thank you @turtletooth !
LGTM

Issue #1120, opened 2017-03-23
Feature Request: Add audio scrubbing to ComboViewer


artisteacher:
It's really wonderful being able to play audio in the ComboViewer, thanks @turtletooth! Right now scrubbing audio is still not possible. Having audio scrubbing added to the ComboViewer would be super helpful. 

Thanks everyone for the great progress!

turtletooth:
Fixed in #1260

PR #1260, opened 2017-06-08
Sound Scrubbing in Combo Viewer and Missing Updates in Scene Viewer

turtletooth:
#1120 
Added audio scrubbing to combo viewer- previously it did work sort of but only if the sound level was selected.

I also added some updates that the combo viewer had that the viewer didn't

### DIFF
--- a/toonz/sources/toonz/comboviewerpane.cpp
+++ b/toonz/sources/toonz/comboviewerpane.cpp
@@ -381,7 +381,8 @@ void ComboViewerPanel::showEvent(QShowEvent *event) {
   // onXshLevelSwitched(TXshLevel*)： changeWindowTitle() + updateFrameRange()
   ret = ret && connect(levelHandle, SIGNAL(xshLevelSwitched(TXshLevel *)), this,
                        SLOT(onXshLevelSwitched(TXshLevel *)));
-
+  ret = ret && connect(levelHandle, SIGNAL(xshLevelTitleChanged()), this,
+                       SLOT(changeWindowTitle()));
   // updateFrameRange(): update the frame slider's range
   ret = ret && connect(levelHandle, SIGNAL(xshLevelChanged()), this,
                        SLOT(updateFrameRange()));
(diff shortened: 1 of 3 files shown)


turtletooth:
I fixed another issue in the sound viewer - It seems the signals weren't getting disconnected, so the signals were being called multiple times.  If I switched to a room with a normal viewer and back to the room with a combo viewer, the signal was firing as many times as the combo viewer had been made active.  So sound samples would play multiple times - causing inaccurate playback.

@shun-iwasawa I also moved isFrameAlreadyCached to the SceneViewer - I'm not entirely sure that's needed, but I am trying to make the code in each viewer as similar as possible.  What are your thoughts?

toonz/sources/toonz/comboviewerpane.cpp
```diff
@@ -413,13 +411,38 @@ void ComboViewerPanel::showEvent(QShowEvent *event) {
...
+             SLOT(changeWindowTitle()));
+  disconnect(levelHandle, SIGNAL(xshLevelChanged()), this,
+             SLOT(updateFrameRange()));
+
+  disconnect(frameHandle, SIGNAL(frameSwitched()), this,
+             SLOT(changeWindowTitle()));
+  disconnect(frameHandle, SIGNAL(frameSwitched()), this,
+             SLOT(onFrameChanged()));
+  disconnect(frameHandle, SIGNAL(frameTypeChanged()), this,
+             SLOT(onFrameTypeChanged()));
```
shun-iwasawa: Thank you so much for fixing the bug!

`disconnect(app->getCurrentScene());` was obviously wrong as  `app->getCurrentScene()` is a sender, not a receiver...

How about to replacing lines 415-442 with the following lines for simplification? :
`disconnect(app->getCurrentFrame(), 0, this, 0);`
`disconnect(app->getCurrentScene(), 0, this, 0);`
`disconnect(app->getCurrentLevel(), 0, this, 0);`
`disconnect(app->getCurrentObject(), 0, this, 0);`
`disconnect(app->getCurrentXsheet(), 0, this, 0);`

According to the [Qt Documentation](http://doc.qt.io/qt-5/qobject.html#disconnect); 
> 0 may be used as a wildcard, meaning "any signal", "any receiving object", or "any slot in the receiving object", respectively.
>The sender may never be 0. (You cannot disconnect signals from more than one object in a single call.)
>If signal is 0, it disconnects receiver and method from any signal. If not, only the specified signal is disconnected.
>If receiver is 0, it disconnects anything connected to signal. If not, slots in objects other than receiver are not disconnected.
>If method is 0, it disconnects anything that is connected to receiver. If not, only slots named method will be disconnected, and all other slots are left alone. The method must be 0 if receiver is left out, so you cannot disconnect a specifically-named slot on all objects.

shun-iwasawa:
Confirmed that the sound scrubbing feature works fine.

I'll soon post another PR based on your commit which will be removed unused codes ( `isFrameAlreadyCached` ) left in the combo viewer and other sources.

toonz/sources/toonz/viewerpane.cpp
```diff
@@ -588,6 +600,12 @@ void SceneViewerPanel::onSceneChanged() {
   TApp *app         = TApp::instance();
   ToonzScene *scene = app->getCurrentScene()->getScene();
   assert(scene);
+  m_flipConsole->setFrameRate(TApp::instance()
+                                  ->getCurrentScene()
+                                  ->getScene()
+                                  ->getProperties()
+                                  ->getOutputProperties()
+                                  ->getFrameRate());
```
turtletooth: I'm not looking at the code, so I may be mistaken, but I believe the combo viewer does it- if I change the scene fps, the combo viewer fps updates (but not the viewer).  This is handy and less confusing if you change the scene fps and wonder why the viewer is still playing at the old fps.

shun-iwasawa: Oh, I see. I agree that the console fps should be synchronized with the scene settings.
However, I think the value should not be set back every moment you change the scene... 
Consider that you are setting the scene fps to 24, if you temporary set console fps to 8 (for checking a movement), you might not want to set it back to 24 once you edit the xsheet.
Well, let me think.

turtletooth: I see.  I never really used the combo viewer- so I didn't see it fire on every xsheet change.  I agree - we don't want that.

PR #1261, opened 2017-06-09
Fix New Level Name for Vector Level

turtletooth:
This PR does two things:

When the first level created is a vector level, the level create popup doesn't correctly set a new name - it gets A (only the first time, but this isn't as it should be).

To recreate:
Open OT
Click the new vector level button in the xsheet toolbar 
Notice the level name is A whether a level A exists already or not.

This PR also selects all text in the new level name field if the warning that the level name already exists, giving the user the chance to give a totally new name without having to backspace or select all.


shun-iwasawa:
It works fine. LGTM

PR #1263, opened 2017-06-09
New Fx: TimeCode Iwa

shun-iwasawa:
This was requested by some Japanese animation production.
Introduced TimeCode Iwa Fx which renders the current time code.

Parameters and usage are basically the same as AfterEffects.

timecode

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1224,6 +1224,16 @@
   <item>"STD_iwa_BokehFx.distance5"		"Layer5 Distance"</item>
   <item>"STD_iwa_BokehFx.bokeh_adjustment5"	"Layer5 Bokeh Adjustment"</item>
   
+  <item>"STD_iwa_TimeCodeFx"	"TimeCode Iwa"</item>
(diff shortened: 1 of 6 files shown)


shun-iwasawa:
Jenkins

artisteacher:
Since the type tool is completely nonfunctional in OS X, can this still work in OS X?

RodneyBaker:
@artisteacher  if it doesn't work that would be very strange as this implementation doesn't appear to have anything to do with the Text Tool.

turtletooth:
Tested and works fine.  LGTM.  Thanks @shun-iwasawa

RodneyBaker:
It took me a couple installs but I finally got the FX Timestamp IWA to appear as an option.
Not sure what happened there but it's working great now and this is a great addition.  Thanks!

artisteacher:
I also had problems getting the new fx to display. I had to delete the fxs folder (stuff/profiles/layouts/fxs) before it would show up. I'm wondering if the fxs.lst file didn't update on install? This could be mac specific. Anyways, now it works well. Thank you!

I am wondering, is there a reason to use semi-colons instead of colons? : is more common than ; for this usage.

PR #1264, opened 2017-06-09
fixed linux build documentation

lordadamson:
* removed the `cd -` in the build tiff because it's wrong and it can confuse noobs.
* replaced make with make -j 4 in the instructions for a faster build


ideasman42:
Thanks for the PR. Committed similar changes cec9b90a59a57c9a97a655a8f78a8bd0230f78ae

(use `nproc` since 8+ cores are common).

lordadamson:
alright cool

PR #1265, opened 2017-06-10
Add autosave options to startup popup

turtletooth:
I have seen a number of people on twitter recently complaining about lost work due to a crash before saving.  I added the basic autosave options to the startup popup:

autosave

If people more easily could access the autosave options, this might be less of an issue.


RodneyBaker:
I'm not a code reviewer but everything appears to be in order here and working properly.
I tried to break it.  Tried to set inappropriate settings.  Looked over the code for any obvious inconsistency or error.  It works seamlessly with setting the autosave settings in Preferences and I like that the popup doesn't try to include additional options better addressed in Preferences.

Aside:  Additional Benefits
The addition of this option in the Startup popup also has the plus of reminding us what timeframe we have set the autosave.  This gives us a better sense of security and confidence with regard to whether our data will be readily available when needed.  Moreover, this visualization upon opening OT has an benefit of better understanding our time sitting in front of our screens as it aids with unconsciously or consciously optimizing work/rest intervals.  OT's startup popup may be the only startup screen that I have any real desire to interact with as most I have used only serve to get in the way or slow down progress of opening my files or being creative.   This popup continues to be well thought out and (most importantly) useful.

I see no reason not to merge.

shun-iwasawa:
Jenkins

shun-iwasawa:
Thank you @turtletooth. ( and sorry for the late response! ) LGTM

PR #1286, opened 2017-06-15
Allow drawing substitution on subxsheet levels

turtletooth:
Right now drawing substitution only works on drawing levels, this change allows drawing substitution on subxsheet levels also.  This is useful if mouth, face, etc. . . are actually made up of multiple levels.

### DIFF
--- a/toonz/sources/toonz/xsheetcmd.cpp
+++ b/toonz/sources/toonz/xsheetcmd.cpp
@@ -28,6 +28,7 @@
 #include "toonz/levelset.h"
 #include "toonz/txshsimplelevel.h"
 #include "toonz/txshleveltypes.h"
+#include "toonz/txshsoundtextlevel.h"
 #include "toonz/tstageobjecttree.h"
 #include "toonz/tstageobjectkeyframe.h"
 #include "toonz/txshcolumn.h"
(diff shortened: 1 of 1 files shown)


turtletooth:
I also added the ability to use the command on empty cells if the previous cell was valid.

turtletooth:
Added note levels too.

shun-iwasawa:
I've tested this PR and found some problem when the empty cells are selected. Please see the attached image:

1. Using "Substitution Forward", it becomes like this.
1. Undoing the command cannot revert the scene.
1. Using "Substitution Backward", drawing number appears in the empty cells in strange order.
1. Undoing the command cannot revert the scene neither.

image 

Actually I'm not quite sure what should happens when these commands are used to the empty cells...

turtletooth:
I think I'll take out empty cells for now.  Thanks for testing @shun-iwasawa

turtletooth:
@shun-iwasawa  I think I figured out a way to keep the empty cells functionality.  I created it as a way to quickly create a xsheet level if the drawings already existed.

I think my changes fix the issue.  Here is my expected behavior:
If a non-empty cell is selected followed by empty cells, they all get the cell id that would be assigned to the non-empty cell.

If only empty cells are selected but they are immediately after a non-empty cell, they will get the cell id of the non-empty cell as a starting point.

shun-iwasawa:
Jenkins

shun-iwasawa:
It works fine ! Thank you @turtletooth for enhancement. LGTM

If only empty cells are selected and they are not after a non-empty cell, then this command will do nothing.
I think it would better to make the command not to register undo in such case. 
This problem exists in the current master so it can be a separate issue.

Issue #1283, opened 2017-06-14
Non-responsive when trying to create new level (Xsheet toolbar)


gtrevelldickson:
When I try to create a new level from the X sheet toolbar I can no longer interact with Opentoonz. It doesn't crash or freeze, because it still auto saves if I let it sit. When this happens the **only** way to close Opentoonz is through windows task manger.
Before this happened, I noticed I can't select both a Vector & Raster level and put them in the same sub x sheet unless I :

1. Put Raster in sub xsheet alone.
2. Open that sub sheet
3. Create vector level
 
I did this several times in my last animation and I feel this may have caused the problem.
Last thing I remember doing was trying to add another vector level into a sub xsheet that already had a Raster & Vector in it. **This problem happens in new projects also** _Running 1.1.3._

turtletooth:
I think the new level popup window is somehow hidden.  What happens if you hit escape?  Can you get control back?

gtrevelldickson:
Wow...It worked! Idk why I never tried that lol thank you! 
The popup window is still hidden though. Any suggestions?

gtrevelldickson:
@turtletooth Ok, I just got the window to show up again and everything looks good for now.
To make it come back I:
1. made the problem happen
2.  then I plugged my laptop up to a 2nd screen.
 That brought it back to the front.

turtletooth:
For now you can delete popups.ini in the stuff->profiles->layouts->settings folder
I'll see about creating a menu option to reset the popups

turtletooth:
Fixed in #1287

PR #1287, opened 2017-06-16
Fix popup locations

turtletooth:
This is a fix for #1283 
If a popup gets  moved off screen, this moves it back.

This also puts the startup popup on the screen OpenToonz is on, not just the primary monitor.


shun-iwasawa:
Jenkins

shun-iwasawa:
This will make the software safer. Thank you @turtletooth !
LGTM

Issue #1270, opened 2017-06-13
Greyed Out windows when deactivating "Main window's full screen"


MikioKimura:
## **_When using “main window’s full screen mode”  and the mouse hovers over another window that sits outside of full screen mode, all of the windows go grey when full screen is deactivated; To currently correct this the mouse must hover over the window that activated full screen mode; the artists that I work with in the studio seem to find this incredibly frustrating._**

![001](https://user-images.githubusercontent.com/19598708/27067456-59e34c26-504e-11e7-949b-16eb3f4d48ea.PNG)

![002](https://user-images.githubusercontent.com/19598708/27067466-6268c70e-504e-11e7-8535-49096ad067ca.PNG)

![003](https://user-images.githubusercontent.com/19598708/27067480-6fdcd452-504e-11e7-8cbb-76de05522ec2.PNG)

MikioKimura:
This Bug appeared on these computers

**Desktop Computer 1 specs**
Operating System: Windows 7 unlimited SP 1
Ram: 32 gigs of ram
Screens: Wacom Cintiq DTZ 2100, Viano TV, Samsung computer screen
Graphics card; GTX 960
Motherboard Intel core i5-2400 CPU @ 3.1GHz

**Desktop Computer 2 specs**
Operating System: Windows 10 Pro
Ram: 16 GB
Screens: 2 Dell Monitors and a Huion Gt 220
Graphics Card: GeForce GTX 970
Processor Intel Core i5-6500 @ 3.20GHz 3.20Ghz

**Surface Pro Computer specs:**
Operating System: Windows 8
Ram: 4 gb
Graphics card: Intel HD Graphics 4000
Processor: Intel Core i5-3317U CPU @ 1.7 GHz 2.6 GHz

**Surface Pro 4 Computer specs:**
Operating System: Windows 10
Ram: 4 gb
Graphics card: Intel HD Graphics 4000
Processor: Intel Core i5-6300U CPU @ 2.4 GHz 2.5 GHz

PR #1288, opened 2017-06-16
Fix for maximized grayed out windows

turtletooth:
Fix for #1270


toonz/sources/toonz/mainwindow.cpp
```diff
@@ -1112,11 +1112,19 @@ void MainWindow::resetRoomsLayout() {
 }
 
 void MainWindow::maximizePanel() {
+  if (m_panelMaximized) {
+    m_activePanel->maximizeDock();
+    m_panelMaximized = false;
+    return;
+  }
```
shun-iwasawa: Confirmed that this PR fixes the bug.

However, for now this PR allows only one room to be maximized.
When you already have maximized panel in some room (say, room 1), to maximize some panel in another room (room 2) will need to hit the shortcut twice, one is for deactivating maximize in the room 1, another is for maximizing the panel in room 2.

How about having `m_panelMaximized` and `m_activePanel` in each room, instead of the `MainWindow`?

turtletooth: Agreed and fixed.

turtletooth:
Changes submitted.

shun-iwasawa:
Jenkins

shun-iwasawa:
The grayed-out problem still occurs when you do as follows:

- Maximize some panel
- Quit and restart OpenToonz
- Release maximization while the mouse cursor is at outside of the original region of the maximized panel

And now, I realized that the information which panel is maximized is already stored in `DoclLayout` !
So sorry for late notice, but instead of keeping the `m_activePanel` and `m_panleMaximized` in `Room`, how about just retrieving it from `DoclLayout` like as follows? :

```
 void MainWindow::maximizePanel() {
  Room *currRoom = getCurrentRoom();
   DockLayout *currDockLayout = getCurrentRoom()->dockLayout();
-  if (currRoom->m_panelMaximized) {
-    currRoom->m_activePanel->maximizeDock();
-    currRoom->m_panelMaximized = false;
+  if (currDockLayout->getMaximized() &&
+      currDockLayout->getMaximized()->isMaximized()) {
+    currDockLayout->getMaximized()->maximizeDock();
     return;
   }
  
  QPoint p                   = mapFromGlobal(QCursor::pos());
  QWidget *currWidget        = currDockLayout->containerOf(p);
  DockWidget *currW = dynamic_cast<DockWidget *>(currWidget);
   if (currW) currW->maximizeDock();
-  currRoom->m_panelMaximized = true;
-  currRoom->m_activePanel    = currW;
 }
```

turtletooth:
Nice solution @shun-iwasawa .  I made the changes requested.  Seems to work great with multiple rooms maximized even after closing and reopening.

shun-iwasawa:
Jenkins again

shun-iwasawa:
LGTM. @turtletooth Thank you for the bug fix!

PR #1289, opened 2017-06-18
Lip Sync Popup - Apply Text Data to a Level for Lip Sync

turtletooth:
lipsync

This popup allows a user to apply data from a text file generated by hand or from another application such as papagayo to a level.  The user can choose which level drawings they want to use for which mouth shape using the arrows below each icon.  The mouth shapes are based off of the standard Preston Blair mouth shapes.

This works for all image level types including sub xsheet levels, with the only exception that sub xsheet levels don't provide icons and therefore use a frame number reference.

This action is undoable, so the original level order can be recovered if the results need changed.

To do:

- [x] Make an option to fill the rest of the column (up to the end marker) with the rest pose.
- [x] Disable apply button if a valid file isn't selected
- [x] Add menu items


turtletooth:
sync2

New version.

bobbybutterfly:
very cool feature!
Another additional feature suggestion: for the more classic approach would it not be nice to have the option to handdraw some letters in the voicetrack with the pen, like in the classic exposure sheets?
lipscript

turtletooth:
Added the drawing number underneath the picture, now activating this from the right click menu on a cell automatically sets the Begin at value to the current frame number.

Added some spacing to the layout and fixed an undo bug.

bobbybutterfly:
@turtletooth 
Tanks fort this great little tool.  I love it.
Great would be another button in this window, where you can offset the papagayo data for 1,2,3 or 4 frames.
There is a rule for lipsyncing "see the mouthshape, b e f o r e you hear it" but the papagayo data is too exacty on the mouth shapes

shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @turtletooth for the addition!

PR #466, opened 2016-06-17
Linux support for tfont (Qt), from pr #51

ideasman42:
Credit goes to François Revol for this work.

PR #1294, opened 2017-06-22
Enable Type Tool in OSX

### LINKED ISSUE #370: Type Tool don't work in OSX

shun-iwasawa:
This PR fixes #370 (finally!) by using Qt-based tfont implementation done in #466 .
#466 was originally for Linux, so I just re-used it in other platforms.
(I replaced the existing implementation by the qt-based for Windows as well, in order to make it simpler.)

For now there are some inconsistency with the previous version in Windows (such as font size, line space, and character space), I think it can be resolved in future work.

I tested in Windows and OSX. I hope someone would check this on Linux.


shun-iwasawa:
Jenkins

artisteacher:
Thank you!!!

artisteacher:
I'm looking forward to using text.

RodneyBaker:
@artistteacher said:
>I'm looking forward to using text.

Not so much 'text' as 'type'.  ;)

I'm in the process of trying to figure out how best to put together a proposal to enhance @shun-iwasawa's TimecodeIwa fx to include actual text.  I've stared at the TimecodeIwa code a little and have been wondering if I might be able to stumble through it enough to code an update on my own (at present I'm guessing not!).  The idea would be to make a 'TimeCodePlusIwa that includes additional fields for entry of text that can accompany the timecode (and with an option to turn the actual timecode off... which would allow for overprinting only the text.  That text could then be placed anywhere on the canvas and animated in the same fashion as the current TimecodeIwa fx.  Adding the additional fields isn't too much of an issue but I'm not sure where the user entered text would be stored unless an option to point to a text file would be included.  One useful option in a TimecodePlusIwa node might be that of pointing to a text file that contains Subtitle or Close Caption in a standard format (but... that would be a pretty tall order and likely would be yet another different fx node... but it is a logical extension of the basic processing of TimecodeIwa etc.)


Sorry to go off topic but I wanted to make the distinction between the Type Tool and the use of actual (editable) text.  Because the Type Tool converts whatever text is entered to vector shapes very early in the process which then cannot be further edited there is a considerable limitationThe current TimeCodeIwa comes very close to this functionality. 
I can imagine what success might look like would be to use the current Type Tool for all graphical text elements that do not require (textual) editing while an fx node in the vein of TimecodeIwa would be use to add editable text.

Hope that makes sense.

turtletooth:
I tried this out and there are a couple thoughts- it seems that text is rendered larger using this version than previous.  Both of the images have fonts at 70 and 150, but notice how much larger the new version is:
Old
text_size_old
New
new_text_size

There are also issues with the final image, but they exist on both versions and will vary depending on what size the font is:
new_text
Notice the little imperfections on the letters?

The new changes work and will really help out mac users.  Thanks @shun-iwasawa

artisteacher:
@RodneyBaker, thanks for the clarification. At this point I have to save images of text in Photoshop and then load those images into OT. Being able to add type directly will save a lot of time - even if it is not editable. 

It would be great at some point, to have the type remain editable by default, with the ability to convert to vector at any point in the process. That is a more expected behavior. However for now, just having access to type in any form will be amazing.

artisteacher:
Any chance this will be merged before school starts? 🙂

turtletooth:
@artisteacher Here you go!
There are still some tweaks worth playing with - such as adding more size options to the drop down and potentially fixing some render issues (which I have no idea how to do). But the functionality is there.
Thank you @shun-iwasawa for the fix.  The mac users should love it.

jpturcotte:
Thank you so much for this fix, @shun-iwasawa!

Issue #611, opened 2016-07-09
[feature] x-sheet /selection tool - preserve line selection when moving between frames


blurymind:
It would make it much easier to edit a line across multiple cells if the line's selection was kept when you move to another frame where the line exists.

Lets say that you want to change the style or the thickness or the shape of a line across multiple drawings.
Currently you have to reselect the line every time you move to the next frame.

Here is a mockup to illustrate:
![selectionacrossframes](https://cloud.githubusercontent.com/assets/6495061/16707734/054b3fee-45d1-11e6-83ab-addbee1fe2d9.gif)

This could of course be an optional behavior.

turtletooth:
That would be nice.  On a totally side note, what did you use to make the gif.  I dig it.

blurymind:
just some basic editing in camtasia :)

blurymind:
This behavior will be most advantageous in cases where you have selected a number of lines that you want to alter between frames - by transforming them , changing their style or their thickness.
It can save hours of work imo. I found myself doing that a lot and re-selecting the lines on every frame change is the thing that took the most of the time.

blurymind:
I wonder if @shun-iwasawa could look into it :)
Is it a technically difficult feature to implement? Preserving vector line selection between frames?

turtletooth:
Based on the discussion in #1207 feature requests are going to be closed here on GitHub unless there is a developer actively working on the feature. Feature requests can be discussed at the OpenToonz Google Group: https://groups.google.com/forum/#!forum/opentoonz_en

Issue #565, opened 2016-07-01
feature - Inbetween (tween) onion skin mode!


blurymind:
This is a request for a new onion skin mode that gets enabled when multiple frames in the level strip are selected and the 'inbetween' buttons appear.
The mode allows you to get a good idea of the order in which lines are drawn in the refference frame and it also shows you the direction.

The easiest way to illustrate it is by looking at Cacani  - 
https://www.youtube.com/watch?v=NQEgL8tZg-0

The onion skin displays only the first previous/next reference keyframe.
It also highlights the one stroke from the previous frame (depending on how many strokes you have made so far on the active one)  that corresponds to the one that you are about to make.

We can get this feature in opentoonz if somebody knows how to implement it. it will be incredibly useful to have!

![onionhighlight-inbetween](https://cloud.githubusercontent.com/assets/6495061/16528892/7e20aa3c-3fb7-11e6-8c79-477f0797b9d4.PNG)

In order to enable the inbetween onion skin mode, simply select multiple frames in the level strip - this also brings the 'inbetween' buttons.

Depending on where your active frame is and last selected frame is - the onion skin looks back or forth.
Even if there are lines in the inbetween frames - the onion skin ignores them. It only cares about the inbetween refference frame (0001 in this case).

The onion skin inbetween reference frame displays in a different color than the normal onion skin (purple?) -  and also draws a direction hint on top of one of the lines (>>>>>>) - that way it highlights the correct line in the order you are drawing and also its direction.

@walkerka Can you really implement this? This would be the second time you radically improve the software with a single blow. ;) 
The OT community will love you even more.
A lot of people have been crying out for this feature on other animation software. 
It's a simple, but very powerful feature that makes the process of automatic tweening much easier and organic.

blurymind:
This is like the holy grail for cleanup artists and inbetweeners. I am not gonna lie. :D 
It is so much better than simply copy and pasting the entire drawing to deform it.

blurymind:
And here it is in action - in tweenmaker - Note how the onion skin outlines the line order you are at and also the direction:
![sequentianonionskin](https://cloud.githubusercontent.com/assets/6495061/16536140/0e107b76-3fe6-11e6-81f2-7377a30ad038.gif)

because a single picture speaks volumes.
This is a really nice way to work btw.

walkerka:
[...]

PR #1297, opened 2017-06-24
Guided Drawing Feature

turtletooth:
This feature adds a new option to the vector brush and geometry tools- Guided Drawing

In order to successfully use auto-tweening of vector drawings, the strokes need to be in the same order and direction.  This change makes the onion skin highlight what stroke is next in order and draws an x at the start of the stroke- making it much easier to keep track of what order the strokes need to be drawn in.

guided


turtletooth:
@blurymind Thought you might like this.

beeheemooth:
Finally !!!
I'm sure that many are waiting for this, as I am.

RodneyBaker:
Very nice!

blurymind:
Sweet! Thank you for implementing this. Would it be possible to share a
test build so i can play with it for a bit. I can give you some feedback or
report if i find any bugs with it :)

blurymind:
I requested this feature a while ago here:
https://github.com/opentoonz/opentoonz/issues/565

And as an inbetween artist am very very very happy to see it implemented

I dont know how hard it would be, but while you are at this, can you also
make opentoonz preserve line selection between frames?
https://github.com/opentoonz/opentoonz/issues/611
It seems like a small thing, but it can get really laborious when you have
to tweak the thickness of a bunch of lines or nudge them - because you have
to select the same lines over and over again when you flip through frames

turtletooth:
@blurymind - This currently works on both level view and xsheet view.  It currently only highlights the immediately preceding frame.  

So for now the workflow would be to have your key poses in order with now inbetweens and draw them in order.  Insert frames between in the level strip, auto inbetween, and expose in the xsheet.

It's not quite the workflow that you requested, but it's a big start.

artisteacher:
I'm excited to try this out! This is a great start and hopefully at some point in the future it will be possible to highlight any frame based on the first drawing.

blurymind:
@turtletooth that sounds like it would be already very usable for production! My proposal was sort of pushing the idea as far as possible. In reality we might not need everything in the proposal. Only thing that could be suggested at this stage is the presentation in the gui. It seems like this is presented to the user as a brush tool feature, while it really is an extra option for the onion skinning in opentoonz. I wonder how it plays with onion skinning- can it work while onion skinning is on and off? Also wouldnt it make more sense to have it activate with the onion skin by default, instead of cluttering the top panel?

https://cdn-images-1.medium.com/max/1600/1*fttJXyuH8z4AlugLwf3T3Q.png

blurymind:
@artistteacher instead of marking the first frame, a better aproach to this
imo would be for opentoonz to automatically recognise frames that are
empty- by not having any strokes on them.That way in the future
@turtletooth would be able to pave the way for the guided tween to also
preview the line spacing of the tween over these empty frames :D that would
in my eyes complete the feature to its full power.

turtletooth:
@blurymind It is a modification to the onion skinning, but since it is only relevant to the brush and geometry tools, it is only available in those tools.  It is not available when onion skin is off since all I am doing is actually coloring the stroke of the onion skin.

blurymind:
@turtletooth that is a good design imo. Was just wondering why not have it
enabled all the time by default and in order to turn it off- have an option
in the preferences window- where all the onion skin settings are. It doesnt
really affect any of the qualities of the stroke of the brush tool, so it
seems strange to find the checkbox to enable it next to the line settings :)

turtletooth:
@blurymind I don't think it makes sense to show this on tools that can't do anything with it.  It's location in the Tool Options Bar makes sense to me since it is a tool option.

blurymind:
@turtletooth thats fair enough :) I just hope to see it merged. It doesn't bother me that much

I am saying that because of the way I've seen it implemented in other software.
It makes sense to me to have it on by default without the need to tick a box.

I think the user will immediately notice that it is not available in some tools - even if you don't have it as a tool option on them.. It's obvious from the viewport.
If you compare it to all the other tool options - it's the only one that doesnt actually affect the stroke that the tool creates - it only affects the onion skin preview. So my intuition connects it to the onion skin viewport settings instead of the tool's settings.

You might also have people tick the box, but not have onion skin enabled - wonder why nothing is happening when they do. They will tick the box, draw a stroke and see no effect - then wonder what that box does to their stroke. It just says 'Guided'. There is no hint whatsoever that it has to do with onion skinning

beeheemooth:
If I suddenly change my mind to draw the current highlighted stroke, and I want to draw the next in order stroke - is it possible to skip it? (That is, switch the highlight of the next or previous stroke)

openanim:
Thanks you turtletooth to make opentoonz great again 💃  Really it's AWESOME (sorry for caps) !
Ty blurymind for this request :D. Another great feature from Cacani could be his timing panel to change timing naturally and easily. That can be great if opentoonz can integret that. Anyway it's a big start, now inbetweening with Opentoonz will be fantastic. It's becoming the best 2D software for me. 

It's possible to share a test or unstable build to see how that works ?

turtletooth:
Quick question- an easy change to this would be to highlight the current stroke number on all the onion skin frames shown instead of just the most recent.  Would that be preferable to just the closest frame?  It isn't quite pick your frame, but it gives more flexibility than there is currently.

beeheemooth:
@turtletooth And to what type of onion skin do you want to apply it: to static or dynamic? My opinion is that for the dynamic onion skin, the nearest frame is sufficient.

artisteacher:
My vote is for all onion skin frames - this would be particularly helpful when making changes/updates to something that has already been auto-tweened.

turtletooth:
So I added the options to show all 3. 

The terminology is closest, farthest and all (relative to the current frame).  I'm not too keen on the word choice but first and last could also be confusing- first in chronological order or first moving away from the current frame.  

Either way, the functionality is there.

openanim:
@turtletooth !! Really nice ty so much for this feature, impatient to play with :D great idea to put these options.

artisteacher:
That will be great to have all 3 options. What about proximity and distance?

turtletooth:
To preserve space in the tool options, I moved this to the right click context menu of the viewer.

artisteacher:
What about a more options button in the tool option bar?

turtletooth:
I moved it down there to not only preserve space, but with it moved to where you activate onion skin, it is accessible whenever you want it - not just brush and geometry tools.  There might be a use case where having it visible with another tool might be useful.

blurymind:
@turtletooth thank you! Makes more sense to find it there too-D

blurymind:
@turtletooth as i expected , having it with the brush options
miscommunicates to users what it does. Here is an example of a guy who
thinks that guided is used for the brush stroke z order:
https://youtu.be/R1n9Ni3QXjk
At 18:25

Having it with the onion skin options makes much more sense also helps
reduce the situation with the clutter up there

turtletooth:
@blurymind Good idea on the status icons in the right click menu.  I've added them.

blurymind:
@turtletooth  thank you!! These features are awesome!

artisteacher:
I'm still looking forward to trying this out! 

In the future, it might be helpful for drawing closed shapes if a short arrow could replace the  x, so that the marker also showed the direction.

RodneyBaker:
>
> In the future, it might be helpful for drawing closed shapes if a short
> arrow could replace the x, so that the marker also showed the direction.


I find my self in agreement there as I have to keep reminding myself
whether the x is the origin or the target to hit.
It seems that an X might signify target.

Of course, if there was a symbol at both ends it might be a small x,
circle, or plus at the base when a directional pointer at the end?


x----->
o----->
+----->

Regardless, this is a great feature.  Love it!  :)

RodneyBaker:
There was an aspect of the Guided Drawing Feature that I had put aside and forgot in favor of some of the other recent additions such as Snap, etc.  I am so glad the lightbulb finally went on regarding the primary aspect of the featuer which simply highlights the next line that should be drawn.  I'm not sure why that aspect didn't sink in before.

turtletooth:
Conflicts resolved.

turtletooth:
How does this look for giving the direction to go:
guide

blurymind:
it looks more intuitive, but the x sign is easier to spot I guess :)

turtletooth:
How about an x with the line then?  Just trying to give a sense of which direction to draw in.

gab3d:
If this shows while drawing, then I think the X is not necessary and it'd
clutter the visibility of the brush radius itself.
personally, I liked how it looked. :)

2017-09-05 16:44 GMT-03:00 Jeremy Bullock <notifications@github.com>:

blurymind:
@turtletooth  that would be ideal! 👍 

Btw there is a new feature request that ties in very well with this feature here:
https://github.com/opentoonz/opentoonz/issues/1423

in the request, I suggested some tweaks to the current gui for inbetweening:

> Instead of having to select any frames in the level strip to define what the first and last drawing are for the tween preview and generation, we just use what vector guided mode is using, like this:
> 
> vectorguidedtweenpreview
> 
> This is a suggestion that can be addition to the current workflow if desired by the devs. I mean that OT could check if vector guided mode is enabled - and if so decide to use what vector guided mode is showing instead of the level strip selection - to determine the start and end frame for tween preview or tween generation.
> 
> You could even add the preview tween and set tween commands added to the Vector Guided drawing menu like this:
> vectorguidedtweenpreview2
> 
> That would reduce the number of steps required to make tweens even more, but it will also make it more intuitive as well, without negating the current workflow

Right now the user has to manually create the inbetween empty frames and select the strip frames. My suggestion is to set the number of inbetween inside the inbetween window, and let opentoonz handle the creation of empty frames and their population instead.

The feature request is for an inbetween preview command, that fits in perfectly with the feature here. The idea for it comes from cacani's quick flip command:
https://www.youtube.com/watch?v=1LIrPyaG03o

On the previewing side of things, in my original feature request here:
https://github.com/opentoonz/opentoonz/issues/565

One of the examples from other software had a fancy line rendering for the preview line:
sequentianonionskin

That would be cool to have too, but not sure how difficult it would be. Honestly your implementation with the x-mark does an excellent job already :)

beeheemooth:
@turtletooth but what about short vectors in this case?

turtletooth:
@beeheemooth The line in my implementation shows up on the first 20% of the line, no matter how long it is.  

@gab3d For circles or closed shapes, it is important to show which direction the line should be drawn in.  If the line is drawn in the wrong direction, auto-inbetween won't work right.

blurymind:
@turtletooth for circles the X will not very well show the direction-especially if the circle is closed, so I guess a combination of an X and a line hint would be ideal 

Now I begin to understand why that other software has the fancy line rendering preview

RodneyBaker:
I'm a fan of the directional hint as I've run up against not knowing which way to proceed.

Now you've guys have got me wondering if the percentage of that direction preview was made to change from say, 0 to 33 to 66 to 100 in rapid succession**, an animated preview of direction might work.  I think keeping it simply green might be easier on the eyes rather than cancani psychedelic.  ;)
It might slow down operations a little because of the computations but animating the percentage is an intriguing idea.

**Hitting 0 and 100 twice might also allow for a slight delay so that users better sense the origin and target locations.

turtletooth:
[guided_animation.zip](https://github.com/opentoonz/opentoonz/files/1279118/guided_animation.zip)

Note: the animation only moves when the cursor moves.  At first I thought the animated lines would be unnecessary, but I did find them to be actually helpful in guiding the stroke.  

The only issue with this implementation is that since they only move when the mouse moves (which I can't control from where this happens in the program), if the mouse stops while the lines are in the middle of the main line, that's where the animation stops until the mouse moves again.

blurymind:
@turtletooth I really like it =)
Perhaps direction can be even more strongly suggested if the lined moved slightly faster, but at half the distance per prame

RodneyBaker:
Very nice Jeremy.  :)

turtletooth:
How about this:
[guided2.zip](https://github.com/opentoonz/opentoonz/files/1279264/guided2.zip)

artisteacher:
I like it!

blurymind:
@turtletooth this is it! :D That  line effect really clearly shows the direction now, and it's not as distracting as I thought it would be, you made it subtle but functional

openanim:
@turtletooth  maybe I'm exaggerating you ask for extra work even though you just improve your superb guided mode but personally I find drawing on tweenmaker tires me a lot more eyes with the line moving after some hours. In tweenmaker it is possible to lower the speed so as to have something close to 1frame by frame, I will like 2 things in relation to your new functionality if possible:

- or there is an option to enable or disable the flow of direction

- or as in tweenmaker, it is possible to adjust the speed of the directional flow until almost canceled or even see that the very beginning of the vector.

https://1drv.ms/i/s!AspsYffPSwbdiKNZitcnNPqADX4xyg

This may seem ungrateful, I apologize for these extra requests :( because i love this new improve.

blurymind:
I agree that having these variables exposed to the preferences window would be quite nice, as it will be impossible to please everybody and in some rare occasions it might help usability

konero:
@turtletooth Nice work!

I'm in the same boat as @openanim, how about the way CACANi does it where both ends highlighted with blue, inside has big arrows indicating direction.

cacani-imo

Would this also have the ability to skip to next or prev stroke? I got the chance to work on a production using CACANi, that made all the difference between a work looking like it was either made in 2004 or 2017.

Perhaps some kind of status indicator somewhere?

strk-no

Depending on style, obviously, there can be hundreds of strokes, and often you need to skip or navigate back, otherwise you have to start the frame all over again if you make a mistake, such as needing to adjust just one area.

caca-an-9556

Also the ability to flip stroke direction, thats the most common mistake I usually make even with CACANi's obvious indicators telling me which way to go.

Sorry for chiming in with so much, feel free to ignore! Great work already. 👍

turtletooth:
@konero I originally wanted to do that, but I couldn't figure out the math to calculate the angles for the arrows.  I agree that the arrows would be clear without the distraction that an animation could be.  

I might be able to put the animation into preferences, but I think if we can do arrows instead of adding more things to the preferences, that would be best.

@shun-iwasawa I realize you are busy, but do you have any idea how I might figure out the end points for the arrows?  I've tried a handful of things, but my math in this area is a bit rusty.

turtletooth:
@konero @blurymind Unfortunately, I think fixing this is as far as I will be taking this feature for now.  After this, I want to work on the cutout side of things and help improve that workflow.

turtletooth:
I made the animations an option in the guided drawing menu:
show_animations

Without the animations, it will show the x and a line coming from the x in the direction you are supposed to draw (as shown above).

blurymind:
@turtletooth thank you for keeping the animated preview as an option :) I personally don't find it distracting at all and really like it - as it clearly shows the direction in all use cases and it doesnt obscure the lines in any way

openanim:
@turtletooth Thanks for this nice option. 

@blurymind I also share the approach of @konero about it but I am very happy with this option. 
 Even if it's softer, as you said, I find it a bit disturbing, and especially with the mode "All drawings of the guided drawing" it would be possible to see a fireworks : ) but now it's good for me.

Btw, about preview command, your second alternative to your first one is also pleasant.

turtletooth:
> If "Show Animations" is a preference - why not move it to the preferences window? Also give it a more descriptive name

The preferences window is getting really crowded.

turtletooth:
Hey look- Arrows!
arrows

I still kept the animation option too.  You get to choose either.

blurymind:
@turtletooth Nice! The arrows are awesome!  👍 

Made a quick and dirty mockup on the toggle thing:
vectorguidedonionskinprefs
The Onion Skin tab of the preferences is not as badly cluttered as some of the other ones..

I guess I would prefer having the option inside the preferences- as it is something that you wouldn't change very often. It is a personal preference, more than a workflow tool that requires switching between the two preview modes very often.

If you add another checkbox item with the checkbox items that are currently used to set the mode, it might confuse the user a little bit. Also naming it "show animations" is not explicit enough. If we are still dreaming to have a "preview tween" command over there - it is also showing the inbetween animation

turtletooth:
@blurymind I didn't realize the Onion Skin preferences had the room.  I don't often go in there.  It makes sense to put it in there.

shun-iwasawa:
@turtletooth  Thank you for adding this. It seems to be very useful.

> @shun-iwasawa I realize you are busy, but do you have any idea how I might figure out the end points for the arrows? I've tried a handful of things, but my math in this area is a bit rusty.

Although my solution is still old-fashioned OpenGL, I think the function `drawArrows()` can be rewrite as follows:

```
void drawArrows(TStroke *stroke, bool onlyFirstPoint) {
  double length          = stroke->getLength(0.0, 1.0);
  int points             = length / 20;
  if (points < 2) points = 2;
  double currentPosition = 0.0;

  TPointD prePoint, point, postPoint;
  glColor3d(1.0, 0.0, 0.0);
  for (int i = 0; i <= points; i++) {
    currentPosition = i / (double)points;
    point           = stroke->getPointAtLength(length * currentPosition);
    prePoint =
        (i == 0) ? point
                 : stroke->getPointAtLength(length * (currentPosition - 0.02));
    postPoint =
        (i == points)
            ? point
            : stroke->getPointAtLength(length * (currentPosition + 0.02));

    if (prePoint == postPoint) continue;

    double radian =
        std::atan2(postPoint.y - prePoint.y, postPoint.x - prePoint.x);
    double degree = radian * 180.0 / 3.14159265;

    glPushMatrix();
    glTranslated(point.x, point.y, 0);
    glRotated(degree, 0, 0, 1);
    glBegin(GL_LINES);
    glVertex2d(0, 0);
    glVertex2d(-3, -3);
    glVertex2d(0, 0);
    glVertex2d(-3, 3);
    glEnd();
    glPopMatrix();

    if (onlyFirstPoint) break;
    // make the arrow blue from the second one
    glColor3d(0.0, 0.0, 1.0);
  }
}
```

Please note that you need to modify `drawFirstControlPoint()` as well in order to use transformation for each arrow:

```
@@ -340,7 +191,6 @@ void drawFirstControlPoint(const TVectorRenderData &rd, TStroke *stroke) {
   glPushMatrix();
   tglMultMatrix(rd.m_aff);
   if (!rd.m_animatedGuidedDrawing) glLineWidth(2.0f);
-  glBegin(GL_LINES);
   glColor3d(0.0, 1.0, 0.0);
   if (!rd.m_animatedGuidedDrawing) {
     drawArrows(stroke, false);
@@ -359,6 +209,7 @@ void drawFirstControlPoint(const TVectorRenderData &rd, TStroke *stroke) {
     drawArrows(stroke, true);
     glColor3d(0.0, 1.0, 0.0);
     j = 0.025 + modifier;
+    glBegin(GL_LINES);
     // draw the first animated section
     for (int i = 0; i < 8; i++) {
       endPoint = stroke->getPointAtLength(length * j);
@@ -389,8 +240,10 @@ void drawFirstControlPoint(const TVectorRenderData &rd, TStroke *stroke) {
         startPoint = stroke->getPointAtLength(length * j);
       }
     }
+    glEnd();
   }
 
+  glBegin(GL_LINES);
   // Draw an X over the first point
   if (rd.m_animatedGuidedDrawing && false) {
     glColor3d(1.0, 0.0, 0.0);
```

openanim:
Thanks very much for theses options, theses arrows will be very useful.
OT now allows great flexibility in tweening workflow. One can choose the type or mix the different modes according to need for more efficiency. From copy-paste to point-to-point, a unique and very fast guided frame range, a guided drawing that continues to improve and can be compatible with the cut-out too...thank you for allowing this, and continue to improve things. 

@blurymind  True, and what about Moho that will be even more. 
You've made  huge improvements to OT though it's not his specialty.

turtletooth:
@shun-iwasawa You are brilliant.  My version was pitiful compared to yours.  I suppose that's what I get for not using any trigonometry in 20 years.  Thank you so much for your implementation.

I also moved the viewing option to preferences, with the arrows being the default.

Thank you all for your feedback.

turtletooth:
@shun-iwasawa I don't know if you have the urge to be a superstar twice in one day :) , but #1308 is pretty close, but just has a few bugs.

shun-iwasawa:
I know that the modification like #1308 is inevitable in the future... I and my colleague once tried to transfer to "modern" style of openGL but could not succeeded. Thank you for tackling this matter!
For now I'm working on implementing the demanded feature by some animation studio, I'll watch it then.

turtletooth:
> For now I'm working on implementing the demanded feature by some animation studio, I'll watch it then.
Awesome! Can't wait to see it.  Thanks again for your help.

konero:
Wow, nice one guys!

shun-iwasawa:
@turtletooth 
> Awesome! Can't wait to see it. Thanks again for your help.

Oh, please do not expect anything for this matter..! I'm trying to read [this](http://www.opengl-tutorial.org/) and [this](https://www.slideshare.net/ICSinc/convert-your-legacy-opengl-code-to-modern-opengl-with-qt-ondemand-video) . 
According to the former tutorial, it says; 
> Forget Everything
> You don’t have to know anything, but you have to forget everything you know about OpenGL. If you know about something that looks like glBegin(), forget it. Here you will learn modern OpenGL (OpenGL 3 and 4) , and many online tutorials teach “old” OpenGL (OpenGL 1 and 2). So forget everything you might know before your brain melts from the mix.

... Well, it must be tough for "old" openGL guy like me.

turtletooth:
@shun-iwasawa You are light years ahead of me.  I'm just looking forward to shaders being usable.  That's all #1308 was done for.  There are some really awesome effects to be had in the shaders.

blurymind:
@turtletooth thank you for this awesome work. It's amazing to think how far OT has come ever since it was open sourced. Thank you for listening to the feedback from the community and making this software truly great. You can't see such close interaction between devs and users with proprietary software

shun-iwasawa:
@turtletooth Sorry for my late response, I think this can be merged now. Could you please resolve conflicts? Thanks!

turtletooth:
Conflicts resolved.  Thanks @shun-iwasawa

shun-iwasawa:
Jenkins

shun-iwasawa:
It must make the vector inbetween feature very useful. Thank you @turtletooth !
LGTM

PR #1300, opened 2017-06-24
Allow Imported Images to Use Camera DPI

turtletooth:
This creates a new setting in preferences to allow OpenToonz to ignore an images built in DPI and use the current camera dpi.  

It also corrects an error where images without a built in dpi were getting the stage::inch as a default instead of stage::standarddpi.


shun-iwasawa:
Jenkins

shun-iwasawa:
It works fine. Thank you for adding the new option and the fix!
LGTM

Issue #978, opened 2016-12-25
Scrubbing or Playing  Audio Causes Crash with Sierra


artisteacher:
Hi! So I just upgraded my home computer and now have OS X 10.12.2 (Sierra). Using the latest build, scrubbing audio now causes a crash - both in the x-sheet and the viewer. This is definitely not a problem with 10.11.4 or 10.11.6.

If I scrub from the beginning of the track, the crash is pretty instantaneous. If I scrub more from the middle the crash takes a little more time to occur. If I hide the audio columns, I can scrub the scene with no issue.

Thanks for all if the hard work you all do! Crash report is attached.

[sierra_crash_audio.txt](https://github.com/opentoonz/opentoonz/files/672016/sierra_crash_audio.txt)

artisteacher:
I've found that scrubbing directly in the x-sheet or with the viewer causes an immediate crash. There might be a slight delay in the crash if I start scrubbing slightly inward. If I have the audio column selected and try scrubbing through the combo-viewer, I can often scrub 2 times consecutively (on average) before a crash occurs at the start of the next consecutive scrubbing attempt. There is still the unfortunate audio distortion #941.

artisteacher:
So, it's not just scrubbing that causes a crash. Playing a scene with audio through the the viewer controls also causes a crash, though it is not as immediate. If preview has been toggled on in the viewer, I can play through the scene multiple times without issue. Scrubbing still causes an instant crash if preview is on. Audio does not cause a crash when rendering nor when previewing the audio itself in the top of the xsheet column.

artisteacher:
So I've found a thread on stack overflow for an audio program that looks like the same issue - the crash report looks the same anyway.  The crash with sierra was eventually resolved by the very end of the thread. I am hopeful that this can point someone in the right direction.

http://stackoverflow.com/questions/40580808/call-to-audioconverterfillcomplexbuffer-results-in-crashifclientprovidedbogusaud

artisteacher:
I found this message in console today. Maybe this is contributing to the various audio issues on OS X. 
![image](https://cloud.githubusercontent.com/assets/21159116/23825275/bbe66c9e-0654-11e7-971c-47bfd7ddca3f.jpeg)

artisteacher:
[...]

PR #1302, opened 2017-06-28
Fix for sierra scrub crash

turtletooth:
#978 
The Sierra Core Audio code (not the OpenToonz code) appears to have a known bug.  This fix bypasses the Core Audio implementation and uses Qt audio for playback.  

@shun-iwasawa I left the original code in case a Mac update fixes it and if performance is better with the Core Audio.  In my limited testing, performance appeared similar.

I don't mind taking out the original code if this seems to work fine.


turtletooth:
@shun-iwasawa Based on your work with the mac font issue, I went ahead and also removed the old code from this PR.  I suppose if we need it, it is in the git history.

The code in txshsoundcolumn could be useful on all platforms, not just windows to make sure that we get a format that is playable on the device.  I could take out the mac check if you agree.

artisteacher:
I'm really looking forward to this being merged!

morevnaproject:
We have same problem with Linux build. I wonder if your fix can help us as well. ^__^

turtletooth:
@morevnaproject I don't see why it wouldn't work.  My solution uses straight Qt audio (with OpenToonz built in audio converter) to find supported audio formats and play them.

morevnaproject:
Yes, I will test that. Thank you! ^__^

shun-iwasawa:
Jenkins

shun-iwasawa:
@turtletooth 
I tested the PR in OSX Sierra 10.12.4
For now it crashes when you do either one of the following operations:

- Click the speaker icon in the sound column's header in order to preview-play the whole column.

- Left drag the right side area (with horizontal stripes) of the sound column in order to specify the range for preview-playing.

turtletooth:
Thanks for catching that @shun-iwasawa .  I was so focused on scrubbing, I forgot to check those options.  I think I fixed it.

shun-iwasawa:
Confirmed the fix. Thank you @turtletooth !
Travis claims error with Linux build. I have no doubt it is not because of this PR itself, but please give me some time to investigate. Thank you for your patience.

turtletooth:
@shun-iwasawa Is this ready to merge?

shun-iwasawa:
@turtletooth Sorry for keep you waiting! Please let me try Jenkins again just in case.

shun-iwasawa:
LGTM Thank you @turtletooth for the fix.

Issue #1058, opened 2017-02-23
Strange pressure behavior when drawing with mouse on vector layer


mcjadski:
(Tested on **Windows 7**, OpenToonz 1.1.2 nightly build.  None of this seems to happen on Mac.)  

When I draw with the mouse on a Toonz Vector Layer when Pressure is checked, three strange behaviors result:  

- [x] The beginning of the stroke is covered up by this circle:  

![mouse_pressure_01](https://cloud.githubusercontent.com/assets/25377429/23243603/8fb3b7ec-f945-11e6-91d6-a7604ec09f03.jpg)

When I let go of the mouse button, the circle goes away.  However, this can be quite obnoxious to deal with if you're more zoomed in on the canvas, as it obscures your view:  

![mouse_pressure_02](https://cloud.githubusercontent.com/assets/25377429/23243639/baa4f79a-f945-11e6-9991-22a9b66df86b.jpg)

The size of the circle doesn't seem to be based on the Max or Min brush size.  The circle is only visible if the brush size is thinner than the blob shapes.  

Also, once you draw a mark with a pen tablet and then go back to using the mouse, this glitch goes away.  

---

- [x]   Sometimes, when I draw fast strokes, I get these teardrop shapes attached to the beginning.  These **do** remain after letting go of the mouse:  

![mouse_pressure_03](https://cloud.githubusercontent.com/assets/25377429/23243659/e61048a8-f945-11e6-983d-fe0988361056.jpg)

Again, the shapes only appear if the brush size is thinner than the blob shapes, and again, once you draw a mark with a pen tablet and then go back to using the mouse, this glitch goes away.  

---

- [ ]   **Before** drawing a stroke with a tablet, the mouse draws with the **Max** brush thickness.  **After** drawing a stroke with a tablet, the mouse draws with the **Min** brush thickness.   

For consistency, it should always use the **Max** brush thickness, because that's what it does when pressure is turned off, and that's what it does on a Mac (whether or not pressure is turned on or a tablet has been used.)  

EDIT:  This last glitch also occurs on Toonz Raster Levels.

artisteacher:
I've had the same issues with OS X 10.11.6 and 10.12.3

artisteacher:
Here's a gif that shows the glitch in action - it really can be hard to see what you are doing. This is on OS X. 
[...]

Issue #618, opened 2016-07-10
Bug: Menu items disabled on OS X


yamisama:
The Quit menu item is often grayed out on OS X (10.9.5), either right after launch, or after performing a few actions in the UI. When the main viewer has focus, pressing Cmd+Q still invokes the Quit command. Tested with a customized room set based on the Ghibli preset, stylesheet set to "gray_128_mac".

Sometimes after a longer time of working in the software, not just the "Quit" command, but pretty much all menu items become grayed out.

This issue has existed for many builds, including recent nightlies and regularly occurs.

Also, the Undo and Redo commands are only labelled "Undo" and "Redo" and do not show the name of the corresponding action like they do in other applications, eg. "Undo Brush Tool" etc.. The History panel shows labelled commands. Not sure if this is a menu handling bug or by design.

artisteacher:
The quit menu icon is always grayed out. I've also had several students where all the menu items are greyed out when they are ready to save. This is with OpenToonz 1.1.1 on the OS X El Capitan.

artisteacher:
This is still happening as of 1.1.2 - but only when using wacom tablets.

artisteacher:
So I've found that quit in the menu is disabled - unless the selection tool is active. However in that case, neither the quit menu nor the quit shortcut actually work. OpenToonz does not close. If the scene was saved before trying to quit with one of those methods (with the selection tool active) an asterisk appears next to the scene name as if there was a change made. When the selection tool is not active, the quit shortcut works but quit in the menu is greyed out.

artisteacher:
The entire menu still becomes greyed out and inaccessible when using a wacom tablet on a mac.  I'm thinking the issue may have to do with incomplete/interrupted signals from the pen. I definitely press against the buttons unintentionally fairly regularly. Disabling both buttons in the tablet's settings (in system preferences) seems to be making a big difference so far. No more disabled menu! It is mildly annoying to have to switch to the mouse or touchpad to right-click but at least the menu is staying accessible.

I wonder if this could be related to #860. 

This has no impact on quit being constantly greyed out and broken, since it never works, even without a tablet.

artisteacher:
With the tablet fix in the recent nightly build, the menus seem to have stopped becoming inaccessible. However quit in the menu still is broken.

artisteacher:
In the Morevna release, the menu button is disabled as well.

artisteacher:
[...]

PR #1304, opened 2017-06-29
Fix for Pen Glitch on Mac

turtletooth:
#756

The way that tablet detection was handled for tool events was a little sloppy.
I performed more careful checks about whether an event is from a tablet or mouse and which event type it is.


artisteacher:
Thanks so much! Does this have any impact on the menus becoming greyed out and inaccessible during prolonged tablet use - part of #618?

turtletooth:
Checked on Windows.  Appears to work correctly there too.

turtletooth:
@artisteacher I don't think this has anything to do with the menus, but I have no idea really.  These past two PRs are the first time I've done any development on the mac.  And it has been painful.  The mac I have that can actually run Sierra is the last model that Apple allows to run Sierra - and it is slow.

artisteacher:
I do think something related to the tablet input causes the menus to become disabled. Have you run into this yourself? It is most likely to happen when there is a vector level in the scene (even if not actively selected) and autosave is on. It is similar to when the startup or about popups is open and (with the exception of services, the hide options, and search) the menus are inaccessible. Though with the tablet menu glitch, "show tab bar" and "enter full screen" are also still accessible. In this case trying to close out of OT with the close button (the red x), gives an alert sound. It's like the program thinks something is active and is waiting for input. 

The other frequent tablet issue distorts the beginning of brushstrokes after a tablet has been used and disconnected #1058.

Anyways, thanks for tackling some of the mac specific issues despite the slow computer! The fixes you've been working on are really necessary to make using OT on a mac more feasible.

shun-iwasawa:
@turtletooth Thank you very much for tackling this problem.
I tried the PR and found that the problem still seems to occur.
I've modified and posted another PR #1360 . Please check if it will solve the problem.

PR #1307, opened 2017-07-01
Vector - Shift for Straight Line

### LINKED ISSUE #639: [bug/feature] zero thickness lines - need to appear in onion skin - need to be more visible as you draw them - and other features
The zero thickness lines feature is nice - it is however missing some much needed details!

essential feature:
- [ ] Zero thickness lines need to be more visible when onion skin is activated - right now they almost disappear if your style is in a darker shade of a skin tone
- [x] Zero thickness lines need to be visible as you draw them - right now you cant see them until you release the stroke. This is very bad - as it makes it impossible to take advantage of @walkerka 's brush smooth feature

nice to haves:
- [ ] A setting in preferences to control their display width
- [ ] A setting in preferences to overwrite their display draw color

Perhaps @turtletooth might be interested? :)

turtletooth:
This improves the vector brush in two ways:
You can now hold shift to draw a straight line - No need to switch over to geometry tool
Zero thick lines are now visible as you draw them.  This partially fixes #639

### DIFF
--- a/toonz/sources/include/toonz/strokegenerator.h
+++ b/toonz/sources/include/toonz/strokegenerator.h
@@ -83,6 +83,9 @@ In caso affermativo li cancella.
   //! Visualizza tutti i frammenti
   void drawAllFragments();
 
+  // Only keep first and last points. Used for straight lines
+  void removeMiddlePoints();
+
   //! Restituisce il rettangolo che contiene la regione modificata
   TRectD getModifiedRegion() const;
 
--- a/toonz/sources/tnztools/brushtool.cpp
+++ b/toonz/sources/tnztools/brushtool.cpp
@@ -27,6 +27,7 @@
 #include "toonz/toonzimageutils.h"
 #include "toonz/palettecontroller.h"
 #include "toonz/stage2.h"
+#include "toonz/preferences.h"
 
 // TnzCore includes
 #include "tstream.h"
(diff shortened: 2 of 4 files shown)


blurymind:
Thank you! :D

shun-iwasawa:
LGTM for this feature.

@turtletooth 
Yesterday some Japanese animation studio pointed that the frame scrubbing feature ( #994 ) may triggered by mistake when they work with Shift + Raster Fill tool and pan in the viewer (with middle drag) rapidly. Note that Shift + Fill tool is used for fill the areas with the maximum fill depth.
This behavior is not a bug but unwanted, so I was considering to make the frame scrubbing feature to be optional in the preferences.

With this PR the vector brush will use the shift key as well. 
If there will be the cases that other tools may use shift key for extra feature, how about changing the frame scrubbing feature to use another modifier key or combination of them? 

Jenkins

turtletooth:
@shun-iwasawa Would changing shift to ctrl in #994 be adequate?  I think leaving the ctrl+shift option for steps would be fine.

shun-iwasawa:
Thank you for the quick response. Well, IMO ctrl+shift would be safer as I'm not sure any tool is not using ctrl.
( It would be helpful for beginners if the instruction of all the modifier features for each tool is displayed in the future. )

turtletooth:
I agee Ctrl + Shift would be safer.  What would you do about the next/prev step command in #994 which already uses Ctrl + Shift?

shun-iwasawa:
Hmm, how about setting Ctrl+Alt+wheel for next/prev step?

turtletooth:
Sounds good.

shun-iwasawa:
@turtletooth Thank you for this feature and posting the fix for shortcuts!
LGTM

Issue #1081, opened 2017-03-03
Full screen mode stops working correctly after the first use


artisteacher:
Full screen mode works the first time I use the shortcut or the visualization pop-up menu. Exiting full screen mode with the shortcut only works after a tool has been used. Returning to full screen mode with the shortcut only works after using a tool. Using the pop-up menu to leave or return to full screen works without using a tool.  

After entering and leaving full screen mode the first time in a room, returning to full screen mode works incorrectly. The workspace becomes a floating window instead of going full screen. The window can be maximized which then enters true full screen mode. The window can also be closed completely - which results in a viewer with no workspace - obviously problematic. Switching rooms returns full screen mode to the expected behavior, but only for the first use. Changing scenes and projects has no impact. 

I'm using OS X 10.12.3 and the latest nightly build. I know the Harlequin 7.1 manual says that this is a Windows only feature, but since it works the first time it ought to be possible to make it completely functional. Thanks for all the great progress!

artisteacher:
Right now, fullscreen mode (like normal fullscreen) causes an immediate crash in OS X 10.11.6 & 10.12.5.

artisteacher:
This is working properly now! Thanks @turtletooth, @shun-iwasawa & @blackwarthog!

turtletooth:
Great!

PR #1308, opened 2017-07-01
WIP: Fix for Shaders and Switch to QOpenGL

turtletooth:
This is still pretty buggy, but shaders don't instantly crash OT anymore.  It works, but throws a lot of assert errors.

shaders

@shun-iwasawa I have no experience doing OpenGL programming and would love some help.


turtletooth:
So here's the limit of what I can find:

I keep getting a "cannot send events to objects owned by a different thread" error around lines that contain:
```
m_imp->m_context->create();
m_imp->m_context->makeCurrent(m_imp->m_surface);
```
Ignoring this doesn't crash the program.  I have tried a variety of fixes but my limited knowledge of threading and/or opengl isn't able to fix the problem.

There is also a problem in the destructor if sceneviewer when trying to clear out l_proxy and l_contexts.

I am going to stop here and hope someone is able/willing to pick up the pieces since I seem to be at a dead end.

turtletooth:
@blackwarthog - would you have any time to take a look at this?  Having usable shaders is so close, but I can't figure out the last bit.

blackwarthog:
At my system I've seen only warnings about wrong thread of context.

Seems you was at the right way, i just removed two other MACOSX preprocessor condiditions to call moveToThread in ShadingContext destructor and in ShadingContext::doneCurrent too:

https://github.com/blackwarthog/opentoonz/commit/1213cfb37b6e7e730f828befbd443de9860aa119

And I've no more warnings about context thread. I've checked it linux, please, tell me how it works at windows.

Also at windows and linux uses OpenGL 2.0 (by default) but for MacOS uses version 3.2. If some shaders requires OpenGL 3, then we may to select OpenGL 3.2 for all platforms (in function ShadingContext::Imp::format()).

turtletooth:
With the new changes I get an instant crash, no assert errors or anything.  It doesn't even trigger Visual Studio where it crashed.  It just goes down hard.

blackwarthog:
ok, i'll try to check it on Windows.

blackwarthog:
I cannot test MSVC build at PC with graphics accelerator. I've tested MinGW build and it works at Win7 64 and GeForce 9600GT. You may to download it from from here: https://dev.icystar.com/downloads/OpenToonz-1.1.3-morevna-2017.08.08-3ea6d-win-64bit.exe

blackwarthog:
May be latest changes in master affects shaders behavior. You may try to do rebase to latest master.

RodneyBaker:
@blackwarthog  This is off topic and I apologize for that.  I'm not sure where exactly to post this.

When installing the Morevna Project version of OT I always get a message that I have to uninstall the previous installation of OT, and I can't install without doing so.  This is problematic in several ways as you might imagine.  Can we get the Morevna build to install into another directory or in such a way as not to interfere with other installations of OT?  This largely prevents my use of the Morevna branded OT because I don't want to uninstall the official release (or for that matter any other install).

turtletooth:
@blackwarthog I checked your build and it works on the first 6 shaders, but the blur ones cause an instant crash still.  I did a build rebased with master and added your changes and if I run a release build, it works, but if I run a debug build, shader preview instantly takes OT down.  I'm not sure why debug would be different than release.

Thanks for your help with this.  I really appreciate it.

shun-iwasawa:
I still do not the way to fix the problem, but confirmed that this PR as-is will fix the issue #1081 (crash with full screen mode on OSX) .

artisteacher:
If this could fix all the fullscreen crashing, including #1396, that would be so amazing!

turtletooth:
Closing - Updated in #1517

Issue #1093, opened 2017-03-10
Closed vector shape is unfillable


mcjadski:
On Win 7, I did some experimenting with [unfillable.zip](https://github.com/opentoonz/opentoonz/files/816428/unfillable.zip) , courtesy of @shun-iwasawa.  For reference, here's the starting position.  The left region is indeed unfillable;  I left the fill-check off for the following screenshots, as it's quite interesting what happens.

![unfillable_start](https://cloud.githubusercontent.com/assets/25377429/23776461/1d95a764-04f4-11e7-8f45-f3287dd6d780.jpg)

If you adjust the angle of one of the handles like this, the fill swaps to the other region!  (The color is whatever color you had in the other region.)  

![unfillable_swap](https://cloud.githubusercontent.com/assets/25377429/23776547/9ccf494a-04f4-11e7-9bf6-030ec93d54e5.jpg)

If you adjust it to this position, the whole thing becomes unfillable:  

![unfillable_none](https://cloud.githubusercontent.com/assets/25377429/23776599/d5ed768e-04f4-11e7-819c-89db66179159.jpg)

Here, the whole thing becomes filled!  I didn't fill it, it just became all the one color:  

![unfillable_all](https://cloud.githubusercontent.com/assets/25377429/23776627/033b1dc6-04f5-11e7-8811-cfb7c64ce5f7.jpg)

You can color them differently then:  

![unfillable_different](https://cloud.githubusercontent.com/assets/25377429/23776637/141f25f6-04f5-11e7-9b79-273b6bb255dc.jpg)

As you tweek it slightly, the left region will sometimes flash between filled and not filled.  (NOTE:  If you color the left region red, then tweek it so the left region becomes unfilled, then tweek it so it becomes filled again, it will be black, not red.)  

![unfillable_near_all](https://cloud.githubusercontent.com/assets/25377429/23776685/63ea290a-04f5-11e7-9e99-e52df2016f22.jpg)

Making the image large using the Edit Tool has no effect:  

![unfillable_big_edit](https://cloud.githubusercontent.com/assets/25377429/23776694/7379a9fe-04f5-11e7-9bd9-5c7d69c937f6.jpg)

However, making it bigger with the Selection Tool makes the whole thing filled:  

![unfillable_big_selection](https://cloud.githubusercontent.com/assets/25377429/23776705/85fac4dc-04f5-11e7-9dc4-15049db91f59.jpg)

Moving it around with either tool seems to have no effect.

[...]

PR #1317, opened 2017-07-04
Fill Tool: Allow Fillable Gap Size Change in Vector Images

turtletooth:
This adds a new option to the vector fill tool- maximum gap size.

gap

In OpenToonz the gap size tolerance is stored in each image, not just on the level.  This PR allows that value to be changed in the fill tool.  Different images on the same level can have a different value.

This also takes care of some of the fill issues such as #1093.  Adjusting the slider makes that shape fillable.  The benefit of this solution is that it only affects images where the slider has been changed.  It doesn't change anything about how gaps, intersections and fillable regions are calculated on a permanent basis.  So there will be no adverse effects on existing work.

Note: Going to a high value can be useful at times but sometimes causes unexpected results. 

Note++: This change will make vector levels created in new versions incompatible with older versions.  This adds a new tag in the vector file format for maximum gap size which older versions won't know how to deal with.


mcjadski:
Hey, that's awesome!!  

"Note: Going to a high value can be useful at times but sometimes causes unexpected results."

Hmm...  What kind of unexpected results?  Do you mean additional glitches, or more just that if the allowed gap is too large, the image will fill in undesirable ways (e.g., half of a shape getting filled in)?

RodneyBaker:
I can confirm (via @turtletooth 's experimental release that this feature works great and helps to resolve a majority of gap/fill issues.

Thanks much!

blurymind:
I am so going to play with this when i get back home :D @turtletooth has
packed so many incredible new features

turtletooth:
@shun-iwasawa What are your thoughts on this?

shun-iwasawa:
I'm a bit concerned about that, according to your comment, it loses compatibility with the older versions.
I don't yet properly understand the modification, but is there a possibility to keep the compatibility at least if you do not adjust the "maximum gap size" parameter from the default value?

turtletooth:
Right now, it will load scenes from older versions the same as always, but vector levels created i this will have a special tag in the vector file that older versions don't know how to handle and will crash on.  

I'm not sure how to handle that since I can't retroactively change how older versions load vector levels.  I think that at the very least a strong announcement should be made upon release.

turtletooth:
How about this- I can make it so the tag isn't written to the vector file if the setting hasn't been changed.

toonz/sources/image/pli/tiio_pli.cpp
```diff
@@ -541,7 +550,13 @@ solo nel costruttore)
...
     tags.push_back((PliObjectTag *)tag);
   }
-
+  // Store the auto close tolerance
+  {
+    int tolerance =
+        (int)((roundf(tempVecImg->getAutocloseTolerance() * 100) / 100) * 1000);
+    PliTag *tag = new AutoCloseToleranceTag(tolerance);
+    tags.push_back((PliObjectTag *)tag);
+  }
```
shun-iwasawa: Sorry if I'm wrong, but skipping this scope (if the parameter is unchanged from the default) can save PLI compatible with older versions?

turtletooth: Yes.  That's what I am building in.

shun-iwasawa:
@turtletooth 

> How about this- I can make it so the tag isn't written to the vector file if the setting hasn't been changed.

Oh, yes! I think it is ideal.

turtletooth:
Would you be agreeable to this change as long as compatibility is preserved if the slider isn't touched?

shun-iwasawa:
@turtletooth 
Again, I'm sorry I still do not review the source at all ( I will do it soon ! ) , but I think it would be more agreeable if this option is really optional - in other word, without touching the slider, they can use the software exactly as the same as before.

turtletooth:
I made it so that if the slider is not touched, images will continue to load properly.  I also tested and images that do have the value changed don't crash the program, they just don't load.

turtletooth:
Fixed conflicts.

turtletooth:
Conflicts resolved

turtletooth:
@shun-iwasawa What are your plans regarding the next release?  I have a few PRs that I am hoping make it in.

shun-iwasawa:
>I have a few PRs that I am hoping make it in.

Can you please list PRs you would like to be included?
My plan toward the next release would be as follows:
- Review #1570 
- Review your PRs in the list
- Make the touch control to be optional, requested in #1579 
- ---Freeze features---
- Spend approx. a week for bug tracking and updating translations. I'm planning to tackle with the tablet control overhaul, especially with OSX.
- Release

I'm sorry that the release date will be behind schedule for a week or two.

jpturcotte:
Improved/fixed tablet support for macOS would be a godsend!

turtletooth:
@shun-iwasawa The one's I use often are 1602, 1353, 1317, 1028 and 744.  Changing the name of the edit tool to animate is the easiest to review- but it is a pretty big change.  It does make the shortcuts make more sense for eraser and edit tool.

shun-iwasawa:
Thanks @turtletooth for the list, I'll review them.

toonz/sources/tnztools/filltool.cpp
```diff
@@ -2032,11 +1932,26 @@ bool FillTool::onPropertyChanged(std::string propertyName) {
...
+  else if (propertyName == m_maxGapDistance.getName()) {
+    if (TVectorImageP vi = getImage(true)) {
+      if (m_maxGapDistance.getValue() != vi->getAutocloseTolerance()) {
+        vi->setAutocloseTolerance(m_maxGapDistance.getValue());
+        int count = vi->getStrokeCount();
+        std::vector<int> v(count);
+        int i;
+        for (i = 0; i < (int)count; i++) v[i] = i;
+        vi->notifyChangedStrokes(v, std::vector<TStroke *>(), false);
+        if (m_level) m_level->setDirtyFlag(true);
```
shun-iwasawa: How about calling `TXshLevelHandle::notifyLevelChange()` as follows in order to update the Level Strip title bar?:
```
--------------------- toonz/sources/tnztools/filltool.cpp ---------------------
index e7a4095..67bfc52 100644
@@ -1947,7 +1947,10 @@ bool FillTool::onPropertyChanged(std::string propertyName) {
         int i;
         for (i = 0; i < (int)count; i++) v[i] = i;
         vi->notifyChangedStrokes(v, std::vector<TStroke *>(), false);
-        if (m_level) m_level->setDirtyFlag(true);
+        if (m_level) {
+          m_level->setDirtyFlag(true);
+          TTool::getApplication()->getCurrentLevel()->notifyLevelChange();
+        }
       }
     }
   }
```
Also, IMO it should have undo for modifying this value as it directly modifies the level data.
( The undo should be stored only on mouse release, but I'm not yet sure how to. )

toonz/sources/image/pli/pli_io.cpp
```diff
@@ -1808,6 +1824,10 @@ void ParsedPliImp::writeTag(TagElem *elem) {
   case PliTag::PRECISION_SCALE_GOBJ:
     elem->m_offset = writePrecisionScaleTag((PrecisionScaleTag *)elem->m_tag);
     break;
+  case PliTag::AUTOCLOSE_TOLERANCE_GOBJ:
+    elem->m_offset =
+        writeAutoCloseToleranceTag((AutoCloseToleranceTag *)elem->m_tag);
+    break;
```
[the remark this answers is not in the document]
turtletooth: That was my first thought initially as well.  But this controls the gap size for the whole level.  This change allows different drawings in the same level to have different gap sizes.

turtletooth:
@shun-iwasawa I figured out getting undo to work (only when the slider is released) and made the requested changes.

toonz/sources/tnztools/filltool.cpp
```diff
@@ -252,6 +252,99 @@ class VectorRectFillUndo final : public TToolUndo {
...
+      , m_newGapSize(newGapSize)
+      , m_oldFillInformation(oldFillInformation)
+      , m_vi(vi) {
+    TTool::Application *app = TTool::getApplication();
+    if (app) {
+      m_row    = app->getCurrentFrame()->getFrame();
+      m_column = app->getCurrentColumn()->getColumnIndex();
+    }
+    ImageUtils::getFillingInformationInArea(vi, m_oldFillInformation,
+                                            vi->getBBox());
```
shun-iwasawa: I'm not sure why you update `m_oldFillInformation` here.
It seems that `vi` is already modified with new gap size and calling `getFillingInformationInArea()` overwrites the old fill information with the new one.

toonz/sources/image/pli/pli_io.cpp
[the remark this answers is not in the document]
shun-iwasawa: Understood.

shun-iwasawa:
One more suggestion:
How about incrementing the file format version only if it has modified gap size?
For now it is set to `c_majorVersionNumber = 71` and `c_minorVersionNumber = 0` defined in `pli_io.cpp`. Which means that the current PLI format is compatible with Toonz Harlequin 7.1 . By setting the greater number (say, `c_majorVersionNumber = 120` and `c_minorVersionNumber = 0` considering that the new format will be available from OT v1.2), opening the file with older versions will just open a warning popup without crash.

toonz/sources/tnztools/filltool.cpp
turtletooth: Thanks for catching this.  It was from a previous attempt.

turtletooth:
@shun-iwasawa I got the version change in there.  If the value of the gap size is not changed, the version is set to 71 to maintain compatibility with older versions.

shun-iwasawa:
I found there still be problems as follows:

1. Undo of changing the gap size sometimes fail to revert the paint. Please do the following with the uploaded sample below:
[gap_test.zip](https://github.com/opentoonz/opentoonz/files/1516467/gap_test.zip)
  
    - Load `gap_test.pli` and change the gap size from 2.56 to lower value (say 1.8) until the fill regions will be unified.
    - Undo it.

1. Switching the levels with different gap sizes fails to update the tool option correctly.
gap_size_problem

turtletooth:
Undo is always the death of me.

turtletooth:
Do you have any idea why might be causing the first issue you mentioned?

shun-iwasawa:
Unfortunately I have no idea yet.
If it may need some time to investigate, would you mind if I postpone merging this after releasing the next version ? I think I can take a look further later on.

turtletooth:
@shun-iwasawa Would you try the changes I made?  I think I figured it out.

shun-iwasawa:
Oh, thank you for the fix! I'll check it.

shun-iwasawa:
Unfortunately I faced a case that undo of changing the gap fails to revert. Also found that the changes does not update the thumbnails in the level strip or in the xsheet.

gap_size2

Hope you can reproduce the problem on your end. I couldn't find the precise condition yet.
I'm sorry to say but I think this feature should be postponed to v1.2.1 . As this changes the file format, careful bug tracking on nightly builds for a certain period seems to be needed.

turtletooth:
I understand @shun-iwasawa - Thank you for looking at it.  I agree that it can wait until 1.2.1.  Once 1.2 is made, do you think you might be willing to look at vector fill in general and try and make it more predictable?

Thanks again for all your time on this.

shun-iwasawa:
Thank you @turtletooth for your kind understanding. I'll take a look on this further later on.

turtletooth:
After 1.2, I think I fixed the new errors you noticed.

blurymind:
@turtletooth @shun-iwasawa I posted an issue that is related to this one:
https://github.com/opentoonz/opentoonz/issues/1636

Can we please make the fill gap size NOT be affected by line thickness (optionally or not)?
Right now it is creating many problems and I can't think of a single advantage in having it the way it is
33514768-9bfa6cae-d751-11e7-8838-ba500205ca58

artisteacher:
I’m really looking forward to this!

shun-iwasawa:
Jenkins

shun-iwasawa:
@turtletooth 
I'm terribly sorry for my late response. Confirmed the errors are fixed now. Thank you!

Sorry for the late notice, but I found another problem as follows:

- Create a new toonz vector level with 2 frames.
- Change the `Maximum Gap` of the 1st frame. Keep the `Maximum Gap` of the 2nd frame unchanged from the default value.
- Save the level.
- Start a new scene and load the level.
- Check the `Maximum Gap` of the 2nd frame.

On my environment, the `Maximum Gap` of the 2nd frame unexpectedly becomes the same value as the 1st frame.

turtletooth:
@shun-iwasawa I found the issue.  The gap size tag wasn't being set if the gap hadn't been changed.  If there is no tag, the frame inherits the gap size of the level, which is based off the first frame.  I made it write the tag in two cases: if the frame's gap size was changed or if the level (first frame) gap size was changed.

shun-iwasawa:
@turtletooth 
Thank you for the fix. I found the fix #1848 is also needed to make this work correctly. Can you please take a look ?

Jenkins

turtletooth:
You bet.  Once this is merged, I can merge that right away.

shun-iwasawa:
Thank you @turtletooth for the nice feature and modifications done so far !
LGTM

PR #1323, opened 2017-07-10
Fix to save the column icon loading policy properly on the non-English environment

shun-iwasawa:
For now the `Preferences -> Loading -> Column Icon` option is not saved properly on non-English environment.


turtletooth:
LGTM.  Thanks @shun-iwasawa

PR #1328, opened 2017-07-13
Geometry Tool Vector Snapping for All Tools

turtletooth:
On vector levels ALL GEOMETRY TOOLS can use snapping.  This allows for a great degree of precision.


blurymind:
Thank you! :D

openanim:
@turtletooth So happy ! Thank you, it's going to be exciting to try it :)

toonz/sources/tnztools/geometrictool.cpp
```diff
@@ -380,6 +402,8 @@ class PrimitiveParam {
     m_capStyle.setQStringName(tr("Cap"));
     m_joinStyle.setQStringName(tr("Join"));
     m_miterJoinLimit.setQStringName(tr("Miter:"));
+    m_snap.setQStringName("Snap");
```
yamisama: Not sure if this is intentional, but you might be missing the `tr()` here for localization.

turtletooth: Good catch.

turtletooth:
I'll add guide snapping to this too once the brush snapping is merged.

toonz/sources/tnztools/geometrictool.cpp
turtletooth: Fixed

turtletooth:
Snapping now works with guides on Geometry Tool too.
This one's ready.

shun-iwasawa:
Jenkins

shun-iwasawa:
It works so nice !
How about adding the snapping behavior to `CirclePrimitive::leftButtonDrag()` and `PolygonPrimitive::leftButtonDrag()` as well ? Both of them only the start point snaps for now.

beeheemooth:
How about a snap using the Control Point Editor Tool?

turtletooth:
@shun-iwasawa That was intentional since snapping sets the radius of the shape, which sometimes is a bit harder to get what you want.  But I played with it some more and think that it could definitely be useful, so I added it back in.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. @turtletooth thank you for the useful feature!

phillycheese007:
When you say "Snapping now works with guides .." are you talking about the field guide?  I see the snapping to other shapes but cant figure out the guide snapping part

shun-iwasawa:
@phillycheese007 
The guides mean not the field guide, but the ones which are added by clicking in the rulers on the top and on the left sides of the viewer.
Please see `To add a custom guide:` in the [Customizing the Work Area](https://opentoonz.readthedocs.io/en/latest/drawing_animation_levels.html?highlight=Ruler#customizing-the-work-area) section in the manual.

phillycheese007:
Thank you for pointing that out about the guides.

When trying to snap to the intersection of a vertical and horizontal guide:
It seems like the cursor is picking one or the other guide, when starting and finishing a shape, but never locking on the intersection.  i would also like to see the guides being used in a snapping operation display a bold colored line across the viewport showing it is active..and a vertical & horizontal bold colored line confirmation when you are activating the intersection of two guides for a snap.


fraz2




i am also wondering is there anyway to get the guides to snap to to the ruler when you are creating them? some way of making a snapping grid without having to freehand place all the guides?

phillycheese007:
what about a toggle for  "all, guides, or strokes" as a shortcut toggle or button in the toolbar?
strokes on top of guides the snap gets visually confusing. this could also just be accomplished by coloring anything that is being used by snap operation. both would be be even better

phillycheese007:
or maybe the simplest way to tell between a guide snap and a stroke snap would be an indication from the cursor/tooltip , right now if i am not mistaken it just tells you that you are snapping, with the green circle, not what sort of thing you are snapping to?

RodneyBaker:
> will snapping for the mesh skeleton vertices be included in this feature?

No.  This feature only deals with the Geometry Tool.

phillycheese007:
I would love to see the select tool work w/ snapping....like when dragging a whole shape or scaling a side....
I believe i have the latest version of this PR  and did not see this feature?

PR #1332, opened 2017-07-17
Russian translation (first draft)

morevnaproject:
Translated by Yury Grubnik (@beeheemooth).


shun-iwasawa:
I don't understand a single word in Russian, but it seems to be nice!
@morevnaproject and @beeheemooth, Большое спосибо

For now the current language is saved not as the language name but as index in the combo box.
I think I must aware Japanese & Chinese users that inserting Russian may cause unexpected change in UI language.

LGTM

morevnaproject:
Great! Much thanks to @beeheemooth for doing all hard work! ^__^

beeheemooth:
^_^, どうもありがとうございます

PR #1335, opened 2017-07-20
Fix Function Editor to be scrolled synchronously with Xsheet 

shun-iwasawa:

The function editor does not scroll synchronously with the Xsheet, especially it is obvious when using the mouse wheel.
- When wheel-down, it does not start scroll when the scroll bar is at the top.
- When wheel-up, it scrolls double.

wheel_bug


shun-iwasawa:
Well, it seems that the `clang-format` is needed in order to see the difference. 
Jenkins

manongjohn:
Forgot about that annoying issue while working on timeline. That only happens on empty scenes. Worked fine otherwise. Thanks for fixing.

shun-iwasawa:
Conflicts resolved.
Jenkins again

turtletooth:
Works great.  Thanks @shun-iwasawa

PR #1343, opened 2017-07-26
Sound Scrubbing in Combo Viewer and Missing Updates in Scene Viewer by turtletooth (Removed unused source)

shun-iwasawa:
This PR is based on #1260 by @turtletooth , just revised slightly as follows:

- Removed unnecessary code `isFrameAlreadyCached`.
- Moved the timing of resetting the console's frame rate from changing the scene to switching the scene.
- Simplified `ComboViewerPanel::hideEvent`

Again, thank you @turtletooth for the fix!

### DIFF
--- a/toonz/sources/toonz/comboviewerpane.cpp
+++ b/toonz/sources/toonz/comboviewerpane.cpp
@@ -714,6 +714,14 @@ void ComboViewerPanel::onSceneChanged() {
   int frameIndex    = fh->getFrameIndex();
   int maxFrameIndex = fh->getMaxFrameIndex();
   if (frameIndex > maxFrameIndex) maxFrameIndex = frameIndex;
+  // update fps only when the scene settings is changed
+  m_flipConsole->setFrameRate(TApp::instance()
+                                  ->getCurrentScene()
+                                  ->getScene()
+                                  ->getProperties()
+                                  ->getOutputProperties()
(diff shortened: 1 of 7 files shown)


shun-iwasawa:
Jenkins

toonz/sources/toonz/comboviewerpane.cpp
```diff
@@ -737,14 +714,6 @@ void ComboViewerPanel::onSceneChanged() {
   int frameIndex    = fh->getFrameIndex();
   int maxFrameIndex = fh->getMaxFrameIndex();
   if (frameIndex > maxFrameIndex) maxFrameIndex = frameIndex;
-
```
turtletooth: I posted this in the other PR, but I think having it on scene changed is actually useful and less confusing for users.  If a user changes the scene fps in the scene settings, but the viewer doesn't update, I think that would confuse users.

shun-iwasawa:
@turtletooth I've modified the flipconsole to update when the scene settings is changed, meanwhile, not to update with other operations emitting `sceneChanged()` in case the fps slider is modified manually.
Could you please take a look? Hope this works.

Jenkins again

turtletooth:
Tested and it works fine.  Thank you @shun-iwasawa for the extra fixes.

Issue #1316, opened 2017-07-04
Updated Column Header Layouts


2dvision:
If user want change column name, preview visibility will be off. I think only when user click on icon, it should be on/off.

![toggle column icon](https://user-images.githubusercontent.com/16509592/27837032-5d061cd0-60f7-11e7-8ae0-bc6da450115c.gif)

adamearle:
To add to this renaming and changing values in it is really clumsy.

It could definitely become more elegant. The excuse of this is a design
feature is not excuse for poor design.

It really is time for a change and for it to become more elegant in
workflow and seamless.

Workflow workflow workflow

manongjohn:
Should add some additional icons and rearrange it to look something like this so it's a little more intuitive, especially to new users. 

![image](https://user-images.githubusercontent.com/19245851/27929919-ca4b8c5a-6262-11e7-95ec-07548e6a065a.png)

Currently waiting for another PR to be reviewed and merged in which would be impacted by this. If/when it does, I can probably take care of this.

turtletooth:
@manongjohn I like that layout.  Can you think of a way to include the layer opacity in that?  What about moving the eye back to the top and putting opacity there somehow?

manongjohn:
The problem with the eye back on top is 2 fold: long names cover the eye and currently the eye control is still competing with the column renaming though I that can be fixed also, I suppose.

I was thinking about a layer opacity button also. Just not sure where to fit it nicely with the other icons.  Then I'd have to figure out where to place it in the timeline view as well.

2dvision:
Toonz Harlequin have this layout. Is it possible to have this?

![column redesign](https://user-images.githubusercontent.com/16509592/27933121-78a09bae-62b5-11e7-9e06-f6894d96be7f.jpg)

manongjohn:
Just a thought: I think I can borrow some space from column type below the image icon as well as reduce some of the vertical spacing and introduce a small slide bar to control opacity below the image:  (May not be this particular slide bar)

![image](https://user-images.githubusercontent.com/19245851/27932775-f9049c94-626c-11e7-9b58-d5a3b72e3ab9.png)

Someone added that filter color to the opacity slide bar which I'd also have to find room for.

2dvision:
OT have a hidden menu With more than one option (Opacity and filter).

[...]

Issue #490, opened 2016-06-22
﻿Change opacity levels


2dvision:
If you want inking, you need change opacity "sketch level" for inking in "ink level" be easy. 

now you can use FX, but this work only on preview mode viewer.

blurymind:
It is already possible to do that, but it is somewhat hidden.
https://www.reddit.com/r/OpenToonz/comments/4d4kgs/any_idea_on_how_to_change_a_levels_overall_opacity/

shun-iwasawa:
In the comment in reddit says:
> In the top of the column you have a yellow button, if you click it the col are made transparent and if you click and hold an transparency bar appears. 

To be precise, you need to click&hold the "orange" button (=camera stand visibility toggle) in order to make levels semi-transparent in the viewer. This feature is available in the current version as well.

![opacity_slider](https://cloud.githubusercontent.com/assets/17974955/16264848/d88dfb94-38b6-11e6-8527-ab83710d3a54.png)

2dvision:
OH GOD! :)))

Whay its hidden? OT is not a video game. Users should be able to easily see it. Perhaps with an icon, percent or etc. What do you think?

blurymind:
the problem is that buttons in OT tend to be sometimes invincible when in a disabled state.

This is similar to this: https://github.com/opentoonz/opentoonz/issues/411
Buttons for level visibility also disappear when disabled. They should have a disabled state instead of disappearing.

In this case it is a slider though

blurymind:
in defense of the current design - it does make for less clutter in the x-sheet. That makes it  extremely clean and sleek if you compare it to tvpaint.

AyameLee:
What about an icon? Does it look too busy?

![col_tz](https://cloud.githubusercontent.com/assets/18118611/16291310/615ed484-38fd-11e6-9b34-21c89695c19d.png)

blurymind:
@AyameLee @2dvision  technically the entire 'camera stand visibility toggle' button IS that 'opacity' button! The gotcha for the user is the fact that the button needs to be clicked and held in order for the opacity slider to appear.
So adding another button next to it is kind of redundant in this case and decreases the clicking area.

Right now we have a button that does two things - depending on whether its clicked or clicked and held.

What you are asking for is to split the button into two buttons that are simply clicked and each serves  one function.

We could hint to the user that the toggle visibility button is also a slider, by indicating percentage of opacity:
[...]

PR #1347, opened 2017-07-27
WIP - Xsheet Column Header Layout Redesign

manongjohn:
Issue #1316 was recently brought up and, as a way to fix it, I had suggested to change the layout of the Xsheet column headers separating the name from the preview toggle buttons as well as address what I though could improve in the current layout.

After some discussions regarding what the layout should be like, I incorporated a number of suggestions and ideas from various people and the last suggestion I proposed was this:

Xsheet Mode:
image

Timeline Mode:
image

These changes incorporate the following:
1) Shifted things around to utilize unused space to allow for more layout options (new button, bigger thumbnail)
2) Separated the Preview toggle button from the Column name, addressing the issue described in #1316
3) Removed the drag bar and made the Column name the drag point. Similar to how the drag bar lit up when over it, the Column name now lights up and brings up a hint that it is the drag point.
4) Option button for easier access to Opacity slider and Filter Color drop down, addressing the hidden options noted in issue #490
5) New icons in SVG format for on/off/disabled/transparent states.
6) Column Number  (not shown) added to right side of Column name. Visibility of the Column number is controlled by preference setting that already controls the column number in Timeline mode.

While we cannot expect everyone to like any particular layout, a few of us believe this new layout will provide a better UI experience for new/newer users while keeping in mind existing users concerns (buttons not too small, keeping color changing buttons)

**This PR is still open to layout discussions and suggestions and will remain in WIP for now.**  

Question: How/when do we know when we're ok with the layout and can merge it in?


turtletooth:
You can try it out in the latest experimental build: https://gum.co/TZmmM

openanim:
@manongjohn Thank you for the work you do, and good luck on this issue. there's no accounting for taste . It's a big issue :D. A small suggestion, it might be interesting to have a visual link between xsheet and stage shematic, when the preview mode is disabled, as in harlequin. Example :

https://1drv.ms/i/s!AspsYffPSwbdh4EO4n3EB0ffJDl5ew

@turtletooth  thx for the new version OTX 💃

Agnyy:
I suggest icons to make colorless. And move the marker color under the icons.

Before (Marker)
28655735-baf93388-726b-11e7-9171-fda8046f3f9c

After (Marker)

Agnyy:
I think this will become more compact and visually easier. What do you think? @openanim @manongjohn  @turtletooth

MaxMixBug:
Will they accept it? if yes, then when?

manongjohn:
@MaxMixBug We're still working through it, but I think we're getting closer to something more acceptable. 

I do have to update the description of this PR when I get a moment because it's slightly changed.  In the future you'll be able to pick from 1 of 2 layouts, the current one or a more compact one.  The layout proposed above will most likely not be offered.  Icons have also been updated.  If you haven't done so already, head over to the PR mentioned in the 1st post for the latest thoughts, suggestions and images of what may be.

Hopefully it will be accepted.  I think the direction we're currently heading would lead more to a "yes" than when I first started this :)

turtletooth:
@manongjohn Since this is still a work in progress, do you mind if I look at @konero 's new PR.  This will break this PR as well as some of mine, but it will ultimately make the styles better.

manongjohn:
@turtletooth Yes please do.  May take a while before this goes in so I expect I'll be getting conflicts every so often and will work them as they come up.

manongjohn:
Default and Light Theme proposed layouts from the latest commits:

image

konero:
I've been testing this out, really nice work by the way, all of the layouts ended up growing on me. I've got some minor feedback and discovered one or two bugs and a proposal to make.

## BUG 1
These relate to the qproperties, some of your timeline qproperties are mixed up.

**PREVIEW OFF** uses **TABLE OFF**
**TABLE ON** uses **LOCK ON**

## BUG 2
When the X-Sheet and the Function Editor windows are docked side-by-side their markers are out of alignment by 1px when the **Compact** layout is used. In this image X-Sheet is on the left.

a2

## CLASSIC LOCK BOX SUGGESTION
The black border around the lock box wants too much attention. Would it be too much hassle to remove it and instead use a floating box? I also found an issue since you decided to share icons across all layouts (good call) is with the border the lock box is 1px² smaller than the icon file causing it to become very aliased.

a4

## ICONS ICONS ICONS
What about using the icons from my pane PR? I made them the same as all other icons for consistency. If you'd like to try them out, see below for a QSS zip. I know they're not flat like intended, but I think they have a better **read**, at least no one can complain they look out of place.

a5

a6

## MISC
This may not be specific to your work, but I noticed the opacity slider is difficult to grab with a stylus. The QLineEdit box could be shorter. I only point this out since this journey all started because of that slider, haha.

a1

## PROPOSAL
Would you mind using my up-to-date stylesheet from #1474 so we don't conflict again? I can then close that one. The PR is just some SVG tweaks and adds a focus state to QPushButton. To make it easy I merged our [QSS ONLY] commits from both PRs and zipped so you could just download and replace.

This contains my pane icons, you'll need to replace those if they're not your cup of tea. 👍 

**Download: [qss_1347_plus_1474.zip](https://github.com/opentoonz/opentoonz/files/1326254/qss_1347_plus_1474.zip)**

turtletooth:
@konero - Don't close that one. I'm looking at your PRs right now.

konero:
@turtletooth OK, thanks! That would still allow @manongjohn to use the QSS zip I uploaded here, it would just show fewer changes on his commit.

manongjohn:
@konero  Thanks for reviewing and providing feedback!  I see your stuff has been merged and will work on doing a master sync to bring it all in.  Nice work! 

I'll work on the things you pointed out.  I like the updated icons. Aside from being more cohesive with the rest of the application, the borders on the images fixes the contrast issues we would run into every time we tried to play with the colors.

So...challenge for you:
I reverted the Selected Column color back to the original Red it was using since I was having a hard time finding a replacement color that was easily noticeable at a glance yet had enough contrast with the different colors currently in use.  Think you can find one? :)

konero:
@manongjohn Thanks! Be sure to use the icons in the zip folder above and not the ones from the actual pane PR. I had to modify them slightly for the x-sheet since they're different viewbox sizes. I also did the timeline/layer icons that also use different viewbox sizes to column head (you likely did/are, just checking).

I tried finding some new color for the selected text but it's pretty difficult, so many variables to take into consideration. I understand why it was suggested for change though. The best I could do was a bright blue but it won't work with classic view due to yellow bar.

Text:
`#51C9F8`
BG:
`rgba(0, 0, 0, 0.2)`
BG:hover:
`rgba(255, 255, 255, 0.25)`

Looks like a lot of hassle and complexity. I think red could work for now. It would be the easiest thing to change later if we come up with something else and that would hopefully speed up merge. 😹

manongjohn:
Latest commit includes fixes/updates noted by @konero 

- Bug 1 - Copy paste error. Code corrected.

- Bug 2 - Shorted Function Editor height by 1 px to match height of Xsheet

- Classic Lock Box now appears as follows:
image

For Classic mode only, if the stylesheet color for Lock Off is completely transparent, it will default to the transparent white as it does in current release.  Removed bounding box for Lock area so I can make the lock icon as big as possible to reduce the effect of aliasing.

- Icons Icons Icons - Updated
Additionally modified Xsheet Table View Transparency Icon to use look like the regular one but with the center of the image semi-transparent.  This is similar to how the current release shows it:
image

- Misc - Widened dialog box so slidebar wasn't being clipped by the %-age box

@konero I noticed you made the Timeline ON-state icons the same size as the OFF-state, the only difference being the color. I think it would be better if the ON-state was larger.  Plus, the Table View Transparency version looks too much like the OFF-state.  I had to switch it to the ON-state icon.  Either way, now we don't know if a transparency was applied.  At least with the larger size you could tell it was ON but the dimmed color indicated it was somewhat transparent.  Can you make the ON-state ones larger?  I'd probably just need 1 file (or 2 if the light version is different) and I can make the rest.

As for the selected column color, thanks for trying.  I agree, it's definitely not worth holding up this PR just for that.  I did correct it so it used the full red like it does currently.  What I had before was the red used by missing level which was hard to read with some of the current colors.

konero:
@manongjohn Whoops, I didn't realise what transp was for! 😹 

tline-ex

- I made on larger
- Transp icon similar to what other programs use

Three files, you'll need to replace both white and black folders. 👍 

[timeline_upd.zip](https://github.com/opentoonz/opentoonz/files/1332692/timeline_upd.zip)

konero:
@manongjohn Whoops, I didn't realise what transp was for. 😹 

Here you go: [xsh_upd.zip](https://github.com/opentoonz/opentoonz/files/1332738/xsh_upd.zip)

1. I made ON larger
2. Made transp icon look similar to other software

tline-ex

tline-ex2

For column view feel free to use either mine or yours, both are OK for me.

manongjohn:
@konero I like your new transparency icons and will use those.  Thanks.

Quick question: The Timeline lock ON icon is the lock.  I originally had the just the lit dot.  Wanted to confirm this change was intentional. :)  I'm fine with either and will use your recommendation.

gab3d:
I think it'd be more coherent if it'd be a dot too.

In spite of that, if it's only the lock icon that shows on each layer, it doesn't seem to look that heavy on the eyes either.

konero:
I agree it could use the dot too. 👍

manongjohn:
I think we're ready for review and merge.  Barring any changes requested by reviewer, here are the final looks:

image

turtletooth:
Great work @manongjohn - I'm going to leave this to @shun-iwasawa to review, since it is a pretty big change but it looks great!

toonz/sources/toonz/xshcolumnviewer.cpp
```diff
@@ -1823,8 +1958,27 @@ void ColumnArea::mouseMoveEvent(QMouseEvent *event) {
...
+      if (o->rect(PredefinedRect::SOUND_ICON).contains(mouseInCell))
+        m_tooltip = tr("Click to play the soundtrack back");
+      else if (o->rect(PredefinedRect::VOLUME_AREA).contains(mouseInCell))
+        m_tooltip = tr("Set the volume of the soundtrack");
+    } else
+      m_tooltip = tr("Click to select column, double-click to edit");
   } else if (o->rect(PredefinedRect::LOCK_AREA).contains(mouseInCell)) {
     m_tooltip = tr("Lock Toggle");
+  } else if (o->rect(PredefinedRect::CONFIG_AREA).contains(mouseInCell)) {
+    m_tooltip = tr("Additional level settings");
```
shun-iwasawa: How about changing this tool tip to "Additional **column** settings" for the purposes of accuracy?

toonz/sources/toonzlib/preferences.cpp
```diff
@@ -625,6 +631,17 @@ Preferences::Preferences()
...
   getValue(*m_settings, "shortcutCommandsWhileRenamingCellEnabled",
            m_shortcutCommandsWhileRenamingCellEnabled);
+
+  QString xsheetLayoutPreference;
+  xsheetLayoutPreference =
+      m_settings->value("xsheetLayoutPreference").toString();
+  if (xsheetLayoutPreference != "")
+    m_xsheetLayoutPreference = xsheetLayoutPreference;
+  else if (existingUser)  // Existing users with missing preference defaults to
+                          // Classic. New users will be Compact
```
shun-iwasawa: It's just a personal opinion - I think the "Classic-revised" layout would be good for new users.
I suppose that the "Compact" layout would be more for experienced user who already know the purpose of each button and would like to gain more working space in the cell area.
And IMO it would be more friendly to offer larger buttons available in the "Classic-revised" (or "Classic") layout for new users, considering that it would be more preferable for them to get familiar with usage of the buttons than to create long footage.

toonz/sources/toonz/xsheetdragtool.cpp
```diff
@@ -1761,12 +1761,18 @@ class VolumeDragTool final : public XsheetGUI::DragTool {
...
-    double v               = range.ratio(frameAxis);
-    if (o->flipVolume()) v = 1 - v;
+    const Orientation *o = getViewer()->orientation();
+    QRect track          = o->rect(PredefinedRect::VOLUME_TRACK);
+    NumberRange range    = o->frameSide(track);
+    int frameAxis        = o->frameAxis(event->pos());
+    if (o->isVerticalTimeline() &&
+        !o->flag(PredefinedFlag::VOLUME_AREA_VERTICAL)) {
+      range     = o->layerSide(track);
+      frameAxis = o->layerAxis(event->pos());
```
shun-iwasawa: For now, with "Classic-revised" or "Compact" layout, you cannot control the volume if you put the sound level in other than the first column.
Should this line be like as follows ? :
```
-      frameAxis = o->layerAxis(event->pos());
+      frameAxis = o->layerAxis(event->pos()) - getViewer()->columnToLayerAxis(m_index);
```

toonz/sources/toonz/xshcolumnviewer.cpp
manongjohn: I agree. Fixed.

toonz/sources/toonzlib/preferences.cpp
manongjohn: I agree with that as well. I've switch the default for new users to Classic-revised.

toonz/sources/toonz/xsheetdragtool.cpp
manongjohn: Good catch.  Fixed.

toonz/sources/toonzlib/preferences.cpp
gab3d: (also personal opinion here)
I just find the Compact layout more easy to new users because of being more:
- Simple: all buttons are on a single line.
- Coherent: no matter if you use xsheet or timeline modes, you always find this buttons in a similar layout (i.e. single line, same left to right order)
- Compatible: it's pretty much how Toonz has them also.
(NOTE: although I"ve just realized we have the CamStand and Render buttons order swapped with respect to how Tonnz shows them). Why that happened?? :-\   Could it be inverted before it goes "official"? ...it would be easier for users having to use both programs). @manongjohn @turtletooth @shun-iwasawa @konero

shun-iwasawa: @gab3d 
Thank you for the suggestion. I understand that the UI-related issue is really a matter of preference and the best answer would vary among users. But I still think that the "Classic-revised" is more for new-users than "Compact" ...
The preview and camstand toggle buttons are used IMO more frequently than the other two. I think that having them larger will afford users such importance. 
Considering the consistency with other software, the "eye" shaped button is commonly recognized as some visible property, but the "animation table" icon is uncommon and quite original in OpenToonz. Making the "preview" and "camstand" buttons the same (large) size will also help to know that both buttons have similar functionalities.

> (NOTE: although I"ve just realized we have the CamStand and Render buttons order swapped with respect to how Toonz shows them). Why that happened?? :-\ Could it be inverted before it goes "official"?

Thank you for pointing the inconsistency in order of icons with Harlequin, I didn't realize that. 
I originally made the current column head layout based on Toonz 4.6 which Ghibli staffs had been already familiar with. In Toonz 4.6 there was no concept of camstand visibility, so I just added it under the preview toggle.


I think having the preview toggle at the first place is somewhat reasonable even if it will lose some consistency with Harlequin, considering that the preview toggle is important as it actually affect to the render result.

shun-iwasawa:
Jenkins

shun-iwasawa:
@manongjohn Thank you very much for the great work! LGTM

manongjohn:
Thanks to everyone in this thread and in #1316 for their input, support and patience to help bring this all together!

artisteacher:
I love the new options! One thing I did notice, for the classic-revised column header in OS X, if a column is too close to edge of the program window, the opened opacity/filter area can get cut off. It was only a problem because my work computer has a fairly small screen. It should probably open to the left if there is not room on the right.

Issue #926, opened 2016-11-18
Feature: Hand-drawn Animation Auto-Complete


ghost:
In my graduation project me and my team are going to implement the Hand-drawn Animation Auto-Complete research: [click here to see video](https://www.youtube.com/watch?v=w0YmWiy6sA4).
This research is based on this one: [click here to see video](https://www.youtube.com/watch?v=m7MEAw46Ojo).

I think it would be easier if we start it from scratch so that the code base would be smaller and more manageable. But at the same time it would be kind of a waste. So we were thinking maybe we could add this feature to opentoonz. Opentoonz has a wonderful qt5 code base. But it's huge and we will spend so much time trying to be consistent with the rest of the code.

The question is: Would you guys be willing to merge this into the master branch if we do it, or would that be against any of the design philosophies of the program?
If you would be willing to merge it then we will definitely start to work on it.
If not then I guess we'll just start from scratch.

Also if you guys have any ideas, comments or critics we would love to hear them.

PS: we're big fans of opentoonz. Thank you for the masterpiece software.

turtletooth:
This looks great.  I think it would be awesome to integrate this into an existing product rather than creating a separate piece of software.  I am one of developers here, but I don't have the power to make the big decisions, but I don't see why something like this wouldn't be accepted and merged.  It would only make the program more functional and capable.  

I would love to see this implemented.  I don't think it competes with the design philosophies since OpenToonz is designed to work with many animation workflows.

Thanks for sharing the possibility and I'm excited to see what happens.

lordadamson:
That sounds encouraging, thank you :D.
So do you advice we start working on it right away or should we let one of the decision makers know first?

blurymind:
I'd say get it started :)
But it would be of course good to study the code base of some of the
relevant tools where this could be integrated. Opentoonz has for example a
basic tweening capability.

If you manage to get a functional prototype of this running in opentoonz
the likelyhood of merging it to trunk is high- as it is a valuable feature
indeed. @shun-iwasawa is one of the core devs and has merged other
contributed features in the past- sometimes with some refactoring after the
functional prototype

lordadamson:
Yes indeed. If we want to get it merged we'll have to have you guys see what we're doing and constantly get your feedback. Because I know we are going to get things wrong at the beginning, especially because we're not familiar with the code base. And this is the first time we code anything similar.

So what parts of the code would you recommend we start looking? @blurymind

turtletooth:
[...]

PR #1350, opened 2017-07-29
[WIP] hand-drawn animation autocomplete

lordadamson:
Hi,
This is a work in progress for a research that aims at providing an autocomplete functionality that should serve an artist as they draw.
I have been unable to dedicate time for this project for a while. That is why I'm opening this PR where I will document here everything I know about this research to make it easier for others to continue our work.
More details about the progress we made can be found at #926


lordadamson:
# Introduction
Our work is based on a paper by Jun Xing et. al. Called [Autocomplete Painting Repetitions](www.liyiwei.org/papers/workflow-siga14/siggraph.pdf) Which basically detects the patterns in the artists strokes and tries to predict the upcoming strokes accordingly.
Here's a [video](https://www.youtube.com/watch?v=m7MEAw46Ojo) by the original authors of a working prototype that shows the capabilities of the Autocomplete Painting Repetitions.

There is another paper by the same author called [Autocomplete Hand-drawn Animations](https://pdfs.semanticscholar.org/a7a9/ef882398bf7f2972519994ca564abfd28ca6.pdf) Which takes the pattern detection to be across frames.
Here's a [video](https://www.youtube.com/watch?v=w0YmWiy6sA4) for the Autocomplete Hand-drawn Animations.

## Note
All we've got is a couple of research papers and a few flashy videos showing only the best results. So until we produce testable outputs and real artists test this thing out and provide us with feedback, we may never know if this can actually practical.

# Method
## Sample Points
Every stroke is broken down into a few sample point that act as representatives to the whole stroke. The original author took a point every **3 pixels** in his implementation. Our implementation is vector based, so we chose the **bezier curve control points** as our sample points. And we don't know if that was a good decision, but hey, this is a research, we should try all methods and then stick to the one that produces the best results.
So allow me to insert a `TODO` here: try other methods to acquire sample points.

screenshot from 2017-07-29 13-32-53
 In OpenToonz a bezier curve can be represented in multiple ways.
* It could be a segment with only two control points
* It could be a quadratic curve with two points on the curve and a third that determines the curvature between them.
* It could be cubic with two points on the curve and two others determining the curvature or the curve.

Here is a picture from the internet that misspells the word bezier but shows what I'm talking about:
bezier_curves

Look at [tcurves.h](https://github.com/lordadamson/opentoonz/blob/AnimationAutoComplete/toonz/sources/include/tcurves.h) for more information.

We decided to go with the quadratic curves. So our data structure for holding a sample point now is called `TThickQuadratic` (which is the same as `TQuadratic` but it also carries the thickness of the line at this point).

In OpenToonz the three control points together make up a *chunk*; `p0`, `p1` and `p2`. Two on the curve one for the curvature, remember?
**Notice** that in the stroke, `p2` of one chunk is the same point as `p0` of the next chunk.
That is why our sample points are always `p0` of any chuck except the very last one, we take `p2`.

*To be continued*

blurymind:
This now looks pretty abandoned and doomed :D

Orphanlast:
Unfortunately it does. I was planning on eventually donating.

lordadamson:
I did my best documenting the whole thing people. If anyone wants to start working on it, this document should make it much easier.
If you guys are so interested in this research maybe you could post it on [bountysource](https://www.bountysource.com/) or a similar service, where someone might be interested.

DoomDesigner:
Why can not this be used now? What is missing?

Orphanlast:
So, wait... this bounty isn't completely dead yet?

lordadamson:
Please do not think of this as a feature as of yet. It is more of a research.
The difference simply put:
# Feature
When you implement something using technologies that are already established and are known to work.

# Research
When you attempt to create/enhance a new technology.

Research usually takes more time/effort. It involves trying different methods and experimenting around.

OpenToonz has been open sourced to support research in the field of CG. And this PR is definitely a research.

I really hope some other researcher pick this up and keep going.

711GM:
Is it possible to try something like that?

RodneyBaker:
Routine review of opened and closed PRs.

There is some very good information on Opentoonz vectors and working with them in this PR.

RodneyBaker:
Closing.  There is no anticipated work being done on this PR.

lordadamson:
Hi,
I'm the ghost that created this pull request then deleted my account later. I now realize that because I deleted my account the repo is not accessible.

So here's a link to the code: https://gitlab.com/lordadamson/opentoonz/-/tree/AnimationAutoComplete

in case anyone would like to look into it.

RodneyBaker:
Thank you @lordadamson 
Much appreciated.

Issue #1348, opened 2017-07-28
Proposal - Change Edit Tool Name


turtletooth:
The edit tool is actually more of an animate tool than an edit tool.  The selection tool is actually more of an edit tool.  
Would it be more accurate to call the edit tool the animate tool or something to that effect?

Just a thought.

2dvision:
I agree with "animate tools" name, But i think we need more merge Position, Rotation, Scale and Shear into one function (it is necessary). 

For example in krita:

![edit tools - merge function](https://user-images.githubusercontent.com/16509592/28736012-264312b6-73fd-11e7-8d67-2bb18e5bc487.gif)

gab3d:
I support a name change, but to me "Transform Tool", as its name, would be more accurately reflecting what the tool does. Because either if the transformations would be animated or not would be an election of the user, but they will always be using it to actually transform an item's position, rotation, scale, etc.

openanim:
I also thought of the name transform tool by seeing this issue as is the case in Toon boom harmony. But after reflection, in harmony the transform tool can animate only if the animate mode is activated. Otherwise, that can only transform. So, I think the name animate tool is better because that reflects the idea that you can animate with this tool.  The  selection tool in opentoonz looks more like the transform tool or select tool in harmony.

turtletooth:
@gab3d I understand what you are saying, but I disagree.  Everything you do with the edit tool sets a keyframe - which is for animation.  The selection tool is also a transform tool - but without the keyframing.

turtletooth:
This also makes for easier keyboard shortcut keys.  Right now eraser is "a" because "e" is taken for the edit tool.  

@2dvision Take a look at #1345 - the last icon in the row is a combo edit tool which does what you are hoping for.

openanim:
@turtletooth Eventually even the selection tool does almost the same thing, it just missing a function shear.

artisteacher:
I think the name change is a good idea. Edit as a name gives the wrong impression of the tool's purpose. If the name is changed, it will be important to have clear documentation. The harlequin manual is becoming less and less relevant as new features are merged.

gab3d:
@turtletooth you know what? you're right, I'm not. :)

both are transform tools, there are two main differences between them:
1. Select Tool allows the transformation of elements inside a cell, whereas Edit Tool transforms the whole cell.
2. Edit Tool generates keyframes, whereas Select Tool does not.

So Transform Tool is not a good name.

[...]

PR #1351, opened 2017-07-29
Change Name of Edit Tool to Animate Tool

turtletooth:
animate
#1348 

This is definitely up for debate, but I think it would be more clearly state what the tool is for.

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -2057,14 +2057,14 @@ void MainWindow::defineActions() {
   createRightClickMenuAction(MI_DeactivateUpperColumns,
                              tr("Hide Upper Columns"), "");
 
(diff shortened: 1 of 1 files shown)


BrundleThwaite:
I do not like the name "Animate" as It is even more general than "Edit" is and it does not animate the plastic or skeletal rigs. I would suggest that "Transform" would be more accurate.

openanim:
I think that's fine. @BrundleThwaite, the name transform tool has already been debated, and it emerges that it is not adapted to what the tool does, see #1348 .

shun-iwasawa:
LGTM

Although I personally have a kind of special feeling for the word "Animate" (as [commented](https://github.com/opentoonz/opentoonz/issues/1348#issuecomment-324596541) in #1348), I would like to emphasize the majority in the discussion done so far.
Thanks @turtletooth for raising the debate on it.

PR #1352, opened 2017-07-29
Change Matte to Alpha in the UI

turtletooth:
m2a

This changes matte (m) to alpha(a) in the interface.


gab3d:
on the same topic...

The HSV A RGB vertical letters next to sliders aren't currently translatable.
Could someone add this to translatable strings?
Thanks!!

turtletooth:
@gab3d I made the strings translatable in this PR.  Thanks for pointing that out.

gab3d:
@turtletooth Thank you for that! :)

shun-iwasawa:
@turtletooth Thanks for this modifications.

The "Matte"s seem to be still alive in the palette gizmo popup. Please check `toonz\pltgizmopopup.cpp` .

turtletooth:
Thanks @shun-iwasawa - I have never actually seen that window until you mentioned it. :)

adamearle:
Personal I think that this is a mistake changing from matte to alpha
because you can matt's for creating alpha channels many ways with current
industries best practices

Just take look at photoshop abs splitting channels and this is a very old
example of workING and creating matt's

Best to maybe do some homework on this.

Ad ;D

turtletooth:
@adamearle  I couldn't fully understand your post above, but I can't think of a commonly used program that uses matte to discuss opacity of a color.  Creating a matte is different than setting the alpha channel of a color.  This change is referring to the latter.  If fully admit I could be (and often am) wrong, so I would be interested in seeing what others have to say.

gab3d:
I vote for Alpha, it's more commonly used. (is also what it says in Toonz
Style Editor) ;)

RodneyBaker:
I appreciate the change in terminology to Alpha.
I must admit that for a moment I did have a rough time finding the setting
(in one of the views (can't recall which one) because subconsciously I
think was looking for the word "Matte"
I had gotten so use to looking for that word!

As for the use of the term "Matte"... now it is freed for better usage as
masking and matting should be used.
There have been a few requests related to such that allow for limiting
where brush strokes can be drawn etc. and if implemented those may use the
Alpha channel but in ways we currently cannot use them.

The standard of RGBA should be followed because that is such a core
standard it makes little sense to deviate from it unless and until such a
time where some revolutionary element of image structure and functionality
arrives that a new abbreviation should commonly be used (and not just by
OpenToonz but by everyone).  But most functionality associated with RGBA
images can be achieved by adjusting the parameters of the channels
externally (even by html and css for instance).  It is important to note
that of the four channels the Alpha channel is the one where most
development is now occurring with (primarily) depth data being stored in
that channel.  To access and utilize that data however a image format such
as EXR that can store that data must be utilized.  All of this does pertain
to resolution but to be the most useful the underlying infrastructure
should be resolution independent.  It is then the output that is given a
specified resolution.

I do shed a tear of sadness at the departure of the M/Matte label in
OpenToonz.
After many years I thought I had learned almost everything there was to
learn about Alpha Channels and through discovery of RGBM I learned that I
still have a long way to go.
In a way the designation of 'M' would make more sense as a channel where
Multiple images and Matting Transparencies can be stored and manipulated
but if its use confuses most user because it conflicts with well
established standards the usage would be further limited.

Of additional noteworthiness:  When considering vector images strictly
speaking there isn't an Alpha to the image outside of the Transparency of
the colors assigned to and filling those lines as long as that image
remains vector.  Where there is an absence of lines there is 'nothing'
until such a time as the image is 'rendered' to a specific image format
where that 'nothing' needs to be interpreted.  That format may treat empty
space differently than another format so the user may need to allow for
that difference.  We see this especially in older RGB image formats that
don't store data in an Alpha Channel (They can't because there is no Alpha
channel associated with that file).  The extension of this is applications
(such as browsers) that while displaying an RGBA image might not take into
account the Alpha channel's data.  Most modern browsers do effectively
display RGBA images.  Those that don't won't be very popular.

At any rate, it may be helpful to think of 'Alpha' not just as transparency
but in the sense of being of a dominant positioning... the first channel
outside the core three used in all images.
There certainly can be others but to introduce others unnecessarily would
be to increase problems (read:  variation) associated with those channels
exponentially.

artisteacher:
I also vote for Alpha.

RodneyBaker:
The long tail of continuous improvement marches on...

I notice there is a Matte Channel reference in the Keyboard Configuration menu (see image).
Similar to another issue, I will assume that label is not being used by macros and scripts but it may be best to verify that as a true assumption.


mattechannel shortcut

shun-iwasawa:
@RodneyBaker

>I notice there is a Matte Channel reference in the Keyboard Configuration menu (see image).

Thank you for finding this.
This command is for switching the "Alpha Channel" toggle in the flip console. I couldn't notice it as the tool tip of this toggle is already "Alpha Channel". The text "Matte Channel" appears only in the Configure Shortcut popup.

@turtletooth Sorry for the late mention, but could you please modify the line 1912 of `mainwindow.cpp` as follows? :
> `-  createRGBAAction(MI_MatteChannel, tr("Matte Channel"), "");`
> `+  createRGBAAction(MI_MatteChannel, tr("Alpha Channel"), "");`

It is true that the definitions of "Alpha" and "Matte" varies depending on the context.
I set channel names to RGB"M" in the style editor of Toonz Ghibli Edition in order to keep consistency with Toonz 4.6 - it used "M" for channel name. For now, in Toonz Harlequin it uses RGB"A" in the style editor. Even in the same software series, usage can be changed.
Both "Alpha" and "Matte" are related to image transparency, but in my opinion, "Alpha" is more used for the name of channel (which is embedded in the image data), and "Matte" is tend to used with some region of image to be transparent (which can be applied externally via composite process).
So, I think it is somewhat reasonable that altering RGBM to RGBA in the style editor and leaving the name of "Matte In/Out Fx" unchanged.

RodneyBaker:
>  leaving the name of "Matte In/Out Fx" unchanged

Oh yes, definitely keep the term 'matte' in the FX.

The Matte In/Out FX is not the same as the Alpha channel in the images where the FX is applied.  They might use the Alpha channel to drive the masking/matting but (as I understand it) they are not 'alpha-ing' anything.  ;)

Regarding something related I did want to look at the convention now used in OpenToonz with the advent of the term "Alpha" to check to see if it is consistent with standard usage.  I haven't investigated this yet but the thought is:  When the Alpha channel's setting in the Style Editor is set to 100% then the color style is fully opaque.  If I set a basic material's transparency to 100% in another program the surface/material is 100% transparent.  When the term was "Matte' this suggested that the a setting of 100% should rightly be 100% opaque BUT have we now inverted this through the use of the term 'Alpha'?  I believe we are okay here because another term for alpha would be 'opacity'.  So 100% opacity would equate to 0% transparent.  So a program that presents the option of Transparency is (rightly) inverting the settings while also inverting the terminology.  So again, unless I'm missing something I say we are okay here with the use of 'Alpha' as a replacement for 'Matte' with regard to color.  :)

turtletooth:
Thanks @RodneyBaker  for catching that.  Fixed.

shun-iwasawa:
Jenkins

shun-iwasawa:
Wow Travis succeeded !

shun-iwasawa:
LGTM, thanks @turtletooth

Issue #1346, opened 2017-07-27
Feature Request: Schematic FX node that does "nothing" (I/O or reroute passthrough node)


snapai:
Could we have a Schematic View (FX Schematic) node that does nothing but pass its single input to its output - for consolidating inputs to a MacroFX node, and other graph cleanup?

When a MacroFX node is created, OpenToonz does not give it input or output sockets for the MacroFX.
![OpenToonz open MacroFX](https://puu.sh/wU4t0/af55e798c5.png)
Instead, it creates a socket for each input entering the captured bit of graph.
![OpenToonz collapsed MacroFX](https://puu.sh/wU4u2/a913468c4a.png)

If you could create a node that simply passes FX render data from its single input to its output, that will allow us to collapse multiple sockets into a single one, where a macro is intended to have a single input plugged into multiple nodes.

Also this is useful just for general graph cleanup, keeping connections from branching until they're closer to what they're being connected to. For Blender's node graphs, these are called reroute nodes and are represented as a node socket floating free in the graph (highlighted):
![Blender Reroute node](https://puu.sh/wT8wG/b1b83d25cf.png)

The Blender screenshot also shows how reusable groups (macros) in Blender have input socket nodes. (they also have multiple-output socket nodes built into the group) I don't anticipate that happening in OpenToonz right away, as I imagine it would involve rewriting a major chunk of how nodes operate. Same goes for macros that have multiple output ports, as it doesn't appear anything in OT's node systems have multiple outputs.

One more example, here's a snippet from a friend of how input ports work in Harmony (this separates out overlay lines and composites them above another input):
![Harmony Multiport Input](https://puu.sh/wU4CK/e1a36dba8f.png)
While I don't use Harmony myself, I've implemented this particular graph in OpenToonz without difficulty using a different palette entry for the overlay lines; however putting it within a MacroFX, without some sort of I/O node to consolidate, would create four inputs (from the external-facing nodes inside the graph) rather than the two it actually requires.

turtletooth:
Done in #1353

PR #1353, opened 2017-07-29
Fx Pass Through

turtletooth:
This is for #1346 

This is an fx that does nothing except pass through what is sent to it.
Exciting huh?

pass

premacro

postmacro

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -288,6 +288,8 @@
   <item>"STD_noiseFx.Black_White"	"Black & White"	</item>
   <item>"STD_noiseFx.Animate"	"Random Animation"	</item>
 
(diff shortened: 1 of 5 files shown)


RodneyBaker:
Nice.  This looks handy.  

The introduction of this node got me thinking that all nodes should have a 'pass through' function and it turns out they in fact appear to have that functionality via turning on/off.  So, I learned several somethings here.  :)

shun-iwasawa:
Nice feature. It would better to have an original appearance in the schematic in the future.
Jenkins

2dvision:
Is it possible add multiple node to Fx PassThrough input? It is useful if user want add one FX for multiple node.

mutiple input node

shun-iwasawa:
LGTM Thank you @turtletooth for addition.

Issue #482, opened 2016-06-20
feature request: (view) command to toggle horizontal flip of canvas viewport


blurymind:
Very often when drawing, the artist's eyes get used to the image and they stop noticing mistakes in it. 
A common approach to look for problems in a drawing with fresh eyes is flipping the canvas horizontally. It is a 'View' command found in most drawing software like photoshop - including tvpaint and toonboom.

![13422425_1645300485795736_874756708222360569_o](https://cloud.githubusercontent.com/assets/6495061/16194636/6f1fe2c8-36ed-11e6-8839-742fd4296c24.jpg)

The command does not do anything to the actual image, just to the display of the viewport.
It can be done as a command under the view menu, a command that can be binded to a keyboard shortcut.

Rayek:
There are some workarounds, but YES a simple flip view command should be implemented.

Workarounds:
1) copy the current camera in the schematic view. You may want to rename it "Flip View"
2) select that camera, and switch to the EDIT tool.
3) switch to SCALE mode, and set the H value (horizontal) to -100%.

To check your work by flipping it you have several options now:

A) activate that second camera (either going into the schematic view, or left top dropdown menu in the EDIT tool properties), and click on the small camera icon at the top right of a view. By clicking on that icon, and switching back to table view (the first one of the three icons) you can now flip the screen easily.

Or:

B) keep the view in camera mode, and activate either the normal or flip camera.

Or:

C) open a second combo viewer, and activate the flip camera. Then in your main view work in table view, and in the second view turn on camera view. You now have two views, the original one, and a flipped version. Even when animating you can check your work against a flipped version. You could even create a dedicated room for this - side by side view :-)

To flip just one object, select the level, and use the level EDIT tool in scale mode, and again apply a negative 100% scale H value.

adamearle:
There is probably a flip and a flop effect. Make sure you check the nodes

Adam Earle

blurymind:
thank you for the workarounds guys. The feature request is for something quick  that is out of the box and can be bind to a keyboard shortcut.

We do this flip back and forth often, so making it easy to get to would be nice.

turtletooth:
Based on the discussion in #1207 feature requests are going to be closed here on GitHub unless there is a developer actively working on the feature. Feature requests can be discussed at the OpenToonz Google Group: https://groups.google.com/forum/#!forum/opentoonz_en

PR #1356, opened 2017-07-30
Added Commands to Flip Viewer Horizontally and/or Vertically

turtletooth:
#482 
This doesn't have default shortcuts set, but the options are in the shortcut manager.


openanim:
@turtletooth I made a similar request from the forum, I didn't know that's was already requested, thank you to do it and sorry for the duplication.

bobbybutterfly:
why not a button to flip objects horizantally.
In cut out style animation it would be really helpful to flip head or body around.
scaling each object negative by hand is tideous

turtletooth:
@bobbybutterfly That would be a separate feature request.  You can make those on the Google Group page.  Good idea though.  This feature is more for hand drawn work.

bobbybutterfly:
@turtletooth ok, thanks, i just put this on the Google Group

blurymind:
very useful!Thank you @turtletooth :)

RodneyBaker:
I confirm that this works great.
For the moment I've assigned Ctrl + Right Arrow to horizontal flip and Ctrl +Up Arrow to vertical flip.
Occasionally I'll forget what I assigned and use Ctrl +Left Arrow and Ctrl + Down Arrow but because the results are immediately seen on the screen it doesn't take me long to remember again.


Note to self:  Do not remove USB with OTX on it and then try to enter the shortcut preferences in the instance of OTX you inadvertently left open on you computer... that will likely result in a crash.  ;)

blackwarthog:
I think will be good to add some kind of label about current flip state. For example at the top of scene view near the zoom percent.

shun-iwasawa:
@turtletooth 
I found a problem as follows:
When one of the flip options is activated, zooming in/out the viewer with mouse wheel causes strange scale factor displayed on the title bar ( like `Zoom : -2147483648%` ). And pressing the zoom In/Out shortcut key causes crash.
Thank you for adding the useful feature.

turtletooth:
Thanks @shun-iwasawa I fixed those and a couple other glitches I found along the way.

shun-iwasawa:
@turtletooth The problem seems to be fixed in the camera stand view but still remains in the 3D View. It doesn't crash but the strange zoom factor appears.

turtletooth:
@shun-iwasawa The zoom factor isn't shown on 3d view for the viewer.  Would you be opposed to removing the zoom factor for the Combo Viewer in 3d view?  The zoom factor doesn't seem to actually reflect the zoom anyway.

shun-iwasawa:
@turtletooth 
> The zoom factor isn't shown on 3d view for the viewer.

In the Viewer the zoom factor appears on 3D view when there is no current level.
Anyway I agree with you to hide the zoom factor on 3D view, both for the Viewer and the ComboViewer.

Related to this, I found that scale marks of the Ruler (on the top and left side of the viewer, NOT the Ruler Tool) disappears on the "flipped" 3D view. It's not a critical as the Ruler will scarcely be used in the 3D view.

turtletooth:
I think I got it now.  Thanks for the help @shun-iwasawa

shun-iwasawa:
Jenkins again

shun-iwasawa:
LGTM Thank you @turtletooth for the nice feature !

artisteacher:
It could also be helpful to have these commands added to the viewer's context menu.

Issue #595, opened 2016-07-07
increase and decrease numeric values by click+dragging on them


blurymind:
This will greatly help workflow with a stylus. Right now we can only change numeric values by entering a new number. Numeric values in the tool options bar for example.

It would be much easier to increase and decrease them with a stylus by just click (mouse or stylus) dragging on them. It will also alleviate the problem of the sliders not getting enough horizontal space.

This feature exists in other animation software - such as Maya and Blender.
![u0gult](https://cloud.githubusercontent.com/assets/6495061/16662961/0ea68324-4472-11e6-88ae-0b3f0407e7da.gif)

its incredibly useful

blurymind:
It would actually be better if there is no need to middle click. Just click and drag on the title of the input box to increase or decrease it's value - as demonstrated in the gif image.

shun-iwasawa:
Is this issue related to #577 ?

blurymind:
It is similar - but it is different approach to the same problem.

He is requesting a new popup to appear. I am requesting no extra pop up - just click and drag. Single click and drag - not multiple clicks.
Dragging to the left decreases - dragging to the right increases.

turtletooth:
Based on the discussion in #1207 feature requests are going to be closed here on GitHub unless there is a developer actively working on the feature. Feature requests can be discussed at the OpenToonz Google Group: https://groups.google.com/forum/#!forum/opentoonz_en

turtletooth:
I'm working on this.

turtletooth:
The way OpenToonz is built, clicking on the label would be a very challenging.  I have it working now with middle mouse + drag on the value itself.  It could also be done with a keyboard modifier like ctrl.

PR #1362, opened 2017-08-01
Middle Mouse Drag to Edit Values

turtletooth:
This is for #595 

This does a number of changes:

- All value fields can be middle click and dragged to change their value.
- All value fields in the tool option bar that don't have a slider can have their label left-clicked on and dragged to change the value.
- The function editor spreadsheet values can be ctrl + clicked and dragged to change their value.  (This also works on empty cells in columns that already have at least one key.  This is handy for quickly setting keys.)
- Alt + click on a key in the function editor will clear they key.

I was torn on using middle mouse for everything except the function editor.  I wanted the commands to be the same everywhere.  The benefit of middle mouse is that it doesn't require two hands and a tablet pen button can often be set to middle mouse - making it useful for tablet users.  However, middle mouse can't be used on the function editor, since it will pan the spreadsheet and that is useful functionality.  

Because of this I opted for different commands for the function editor and all other value fields- trying to do what is most convenient for users.  ( I realize that remembering two commands isn't convenient, but regularly using two hands for a command isn't also.)

I personally am most excited about the function editor commands as they seem to make adjusting key values and setting/clearing keys a breeze.


blurymind:
Thank you! It should really help people with smaller screens. One
suggestion to add to it- mid click drag + shift could change the value at
smaller increments .. similar to how it is in blender

artisteacher:
I'm really excited to have this functionality. I do wish it was left-click and drag like Blender and Krita. I almost always just use  a touchpad and am not sure it is possible to middle click and drag on a touchpad.

blurymind:
We could use left click as well, it will work better with tablets too..
An alternative way to do it that would be neat for pc users is also mouse
scrolling :)

artisteacher:
If left-click and drag isn't possible, scrolling would be a great solution. The addition of shift for more precision is a good idea.

blurymind:
@turtletooth it doesnt have to use labels. Click dragging on input boxes actually makes a lot sense imo

turtletooth:
Question about the function editor- I currently have it so you can ctrl+click and drag on a value in the function editor spreadsheet to change the value.  (This can be used to also set keyframes on unkeyed cells as long as the column has a key.)  I also set alt+click to remove keys.  

So far this makes editing values much faster than double clicking and typing the value.

The question is "What should the keys be that go with the mouse move?"
Shift + Click
Ctrl + Click
Alt + Click

There needs to be some modifier because click on its own needs to be available for selecting cells, and middle mouse is used to pan the spreadsheet.

turtletooth:
I am also working on getting label click and drag working for only the edit tool and selection tool.  It's doable, but tedious.  But since I also use tablets alot, it would be a nice option to have.   Most other tool values have a slider. Plastic tool seems to be the only other one that doesn't - I'll look at that too.

All values will be able to do the middle mouse button as well.

openanim:
First of all, thank you for your work, I don't understand everything, I don't know if it is related or not but i wanted to know 2 things

- We will always be able to navigate between the different drawings , for example to navigate frame by frame, skipping 3 frames by 3 frames, etc. thanks to the mouse wheel or ctrl+mouse wheel ?

2) If it will be possible to change the values ​​of the brush and geometry tools with this ? Or if I am completely off topic ?

turtletooth:
Almost done- first post edited to include current functionality.

Todo:

- [ ] Adjust drag value if the measure in a field is inches or cm.
- [ ] Add better comments
- [ ] Squash commits

turtletooth:
Done - man that was a lot more involved than I thought it would be.

blurymind:
@turtletooth you are on fire! Thank you. It's perfect :D

blurymind:
i think ctrl would be a good default modifier, but ideally it should be configurable.
Mid click drag is perfect imo

RodneyBaker:
These are great time savers.
As I click and drag my middle mouse my my fingers and brain simultaneously say, "Smooooooth."  :)

turtletooth:
Don't forget to try clicking the labels of the edit tool with left click, or ctrl + left click and drag on the function viewer.  Lots of fun.

RodneyBaker:
@turtletooth said:
Don't forget to try clicking the labels of the edit tool with left click, or ctrl + left click and drag on the function viewer. Lots of fun.

I think I'm following all of that.
clicking
labels
labels of the edit tool
click with a left click (or ctrl click) and then drag
click and drag on function viewer
...
Okay, I think here is where I get lost.
The function viewer is the pink gnomon in the center of the canvas right?

Aside:  And... I never realized we could scale the Table itself.  Very interesting.

shun-iwasawa:
Jenkins

toonz/sources/toonzqt/doublefield.cpp
```diff
@@ -461,6 +494,58 @@ void MeasuredDoubleLineEdit::timerEvent(QTimerEvent *) {
...
+    m_mouseDragEditing = true;
+  } else
+    QLineEdit::mousePressEvent(e);
+}
+
+//-----------------------------------------------------------------------------
+
+void MeasuredDoubleLineEdit::mouseMoveEvent(QMouseEvent *e) {
+  if ((e->buttons() == Qt::MiddleButton) || m_labelClicked) {
+    m_value->modifyValue((e->x() - m_xMouse) / 2);
```
shun-iwasawa: In some Fx there are parameters with small value range. For instance, the `Opacity` parameter of the `Over Ino Fx` has 0.0-1.0 value range. In such case middle-dragging of the field cannot get the values other than 0 and 1 for now.

How about modifying this line, adding the `precision` argument like as follows ?:
```
 int precision = (m_maxValue > 100) ? 0 : ((m_maxValue > 10) ? 1 : 2);
 m_value->modifyValue((e->x() - m_xMouse) / 2, precision);
```
I once thought of referring `MeasuredDoubleLineEdit::m_decimals` variable but using the maximum value of the field seems to perform better.

toonz/sources/toonzqt/doublefield.cpp
```diff
@@ -461,6 +494,58 @@ void MeasuredDoubleLineEdit::timerEvent(QTimerEvent *) {
...
+    QLineEdit::mouseMoveEvent(e);
+}
+
+//-----------------------------------------------------------------------------
+
+void MeasuredDoubleLineEdit::mouseReleaseEvent(QMouseEvent *e) {
+  if ((e->buttons() == Qt::NoButton && m_mouseDragEditing) || m_labelClicked) {
+    m_xMouse   = -1;
+    m_modified = true;
+    onEditingFinished();
```
shun-iwasawa: How about inserting `clearFocus();` just below this line?

shun-iwasawa:
Related to this feature, I found a crash problem when dragging the frame field in the console and modified in #1574 .
Anyway this feature is very useful. Thanks @turtletooth !

turtletooth:
Thanks for the review @shun-iwasawa - I made the requested changes.

shun-iwasawa:
Jenkins

toonz/sources/include/toonzqt/lineedit.h
```diff
@@ -33,6 +33,7 @@ class DVAPI LineEdit : public QLineEdit {
 
   bool m_isReturnPressed;
   bool m_forbiddenSpecialChars;
+  bool m_mouseDragEditing;
```
shun-iwasawa: @turtletooth Sorry for the late notice, but it seems that `bool m_mouseDragEditing` should be initialized to `false` here or in ctor.

turtletooth:
Done.  Thanks @shun-iwasawa

shun-iwasawa:
Jenkins again

shun-iwasawa:
LGTM Thank you @turtletooth for addition !

artisteacher:
I love this! It could also be helpful for text size. Just a thought, but replacing the sliders in the tool option bar with the click + drag on the label could save a lot of room in the tool option bar.

Issue #1338, opened 2017-07-23
Timeline UI bugs


artisteacher:
I'm loving the new timeline variant of the Xsheet - thanks so much @manongjohn for getting this finished! I have noticed a couple small UI issues. Some of these might be OS X specific.

The text in the timeline/xsheet toggle button is slightly cut off on each end:
<img width="92" alt="screen shot 2017-07-20 at 12 54 13 pm" src="https://user-images.githubusercontent.com/21159116/28430045-fc903eaa-6d4c-11e7-91b9-d66b4173b4ef.png">
<img width="74" alt="screen shot 2017-07-20 at 12 54 25 pm" src="https://user-images.githubusercontent.com/21159116/28430046-fea09744-6d4c-11e7-838e-59ffaef77bef.png">

The drawing number can overlap the level name, or the ~ that represents a longer name. Keyframes can also overlap the level name.
<img width="51" alt="screen shot 2017-07-20 at 12 58 58 pm" src="https://user-images.githubusercontent.com/21159116/28433252-eb4bf268-6d58-11e7-8430-0180e63d4837.png">

manongjohn:
Not sure but the button issue is more likely OS X specific in that its a different font used than on Windows so it takes up more room. I sort of happened in Windows when I accidentally shrunk the button.

Unfortunately I don't have a Mac to test on but will see what I need to change to widen that button's text field without changing button width so it looks right on OS X. I think it just needs to be widened about 10px to be safe?

The name issue is in both platforms. Will fix.

I'll wait a little bit on submitting fixes to see if any more things need to be fix.

artisteacher:
10px would probably be fine but after some measuring I recommend 12px. Thanks!

2dvision:
What do you think if remove timeline/xsheet toggle button and add this to windows menu?

![timeline-ui-01](https://user-images.githubusercontent.com/16509592/28516276-e090623a-7075-11e7-9175-dd4a10219b40.jpg)

 I think we have more horizontal space in timeline.

![timeline-ui-02](https://user-images.githubusercontent.com/16509592/28516275-e07d7cd8-7075-11e7-982f-d1770e766529.jpg)

artisteacher:
@2dvision it may make more sense to rename the xsheet in the window menu to something like Xsheet/Timeline 

[...]

PR #1363, opened 2017-08-01
Theme Updates

konero:
I had some friends from Japan request of me some changes to the themes. The gist is they were missing some design elements from the original themes and the light theme needed some work.

**Requested Changelog**
- Console bar buttons have been condensed
- Removed grip bars for toolbar titles as this caused an experience issue grabbing them undocked
- Focus color removed on button press
- Lock/FSP icons now have a clearer presentation of their states
- Tool buttons checked state for 048 variant has been reverted back prior to 1.1.3

**General Changelog**
- Light theme redone (I feel this one needs to be reviewed by @shun-iwasawa)
- Scrollbar handles are now rounded and with vastly improved contrast and edge hinting
- Border color on the textfield control for 048 variant has been darkened to match After Effects
- The dialog button frame background color makes a return
- Flipmarker slider has been tweaked so its easier to see when in preview mode
- MUCH improved styling for File Browser
- Updated lock icon in toonzqt/resources to match the one in the themes
- Updated the SVG files for the Console Playbar buttons for better contrast across all themes

console-buttons-new

console-buttons-new2

The icons now use a pure black outline set at 50% alpha, this way the outline will be somewhat responsive when the theme color is changed instead of using an opaque gray that causes the icons to blur on the VeryDark theme.


artisteacher:
I like the grip bars but they are less helpful when undocked, since the close button takes up a good deal of the space. The whole docking/undocking experience is pretty fussy in general, at least in OS X.

konero:
@artisteacher Me too, not worth the hassle really since it was a very fiddly smoke and mirrors job.

I've done a lot of tweaking while I've had stuff open, you can download the themes and try them out but remember to backup your current themes.

Friends in Japan were using the light theme so I really had to redo it as it wasn't polished enough. Although @shun-iwasawa might want to try it out first as it seems fairly important. It's the theme color I see in the Studio Ghibli artbooks and in press releases so let me know if I changed anything you put there specifically you need back.

Changes listed in OP.

**Download: [ThemeUpdate.zip](https://github.com/opentoonz/opentoonz/files/1194664/ThemeUpdate.zip)**

gab3d:
I like "Neutral" theme over the others too. :)


2017-08-02 14:08 GMT-03:00 Kite <notifications@github.com>:

manongjohn:
@konero  @artisteacher reported in Issue #1338 that the new xsheet/timeline toggle button text was getting clipped at the edges. This apparently only happens on macs. I believe it's because different fonts are being used between macs and windows.

The easiest way to fix it is to put this in the mac-only stylesheets:

#flipOrientationButton
{
  padding: 2px 12px;
}

For the windows version it needs to be "2px 15px" or not styled so the button doesn't shrink.

Since you are updating themes, can you include this change? This will cause the base and mac stylesheets to differ. Is there a better solution to stop the text clipping such that it works on both mac and windows machines?

konero:
@manongjohn I remember having buckets of issues with Mac fonts. 👿 

I've tested your change on Windows, I don't have a Mac, and it looks OK. Are you seeing more shrinkage than me? I'm pretty OCD but I'm OK with how the button looks.

xsheetbutton

I'd rather not walk down the route of applying Mac only fixes, too much overhead. I'll apply the fix for now. 👍

manongjohn:
Yes, my OCD is kicking in now too :)  On the right edge of the button in Xsheet mode, there is a space in Windows version when you shrink it down to 12px.  It is normally the same size as the Frame button bellow it when not styled. There is shrinkage in the timeline mode also, but not as noticeable since it's all on 1 line.

I didn't want to have Mac only fixes either so wasn't sure if there was a better method.  My only other solution is to widen the box by some amount to avoid the clipping issue but didn't really want to do that.  I wouldn't know what that value would be because I don't have a Mac either.

Oh..one other question/request.  In the Dark theme, the xsheet toolbar color appears to match exactly to the color of the Frames area of the xsheet in timeline mode so they blend into each other.  In other themes, they are different colors so there is a definitive break between xsheet toolbar and Frames area.  

Can you modify the xsheet toolbar color a smidge or modify the Frames area color a little to make that separation?

konero:
toolbar-issue

I tried adding a border to the bottom to match other toolbars, but it seems the xsheet toolbar doesn't have expanding width.

I could add a border back to all xsheet components, but that looks very broken which is why I removed it in the first place (it displaces the scrollbar). Unfortunetely there are not enough objectNames to reference for styling.

Do you think you could separate each component into its own object name? That would then allow targeted styling. This is probably difficult to do, sorry for suggesting. 😬 

xsheet-ids

I'll tweak the the color for now.

manongjohn:
The colors used in the column head, xsheet frame and xsheet spreadsheet are actually qproperty settings under "XsheetViewer" in the stylesheet.  Most of the names are descriptive enough so you know what color you are actually editing.  I think the XSheet Frame one uses qproperty-BGColor.  

The object name for the Xsheet Toolbar is "XSheetToolbar", I see the buttons are being styled, but I think the actual bar itself might be inheriting it from somewhere, I think.

artisteacher:
I took a look and the update is looking good. The console works well more compact, though it would be nice to see a little more space/division between the different areas.

manongjohn:
@konero Were all these changes abandoned or were they incorporated in another PR?  Need to know if I should take the change for timeline button into my Xsheet PR or not.

konero:
@manongjohn You can add the change. I'll submit all the fixes and changes in this PR again. 👍

PR #1366, opened 2017-08-02
Enable to use template settings for preferences

shun-iwasawa:
This feature is demanded by some Japanese animation studio.
In the studio, animators use OpenToonz only for pencil testing.
There are some preferable options for pencil testing in preferences, so they would like to introduce the template settings in order to cut out the need of initial customization for new users.

With this PR, if there is no `preferences.ini` in  `$TOONZPROFILES\layouts\settings.username` , then try to load the template settings `$TOONZPROFILES\layouts\settings\preferences.ini` .
Nothing is changed in behavior if there is no template settings.


shun-iwasawa:
Jenkins

turtletooth:
Seems to work fine.  LGTM.  Thanks @shun-iwasawa for your continuous work.

Issue #791, opened 2016-09-05
Modern color whele


2dvision:
**UPDATE**

[Coolorus](https://coolorus.com/) is a color wheel plugin for Adobe Photoshop:

![cooloruse color wheel](https://cloud.githubusercontent.com/assets/16509592/24871113/1b67459c-1e08-11e7-99b8-4a866e16f49d.png)

TVPaint Color whele:

![tvpaint_colorwheel](https://cloud.githubusercontent.com/assets/16509592/24871217/6b04d4de-1e08-11e7-9753-b7ba2f5e42c8.png)

Toon Boom Color picker:

![toonboom_colour_picker](https://cloud.githubusercontent.com/assets/16509592/24871419/0a32e6c2-1e09-11e7-8f2e-6dd725db183a.png)

In krita you can Choose Color wheel like this:

![screenshot from 2016-09-05](https://cloud.githubusercontent.com/assets/16509592/18250979/ba0033a2-739b-11e6-9370-67863f5e23a0.png)

MyPaint:
![mypaint-color](https://cloud.githubusercontent.com/assets/16509592/25784831/5b467e42-3363-11e7-8b9f-c93d7a7b441d.gif)

turtletooth:
![color_wheel](https://cloud.githubusercontent.com/assets/4576381/22915768/ced3fa7e-f236-11e6-9316-129feac861b9.JPG)

I found this hanging out in the code.  You can change the modes between R, G, B, H, S and V.  I think it would be neat to have the option of which one to use.

turtletooth:
@artisteacher What do you think?

artisteacher:
I think having options is always a good thing. I'm all for it!

ideasman42:
[Options don't come for free](http://enterprise-dashboard.com/img/cockpit-dashboard.jpg), more buttons for users, more code for devs to maintain - am not really against, just there *is* some cost in adding options and spending some of the [complexity budget](http://scattered-thoughts.net/blog/2015/10/25/complexity-budgets/).

artisteacher:
Good point. I definitely wouldn't add all of the options displayed in the example - the style editor is certainly complex enough. I do still think adding the regressed mode back in could be helpful.

2dvision:
In [Toonz](http://www.toonz.jp/) we have a nice color selector, Can we have it in OpenToonz?

![](http://www.toonz.jp/images/immagini%20interfaccia%20Toonz/o1.png)

[...]

PR #1370, opened 2017-08-03
(WIP) Advanced color selector

caryoscelus:
I'm currently developing an advanced color selector (based on https://github.com/mbasaglia/Qt-Color-Widgets and so usable for any Qt apps, not just OpenToonz) as requested by @Agnyy. While it's far from being complete or advanced yet, i've already made it to work with OT and so have a few questions regarding integration:

- what is the preferred way of source storage and building? As i understand, OT doesn't use many third-party dependences and uses static linking, so for now i used git submodule and add_subdirectory in CMakeLists

- are there any special cases related to color selecting? I already run into "color 0" which is displayed as invalid style in Style Editor (though changing it through my dock works)

- i see you're using custom icons for everything so i suppose using icons bundled with Qt-Color-Widgets would be ok?

Since my color selector widget isn't limited to OT, i'd like this thread to be kept for OT-related stuff and discuss universal features in https://github.com/mbasaglia/Qt-Color-Widgets/issues/29

screenshot

Issues List:
- [ ] Fix the color picker not picking both colors for gradients and stuff. Only lets you pick one color.
- [x] Fix black text in left most box when the color picker is opened for the 1st time
- [x] Fix Linux compile target in `toonz\sources\toonz\CMakeLists.txt` (see comment belot)
- [x] Suggestion: Relocate `MI_OpenAdvancedColorSelector` option in Windows Menu to after `MI_OpenColorModel`
- [ ] Add `MI_OpenAdvancedColorSelector` to ` stuff\profiles\layouts\rooms\Default\menubar_template.xml` to match location in Windows Menu
- [ ] Fix styling issues (not required, but would be nice)


MaxMixBug:
@turtletooth Please add this to OTX

2dvision:
Finally #791 , Thanks.

tGuMedia:
this is really cool, not a fan of the UI yet though but that'll probably change in the future

caryoscelus:
Yeah, the UI is temporary, to avoid dealing with overlapping widgets
for now.

Agnyy:
Test video (New) 
image

https://youtu.be/DSSkQtJxhv4

2dvision:
That's perfect. Is it possible add HSV and Alpha slider too?

adamearle:
this is just to throw a spanner in the works for people to have a think about. http://paletton.com/#uid=c0d310G3v0klllltksSs0sAeFe5f4dm

toonz/sources/toonz/advanced_color_selector.cpp
```diff
@@ -0,0 +1,90 @@
...
+      wheel,
+      &AdvancedColorSelector::colorChanged,
+      [palette_handle](QColor c){
+        auto palette = palette_handle->getPalette();
+        auto styleIndex = palette_handle->getStyleIndex();
+        if (!palette || styleIndex < 0)
+          return;
+        if (palette->getStyle(styleIndex)->getMainColor() == qColorToTPixel(c))
+            return;
+        palette->setStyle(styleIndex, qColorToTPixel(c));
```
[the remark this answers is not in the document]
caryoscelus: Fixed, thx.

caryoscelus:
>Is it possible add HSV and Alpha slider too?

HSV should be pretty trivial. As for alpha, i have plans for it, but it may need a bit of tweaking (e.g. to make it optional).

artisteacher:
Will the HSV and RGB sliders be optional? Does the widget fit well in a small but horizontal space? Could there be a way to create new styles in the level palette directly from the widget?

caryoscelus:
@artisteacher:

>Will the HSV and RGB sliders be optional?

RGB sliders are already optional, HSV will be also when it is implemented.

>Does the widget fit well in a small but horizontal space?

It doesn't fit really well into horizontally oriented space, except for "Rectangle" tab which can simply takes all space available. It should be possible to implement orientation change eventually.

>Could there be a way to create new styles in the level palette directly from the widget?

Currently not possible, but should be doable in theory. One of ideas was ability to integrate any kind of palettes into the widget. Another option would be to add custom context menu on colors and add "add as new style" action there. Yet another would be to implement drag&drop

caryoscelus:
Added HSV.

artisteacher:
The functionality of the advanced color selector works great. It will pair particularly well with a workflow using the my paint brushes. 

There are some areas where the UI can be refined - both in terms of blending in with the existing themes and in terms of fluidity. 
- The white border is a bit distracting since it is high contrast, not used anywhere else & has positioning issues when changing the tabs
- The tabs should use the same styling as the style editor & palettes (i.e. no rounded corners and a darker value for the inactive tab)
- The configure button should use an icon instead of text
- If the selector is too narrow, you can see the tabs through the middle of the arrows which is an existing behavior with the other tabs but is more distracting here due to the configure button
- If the selector is too narrow, the color ring placement feels out of balance (skewed a little too left)
- If the selector is too narrow, the mode buttons can overlap the color ring
- There should probably be a minimum height for the wheel area
- While the black background is visually pleasing, the margin takes up unnecessary space for those of us with small screens

In addition to the existing modes, it could be helpful to have a mode that selects a set of values in a single color range.

Drag and drop for the palette colors would be fantastic, especially if you could select multiple styles at once. Even just a shortcut would work well (i.e. shift-left-click creates a new style while left-click modifies the selected style).
Thanks @caryoscelus for this great addition!

morevnaproject:
We have found that with this addition it is impossible to change alpha. Here is the fix - https://github.com/morevnaproject/opentoonz/commit/7fe91566b67045a36e2f5170787ff61b49ad4e1b

turtletooth:
I tried building this on Visual Studio and there are a number of errors.

caryoscelus:
Short update: alpha works with a slider, everything that i could test seems to work. I can't help with building on system i don't use without details.

Hopefully i'll answer to suggestions in more details later.

711GM:
Add this PR to the new version of the OT. I ask you to!

711GM:
@shun-iwasawa When do I add this feature?

jpturcotte:
@711GM Please be patient, and try not to double post. Thank you!

openanim:
@711GM If you want to use this selector, you can also download the morevna version from their website.

turtletooth:
@caryoscelus It looks like this is failing on Windows due to a missing CMakeLists.txt file in the Qt-Color-Widgets directory

caryoscelus:
On Sat, 18 Aug 2018 22:07:40 -0700
Jeremy Bullock <notifications@github.com> wrote:

> @caryoscelus It looks like this is failing on Windows due to a
> missing CMakeLists.txt file in the Qt-Color-Widgets directory
> 

Have you initialized git submodules? If your Qt-Color-Widgets directory
is empty, you should try that.

RodneyBaker:
I'm curious how we can get this to pass the AppVeyor tests so that interim releases can be tested.

Agnyy:
What needs to be done to make this PR accept?

shun-iwasawa:
@caryoscelus @Agnyy 
Sorry for my late response. 

Here are my thoughts:
- Please try adding `git submodule update --init --recursive` to the `install` script in opentoonz/appveyor.yml in order to advance the Appveyor build.
- I managed to build the PR on Windows but it was not yet straightforward:
  - Needed to call `git submodule update' which should be added to the document.
  - CMake adds 5 projects ( `ColorWidgetsPlugin` , `ColorWidgetsPlugin_install` , `ColorWidgets-qt5` , `gallery` , and `screenshot_bin` ) only for this widget. I think it should be minimized for simplification.
  - As @turtletooth pointed, there were a number of errors appeared. To resolve errors I did the following;
    - Removed `screenshot_bin` project which seems to be unnecessary.
    - Added `QTCOLORWIDGETS_LIBRARY` to the preprocessor definition of `ColorWidgets-qt5` project.
- Can you please cope with the above? Or I personally think it would be somewhat reasonable to simply make your library pre-built and put in `thirdparty` folder instead of using submodule.
- The license notice seems to be needed to the stuff folder `opentoonz/stuff/doc/LICENSE/` as the original source is under LGPL.

The widget looks neat and useful. Thank you for your contribution!

caryoscelus:
@shun-iwasawa 

>Or I personally think it would be somewhat reasonable to simply make your library pre-built and put in thirdparty folder instead of using submodule.

I guess it can also be just embedded, but then it'll be harder to update. The middle option would be to use git subtree, so that it still retains ties to original history, but also checkouts automatically. That would also cover the appveyor issue.

>Added QTCOLORWIDGETS_LIBRARY to the preprocessor definition of ColorWidgets-qt5 project.

Hmm, that sounds unfamiliar. I guess i'll have to review build files..

Will update PR soon.

caryoscelus:
Sorry for the delay. I've switched library to git-subtree, removed some unneeded stuff and added license notice. Still not sure what to do with QTCOLORWIDGETS_LIBRARY though.

caryoscelus:
Hmm, judging by the logs, travis failure on mac os seems to happen before it even starts building color widgets. Appveyor's error is still a mystery to me.

> [00:05:10] C:\projects\opentoonz\toonz\x64\Qt-Color-Widgets\include\moc_color_wheel.cpp(248): error C2491: 'color_widgets::ColorWheel::staticMetaObject': definition of dllimport static data member not allowed [C:\projects\opentoonz\toonz\x64\Qt-Color-Widgets\ColorWidgets-qt5.vcxproj]

Unless moc works differently under windows, this is what is on 248 line:

```c++
QT_INIT_METAOBJECT const QMetaObject color_widgets::ColorWheel::staticMetaObject = {
    { &QWidget::staticMetaObject, qt_meta_stringdata_color_widgets__ColorWheel.data,
      qt_meta_data_color_widgets__ColorWheel,  qt_static_metacall, nullptr, nullptr}
};
```

RodneyBaker:
@caryoscelus Is there anything we can do to assist?  If so, please let us know.

caryoscelus:
@RodneyBaker @shun-iwasawa 

I think the only obstacle left is CI failure (and possibly windows build in general). I also tried updating Qt-Color-Widgets to upstream, hoping that it might fix that (build system is somewhat changed there),
but unfortunately it got sliders broken in OT (some styling issue). I'm pushing a few more minor improvements (as well as update to master), but this is unlikely to fix CI.

If anyone can help fix windows/ci build, please do (i think github has a feature enabling collaborators to push here directly; or just publish commit separately and i'll update the pr). I cannot test it other than by pushing to CI and error messages there weren't exactly helpful

RodneyBaker:
Thanks for the update @caryoscelus 

Huge guessing and speculation on my part follows:

I'm curious if the CI failures are due to expired or missing dependencies.
In the case of windows this could very well be related to the content of the stuff directories.
It was noted that something similar was happening but in reverse where Mac builds were running better when the windows foundation was portable... not being dependent on entries in the Windows registry.

I'm way out of my comfort zone on this one.

cc @manongjohn I know you are busy with last minute changes but if you could weigh in on the windows side of this it will be appreciated and help us determine if inclusion the Advanced Color Selector in v1.3 needs to be delayed.

toonz/sources/Qt-Color-Widgets/CMakeLists.txt
```diff
@@ -0,0 +1,114 @@
...
+include/color_names.hpp
+include/component_color_selector.hpp
+include/advanced_color_selector.hpp
+)
+
+qt5_wrap_cpp(SOURCES ${HEADERS})
+qt5_wrap_ui(SOURCES src/color_dialog.ui src/color_palette_widget.ui)
+qt5_add_resources(SOURCES src/color_widgets.qrc)
+
+# Library
```
shun-iwasawa: How about inserting 
`add_definitions(-DQTCOLORWIDGETS_LIBRARY)`
just like ColorWidgetsPlugin project?

toonz/sources/Qt-Color-Widgets/color_widgets_designer_plugin/CMakeLists.txt
```diff
@@ -0,0 +1,102 @@
...
+# Qt
+find_package(Qt5Designer REQUIRED)
+set(CMAKE_AUTOMOC OFF)
+set(CMAKE_AUTOUIC OFF)
+set(CMAKE_AUTORCC OFF)
+set(CMAKE_INCLUDE_CURRENT_DIR ON)
+include_directories(${Qt5Designer_INCLUDE_DIRS})
+
+# Debug
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Wall -pedantic -Werror")
```
shun-iwasawa: Same as the above-mentioned, 
```
+ if(CMAKE_COMPILER_IS_GNUCXX)
  set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Wall -pedantic -Werror")
+ endif()
```

toonz/sources/Qt-Color-Widgets/CMakeLists.txt
caryoscelus: Ah, thanks, so that's where it was supposed to go. Now it seems to build fine.

caryoscelus:
@shun-iwasawa @RodneyBaker 

So now with ci passing, i believe these are all changes strictly required for merge?

RodneyBaker:
All  appveyor checks have passed... let's give this feature a try and see what is there to be seen.
I'll report right back! :)

shun-iwasawa:
@caryoscelus 
Although passing CI is a big step, I think we need to check out the source and test the feature itself before introducing this PR. Thank you for your patience.

@RodneyBaker 
I'll try running Jenkins in order to make the installer for testing this PR.

RodneyBaker:
Very nice.   I'm liking this.  :)

shun-iwasawa:
@caryoscelus
I'm so sorry but I decided to postpone including this PR before releasing V1.3. It is not only because the size of this PR is too large for me to review within the RC period, but also because this PR will introduce sources under LGPL which needs careful investigation in order to avoid unexpected license conflicts. 
As @RodneyBaker remarked, I also believe this feature will make the software more useful. Thank you for your understanding and patience!

Agnyy:
@shun-iwasawa Any news? I would like to accept this request.

manongjohn:
@shun-iwasawa  Can you or Dwango please review the license and let us know if this is a GO or NO GO situation?  This is the major stumbling block as far as I am concerned.

If so, there is still the matter of a minor conflict that needs to be resolved.

toonz/sources/Qt-Color-Widgets/CMakeLists.txt
```diff
@@ -0,0 +1,113 @@
...
+add_library(${COLOR_WIDGETS_LIBRARY} ${SOURCES})
```
shun-iwasawa: According to the [CMake doc](https://cmake.org/cmake/help/latest/command/add_library.html#id2), it seems to need `SHARED` type like
`add_library(${COLOR_WIDGETS_LIBRARY} SHARED ${SOURCES})`
in order to create a dynamic library.

shun-iwasawa:
I'm sorry for the late response. I've been thinking about this PR.

First of all, as long as I have tested in Windows environment, currently this PR seems to output a static library. It is not possible for OT to static link with LGPL library.
(By doing so, we need to release the source code or object code of conbined work on conveying the software based on the term 4-d-0 of LGPL v3 and it will add an additional restriction to the existing 
new BSD license.)

This library has a license exception, but it allows only when the conbined work is `under terms of any of the GNU General Public licenses` . It does not allow to distribute this library under new BSD lincense.

Therefore, as a precondition, this PR must be modified to dinamically link with OT. In particular, it seems that you need to modify CMakeLists.txt as pointed above.

Furthermore, I'm still concerning about the following:

1. How can we let the developpers to realize the fact that only when they modify the sources under `Qt-Color-Widgets` , their contributions are subject to different license. (IOW they are actually modifying another software from OpenToonz.)

1. How can we prevent potential risk that an innocent developper may accidentally reuse (copy & paste) a part of the source code from this library to another part of OpenToonz and causes license violation.

I think my concern is because the sources of this library are currently placed under `toonz/sources`. IMO it should be more clear that this module is a separate work by different author from OpenToonz.

Therefore, I would suggest moving all `Qt-Color-Widgets` sources from `toonz/sources` to `thirdparty`.
By doing this, I think we can maintain consistency with the conventional license and introduce this feature with smaller risk and confusion.

I'm not 100% sure if the above measure would be adequate. I would like to hear from you all.

caryoscelus:
@shun-iwasawa 

That seems reasonable. Unfortunately i'm not sure when will i be able to update and test this PR

> It is not possible for OT to static link with LGPL library.
> (By doing so, we need to release the source code or object code of conbined work on conveying the software based on the term 4-d-0 of LGPL v3 and it will add an additional restriction to the existing new BSD license.)

(Of course, i'm not a lawyer or anything, but) it wouldn't be additional restriction on source code.. rather, on a combined work which only arises from static linking (compiled binaries). For source code it would be an additional option to use it on terms of LGPL (but still retaining original BSD license), which is already present because BSD allows being part of LGPL project (as long as original notice is kept), so it would amount to nothing

Agnyy:
@shun-iwasawa Hello! This request can be resolved in the near future?

shun-iwasawa:
@caryoscelus 
As far as I understand there is a obvious difference between static and dynamic linking, when linking LGPL library to the software under another license.
Please read the terms 4-d-0 and 4-d-1 in [LGPL v3](https://www.gnu.org/licenses/lgpl-3.0.en.html) .
Using dynamic linking means taking the option 4-d-1. With the static linking you cannot choose the option 4-d-1. Therefore you have to take the option 4-d-0, which is to `Convey the Minimal Corresponding Source` of the combined work (i.e. OpenToonz).
BSD 3-Clause License does not stipulate licensee to convey the source code on redistributing.
So, only by static linking with LGPL library licensee will need to disclose the source code of the software. It's what I called the "additional restriction".

@Agnyy 
I myself am not working on this, but I'll review it once the PR is modified according to the above comment.

caryoscelus:
@shun-iwasawa yes, there's additional "restriction", but i was just saying it only applies to binary distribution rather than source code itself

@Agnyy et all: i'll try to find some time this or next week. Although if someone else more involved with OT would want to try and fix linking method, then by all means, please do so: i'm afraid i'd need additional time on setting build environment back up..

caryoscelus:
Shared library seems to work fine for me; still have to do source moving, though (add_subdirectory obviously doesn't work as well with non-subdirectories..)

caryoscelus:
I'm hoping i didn't break anything about git-subtree with this, but otherwise it does build fine now

@shun-iwasawa @RodneyBaker i think i've addressed the previously mentioned issues. Sorry it took a little long for just a minor build system tweak

RodneyBaker:
While awaiting review by @shun-iwasawa...

I noticed this post over in the Opentoonz forum that relates having an issue with special styles and the Advanced Color Selector.  My initial thought is that it's a settings problem but I haven't yet looked into the matter more deeply.  I'm linking it here only because it is said to relate to the Advanced Color Selector:

caryoscelus:
@RodneyBaker your link gives me "The requested topic cannot be found. It might have been deleted." message

RodneyBaker:
Ha!  Perhaps that is a good sign.
In reading the topic I couldn't help but think the problem might relate more to a setting than anything else.  Perhaps that was the case.

So, false alarm.  Looks like you resolved that problem without even having to read the issue.  :)

artisteacher:
I took another look and function seems to work well. I really look forward to this being merged, though I still think the UI could use some work to be more consistent with the rest of OT.
Screen Shot 2019-10-29 at 11 36 30 PM
The "tabs" should really have the same style as the rest of OT - like the style editor. It would also be great to stop the mode buttons overlapping the color wheel. It might be easier to align them either horizontally or vertically rather than in an arc. Instead of the partial white border on the right/bottom, there should be a dark division (like in the style editor) between the wheel and the color sliders.

Screen Shot 2019-10-29 at 11 34 50 PM<br>There's already an icon that's used for configuring in the other windows - it really should be used here instead of writing out configure.

Screen Shot 2019-10-29 at 11 37 18 PM
It's fun having the context menu color based on the color chip, but the text may not always be readable depending on the amount of contrast.

artisteacher:
>  all the tabs has icons on them, but they are using system ones so if those are missing you get labels.

If all of the tabs have icons, the tab style makes more sense. However when the icons are missing, the style feels out of place with the rest of the interface. Personally, I wouldn't rely on system icons that are not universal - unless the fallback looks seamless with the existing interface. The  configure icon is located at toonz/sources/toonz/Resources/menu.svg.

toonz/sources/toonz/menubar.cpp
```diff
@@ -1414,6 +1414,7 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   windowsMenu->addSeparator();
   addMenuItem(windowsMenu, MI_OpenFileBrowser);
   addMenuItem(windowsMenu, MI_OpenFileBrowser2);
+  addMenuItem(windowsMenu, MI_OpenAdvancedColorSelector);
```
manongjohn: If you have to make another update to this PR, can you move this command's menu location up, below one of these lines:

>1403  addMenuItem(windowsMenu, MI_OpenColorModel);
or
>1400  addMenuItem(windowsMenu, MI_OpenStyleControl);

It would make more sense under that grouping of commands.

Please also add the command to `stuff\profiles\layouts\rooms\Default\menubar_template.xml` in the same position.

manongjohn:
I'm on Win10 version and find the following issues with the boxes below the color wheel:

image

1. Height of boxes could be a little taller.
2. The left most box text color is always black and doesn't change from black to white like the other boxes do when changing colors.  It does change if you click on it first.

konero:
The style is not so good unfortunately, I'm too busy to go through the code though, there's a lot.

@manongjohn You'll probably know the most about this, I think he's set a margin to the whole window, I don't know what's inheriting it, but it's after the titlebar, this needs to be removed so the `QTabWidget` can be contiguous with the window, but in doing so you'll need to wrap all the other stuff after the `QTabWidget` in a container to give them margin.

Then in the stylesheets if you navigate to `less/layouts/palette.less` and go straight to the bottom, add this block to style the tabs.

``` CSS
/* -----------------------------------------------------------------------------
   Advanced Color Selector
----------------------------------------------------------------------------- */
#AdvancedColorSelector {
  & QTabWidget::pane {
    // sets border under tabs
    border: 0;
    border-top: 1 solid @accent;
  }
  & QTabWidget > QTabBar::tab {
    // set tab style
    .tab-flat();
  }
  & QTabWidget::tab-bar:top { top: 1; }
  & QTabWidget::tab-bar:bottom { bottom: 1; }
  & QTabWidget::tab-bar:left { right: 1; }
  & QTabWidget::tab-bar:right { left: 1; }
}
```

If it's too difficult to work out for some reason, you can switch to round tabs. To do that change the following from the above...
``` CSS
& QTabWidget::pane {
    border: 1 solid @accent;
```
And...

``` CSS
& QTabWidget > QTabBar::tab {
  // set tab style
  .tab-round();
```

And you'll end up with this...

image

To add separators between slider groups etc, I think you can use `#PlainColorPageParts`, that's what the Style Editor uses.

Just wanted to note that, in-case anyone has trouble figuring out how to style tabs.

RodneyBaker:
The reported bug with multiple fill colors has been reposted in the Opentoonz forum again.
Hopefully this link will remain active:

https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!category-topic/opentoonz_en/ibZ3HHC23rg

manongjohn:
@RodneyBaker I already posted that bug here about 3 hours ago

RodneyBaker:
Thanks @manongjohn.  I missed that posting.

hayoko:
Not to offend, but did no one try to just add different wheels as options to the current style editor? That would be easy to do, I found a square one in the code already... you could just add a circle type with multiple pointers to get that multi-color picker feature with just a few lines. This PR is massive for what is actually feature duplication, which is going to increase build time.

Has anyone thought about this at all? 🤔 

I get the OP said he's building a self-contained widget, but that branch hasn't been updated in three years and this implementation has so many changes it's basically not the same thing anymore.

_tl;dr: It would be quicker and require less overhead just adding options to the current selector._

caryoscelus:
@artisteacher 

> Personally, I wouldn't rely on system icons that are not universal - unless the fallback looks seamless with the existing interface

This makes sense for OT which has its own icons, but for other programs that use system ones that would break consistent look. I know, currently it isn't really used anywhere else anymore, so it's not a big issue to just throw in the icons or something..

@hayoko 

What are the changes of the actual widget do you see? Maybe i've missed to push some update, but i doubt it; as far as i recall there wasn't much updates of the widget, the changes are mostly about its connection to OT. Anyhow, as i mentioned, it's not used elsewhere anymore, but that doesn't change the fact that PR was already pending at that point. If anyone would make alternative one, that would only be easier for me

@manongjohn 

Thanks, i'll look into the gradient issue. Couldn't reproduce black text issue, though. Is it only happening until the first click and then goes on to work until program restart? Or does it stop working again if you select another box?

@konero 

I'll try to look into that along with other changes.. but personally i don't think styling issues are a priority now

manongjohn:
The black text appears the 1st time I open the color picker. It stays black no matter what color I choose until I click that 1st (left most) box. It will then function normally until next OT restart.

manongjohn:
For Linux cmake installations, can you please add the following line just before line 649 in `toonz\sources\toonz\CMakeLists.txt`

```
648 +            "$<TARGET_FILE:ColorWidgets-qt5>"
649          DESTINATION lib/opentoonz
```

RodneyBaker:
Unfortunately, this PR appears to be slipping to 2020 Development cycle.
Is the consensus to allow more development time for this feature?

RodneyBaker:
If the conflict with CMakeLists.txt is resolved I believe this PR can be merged.
Additional changes (if any) can then be made.

manongjohn:
Conflicts, not withstanding, there were 2 issues that I mentioned in prior comments that needed fixing.

manongjohn:
Not sure what is going on with this PR, whether this will ever see the light of day, be abandoned completely or abandoned but integrated into the existing color selector. 

At any rate, if this is to resume in some way, I've added an Issues Checklist to the body of the PR for stuff that was brought up.  At minimum, the 1st 3 should be fixed. The 2 after that are quick changes I would like to see included. Last one can be done in a later PR.

caryoscelus:
Well, i can hardly find enough time to fix things properly, but i did fix the most urgent stuff. "Gradient editing" (for the lack of better term) is working though unuserfriendly, just as a PoC for now. Still have to merge to latest master and fix the ui

manongjohn:
> "Gradient editing" (for the lack of better term) is working though unuserfriendly, just as a PoC for now.

Not quite sure what you mean here.  I tried the artifact and it's still only updating 1 color box. When switching to the second color, it keeps changing the same box (1st box). 

I saw you changed the position of the Advanced Color Selector in `menubar.cpp`.  Thanks.  Need to update it in `stuff\profiles\layouts\rooms\Default\menubar_template.xml` also for it to be picked up on Windows machines.

caryoscelus:
> Not quite sure what you mean here. I tried the artifact and it's still only updating 1 color box. When switching to the second color, it keeps changing the same box (1st box).

When i was testing, i didn't do it on gradients, but on whatever two-color style i found working; and it worked for me, updated the second box after switching. If you would find time for that, a video record of how things went wrong might be helpful

> I saw you changed the position of the Advanced Color Selector in menubar.cpp. Thanks. Need to update it in stuff\profiles\layouts\rooms\Default\menubar_template.xml also for it to be picked up on Windows machines.

Ok, thx, will do

stuff/profiles/layouts/rooms/Default/menubar_template.xml
```diff
@@ -307,6 +307,7 @@
         <separator/>
         <command>MI_OpenFileBrowser</command>
         <command>MI_OpenFileBrowser2</command>
+	<command>MI_OpenAdvancedColorSelector</command>
```
manongjohn, line 310: To match the location of if in code, please move this up below `MI_OpenColorModel`

caryoscelus:
@manongjohn i've fixed both positions, in .cpp and .xml

thirdparty/Qt-Color-Widgets/src/color_wheel.cpp
```diff
@@ -0,0 +1,716 @@
...
+        qreal height = p->triangle_height();
+        qreal slice_h = side * p->val;
+        qreal ymin = side/2-slice_h/2;
+
+        selector_position = QPointF(p->val*height, ymin + p->sat*slice_h);
+        QPolygonF triangle;
+        triangle.append(QPointF(0,side/2));
+        triangle.append(QPointF(height,0));
+        triangle.append(QPointF(height,side));
+        QPainterPath clip;
```
manongjohn, line 412: When compiling the osx version for OTX using Qt 5.15, I get the following compile time error:
```
/Users/travis/build/manongjohn/OTX/thirdparty/Qt-Color-Widgets/src/color_wheel.cpp:412:22: error: variable has incomplete type 'QPainterPath'
        QPainterPath clip;
                     ^
/usr/local/opt/qt/lib/QtGui.framework/Headers/qmatrix.h:54:7: note: forward declaration of 'QPainterPath'
class QPainterPath;
      ^
```
Can easily be fixed by adding `#include <QPainterPath>` at the top of this file and still compiles for prior versions.

RodneyBaker:
The current consensus appears to be that this PR will not be added to Opentoonz.
If this is the case we (collectively) should identify (or reiterate) the reasons why this is to be the case.

For those desiring to work with the Advanced Color Selector it can be found in the Morevna release as well as via the appveyor build here:

https://ci.appveyor.com/api/buildjobs/twfy7xoy15wq5ly6/artifacts/toonz%2Fx64%2FOpenToonz.zip

Outlook for adding this Advanced Color Selector to various forks is estimated as the following:

Opentoonz (master):   Not currently planned for v1.5 and unlikely thereafter without major changes 
OTX:  Unlikely (although some releases do include the feature)
Tahoma:  Unlikely (use of similar feature set more likely)
Morevna:  (as primary sponsor of the feature) addition and maintenance of the feature is likely

All interested parties are encouraged to provide feedback and status regarding decisions in their specific areas of interest (i.e. forks, additional or competing developments, etc.)  As @caryoscelus has put a lot of work into the Advanced Color Selector it will be good to keep him well informed of plans moving forward.

Above all, @caryoscelus we want to keep you in the loop.
As always, thanks for all you do on behalf of the Opentoonz community.

RodneyBaker:
Closing this for now pending renewed interest.

Agnyy:
Hello! Why did they decide to reject this request? It took a long time and it will be very sad if it is thrown out like that.

RodneyBaker:
> Why did they decide to reject this request? 

The primary reason appears to be loss of interest in merging this feature into Opentoonz and its various forks from all parties with last update of code toward resolution of various recommendations in Sep 2020.

Your post is the first signal of interest since that time.

Agnyy:
What could I do to get this request accepted? This request was made not so much for me as for the artists. This system is specially designed for painting.

manongjohn:
I think there was still 1 outstanding issue with Color Gradients not picking the different colors that never got resolved.  It only updated the 1st color.

jeremybullock:
I'm going to throw in my thoughts. . .
I think this feature adds even more complexity to an already complex program.  Having two different panels that essentially do the same thing doesn't make a lot of sense to me.  If there are issues with the main color selector, I think it makes much more sense to fix/enhance that one rather than add an additional one.

Issue #1354, opened 2017-07-29
Bug with "Hide All" audiolevels in Xsheet / Timeline


beeheemooth:
When using the command "Hide All" (both in Xsheet and Timeline), the levels containing audio files continue to work when the frames are scrolled. They "hide" when you apply two commands: "Hide All" and "OFF All".
As I recall, this was not before.

The "Soundtrack" button in the Viewer window turns off the sound only when playing a scene. If you scroll frames in Xsheet, the sound is still reproduced. In my opinion this should not be, since the "Soundtrack" button is (as I know) the only one that can turn off the general sound if you have a lot of audio tracks. Sometimes editing frames is easier to do "in silence", and you have to manually turn off (and turn on) all audio levels. Is it possible to mute the sound with the "Soundtrack" button when scrolling frames? Users pressing a button can decide for themselves whether they need to hear the sound at the moment of animation editing, or not.
 (I hope, I correctly stated the idea)

manongjohn:
I just tested the official 1.1.3 release. The sound still plays there also when you "Hide All". It doesn't play sound if you click the button directly to turn it off.  Seems the bug is in "Hide All".  Sound still plays if you turn preview off or use "Off All". Not sure if that is intentional.

PR #1376, opened 2017-08-04
Xsheet fixes

manongjohn:
KY This PR fixes the following Xsheet issues that have been brought up since implementing the timeline.

1) Issue #1354 - Bug with "Hide All" audiolevels in Xsheet / Timeline

This is a pre-existing issue. Using Hide All or any indirect toggle off of camstand button in a sound column didn't change sound playback settings.  It worked fine if it was directly toggled off/on.

Added the logic needed to correct this issue.

2) Keyframe Context Menu displays "Row" even in Timeline mode (mentioned in Issue #1316)

This can be confusing since "Row" typically means a horizontal line which, in timeline mode, is actually the "Xsheet Column" and could be a source of confusion.

Replaced the text "Row" in keyframe context menu to "Frame".

3) While renaming cell, copying selected text and pasting elsewhere pastes "TCellData" (mentioned in Issue #1316)

This is a pre-existing issue. The current RenameCell object has an event filter which is checking the key/command strokes against OT's list of shortcuts. Under certain conditions when no OT action is performed and the filter returns true or false instead of passing the event forward in case something else might be interested in the event. This was not good for standard text commands like copy/paste/cut/delete since it was trying to perform OT's action of copy/paste/cut/delete cell instead of the standard text editing functions.

Modified the logic to only return true/false if it actually executed an OT action. In all other cases it passes the event forward in case something else wants to act on the event. Additionally added exceptions so standard Ctrl-c/Ctrl-v/Ctrl-x/Delete commands work as standard text command edits instead of OT actions which really shouldn't be happening while trying to rename a field.

Note: I would argue that this event filter shouldn't even be there for renaming cells. While renaming cells, OT specific actions shouldn't be allowed...you are renaming a cell, not trying to cut/copy/paste/delete the xsheet cell. However, I'm not sure why it was added so I'm not removing it, just fixing it.

4) Reduced height of note button in timeline mode. (mentioned in #1338)

The note button was taller than the other buttons around it making the UI look off especially in the neutral theme.

Reduced the height of button only in timeline mode so all buttons were level with each other

5) Level name in cells overruns frame number (mentioned in #1338)

Pre-existing issue. Level name in cells, even when elided, would run into and over frame number.

Added logic to adjust for frame number width.

6) Timeline button text clipped in Mac version (Issue #1338)

Mac uses a wider font than Windows so it was clipping the text on the button.

Updated stylesheet to fix the padding on the button. The padding settng works for both Mac and Windows so the stylesheets are still exactly the same.

7) Frames area blends into Xsheet toolbar in Dark theme (Issue #1338)

Tool bar color and frames area are the same color in Dark theme so they blend right into each other. All other themes are ok.

Now draws bounding box using vertical line color. This creates a separation between toolbar and frames as well as draws the missing vertical line on the 1st column.

8) Hotkeys for Select keyframes crashes OT (issue #1415)

Pre-existing issue. When no keyframes are selected, code still tries to execute rather than returning and not doing anything.

Corrected logic.


turtletooth:
@manongjohn Thanks for doing this.  I really appreciate the way you've jumped in and have really helped out.

I don't feel like frame is the right replacement for row.  Could you do something like this in CellArea::createKeyMenu:

`QAction *action = cmdManager->getAction(MI_SelectPreviousKeysInRow);
  if (whateverFunctionGetsTheStatusOfTheXSheet)
	  action->setText("Select Previous Keys in this Row");
  else action->setText. . .`

manongjohn:
I guess I have mixed feelings about that approach. We don't really use the term "Row" anywhere else in the software. We usually refer to it as "Frames". Though, having said that, we usually say "Level" not "Column" though default column names are usually "Col#".  Maybe this should change as well. Use "Level" and default columns to "Lvl#"?

According to Google, the current Japanese and Chinese translations already translate to Frames, not Rows though I'd need someone fluent in that to confirm.

If you really want me to go that route, I will. These are just more food for thought,

Anyway, I love this software and want to make it even better for those new/hobbist animators like me. Had this been around when I was a lot younger, who knows what I would be doing now. 😊

gab3d:
@manongjohn, if I'm not mistaken, a "Level" is not equal to a "Column":
- "Column" is akin to a Layer in other software.
- "Level" is really Footage/Images within a column/layer.

turtletooth:
I agree with @gab3d 
A frame is a single drawing
A level is a collection of drawings that can be added to one or more columns in the xsheet.

@manongjohn I totally agree with you on the software.  I wish this was around when I was younger.  It's a blast.

manongjohn:
If you assign only 1 level per column, then it is equal, but 1 column can have multiple levels in it so I agree it is not 100% equatable.  The default names for empty columns is Col# and when there was only the Xsheet, that made sense.  That doesn't make sense in Timeline now.  I guess they should really be called Layers, but somehow I think relabeling that would not be received well.

Each row's unit of measure is a single frame of animation. At the end of the day, any changes you make to a Row in Xsheet mode affects that single Frame of animation. In Xsheet mode "Select All keyframes in Row" means you are selecting all the keyframes for that single Frame of animation.  Switch it to Timeline and it doesn't make sense because the Frame of animation is now in a column.  Add to that the confusion that we refer to Xsheet columns as Column.

Either we need to keep true to the definitions of Columns/Rows and switch the code depending on orientation, or keep to how we reference things Columns/Frames (of animation) and just relabel the context menu.  

I guess the question is, which one makes more sense?  I checked a few other translation sites and Japanese and Chinese translations already translate to Frame, not Row.

I apologize if I'm coming across as difficult over something that seems so minor.  I'm just trying to be somewhat consistent.  If there is still disagreement, then I'll revert this change and fix the code to follow the true definition of the words.

turtletooth:
I'm not sure about the other languages, but in the English sense, a frame is one single cell or drawing.  So, selecting all keys on this frame doesn't make sense if it is selecting all keys that are actually on the col/row/layer.  I personally think just switching the text based on orientation is an easy fix, and am cautious about using the word layer for two reasons: the word layer is close to level and may cause confusion between the two, and the timeline is currently top to bottom as far as order goes which is the opposite for those who are used to the term layer.

manongjohn:
We'll just leave it at "we'll agree to disagree".  I'll revert the label change and put in the code to switch based on orientation.

Yeah, I still have to work on reversing the order of the timeline.

gab3d:
I think when we refer to the concept of "Frame" in the Xsheet or Timeline,
it is to the frame of the final composition, not a drawing in a column. So,
in that sense it majes perfect sense the sentence "Select all keys in this
frame", for example.

Regarding, layer order in Timeline mode, I do think we should reverse it to
"bottom-to-top", as in any other application dealing with compositing.


El ago. 7, 2017 5:26 PM, "manongjohn" <notifications@github.com> escribió:

We'll just leave it at "we'll agree to disagree". I'll revert the label
change and put in the code to switch based on orientation.

Yeah, I still have to work on reversing the order of the timeline.

—
You are receiving this because you were mentioned.

<https://github.com/opentoonz/opentoonz/pull/1376#issuecomment-320771223>,
or mute the thread

openanim:
If this can help, the manual says this :

_"The xsheet, the digital version of the traditional exposure sheet, allows
you to control the timing of all of the elements of a scene. It is organized
in columns and rows: each column contains a layer of the animation and
each row represents a frame contents. Columns are divided into cells,
representing the content of that column in a particular frame.
Different items, generically called here as levels, can be loaded into the
scene: animation levels, images for the background and overlays, and
clips.
Other Toonz scenes can be loaded as well: they will be nested inside a
column and considered a sub-xsheet of the current scene.
All of the xsheet contents can be checked in the viewer, where the
animation can be played back, and the scene contents edited."_

turtletooth:
Hey guys.  My bad.  I was thinking you were talking about "Select all keys in this COLUMN" instead of row, even though the originial post clearly says row.  I can totally understand that an entire row could be referred to as a frame.  I was opposed to changing the word column to frame, but am totally on board with the changes as they are right now.

I feel like an idiot.  I'm sorry for the confusion.

I do think that the word column should be changed based on the orientation.  It would be easiest to change that to layer in timeline view(in spite of my statements above), but I would lean toward keeping the word column in xsheet view.

Again, my apologies.  Thank you all for your help and feedback.

manongjohn:
No prob. With the introduction of timeline, there is bound to be some confusion when we use the word "Column".

I'll change the context menu to say Layer instead of Column in timeline mode only.

gab3d:
To me it makes more sense, terminology wise, to adopt the term "layer" for
both cases (xsheet and timeline), and probably it'll be a matter of time
for that change to finnally be adopted.
But for now, I think I'm ok with at least calling them layers in Timeline
mode. :)


El ago. 7, 2017 9:16 PM, "manongjohn" <notifications@github.com> escribió:

No prob. With the introduction of timeline, there is bound to be some
confusion when we use the word "Column".

I'll change the context menu to say Layer instead of Column in timeline
mode only.

—
You are receiving this because you were mentioned.

<https://github.com/opentoonz/opentoonz/pull/1376#issuecomment-320814710>,
or mute the thread

RodneyBaker:
Hehe.
Perhaps I'm the only one not confused here.<joking!>
I think of the 'rows' in the Timeline as Columns that have been tilted over
on their sides.   ;)

I"m not sure why I resist the term Layer but I do.
It may be that I always consider layering as something placed on top of
somethings else (cakes, sediment, etc. etc.).
This doesn't mean that layers can't be tilted vertically but it often means
the structure has been adjusted from its original orientation when found
vertically..
I will further guess that layering has something to do with the concept of
gravity, and I say this in all seriousness (as we don't often consider how
gravity works in virtual spaces).
There is a force at work that is layering all of these things down on top
of each other.

The desire to standardize is a powerful one but there are times when a
standard may be optimal so we need to at least consider that.

@manonjohn  Thanks for your hard work and also for your thoughtful
consideration.  I'm confident that In time all these rough edges will
smooth out.  Good ideas always come back!  :)

manongjohn:
I've decided not to switch Column to Layer in Timeline mode for now.  There are a number of places where "Column" is used in the verbiage and, to be consistent, I'd have to put in logic in many places to switch based on orientation.

For now we'll maintain Column/Frame labels.  Perhaps a future PR can change everything to Layer. Probably should be a topic of discussion.

artisteacher:
The vocabulary certainly can be very confusing, especially now that there are 2 opposite orientations.  The problem that I see with using layers is the fact that vector layers will eventually be exposed in the timeline.  I think there really should be some sort of terminology that distinguishes between the two different layer types. Timeline layers and vector layers? Layers and sub-layers? Rows and layers? Further discussion is a good idea.

adamearle:
You 'keyframe' a column ie. SO (stack order)
Draw in a 'cell'
That makes a 'Frame' in your level

As an animator it doesn't seem that complicated.

If your worried about language there are few ways to look at this I believe.



Object and time



You have already talked about stacking layered objects

Each animation is a block of time isn't it.

Block could be replaced with anything thougg

Housing frames a house of frames.
Block of time
Collection of drawings etc

I think you guys get the idea

I love this development. Really interesting  conversation guys.

Love your work.

Ad





On 9 Aug 2017 2:38 AM, "artisteacher" <notifications@github.com> wrote:

The vocabulary certainly can be very confusing, especially now that there
are 2 opposite orientations. The problem that I see with using layers is
the fact that vector layers will eventually be exposed in the timeline. I
think there really should be some sort of terminology that distinguishes
between the two different layer types. Timeline layers and vector layers?
Layers and sub-layers? Rows and layers? Further discussion is a good idea.

—
<https://github.com/opentoonz/opentoonz/pull/1376#issuecomment-321012071>,
or mute the thread

gab3d:
@artisteacher, what you refer to as "vector layers" are really not layers,
but more "paths" or "objects" inside a vector drawing. If you think it that
way, there's really no confusion.

LAYERS / COLUMNS
Anyway, going back to thinking about the "Layer" concept for the
timeline/xsheet, I've come to a revelation :)
What we have in OpenToonz are not really layers (as in After Effects, for
example), but we have "Tracks" (as in PremierePro or Kdenlive).
The difference is:
Timeline based on Tracks -->  several footage items can live sequentially
in the same level of the composition.
Timeline based on Layers -->  each footage lives at its own level in the
composition, and no other footage item can be at the same level.

On top of that, another confusion we currently have in OT is calling
"Level" to a "footage item"; if a Level would mean a column/layer/track
itself, the terminology would be much clearer to the user.
We could have "Drawings" or "Clips" (whatever you want to call them) living
inside Levels (or tracks/layers).

The confusion currently derives from the idea that a "Level" (a fottage)
can be used at several levels (columns) in the composition! That doesn't
make much sense...
I mean, the word "level" indicates a certain preeminence, or order, respect
other things, but in fact currently it's just an image or a series of
images, that can be used in any order, or any number of times we want. So
in the end "Level" doesn't seem to be a good term to describe what it is or
what it does.

Well, I just meant to bring in some thoughts to the table, for future
decisions.




2017-08-08 13:38 GMT-03:00 artisteacher <notifications@github.com>:

artisteacher:
@adamearle the language does seem complicated to my students when they are completely new to animation. Your explanation is helpful though, thank you!

@gab3d you are totally right, thanks for the clarification.

snapai:
@gab3d I really like the idea of "tracks". I think the issue with "Level" vs "Layer" is just that they're similar-looking words that start with the same letter; so even if English is your first language, parse errors are really easy there. (particularly with a double-story 'a' looking like an upside-down 'e', and 'y' being a 'v' with a tail.)

For an example to complicate ideas further, Blender's video editor uses "[current] frame" for time, "channel" for what layer footage or effects sit in, and "strips" to refer to the individual blocks in a channel - which can be combined like "collapse x-sheet" with "make meta-strip".

(I also like their convention of [tab] in the node and video editors to enter or exit editing sub-containers depending on whether something enter-able is selected, but that's for a different pull's discussion entirely!)

manongjohn:
This is ready to be reviewed and merged.

adamearle:
I have to say I like tracks as well. I think it really works and not only
does it sound kool  but it ties in with audio as well.

Animation tracks
Sound tracks
Note tracks

Sounds great right?

Ad

manongjohn:
@shun-iwasawa @turtletooth  I know you guys are busy with your own fixes and enhancements (lots of good stuff btw!) on top of everything else. Any chance this is going to be reviewed/merged soon? I'll continue to keep adding little Xsheet related fixes to it as they come up until it does.  Just wanted to make sure you aren't waiting for something specific to review.

Not sure what criteria you guys are using for when something should be reviewed/merged, by whom, and whether or not the PR should be accepted or rejected as a whole.  Is there a regular schedule when you guys review open PRs and decide whether or not it will be allowed and who would be reviewing it?

This is not a complaint or criticism in any way. There are quite a number of PRs that have been sitting  a long time and not sure why.  Some of them seem like minor changes but would help with workflow or other things so I'm a little surprised they are still waiting.

While it's good that we don't have too many people just merging things when they want and can help control and steer the development of the application, it's also is a bit of bottleneck and I'm sure things can be overlooked or forgotten easily in this environment.

Anyway, thanks for all that you guys do here, both from an administrative and development point of view. I imaging trying to keep up with everyone's issues, thoughts, opinions and PRs can be quite maddening. :)

konero:
@manongjohn 

If you want to fix a few more bits, these need doing.

**Add a qproperty for the column area outline.**
Because its so bright... there's probably an existing one that can be used.



**Extend xsheet toolbar all the way across the window.**
It has a fixed width, so we can't do anything with it to separate it from the rest of the xsheet, such as giving it a bottom border of its own. It's also missing scrollable toolbar buttons for when the window is too small.



**Separate each spreadsheet block into its own class.**
So we can add some accent borders to separate them better. Right now its all or nothing with #xsheetArea I think it was.



You don't need to, they're not really bugs, I'm just being cheeky. 🥂

manongjohn:
@konero I believe the border for the column area is already in the stylesheet.  I'll take a look, however I'll look to add these changes to the Column UI Redesign PR since I'm already adding a bunch of new qproperty settings for styling the xsheet in general.

@turtletooth If I am not mistaken, you have some changes coming up for the Xsheet Toolbar.  Perhaps you can fix the issue @konero noted?  I did notice when the window gets too small for whatever reason, the toolbar doesn't adjust accordingly and you can't get to the buttons that get hidden because of it.

turtletooth:
@manongjohn I fixed the xsheet toolbar issues in #1425

turtletooth:
@manongjohn I understand what you are saying about the review delay.  I'll take a look at this today.  

As for reviews, anything that makes big and deep changes I tend to let @shun-iwasawa take a look at since he understands the inner workings better than I do.  I try to pull my head out of my code and do reviews pretty regularly.  I tend to look at @shun-iwasawa changes first since his changes are often because a studio is requesting the change, but I'll try and make sure I am not missing other PRs that I can review too.

turtletooth:
@manongjohn Could you please run beautification on this?

manongjohn:
@turtletooth Sure. Mind explaining how? :) I do see a beautification.sh script but I get this error when I run it: 

xargs: clang-format: No such file or directory

I'm guessing I've missed installing something or maybe something in my environment isn't set right?

turtletooth:
I usually have to (from git bash) run: bash beautification.sh 
Including the bash

turtletooth:
Oh.  I remember now.
Hold on.

turtletooth:
[clang-format.zip](https://github.com/opentoonz/opentoonz/files/1244098/clang-format.zip)
You'll need this file.  I keep it in a c:\bin folder and add that folder to the windows path.

manongjohn:
@turtletooth Thanks for reviewing. Beautification done and committed.

Probably should add the clang-format setup and maybe even WinLess or other .less compiler setup and usage instructions to the Windows setup for other potential contributors.

turtletooth:
@shun-iwasawa Does this need to wait for a Jenkins?

shun-iwasawa:
@manongjohn Thank you for the fix and @turtletooth thank you for reviewing this!
Jenkins

konero:
I suggest using VSCode with EasyLess extension like @turtletooth uses, its easier to setup and seems to be more commonly used. The only downside is you need to save the theme file to compile when you make changes into the layouts whereas WinLess is like a watch folder GUI substitute for NPM more ideal for large scale or ground up development since it will compile the theme file when you save any file in the project tree.

EasyLess and WinLess also use different case letters for their hex code output from less color functions, so it might be a good policy to suggest EasyLess all-round so QSS pulls are easier to review without so many deletions and additions being added for no reason.

I also suggest installing Less IntelliSense extension so variables can be read from the default theme file as you type.

vscode-ex

I've changed the file order so I could use glob patterns to exclude all theme files except for the default one, otherwise variables are listed for as many times as there are theme files. I'll submit these changes in my next PR and a vscode settings.json file.

turtletooth:
LGTM Thanks @manongjohn I really appreciate your help on OpenToonz.

manongjohn:
@konero Thanks for info. Will switch to VSCode/EasyLess going forward.  Winless was a little annoying whenever I switched branches, if I forgot to turn it off. :)

@turtletooth Thanks for taking the time to review and merge.

manongjohn:
The following issues can be closed: 1354, 1338, 1415

turtletooth:
I closed the issues.  Thanks @manongjohn

PR #1380, opened 2017-08-05
Update Playbar Icons

konero:
playbar-icons-2

Changes the stroke color to make icons appear sharper on the very dark theme. The stroke is now transparent, so it can mix with the theme color. It was feathering before making them look blurry.

Base Fill
`#E6E6E6`

Hover/Click Fill
`#FFFFFF`

Border (140 Alpha)
`#000000`

**NOTE:** Can someone test these on Mac? Last time I submitted SVG they didn't fare so well on Mac.


turtletooth:
These look good.  Why did you close the other PR?

turtletooth:
What do you think about flipping the arrow on the loop icon?  Right now it comes off a little bit like a reverse icon.

gab3d:
play
It be nice if the playback buttons layout would be like in this image.

- Pause button would appear instead of Play button when playing, it's space in the UI could be saved and the user would not have to move the mouse (and change his sight focus) just to pause a playing animation.
- Cycle button which is really a "mode", would be next to playback control buttons, but not "in the middle" of them, which makes it easier to read the whole set of buttons.

just my two cents.

konero:
@gab3d Good point, this PR will remain as a SVG swap only though.

@turtletooth I updated the loop icon, I'll re-submit a new PR for theme fixes, there were some big changes so I wanted to use it in production before pushing anything.

turtletooth:
LGTM.  Thanks @konero

PR #1381, opened 2017-08-05
Allow All Windows Builds to be Portable 

turtletooth:
This allows all Windows builds to be converted to a portable version simply by placing a "portablestuff" folder in the same directory as the main executable.  The portablestuff folder should have the same contents as the normal stuff folder.

This is very useful for being able to test builds without having to reinstall OpenToonz from the nightlies.  Users would simply be able to download the artifacts from AppVeyor, add a "portablestuff" folder and test it.

This also streamlines the process for creating portable versions for school or other settings where installing the software is limited.


turtletooth:
Just tested on the AppVeyor build for this PR - works great.  Just copy over the mypaint dlls and the stuff folder - instant portable version.

RodneyBaker:
I have been looking for a word to describe how interesting the abiity to make any build portable is/will be.  At present all I can come up with is 'crazy cool'... because, well, it is.  :)

I'm not sure how you come up with this stuff but keep it up!

Thanks Jeremy!

turtletooth:
@RodneyBaker This idea came from the number of people asking to try out new features before they were merged.  This would simplify things a whole lot.

gab3d:
this is really great.
thank you!! :)

turtletooth:
@shun-iwasawa While I am fixing the merge conflicts on other PRs, would you mind looking at this when you get the chance?

toonz/sources/toonz/main.cpp
```diff
@@ -109,6 +109,10 @@ const char *applicationFullName = "OpenToonz 1.1.3";
 const char *rootVarName         = "TOONZROOT";
 const char *systemVarPrefix     = "TOONZ";
 
+QString workingDirectoryTmp  = QDir::currentPath();
+QByteArray ba                = workingDirectoryTmp.toLatin1();
+const char *workingDirectory = ba.data();
```
shun-iwasawa: Maybe obtaining the path to the working directory can be moved from here to the constructor of `EnvGlobals` as well ?

turtletooth:
Thanks for the review @shun-iwasawa I made the suggested changes.

shun-iwasawa:
Jenkins

toonz/sources/toonz/main.cpp
```diff
@@ -161,6 +161,7 @@ DV_IMPORT_API void initColorFx();
 static void initToonzEnv() {
   StudioPalette::enable(true);
 
+  TEnv::setWorkingDirectory();
```
shun-iwasawa: How about calling `EnvGlobals::setWorkingDirectory()` in the constructor of `EnvGlobals`, instead of calling it externally here?
By doing this, I think we can share the portable-mode with other command-line binaries such as `tcomposer`.

turtletooth: Good idea.

turtletooth: I'm making the changes now.  What are your thoughts on changing the portablestuff folder to just stuff?  I don't think any platforms have the stuff folder in the app folder by default.  portablestuff feels a little long, but I can go either way.

shun-iwasawa: @turtletooth Thank you for the modifications. Well, I prefer "portablestuff" as it is more self-describing.

shun-iwasawa:
Jenkins again

toonz/sources/toonz/main.cpp
turtletooth: Cool.  I'm good with that.

turtletooth: If you are feeling daring, you could take a look at the sound pr.  I think I fixed it.  Development on a mac isn't nearly as fun as Visual Studio.  Are you using Qt Creator or XCode on Mac?

shun-iwasawa: I'm checking the sound PR now. Thanks for fixing that too.
I'm using CLion for OSX development, I totally agree that it is not comfortable as working on MSVC.

shun-iwasawa:
This will enhance usability. Thanks @turtletooth !
LGTM

RodneyBaker:
So let me get this straight.

If a user places a folder named "portablestuff" in their installation folder OpenToonz automatically becomes portable?

turtletooth:
Yes, you can turn your install into a portable one by putting a portable stuff folder in the installation folder and then moving it wherever you like.

That is how OTX is built.  I just build it normally and then put the portablestuff folder in the directory.

RodneyBaker:
Thanks for the clarification.  That is what I thought you were saying and hoping too.

That is indeed 'crazy cool'.  :)

jcome:
Amazing.

Can we add a portablestuff folder in default build, so every installation is natively "Portable" ?

RodneyBaker:
I like @jcome's suggestion.
Then if anyone want's a non-portable release all they have to do is delete (or rename) that folder.

I do find myself wondering... if OT is set up as portable and then the folder renamed or removed, does OT then automatically create the stuff folder in the root directory of C:Drive? 

I guess there is one way to find out and that is to install and check out the results.  ;)

turtletooth:
If a version of OT is not already installed, the required registry entries wouldn't be there to locate the stuff folder.  The program also can't recreate the contents of the stuff folder from scratch.  

If an installation already exists, you should be able to delete the portablestuff folder and it should then be able to find the other stuff folder from the installation - however, in the case of OTX or any build where the contents of the stuff folder have changed, you might be missing required pieces.

I do like idea of making portable versions available for releases and especially nightly builds.  If users could test the nightly builds without reinstalling anything, that would be awesome.

turtletooth:
One issue with that - and you may have noticed in OTX.  Zipped files (at least made in Windows) don't support many characters from other languages, so I have had to rename the languages to their English name.

jcome:
> so I have had to rename the languages to their English name.
might be a little out of topic, is it better to not use folder name as language list item label? Instead, we localize UI language list item via standard UI translation method as most of software does.

RodneyBaker:
Both good and bad news here.

I installed the latest nightly build.
Created a 'portablestuff' folder in the installation directory.
Launch OpenToonz (directly from the executable)
OpenToonz tries to launch but crashes due to missing files.
(I sometimes got no response but once saw dialogue that suggested a reinstall and specifically mentioned some menu files.  I wish I would have captured those)

Reinstall got same result.

Now the good news.
I decided to copy the contents of the stuff folder in root of C:Drive into the portablestuff folder.
I copied and opted to skip any duplicate files... because I figure there could be one or two that operate differently on a portable install.

And it works!

So if you get reports of the neo-portable not working that's one way to get it to successfully launch.
That was running the neo-portable from my desktop.
Now to run it from my USB.  :)

turtletooth:
Yes, just making a folder called portablestuff isn't enough.  It has to have all the contents of the normal stuff directory.

RodneyBaker:
Ha!  I thought that just making an empty folder would be too simple.  ;)
But seriously though, still awesome.

AND... Confirmed.  It works great via USB too!  :)

RodneyBaker:
Apparently this feature allowing all windows builds to be portable no longer works.
Will have to investigate...

turtletooth:
@RodneyBaker I just tested and it still works, except that if you have a current install of OpenToonz it will find that stuff folder first and not load as portable.  I had to rename my installed stuff folder for it to find the portablestuff folder.  I think that got changed somewhere.

manongjohn:
This still works and looks at portable first for me even when I still have my installed stuff folder present.

turtletooth:
Weird.  Just tested again and it worked as @manongjohn mentioned.  So I think we're good.

RodneyBaker:
Weird indeed.  I just created another portable and this time it worked just fine..
If I run into the glitch again I'll post more info on it.

PR #1386, opened 2017-08-08
Pick a Font

turtletooth:
This PR makes Segoe UI the default on Windows, but it also allows you to choose whatever font you want and set it to normal or bold.

Here is Segoe:
segoe

Now with more Wingdings:
wingdings

Wingdings makes everything better:
moredings

Here is where you set up the font:
prefs


RodneyBaker:
Oh dear!  I sense the Klingons will now be animating war in OpenToonz!


Very nice Jeremy.
The segoe font is very easy on the eyes.   :)


Unrelated aside:  Since you posted a screenshot of the popup I"ll use this
as an opportunity to highlight a workflow issue I've been having with it.
Feel free to ignore as off topic and/or irrelevant but it's something I've
noticed.
When I navigate the startup popup my hand/brain tends to want to click on
that right-most button 'Open Another Scene' rather than the 'Create Scene'
button on the left.
I'm wondering if having the Open Scene area on the left and the Choose
Project area (with it's associated Scene creation area) might not be more
appropriate.
I tend to create a new Scene almost every time I open OpenToonz so I
inadvertently try to click the wrong button a lot.
Regardless, I am in awe of OpenToonz and the thought that has gone into it.

Somewhat related aside:  The japanese word 'segoi' (See-go-ee = amazing!)
comes to mind every time I open this software although I understand that is
pronounced differently than 'segoe' (which I read is pronounced 'See-go').

yamisama:
Cool, thanks for your fantastic work, as always!

May I suggest one point for testing: What happens when the language is set to something like Japanese, Chinese or Korean? A lot of fonts don't actually contain the necessary characters. So for a user using the UI in one of these languages, the UI text might disappear on them or become otherwise unreadable depending on the font that is chosen. That could potentially make it rather difficult for a user to return to a "normal" CJK font, a bit like in the Wingdings example.

For Japanese users, even just switching the UI to Japanese by navigating the English UI when they they start the software for the first time can be quite a hurdle as it apparently always defaults to English instead of the system language, so I imagine an inadvertent font change might trap a few people as well.

turtletooth:
@yamisama I understand the concern.  My original PR was to simply switch the font to Segoe UI, but it was requested that I give the user the chance to choose for themselves.  I'm not sure there is any way to prevent a user from choosing a font that has the effects you mention without limiting them to the choices of Arial or Segoe UI.  I think that would actually be enough, and would be an easy change if others would like that.  One benefit of this model is for dyslexic users.  There are fonts designed to help users with dyslexia and it's cool to be able to let them choose a font that works best for them.

jpturcotte:
Is this helping in any way?
https://github.com/Microsoft/vscode/issues/5260#issuecomment-211138262
https://github.com/Microsoft/vscode/issues/5260#issuecomment-211505841

turtletooth:
Fixed conflicts.

shun-iwasawa:
Jenkins

toonz/sources/toonz/main.cpp
```diff
@@ -636,9 +637,14 @@ int main(int argc, char *argv[]) {
...
-
-  std::string family = EnvSoftwareCurrentFont;
-  myFont             = new QFont(QString(family.c_str()));
+  std::string fontName =
+      Preferences::instance()->getInterfaceFont().toStdString();
+  std::string isBold =
+      Preferences::instance()->getInterfaceFontWeight() ? "Yes" : "No";
+  if (fontName != "") EnvSoftwareCurrentFont = fontName;
+  EnvSoftwareCurrentFontWeight               = isBold;
+  std::string family                         = EnvSoftwareCurrentFont;
```
shun-iwasawa: It works fine as it stands.
But I think it would be clear if you could completely remove ENV values `EnvSoftwareCurrentFont` and `EnvSoftwareCurrentFontWeight` as they will not be used anymore.

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -339,6 +340,23 @@ void PreferencesPopup::onRoomChoiceChanged(int index) {
 
 //-----------------------------------------------------------------------------
 
+void PreferencesPopup::onInterfaceFontChanged(int index) {
+  QString font = m_interfaceFont->currentText();
+  m_pref->setInterfaceFont(font.toStdString());
+  if (font.contains("Comic Sans"))
+    DVGui::warning(tr("Life is too short for Comic Sans"));
+  if (font.contains("Wingdings"))
+    DVGui::warning(tr("Good luck.  You're on your own from here."));
```
shun-iwasawa: Nice warning ! :-)

shun-iwasawa:
@yamisama 

> May I suggest one point for testing: What happens when the language is set to something like Japanese, Chinese or Korean? 

I don't know the exact reason, but we can display Japanese UI with Arial font in the current release (it seems to show Japanese text with MS P Gothic in my environment).  It may be related to the [font linking feature](https://msdn.microsoft.com/en-us/library/ee489864.aspx) in Windows. I'm not sure though.

turtletooth:
Changes made.

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed it works fine on OSX as well. Thank you @turtletooth for the enhancement! LGTM

PR #1391, opened 2017-08-10
Fix cleanup problem when the source image has no DPI

shun-iwasawa:
This PR fixes the following problem which is reported by some Japanese animation production:

#### PROBLEM
When cleanup raster levels which has no dpi (for instance, non-antialiased TGA images), the first frame looks OK, but the following frames change its size irregularly and the images are not resampled nor smoothed. 
This problem does NOT occur if the source raster level has dpi information in its file format  (such as TIF images).

#### CAUSE OF THE PROBLEM
For now, in cleanup procedure the dpi of source image is obtained for each frame. 
If the source raster level has no image dpi, then the custom dpi in the level settings is retrieved as the source image dpi. 
However, just after processing the first frame, the custom dpi will be overwritten by the dpi of cleanup camera. This is why the second and the following frames becomes wrong size.

I fixed this behavior to use always the dpi of the first frame throughout the level.

In this PR I also fixed the following cleanup-related problems:
- Added missing `m_firstLevelFrame` initialization in the `CleanupPopup` constructor.
- Fixed color of cells of "partially cleanupped" level.

#### CAUTION
Cleanup of the raster levels with no dpi still has issues, in the case you cleanup partial frames and then cleanup the rest later.
Even if you cleanup the level partially, the level's dpi will be overwritten by the cleanup camera's dpi. So you need to revert the custom dpi before you start cleanup of the remaining frames.

### DIFF
--- a/toonz/sources/include/toonz/tcleanupper.h
+++ b/toonz/sources/include/toonz/tcleanupper.h
@@ -57,7 +57,7 @@ public:
 
 class DVAPI TCleanupper {
   CleanupParameters *m_parameters;
-  TPointD m_customDpi;
+  TPointD m_sourceDpi;
 
 private:
   TCleanupper()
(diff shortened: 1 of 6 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1397, opened 2017-08-11
Style Editor Overhaul

turtletooth:
Summary of Changes:
- The "Plain"(now Color) tab remembers which color sections (hsv, matte, rgb, wheel) are visible on restart for each room.
- The "Plain" tab has a new button next to rgb to allow switching the style editor from vertical to horizontal, so the style editor can be placed below the viewer without all the sliders being squished.
- The Special, Custom and Vector Brush tabs are now consolidated under the new Vector tab.  There are buttons in there to toggle on or off the different types.  (See Photo)
- The Plain tab has been renamed Color
- Special has been renamed Generated
- Custom has been renamed Trail
- MyPaint Brush has been renamed Raster.  This isn't to take away from the MyPaint-ness of the feature, but to more clearly communicate with users what the tab is for.

style

vector


gab3d:
currently HSV M RGB strings, next to sliders, are not translatable.
could ths be done?


El ago. 11, 2017 19:42, "Jeremy Bullock" <notifications@github.com>
escribió:

This will eventually have a bunch of fixes.

Right now it has:

   - The "Plain" tab remembers which color sections (hsv, matte, rgb,
   wheel) are visible on restart for each room.
   - The "Plain" tab has a new button next to rgb to allow switching the
   style editor from vertical to horizontal, so the style editor can be placed
   below the viewer without all the sliders being squished.

[image: style]
<https://user-images.githubusercontent.com/4576381/29234414-0dfa30c4-7eb4-11e7-97f3-166a6f2e00d4.PNG>
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/opentoonz/opentoonz/pull/1397
Commit Summary

   - remember style editor settings and add horizontal

File Changes

   - *M* toonz/sources/include/toonzqt/styleeditor.h
   <https://github.com/opentoonz/opentoonz/pull/1397/files#diff-0> (29)
   - *M* toonz/sources/toonzqt/styleeditor.cpp
   <https://github.com/opentoonz/opentoonz/pull/1397/files#diff-1> (179)

Patch Links:

   - https://github.com/opentoonz/opentoonz/pull/1397.patch
   - https://github.com/opentoonz/opentoonz/pull/1397.diff

—
<https://github.com/opentoonz/opentoonz/pull/1397>, or mute the thread

gab3d:
thank you! :)

2017-08-15 18:55 GMT-03:00 Jeremy Bullock <notifications@github.com>:

blurymind:
It would be awesome if the style editor is also context sensitive and only show the vector brush tab only if the active level's type is 'vector' and raster brush tab only when the active level is raster

That will prevent the user from selecting a brush style that doesnt do anything on the level type they are on

konero:
How about placing the channel buttons and the layout options into a drop-down menu somewhere like the palette window?

style-menu

turtletooth:
@konero That's a good idea.  I'll look into that in the future.

artisteacher:
I'm looking forward to this! It will be very helpful for my students.

gab3d:
...just trying to visualize how compact the current Style Editor could get.
I think the more space for other editors, the better.

compact_style_ed

artisteacher:
@gab3d, I like the compactness. I would move the menu button to the top though.

gab3d:
@artisteacher 
I know... I first I tried to put them at the top, then I realized they had to sit in a place "inside" the Color tab (so they won't be visible when other tabs are selected) AND without taking a full row for themselves, so having removed the wheel/rgb/hsv row of buttons to save space, the only place remaining was the bottom row. 
Perhaps someone could come up with a better solution.

Thinking about making the color panel even smaller, I think the color sliders' height could also be reduced a little without sacrificing its function.

adamearle:
This is just a thought and before mypaint brushes where introduced this may have worked but could settings sit int eh same tab as to what you have selected?

A lot of people don't seem to realize what they have selected has an effect on the settings tab. this might be well worth having a look.

Drawing a complete blank on this one as we should try and future proof it as much as possible.
Perhaps some kind of hotlink within the tab? or when people select what they need settings might gently breathe changing color letting people know there are setting in there? Perhaps this could be a setting that is on by default and then after people learn it they may turn the setting off?

Night guys
Awesom work ;D

Ad

gab3d:
@adamearle 
To me, the best long term solution for this matter (and others) would be that there was a general Settings panel that showed the parameters and options of what item is on focus in OT. That is, a level, a column, a palette style, a camera, an fx node, etc.
That way, with a single screen area reserved for settings, we could always be able to be aware of and, if needed, change the relevant settings for the current selected item.

But currently there's no effort towards that goal, so that's just a hopeful wish. :)

toonz/sources/toonzqt/styleeditor.cpp
```diff
@@ -1505,46 +1505,49 @@ PlainColorPage::PlainColorPage(QWidget *parent)
...
-  QPushButton *hsvShowButton   = new QPushButton(tr("HSV"), this);
-  QPushButton *alphaShowButton = new QPushButton(tr("Alpha"), this);
-  QPushButton *rgbShowButton   = new QPushButton(tr("RGB"), this);
-
-  QFrame *wheelFrame = new QFrame(this);
+  m_wheelShowButton         = new QPushButton(tr("Wheel"), this);
+  m_hsvShowButton           = new QPushButton(tr("HSV"), this);
+  m_alphaShowButton         = new QPushButton(tr("Alpha"), this);
+  m_rgbShowButton           = new QPushButton(tr("RGB"), this);
+  m_toggleOrientationButton = new QPushButton(tr("↔"), this);
```
shun-iwasawa: In my environment with MSVC 2013 Japanese version, build failed because of the characters "↔" and "↕".
Can you please replace `"↔"` in lines 1512 and 1745 by `QChar(0x2194)`, and `"↕"` in line 1752 by `QChar(0x2195)` ? They will specify the same character with unicode.

shun-iwasawa:
And one more point I would like you to modify: 

`m_styleChooser->setCurrentIndex(7);`

in lines 3155, 3251, and 3298 should be replaced by 

`m_styleChooser->setCurrentIndex(5);` or 
`m_styleChooser->setCurrentIndex(m_styleChooser->count()-1);` (this seems to be more robust).

Since the index of the empty page is changed.

turtletooth:
Thanks @shun-iwasawa I made the changes.

toonz/sources/toonzqt/styleeditor.cpp
```diff
@@ -1691,6 +1698,81 @@ void PlainColorPage::setColor(const TColorStyle &style,
...
+  // if (m_isVertical == isVertical) return;
+  // not returning even if it already is the same orientation
+  // to take advantage of the resizing here
+  // this is useful for the first time the splitter is set
+  // afterwards, it will be overridden by the saved state
+  // from settings.
+  m_isVertical = isVertical;
+  if (isVertical) {
+    m_vSplitter->setOrientation(Qt::Vertical);
+    m_toggleOrientationButton->setText("↔");
```
shun-iwasawa: Sorry for my late reply, but can you please replace this "↔" by `QChar(0x2194)` as well?

shun-iwasawa:
Jenkins

shun-iwasawa:
Thank you @turtletooth for the UI improvement ! Especially I like the organized tab.
Although there still be the ideas for refinement, IMO it can be done in the future. LGTM

PR #1400, opened 2017-08-13
Remove Mac qss Files - They are identical to the main qss files.

turtletooth:
Maintaining duplicate qss files is unnecessary.

When OpenToonz only supported two operating systems, the potential to have two qss versions in case we needed to factor for differences made sense, but now that there are more (and even more if other PRs get approved) it makes more sense to me to make sure that the single qss version works for all systems.


shun-iwasawa:
LGTM

PR #1403, opened 2017-08-14
Saving strings for the "style sheet" and the "language" preference options

shun-iwasawa:
This PR will change the behavior of preferences to save both the "style sheet" and the "language" option as string ( name of the option ), not as integer ( index of the combo box ).

For now, when you introduce a new style or language, an appearance of UI may be unexpectedly changed since the new item is inserted above the current item.
After merging #1332 ,  I'm going to think that this modification is needed for introducing new style or language safety in the future.

### PLEASE NOTE
This PR will reset the language to "English" and the style to "Astral_072_Dark" respectively for the first time you launch the software.


shun-iwasawa:
Jenkins

shun-iwasawa:

masafumi-inoue:
LGTM

Issue #471, opened 2016-06-18
insert frame between two frame in level strip, change every frame in xsheet


2dvision:
I'm confused. When insert frame between frame 0001 and 0002 in level strip, new frame number change to 0002 and change every frame number and drawing in xsheet.

what do you think? Is this good or bad?

![number frame](https://cloud.githubusercontent.com/assets/16509592/16172433/34324cd2-359e-11e6-950a-42fcb1e8a122.gif)

Rayek:
This is expected wanted behaviour: suppose the animator wants to add inbetweens - in that case this makes a lot of sense.

So, very good indeed.

blurymind:
I think that this is not good design, because some times you would like to insert new drawings between the ones you have.
@Rayek precisely because you might want to add inbetweens the current design is a problem.
You add a new inbetween in your flipbook and suddenly all of your xsheet drawings change unexpectedly.

A better design would be toonboom's and tvpaint's approach. Each drawing has a unique ID that is not affected by the order that it has in the flipbook. The x-sheet uses the unique id of the frame instead of it's order number.

blurymind:
or wait.. i need to investigate this. It might be beneficial in some cases.

blurymind:
in my case it is a bit of a nuisance because it makes me reorder frames in the x-sheet.

blurymind:
With this design you discourage the user from inserting any inbetween frames in the flip book. I tried it on a simple animation and I have to say that it creates needless work for the artist with no benefits!

The workflow is usually:
1. draw key poses in flipbook
2. drop them onto the xsheet and adjust their timing (hold some longer than others)
3. insert breakdowns (flipbook, then drag to xsheet)
4. insert inbetweens (flipboox and then drag to xsheet)

from step 3 - immediately all the timing that I have set for myself in step 2 becomes worthless - as all of the drawings are not at the right place. So now I have to redo step 2

Please implement (as an option if some are against it) the ability for drawings to have a unique ID variable  under the hood (each time a drawing is created, it gets a unique ID) and for the xsheet to use their Unique ID variable instead of their order in xsheet variable.

blurymind:
Actually one advantage of having the current design is numbering clarity in the xsheet. Inbetween drawings and key drawings become very obvious - as this forces you to have sequences like this:

1,4,5,6,2,7,8,9,3

4,5,6 is an inbetween
7,8,9 is an inbetween
and 1,2,3 are keys.

[...]

PR #1405, opened 2017-08-14
Level Strip Overhaul

turtletooth:
This fixes a bunch of things in the level strip:

- Level strip will set to a fixed width based on icon size
- Drag select works in the level strip (reset a group selection by clicking a single frame)
- Drag and drop to xsheet now works consistently for group selections too
Note: if a frame is not selected and is dragged it will attempt to create a drag selection.  If you want to drag a single frame that isn't already selected to the xsheet, clicking and dragging the number area below the frame will trigger a drag and drop to xsheet.  This sounds more complicated that it feels in actual use.
- Scroll bar stops at the bottom ( + 1 frame)
- A frame navigator will show up for toonz raster and raster levels that indicates the zoom and location of the viewer.  This navigator can be moved with middle click and drag. (This does not change how it works for Studio Ghibli rooms.)

I thought about changing the colors since red is a little harsh, but that can wait for another PR.

viewer


2dvision:
is it possible to add (optional) unique ID for Level Strip?  #471

turtletooth:
@2dvision I'm sorry but that isn't something I'll be able to work on right now.

toonz/sources/toonz/filmstrip.cpp
```diff
@@ -400,13 +402,31 @@ void FilmstripFrames::hideEvent(QHideEvent *) {
...
+  ComboViewerPanel *inknPaintViewerPanel =
+      TApp::instance()->getInknPaintViewerPanel();
+  if (m_isGhibli && inknPaintViewerPanel) {
+    m_viewer = inknPaintViewerPanel->getSceneViewer();
+  } else {
+    m_viewer = TApp::instance()->getActiveViewer();
+  }
+  if (m_viewer) {
+    connect(m_viewer, SIGNAL(onZoomChanged()), this, SLOT(update()));
+    connect(m_viewer, SIGNAL(refreshNavi()), this, SLOT(update()));
```
shun-iwasawa: For now, this function `getViewer()` connects signals every time you enter the cursor in the Level Strip and causes multiple connection.
(Just as a side note, you can dump the connection info with `QObject::dumpObjectInfo()` under the debug mode)
I think you meant (and forgot) to use `viewerChanged` flag for avoiding such behavior.

How about modifying this function like as follows? Please note that the former block of modification is for connecting the signals properly on launch:

```
@@ -412,9 +412,11 @@ void FilmstripFrames::hideEvent(QHideEvent *) {
 
 void FilmstripFrames::getViewer() {
   bool viewerChanged = false;
-  if (m_viewer && m_viewer != TApp::instance()->getActiveViewer()) {
-    disconnect(m_viewer, SIGNAL(onZoomChanged()), this, SLOT(update()));
-    disconnect(m_viewer, SIGNAL(refreshNavi()), this, SLOT(update()));
+  if (m_viewer != TApp::instance()->getActiveViewer()) {
+    if (m_viewer) {
+      disconnect(m_viewer, SIGNAL(onZoomChanged()), this, SLOT(update()));
+      disconnect(m_viewer, SIGNAL(refreshNavi()), this, SLOT(update()));
+    }
     viewerChanged = true;
   }
   ComboViewerPanel *inknPaintViewerPanel =
@@ -424,7 +426,7 @@ void FilmstripFrames::getViewer() {
   } else {
     m_viewer = TApp::instance()->getActiveViewer();
   }
-  if (m_viewer) {
+  if (m_viewer && viewerChanged) {
     connect(m_viewer, SIGNAL(onZoomChanged()), this, SLOT(update()));
     connect(m_viewer, SIGNAL(refreshNavi()), this, SLOT(update()));
   }
```

gab3d:
@turtletooth 
One behaviour of the Level Strip that is not very easy to understand at first (and I don't know if your PR addresses it in some way) is the fact that if you currently go to the last frame of the Level Strip and use the mouse wheel to scroll it even further (yes, it allows you to do that), then select that last unexistent frame and draw something in the Viewer, you are creating a new frame for that Level that is not the inmediate consecutive number to the last one, but instead the newly created frame is numbered according to that non existant frame that you've previously had selected.

I don't know if I consider this as inconvenient or not, but what is indeed unfriendly is that the user has no idea of the frame number he/she wil be creating before it is already done. 

In my opinion that could be improved by:
- also showing the numbers of that unexistant frames, when you scroll down past the current last one.

and/or

- perhaps it'd be a good idea too, for that non existant frames, to also display a + symbol somwhere, to iindicate and allow to create such new frame by clicking on it, if the user whishes to do so.

[Offtopic comment: This last idea could also be used in the Palette Editor, to allow for the creation of a new Style. Currently the button to create a new style sort of "hides" between other very similar (and less often used) ones. Besides that, depending of the place the user places the Style Editor, the create new style button could end out of screen sight (when the panel is docked in somewhat narrow places.
So I think that having a last empty Style rectangle with a big plus symbol inside would inmediatelly communicate to the user the way for creating a new Style, should he/she whishes to do so, in a much more direct and friendly way.]

turtletooth:
@shun-iwasawa I tool care of the changes you requested.  

@gab3d This change also fixes the infinite scroll on the level strip.  You can now only scroll down one empty frame, so the frame numbers will be consecutive.

artisteacher:
Will the level strip be responsive? I'd love to be able to see the full icon at a narrower size - based on the width of the level strip. Though I suppose that could be unfortunate if the level strip was expanded.

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed it works fine. Setting the active scene viewer for the navigator feature is nice. I think I'll take the same strategy as yours in the StudioGhibli layout as well.
Thank you @turtletooth for the improvement! LGTM

artisteacher:
This is great! Is there an alternative shortcut for moving the navigator? Middle mouse click is not possible with a trackpad.

gab3d:
@artisteacher 
good point. 
for all notebook users it'd be good if there was a default way to use it via trackpad too.

PR #1355, opened 2017-07-30
Show Renamed Names in MacroFx

turtletooth:
renamed

If a user has renamed a fx (ctrl + double click), the new name isn't used for the port name in macroFx.  The renamed name is probably more descriptive for the user than the generic name - if they have gone through the work of renaming it.  They can still see the original name/type by hovering over the port.

PR #1407, opened 2017-08-16
Camera Capture enhancement : Added the information label and the previous level button

shun-iwasawa:
This PR enhances the Camera Capture feature in following 3 points: 

* Added the information label below the level name field, which indicates how many frames exist in the level and what will happen when you hit the capture button with the current parameters. It shows one of the following 4 status:
  * NEW - The captured frame will be the first frame of the newly-created level.
  * ADD - The captured frame will be added to the existing level.
  * OVERWRITE - The captured frame will overwrite the existing frame.
  * WARNING - The frame cannot be captured for some reasons.
Furthermore, you can see the detailed information in the tool tip of the information label.

* Added a button to select the previous level name on the left side of the level name field.

* Modified the cell placement behavior after the image captured.

These features were requested by some Japanese animation studio.

camcap_frameinfo

### DIFF
--- a/toonz/sources/include/toonz/namebuilder.h
+++ b/toonz/sources/include/toonz/namebuilder.h
@@ -33,7 +33,9 @@ public:
 //-------------------------------------------------------------------
 
 // NameCreator genera la sequenza 'A', 'B', ...
-class DVAPI NameCreator final : public NameBuilder {
+// inherited by FlexibleNameCreator
+class DVAPI NameCreator : public NameBuilder {
+protected:
   std::vector<int> m_s;
 
 public:
--- a/toonz/sources/toonz/penciltestpopup.cpp
+++ b/toonz/sources/toonz/penciltestpopup.cpp
@@ -450,6 +450,19 @@ bool findCell(TXsheet* xsh, int col, const TXshCell& targetCell,
   return false;
 }
 
+bool getRasterLevelSize(TXshLevel* level, TDimension& dim) {
+  std::vector<TFrameId> fids;
+  level->getFids(fids);
+  if (fids.empty()) return false;
+  TXshSimpleLevel* simpleLevel = level->getSimpleLevel();
+  if (!simpleLevel) return false;
+  TRasterImageP rimg = (TRasterImageP)simpleLevel->getFrame(fids[0], false);
+  if (!rimg || rimg->isEmpty()) return false;
+
+  dim = rimg->getRaster()->getSize();
+  return true;
+}
+
 }  // namespace
 
 //=============================================================================
@@ -2318,7 +2331,7 @@ void PencilTestPopup::refreshFrameInfo() {
                             .arg(fidsToString(fids, letterOptionEnabled));
         // if the frame exists, then it will be overwritten
         if (frameExist) {
-          labelStr += tr("OVERWRITE one of");
+          labelStr += tr("OVERWRITE 1 of");
           infoType = OVERWRITE;
         } else {
           labelStr += tr("ADD to");
@@ -2348,10 +2361,18 @@ void PencilTestPopup::refreshFrameInfo() {
     level_sameName->getFids(fids);
 
     // check resolution
-    TDimension dim = level_sameName->getSimpleLevel()->getResolution();
+    TDimension dim;
+    bool ret = getRasterLevelSize(level_sameName, dim);
+    if (!ret) {
+      tooltipStr +=
+          tr("\nWARNING : Failed to get image size of the existing level %1.")
+              .arg(QString::fromStdWString(levelName));
+      labelStr += tr("WARNING");
+      infoType = WARNING;
+    }
     // if the saved images has not the same resolution as the current camera
     // resolution
-    if (camRes != dim) {
+    else if (camRes != dim) {
       tooltipStr += tr("\nWARNING : Image size mismatch. The existing level "
                        "size is %1 x %2.")
                         .arg(dim.lx)
@@ -2378,7 +2399,7 @@ void PencilTestPopup::refreshFrameInfo() {
       }
       // If there is already the frame then it will be overwritten
       if (hasFrame) {
-        labelStr += tr("OVERWRITE one of");
+        labelStr += tr("OVERWRITE 1 of");
         infoType = OVERWRITE;
       }
       // Or, the frame will be added to the level
@@ -2403,13 +2424,9 @@ void PencilTestPopup::refreshFrameInfo() {
               .arg(QString::fromStdWString(levelName))
               .arg(toQString(anotherPath));
       // check resolution
-      std::vector<TFrameId> fids;
-      level_sameName->getFids(fids);
-      TRasterImageP rimg =
-          (TRasterImageP)level_sameName->getSimpleLevel()->getFrame(fids[0],
-                                                                    false);
-      TDimension dim = rimg->getRaster()->getSize();
-      if (camRes != dim)
+      TDimension dim;
+      bool ret = getRasterLevelSize(level_sameName, dim);
+      if (ret && camRes != dim)
         tooltipStr += tr("\nWARNING : Image size mismatch. The size of level "
                          "with the same name is is %1 x %2.")
                           .arg(dim.lx)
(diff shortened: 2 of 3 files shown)


shun-iwasawa:
Jenkins

RodneyBaker:
That is very interesting.
I like how that informs the user and having the tooltip be updated with useful information... very nice.

The line 'Add:  4 frames exists' reads a little oddly but 'Add to 4 existing frames' is too long.
Folks should be able to figure that out though.  
I will guess that the kanji captures that more succinctly and with brevity.

I'm trying to imagine something that would flow a little smoother on the english side.  

'Add to: (4 frames)'    Perhaps not quite but that seems to be in the right direction.
Something like that would perhaps suggest the following for the other options:

'New: (empty)'
'Overwrite: (All)'
'Warning: (Error)'

The text inside the brackets would then be replaced by the number of frames (if any).

This approach does require the user to initially figure out what the number between the brackets actually means but I think they can figure it out and when in doubt the Tooltips help in that area by spelling the whole thing out.

Just a few random thoughts.
Nice addition!

shun-iwasawa:
@RodneyBaker Thank you for the advice regarding the English messages. It's helpful to me.
I modified the texts and here are variations of the info messages:

camcap_info_variations

Jenkins again

turtletooth:
@shun-iwasawa Small correction - it should be "There already is a level. . ." 
This is for the bottom two tooltips on the right.

This is a neat feature.  Thanks for doing it and I'm looking forward to testing it.

shun-iwasawa:
@turtletooth Thank you for the notification. I fixed them.

Jenkins

RodneyBaker:
Again, I really like the smart tooltips introduced with this feature.
I don't even know what such a thing might be called in the programming world because I can't remember ever seeing such a thing.  It compares to the usefulness of Right Clicking which always provides context sensitive options related to what the user needs at that specific time.  This information revealed to the user in real time (just-in-time to be useful) is yet another aspect of OpenToonz that sets it ahead of other programs in little ways.

Perhaps someday I'll be able to contribute to OpenToonz programming and scrub through the other various tooltips with an eye toward seeing of they too might benefit from tooltips that provide deeper information to the user.  

-----------

In other thoughts...
(Disclaimer:  I am not an expert in use of the English language)
Read the following only as a means of taking a break from the pressures of software development as it covers territory that is largely irrelevant but perhaps educational never-the-less.

@turtletooth's spotting of the English nuance of "a" versus "the" is something I experienced while in Japan.  The leading modifier doesn't have a direct equivalence in Japanese but comes close to the use of modifiers such as 'koko' (here), 'soko' (there) and 'asoko' (way over there) or more appropriately 'kono' (this one close by), 'sono' (that one a little further away) and ano' (that one way over there) although Japanese takes it even deeper still with 'dono' which (I think) departs further from English equivalence in that it suggests enumeration (3 or more?) but seems to me to also imply a question (i.e. it is important for us to consider how many things are under consideration. So having said that, please see the very end of this post where I return to thoughts on counting and enumeration).  

In the English language I love how an emphasis can be placed on something (or the thing under consideration de-emphasized) through use of 'a', 'an' and 'the'; as in "The King has returned.", signifying a uniquely qualified person not only by the capital letter applied to the title but by the preceding modifier 'the' which signifies 'the only one', 'this particular one' or 'the specific one we have under consideration'.  This is unlike the modifier 'a' which would suggest there is more than one of this same type and it is not of any particularly greater significance than another.  "A king has returned." reads a bit oddly because it suggests there are others.  The only reason the capital 'a'  exists is to suggest the beginning of the idea/sentence.    To say "An King has returned." would of course be incorrect and a mismatch for two reasons , firstly because the context itself suggests there is only one king under consideration but also because of convention that would suggest a consonant (the letter 'n') appear between two vowels that lack such a connection.  Ex:  "A warrior delivered his message." versus "An angel delivered his message".  Note that the only reason in this instance to change the modifier is to bridge that gap between a word with or without a consonant.  They both suggest there are more than one and there is no elaboration or special significance of that messenger under consideration.   It is 'the message' that receives emphasis so we might anticipate (expect to see) more information about that particular message in the following sentences.  We should not expect the warrior or angel to receive more emphasis unless the storyteller changes the subject of our interest by calling attention to it and emphasizing greater detail.

I add the proceeding as an illustration of where words and imagery begin to meld together.  To me words and images are but different views on the same thing.  This has long been a fascination for me.  As a young kid I even created a series of images in book form titled, "Iconoci:  woids and pitchas" to try to capture the idea of words and images working together seamlessly.  But I digress... 

---------------

Slightly more relevant to this PR...  
(Thanks for sticking with me if you are still reading!)

In the second box on the left where the field relates, "OVERWRITE one of 4 frames" the tooltip gets it right by stating 1 - 4.  While the convention is to write out numbers under ten would suggest we should spell out the number one in a sentence when itemizing or quantifying numbers it is quite appropriate to use the number itself, as in:  "OVERWRITE 1 of 4 frames".  Many (most?) native English speakers get this wrong so we are in good company here but consider that if we spell out the number one... we should also spell out the number four as in, "OVERWRITE one of four frames" (as four is less than ten).  However, because we are enumerating it is appropriate to use the numbers instead.  It is surely convenient as well because the number 4 here is surely an integer and should have to be converted to a string in order to be translated for use in a sentence.  Regardless, the current wording works but "OVERWRITE 1 of 4 frames" would appear to be best.

turtletooth:
@shun-iwasawa Thank you for doing this.  I love the way that it consistently places the new frames at the bottom of the correct level even when I try and mess it up.  I did manage to crash it by taking a few pictures, creating a vector level with a drawing and trying to set the current camera capture level to the vector level.  

It seems to catch on penciltestpopup.cpp line 2409.

shun-iwasawa:
@turtletooth Good catch. Thank you for finding the crash. Fixed.
Also changed the message to "OVERWRITE 1 of ..." according to @RodneyBaker 's advise. Thanks!

Jenkins again

turtletooth:
@shun-iwasawa I don't know if it is just my build, but the error window looks like this:
critical
And OpenToonz locks up.

shun-iwasawa:
@turtletooth In my environment it seems to be OK... Is there a way to reproduce the problem?

In the Camera Capture feature, it seems that the "Critical" popup appears only when creating the sub-folder is failed. It is a result of catching some error in `TSystem::mkDir()`, so IMO the crash might be not caused by the camera capture feature itself, but by some another reason related to the file system.

turtletooth:
@shun-iwasawa I got the same error on my work computer.  Here are steps to reproduce:
Before camera capture:
Create a new vector layer with the default name
Don't save.
Open camera capture
Set the level name to the name of the new vector level
Hit the capture button.
Critical error with white dialog.

critical_at_work

shun-iwasawa:
@turtletooth Thank you for the information.
I followed the procedure but unfortunately could not reproduce the crash...
Please see [the movie](https://drive.google.com/open?id=0BxLlAdFKGRXTOVR3cGE3Z0RIVDA) , and let me know if I overlook something.
Does the debug mode of MSVC provide any info regarding the crash?

turtletooth:
Thank for the video @shun-iwasawa You followed the steps correctly.  I don't get a crash, so there is nothing happening in debug.  The critical error window is just white.  I'm going to try a couple of other things on my side to test.

On a different topic, could you please let me know the time frame regarding my other PRs.  I currently have 30 open PRs - a couple of which are a year old.  I understand that you are busy and you have done some really incredible things with your PRs.  However, it is difficult to make PRs that may sit there for months.  I know I'm not the only one who shares the concern and I don't want to lose developers who may feel like their contributions aren't looked at.

shun-iwasawa:
Thanks @turtletooth , can you please also check if the problem occurs in the current master branch, then it can be separated as another issue?

> On a different topic, could you please let me know the time frame regarding my other PRs.

I terribly sorry for keep you and other developers waiting for PRs being merged. 
It depends on how large the modification is, but I'm trying to review 2-3 PRs per a day.
For now, I'd rather pick up PRs which are bug-fixes or pure-enhancements which will obviously have not a bad effect to the current workflow. Please let me know if there are items which you would like to be reviewed with higher priority.

One thing I would like you to try - can you restart the timed out Travis CI job ?

turtletooth:
Here's what I tried:  I tried with a clean stuff folder in case somehow it was a qss making the error not display.  That wasn't it.  

I changed my build from debug to release and that worked.  I'm not sure why it worked, but it did.  With that, I'm content merging this PR.  I still am baffled why my debug build did that.

As far as PRs go, I made some notes that might be helpful for tackling the easiest or most critical PRs:
PRs not affecting any code:
- #1301 Just an instruction change – test by @damies13
- #1409 Just a markdown file

PRs that have already been reviewed:
- #1256 I made the changes requested.
- #1286 I fixed the error you found.
- #1288 I made the requested changes
- #1352 I’ve made the requested changes
- #1381 Very helpful for me, and I’ve made the requested changes.

PRs that fix broken functionality:
- #1302 Critical for mac users, small code change, currently only affects mac sound, but could be useful to add to other sound.
- #1425 Fixes some broken functionality on the xsheet toolbar
- #1429 one word change to fix a critical function

PRs with easy reviews:
- #1351 Is a very simple change, but I’d like your thoughts on the change
- #1353 Pretty small code, doesn’t change any other functionality
- #1355 Very small fix

shun-iwasawa:
@turtletooth Thanks! It's very helpful.

turtletooth:
> One thing I would like you to try - can you restart the timed out Travis CI job ?

How do I do that?

shun-iwasawa:
@turtletooth 

- Click "Details"

image

- You will need to sign in with your github account
- Click the failed job 

image

- Check the log and confirm that the job failed on time out
- Click "Restart job"

image

Issue #1036, opened 2017-02-12
(WIP) Grouping/Ungrouping problem with Merge Levels


beeheemooth:
Hello! I'm new in Opentoonz. I need your advice. Please watch my little video to understand the problem. Look to desription. It's about ungrouping. I would be grateful to hear your opinion in solving this problem! :)

https://www.youtube.com/watch?v=LhT1WH8_8zI

mcjadski:
For some reason your video doesn't seem to work.

artisteacher:
Are you sure you are trying to fill closed shapes? You might want to enable gap check to see where the gaps are  (shift-c). Sometimes a gap is not easily noticeable. Then you can use the tape tool to fill any gaps.

beeheemooth:
I use only two layers: one doing Merge Levels (to fill the drawing) and the other to morph lines. Each new layer after morphing I combine with the first. Perhaps in this case formed groups within groups. Grouped line is impossible to fill, even if they intersect and form a closed shape. Is there a way to do Merge Levels without grouping? How do I ungroup groups within groups quickly and easily? 
(Thanks for answers, and sorry for my English)

artisteacher:
Using your workflow, I've been able to fill merged levels successfully, but most often after using the tape tool to join gaps together. However, the fill sometimes covers the other lines/areas - due to multiple groups placed in the wrong order. I have run into some cells not filling as expected (#1007), and by erasing the overlapping points and using the tape tool to reconnect, have been able to fill them.

In regards to quickly ungrouping, or merging without groups, I don't have any suggestions. I don't know if there is an easy answer there. Good luck!

beeheemooth:
It's easy to create an animation in OpenToonz this way, but it is very difficult to fill every frame! I wanted clean the drawings animation, and fill in another program, but "Remove Vector Overflow" does not work with groups.

Within half an hour of work, I ungroup just one frame (112 drawings). After ungrouping, I easily painted over 90% closed shapes without the use of "Tape Tool". And the "Remove Vector Overflow" worked well this time.

As a result, I spent about an hour on a single frame. 

Drawings animation with "Merge Levels" (7 frames) I've done about 10 -15 minutes.

beeheemooth:
Excuse my joke, please, but I really need this thing :)) Peace! V.
![ungroups](https://cloud.githubusercontent.com/assets/25724133/23085351/f17410aa-f56f-11e6-9555-fa930533cdf7.jpg)

[...]

PR #1410, opened 2017-08-18
Two New Vector Conversion Commands

turtletooth:
This is a time saving PR.  Right now - to convert a vector cell to a toonz raster cell you have two choices:

1:
- Select all the strokes using the select tool
- Copy the strokes
- Create a new Toonz Raster level
- Select the new cell and paste the strokes

2:
- Export the level as a TLV using the export level dialog
- Import the new level into the scene

This will (with one click):
- Take a selection of vector cells
- Create a new Toonz Raster level
- Copy the drawing info to the cells
- Expose the new level in the xsheet with the same order and timing as the vector level.

The other command will do almost the same thing as above, but instead of exposing the new level, it will convert the Toonz Raster back into vector and replace the original images with the simplified vectors.


beeheemooth:
It's great! But this solution is only half the problem. A reverse command is also needed: from the raster to the vector. Together, these two operations are the only way to ungroup vector levels at this time. #1036

turtletooth:
@beeheemooth That is actually why I'm building this.  This is the first half, but I am also going to build a command that converts to Toonz Raster and back to vectors with one click.  Going back to vectors will perform the same thing as the "Convert to Vectors" popup, but bypassing the popup.

artisteacher:
It will be great having a single command to convert back and forth. There is currently a 3rd way to convert from a vector level to toonz raster - selecting & copying frames from the vector level in the Level Strip and then selecting and pasting into the Toonz raster level in the level strip.

turtletooth:
@artisteacher Thanks!  I didn't know that.

turtletooth:
Added the command to convert to toonz raster and back again.  Right now the command name is "Replace with simplified vectors" - Does that make sense?

To Do:

- [ ] Add commands to menu items
- [ ] Add option in preferences to keep fill when replacing with simplified vectors
- [ ] Raise Toonz Raster image dpi on replace to get greater fidelity on converting back.

RodneyBaker:
> "Replace with simplified vectors" - Does that make sense?

"Replace with simplifed vectors" almost sounds like "Simply Vectors" in the sense of taking a vector image and optimizing it so that it contains fewer control points and yet conforms to the same shape.

Is there a reason not to use "Convert to Vectors"?
Another simple option might be "Raster to Vector" and "Vector to Raster".
Or if you are desiring only one menu item it could be "Raster <> Vector"
Although I'm not sure if those symbols are allowed in menus.

Edit:  In re-reading I may have missed that your considering the working command in the code and not text for the menu.   Hmmm...   Perhaps "raster or vector"?  "rasters and vectors"?  Now you've got me curious what the process might actually be called.  One one from raster to vector would usually be 'vectorization'.  I like the term 'conversion' as it applies to the process itself.  "raster vector convert"?  Getting closer (perhaps).


Whatever the command name... this is going to be a very welcome addition.
Thanks for adding it!

DemonRem:
בתאריך 20 באוג׳ 2017 08:58,‏ "Jeremy Bullock" <notifications@github.com>
כתב:

turtletooth:
@shun-iwasawa I don't know if you have time to answer this, but I have a little question.  When I use the convert to simplified vector command above, then undo, then redo, and undo again.  (I think it's that many)  It crashes.  I created new clones of the images for the purpose of undo and redo, but does the original image get destroyed after setImage?  These commands work, but undo has always been the part that is hardest to figure out for me.

turtletooth:
@shun-iwasawa I don't know if you would have time to look at it, but I still can't figure out how to best implement undo for the convertVectortoVector (Convert to Simplified Vector) command.

shun-iwasawa:
I'll look into this

shun-iwasawa:
@turtletooth You should use the smart pointer `TImageP` instead of `TImage*` in order to avoid crash.

```
@ -2327,14 +2327,14 @@ class VectorToVectorUndo final : public TUndo {
   TXshSimpleLevelP m_sl;
   std::vector<TFrameId> m_frameIds;
   // std::vector<TFrameId> m_newFrameIds;
-  std::vector<TImage *> m_oldImages;
-  std::vector<TImage *> m_newImages;
+  std::vector<TImageP> m_oldImages;
+  std::vector<TImageP> m_newImages;
   HookSet *m_oldLevelHooks;
 
 public:
   VectorToVectorUndo(TXshSimpleLevelP sl, std::vector<TFrameId> frameIds,
-                     std::vector<TImage *> oldImages,
-                     std::vector<TImage *> newImages, HookSet *oldLevelHooks)
+                     std::vector<TImageP> oldImages,
+                     std::vector<TImageP> newImages, HookSet *oldLevelHooks)
       : m_sl(sl)
       , m_frameIds(frameIds)
       , m_oldImages(oldImages)
@@ -2553,7 +2553,7 @@ void TCellSelection::convertVectortoVector() {
   TFrameId frameId = firstCell.getFrameId();
   std::set<TFrameId> frameIdsSet;
   std::vector<TFrameId> frameIds;
-  std::vector<TImage *> oldImages;
+  std::vector<TImageP> oldImages;
   frameIds.push_back(frameId);
   oldImages.push_back(firstCell.getImage(false).getPointer()->cloneImage());
   for (i = r0 + 1; i <= r1; i++) {
@@ -2563,7 +2563,7 @@ void TCellSelection::convertVectortoVector() {
       if (std::find(frameIds.begin(), frameIds.end(), newCell.getFrameId()) ==
           frameIds.end()) {
         frameIds.push_back(newFrameId);
-        oldImages.push_back(newCell.getImage(false).getPointer()->cloneImage());
+        oldImages.push_back(newCell.getImage(false)->cloneImage());
       }
     }
   }
@@ -2612,10 +2612,10 @@ void TCellSelection::convertVectortoVector() {
                                    // question about the palette
 
   // get clones of the new images for undo
-  std::vector<TImage *> newImages;
+  std::vector<TImageP> newImages;
   for (i = 0; i < totalImages; i++) {
     newImages.push_back(
-        sourceSl->getFrame(frameIds[i], false).getPointer()->cloneImage());
+        sourceSl->getFrame(frameIds[i], false)->cloneImage());
   }
 
   HookSet *oldLevelHooks = new HookSet();
```

You are calling `TXshSimpleLevel::setFrame()` in `VectorToVectorUndo::undo()` and `VectorToVectorUndo::redo()` . `TXshSimpleLevel::setFrame()` will overwrite the image cache and release the registration of old image accordingly. Since the image cache is using the smart pointer, releasing the registration will release the reference to it and will automatically delete the instance if the reference count will become 0. Using `TImageP` instead of `TImage*` in `VectorToVectorUndo` will add reference count to the smart pointer, so that the instance will not be automatically deleted before `VectorToVectorUndo` 's destruction.

turtletooth:
Thank you so much @shun-iwasawa - you are awesome!

toonz/sources/toonz/cellselection.cpp
```diff
@@ -2303,3 +2313,478 @@ void TCellSelection::renameMultiCells(QList<TXshCell> &cells) {
...
+  }
+  return true;
+}
+
+TXshSimpleLevel *TCellSelection::getNewToonzRasterLevel(
+    TXshSimpleLevel *sourceSl) {
+  ToonzScene *scene       = TApp::instance()->getCurrentScene()->getScene();
+  TFilePath sourcePath    = sourceSl->getPath();
+  std::wstring sourceName = sourcePath.getWideName();
+  TFilePath parentDir("+drawings");
```
shun-iwasawa: For now it seems that a new Toonz Raster Level is created always under `+drawings` . 
Instead, how about making the new level under the same folder as the source .pli file belongs to?

toonz/sources/toonz/cellselection.cpp
```diff
@@ -2303,3 +2313,478 @@ void TCellSelection::renameMultiCells(QList<TXshCell> &cells) {
...
+  std::wstring newName = sourceName;
+  while (TSystem::doesExistFileOrLevel(actualFp)) {
+    newName = sourceName + QString::number(i).toStdWString();
+    fp      = scene->getDefaultLevelPath(TZP_XSHLEVEL, newName)
+             .withParentDir(parentDir);
+    actualFp = scene->decodeFilePath(fp);
+    i++;
+  }
+  parentDir = scene->decodeFilePath(parentDir);
+  if (!TFileStatus(parentDir).doesExist()) {
```
shun-iwasawa: Here you are checking the existence of the parent directory.
I think this operation is not needed since the `parentDir` is always `+drawings` for now.
Even if you will change `parentDir` to any other non-existent path, it will surely be created on saving the level.

toonz/sources/toonz/cellselection.cpp
```diff
@@ -2303,3 +2313,478 @@ void TCellSelection::renameMultiCells(QList<TXshCell> &cells) {
...
+    } catch (...) {
+      DVGui::error(QObject::tr("Unable to create") + toQString(parentDir));
+      return NULL;
+    }
+  }
+
+  TXshLevel *level =
+      scene->createNewLevel(TZP_XSHLEVEL, newName, TDimension(), 0, fp);
+  TXshSimpleLevel *sl = dynamic_cast<TXshSimpleLevel *>(level);
+  sl->setPath(fp, true);
```
shun-iwasawa: I think this line 2479 is not needed as ‘setPath()‘ is called in `ToonzScene::createNewLevel()` above.

toonz/sources/toonz/cellselection.cpp
```diff
@@ -2303,3 +2313,478 @@ void TCellSelection::renameMultiCells(QList<TXshCell> &cells) {
...
+  TApp::instance()->getCurrentColumn()->setColumnIndex(col);
+  xsh->insertColumn(col);
+
+  CreateLevelUndo *undo = new CreateLevelUndo(row, col, totalImages, 1, true);
+  TUndoManager::manager()->add(undo);
+
+  // expose the new frames in the column
+  for (i = 0; i < totalImages; i++) {
+    for (int k = r0; k <= r1; k++) {
+      TXshCell oldCell = xsh->getCell(k, col - 1);
```
shun-iwasawa: Please modify this line like as follows:
`TXshCell oldCell = xsh->getCell(k, c0);`

Because the cell selection is not always the same as the current column. For instance, you can move the current column by pressing right/left arrow key on the XSheet keeping the cell selection unchanged.

turtletooth:
@shun-iwasawa I think I took care of the changes you asked for.  Thanks for the review.

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -1283,6 +1295,10 @@ PreferencesPopup::PreferencesPopup()
       new CheckBox(tr("Minimize Savebox after Editing"), this);
   m_useNumpadForSwitchingStyles =
       new CheckBox(tr("Use Numpad and Tab keys for Switching Styles"), this);
+  m_keepFillOnVectorSimplifyCB = new CheckBox(
+      tr("Keep fill when using \"Replace Vectors\" command"), this);
+  m_useHigherDpiOnVectorSimplifyCB = new CheckBox(
+      tr("Use higher DPI for calculations - Slower but more accurate"), this);
```
shun-iwasawa: @turtletooth 
Sorry for the late notice, but how about opening some dialog each time you trigger the `Replace Vectors with Simplified Vectors` command, containing these options instead of put them in the preferences popup?
Since the preferences popup has already many items and the `Replace Vectors with Simplified Vectors` command seems not to be used so quickly as it cannot have the dialog before execution.

shun-iwasawa: > However, I actually use the commands often (there are still often fill issues) and for workflow, being able to minimize clicks is nice.

OK, I didn't imagine that. Then putting these options in the preferences is reasonable.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @turtletooth for adding the useful commands!

beeheemooth:
It works! Hooray! Thanks so much! But these functions are not present in the specified place (Levels)

shun-iwasawa:
@beeheemooth 
It seems that the old version of menubar settings remains in your personal folder. Try removing a folder `OpenToonz 1.1 stuff\profiles\layouts\personal\Default.username` and launch the software again.

beeheemooth:
@shun-iwasawa It works, thanks!

Issue #102, opened 2016-03-29
Plastic Tool Mesh: Ability to Parent Things to Mesh Bones (Vertex)


blurymind:
Right now you can create a mesh over a drawing that is drives the drawing's deformation via bones.

That mesh can be made a child to other things - as it is exposed to the schematic view.
But you can not make anything a child to any of the bones, since they can not be exposed to the schematic view.

As an example lets pretend that you have a character whose torso is driven by the platic tool mesh. You want to attach some arms to that character, to do that you need to make them a child of a specific bone (chest). You cant.

blurymind:
Here is an image to illustrate it:
![meshboneparentexternalobject](https://cloud.githubusercontent.com/assets/6495061/14119745/7e910166-f5e7-11e5-9406-6cfefa029944.PNG)

I want to be able to attach things to mesh bones

daniele-pini:
I remember the original Toonz having several ports on one side of a schematic node with letters and numbers to choose from.

Port letters were bound to pegbar holes (everyone always only used the center one 'B').

Numbers were introduced to bind hooks and plastic skeleton vertexes (the numbers between parenthesis), so you could parent skeleton vertexes of a mesh column to other columns.

Now all I see are a blue and red dot per node, with no other indication. It's probably a feature lost in the open sourcing.


As a workaround, you could try merging the two drawings in one and remesh. The apple should result in a separate mesh component in the same mesh drawing - then you can use a single skeleton to animate both components.

digitallysane:
Yes, hooks are already doing what you need, but from the Harlequin manual it seems indeed that Harlequin's UI is more useful for this.

In the Harlequin manual: http://www.toonz.com/cgi-shl/download/71H/Toonz%20Harlequin%2071%20User%20Guide.pdf
Page 271 (288 of the PDF)

blurymind:
Can we then please get this feature back? I can think of many scenarios
where the workarounds will fail or be a pain in the neck to use!

blurymind:
I think we have enough data to call this an Enhancement feature

digitallysane:
Interesting, if you download the Paperless Tutorial files
http://www.toonz.com/htm/support/suppw.htm
And open the scenes in OT and examine the schematic, you'll see that the
multiple hook connections ARE there.
So at least it can understand them in existing Toonz scenes.
Wondering if there's some shortcut to create them?

blurymind:
[...]

PR #1413, opened 2017-08-18
Replace  'zanzo' with 'Afterimage' in current.txt

RodneyBaker:
LIne 1043, 1044 and 1130 of current.txt updated with translation of term 'zanzo' to 'afterimage' in the dialog settings of Motion Blur Ino and Motion Blur Iwa.  This is a UI update only.   Occurrences of 'zanzo' in other locations are not effected.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1040,8 +1040,8 @@
   <item>"STD_inoMotionBlurFx.y2"	"Y2"	</item>
   <item>"STD_inoMotionBlurFx.scale"	"Scale"	</item>
   <item>"STD_inoMotionBlurFx.curve"	"Curve"	</item>
-  <item>"STD_inoMotionBlurFx.zanzo_length"	"Zanzo Length"	</item>
-  <item>"STD_inoMotionBlurFx.zanzo_power"	"Zanzo Power"	</item>
+  <item>"STD_inoMotionBlurFx.zanzo_length"	"Afterimage Length"	</item>
+  <item>"STD_inoMotionBlurFx.zanzo_power"	"Afterimage Power"	</item>
   <item>"STD_inoMotionBlurFx.alpha_rendering"	"Alpha Rendering"	</item>
   <item>"STD_inoMotionWindFx"	"Motion Wind Ino"	</item>
   <item>"STD_inoMotionWindFx.direction"	"Direction"	</item>
@@ -1127,7 +1127,7 @@
   <item>"STD_iwa_MotionBlurCompFx.startCurve" "Start Curve"	</item>
   <item>"STD_iwa_MotionBlurCompFx.endValue" "End Value"	</item>
   <item>"STD_iwa_MotionBlurCompFx.endCurve" "End Curve"	</item>
-  <item>"STD_iwa_MotionBlurCompFx.zanzoMode" "Zanzo Mode"	</item>
+  <item>"STD_iwa_MotionBlurCompFx.zanzoMode" "Afterimage Mode"	</item>
   <item>"STD_iwa_MotionBlurCompFx.premultiType" "Source Premultiply Type"	</item>
 
   <item>"STD_iwa_PerspectiveDistortFx"			"Perspective Distort Iwa"</item>


RodneyBaker:
Initial PR submitted which includes another instance of term zanzo found but this time for @shun-iwasawa 's Motion Blur Iwa FX where it is a mode.  I want to emphasize that no change is made to underlying code at this time which still uses the term zenzo in a few different ways.  Only the text displayed in the FX dialog box is changed.

turtletooth:
@jpturcotte raised a concern about translation to other languages, which is valid.  However (and unfortunately) I don't think any of the text in the xml or .txt files is able to be translated on the fly like the rest of the text in the application.  Since the rest of the controls for fx are usually in English, I think it makes sense to translate zanzo to English too.  BTW- If I am wrong about these files not being able to be translated on the fly - Yay!

Quick question- I have never really played with the zanzo options.  Is "afterimage" an accurate word to describe what is being affected here?  I realize that is the zanzo translation, but is there a more commonly used English word that reflects the idea?

RodneyBaker:
> Quick question- I have never really played with the zanzo options. Is "afterimage" an accurate word to describe what is being affected here? I realize that is the zanzo translation, but is there a more commonly used English word that reflects the idea?

That is a very good question.
I would think that 'Blur' or 'Motion' might work almost as well but they don't quite capture that sense of 'image trail' that 'afterimage' suggestions.

RodneyBaker:
Errors appear to be related to Linux and Mac (environmental variables).
I need to get smart fast on what is going on here.

turtletooth:
The mac build just timed out.  Nothing to worry about - happens all the time.

RodneyBaker:
@turtletooth Ah, thanks for that.  
I had seen the same error on a few other PRs and it didn't look like there had been any effort to correct anything.  A similar error with Linux except in that case it looked like just a notification and the final PR passed.  Your explanation makes good sense.

When you find some extra time (like you have that!) can you look at this PR and advise if it is good to merge or... is that concern you raised before still a issue we need to look into?

shun-iwasawa:
Regarding this PR, please let @masafumi-inoue check it.

> This is a UI update only.

Here is one point I need to note: The fx parameter names are used as variables for expressions. Changing the parameter name will lose compatibility for some scenes in which the expressions refer it.
For instance, expression for `Zanzo Length` will be changed from `fx.motionblurino01.zanzolength` to `fx.motionblurino01.afterimagelength` . IMO it would be better if we could still load "obsolete" expression even if you changed the UI text.

adamearle:
Is there a library for these expression or code for OpenToonz on the net?


On 21 Aug 2017 12:45 PM, "shun-iwasawa" <notifications@github.com> wrote:

Regarding this PR, please let @masafumi-inoue
<https://github.com/masafumi-inoue> check it.

This is a UI update only.

Here is one point I need to note: The fx parameter names are used as
variables for expressions. Changing the parameter name will lose
compatibility for some scenes in which the expressions refer it.
For instance, expression for Zanzo Length will be changed from
fx.motionblurino01.zanzolength to fx.motionblurino01.afterimagelength . IMO
it would be better if we could still load "obsolete" expression even if you
changed the UI text.

—
<https://github.com/opentoonz/opentoonz/pull/1413#issuecomment-323634116>,
or mute the thread

shun-iwasawa:
@adamearle 

> Is there a library for these expression or code for OpenToonz on the net?

In the Toonz Harlequin Manual ([a link posted in the previous comment](https://github.com/opentoonz/opentoonz/issues/102#issuecomment-203073578)) , the expression is described in page 358. The variable texts can be referred to in the tool tips of the column headers in the function editor. Also you can drag & drop (with the MIDDLE mouse button) from the column header or the parameter tree node to the expression field to input the variables without typing.

RodneyBaker:
@shun-iwasawa said:

> In the Toonz Harlequin Manual (a link posted in the previous comment) , the expression is described in page 358. **The variable texts can be referred to in the tool tips of the column headers in the function editor. Also you can drag & drop (with the MIDDLE mouse button) from the column header or the parameter tree node to the expression field to input the variables without typing.**

Oh, wow!  That was information I had been looking for.  Thanks!  Very useful.

RodneyBaker:
@shun-iwasawa   Yes, Inouesan should be more than able to validate this.

I think we are okay here although I certainly may be wrong.
As I see it, what has been changed it the UI text only that relates to the parameter and not the parameter itself.  For instance:

"STD_inoMotionBlurFx.zanzo_length"
That would be the underlying parameter of the afterimage length.
I'm not sure what the STD designation is unless it stands for 'standard'.
If so, this would be a wise addition because it warns the developer not to alter the STD because it will break something else.   This is my hope.

I can't see logically why the UI text label would be used in an expression although it may have developed that the label is used to refer to the variable and so using the label becomes the equivalent of using the parameter... I don't know enough to speculate further.

As I say, I think... and hope... that we are okay here.

@masafumi-inoue  Onegaishimasu!

shun-iwasawa:
I asked @masafumi-inoue and Studio Ghibli, and confirmed that there will have little impact on compatibility.
IMO it would be the right moment for changing the parameter name since we will release the new version (v1.2) which will alter the default name of stuff folder as well.

Thank you @RodneyBaker for the modification. LGTM

PR #1419, opened 2017-08-21
Add Edit In Place to XSheet Toolbar and Menus

turtletooth:
Edit In Place is currently only available as a shortcut for the default rooms.  This adds the command to the xsheet toolbar and the menus.

edit_in_place_screenshot


artisteacher:
Another helpful addition would be a load level button.

turtletooth:
@artisteacher If you can get me an svg at 18 x 18, I can add it in.

artisteacher:
I've done a couple different versions. Let me know if you'd like to go in a different direction. 
basicload
I don't think I gave you svg versions of the original icons, so I've included those. I've also included some alternative versions where the color could be coded in the css at some point.
maskiconsxsheetbuttons_mask

[level_svg.zip](https://github.com/opentoonz/opentoonz/files/1244152/level_svg.zip)

PR #1420, opened 2017-08-21
Allow Shortcut to Cycle Dropdown Options

turtletooth:
This is based off of:
https://groups.google.com/forum/#!category-topic/opentoonz_en/sE-Cth_LYr0

This allows the keyboard shortcut for a dropdown to cycle the options.  This can be enabled through the preferences.  

This also creates a new category in preferences: Tools.  There didn't seem to be a category that best fit this option except for interface which is already getting crowded.  There are probably a few other preferences that could be moved to the Tools category later.


openanim:
THANKS !!! Very pleased with this addition. By the way, combining with a modifier key (or another shortcut) to go back to the previous tool could be cool too. But that would be a bonus, you did, thank you again.

shun-iwasawa:
LGTM Thank you @turtletooth for adding the useful option !

PR #1425, opened 2017-08-22
XSheetToolbar Fixes and New Pane

turtletooth:
This does a few things:
It fixes the xsheet toolbar not showing a dropdown arrow when not all icons are visible.
It also makes a new panel out of the xsheet toolbar that can be docked anywhere.  The normal xsheet toolbar is still available, the panel is a duplicate just for the flexibility of having it where you want it.

Note:
 - A side effect of this is that I had to make the new vector, new toonz raster and new raster level buttons an actual command, or they also didn't show up in the dropdown.  That means you can map them to shortcuts if you wish.
 - The keyframe commands don't show up in the dropdown because the widget they use isn't recognized as an action to add to the dropdown.

 - The new panel is called the Command Bar since XSheet Toolbar wouldn't be a good name since it can be docked anywhere.

Here is the new panel above the viewer:
anywhere

Here is the dropdown:
drop

### DIFF
--- a/stuff/config/qss/Astral_048_VeryDark/Astral_048_VeryDark.qss
+++ b/stuff/config/qss/Astral_048_VeryDark/Astral_048_VeryDark.qss
@@ -1489,7 +1489,7 @@ FxSettings QToolBar QToolBar {
    Tool Options
 ----------------------------------------------------------------------------- */
 #ToolOptions TPanelTitleBar,
-#XSheetToolBar TPanelTitleBar {
+#CommandBar TPanelTitleBar {
   background-color: #303030;
   background-image: url('../Astral_072_Dark/imgs/white/title_grip_vertical.svg');
   background-repeat: no-repeat;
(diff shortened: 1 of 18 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
The command bar will provide a quick access to specific command by touching icons, instead of typing shortcut keys. It's helpful when working with LCD tablet without keyboard. IMO it would be nice in the future if users can configure the content of it, by picking from all commands available in the software.

It seems to be fine regarding the functionality. 
Just an appearance of the command bar looks a bit strange when it is floating. Here is a comparison with the tool options bar, note that the bottom line of the command bar panel is behind the content : 
image
Can you fix this and include the modification in this PR? Or would you like this to be merged first?

turtletooth:
@shun-iwasawa I moved the new panel to its own class to both take care of the appearance issue as well as preparing it for it's own actions and customization.  Since the new level commands are accessed now accessed by two different classes, I placed the command handler for them in mainwindow.

shun-iwasawa:
Jenkins again

shun-iwasawa:
LGTM. Thank you @turtletooth for implementing this!

artisteacher:
In preferences, for consistency, "show toolbar in the xsheet" should probably be changed to "show command bar in xsheet"

PR #1429, opened 2017-08-24
Fix for arrow commands on brush tool

turtletooth:
I broke the arrows for next frame and previous frame on a previous PR.  This fixes it.  One word can change a lot.


shun-iwasawa:
Confirmed the problem occurs in the current version and fixed in this PR. Thanks @turtletooth for the fix.
No need to run Jen.kins as it is obviously no problem. LGTM

Issue #67, opened 2016-03-27
Support Multitouch Gestures 


SafariMonkey:
Many devices used by artists nowadays, such as Cintiq and Surface, have models with multitouch capabilities. Being able to use touch gestures such as pan and pinch-to-zoom without changing tool would be very useful. 

As an example, in Clip Studio, there is an option (on by default I believe) to use touch on the canvas for pan/zoom/rotate only, so effectively the fingers are used for navigation and the pen is used for tools. UI elements treat them equally.

TheoXD:
Thumbs up for this one, I was a bit disappointed when they've spoken about paperless drawing in toonz videos but do not have basic multitouch gestures, nor even an undo/redo button easily accessible.

odysseywestra:
Could you provide code examples of how we could accomplish this in QT?

relgukxilef:
There is the [MultiPointTouchArea](http://doc.qt.io/qt-5/qml-qtquick-multipointtoucharea.html) ([example code](http://doc.qt.io/qt-5/qtquick-touchinteraction-example.html)) available in QML. I don't know how useful this is for this project.

metellius:
Incoming mouseevents are in SceneViewer, which is a qwidget. Handling basic gestures with qwidget is not hard, using the gestures framework: http://doc.qt.io/qt-5/gestures-overview.html

turtletooth:
Based on the discussion in #1207 feature requests are going to be closed here on GitHub unless there is a developer actively working on the feature.  Feature requests can be discussed at the OpenToonz Google Group: https://groups.google.com/forum/#!forum/opentoonz_en

turtletooth:
Added in #1437

Issue #755, opened 2016-08-24
Android/iOS mobile port


blurymind:
I would vouch to test it and even support it's development by purchasing it on the app store.
Even as an apk that is not hosted at the android market, i would still love to have it.

Almost everyone now has a tablet or a smartphone. A lot of these devices come with a pen - if not pressure sensitive pens for them are easy and cheap to acquire. Even wacom now has android tablets sold.  Pressure sensitive or not - varied thickness can be faked via acceleration.

If anyone else vouches to test a future port or has some information, please state your device and why you would love to have OT running on it.

For me personally these are the reasons:
- the galaxy note is an affordable alternative to cintiq. Many tablets nowadays are!
- I can carry it around the house and hold it in a more comfortable pose on the sofa - as opposed to sitting in front of my laptop to draw - it's much easier to pick up and thus the incentive to casually start OT is bigger.
- I can use it while travelling on the train, bus, or generally being outside - making OT easier to show to other people and taking it out in the world - to the eyes and hands of more people. It will make it easier to show OT to others and get them involved!
- There is currently no good basic animation software for android - especially one that is open source. 
- Multitouch makes it very easy and nice to pan,rotate and zoom on the canvas naturally. Combined with a pen - it is a combo that is hard to beat by a desktop computer. It feels so natural :) 

I know that there would be some design challenges in this- but i strongly believe that it would be worth the effort - especially if the goal is to get OT in the hands of more people!
Tvpaint is an example of a port being done right -as their port is basically the same as the desktop app almost! However tvpaint on android costs 1500£ - which in my humble opinion is an insane price tag 

There is some other software on android, but it is not professional in the slightest. It's really basic.

Open toonz is a complex piece of software, but I believe that it is not impossible to port it over to a mobile platform. You might have to change some interface elements, and cut down some of the tools at first, but it can be done.
You could even put it on the app market as freemium. It will help the popularity of OT imo.

Artrage devs made the transition and it has been incredibly successful. 

So for now, until there is actual developer interest or more information on the posibility for such a port, I vouch to test any possible ports on my tablet.

If you are a tablet user, with a stylus and are interested in such a port, please state what model tablet you have and if you can test it.

manongjohn:
Wow. Funny, last night i decided to see what animation software was available for my Samsung Galaxy Tab S2. I found an ok freeware version called 'Animator'. Has onion skin and layers which is why i liked it better than others.

Would be cool to see an OT-lite/mobile version. Can help test too.

[...]

PR #1437, opened 2017-08-27
Touch Screen Controls

turtletooth:
This adds touch screen controls to the viewer.
One finger drag to pan
Pinch to zoom in/out
Two finger twist to rotate
Double tap to fit to viewer.


bobbybutterfly:
i would prefer one finger drag to pan (like in Clip Studio)
two finger pan (like in sketchbook pro) leads to accidentally drawn strokes, if one of the two fingers touches the screen first.

openanim:
@turtletooth Oh thanks, It will really help! As suggested by @bobbybutterfly , is not it possible to have 1 finger for the pan? Optional if other prefers 2 fingers. Personally, I prefer CSP approach, it's much better with a finger. In any case thank you for having realized it.

openanim:
Anyway if it works with a finger, with two fingers it should also work :)

bobbybutterfly:
@turtletooth thanks, touch Support for OT will be great.
(TV-Paint users are begging unsuccesfully for touch for years now)

one more suggestion: "snap" for horizontal Position when twist rotate (-1, 0 and 1 degree)
in Toonboom Storyboard pro 5 for example it is kind of impossible to rotate your board back to exact horizontal Position, because it has no snap support like CSP

turtletooth:
Quick question for those who have used the other apps. Once you trigger an action, does it lock to that action or can you do pan rotate and zoom all at once. This currently let's you do all at once, but it makes it hard if all you want to do is pan for example, not too also trigger one of the other actions.

turtletooth:
The reason I did two finger pan is that at certain angles using my surface pro I rest my pinky instead of my whole palm and that would trigger a pan. I'll look into making an option but I can't promise anything. I did disable finger painting with this so there shouldn't be a risk of accidentally using a tool with a finger.

bobbybutterfly:
touchuse
@turtletooth this a a screenshot of CSP; if this button is activated, you have the one finger pan, else the two finger pan.
There is a Website: surfaceproartist.com and
i know that a lot of People there complain about two finger pan in Sketchbook
(on the other site nowbody complains about one finger pan in CSP)

I just tested my paint progs with following results:
- painter 2018: pan or zoom or rotate; only one at a time
- CSP: also
- Sketchbook pro: pan and zoom and rotate; all actions together possible

i personally like the "only one at a time"-behavior a bit more. it seems more precise.

openanim:
For painter, moho we can do it. Photoshop also for zoom and rotation at the same time. CSP no, it's one action at a time. It is also the only one, for whom it is a finger for the panoramic. If you can, making this option will be a good bonus, it's really flexible to use on csp, but it's already great to be able to use tactile gestures.

openanim:
@bobbybutterfly I just saw your post, so I think they changed to painter because previously we could do all the actions zoom pan rotation at the same time, the work plan gave me the impression of a flying leaf, which moved too much in all directions.

bobbybutterfly:
@openanim or maybe its in the preferences. I am on a Surface 3 and had to choose "RTS-Realtime-Stylus" and "Windows-Multitouch" in Preferences/Tablet
are you on a Wacom?

"flying leaf" nails it. Ist just to much movement.
rotate, move, scale: one at a time, its also the standard use in 3D-Programs, for a reason

openanim:
yes, I'm on a cintiq wacom and you are probably right. I never tested it preferences. :)

turtletooth:
Alright - so I made some changes.  Here is what we have now:
One finger pan.
Two fingers for zoom or rotate, but only one at a time.
There is a slight delay to see which movement is going to win over the other at the beginning, but once a the program decides whether to zoom or rotate, controls are pretty smooth.
Double tap still fits to the viewer.

I'm hesitant to add anything to the preferences since we run the risk of preferences overload.

The only thing I couldn't figure out was snapping the rotate back to zero, since I couldn't figure out how to actually get the current rotation out of the viewer's affine.  Double tap will take care of resetting rotation for now.

artisteacher:
Will this impact touchpad performance at all? Right now 2 fingers work to zoom but a pinch motion is more standard, at least for Apple products.

RodneyBaker:
Looking forward to trying this on my touch screen/ monitor.  :)

turtletooth:
I added three finger drag for undo and redo.

blurymind:
are we getting closer to the possibility of #755  :D

shun-iwasawa:
Jenkins

shun-iwasawa:
@turtletooth 
It works fine on my MobileStudioPro.
On OSX, I tried using Magic Trackpad 2 and faced a problem: After rotating or zooming with the trackpad, the viewer seems to be disabled to show right-click menu.
I'm not sure if it is related to this PR, but please give me some time to investigate.

turtletooth:
If OSX is posing a problem, I could make this Windows only for now.

shun-iwasawa:
@turtletooth I tested in the previous version and confirmed that the similar problem happens as well. I think it is the separate issue with this modification.

shun-iwasawa:
LGTM Thank you @turtletooth for adding the nice feature!
I think I need to overhaul the tablet behavior on OSX.

artisteacher:
It looks like this is definitely causing problems in OS X with trackpads. Since only one finger is used for normal navigation on a trackpad, I keep unintentionally triggering the 1 finger pan which makes it really difficult to keep the focus where I need it to be in the viewer. The pan movement is sometimes very delayed. The two finger twist to rotate doesn't seem to be working. Also, now I can't get the hand tool to work at all.

On a positive note, I really love the double tap to fit to viewer.

RodneyBaker:
Routine review of old issues and PRs.
This PR resolves #67

Issue #1438, opened 2017-08-28
Brush Not Loading all Presets


beeheemooth:
It's_ latest nightly build. Strangely, all the presets of the brush disappeared. Creating new presets or deleting them causes a crash. 
![ezgif-2-f16cc930ea](https://user-images.githubusercontent.com/25724133/29755504-e5a04e56-8ba1-11e7-9748-37a95ec59882.gif)

konero:
I confirm this.

1. Add new preset.
2. Switch preset.
3. Delete preset.

Hangs, then crash.

turtletooth:
I probably broke it.  Sorry.  Looking into it.

turtletooth:
Yep.  I broke it.  PR submitted in #1440

beeheemooth:
Sorry for reopened that issue, but now for some reason new presets are missing after restart the program.
![ezgif-2-c8e56bc2c4](https://user-images.githubusercontent.com/25724133/29795265-2dad1564-8c54-11e7-81ae-6834a9ef709e.gif)

turtletooth:
I've looked at this for a couple hours and I can't figure out what's going on.  The data is in the brush_vector.txt file, but only one preset is loading in OT.

turtletooth:
Fixed in #1447

PR #1440, opened 2017-08-28
Fix for Preset Crash

turtletooth:
Fix for #1438


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed that the problem resolved. Thanks @turtletooth for the hot-fix !
LGTM

PR #1441, opened 2017-08-28
Fix crash on clicking "Save Task List" button with non-English environment

shun-iwasawa:
To reproduce the crash: 

1. Change Language to non-English (such as Japanese), restart the software.
1. In Tasks window, add some task.
1. Click "Save" (Save Task List) button.


RodneyBaker:
Confirmed the crash.  Everything looks good here but need someone to fully confirm fix.

masafumi-inoue:
LGTM

PR #1442, opened 2017-08-28
Theme Update

konero:
**Gamma Correction**

Some areas were difficult to see depending on setup, although it can be explained to people to use a different theme, I've adjusted the gamma as best I can to prevent random complaints like this.

**Fixes & Requests**
- FpsSlider buttons now float on their own to prevent them looking broken on specific setups
- QSlider now has a disabled style
- Selected ToolButtons now display bright to make them much quicker to find
- Lock icon displays red to make it obvious something is locked, some guys who were doing some compositing asked for this to make using the edit tool easier
- FlipPlayBar now displays the same color for the slider when in preview mode after frames have been rendered instead of white, I found there was already a qproperty for this, this was done to make the marker easier to find
- Keyframe icons converted to SVG
- Some TreeLists now show an alternate background color for items
- ScrollBars redesigned
- ConsolePlayBar buttons have been greatly condensed
- Titlebar grip icons have been removed because it can be buggy depending on setup
- The background for QDialog buttons is back
- Darkest theme XSheet has been darkened
- Light theme has been redone

**Renamed**

I renamed the themes so they were easier to understand, I thought it would be good to have a theme called "Default" so both users and devs can find their way back.

- Blue
- Dark
- Default
- Light
- Warm

**Aesthetic Update**

I altered the aesthetic slightly. All I've altered is removing the borders around buttons and textfields if the themes were darker than a specific value. I took this aesthetic from other popular software such as Maya and TVPaint where they use the ultra flat look, it makes for nice arrangements and greatly reduces the visual noise coming from the UI.

I know it can seem plain looking at it on its own, I suggest reviewing with an open scene file that has a bunch of assets. I thought the same way about 3DS Max, but once you're working with it its so much nicer not to have the UI screaming for attention.

**New Theme Colors**

I've added two new theme colors which are both very low maintenance.

The first is "blue" and it was created to fit in with the UI color from DaVinci Resolve/Fusion and Clip Studio Paint which use a blue tint. The second is "warm", it was modelled from Softimage 3D, the guys who requested it say they prefer the warm color of the UI, otherwise I don't really get it, but it sure was requested a lot.

theme-color

**Backend**

I've changed the backend to support EasyLess and Less IntelliSense VSCode extensions. There is an included VSCode settings file so everyone can output the same way.

The default theme file now resides in the less folder, not the themes folder which contains alternate theme files, this was to support a glob pattern that excludes specific folders when using IntelliSense.

Compiling is made much easier, just compile the theme files directly, no longer the main file for the default theme.

I've compressed the QSS files to a single line here, this is in the settings file for EasyLess, this makes it impossible to review or modify but it makes it load faster in OpenToonz. Since this is built on LESS, the ability to edit the QSS files was not a concern to me. If this is a problem I'll output them normally.

**Download**

[qss.zip](https://github.com/opentoonz/opentoonz/files/1263539/qss.zip)


turtletooth:
These look really good @konero -  Thanks for the amazing work as usual.  To avoid any sort of potential legal issues, would you mind renaming Softimage (perhaps Warm)?  I doubt there would be any issue, but it's conceivable that someone could be upset about using a trademarked name.

openanim:
@konero, Thanks for news themes, i love this blue theme :)

After loading a theme, I have a little issue in the display.
https://1drv.ms/i/s!AspsYffPSwbdh-x2vQIqoZBLzmZabA

konero:
@turtletooth I changed the name to warm, thats a pretty good name actually.

@openanim I thought I wouldn't see that bug again, alas. I've added a fixed font size like before so hopefully that fixes it, its not as pretty though. 😬 

I updated the qss zip above with the changes.

turtletooth:
Thanks @konero

turtletooth:
LGTM.  It's a pretty huge PR, so if I missed anything, we can fix it later.  Thanks @konero for your great UI eye.  These are really good.

PR #1447, opened 2017-08-29
Fix for Broken Brush Presets

turtletooth:
This is for #1438 

This removes snap, snap sensitivity and frame range from the presets, but it's probably better that way since they don't actually affect the brush stroke but rather are general tool options.

I don't know why adding them in broke it, but this should fix it.

### DIFF
--- a/toonz/sources/tnztools/brushtool.cpp
+++ b/toonz/sources/tnztools/brushtool.cpp
@@ -2391,9 +2391,6 @@ void BrushTool::loadPreset() {
       m_capStyle.setIndex(preset.m_cap);
       m_joinStyle.setIndex(preset.m_join);
       m_miterJoinLimit.setValue(preset.m_miter);
-      m_frameRange.setIndex(preset.m_frameRange);
-      m_snap.setValue(preset.m_snap);
-      m_snapSensitivity.setIndex(preset.m_snapSensitivity);
     } else {
       m_rasThickness.setValue(TDoublePairProperty::Value(
           std::max(preset.m_min, 1.0), preset.m_max));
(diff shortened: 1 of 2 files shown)


toonz/sources/tnztools/brushtool.cpp
```diff
@@ -2616,12 +2595,6 @@ void BrushData::loadData(TIStream &is) {
       is >> m_join, is.matchEndTag();
     else if (tagName == "Miter")
       is >> m_miter, is.matchEndTag();
-    else if (tagName == "Frame_Range")
-      is >> m_frameRange, is.matchEndTag();
-    else if (tagName == "Snap")
-      is >> m_snap, is.matchEndTag();
```
shun-iwasawa: Looking how another boolean values are loaded, maybe this line should be 
`is >> val, m_snap= val, is.matchEndTag();` ?

turtletooth: @shun-iwasawa You are right.  That does fix it.  Do you think that snap and frame range should be part of the preset or left as separate values?

shun-iwasawa: Well, excluding them would be somewhat reasonable as you initially mentioned.

toonz/sources/tnztools/brushtool.h
```diff
@@ -33,11 +33,14 @@ class BluredBrush;
 
 struct BrushData final : public TPersist {
   PERSIST_DECLARATION(BrushData)
-
+  // frameRange, snapSensitivity and snap are not included
+  // for some reason, including them causes presets to break
```
shun-iwasawa: Confirmed that it works fine. Then can you please remove this line?
As excluding snap and frame range is intentional for now.

turtletooth:
Changes made.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. Thank you @turtletooth for the fix !

PR #306, opened 2016-05-16
Added menu item 'Toggle Edit in Place' (with small modification)

shun-iwasawa:
This PR is for the issue #264 and is based on Banbury's modification https://github.com/opentoonz/opentoonz/pull/304

Here is Banbury's comment on his original PR:
> This readds the ability to edit Sub-XSheets in place. The menu item has to be
added to the menu of a room, before it can be used.

I just added a line to his code for notification of the change of the state in order to update the viewer.
And also I added this command to the default menubar of Xsheet Room.

Thank you very much, @Banbury!

PR #1448, opened 2017-08-30
Customize Command Bar and XSheet Toolbar

turtletooth:
This allows the XSheet Toolbar and the Command Bar to be customized with any command available in OpenToonz.  

They are separate items- the xsheet toolbar doesn't have to have the same items as the command bar.

Note:  Not all commands have icons, so some commands will show up as text for now.

customize

This is also a replacement for #1419 
It was easier to redo it than fix the merge conflicts.

Edit In Place is currently only available as a shortcut for the default rooms.  This adds the command to the xsheet toolbar, the command bar and the menus.

edit_in_place_screenshot


turtletooth:
Scope of the PR changed.  See notes above.

turtletooth:
I'll make a new OTX this week.

toonz/sources/toonz/commandbar.cpp
```diff
@@ -1,84 +1,196 @@
...
+void CommandBar::updateEditInPlaceStatus() {
+  TApp *app         = TApp::instance();
+  ToonzScene *scene = app->getCurrentScene()->getScene();
+  int ancestorCount = scene->getChildStack()->getAncestorCount();
+  if (ancestorCount == 0) {
+    m_editInPlace->setChecked(false);
+    return;
+  }
+  m_editInPlace->setChecked(scene->getChildStack()->getEditInPlace());
+}
```
shun-iwasawa: In many part of this PR you are trying to synchronize the state of `Edit In Place` property.

I'm so sorry for not coming to realize when I merged the `Toggle Edit in Place` command PR ( #306 )...,
but now I wonder why this toggle is stored in the scene data despite the fact that (as far as I can see) it seems to affect only display behavior of the viewer.

If this toggle is related only to the interface, then how about bringing this value out from `ChildStack` and making it one of the viewer toggle (just like `MI_ViewTable`, `MI_ICheck`, etc. ) by using `MainWindow::createToggle()` instead of `MainWindow::createMenuXsheetAction()` ? 
I think it will make it easier to handle this command in the command bar.

Anyway, making this command easily accessible via command bar will be helpful especially for cut-out animators. Thanks!

toonz/sources/toonz/commandbarpopup.cpp
```diff
@@ -0,0 +1,480 @@
...
+  TFilePath fp;
+  if (TFileStatus(path).isWritable())
+    fp = path;
+  else {
+    if (path.getName() == "xsheettoolbar") {
+      fp = ToonzFolder::getTemplateModuleDir() + TFilePath("xsheettoolbar.xml");
+    } else {
+      fp = ToonzFolder::getTemplateModuleDir() + TFilePath("commandbar.xml");
+    }
+  }
```
shun-iwasawa: - By default, `ToonzFolder::getTemplateModuleDir()` will return `stuff\profiles\layouts\settings` .
But for now `xsheettoolbar.xml` and `commandbar.xml` are under `stuff\profiles\layouts` so it fails to load the template properly.

- You can use `ToonzFolder::getModuleFile(TFilePath filename)` to obtain path to the setting files. This function will return a path to the personal setting (under `stuff\profiles\layouts\settings.username` ) if it exists, or return a path to the template setting if the personal one does not exist.

toonz/sources/toonz/commandbarpopup.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+  QTreeWidget::mousePressEvent(event);
+}
+
+//=============================================================================
+// CommandBarPopup
+//-----------------------------------------------------------------------------
+
+CommandBarPopup::CommandBarPopup(bool isXsheetToolbar)
```
shun-iwasawa: It seems that `CommandBarPopup` overlaps in many part with `MenuBarPopup` .
Is there a possibility to share the same code, in order to save the maintenance cost?

toonz/sources/toonz/commandbarpopup.cpp
turtletooth: Good catch. I fixed this and also moved the editinplace command to a toggle like you suggested.

toonz/sources/toonz/commandbarpopup.cpp
turtletooth: I took a look at the files and while the basics are very similar (I used MenuBarPopup to make CommandBarPopup), most of the functions have small changes that make sharing the same code difficult.  If they both inherited from a common base class, most functions would need to be overridden anyway.  I'm sure there is a way to do it, but honestly with this school year being as busy as it is, I wouldn't have time to take it on right now.

shun-iwasawa: No problem, then let's just leave it for the future issue.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @turtletooth for this improvement !

2dvision:
Is it possible to add search field in the toolbar item?

PR #1451, opened 2017-09-01
Enable shortcut command while renaming cell

shun-iwasawa:
Executing OT's command should not be interrupted by `RenameCellField` if the commands are enabled in the cell selection. Since they are merely used for cell renaming but are frequently used for xsheet editing.


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
Also made the `RenameCellField` not to cover left side the cell in order to make it enable to grab drag handle.

Both are just reverting the previous features missing in the current master.

shun-iwasawa:
Jenkins again

manongjohn:
Sorry, but I disagree with this change.

I don't understand why you would want to allow OT commands while you are renaming a field.  If I'm in the rename edit box trying to cut and paste the text while I'm renaming, I should be cutting and pasting text, NOT the cell beneath it.  I am purposefully in the renaming box for a reason, so why shouldn't standard shortcuts apply when it comes to text editing?

When I hit DELETE to clear the text while renaming, I've deleted the CELL, not the text, which in my opinion is wrong.

In my opinion, if you want to do the OT command, get out of renaming field box first.

turtletooth:
@manongjohn The issue is that there is an option to edit the cell entry without double clicking.  With that enabled, single clicking enables cell rename, but you would still need a way to trigger the OT commands even though the rename field is active.

I'm sorry I didn't think about that when I approved the change.  I never use the single click option.

I'm not sure what the answer is.

turtletooth:
To add to my thoughts before.  In general, I agree with @manongjohn and think the commands should only be available if you have the single click option enabled or aren't in the rename box.

manongjohn:
Ok, I enabled the single click option just to experience using it.  So whenever you click a cell, you automatically go into rename mode. and can't really get out of it on your own.  Have to wait for the rename box to disappear on its own before you can get to the cell.  Escape doesn't even clear the rename box as one might expect. 

I honestly believe this is a bad feature or a poorly thought of execution, however, since it is established, it guess needs to stay?

If I were to redo this, I'd suggest removing the command rename override logic and putting a button somewhere on the Xsheet that puts you in Cell Edit mode so any cell you click brings you right to rename box which would be useful if you are doing a lot of renaming/manual renumbering.  It's basically the preference setting but easily accessible. Whether you are in Cell Edit mode or not, allow escape key to clear the rename box in case you need to get back to the cell so you can perform OT commands as normal. 

FYI,  I see you took out the UNDO and REDO command check.  That was there before I made any changes to that logic.  If you are going to restore the functionality to what it was, you might want to keep that in there.

artisteacher:
Will this allow access to the context menu? I tried enabling the single click option a while back and found it frustrating trying to load levels directly in the Xsheet & etc. I did like being able to expose drawings in a level to the Xsheet easily and with a lot of control though.

I like @manongjohn's suggestion of having a button to toggle in & out of cell edit mode. The Xsheet toolbar might be an appropriate place for that.

shun-iwasawa:
@manongjohn

>I don't understand why you would want to allow OT commands while you are renaming a field. 

I understand your opinion that it is very straightforward that the input field accepts standard editing shortcuts when it has key focus.
Why I initially implemented and reverted this feature - enabling the OT command available in renaming field was because of requests from Japanese animation studios.

In Japanese animation studios, it is still very common to draw animation on papers and write xsheet on paper as well. OpenToonz is used (commonly, for now) for pencil testing, ink&painting or digital compositing workflows. In such workflows, usually the xsheet is already written on paper in upper process.
In such case, inputting OT's xsheet is rather "copying" from paper rather than "building" from scratch. And for such occasions, we attach great importance to "efficiency" as they need to work on tons of scenes per a day.
The option for enabling inputting the cell without double-clicking (as @turtletooth pointed) is one of the measures to achieve speed-up of the xsheet input. Furthermore, it was added also to keep consistency with Celsys's QuickChecker which is de-facto standard for pencil testing in Japan.
I enabled OT's command in the renaming field for the same reason - to gain efficiency of inputting xsheet. Unlike regular input fields, the renaming field is only allowed to type "level name" and "number". IMO it requires less try-and-error for which the standard editing shortcuts are needed.

Anyway I realize that there may be some demands for enabling standard editing commands on the field. I'll make this feature optional in preferences. Thanks!

shun-iwasawa:
@manongjohn 
> I see you took out the UNDO and REDO command check. 

I included them in `TCellSelection::isEnabledCommand()` .

shun-iwasawa:
Made the feature optional. It is OFF by default so the OT's commands will not override the standard editing commands for now.

Jenkins again

gab3d:
@shun-iwasawa perhaps we could have a special key to enable OT command
listening while in a input field (say TAB for example).
That way, if you need to issue an OT command you first hit the TAB key and
afterwards the intended OT shortcut.

El sept. 4, 2017 2:17 AM, "shun-iwasawa" <notifications@github.com>
escribió:

Made the feature optional. It is OFF by default so the OT's commands will
not override the standard editing commands for now.

Jenkins again

—
<https://github.com/opentoonz/opentoonz/pull/1451#issuecomment-326868044>,
or mute the thread

shun-iwasawa:
@gab3d

> if you need to issue an OT command you first hit the TAB key and afterwards the intended OT shortcut.

Thank you for the idea. However, I think the Tab-key solution cannot be the replacement of this option.

It is alway a tough decision to balance versatility+flexibility VS specialty+efficiency.

But again, in Japanese animation industry, we place great importance on efficiency (especially for this feature). They prefer to skip even single click or even 5cm of mouse move if such operations are skippable by setting some preference option once.


Well, I still do not exactly understand the importance of the standard editing commands for this field. AFAIK, I (and I had imagined, everybody) type only "numbers" into this field in 95% of all cases. The rest 5% are together with "level name", which is usually not so cumbersome. There will never be a long text or file path for which you may prefer to copy & pasting rather than typing from scratch. Furthermore, IMO the `Copy Cell` & `Paste Cell` commands can be used virtually the same way as Copy & Paste text from another cell.

Anyway, I think further enhancement can be done in the future in another PR. For this PR, I just would like to revert the functionality.

konero:
I agree with @shun-iwasawa, I'm in favor of making this PR the default experience for the sake of posterity because won't users expect this to be the natural way to work?

It hasn't been explained why generic key commands is more important, they're appear useless IMHO.

manongjohn:
I don't think generic key commands is more important in the grand scheme of things, but its not entirely natural for new users to see a cell edit box and not have standard text edit commands working.

On few occasions I instinctively have tried to copy the text in the cell and paste outside of OT or as column name and got TCellData instead. This looks like a bug to me. I wasn't the only one since someone else mentioned it on forum.  Other cases, I hit DEL key to delete a char or number and wound up deleting my cell entirely. Is it useless? No.

I understand and appreciate the need for this functionality now that @shun-iwasawa explained it. Restoring the functionality is prudent for this reason. I just disagree with overall implementation at this point.

I think if you were going to allow OT commands while in rename box, it really should only be effective in single-click mode where it makes sense since it brings up the rename box by default, regardless of your intentions. If not in single-click mode, I purposefully brought up the rename box with the intention I am editting text and would expect standard commands to work.

We should also allow ESC to clear the rename box and keep cell focus in any mode.

gab3d:
Just to clarify a bit.

1. 
When working in input fields (not in OT, but in any program) I usually use "Ctrl+a" to select all of the text (if it isn't already selected when you enter in the field) to be able to quickly replace it with what I need to. But I know in OT when you enter cell input mode, text is already selected (that's good!), so there's frankly not much use for this shortcut.

2.
Also, sometimes I use "Ctrl + left/right arrow" keys to quickly go to one end or the other of the current text, if I need to add a prefix/suffix to what's already in there. This is more problematic, as when I press arrow keys or "Ctrl+arrow" keys currently OT moves me to another cell. For sure that was not what I expected, the first time I used it, and I had to dedicate some amount of time and experimentation to discover how it really behaved with different key combinations. (I know that's more than a lot of users usually do, when using new software) :)

3.
Another thing I'd like OT to implement, regarding cell input is the use of modifiers for the Enter key. For example to be able to go upwards (instead of downwards) when you hit Shift+Enter, or to be able to go to the right (Ctrl+Enter), or even left (Ctrl+Shift+Enter). Spreadsheet software have this resolved since a long ago.

4.
And finally, it would be great to be able to get out of cell input mode when pressing the Enter of the main keyboard (opposed to the one in the Numeric Keyboard), for example. 
Maya has this functionality, and I think it'd be a very useful way to get out of cell input mode without having to click with the mouse elsewhere.

shun-iwasawa:
@manongjohn 

Thank you for clarifying. Now I understand that option for enabling standard text edit commands working will be access-friendly for new users.

> I think if you were going to allow OT commands while in rename box, it really should only be effective in single-click mode where it makes sense since it brings up the rename box by default, regardless of your intentions.

It's somewhat reasonable to unify the single-click option and the allowing OT command option, but for now I would like to keep them separated and ask for users' feedback. 

> We should also allow ESC to clear the rename box and keep cell focus in any mode.

That's nice. I added it.

@gab3d 

> 2. Also, sometimes I use "Ctrl + left/right arrow" keys to quickly go to one end or the other of the current text, if I need to add a prefix/suffix to what's already in there.

OK, I enabled this.

>3. Another thing I'd like OT to implement, regarding cell input is the use of modifiers for the Enter key.

I'm not sure how often you would like to do such operation... Anyway, this must be another issue and needed some discussion. Please post a feature request for this.
 
>4. And finally, it would be great to be able to get out of cell input mode when pressing the Enter of the main keyboard

I think this can be another feature request as well.
Does the "cell input mode" mean "enabling `Enable to Input Cells without Double Clicking` preference option" and would you like to have fast-access for toggling it ?

Jenkins again

gab3d:
@shun-iwasawa, thank you for adding it! :)

Regarding point 4, I mean in any input mode (single or double click) when you press Enter the cursor goes to the following cell, and there's no way to escape this mode. It'd be great to have any sort of key that allowed for escaping the cell input mode. Being Esc, main keybord Enter or whatever key that sounds reasonabilly logical to a user. :)

Thanks again!

shun-iwasawa:
@masafumi-inoue I think for now it satisfies all cases of usage. Can you please review this when you have the time?

shun-iwasawa:
Conflict resolved. I think we can ignore Travis OSX timeout if Jenkins succeed in OSX build.

masafumi-inoue:
Ignore Travis OSX timeout... and Do merge!

PR #1453, opened 2017-09-02
Use New Lock Icon for Palette

konero:
Uses the lock icon from the theme for the palette window. Forgot to add this in previous PR.


shun-iwasawa:
LGTM. Thank you @konero !

PR #1457, opened 2017-09-04
Fix Env File Path Setup

shun-iwasawa:
You can set arbitrary path to  `$TOONZPROFILES` other than `$TOONZROOT\profiles` so that managing the user profiles can be flexible.
For now, this feature is broken regarding .env file because the path to the .env file ( `m_envFile` ) is not updated after setting `m_systemVarPrefix` to proper value. This PR will fix such problem.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1456, opened 2017-09-04
Fix crash on using preference template

shun-iwasawa:
This PR will fix crash when using template preference file store in `$TOONZPROFILES/layouts/settings/` .
Please do the following to reproduce the crash:

1. Create a folder `$TOONZPROFILES/layouts/settings/` and copy `$TOONZPROFILES/layouts/settings.username/preferences.ini` as a template preference file.

1. Remove or rename `$TOONZPROFILES/layouts/settings.username/` to pretend the situation that a new user lauch OpenToonz for the first time.

1. Lauch OpenToonz.

OpenToonz crashes for failing to copy the setting file because the parent directory is missing. Sorry for the unforced mistake in my previous PR.

PR #1460, opened 2017-09-05
Fix for initial preference saving issues

manongjohn:
This PR fixes the following scenarios

1) Initial/default preferences for user not set correctly in preferences.ini file.

In this scenario, the C:\OpenToonz 1.1 stuff\profiles\layouts\setting directory is missing or the preferences.ini is missing AND the C:\OpenToonz 1.1 stuff\profiles\layouts\setting.user directory is missing completely.  This scenario may be likely on a fresh install of OT.

Result: The setting.user\preferences.ini is created, but does not contain the default information found in setting\preferences.ini

Fix: Performed a QSettings sync() which forces the data to be written out.  Afterwards, it can be copied to settings.user and reloaded properly.

NOTE: This issue is probably not really noticed since the next time they open OT, the defaults get loaded again and saved properly in the setting.user\preferences.ini.

2) setting directory exists but setting.user does not leads to a crash

In this scenario, the C:\OpenToonz 1.1 stuff\profiles\layouts\setting\preferences.ini file exists, however the setting.user directory is missing completely.  This may be likely if OT is already loaded on the system and either another user of the machine has logged in and is using it for the first time or the user has somehow deleted their setting.user directory.

Result: When it tries to copy the file from setting to setting.user, since the directory is not there, it crashes.  If the directory is there but does not have preferences.ini it will work as normal.

Fix: Prior to copying, create the parent directory if it isn't already there.  Fix was done in copyFile() so it fixes potential crashes due to this issue.

### DIFF
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -362,6 +362,9 @@ void TSystem::copyFile(const TFilePath &dst, const TFilePath &src,
 
   if (dst == src) return;
 
+  // Create the containing folder before trying to copy or it will crash!
+  touchParentDir(dst);
+
   const QString &qDst = toQString(dst);
   if (overwrite && QFile::exists(qDst)) QFile::remove(qDst);
 
--- a/toonz/sources/toonzlib/preferences.cpp
+++ b/toonz/sources/toonzlib/preferences.cpp
@@ -327,11 +327,18 @@ Preferences::Preferences()
   TFilePath layoutDir = ToonzFolder::getMyModuleDir();
   TFilePath savePath  = layoutDir + TFilePath("preferences.ini");
 
-  // If no personal settings found, then try to load template settings
-  TFilePath loadPath = ToonzFolder::getModuleFile(TFilePath("preferences.ini"));
+  bool existingUser = true;
+
+  // If there is no personal settings file, this is the user's first time use.
+  if (!TFileStatus(savePath).doesExist())
+  {
+	  // create the user's settings directory if it isn't there already
+	  TSystem::touchParentDir(savePath);
+	  existingUser = false;
+  }
 
   m_settings.reset(new QSettings(
-      QString::fromStdWString(loadPath.getWideString()), QSettings::IniFormat));
+      QString::fromStdWString(savePath.getWideString()), QSettings::IniFormat));
 
   getValue(*m_settings, "autoExposeEnabled", m_autoExposeEnabled);
   getValue(*m_settings, "autoCreateEnabled", m_autoCreateEnabled);
@@ -607,16 +614,8 @@ Preferences::Preferences()
   getValue(*m_settings, "importPolicy", m_importPolicy);
   getValue(*m_settings, "watchFileSystemEnabled", m_watchFileSystem);
 
-  // in case there is no personal settings
-  if (savePath != loadPath) {
-    // copy the template settins to the personal one
-	// Save file to directory if it doesn't already exist.
-	if (!TFileStatus(loadPath).doesExist()) m_settings->sync();
-    TSystem::copyFile(savePath, loadPath);
-    m_settings.reset(
-        new QSettings(QString::fromStdWString(savePath.getWideString()),
-                      QSettings::IniFormat));
-  }
+  // Save settings to directory now, just in case
+  m_settings->sync();
 }
 
 //-----------------------------------------------------------------
--- a/toonz/sources/toonzlib/preferences.cpp
+++ b/toonz/sources/toonzlib/preferences.cpp
@@ -610,8 +610,11 @@ Preferences::Preferences()
   // in case there is no personal settings
   if (savePath != loadPath) {
     // copy the template settins to the personal one
-    if (TFileStatus(loadPath).doesExist())
-      TSystem::copyFile(savePath, loadPath);
+	// Save file to directory if it doesn't already exist.
+	if (!TFileStatus(loadPath).doesExist()) m_settings->sync();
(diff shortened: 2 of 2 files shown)


shun-iwasawa:
@manongjohn It's partially duplicated with #1456

manongjohn:
Ah. Didn't see that PR.

Beyond how we make the missing directory, I put my directory fix in copyFiles so where ever this function is used won't run into this issue.

If you want to keep yours, then I can just close this.  Though please add the m_settings->sync() line that I have in preferences.cpp as it's needed to create the initial file in the settings directory so it has something to copy into the settings.user directory.

Let me know how you wish to proceed.

shun-iwasawa:
@manongjohn 
Your solution adding `touchParentDir();` to `TSystem::copyFile()` seems to be better for avoiding potential crash problem related to this function in the future. Let's take this. Thanks!

However, I think that the preferences template `setting\preferences.ini` should not be created automatically by `m_settings->sync()` for reasons as follows:

- Consider that the template file is saved by `m_settings->sync()` when the first user launches the software. For the second user (sharing the same `$TOONZPROFILES` ), in most cases there is no difference between loading the preference values from the template or using hard-coded initial values as they are identical.

- If, by any chance, there will be the case that the hard-coded initial values are changed in the future release for some reasons, they cannot follow such change because they already have old and automatically-generated preferences template. I think which may cause malfunction or strange behavior for users, as they aren't saved intentionally.

manongjohn:
Here's the problem with the existing logic, assuming you don't run the sync():

So the 1st time user starts OT:

Based on the current logic, it establishes m_settings based on loadPath which is the settings\preferences.ini (template) because the savePath (settings.user\preferences.ini) doesn't exist yet.  It runs through the logic and sets up any defaults that are established. It gets to the logic at the bottom and checks to see if the template file exits.  This check **always** returns FALSE for the 1st time OT is run because the file hasn't been written out yet and therefore the copyFile() never runs. So then it resets m_settings using savePath.  The reset action inherently performs 2 things: 1) the contents of m_settings finally gets written to the template file. 2) Attempts to load the data in savePath, but that file does not exist because the copy never happened so any defaults that were set have been cleared. This is actually bad because if a default is not set properly and code using that doesn't account for not being set, OT may crash or produce odd behavior.  Somewhere else in the startup process the settings.user\preferences.ini finally gets created but if you compared it to the template you will see that they are not identical.  This could explain some crashes whose sources may not have been identified yet?

When the 1st time user restarts OT later, since there is now a valid settings.user\preference.ini, it will load that file directly instead of the template. At this point any missing defaults are loaded properly and saved under the user's preference file.

So now the 2nd user logs in:

When it runs, it will use the already established template created by the 1st user because it already exists.  At that point, if something has changed between releases, it won't matter.  The template file will never change existing defaults unless coded otherwise to force the change.  It will only add new defaults. Still, it won't apply any new defaults correctly to the 2nd user's 1st run because it will actually copy the existing template that has not been updated yet. So now, the 2nd user's 1st run is using defaults set from the 1st user's run, missing any new defaults until they restart OT later. 

Calling sync() where I have it fixes these issues.  I think to avoid using old default values, we can add logic so that after m_settings is reset to use savePath, we can delete the template file. Or perhaps, what should really happen is the settings\preferences.ini file is **moved** to settings.user\preferences.ini instead of a copy.  In either case, the effect is the same: another user will always force a new template to be created and used thereby avoiding any issues caused by old default settings.

manongjohn:
Actually, do we really need to create a template file at all? Wouldn't it just be better to write the settings.user\preference.ini file directly if it doesn't exist?

shun-iwasawa:
@manongjohn

Thank you for detailed explanation! Oh, I think I forgot some behavior of `QSettings` on its destruction.

>Actually, do we really need to create a template file at all? Wouldn't it just be better to write the settings.user\preference.ini file directly if it doesn't exist?

Indeed. We don't need to create a template if it doesn't exist. Can you please modify to do so?

manongjohn:
Done.  Tested combinations with and without existing user setting directory and file.

I added a variable to determine if it its an existing user or new user in case someone wants to know and act differently when loading preferences like forcing a preference change on existing users if we're deprecating something they might have settings for.

shun-iwasawa:
I'm terribly sorry for making you confused. But here is what I think the preferences should act:
We don't need to create a template if it doesn't exist. However, **If there is a template in `settings/preferences.ini` (which is to be created manually), new user will use such values instead of the default hard-coded values.** 
For now it seems that it totally ignores the template. Isn't it?

shun-iwasawa:
@manongjohn 

Here are my expected flow for loading preferences;

* If there is already a personal preferences (in `settings.username/` ), use such file.

* If there is not the a personal preferences ( for new user )
  * If there is a template preferences (in `settings/`), copy data from the template and create the personal one.
  * If there is not the template, create the personal one from the hard-coded default values. Do not create template. 

What do you think? If it's reluctant for you, I think I can try to implement this based on your advise so far.

shun-iwasawa:
@manongjohn I've modified my PR #1456 . Could you please take a look ? Thanks!

manongjohn:
Hi. Sorry, had gone to bed shortly after submitting the commit. 😊

Is there a particular reason we want to use a template if one exists? It could have old default values in it since it never changes. 

The preference constructor already defaults things for what we would want new users to start with. A template file could change it to something we don't want.

manongjohn:
Sorry. Had gone to bed shortly after I submitted the commit. It was 2AM 😊

Ok. I guess loading a manually created template might be usefull in a shared environment if the admin wanted to force certain settings on initial startup of a new user. I'll review your changes and close this PR.

shun-iwasawa:
Thank you @manongjohn

Issue #1388, opened 2017-08-08
OT does not refresh altered image file


bobbybutterfly:
import layered psd file into column. Delete the column.
Modify the image in Photoshop, save it.
Drag the new file into a column.
Still the old Image is shown not the new modified one.
Happends with loaded and imported files.

Only restart of OT solves the Problem; the altered Image is shown correct now

a simple reload button could help

shun-iwasawa:
For non-PSD image files, you can reload them with "Revert to Last Saved Version" command in the right-click menu of the xsheet cells containing the target levels.
Regarding the PSD image levels, this command will not be shown for now since (as far as I guess) the command cannot reproduce the loading option. 

Even the levels are removed from the xsheet, the level registration will remain in the scene cast (and the image cache remains as well). So, as a quick workaround, you can do as follows:
- After deleting the column, open the Scene Cast window.
- Right click on the cast pane and select "Remove All Unused Levels".
- Load the updated PSD file again.

turtletooth:
@shun-iwasawa To make this more clear for non-psd levels, what do you think of changing the command to "Load Saved Version" or something like that so users might get a hint that it can also be used to refresh a file.

turtletooth:
Maybe replace with saved version?

shun-iwasawa:
@turtletooth 
It's reasonable to alter the command name. If possible, I would prefer to set some name which is already used for the similar functionality in another (major) software in order to gain consistency.
Your suggestion "Replace with Saved Version" would also be fine.

bobbybutterfly:
Thanks, your workaround "remove all unused levels" works reliably.
"Replace with saved version" sounds to long for me. Simply "reload" 
MAYA for example has a "reload" button in its Attribute Editor.

Btw: i cant find "Revert to Last Saved Version" command in the right-click menu of the xsheet cells:
 
![rightclick](https://user-images.githubusercontent.com/29300971/29109976-54eb1d48-7ce5-11e7-9ad7-8d80dd88e3d2.jpg)

PR #1465, opened 2017-09-09
Rename "Revert To Last Saved Version" Command

turtletooth:
This changes the "Revert To Last Saved Version" command to Reload.
This is based on the conversation in #1388


shun-iwasawa:
LGTM Thank you @turtletooth for the modification. Also thanks @bobbybutterfly for good suggestion of the command name.

shun-iwasawa:
I think we don't need to call Jen.kins as it is obviously just a command name change.

PR #1467, opened 2017-09-10
Move AutoClose Value to TEnv

turtletooth:
This moves the inc value in tl2lautocloser.cpp to TEnv.  This doesn't actually change the value but allows the user to set it by hand.  I wouldn't recommend that most users mess with this, but I think having it in TEnv will allow for some troubleshooting.

I didn't put this in preferences for 2 reasons:
Preferences isn't accessible in this file. (Good reason)
This value isn't something people should generally mess with.

### DIFF
--- a/toonz/sources/common/tvectorimage/tl2lautocloser.cpp
+++ b/toonz/sources/common/tvectorimage/tl2lautocloser.cpp
@@ -3,11 +3,13 @@
 #include "tl2lautocloser.h"
 
 #include "tgl.h"
-
+#include "tenv.h"
 #include "tvectorimage.h"
 #include "tstroke.h"
 
 #include <QDebug>
+
+TEnv::DoubleVar VectorCloseValue("VectorCloseValue", 5);
 
 //=============================================================================
 #ifdef _WIN32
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Sorry, I don't precisely remember a role of this value. Is this value an interval between "points" in strokes, which are used to separate regions at crossing point, right?
I just afraid that making this value adjustable for each user may cause difference of the rendered results of the same scene between users.

turtletooth:
> I just afraid that making this value adjustable for each user may cause difference of the rendered results of the same scene between users.
That's why I made it basically a hidden value.  You have to dig into TEnv to adjust it.  But I would like to be able to play with the value for testing without having to build each time, and there may be a rare case where knowing where to adjust that value helps a user out.

shun-iwasawa:
> But I would like to be able to play with the value for testing without having to build each time, and there may be a rare case where knowing where to adjust that value helps a user out.

Understood, so you would like to have a sort of "maintenance hole" for this feature. That's make sense.

Jenkins

shun-iwasawa:
LGTM

PR #1468, opened 2017-09-12
Foreign letter and Apostrophe fix

### LINKED ISSUE #666: [BUG] Filenames with foreign letters like "äöü" will cause corrupted saves and crashes. - Windows
Filenames with foreign letters like "äöü" will cause corrupted saves and crashes.

### LINKED ISSUE #1226: Apostrophe in level name breaks path
You can create or load a level with an apostrophe ' in the name. After saving all, closing and reopening, the level displays as missing - red name and all. The level is actually still in the original location, whether that is the scene's drawing folder or somewhere else. Going to the level settings, it looks like OpenToonz misinterprets the name. It adds a forward slash where there should not be, resulting in a false path. No other symbols seem to have this issue. 
img_3931

manongjohn:
This PR does the following:

1) fixes #666 

Problem: Foreign characters in Level names, when written to the .tnz file, can cause the scene not to load correctly later if the entire Level name is not double-quoted.

Resolution:  Inspect the characters in the string for anything outside the ASCII range 32 (space) and 126 (~) and force the entire string to be double-quoted.

2) fixes #1226

Problem: When an apostrophe is in a filename, an escape character (\\) is inserted before it.  Later when the scene loads, it cannot find the file because the backlash is being treated as part of the path.

Resolution: Windows and OSX builds do not need to have apostrophe's escaped.  Added logic so it only escapes apostrophes for LINUX builds.

These fixes will only prevents the issue going forward.  This does not fix existing .tnz files with these issues.

**These fixes were tested on a Windows build. Need someone to verify OSX and LINUX builds.**

### DIFF
--- a/toonz/sources/common/tstream/tstream.cpp
+++ b/toonz/sources/common/tstream/tstream.cpp
@@ -24,7 +24,12 @@ namespace {
 string escape(string v) {
   int i = 0;
   for (;;) {
+// Removing escaping of apostrophe from Windows and OSX as it's not needed and causes problems
+#ifdef LINUX
     i = v.find_first_of("\\\'\"", i);
+#else
+    i = v.find_first_of("\\\"", i);
+#endif
     if (i == (int)string::npos) break;
     string h = "\\" + v[i];
     v.insert(i, "\\");
--- a/toonz/sources/common/tstream/tstream.cpp
+++ b/toonz/sources/common/tstream/tstream.cpp
@@ -348,7 +348,10 @@ TOStream &TOStream::operator<<(string v) {
   }
   int i;
   for (i = 0; i < len; i++)
-    if (!iswalnum(v[i]) && v[i] != '_' && v[i] != '%') break;
+    if ((!iswalnum(v[i]) && v[i] != '_' && v[i] != '%')
+		|| v[i] < 32   // Less than ASCII for SPACE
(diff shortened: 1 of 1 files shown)


artisteacher:
Thanks for taking care of this!

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed that this PR fixes the both problems. Thank you @manongjohn for the fix !

Also I tested the fix on OSX. In my environment it seems that the issue #666 does not occur on OSX (level name like "äöü" is already double-quoted in the current master). Anyway, it does work with no problem.

I can't verify Linux environment but it seems to be OK as far as I can see the PR.
LGTM

manongjohn:
> Also I tested the fix on OSX. In my environment it seems that the issue #666 does not occur on OSX (level name like "äöü" is already double-quoted in the current master). Anyway, it does work with no problem.

Did you have any other special character (non-alphanumeric) in the text with the foreign letter? The way the old code worked, any non-alphanumeric characters would cause the text to be quoted anyway, so it didn't matter if there was foreign letter or not.  That's the reason I discovered quoting the text would fix the issue.

I still wonder if the escaped apostrophe was really needed even for LINUX build if the text is double-quoted. If they have this problem also then we just need to remove the macro condition.

Thanks for reviewing & merging!

PR #1474, opened 2017-09-13
Add Button Focus Color and General Theme Improvements

konero:
### Issue
A friend who was testing some scenes by accident pressed return key when closing which caused the scene to `Save All` and caused a bit of trouble to his project. I assume some software he uses has focus set to Cancel which caused him the habit.

### Solution
This adds a focus color to buttons only inside a `#dialogButtonFrame`. You can switch focus by using the arrow keys a bit easier. Otherwise normal OS focus applies (the dotted border).

focus-examp2

### Note
This is purely `QSS`. This also does the following; as usual I pack these PRs with a lot of fixes.

1. Removes the `warm` theme (it needs work)
2. Light theme improvements as per requests
3. Changed scope of QToolButton to style those in `PencilTestPopup` and `XSheet` notes area
4. Reconfigured dual slider handles so they no longer 'combine'
5. Add lock icon hover color
6. Change value of FlipSlider/Marker images for `light` theme
7. Added a 'pressed' effect on scrollbar buttons
8. Tweak spacing and margin of `XSheetToolbar` buttons
9. And other small tweaks.

### Download

[qss.zip](https://github.com/opentoonz/opentoonz/files/1322154/qss.zip) (updated 21-Sep)


turtletooth:
LGTM - Thanks @konero - You do great work.

Issue #783, opened 2016-09-02
Feature : Automatic input cell number


shun-iwasawa:
This is requested from some Japanese animation production company.
Introduce a new xsheet-input feature, which is available in Retas and QuickChecker.

Here is a dialog image captured from CoreRETAS (Japanese version):
<img src="https://cloud.githubusercontent.com/assets/17974955/18200781/8c1f645a-7141-11e6-8ca0-8182216ac857.png" width=400>

1. From level frame.
1. Level frame intervals. If you type 0, all frames in the level will be automatically picked.
1. To level frame.
1. Empty cell intervals.
1. Cell steps.
1. Repeat count. If you type 0, it will be repeated to the end of the scene.

And here is examples of how it works, according to Quick Checker user's manual:
<img src="https://cloud.githubusercontent.com/assets/17974955/18201369/af8b3d30-7144-11e6-8309-0a35979e84ab.png" width=400>

The advantage of this feature is IMO inserting empty cells automatically.
For the actual implementation to OpenToonz, I think it would better to add a level selection combo box and a checkbox for selecting insert/replace cells option.

RodneyBaker:
Did #1479 implement this feature (#783)?
From what I can see it did.

If implemented this can be closed.

PR #1479, opened 2017-09-15
Auto Input Cell Number command

shun-iwasawa:
This implements #783 , which is almost the same feature available in Retas and QuickChecker.
From the specification described in #783, I've changed in the following two points:

- "Level frame intervals" is changed to "Level frames increment". And behavior is as follows:

  - If the increment is set to 1, frames will be placed with no interval (1,2,3,4... , just like the interval is set to 0).
  - If the increment is set to 2, frames will be placed every other frame (1,3,5... , just like the interval is set to 1).
  - And, if the increment is set to 0, every existing frames in the level will automatically be picked and placed. (This is new and default value. The description can be found in the tool tip of the field)

- Enabled to select the placing behavior from "Insert" and "Replace".

As this command is available when cells or columns are selected, I added this command to the right-click menu of the xhseet cell area and column-head area, as well as to the "Cells" menu in the main menu bar.

image


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

gab3d:
@shun-iwasawa, at least from my point of view, the words "frame" in the dialog from above should be better placed to the left of the number fields:
  From frame 0001
  To frame 0006

...at least in Spanish (and I think English too) it makes more sense.

shun-iwasawa:
Text fixed. Thanks @gab3d .

image

Jenkins again

masafumi-inoue:
LGTM

PR #1480, opened 2017-09-15
Toonz raster merge crash fix

### LINKED ISSUE #1238: Crash when merge more than two Toonz Raster Levels (gif)
ezgif-1-506444365e

manongjohn:
This PR fixes #1238 

This also fixes the issue with the palettes not merging correctly for Toonz Raster level merges.

### DIFF
--- a/toonz/sources/toonz/matchlinecommand.cpp
+++ b/toonz/sources/toonz/matchlinecommand.cpp
@@ -473,8 +473,8 @@ public:
 
     cloneColumn(cells, newLevelPath);
 
-    QString newLevelPathString =
-        QString::fromStdWString(newLevelPath.getWideString());
+    TFilePath tmpPath = newLevelPath.withName(
+        QString::fromStdString(newLevelPath.getName() + "_tmp").toStdWString());
 
     it = indices.begin();
     ++it;
@@ -482,8 +482,8 @@ public:
       int index = *it;
       it++;
       mergeCmapped(destColumn, index - count,
-                   it == indices.end() ? newLevelPathString
-                                       : newLevelPathString + "_tmp",
+                   it == indices.end() ? newLevelPath.getQString()
+                                       : tmpPath.getQString(),
                    false);
       ColumnCmd::deleteColumn(index - count);
       progress.setValue(++count);
--- a/toonz/sources/toonz/matchlinecommand.cpp
+++ b/toonz/sources/toonz/matchlinecommand.cpp
@@ -473,7 +473,8 @@ public:
 
     cloneColumn(cells, newLevelPath);
 
-	QString newLevelPathString = QString::fromStdWString(newLevelPath.getWideString());
+    QString newLevelPathString =
+        QString::fromStdWString(newLevelPath.getWideString());
 
     it = indices.begin();
     ++it;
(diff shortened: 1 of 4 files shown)


shun-iwasawa:
@manongjohn Confirmed that the crash problem seems to be fixed. Thank you so much!
Here are some comments on this PR:

1. It seems not to be `clang-format` ed. Please apply it.

1. I found a minor problem as follows: If you merge more than 2 levels, the result level name becomes not the same as specified. For instance, if you set the level name to "A_merged", the result level name becomes "A_merged_1". This seems to be because that the level name is acquired each time a pair of level is merged and which causes level name conflict.

This PR definitely improves the software as-is (you modified the crash problem!) So I think it is OK to merge this once `clang-format` will be applied and separate the second problem to another issue.

manongjohn:
@shun-iwasawa Thanks for reviewing and finding that problem with the final level name.

I've corrected the level name problem and ran beautification on my branch.  The beautification script found 2 additional files that aren't related to this PR and I didn't touch.  Since they are minor spacing changes so I included them  anyway.

toonz/sources/toonz/matchlinecommand.cpp
```diff
@@ -473,13 +473,15 @@ class MergeCmappedCommand final : public MenuItemHandler {
...
+	QString newLevelPathString = QString::fromStdWString(newLevelPath.getWideString());
+
     it = indices.begin();
     ++it;
     for (int count = 0; it != indices.end();) {
       int index = *it;
       it++;
       mergeCmapped(destColumn, index - count,
-				   QString::fromStdWString(newLevelPath.getWideString()),
+				   it == indices.end() ? newLevelPathString : newLevelPathString + "_tmp",
```
shun-iwasawa: It seems that you are trying to set temporary name for intermediate levels.
However, as it now stands this line adds `_tmp` at the end of the file path (i.e. after the file extension). For example, if the `newLevelPath` is `+drawings/A_merged.tlv` , temporary path becomes `+drawings/A_merged.tlv_tmp` which seems to have no effect in the level name.

How about modifying this part like as follows? :

```
ndex 49bcd65..c0a2484 100644
@@ -473,8 +473,7 @@ public:
 
     cloneColumn(cells, newLevelPath);
 
-    QString newLevelPathString =
-        QString::fromStdWString(newLevelPath.getWideString());
+    TFilePath tmpPath = newLevelPath.withName(QString::fromStdString(newLevelPath.getName() + "_tmp").toStdWString());
 
     it = indices.begin();
     ++it;
@@ -482,8 +481,8 @@ public:
       int index = *it;
       it++;
       mergeCmapped(destColumn, index - count,
-                   it == indices.end() ? newLevelPathString
-                                       : newLevelPathString + "_tmp",
+                   it == indices.end() ? newLevelPath.getQString()
+                                       : tmpPath.getQString(),
                    false);
       ColumnCmd::deleteColumn(index - count);
       progress.setValue(++count);
```

manongjohn: Although both somehow seemed to work, your change makes more sense. I've made and commited the change per your suggestion.  Thanks!

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM. Thank you @manongjohn for the fix!

shun-iwasawa:
@manongjohn
Thank you very much for your great works done so far.
We'll be happy if you could contribute OpenToonz continuously and would like to assign you as a collaborator :-)

manongjohn:
@shun-iwasawa Wow. Thanks :)

PR #1482, opened 2017-09-16
Fix crash when viewing audio levels in Scene Cast

### LINKED ISSUE #1240: Audio View Crash (gif)
ezgif-1-67d05e9bd4

manongjohn:
This PR fixes #1240 

The "View" menu item does not support Audio and Palette level types.  Disabled this option if either of those types are selected to prevent crashing.

Additionally allowed "Remove Level" and "Remove All Unused Levels" to be options when selecting Audio level types as these are still valid actions.

### DIFF
--- a/toonz/sources/toonz/castviewer.cpp
+++ b/toonz/sources/toonz/castviewer.cpp
@@ -725,6 +725,7 @@ QMenu *CastBrowser::getContextMenu(QWidget *parent, int index) {
   bool audioSelected       = false;
   bool paletteSelected     = false;
   bool vectorLevelSelected = false;
+  bool meshLevelSelected   = false;
   bool otherFileSelected   = false;
   int levelSelectedCount   = 0;
   for (it = indices.begin(); it != indices.end(); ++it) {
@@ -742,6 +743,8 @@ QMenu *CastBrowser::getContextMenu(QWidget *parent, int index) {
     levelSelectedCount++;
     if (sl->getType() == PLI_XSHLEVEL)
       vectorLevelSelected = true;
+	else if(sl->getType() == MESH_XSHLEVEL)
+	  meshLevelSelected = true;
     else
       otherFileSelected = true;
   }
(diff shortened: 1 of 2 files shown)


toonz/sources/toonz/castviewer.cpp
```diff
@@ -752,11 +752,12 @@ QMenu *CastBrowser::getContextMenu(QWidget *parent, int index) {
...
-  menu->addAction(cm->getAction(MI_ViewFile));
+  if(!audioSelected && !paletteSelected)
+	  menu->addAction(cm->getAction(MI_ViewFile));
   menu->addAction(cm->getAction(MI_FileInfo));
-  if (audioSelected && !paletteSelected && !vectorLevelSelected &&
-      !otherFileSelected)
-    return menu;
+//  if (audioSelected && !paletteSelected && !vectorLevelSelected &&
+//      !otherFileSelected)
+//    return menu;
```
turtletooth: Thanks for this.  I think these commented lines can be fully removed.

manongjohn: Done. Thanks for reviewing.

turtletooth:
@manongjohn The crash can still happen via keyboard shortcut.

adding this to levelsettingspopup.cpp seems to solve it:

`for (i = 0; i < selectedLevels.size(); i++) {
		  if (selectedLevels[i]->getType() == SND_XSHLEVEL) return;
		  simpleLevels.push_back(selectedLevels[i]->getSimpleLevel());
	  }`

manongjohn:
@turtletooth Thanks for reviewing and finding the issue with the View shortcut.

I've corrected the logic to make sure it doesn't try to work on levels it shouldn't.  Really, the only valid level types that View should work on are: Toonz Vector, Toonz Raster and standard Raster.

Please note, I also fixed the Scene Cast windows so View doesn't work on Mesh levels.  The option is not usually allowed in the Xsheet context menu for this type.  Viewing a mesh level doesn't cause an immediate crash but I was sporadically crashing every time I closed the flipbook that opened for the mesh level.

turtletooth:
LGTM.  Thanks @manongjohn

PR #1485, opened 2017-09-19
Reframe with Empty Inbetweens command

shun-iwasawa:
This PR will add "Reframe with Blanks" command, which will reframe the selected cells, inserting blank frames after each chunk of frame step.

As this feature was requested by some Japanese animation studio, specification of this feature is a little optimized for Japanese animation industry: The amount of blank frames is specified by the quotient, divided by the reframe step. For example, "3 steps - 2 blanks" will insert 6 blank cells. We call such cell order as `3コマ 中2` (3 koma - naka 2) in Japanese animation industry. 
For ease of comprehension, how many blank cells will be inserted is shown in the label.

image

This command will be added in the `Reframe` submenus of right-click menus of cell and column selections.


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
Conflicts resolved. Jenkins again

gab3d:
@shun-iwasawa,
sorry to ask, but I think I couldn't figure out how the blanks are
inserted. :)

in your example above (3 steps, 2 blanks):
A - will each selected cell be reexposed to last 3 frames, and after each
one of them there will be 6 blanks?
or
B - will each selected cell exposure be extended to last 3 times its
current value, and after each one of them there will be 6 blanks?
or
C - a different one...?

thanks!!

turtletooth:
@shun-iwasawa I understand what @gab3d is saying.  Is there a way to phrase this so that those who don't use Retas will understand what they are about to do?  I'm not sure what the final outcome will be from looking at the info.

shun-iwasawa:
@gab3d  @turtletooth 

> A - will each selected cell be reexposed to last 3 frames, and after each one of them there will be 6 blanks?
> or
> B - will each selected cell exposure be extended to last 3 times its current value, and after each one of them there will be 6 blanks? 

The behavior A seems to explain the "Reframe" command and B is for the "Step" command. (Please correct me if I'm wrong!)
And for this feature will work as the behavior A, just like other "Reframe - n's" commands as suggested by its command name. Only difference from the existing "Reframe" commands is that the blank frames in selected cells will be ignored on reexposure.

Here is the image for explaining this feature. Hope it helps.

image

turtletooth:
@shun-iwasawa Thank you for the explanation, but I think that the blanks are the most confusing part here.  Having it say 2 blanks and 6 blanks on the same window is a bit confusing.  Is there a way to reword the window?  Is having the number of blank frames as the quotient of the entries a key part of the process, or could it also work with directly entering the number of desired blank frames?

shun-iwasawa:
This feature is demanded by Japanese animation studios, mainly for the pencil testing.
In the pencil testing, key frame animator would like to adjust timing of drawings without inbetweens. And in such case, cells in where inbetweens to be placed are left blank.
So, the feature for reframing and adjusting blanks easily will be helpful for them.

@turtletooth 

> Is there a way to reword the window? Is having the number of blank frames as the quotient of the entries a key part of the process, or could it also work with directly entering the number of desired blank frames?

For above reason, they are familiar to think the amount of blank as "how many inbetweens to be inserted". (In Japanese ‘3コマ 中2' , '中' means "inbetween".) So I would like to keep this field as the quotient, and would like to find another appropriate wording for it... Do you have any idea?

turtletooth:
I get it.  Let me think about it.  Thanks for the explanation.  It didn't click until you mentioned inbetweens and then I understood that it is inserting blank inbetweens with the set steps.

turtletooth:
This is perhaps too wordy, but 'empty inbetweens' seems to describe what that number is defining.  I could also be way overthinking this.  If other people understood the idea with the original wording, I'm fine with that.

shun-iwasawa:
> This is perhaps too wordy, but 'empty inbetweens' seems to describe what that number is defining. 

Sounds good. Thank you @turtletooth

shun-iwasawa:
Changed wording. Jenkins again.

image

@gab3d  Is this more understandable than before?

gab3d:
@shun-iwasawa, absolutely!
LGTM

thanks!!

shun-iwasawa:
@turtletooth Could you please take a look at this PR?

shun-iwasawa:
@turtletooth I'm terribly sorry for fluttering, but please forgive me for asking @masafumi-inoue to review this, since some Japanese animation studio demands this feature to be included soon.
I know you are busy and realized that this feature is quite Japanese-specified.

masafumi-inoue:
LGTM

Issue #1039, opened 2017-02-14
Error Scanning


OliverGee:
We recently purhcased an Epson GT-15000 specifically for use with OpenToonz. While it works in GTS, whenever we try it in OpenToonz (or Toonz Harlequin 7.1 for that matter) we get an 'Error scanning' and a flashing red light on the scanner itself. It is listed as a supported scanner in Toonz 7.1.

OpenToonz sees the scanner in Scan Settings, and is set to use the Internal driver, but it will not scan. It is also only showing A4 paper as the option (Toonz 7.1 shows all paper options).

Any help would be greatly appreciated!

gab3d:
@shun-iwasawa 
I am also considering buying a GT-1500 for using with OT (or GTS), any advice you could give me on the subject would be deeply appreciated.

on another (but related) matter I tried using last GTS version with a Epson Lide 110 (twain) scanner, and when I touch the crop regions controls on screen to move them, the region measure values jump to half their supposed values, and never go back, so I cannot use cropping at all.
...I don't know where to report that issue either  :-/

thanks in advance!!

Teodoro177:
What procedure should I do?
Every time I try an action or click on Scan GTS to work:(
(Until now unable to do anything)

#help
(Sorry if there is any error in the language, I am not fluent in English and I am using Google Translator)

shun-iwasawa:
@gab3d and @Teodoro177 
Thank you for posting the issue, but could you please re-post the GTS-related problems to https://github.com/opentoonz/GTS/issues in order to organize? Thanks!

@OliverGee 
> OpenToonz sees the scanner in Scan Settings, and is set to use the Internal driver, but it will not scan.

Since we used GTS for production work in Studio Ghibli, we did not use or maintain the scan feature in OpenToonz. For now AFAIK the feature is unchanged from Toonz Harlequin and released as-is. And it would be possible that development of other features in the past broke some functionality of the scan feature, unfortunately. 
Since there already is GTS, I'm sorry to say that for now I myself am not planning to modify the scan feature in OpenToonz.

OliverGee:
@shun-iwasawa 

Thanks for the update on the status of scanning in OpenToonz. In that case could you please recommend an A3 feed scanner that you know will DEFINITELY work with OpenToonz? We have had to revert to Toonz 4.6 running on Windows XP since our old Fujitsu scanner died (in use since the 90's) and it is not ideal.

GTS does not have a feed option, does it?

[...]

PR #1491, opened 2017-09-27
Allow to compile 32bit version with MINGW

morevnaproject:


shun-iwasawa:
Please let me clarify some points:

- What is this PR for ? As you can see [how_to_build_win.md](https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_build_win.md#creating-the-files-for-the-srv-folder), there already is the way to build the 32-bit version on Windows.

-  According to the commit message, it seems that this PR disables QuickTime, right? If so, sorry but it is unacceptable since one of the major reasons for keeping 32-bit build available is to use QuickTime. In Japanese animation industry MOV is still used commonly and indispensable for now.

morevnaproject:
Dear @shun-iwasawa ! Apologies for the delay. Here are the answers:

- This PR is made as requested by @turtletooth in comment here - https://github.com/opentoonz/opentoonz/issues/1039#issuecomment-332400345. This PR adds support 32bit version using MINGW toolchain (configuration of MSVC should remain unchanged).
- The PR disables QuickTime for MINGW toolchain only. MSVC configuration remains unchanged.

PR #1494, opened 2017-09-28
Update Simplifed Chinese UI Translation

jcome:
- Synchronizezh_CN .ts files with src changes and update translations accordingly
- Update zh_CN .qm files


jcome:
why the merged commits still here?
let me close this and push to a new branch.

PR #1498, opened 2017-09-29
Renumber crash fix

### LINKED ISSUE #1312: Autorenumber + Save + Revert to Last Saved Version = crash (gif)
ezgif-1-dbcb8b01d2
I'm sorry, I probably should have written more in detail on the video, (I can not do gif for more than 30 seconds)

manongjohn:
This PR fixes #1312

The main cause of the crash in this scenario is the disappearing drawing(s) when trying to display something that was renumbered. This crash appears only to happen with Toonz Raster and normal Raster levels that have been renumbered.

In a nutshell, when you perform a Renumber (Auto or manual), there are a few places internally where the image-to-frame (level strip, not xsheet) relationship is not being updated properly.

The issue usually isn't obvious until you actually save the level/scene because it changes where it might be pulling the images from, cache or file. Depending on what you are doing you may or may not notice a drawing has disappeared. Once the drawing disappears like that, it could crash OT depending on which internal structure is being referenced to display it.

NOTE: Technically this Renumber bug also impacts Toonz Vector levels, but because of the way Vector level images are saved and displayed, the discrepancy of internal dependencies has almost negligable impact.

### DIFF
--- a/toonz/sources/include/toonz/imagemanager.h
+++ b/toonz/sources/include/toonz/imagemanager.h
@@ -288,7 +288,7 @@ protected:
                                           std::vector<TFrameId>,
                                           std::vector<std::string>, bool){};
 
-  virtual void setFid(const TFrameId &fid) {};
+  virtual void setFid(const TFrameId &fid){};
 
   //! Clears the builder's cached data.
   virtual void invalidate() {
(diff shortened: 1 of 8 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed the fix. Thank you @manongjohn !

PR #1501, opened 2017-10-02
Fix cell selection behavior

shun-iwasawa:
This PR fixes a niche, but very very frustrating behavior in the software as follows:

"Sometimes the xsheet cell selection released on right-click. "

After investigation, I found that this behavior happens when the Level Strip panel is opened along with the Xsheet. Right-clicking on the xsheet causes unwanted change of the current selection to the `TFilmStripSelection` which causes the cell selection's release.

Also I fixed in this PR the incorrect disconnection done in `FilmstripFrames::hideEvent()` .


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1505, opened 2017-10-04
Update Japanese translations

shun-iwasawa:
This PR mainly updates Japanese translations.
Also fixed the following two points:

- Made a window title of `PencilTestSaveInFolderPopup` translatable.

- Made the option texts of xsheet column layout in the preferences popup NOT translatable. They are used as key values in the preferences, so replacing with another string will break functionality.


shun-iwasawa:
@manongjohn Sorry for my late notice, but I modified the option texts in the preferences popup ("Classic", "Classic-revised", and "Compact") to be non-translatable. Can you please check if it is correct?

Jenkins

manongjohn:
@shun-iwasawa LGTM

shun-iwasawa:
Thanks @manongjohn for taking a look. I'll merge this.

Issue #1506, opened 2017-10-04
XSheet Update Issues. 


BrundleThwaite:
I really like the new column headers but there are 2 problems that I am having.

1: I can't select a frame. It always prompt's me to rename the frame and I can only perform other actions by right clicking the behavior should be the other way round this behavior makes it almost impossible to use shortcuts as it is always trying to change the frame name.

2: I can no longer use undo when the cursor is over the XSheet

manongjohn:
That behavior sounds like you have "Enable to Input Cells without Double Clicking" enabled in the Xsheet Preferences.

BrundleThwaite:
Sorry, My bad

PR #1507, opened 2017-10-04
Set default Xsheet icons and background colors

manongjohn:
This came about from Issue #1506.

When upgrading from an existing installation of OT, and either they are still using a deprecated theme or they somehow have one of the newer themes missing the Xsheet icon/bg color configurations, the Xsheet displays no icons and black backgrounds.

This PR will use internally set default Xsheet/Timeline icons and background colors, if the stylesheet is missing any information.


shun-iwasawa:
@manongjohn
Thank you for posting this but for now I'm quite hesitant to introduce this PR.
Missing the Xsheet icon/bg color configurations implies that it also lacks other important setting files which must be updated such as other icons, the fx list, the default menu bar settings, etc.
Providing only the default Xsheet icons will not be a complete solution and may make users to overlook such failure. IMO it would better to lead them to re-install the software in such case.

manongjohn:
@shun-iwasawa 
Ok.  I can see where you are coming from.  It's such a minor issue at this point.  I'll close it.  Thanks for reviewing and considering.

shun-iwasawa:
Thank you @manongjohn for kind understanding.

Issue #1488, opened 2017-09-22
Enable to Specify Level Paths as the Relative Path from the Scene's Parent Folder


shun-iwasawa:
This feature is what I mean to implement to OpenToonz, based on a request by Japanese animation studio.
For now, level paths are specified in the scene by one of the following ways;
 
1. Absolute path.

1. Relative path from the project folder ( using the project folder aliases such as `+drawings` ).

Using the project folder is powerful for managing the working data through production. However, for now both cases lacks **portability** of data. Once you move levels to another folder (for organizing, backup or another purposes) your scene cannot find the level resources unless you specify the paths in level settings.

In that context, I'm thinking of implementing a new variable (say, `$scenefolder` ) which can be used in the beginning of the file path. On loading the level, `$scenefolder` will be replaced by the folder path where the current scene file exists. By putting the scene and levels in the same relative position and using this variable, links to the levels will be maintained even if you move the folder anywhere or rename the folder.

(Please note that OpenToonz already has the `$scenepath` variable but it works quite different. Please refer to [the manual](http://opentoonz.readthedocs.io/en/latest/managing_projects.html#using-the-scenepath-variable-in-folder-definition) for details.)

![scenefolder_concepts](https://user-images.githubusercontent.com/17974955/30735858-75d801bc-9fbb-11e7-9089-602bbe0b3f1c.png)

Not only gaining data portability, implementing this feature will cause another beneficent influences as follows:

- Introducing this feature will allow us to make `Collect Files` command for collecting all resource file of the scene into one folder and make it portable, which is available in AfterEffects.

-  In June, [Celsys announced](https://www.celsys.co.jp/topic/20170626) that they started to study for formulating the universal x-sheet file format which is meant to be handled by multiple animation tools. In their initial proposal, all levels/resources used in the scene MUST be contained in the same folder as the scene file. IMO limiting the resources to be placed in one place would be inconvenient considering that there will be resources shared by multiple scenes. However, introducing `$scenefolder` will become capable to follow their concept. 
<img src=https://www.celsys.co.jp/files/user/news_release_20170626_02-600x0.jpg width=500>

gab3d:
...sounds useful to me.

as always, it'd be a bonus to have this option exposed somehow in the UI,
so non technical users know they could make such a choice, when creating or
saving a project.


El sept. 22, 2017 5:54 AM, "shun-iwasawa" <notifications@github.com>
escribió:

[...]

PR #1513, opened 2017-10-10
Introducing $scenefolder Alias

shun-iwasawa:
Implemented #1488 

- Introduced a new alias `$scenefolder` . If you put it at the head of file path, when the file is actually saved/loaded such alias will be replaced by a path to "parent folder of the current scene file" .

- Added a new preference option to specify which alias will be prioritized between the project folder aliases (e.g. "+drawings", "+extras", etc.) and the `$scenefolder` alias.
  - Let’ say that the current scene is saved in `C:/myproject/scenes/foo/bar/myscene.tnz` under the project with the alias `+scenes` is set to `C:/myproject/scenes`. When you save a level "A" at `C:/myproject/scenes/foo/bar/baz/A.tlv` , such path can be expressed in two ways - `+scenes/foo/bar/baz/A.tlv` and `$scenefolder/baz/A.tlv` . So, this option can specify which aliases will be taken on this alternative.
  - There are 3 options as follows; Prioritize `Project Folder Aliases`, `Scene Folder Alias`, and `Use Project Folder Aliases Only` . The default value is `Use Project Folder Aliases Only`, which will never use the `$scenefolder` alias on coding the path, and can be used as the same condition as the existing version.

- Added a new tree node "Scene Folder" in the File Browser. It will give direct access to `$scenefolder` .

- Please note that saving with the `$scenefolder` alias and the "Scene Folder" node is only available when the scene is already saved (i.e. not "untitled"). If you save some level using the alias in some untitled scene, a warning popup appears with a message saying "Save the Scene First".

### Disadvantage of $scenefolder

- If you use `Save Scene As` command and save the current scene in the different folder, levels specified with `$scenefolder` alias will lose access to file and will not open properly when you load the scene later. I made a warning popup to open for such situation to prompt users to update the level path.


shun-iwasawa:
Jenkins

turtletooth:
> If you use Save Scene As command and save the current scene in the different folder, levels specified with $scenefolder alias will lose access to file and will not open properly when you load the scene later. I made a warning popup to open for such situation to prompt users to update the level path.

I don't know if this is practical, but could all loaded levels be saved in the new location as well?

shun-iwasawa:
> I don't know if this is practical, but could all loaded levels be saved in the new location as well?

Well, I intentionally made the levels not to be saved in the new location automatically, because there may be a case that some another file already exists in the new location. 
By adding the dirty flags, you can choose the option to save all of them in the new location from the warning popup on closing the scene. I think `Save All` command will also work for saving them in the new locations.

shun-iwasawa:
@turtletooth 
I realized that it is difficult to save the levels afterwards if there is no image cache left.
So, I changed my mind and modified the behavior based on your idea. (thanks!) 
On using Save Scene As command and saving the current scene in the different folder, the question dialog will appears asking to choose from the options as follows:

image

Jenkins

shun-iwasawa:
Conflicts resolved. 
Jenkins again

masafumi-inoue:
LGTM

PR #1515, opened 2017-10-10
Modification in Camera Capture : Increment Alphabet on Capture

shun-iwasawa:

It changes behavior of the frame number field of the Camera Capture popup, with the preferences option `Show "ABC" Appendix to the Frame Number in Xsheet Cell` being ON.
In the current implementation, the frame number field will always be incremented to the next "number" without alphabet appendix.
(i.e. if the frame is "001" or "001A" or "001B" before capturing, the next frame will become "2" in all cases.)

I changed the behavior, to be incremented to the next "alphabet" if the previous frame has the alphabet appendix (i.e. "001A" -> "001B", "001B" -> "001C", "001I" -> "002", "001" -> "002"), because it is more frequent to do so.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1516, opened 2017-10-10
Refine Right-Click Menu of the Xsheet Cell Area

shun-iwasawa:
This PR is to organize the right-click menu of xsheet cell area, which seems to be complicated with a lot of commands for now.

- Some relevant commands are put into submenus.

- Edit commands (such as `Copy`, `Cut`, `Insert Paste` etc.) are newly added. It would be useful especially for  LCD pen tablet users who tend to use click operations rather than shortcut keys.

Please see the following image for the details of changes:


shun-iwasawa:
Jenkins

RodneyBaker:
Most excellent.  These improvements greatly help us navigate through the UI.
I find myself smiling that the xsheet is again receiving such valuable attention.
 
Thanks!  

Added:
Auto-sorting aside for general organizational consideration in OT:  
I'd like to recommend an auto-alphabetizing approach to menus as this would help users navigate through listings quickly.  I do note the un-alphabetized listing does start with Cut which while that option might be advantageous in some situations it does raise some concern regarding inadvertent data loss if/when the first option available is accidentally selected.   Assuming auto-alphabetizing the order of the new commands area would change to:

Copy  
Cut
Delete
Insert
Insert Paste
Paste Into

So, at least in this case auto-sorting of menus would appear to be an improvement in several ways perhaps most importantly in the safety of the first option being 'Copy' rather than 'Cut'.
Another example of a menu listing that would benefit greatly from auto-alphabetizing might be long lists of options such as FX.  

Other considerations:
I'm not sure how well this auto-sorting would work for other languages but I must assume it would benefit those as well.

Alphabetical sorting should be easy (mostly tedious) to implement assuming that such an approach is preferred.  

Auto sorting of menus would further standardize all of the menus in OT and any menus not auto-sorted would then emphasize those non standard menu orders even more precisely because they are not auto-sorted.

gab3d:
@Rodney
I'm not against menu order being alphabetically sorted, but I think "Cut / Copy / Paste" have a historical order that is standard in every computer user expectations, and would be just weird if presented otherwise.
Other options would be fine if sorted that way, the problem would be that if that applied to every langage, then we would have different option order in each one, and that wouldn't be very good for following tutorials in other languages, for ezmple.

RodneyBaker:
Thanks @shun-iwasawa  for the feedback.
You raise important points.  I do think the historical order is generally best maintained.  My feel is that historical context doesn't take into account use of modern day input devices BUT that is offset by our ability to define how sensitive those input devices can be or by granting double click options, etc.  And of course the failsafe for accidental deletion is usually a confirmation screen.  What?   Delete?  No!  Cancel!  Cancel!  Cancel!  The further safety net being that of non-destructive workflows and Undo histories.  OpenToonz is pretty well equipped in this arena.

Very good point about the variation that would be generated with translations.  We definitely wouldn't want that.

I will guess that most customizable approaches mainatain an index and the user then modifies that index which in turn changes the order of options on the menu.

Sorry to go off topic.  I'm easily convinced.  :)

shun-iwasawa:
Conflict resolved. `Paste Numbers` command is put into `Paste Special` submenu together with `Paste Into` command like as follows:

image

Jenkins again

turtletooth:
LGTM.  Thanks @shun-iwasawa

turtletooth:
BTW - Is Import Magpie File used by anyone you know?

PR #1517, opened 2017-10-11
Fix for Shaders and Switch to QOpenGL by turtletooth and blackwarthog (modified)

shun-iwasawa:
This PR is based on #1308 by @turtletooth and @blackwarthog , with a little modification to reduce assert errors.
Although I still do not properly understand the "modern" OpenGL , it seems to work for zerary shader fxs with no errors. Unary shader fxs seems to have some errors yet. 
As I mentioned in #1308, I confirmed that fortunately this change will also solve the issue #1081 (crash with full screen mode on OSX) .

### What I did

I found in the debug message saying `Attempting to create QWindow-based QOffscreenSurface
outside the gui thread.` on crash. So I Google'd the message and found [this website](https://git.merproject.org/mer-core/qtdeclarative/commit/fae16f9d2fca773276eba150d0414fb2f303d179?view=parallel).
It seems to tackling the same problem so I tried to follow the way done in the website - that is, bringing out the creation of `QOffscreenSurface` from the worker thread.


turtletooth:
Thank you so much for tackling this.  I'm going to test it right now!

shun-iwasawa:
@turtletooth 
>Thank you so much for tackling this. I'm going to test it right now!

Thank you!

I think this will be worth merging as soon as we confirm that the modification seems not to have a bad effect in the existing functionality - even if there still be the problem in shader fx, it does fix the crash bug in OSX anyway!

Jenkins

turtletooth:
@shun-iwasawa Are there any areas that switching OpenGL is likely to affect that I should really hammer on and try and break?  I don't know enough about where OpenGL is being used and what should be tested.

One thing I did in my original PR that I couldn't figure out how to fix so I just commented out was the assert on line 574 of sceneviewer.cpp.

turtletooth:
@shun-iwasawa I found that changing line 89 of shaderfx.cpp to:
 `~ProgramBinder() { glUseProgram(0); // m_prog->release();`

seems to fix the crash on using the blur shaders.  According to http://doc.qt.io/qt-5/qopenglshaderprogram.html#release they are functionally equivalent commands.  (But I never mess with OpenGL, so I may be missing something.)

shun-iwasawa:
@turtletooth Thanks! I'll check and try it.

shun-iwasawa:
You're awesome. It works!

turtletooth:
When I closed OpenToonz, I did get a failing assert on line 64 of tgldisplaylistsmanager.cpp.  I can recreate that just by adding a line to an empty scene and closing - no fx needed.

turtletooth:
That failing assert doesn't seem to be related to the fix you just implemented.  I still get it before the fix.

shun-iwasawa:
Thanks for pointing this. I'll check it.

shun-iwasawa:
It seems that this assertion failure does not occur with the current master using `QGLWidget`. Actually I do not yet understand the purpose of `TGLDisplayListsManager` - I'll try to investigate it.

shun-iwasawa:
It looks like that `TGLDisplayListsManager` is used for sharing context between `SceneViewer` panels and is specific for `QGLWidget`. I'll try to take different way.

shun-iwasawa:
Updated as follows:
- Stopped using `TGLDisplayListsManager` for sharing context and use global share context ( enabled with `Qt::AA_ShareOpenGLContexts` flag ) instead.
- Made `GLWidgetForHighDpi` to inherit `QOpenGLFunctions` according to[ the documentation](http://doc.qt.io/qt-5/qopenglwidget.html#code-examples).

@turtletooth AFAIK OpenGL is being used in as follows:
- `SceneViewer` (for the Viewer and the Combo Viewer)
- `ImageViewer` (for the Flipbook and the Color Model panels)
- `HexagonalColorWheel` (for the Style Editor panel)
- `PlaneViewer` (for the Swatch in Fx Settings popup, etc.)

As far as I can see the above panels seems to work as properly as before.
I'm a bit concerned especially about the plastic tool which is using textures. It looks OK but I may miss something as I'm not used to using it.

Jenkins

turtletooth:
@shun-iwasawa Confirmed that it is fixed.  What do you think about me releasing this in a version of OTX for people to play with and give it a week or so to see if any big bugs pop up?

turtletooth:
@shun-iwasawa I found a deal breaker - high dpi.  

I was trying to fix the cursor offset that happens sometimes on high dpi monitors and was reviewing your high dpi PRs.  You mentioned that you had to revert to qglwidget because of a glitch you were getting with qopenglwidget.  So I tried this PR with my monitor set to 150% scaling and got this:

highdpi_shaders

It seems that when the Windows display scaling is not set to the recommended value, there are issues.

I spent 4 hours on this tonight trying to figure out what was going on and didn't really get anywhere.  I'd hate to lose the functionality gained with this PR - and everything seems to work right at the recommended resolution.  

Do you have any thoughts?

shun-iwasawa:
@turtletooth Thank you for finding this. I'll take a look.

turtletooth:
BTW- with this PR, I'm not sure that getDevPixRatio() is needed anymore. Drawing with the screen scaling modified, the cursor tracking was all over the place, but when I made getDevPixRatio() return 1, it evened out.

shun-iwasawa:
To recap, there seems to be two problems regarding the high dpi.

1. Strange artifacts appear in the main window and in the viewer, such as doubling menu bar.

1. `getDevPixRatio()` does not work properly and some elements offset in the viewer.

I could not found the exact cause of the problem 1, but confirmed that the artifacts disappear in the build with Qt 5.9.2 . So, one option might be transferring Qt to 5.9 series - It'll be a big change but IMO it is somewhat reasonable since the [Qt 5.9 becomes a Long Term Supported release](https://blog.qt.io/blog/2017/05/11/introducing-long-term-support-qt-5-9/) .

The problem 2 occurs regardless of the Qt version. As far as I can see it cannot be fixed by just removing the `getDevPixRatio()`. I'll try to tackle with it.

shun-iwasawa:
I tried combining this with Qt5.9 migration ( #1520 ) as I would like to see if it works on OSX.
Jenkins

shun-iwasawa:
@turtletooth 
Confirmed it works fine on OSX now. Can you please try this? I fixed the offset problem in the viewer as well.

>@shun-iwasawa Confirmed that it is fixed. What do you think about me releasing this in a version of OTX for people to play with and give it a week or so to see if any big bugs pop up?

Thank you so much for your proposal. It would be nice to test the modification in OTX for about a week before merging it. (Actually, Jenk.ins is already switched to Qt5.9.2 so the current nightly build should not work properly on OSX..)

turtletooth:
@shun-iwasawa Thanks for all your work.  You are amazing.  I will make a new OTX today or tomorrow.

turtletooth:
@shun-iwasawa Thanks for your work.  I confirmed that the offset seems to be fixed.  I found a crash that I have duplicated on two different machines.  If the machine has two (or more) displays, and the displays don't have the same scaling, it crashes when switching rooms.  The crash happens on 1403 of sceneviewer.cpp - SetWindowRgn(...)

turtletooth:
Just commenting on the crash - If I comment out the line, the crash doesn't happen, but I do get some display issues at certain scaling- here is 200% (the recommended scaling) on my surface:

issue

turtletooth:
Another thing I noticed - the UI issue for the schematic seems to actually still be there when the displays aren't at the same scaling:
sch

fx

This shows up on two different machines.

shun-iwasawa:
@turtletooth 
Thank you for pointing the problem.
For now, the device pixel ratio is obtained once from one of the connected monitors (maybe, from the screen where the main widget resides) and will not be updated when the application is moved to the other monitor with different dpi.
IMO it can be a separated issue because such behavior is just the same as the current master branch.

turtletooth:
@shun-iwasawa I just tested, and the schematic issue is there in the current master, but the crash on changing rooms is not there.

I agree that the UI issue can be a separate issue, but I think that the crash is better fixed here.  What do you think?

shun-iwasawa:
@turtletooth I'll check the crash right away. Does it crash as well, when the multiple monitors have the same dpi ?

turtletooth:
> Does it crash as well, when the multiple monitors have the same dpi ?

No.  The crash only seems to happen when the monitors have different scaling.  I was testing this all day to make sure it wasn't just my configuration or if I was missing something.  But when the two monitors have different scaling, I could consistently make it crash by changing rooms (at least to rooms that had a viewer).

shun-iwasawa:
Understood. I'm not sure if I can prepare the instruments for reproduction, but I'll try it anyway!

shun-iwasawa:
@turtletooth 
Here are my verification result:
Tested with Mobile Studio Pro (with high DPI, Win10 machine) + An extra monitor (with normal DPI).
The crash happens when I launch OpenToonz on the sub (non-main) monitor.
If I launch OT on the main monitor, it works properly even if I move the main window to the sub monitor.
( The "main monitor" means the monitor with which the option `Display>Multiple displays>Make this my main display` being ON. )

Unfortunately, I think it would be difficult for me to investigate this problem further for now, since my development machine is Win7 and Win7 doesn't have capability to set the different dpi to each of the connected monitors.
I'll try to setup the development environment in Mobile Studio Pro, but some time is needed to do so.

It is difficult to valuate pros and cons of introducing this PR. But IMO it is rather important to solve the crash problem under single-monitor environment like #1081 .

One quick and possible patch would be to skip the line 1403 of `sceneviewer.cpp` if the current OS is Windows 10. ( From Qt 5.9, it seems that we can easily detect the current OS by using [QOperatingSystemVersion::current()](https://doc.qt.io/qt-5/qoperatingsystemversion.html#current) :-) ) It can, according to your report, at least avoid the crash. How do you think about it?

turtletooth:
I found another thing this seems to break - z depth.  To test:
create a simple level
use the edit tool and try and enter 10 in the z depth.

I can't do it.  I changes it back to 16.something.

turtletooth:
Moving the z depth with ctrl in the edit tool seems to be very slow compared to previous builds.

turtletooth:
Here is the OTX that I'll be releasing tomorrow with this in it: 
https://github.com/turtletooth/opentoonz/tree/otx_10-17-17

shun-iwasawa:
> Moving the z depth with ctrl in the edit tool seems to be very slow compared to previous builds.

Thank you for notifying this. I could reproduce the problem and will check this.

I found a new problem with right-clicking on the full-screen mode viewer. On Windows7, the viewer blacks out. On Windows10, it fails to show the context menu. In both cases the software doesn't crash and back to normal if you blind-click the position of the "Exit Full Screeen Mode" command.

shun-iwasawa:
It seems that in the line 1653 of sceneviewer.cpp the function `glGetDoublev(GL_PROJECTION_MATRIX, projection);` fails to obtain the projection matrix properly with this build.
I'll fix this by introducing a projection matrix variable into `SceneViewer` . It is only for fixing this problem for now, but will be alternative to the existing GL_PROJECTION_MATRIX considering that in modern OpenGL there is no longer a transformation matrix state.

shun-iwasawa:
The problem regarding the full screen mode seems to be related to [this](http://doc.qt.io/qt-5/windows-issues.html#fullscreen-opengl-based-windows). I'll tackle with it.

shun-iwasawa:
I think I could fix both of the problems.

@turtletooth As mentioned in #1527 , I tried to remove the line 1403 of sceneviewer.cpp. As far as I can see, it works with no problem on Win7.

Jenkins

turtletooth:
@shun-iwasawa Thanks so much- It seems to work great.  The titlebar glitch I posted above even seems to be fixed as well as the brush stroke offset that sometimes happens on certain devices.

I found one bug related to my multiple screens with different scaling, but it seems to not be related to this since it also happens in previous builds.  I'll post that as a separate issue.  Thanks for all your hard work on this.  I think this is going to be a great move for OT.

turtletooth:
@shun-iwasawa Here is the latest OTX if you want to play with any of the other features: https://gum.co/TZmmM

shun-iwasawa:
Thank you @turtletooth , 
I temporary reverted Jenkins to Qt5.6 now. After merging this , I'll change it to 5.9 again.

turtletooth:
Just to give an update- I haven't heard any major issues so far with this.

shun-iwasawa:
@turtletooth Thank you for letting me know the good news. How about considering to merge this on Thursday, if any critical problem will not be reported by then?

After completing some development, I will prepare for the next version (v1.2) to be released next month.

gab3d:
@shun-iwasawa I've just had a reproducible crash with Glitter effect applied to a raster layer (jpg image), after I've raised the "Halo" parameter to 10. The program stopped responding and the crashed.
I'm on a Windows 10 notebook, i7, 16GB RAM, Geforce GT650M with latest drivers.

It seems something related to memory, because when previewing a region I can raise the parameter much more before the program hangs. I've also tested it with a smaller image and can also raise the number more than with a bigger image, before it hangs.

turtletooth:
@shun-iwasawa Sounds good.  I don't know if you want to try and tackle the issue from @gab3d or leave that for a later PR.  I like the idea of moving toward 1.2.  What are you plans on the current PRs?  Do you have the time to try and look at most of mine?  I'll try and start looking at yours tomorrow.

shun-iwasawa:
@gab3d 
Thank you for reporting the problem. I think it can be another issue to be tackled later since the problem seems to be only related to the shader fxs and not so critical that it will interfere the existing usage of the software. Thank you for your understanding.

@turtletooth 
> What are you plans on the current PRs? Do you have the time to try and look at most of mine?

Yes, for a period of weeks before releasing the next version I'd like to concentrate on reviewing the current PRs and bug-fixing.

>  I'll try and start looking at yours tomorrow.

Thank you !

shun-iwasawa:
@turtletooth 
Brush delay was reported in the forum. In my environment it seems to work OK...
Can you reproduce this problem?
https://groups.google.com/forum/#!msg/opentoonz_en/G9OI5Tc_QJA/-ZcTWX3CBwAJ

turtletooth:
I saw that too.  I tried to reproduce it both yesterday and today and I don't really see any difference between current drawing and new drawing.  The only difference is that on my surface, the cursor is now actually where I want it.

turtletooth:
@shun-iwasawa I have had 60 kids using OTX this week and drawing seems to be working fine on our machines. Are you ready for this to be merged? I think it's fine, and issues that pop up can be handled as they arise.

shun-iwasawa:
@turtletooth Thank you for testing the software in such a large scale ! Yes, I agree it is fine to merge.
Please give me 15 minutes, for updating Jen.kins to Qt 5.9 before merging this.

shun-iwasawa:
@turtletooth Update of Jen.kins is done. I think it's ready to merge.

turtletooth:
Thanks @shun-iwasawa @blackwarthog for your great help with this.  LGTM.

PR #1519, opened 2017-10-12
Spanish UI - update up tp latest Japanesse files.

gab3d:
Hi @shun-iwasawa,
This is my first attempt to do a PR that only has my latest changes and not all the history included.
I'm still not sure if I'm doing it right, so please bear with me :)
Thank you for your patience!


shun-iwasawa:
@gab3d Thank you for updating the translation.

If possible, could you try to squash your commits into one before pushing PR in the future ?
[Here](http://gitready.com/advanced/2009/02/10/squashing-commits-with-rebase.html) is a description of `git rebase -i` command.

The above operation is mainly for easy-reviewing. You don't need to worry about the commit history of upstream as we have an option "Squash and merge" in github.

LGTM

Issue #1422, opened 2017-08-21
(WIP) Timeline Enhancements


Agnyy:
You all already know me and should understand that in the Timeline I invest a lot of energy and all the more so in the idea of functional reals. I ask you to read my plan for the development of the timeline, I take into account many of your requests, but I also add new functions. This timeline will be integrated into the concept of the timelines sbeb (After Effects, TVPaint, Flash) https://docs.google.com/document/d/1CyVbJ6dlhFC-IjEIcG52RoUt45PQ6dvDvEO7eL5GQcw/edit?usp=sharing 
Now the concept is implemented only by 60% in the future will be added and other functions. I very much ask you to respect my concept. Thank you in advance!

Agnyy:
P.S. 
I apologize for the bad translation. I do not know English therefore I translate through Google translator.

bobbybutterfly:
@Agnyy  Thanks for your hard work!
Functionality like in After effects would be a dream come true. Especially the zoom/increase decrease frame size would be wonderful for scenes longer than 2 or 3 seconds.
Also: The function curves should be accessabble from the timeline directly
Thank you:)

RodneyBaker:
Looking very good!
Thank you for your work and dedication to adding and enhancing the Timeline.

openanim:
Impressive work ! Will this be compatible in xsheet mode? Or specific to the timeline ?

manongjohn:
@Agnyy You should post a link to the video you have of the timeline enhancements so people can kind of see it in action.

Agnyy:
https://www.youtube.com/watch?v=ZNJgmsilyt4

artisteacher:
I'm super excited about the addition of keyframing the contents of the vector levels!

I'm curious about the new process for adding levels as new layers vs inside an existing layer once the empty columns are removed. Right now, the selected cell determines whether a new level makes a new column/layer or if the new level is added to an existing column/layer.

gab3d:
@Agnyy, wooow!!  can I be your friend?!  :D 
If you complete this project, the workflow in OT for digital animation would be revolutionized, I am so excited!

I'd just like to comment on some topics that could be added, as well, to improve it even more:

[...]

PR #1522, opened 2017-10-14
Current time timeline indicator

manongjohn:
This PR adds a current time indicator to the Timeline mode.  This is similar to the indicators found in other software and will become more useful once the ability to zoom in/out of timeline is added.

Notes:
-  This currently does not apply to the Xsheet mode.
- The indicator can be turned off via a preference setting or by right-clicking in frames area and using option in context menu.
- The indicator is initially off by default 
- Line appears behind cell text to maintain readability.

Here is what it looks like with and without Onion Skin mode turned on:

image

This was requested as part of the timeline enhancement proposed by @Angyy (Issue #1422, Change # 6)


bobbybutterfly:
Very cool and handy! I missed this!

Agnyy:
Good!

gab3d:
nice!
it makes it easier to spot at a glance where the current frame is at.

turtletooth:
Tested a bit and it works well.  I actually think it should be on by default.

manongjohn:
@turtletooth Thanks for testing it out. I know you guys are busy with the upgrade to 5.9.

I was recently thinking the same thing and see no harm in defaulting it to on either. I will change it shortly.

manongjohn:
@bobbybutterfly You need to turn on Onion Skin under Preferences.

manongjohn:
Oops, i meant you need to turn it off to see the blue marker.

bobbybutterfly:
Ok, found it:) Thanks @manongjohn

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @manongjohn for adding the useful feature!

PR #1523, opened 2017-10-16
Remove console for MinGW build

blackwarthog:
In current MinGW build OpenToonz opens with a console window. This patch removes this console.


turtletooth:
LGTM.  Thanks.

PR #1527, opened 2017-10-18
Fix for Crash with Multiple Monitors with Different Scaling on Qt 5.9

turtletooth:
This is a fix for a specific use case.  When using Qt 5.9 and multiple monitors with different scaling, OpenToonz will crash when switching rooms.  

@shun-iwasawa If you want to work this into your PR, that is fine.  I made this so I can build OTX for testing.

### DIFF
--- a/toonz/sources/toonz/sceneviewer.cpp
+++ b/toonz/sources/toonz/sceneviewer.cpp
@@ -82,7 +82,7 @@
 #include <QInputContext>
 #endif
 #include <QGLContext>
-
+#include <QOperatingSystemVersion>
 #include "sceneviewer.h"
 
 void drawSpline(const TAffine &viewMatrix, const TRect &clipRect, bool camera3d,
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
@turtletooth 
Thanks! I'll try this in my environment.
Please let me know just to make sure - does the UI glitch you mentioned in [this comment](https://github.com/opentoonz/opentoonz/pull/1517#issuecomment-337042687) occur with single monitor environment as well?

turtletooth:
@shun-iwasawa I don't think so.  I think it's only with multiple monitors with different scaling.

shun-iwasawa:
I confirmed that this PR avoids crash and seems not to affect the usage in other cases. I'll merge this once #1517 will be merged. Thank you @turtletooth !

shun-iwasawa:
@turtletooth I found something in the [Qt documentation](http://doc.qt.io/qt-5/qopenglwidget.html#relation-to-qglwidget), saying as follows:

> Note: Avoid calling winId() on a QOpenGLWidget. This function triggers the creation of a native window, resulting in reduced performance and possibly rendering glitches.

Also I have tested the usage on Windows7 commenting out the line 1470 and found that it seems to work in any case of multiple monitors or high-dpi conditions.
This line comes from Toonz Harlequin source. IMO "the problem related to Windows 7" does not exist anymore ( in `QOpenGLWidget` or in the current Qt version ) and we can remove this line at all.

turtletooth:
Closing - updated in #1517

PR #1530, opened 2017-10-19
Paste Numbers Command

shun-iwasawa:
This PR will introduce a new command "Paste Numbers" . This feature is demanded by some Japanese animation studio.

- This command works when the xsheet cells are copied in the clipboard.
- Unlike normal "Paste" command, it will paste only numbers of copied cells, keeping the level unchanged.
- This command will overwrite the cells. No matter how many cells are selected on trigger this command, it will paste numbers of all the copied cells starting from the upper-left corner of the selected cell region.  
- If the destination cell is empty, search upper and lower cells in the column and use the level of occupied neighbor cell. This command will do nothing to empty column or columns without cell numbers.
- As a special option, this command can paste single column to multiple columns. In such case, the same number sequence will be duplicated to the selected columns.

For now this command is tentatively added to the end of the right-click menu of the xsheet cell area. The command position will be modified after merging #1516 .

### DIFF
--- a/toonz/sources/include/toonz/txshlevelcolumn.h
+++ b/toonz/sources/include/toonz/txshlevelcolumn.h
@@ -85,6 +85,9 @@ Return \b TFx.
 */
   TFx *getFx() const override;
 
+  // Used in TCellData::getNumbers
+  bool setNumbers(int row, int rowCount, const TXshCell cells[]);
+
 private:
   // not implemented
   TXshLevelColumn(const TXshLevelColumn &);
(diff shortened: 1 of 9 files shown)


shun-iwasawa:
Jenkins

gab3d:
Sounds useful!
I like it. 👍

shun-iwasawa:
@masafumi-inoue Can you please take a look on this ?

masafumi-inoue:
LGTM

beeheemooth:
Oh! That's super useful feature, thanks!

Issue #1490, opened 2017-09-26
UI:  Unstable panel sizing


beeheemooth:
This thing happens every time you start the program or while minimizing the window.

![ezgif-2-97998ea432](https://user-images.githubusercontent.com/25724133/30871756-8b77068e-a2f0-11e7-9478-af35769915f1.gif)

blurymind:
yeah, it is a really old bug, still not solved :D
I think @turtletooth tried it once

turtletooth:
I made the level strip width fixed based on the thumbnail size in the level strip overhaul PR I made, but undid it since people might want to resize the level strip.  I guess I could make it a preference.

What are your thoughts?

blurymind:
It would be nice if OT saves the width somewhere and loads it when
required. Not only when resizing the window, but also when the application
is restarted

blurymind:
I am actually surprised that it already doesnt do that. Other panels seem
to remember their width and restore it

beeheemooth:
I think that each user can stretch or narrow the panel as he wishes, the main thing is for the program to remember this size. It is not at all necessary to snap the panel size to the size of the thumbnails. 
There are many other panels that also change the size when minimizing or restarting.

openanim:
I also think OT should remember the last workspace of the panels at the closing and opening. Imo, that would be  ideal.

2dvision:
There is also a bug for the "FX Edit" panel size.

![fx panel size](https://user-images.githubusercontent.com/16509592/30926534-feebd568-a3a4-11e7-8fcf-4089e8e5e564.gif)

Orphanlast:
I agree, this is something that should probably be given some attention.

shun-iwasawa:
@2dvision 
Thanks for pointing, Fx Settings issue would be owed by me. I originally made the popup to be resized to the same as its contents in order to avoid the trouble of having to scroll or resize it manually. But for now it seems to fail.

artisteacher:
It would be great if the icon size was more flexible - ie. based on the width of the level strip. I’d love to be able to set the level strip to a narrower width while still being able to view the entire icon. My icons display significantly larger than they look in the gif from @beeheemooth.

Orphanlast:
Does anyone know if anyone is working on this onem

artisteacher:
It's definitely still an issue.

openanim:
When you have a defined workspace for a room, it will be modified at the next opening of OT.

[...]

PR #1531, opened 2017-10-20
Template Env File

shun-iwasawa:
The user env file ( in `$TOONZPROFILES\env\username.env` ) is for storing the current state of the software in order to reproduce the same state on next launch.  The env file differs from the preferences in that, the preferences are manually modified via the Preferences popup, OTOH the env file is automatically saved on quitting the software.

Most of parameters stored in env file are "tend to fluctuate" on a short-term basis ( such as a size of brush tip ). But some of the parameters are more "stable" through production and obtaining default values from the template env file can avoid the trouble of having to setup such parameters for new users ( such as a current camera name and a current resolution set in the Camera Capture popup ). 

This PR will enable to use the template env file. If the personal env file is missing on launch ( i.e. the user firstly launch the software ), then the template env file ( `$TOONZPROFILES\env\template.env` ) will be loaded, if it exists.
This feature is demanded by some Japanese animation studio.

### DIFF
--- a/toonz/sources/common/tapptools/tenv.cpp
+++ b/toonz/sources/common/tapptools/tenv.cpp
@@ -171,6 +171,9 @@ public:
   }
 
   TFilePath getEnvFile() { return m_envFile; }
+  TFilePath getTemplateEnvFile() {
+    return m_envFile.getParentDir() + TFilePath("template.env");
+  }
 
   void setApplicationFullName(std::string applicationFullName) {
     m_applicationFullName = applicationFullName;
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

jpturcotte:
@shun-iwasawa Could this address #1490 as well?

shun-iwasawa:
@jpturcotte 
> Could this address #1490 as well?

Unfortunately, this is not related to the issue mentioned in #1490 . The main window geometry is stored in another file, `$TOONZPROFILES\layouts\settings.username\mainwindow.ini` which is for reproducing the window geometry on launching the software.

turtletooth:
LGTM.  Thanks @shun-iwasawa

PR #1417, opened 2017-08-19
Add Mojo for MyPaint brushes

RodneyBaker:
Add the Mojo for MyPaint directory and brushes

PR #1416, opened 2017-08-19
Add Mojo for MyPaint (brush set license)

RodneyBaker:
Prep work for adding Mojo MyPaint brush set by Saeger Ryman
Also:  Corrected spelling of word 'experimental' at top of document.

PR #1539, opened 2017-10-25
Mojo Brushes

turtletooth:
This combines #1416 and #1417


RodneyBaker:
Thank you Jeremy!

shun-iwasawa:
LGTM Thank you @RodneyBaker and @turtletooth for introducing them!

RodneyBaker:
Post mortem follow up in anticipation of (possibly) adding a few more mypaint brushes to the Opentoonz release.  This is largely possible due to the small footprint/file size of mypaint brushes.

Regarding licensing the following is the standard for including any mypaint brushes in the MyPaint distrubution.  It does not prevent other licenses from applying but does explain their position and the intended purpose of brushes:

> Brush packs
> Ideally, use Public Domain. Alternatively, use Creative Commons "CC-BY" or "CC-BY-SA" version 4.0, but release the raw brush settings into the Public Domain.
> 
> Long form: brush packs for inclusion into the main MyPaint distribution should be placed into the public domain just like background images, but they do not have to be. Instead you can use the Creative Commons "Attribution" or "Attribution-ShareAlike" licenses, both of which require redistributors to give you credit. Note that if you choose one of these fancier licenses, the pack's internal brush settings must still be Public Domain. Here's an example of how to say that in your brush pack's README:
> 
> Copyright <<YEAR>> <<YOUR NAME>>
> 
> Except where otherwise noted, this work is licensed under a Creative Commons Attribution 4.0 International License. See http://creativecommons.org/licenses/by/4.0/ for details.
> 
> As an exception to this and to the extent possible under law, <<YOUR NAME>> has waived all copyright and related or neighboring rights to the raw brush settings (those files ending with ".myb"). See http://creativecommons.org/publicdomain/zero/1.0/ for details.
> 
> We need the exception for the raw brush settings because those bits are copied around very enthusiastically by the MyPaint program — including into files others might want to distribute with their own license. Making the settings Public Domain allows them to do that. Of course, you can still claim copyright on the pack as a “work as a whole”, including all the preview packages and supplemental cover-type artwork in it, but excluding the raw settings.
> 
> If you want to use a Creative Commons license, it must be version 4.0, and it cannot be a NonCommercial or a NoDerivatives one.
> 
> If you choose Public Domain for the entire pack, you'll still be credited by us in the program About box unless you ask not to be named. ;)

PR #1540, opened 2017-10-25
Fix Fill Behavior on Toonz Raster

shun-iwasawa:
This PR fixes the problem when painting on Toonz Raster levels as follows:

When you draw some character which has light and dark parts, usually the self-colored line is drawn at the border, with the "Autopaint Lines" option activated. Conventionally, the self-colored lines are to be painted with a color of the dark area. So the workflow would be as follows:

1. Prepare an unpainted drawing. The self-colored line (red line) is drawn with the style with the "Autopaint Lines" option being ON.
1. Fill the dark area. The self-colored line is automatically painted with the dark color.
1. Fill the light area.

However, for now the result of the above procedure causes unwanted artifacts - which is because the light color protrudes too much, behind the center of the line. ( as seen in the left figure below )
To prevent such artifacts, for now you need to do quite troublesome way - first, fill the light area with the "Autopaint Lines" option being OFF, secondly activate the "Autopaint Lines" option, then finally fill the dark area. ( the result is in the right figure below ) 

autopaint_comparison

This PR will solve such issue.
With this PR, when filling the area the paint will not protrude the center of self-colored line anymore, as illustrated as follows:

fix_fill_behavior

I set this PR as WIP because I need some time to test this modification by some Japanese animation studio who demanded this fix.

### DIFF
--- a/toonz/sources/include/toonz/fill.h
+++ b/toonz/sources/include/toonz/fill.h
@@ -27,6 +27,7 @@ public:
   bool m_shiftFill;
   TPoint m_p;
   TPalette *m_palette;
+  bool m_prevailing;
 
   FillParameters()
       : m_styleId(0)
@@ -37,7 +38,8 @@ public:
       , m_maxFillDepth(0)
       , m_p()
       , m_shiftFill(false)
-      , m_palette(0) {}
+      , m_palette(0)
+      , m_prevailing(true) {}
   FillParameters(const FillParameters &params)
       : m_styleId(params.m_styleId)
       , m_fillType(params.m_fillType)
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins again

shun-iwasawa:
Jenkins again

shun-iwasawa:
The animation studio check is done. Although there is still some room for improvement regarding the Rect Fill, it can be done in the future.
Ready to be merged now.

masafumi-inoue:
LGTM

PR #1541, opened 2017-10-26
Customize Tool Cursors

shun-iwasawa:
This PR will make the tool cursor possible to be synthesized from multiple resource images ( base cursor image + decoration images for correspondent options ).
Now you don't need to prepare the cursor images for each combination of options.

For instance, Fill tool cursor with "Area" and "PolyLine" options under the Black BG view will be synthesized as follows:


shun-iwasawa:
Jenkins

turtletooth:
Seems to work fine.  Thanks @shun-iwasawa for the update.

PR #1542, opened 2017-10-26
Fix Boolean Tool Options Behavior with Shortcuts

shun-iwasawa:
This PR fixes the following two problem regarding the boolean (checkbox) tool options:

- The `Global Key` option of the Edit tool does not accept shortcut key control.

- Any of the boolean options do not change via shortcut key control, if even number of tool option bars are opened.


shun-iwasawa:
Jenkins

turtletooth:

PR #1550, opened 2017-11-02
Xsheet fixes

manongjohn:
This PR fixes the following issues:

1) The Lock OFF icon for the Default theme was shifted right after it was updated by another PR

   Corrected the position of the lock icon and widened both the white and black versions by 1 px to match the width of the Lock ON icon.

2) When selecting mutliple columns and then drag from any dragbar except the left most one (top most in Timeline), it causes the columns to immediately shift so that the left/top most column is the focus.

   Modified logic so prevent the jumping so regardless of which dragbar you pick, it will move based on that dragbar.

   **Change # 2 was removed from this PR and handled in the Timeline layer flip PR.**

3) When right-clicking an empty column that is part of a multi-column selection, the selection is cancelled and only the single empty column is selected. This prevents multi-column selection operations if the selection includes 1 or more empty columns and the empty column is right-clicked.

Modified logic to not cancel the selection when right-clicking an empty column when it is part of a selection.


toonz/sources/toonz/xsheetdragtool.cpp
```diff
@@ -1608,9 +1608,10 @@ class ColumnMoveDragTool final : public XsheetGUI::DragTool {
     assert(m_lastCol == *indices.begin());
 
     if (col < 0) col = 0;
-    if (col == m_lastCol) return;
-    int dCol  = col - m_lastCol;
+    if (col == (m_lastCol - m_offset)) return;
+    int dCol  = col - (m_lastCol - m_offset);
     m_lastCol = col;
+    m_offset  = 0;
```
shun-iwasawa: @manongjohn 
I detected a crash problem on the column drag.
To reproduce:

1. Put some level on the Column 3
1. Select the Column 1,2 and 3
1. Grab the Column 3 and drag the selection to left (on the xsheet mode, Up on the timeline mode).

How about modifying here like as follows to fix the problem?:
```
-------------------- toonz/sources/toonz/xsheetdragtool.cpp --------------------
index 2902bed..c750c29 100644
@@ -1609,9 +1609,9 @@ public:
 
     if (col < 0) col = 0;
     if (col == (m_lastCol - m_offset)) return;
     int dCol  = col - (m_lastCol - m_offset);
-    m_lastCol = col;
-    m_offset  = 0;
+    if (dCol < -(*indices.begin())) return;
+    m_lastCol = col + m_offset;
 
     assert(*indices.begin() + dCol >= 0);
```

Other modifications seems to be perfect. Thanks!

manongjohn: Thanks for reviewing. Actually. I think I will revert that specific change and take it out of this PR.  When I was working on the timeline flip PR, I had to include the fix there and did run into this issue.  I corrected it in that PR as it needed a bit more work so this change would be pointless.

shun-iwasawa: OK, it is fine to move this fix to the timeline flip PR. I'm planning to review that by the end of this week.

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed that the other two fixes work. Thank you @manongjohn ! LGTM
I'll ignore the Travis CI failure on timeout.

PR #1123, opened 2017-03-24
New Fx: Bokeh Iwa

shun-iwasawa:
This PR will introduce a new Fx named Bokeh Iwa, for defocusing the scene.

- It has 5 ports for the Source layers and 1 port for the Iris image.
- It considers characteristics of films (which is known as Hurter–Driffield curves) or human eye's perception (which is known as Weber–Fechner law). Both of them respond to stimulation of light in logarithmic characteristic.
- For filtering process I used KissFFT, an FFT library by Mark Borgerding, distributed with a 3-clause BSD-style license. 

image

For the usage of this Fx in detail, please refer to the movie uploaded [here](https://youtu.be/ooRVvJH-NX0).

PR #1555, opened 2017-11-06
New Fx: Bokeh Ref Iwa

shun-iwasawa:
This PR will introduce a new Fx named Bokeh Ref Iwa, for defocusing a single image by using a "depth reference image".

Most of specification is very similar to the Bokeh Iwa fx (implemented in PR #1123 ), except as follows:
- Bokeh Ref Iwa Fx only accepts one source layer.
- Also you need to input the "depth reference image", which represents the depth of the image in gray scale. (The nearest pixel is black and the farthest pixel is white.)

bokehref

By changing the on-focus distance, you can obtain racking focus shot from one deep-focused photo like this:
bokehref_onfocusdistance


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa Could you provide a test scene or settings for this?  I have tried testing this and keep getting a white screen or a distorted image.  I'm sure it's an error on my side, but I can't quite figure it out.

bokeh


By the way - is this supposed to be very ram intensive.  There were times that on a single 1920 x 1080 image this was using 2GB of ram.

shun-iwasawa:
@turtletooth Thank you for trying this fx. Here is my configuration: 

image

In your scene, please try setting the `Hardness` value to larger value. 
In this fx, exposure is calculated from RGB values with powers of 10.  The reciprocal of `Hardness` is used for the exponent. With very small `Hardness` value may overflow the maximum range of `float` in the some scene.

> By the way - is this supposed to be very ram intensive. 

Sorry to say, Yes. With 1920x1080 camera size it will allocate about 350MB of memories just for raster buffers used in this fx. Could you please try smaller camera and check if the scene can be rendered?

turtletooth:
Took a look at this again today.  I am getting an index out of range error.

Here is my setup:
Photo on Col 1 (500 x 500)
Outline of the man painted in black on a raster level on Column 2
6 sided polygon Toonz Raster level on column 3.

All columns exposed for 1 frame.

When I go to preview the fx, I get a crash.

turtletooth:
@shun-iwasawa I still can't get good results with this.  I get either no noticeable effect or a white screen.   Here is my setup:
fx
Here is the image I am using:
bikes-talon-29-vibe-3

shun-iwasawa:
@turtletooth 
The iris image (in column 3) should be painted in white as the fx uses its brightness. 

If you will still unable to render, can you please upload all the materials including the depth and iris images and the scene to reproduce on my side ?

shun-iwasawa:
[bokeh_ref_sample.zip](https://github.com/opentoonz/opentoonz/files/1480831/bokeh_ref_sample.zip)

@turtletooth This is another sample data which can be rendered on my environment. Please put the level files into `+extras` before rendering the scene.

image

turtletooth:
Thanks @shun-iwasawa - I was able to get it working with my tree scene.  It's always something simple.  

Is there any benefit in letting hardness go as high as it does?  I also think that fixing the crash on transparent images would be good, but can wait for a later time.

shun-iwasawa:
> The raster level had transparent pixels. Painting the empty pixels white stopped the crash.

> I also think that fixing the crash on transparent images would be good, but can wait for a later time.

@turtletooth 
This means that, you firstly made the depth image only by painting the near area in black and keeping the far area unpainted which was actually needed to be painted in white, right? 
If so, I can imagine the cause of crash and I think I can add some modification for avoiding it.

For the depth image this Fx uses its brightness as well, so with the image with only transparent and black area the Fx will fail to compute anymore since the brightness is 0 at all pixels of the depth image.

shun-iwasawa:
> Is there any benefit in letting hardness go as high as it does?

By having the large value of `Hardness` you can make the image like blur-filtered, in which the highlight is not emphasized.
I set the range of `Hardness` value from 0.05-20.0 because, again, this value is used as the exponent for computing the exposure. Which means that the higher the 'Hardness' the less effect by changing its value. I needed to set the maximum value to 20.0 ( = 1/0.05, inverse number of the minimum value ) to get enough effect.

turtletooth:
Makes sense.  If you would like to save the crash fix for later, you can go ahead and merge this.  I got it working and LGTM.  Thanks for this- it's a pretty cool effect.

shun-iwasawa:
@turtletooth Can you please merge this first? I'll push the fix in another PR soon. Thanks!

shun-iwasawa:
> you can go ahead and merge this

Oh, I was overlooking it. Then I'll merge this. Again, thank you @turtletooth for reviewing.

RodneyBaker:
Post mortem pondering of how to best use the Bokeh and Boken Ref FX...

Here's an article that discusses some use of Bokeh.
I'm posting it here mostly so that I'll do more investigation and be better able to use the FX.

Article:  [Bokeh depth of field in a single pass](http://tuxedolabs.blogspot.com/2018/05/bokeh-depth-of-field-in-single-pass.html)

PR #1564, opened 2017-11-08
Make the "Autopaint Lines" Optional in the Fill Tool

shun-iwasawa:
This PR will add a new option "AutoPaint Lines" to the Raster Fill Tool.
This option is ON by default. By disabling this option, you can cut off the autopaint behavior at all regardless of the autopaint flag of each style.

This option is demanded by some Japanese animation studio.

During ink&paint workflow, there are some cases they would like to toggle the autopaint flags.

For example, in order to paint the face properly while keeping the autopaint flags (of red & blue styles for self-colored lines) being activated, there is only a limited choice of ways like as follows:

1. Change style of intersections of red and blue lines by using the Paint Brush Tool
1. Fill a dark region of the cheek
1. Fill a light region of the cheek
1. Fill a dark region of the face
1. Fill a light region of the face

If you can temporary disable the autopaint flags easily, other approaches will become possible such as filling all the areas first then filling the lines. 
However, for now, in order to toggle the autopaint flag you need to change the current style first, which is quite troublesome.

The "AutoPaint Lines" option will act as the "main valve" of the autopaint feature so that you can switch the fill behavior on the fly.


shun-iwasawa:
Jenkins

openanim:
@shun-iwasawa 
Awesome It would be really great if this feature could also work on the vector level in the future.

masafumi-inoue:
LGTM

PR #1568, opened 2017-11-09
Shortcut Override for Tool-Specific Keys

shun-iwasawa:
For now, arrow keys (Up/Down/Right/Left) can be assigned as the command shortcut.
However, they are also defined as special command keys in some tools . For instance, in the Selection Tool arrow keys are used for moving selected region.

The command shortcuts  (processed as a `QAction` shortcut ) take priority of the tool-specific commands ( processed in `keyPressEvent` ) for now. Therefore, if any of arrow keys is assigned to some command ( such as assigning the Up arrow key for "Increase Brush Size" ),  such key cannot be used for tool-specific commands anymore.

In this PR I modified such behavior, preventing the shortcut action if the current tool will accept the pressed key.

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -441,6 +441,11 @@ return true if the method execution can have changed the current tool
     return 0;
   }  //!< Returns the type of cursor used by the tool.
 
+  // returns true if the pressed key is recognized and processed.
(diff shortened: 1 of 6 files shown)


shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa

PR #1569, opened 2017-11-09
Fix Vector Level Disappearing with Plastic Tool in OSX

### LINKED ISSUE #1543: Visibility issue after plastic skeleton is added to vector level
When using the plastic tool, once a skeleton is added to the mesh, the drawing stops being visible in the plastic tool’s animate mode and also when other tools are active. The drawing remains visible in preview and the other plastic modes. Removing the skeleton returns visibility to normal. Trying to animate with preview toggled on causes a crash. This makes animating with the plastic tool impossible. The plastic tool’s background transparency issues #856, #939, #533 and the vector mesh issue #780 are still ongoing.

This is with the latest nightly build and OS X (both 10.11.6 and 10.13)

shun-iwasawa:
This PR fixes #1543 

image

It seems to be OpenGL context related problem. I modified it by using global-shared context which is available with `Qt::AA_ShareOpenGLContexts` attribute.

### DIFF
--- a/toonz/sources/toonzlib/imagebuilders.cpp
+++ b/toonz/sources/toonzlib/imagebuilders.cpp
@@ -270,8 +270,6 @@ TImageP ImageRasterizer::build(int imFlags, void *extData) {
       TVectorRenderData rd(TTranslation(-off.x, -off.y), TRect(TPoint(0, 0), d),
                            vpalette, 0, true, true);
 
-      TGlContext oldContext = tglGetCurrentContext();
-
       // this is too slow.
       {
         QSurfaceFormat format;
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

artisteacher:
The screenshot makes it look like this might also resolve the issue with the mesh shape in OS X? That would be lovely.

turtletooth:
Meshes still seem to work fine on Windows with this.  LGTM.  Thanks @shun-iwasawa

artisteacher:
Thanks for such a quick fix! This does not seem to have an impact on the shape of the vector mesh but at least the plastic tool has returned to it's previous level of functionality.

PR #1573, opened 2017-11-13
Zoom in/out of Timeline

manongjohn:
This PR adds a zoom control allowing you to zoom in and out of the timeline.

timeline_zoom

### Zoom Control Code

The Zoom control allows for a 0-80% reduction in row width. The current max width of each row is 50px and can be scaled down to 10px.

### Zoom Control Notes

- This only applies to Timeline Mode
- Setting of the zoom factor is saved to and loaded from room configuration files.
- To change zoom factor, can use:
	- The slide bar at bottom-left of timeline pane (just below the layers area).
	- Ctrl-wheelmouse to change zoom factor.
- When the zoom factor is below 50%:
	- The cell frame numbers are replaced with a dot
	- The frame header only shows frame numbers every 5 frames when using "Frames" and 5 or 6 frames for other frame designators (SecAndFrame, etc) depending on frame rate.
- Keyframe icons cover cell dots if present.


manongjohn:
Aside from any suggestions for improvements, if anyone is up for it, I could use Zoom In/Out icons (Dimensions: 14px x 14px) for either side of the slide bar.

turtletooth:
Dude!  This is amazing.  This is what has kept me from using timeline.  Thanks so much!

[zoom.zip](https://github.com/opentoonz/opentoonz/files/1468222/zoom.zip)
Here are some icons I threw together real quick.  I'm sure someone can do better also.

bobbybutterfly:
Wow, a dream come true:D Thank you!
As an alternative to turtletooths enlargementglass i build two After Effects style hills 
[zoomSVGs.zip](https://github.com/opentoonz/opentoonz/files/1468718/zoomSVGs.zip)

toonz/sources/toonz/xsheetviewer.h
```diff
@@ -624,6 +627,11 @@ class XsheetViewer final : public QFrame, public SaveLoadQSettings {
   void drawPredefinedPath(QPainter &p, PredefinedPath which,
                           const CellPosition &pos, optional<QColor> fill,
                           optional<QColor> outline) const;
+
+  void XsheetViewer::drawPredefinedPath(QPainter &p, PredefinedPath which,
+                                        QPoint xy, optional<QColor> fill,
+                                        optional<QColor> outline) const;
```
shun-iwasawa: Travis CI claims an error as follows:

```
/home/travis/build/opentoonz/opentoonz/toonz/sources/toonz/xsheetviewer.h:631:22: error: 
      extra qualification on member 'drawPredefinedPath'
  void XsheetViewer::drawPredefinedPath(QPainter &p, PredefinedPath which,
       ~~~~~~~~~~~~~~^
```

Thank you for the great improvement !

manongjohn:
@turtletooth @bobbybutterfly Thanks for icons options. I went with the AE style hills.

@shun-iwasawa Thanks for looking at the build error. Corrected with latest commit.

beeheemooth:
Amazing! It would be great if the same could be done with the columns in the Xsheet

gab3d:
@manongjohn 
Great addition!!
Big thanks!! :)

gab3d:
@manongjohn 
in the demo video it doesn't show, so I ask:
is zoom centered around the current frame? that would be optimal for a good workflow.

manongjohn:
@gab3d Unfortunately no.   Currently it will just shift all the columns left, regardless. 

We can probably add that later as a feature request, perhaps tie it to a Ctrl-key and either using zoom control (uses current row focus) or wheel mouse (uses row under mouse pointer position).

RodneyBaker:
This is a bit more off topic so... take it for what it is worth as general discussion.
I am going to suggest a direction forward but I don't really expect such a thing to be implemented.

I wasn't going to comment just to say 'Great edition' but @gab has raised an interesting question that suggests some interesting possibilities.  By centering the focus of the zoom on the current frame there is an opportunity to at a later date change the paradigm of scrolling through timeline/xsheet giving a sense that cells are receding into the distance (z depth) at the peripheral edges while the active/current frame receives full attention (and scale).

This would not be unlike the hovering effect (bubble up effect) at the bottom of Mac's that lets the user fully see the focus of attention while hinting at other information on the periphery.  Another way of putting this would be to think of frame 1 and the last frame of the timeline all being on a circular platter with the current frame appearing larger because it is closest to the user.  In other words, there is a potential beyond that of to expand the timeline in (perceived) depth through this scaling capability.  The code would have to allow for a falloff of scaling percentages the farther away from the current/focus cell.

At any rate, this is going to be a very welcome addition as is now demonstrated.
Great addition @manongjohn!

gab3d:
@manongjohn 
no problem, I understand we should go one step at a time, and it's a great job you've made!! :)

nevertheless, I know from my experience with after effects' timeline, that you expect not to lose focus on what you are working, and if zoom makes the current frame go "out of sight", that in the end will limit zomming in/out usefulness.
But I know with a little time that will come also.

Personally, I think zooming centered around the current frame should be the default behaviour, and other possible ones done via modificator keys, like Ctrl, as you suggest.
I also find interesting the option to use the mouse pointer position as the zoom center! 👍

manongjohn:
@gab3d Apparently it wasn't as difficult as I thought it would be to get it to zoom based on the current frame you are on.

So now, by default, when using the zoom slide bar, it will zoom in/out based on the currently selected frame.  Control+wheelmouse will zoom based on whichever frame the pointer is over.  :)

gab3d:
@manongjohn 
wow! that's awesome! :)
thank you

toonz/sources/toonz/xsheetviewer.cpp
```diff
@@ -397,6 +411,18 @@ void XsheetViewer::positionSections() {
...
+  if (o->isVerticalTimeline())
+    m_frameZoomSlider->hide();
+  else {
+    if (m_frameZoomSlider->isHidden()) m_frameZoomSlider->show();
+    QRect sliderRect = QRect(0, 0, XsheetGUI::ZOOM_SLIDER_WIDTH,
+                             XsheetGUI::SCROLLBAR_WIDTH - 2)
+                           .translated(m_columnScrollArea->rect().right() -
+                                           XsheetGUI::ZOOM_SLIDER_WIDTH - 17,
+                                       m_noteScrollArea->height() +
+                                           m_columnScrollArea->height());
```
shun-iwasawa: For now the slider seems to be placed irregularly if you toggle the Xsheet toolbar to visible.
How about modifying the vertical position of `sliderRect` from 
`m_noteScrollArea->height() + m_columnScrollArea->height()` to
`height() - XsheetGUI::SCROLLBAR_WIDTH + 1`  ?

manongjohn: @shun-iwasawa Thanks for catching that. I've modified the logic per suggestion and verified it worked properly.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @manongjohn for the very nice feature!

As @beeheemooth commented, it would be nice if the Xsheet view can be zoomed out vertically as well in the future..  :smile:

bobbybutterfly:
@manongjohn this is soo awesome! Thanks a lot!

PR #1574, opened 2017-11-16
Fix Crash on Setting Negative Frame

shun-iwasawa:
In most cases, the current frame is positive. However, in some rare cases it can be negative such as users manually type the frame number in the console or move the current frame marker in the Function Curves panel.
When the preview mode is activated `Previewer` tries to obtain the frame status from `std::vector<UCHAR> Previewer::Imp::m_pbStatus[frame]` and for now if the current `frame` is negative the returned status becomes undefined and may causes crash.

I would like to keep availability of the negative frame itself because there may be a case that the animation key should starts from the negative frame in order to get ”midstream” movement at the first frame.


shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa

PR #1322, opened 2017-07-10
New Note Level

turtletooth:
This adds a note level column type to OT.  It was already there without much functionality and was only visible when a magpie file was imported.

You can enter text on any frame by double clicking.  You can move and hold frames like other frame types.  I believe @manongjohn will be doing the header UI since he is already tackling the timeline PR. Thank you @manongjohn  (I don't know if these notes should be visible in the timeline) 

This is useful for mapping dialog, noting specific actions, marking keys and breakdowns or anything you want.

notelayers

Similar requests have been made, including #325 #368

PR #1575, opened 2017-11-16
Fix Loading Note Level with Non-ASCII Text

shun-iwasawa:
This PR will fix the problem that the note level ( just added by #1322 ) with non-ASCII text fails to be loaded properly. Also refactored `levelcreatepopup.cpp` in order to clarify.

An image on the left-below is the original scene for example and the right-below is the loaded scene with the current master.

image


shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa

PR #1580, opened 2017-11-17
Fix Right Click Crash - Audio Popup

turtletooth:
If you have a scene set up like this:

crash2

And select all 3 columns and then right click in column 2 - it will crash.


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed the fix. Thanks @turtletooth ! LGTM

PR #1584, opened 2017-11-19
Remove duplicate Cell Drag Behavior options

manongjohn:
This PR removes the duplicate entries in the Cell Dragging behavior drop down on Xsheet Preferences tab.

It appears the preference was moved from the General tab to the Xsheet tab, but the code remained causing duplicates.

Other than there are duplicate entries, the duplicates need to be removed because only the 1st instance of "Cells and Column Data" actually triggers the behavior, the 2nd does nothing.


shun-iwasawa:
Wow, I realized that [Appveyor updated the Qt version from 5.9.1 to 5.9.2](https://www.appveyor.com/updates/) on November 18 . I'll follow the change before testing this.

shun-iwasawa:
Jenkins

shun-iwasawa:

PR #1588, opened 2017-11-22
Fix Crash on Launch with Flipbook

shun-iwasawa:
This PR will fix both of the following problems:

- OT crashes on launch if the Flipbook panel is embedded in the room.
- Each time you click the Pause button in console, redundant and unwanted signal-slot connections are made. (You can observe the current connection information by calling `QObject::dumpObjectInfo()` in Debug mode.)

I fixed the behavior of the Pause button, not to proceed if it is not playing.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1596, opened 2017-11-24
Only the First Click to Select All

shun-iwasawa:
This PR is a modification related to #1362 .
For now anytime you click on the numeric input fields, it causes selection of all the texts.
It is useful for inputting the numbers from scratch, but is difficult to edit by inserting numbers.

I modified such behavior, to select all the texts only when firstly clicking (i.e. clicking and focusing in) the field.


shun-iwasawa:
@turtletooth Can you please check if this behavior follows your intention? Thanks!

Jenkins

turtletooth:
Thanks for the fix!  This is great. LGTM.

PR #1597, opened 2017-11-24
Preview Range Color for the Light Style

shun-iwasawa:
This PR will set dark blue color for frame numbers in the preview range with the Light style.

image


bobbybutterfly:
i realized the difference at only at second watch. i think the blue is a little to dark

gab3d:
@shun-iwasawa 
It's great to be able to quickly see where the current preview range is!
Instead I'd better go for something like this:

range

I think is much better spotted at a glance than just a difference in text colors.

shun-iwasawa:
@bobbybutterfly Thank you for the opinion I made the text slightly vivid.
image

Also thanks @bobbybutterfly @gab3d for the suggestions. They will be another future issue as this PR is just a hot-fix for setting the text color which is already available in other styles.

turtletooth:
LGTM.  Thanks @shun-iwasawa

bobbybutterfly:
@shun-iwasawa 
I want to make a suggestion regarding ui-colors in Opentoonz:
the human ability for proper recognizing colors differs from person to person.
6% of male limited in their red and green vision.
For me it is very difficult to read the selected column text red on blue-gray.
I attached a picture to illustrate, that the contrast is very low for proper reading the text. This becomes more obvious for color-healthy people, when the picture is desaturated like in my example.
If the text would turn white instead of read, the contrast would be much bigger and better readable for all people
colorprobs

manongjohn:
I know there has been discussions on changing that red color to a different color since the connotation suggests there is something wrong with the column. We tried other colors but couldn't find one that worked with the various colors we have for different column types and styles. True White does seem to be the best option for this.

PR #1598, opened 2017-11-24
Fix filmstrip renumber issues

### LINKED ISSUE #1581: Renumbering issue
To recreate- 
- Create a level with a handful of frames:

preinsert
- Save
- Close OpenToonz (This is important)
- Reopen the scene
Insert a blank frame in the middle somewhere, a frame is lost:
inserted

I have tried to look into this but can't quite figure out what is going on.

manongjohn:
This PR fixes #1581 

There were 2 issues contributing to this issue:

1) The simplelevel->renumber() automatically updates the icon image cache so when an invalidateIcons() was called, it removed the updated image from the cache. Later when it loads, since the image was not in the cache, it went to the original level file for the image, thus loading the wrong one.

FIX: Removeed unnecessary calls to invalidateIcons() that followed a simplelevel->renumber();

2) When renumbering an image that was not cached, nothing would happen, so when loading the image later, it went to the original level file for the image and loaded the wrong one.

FIX: Attempt to pre-load the icon image before renumbering to ensure it is there.

### DIFF
--- a/toonz/sources/toonz/filmstripcommand.cpp
+++ b/toonz/sources/toonz/filmstripcommand.cpp
@@ -83,8 +83,8 @@ void makeSpaceForFids(TXshSimpleLevel *sl,
   }
   if (!touchedFids.empty()) {
     sl->renumber(fids);
-    invalidateIcons(sl, touchedFids);
-    sl->setDirtyFlag(true);
+//	invalidateIcons(sl, touchedFids);
+	sl->setDirtyFlag(true);
   }
 }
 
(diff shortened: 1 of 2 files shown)


turtletooth:
@manongjohn What do you think about just removing the commented lines?

manongjohn:
@turtletooth Done. :)  Thanks for reviewing.

turtletooth:
@shun-iwasawa I've checked this and it seems to fix the issue.  Would you mind calling Jenkins on this?  Or would you like to review it?

shun-iwasawa:
Thank you @manongjohn for the fix and thank you @turtletooth for reviewing.
It is fine if you find this seems to be OK.

Jenkins

turtletooth:
LGTM.  Thanks @manongjohn for the fix.

PR #1602, opened 2017-11-26
Skeleton Tool Tweaks

turtletooth:
I ran into an issue with the skeleton tool.  I was using a character made outside OpenToonz with really large dimensions (overall 2000px by 3000px) and the parts saved as separate pngs.  I set the character up as a skeleton rig and scaled the parent node down to 15% scale.  With this, when you select a child node with the skeleton tool, the nodes are too big to be usable:
before

This change makes a slider at the top to change the scale of the UI overlay:
after

It also makes the Skeleton Tool remember the mode of the tool between sessions and remember the status of "Show Only Active Skeleton" between sessions.

### DIFF
--- a/toonz/sources/tnztools/skeletontool.cpp
+++ b/toonz/sources/tnztools/skeletontool.cpp
@@ -54,7 +54,9 @@ using namespace SkeletonSubtools;
 
 TEnv::IntVar SkeletonGlobalKeyFrame("SkeletonToolGlobalKeyFrame", 0);
 TEnv::IntVar SkeletonInverseKinematics("SkeletonToolInverseKinematics", 0);
-
+TEnv::DoubleVar SkeletonUIScale("SkeletonUIScale", 1.0);
+TEnv::IntVar SkeletonMode("SkeletonMode", 0);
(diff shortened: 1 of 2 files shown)


turtletooth:
[bear.zip](https://github.com/opentoonz/opentoonz/files/1533861/bear.zip)

To duplicate the problem:
- parent the head to the body
- scale the body to 15%
- use the skeleton tool
- select the head
The UI is huge.  It properly deals with the body layer being scaled down, but doesn't seem to reflect that on the head layer.

shun-iwasawa:
How about trying to revert  `m_unit = getPixelSize() * sqrt(fabs(aff.det()));` in line 1443 of `skeletontool.cpp` and use `m_unit` instead of `getPixelSize()` ?
This line was already commented out from Toonz Harlequin's source.

I think we need to verify in detail (and to guess why they commented out this line) but the handles seems to become proper size.

turtletooth:
@shun-iwasawa I can confirm that this fixes the scale issue and everything seems to work well.  I don't know enough matrix math however to figure out why it was commented out in the first place.  I also discovered that the edit/animate tool also has the same scaling issue.

Issue #1603, opened 2017-11-26
Mesh in sub-xsheets crashes OT


manongjohn:
I can repeatedly crash OT (1.1.3 latest nightly; Windows) when there is a mesh in a sub-xsheet.

This happens when
1) You open a scene with a sub-xsheet that contains a mesh and the column it is linked to
2) Try to collapse a mesh and the column it is linked to

I have determined, it has something to do with loading of the Column Icons. 

When "Column Icons" preference is set to "On Demand", you can open the scene without immediately crashing.  However, if you try to select the column, it crashes.  In the second scenario, if you try to collapse the mesh, even with Columns Icons set to On Demand, it still crashes.   

I believe this is because the mesh column always loads its Column Icon whenever you select that column or the sub-xsheet that it is in.

shun-iwasawa:
@manongjohn
I've tested with all 3 types of levels but couldn't reproduce the crash yet on my either Win7 or Win10 environment. I think I miss something...
Could you please let me know if there would be any further conditions or upload some simple example files for reproducing the crash?

manongjohn:
Here is an example.

[example.zip](https://github.com/opentoonz/opentoonz/files/1504628/example.zip)

As soon as I collapse the 2 columns in the scene, it abends.

I'm on Win7 Prof.  I used a locally built OT master containing updates through 11/24 (last PR merged was # 1593)

shun-iwasawa:
@manongjohn  I could reproduce the crash. Thanks!

PR #1604, opened 2017-11-26
Allow Mesh Icon On Demand

manongjohn:
This PR adds the ability to load mesh icons on demand.

Currently if the "Column Icons" preference setting is set to "On Demand", when a scene loads the mesh columns have no icons like most other columns (excluding palettes and note levels).  However, the moment you select the mesh column, the icon immediately loads without any way to turn it off or on as desired.


turtletooth:
Does this fix #1603 ?

manongjohn:
No, but it should help with the work around until that issue can be fixed.

turtletooth:
Sounds good.  You have been doing some great stuff lately.  Thanks!

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thanks @manongjohn for this modification!

Issue #1586, opened 2017-11-20
Broken levels cause constant crashing


artisteacher:
One of my students has run into an issue where loading or accessing a specific scene causes a crash.  It seems to be a problem with the levels themselves. This started after a crash while working with the build skeleton mode of the skeleton tool with vector levels.  Looking at the levels in the browser, it looks like they are empty but there had been drawings there. The levels and scene had been saved previously with the CMD-S shortcut. This occurred with OS X 10.11.6 and OT nightly build from 8/14 (the computer has since been updated to one of the latest nightly builds). I'm sharing one of the levels and the crash report from trying to access the levels. Is there any way to recover the levels or are they lost?
[Head of Roboto.pli.zip](https://github.com/opentoonz/opentoonz/files/1488327/Head.of.Roboto.pli.zip)
[corruptlevel.txt](https://github.com/opentoonz/opentoonz/files/1488330/corruptlevel.txt)

shun-iwasawa:
This is all I could salvage.
[HeadofRobotSalvaged.zip](https://github.com/opentoonz/opentoonz/files/1489962/HeadofRobotSalvaged.zip)
![image](https://user-images.githubusercontent.com/17974955/33054163-f190dd0a-ceba-11e7-8eb0-1a49da9955f2.png)

artisteacher:
Thank you! Are there steps I can do on my end to recover some of the other levels?

Crow1992:
I encountered the exact same problem in both Windows 8.1 and Windows 10.

-Opentoonz would crash if i try scrolling through the Level Strip or if I play the animation. Levels in Level Strip display red with an error message in them ("No Icon")

However:

If I close the Level Strip, I can playback the animation just fine but if I try to X out of the program, it also gives me a crash on exit.

shun-iwasawa:
As @Crow1992 pointed, crash when opening broken Toonz Raster level may be avoidable by hiding the Level Strip if the defect exists only in thumbnail data in the file.
When I was in Studio Ghibli, in such case I tried to recover the level with the following steps:
- Open the level with the Level Strip being hidden.
- Mark a dot all the frames with Brush Tool, so that OT will update thumbnail data on saving.
- Overwrite the level.
- Try reopen the level with the Level Strip being displayed.
- If it luckily opens, delete all the dots and overwrite the level.

Crow1992:
@shun-iwasawa I've tried your suggestion. The crash still happens even after updating the levels with new data in them.
[...]

PR #1608, opened 2017-11-26
Interim temp save scene file

manongjohn:
This PR attempts to address an issue where OT crashes while trying to save the scene file (.tnz).

Currently, OT loads and saves to the same file. If the save operation is interrupted, the scene file may only be partially written, essentially corrupting it.

The changes made here include:
1. When saving the scene (i.e. scene.tnz), the save is first written to a .tmp file (i.e scene.tnz.tmp) in the same directory. Upon successfull completion, the temp file is renamed to the real scene file.

This way, if OT is interrupted while saving or the rename fails for some reason, all is not lost.

2. When loading the scene, if a .tmp file of the scene is detected, the user will be prompted to either load the last good save or stop loading.

This way the user may be able to somehow salvage changes from the failed save file.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1762,6 +1762,25 @@ bool IoCmd::loadScene(const TFilePath &path, bool updateRecentFile,
   }
   if (!TSystem::doesExistFileOrLevel(scenePath)) return false;
 
+  TFilePath scenePathTemp(scenePath.getWideString() +
+                          QString(".tmp").toStdWString());
+  if (TSystem::doesExistFileOrLevel(scenePathTemp)) {
+    QString question =
+        QObject::tr(
+            "A prior save of Scene '%1' was critically interupted. \n\
+\nA partial save file was generated and changes may be manually salvaged from '%2'.\n\
+\nDo you wish to continue loading the last good save or stop and try to salvage the prior save?")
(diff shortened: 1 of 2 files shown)


turtletooth:
This is really cool.

manongjohn:
Thanks,  Going to look at doing the same with level files as well but thought to at least start with this.

gab3d:
Great!
thanks! :)

artisteacher:
One of my students had corrupted levels recently after a crash. A fix for the levels too would be amazing.

manongjohn:
@artisteacher 
What level type? Was it a level that was brand new and not saved yet?

I was looking at level saving logic the other day and noticed that Vector levels already use an interim file and rename, if the level file was already saved. So if it was previously saved and it got corrupted during aanother save, I'm not sure why.

artisteacher:
They were vector levels and I believe they had been saved previously. #1586

manongjohn:
@shun-iwasawa @turtletooth 
Please review and merge at your convenience.  I will not be adding any further changes to this unless requested.

I will handle Level saving in a different PR because that is a lot more involved (for Toonz Raster anyway) and also want to see if/how the Skeleton Tool crash from #1586 might have corrupted the vector level.

turtletooth:
I would prefer @shun-iwasawa look at the PRs that tackle more prominent changes.  He certainly has more experience and knowledge than I do.

shun-iwasawa:
Jenkins

Issue #1131, opened 2017-04-07
toonzfarm & OpenToonz on different platforms


damies13:
toonzfarm & OpenToonz both compile for Windows, Linux & Mac OSX, however currently you must use the same platform for both your toonzfarm servers and your OpenToonz desktops.

This issue is to track the resolution to this problem as I would like to have a large toonzfarm running on linux that can be accessed by either Windows or Mac desktops.

I have started thinking about this but need to spend more time reviewing the code to work out the best solution for this issue. In the meantime if anyone has suggestions please comment.

jpturcotte:
Best of luck, @damies13! You have done a great job so far working on tfarm. It is appreciated!

damies13:
Thanks @jpturcotte, great to know my efforts are appreciated.

My initial thoughts on this were to break it into 2 phases:

1. Phase one would leave things fairly much the same, still use a common network share, but make minimal changes to the code so that executable names and paths are adjusted for each platform.

In my earlier investigations ( Issue #1070 ) I found this was partially happening already, now I found the code (or at least some of it) where this happens <https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonzfarm/tfarm/tfarmtask.cpp> (around Lines 360-365). I notice that code for linux is missing here, but I can fix this easily.

2. Phase two would involve a bigger rewrite of the network protocol between the various components to change the way the data is sent, to enable a variety of features
- Remove the need for a common network share, instead send the project and resources through the network protocol. I still need to investigate the best way to do this.
- Auto configuration of servers and controller, at the moment this is configured by text files on the common network share, take a server offline and it all stops working :( What I would like to happen is you start a server and it registers itself with the controller, stop it and it de-registers itself, Start a task on the controller, the controller assigns work to various servers as available, new server comes on line and some work is redistributed to it, sever goes offline (intentionally or not) and it's work is reallocated.
- multiple farms, this is more a client side change, but I would like to enable a client to change farm configurations to choose between multiple farms.

more ideas coming....

damies13:
I have started working on this issue, but it's no where ready for a pull request for now and after easter my work is going to get busy for for a few months. If the pull requests i have submitted so far are combined there should be enough to get an OpenToonz Farm running on any platform, as long as all parts of the farm use the same platform.

OpentoonzTester:
I am not a coder, so please forgive me if this is not workable.

[...]

PR #1609, opened 2017-11-26
Issue #1131 toonz farm & open toonz on different platforms 

damies13:
This PR doesn't completely fix Issue #1131, however combined with PR #1141 the as long as you can find a common network file path then it will enable cross platform farms.

- Without PR #1141, this PR ensures cross platform operation for Windows & Linux. 
- PR #1141 is required for Mac OSX as the tfarm binaries are not currently included in the OSX Application package (OpenToonz_1.1.app)


damies13:
it seems I don't have this right yet, will reopen after more testing...

damies13:
Ok this PR should now be complete.

Added bonus, I noticed that the current version on OSX using batch render in local mode also doesn't work, but combining this PR and PR #1141 should fix that too.

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed this fix will enable batch rendering on OSX, in combination with #1993 and #1994 .
Thank you @damies13 for the fix ! LGTM

PR #1589, opened 2017-11-22
(WIP) Translation korean

haminu0108:
Translate toonz.ts in Korean

PR #1610, opened 2017-11-27
Update .ts Files for All Languages

shun-iwasawa:
This PR applied `lupdate` for all languages and update all .ts files, in anticipation of the next version release.
Please note that .qm files are left unchanged.

Also added `korean` language option as @haminu0108 started Korean translation in #1589 .

@gab3d  @jcome @morevnaproject 
I believe that it will just add new untranslated texts and should not remove the existing translations.
Just in case could you check briefly if your translations are still there? 

@haminu0108 
It seems that in #1589 you copied German .ts files and overwrote them. Instead, can you please use files included in this PR for the future? Thanks!

### DIFF
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -540,7 +540,7 @@ if(BUILD_ENV_MSVC AND MSVC_VERSION EQUAL 1800)
 endif()
 
 # generate Qt translations and messages
-set(LANGUAGES japanese italian french spanish chinese german russian)
+set(LANGUAGES japanese italian french spanish chinese german russian korean)
 
 function(add_translation module)
     set(translation)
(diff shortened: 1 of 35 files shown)


gab3d:
@shun-iwasawa everything seems to be in good shape (I see some previously translated terms are now marked as "vanished", but appart from that, everything else looks ok.
Thank you.

gab3d:
I've noticed the "Drawing:" string in the Apply Lip Sync Data dialog is not translatable.

jcome:
> Just in case could you check briefly if your translations are still there?

@shun-iwasawa Thank you for the effort, I confirm Chinese translations are still there. 
I get 478 new strings to translate as my weekend homework :) 

Regards

shun-iwasawa:
@gab3d @jcome Thank you for the confirmation.
As the features and the UI are improved rapidly these days, I'll push another PR for .TS update soon.
Sorry for your trouble. Merging this.

PR #1611, opened 2017-11-27
Prevent Crash on Plastic-deformed Sub-xsheet Icon

shun-iwasawa:
This PR is a quick workaround for the issue #1603 .
For now generating icons of plastic-deformed image causes crash with message `Attempting to create QWindow-based QOffscreenSurface outside the gui thread` due to `QOffscreenSurface` is created outside the gui thread. As a quick (and dirty) workaround, I made the `RasterPainter` to ignore the deformation if this is called from non-gui thread (i.e. icon generator thread).


shun-iwasawa:
Jenkins

turtletooth:

PR #1612, opened 2017-11-27
Fix Lip Sync Popup

### LINKED ISSUE #1577: Issue with applying lip sync data to wrong column type
I love the new addition of applying lip sync data! Thanks @turtletooth!

However, at least in OS X, if you accidentally try to apply lip sync data to the wrong column - ie. an audio column, a warning pop-up notifies that you can't apply lip sync data to that column type. Once you press "ok", the lip sync window opens anyways. After closing out of the window it is impossible to do any actions in OT. Trying to select a tool, change columns or etc. results in a warning sound but no actual action. The menu is also greyed out. At that point the only way to make OT functional again is to restart.

shun-iwasawa:
This will fix #1577 .
If any invalid column is selected, `hide()` is called in `paintEvent()` , followed by `raise()` and `activateWindow()` . It seems to occur malfunction of window modality.


shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa  for the fix.

PR #1613, opened 2017-11-28
Remove Hyphens in Fx Parameter Expression

shun-iwasawa:
This PR solves the following problem:
 
`On-Focus Distance` parameter in the `Bokeh Iwa` and `Bokeh Ref Iwa` Fxs cannot be referred by expression, since the expression `fx.bokehrefiwa01.on-focusdistance` is recognized as separated at hyphen.

I fixed it by removing hyphens from the parameter name when obtaining the text for expression variable.
Now `On-Focus Distance` can be referred with `fx.bokehrefiwa01.onfocusdistance` .


shun-iwasawa:
Jenkins

turtletooth:
Tested and confirmed.  Thanks @shun-iwasawa

PR #1617, opened 2017-11-29
New Timeline Room

turtletooth:
With the great changes made to the timeline, I would like to propose a new room that be the default room on startup:

new_room


gab3d:
It'd be nice to have a room with the timeline, not sure if as default room though.

...but more important: anybody please do something with this mammoth in the bazaar, that is the Style Editor! :D
there seems to be no place where to fit it comfortably

bobbybutterfly:
room
very good idea!
But for Animation the Style editor is not so important.
For me, the Function Curve Editor in combination with the Chanels on the side is way more important for my Animation workflow.
Sad, that the Function Curve Editor is not dockable.
It would be great to toggle between Function Curve Editor and Function Editor

Agnyy:
@bobbybutterfly  
Your request will be implemented in the Timeline itself.  
change # 14 https://docs.google.com/document/d/1CyVbJ6dlhFC-IjEIcG52RoUt45PQ6dvDvEO7eL5GQcw/edit 


Here is a discussion on the development of the timeline. #1422

bobbybutterfly:
@Agnyy  wow, how cool is that :D

beeheemooth:
Sorry, I'm not against the new room, but I think the priority should remain for X-sheet.

shun-iwasawa:
I think it would better to re-organize the `Default` room-set based on expected workflow for first-time users, rather than just adding the new room.
And IMO the first room should still be `Basic` which includes some reasonable panels for them.

turtletooth:
Thanks for the feedback everyone.  I'm going to close this.

PR #1618, opened 2017-11-29
Note level zoom out update

manongjohn:
This PR updates the zoom out behavior for Note levels similar to other levels.

This does the following:

1) Replaces the text with a frame dot when zoomed out more than 50% like other levels.
2) Zoomed out or not, the tool tip for the cell displays the contents of the cell instead of the frame number.

image

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -1919,13 +1919,15 @@ void CellArea::drawSoundTextCell(QPainter &p, int row, int col) {
   QString text =
       cell.getSoundTextLevel()->getFrameText(cell.m_frameId.getNumber() - 1);
 
+  QFontMetrics metric(font);
+
   if (text.size() && text != "" && m_viewer->getFrameZoomFactor() <= 50) {
-    drawFrameDot(p, xy, false);
-    return;
+    text          = text[0];
+    int charWidth = metric.width(text, 1);
+    if (charWidth > nameRect.width()) nameRect.adjust(-2, 0, 4, 0);
   }
 
 #if QT_VERSION >= 0x050500
-  QFontMetrics metric(font);
   QString elidaName = elideText(text, metric, nameRect.width(), "~");
 #else
   QString elidaName   = elideText(text, font, nameRect.width(), "~");
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -1561,6 +1561,17 @@ void CellArea::drawCurrentTimeIndicator(QPainter &p, const QPoint &xy) {
   p.drawLine(cellMid, cellTop, cellMid, cellBottom);
 }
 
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
> Replaces the text with a frame dot when zoomed out more than 50% like other levels.

I wonder if there may be a case that showing the first letters in the note level frames is enough informative.
For instance, using the note level for marking dialog timings like as follows:
image

manongjohn:
@shun-iwasawa 
I like the suggestion of just the 1st char.

Problem is when you are zoomed out all the way. The text is too big to fit unless I scale it down. Let me see what I can do.

manongjohn:
@shun-iwasawa
Thanks again for the great suggestion!

I've modified the logic to only show the 1st character once zoomed out more than 50%.

image

I am currently not adjusting font size and just adjusting the position of the name rectangle to avoid clipping but it is still possible depending on how wide the character actually is.

toonz/sources/toonz/xshcellviewer.cpp
```diff
@@ -1921,11 +1921,8 @@ void CellArea::drawSoundTextCell(QPainter &p, int row, int col) {
 
   QFontMetrics metric(font);
 
-  if (text.size() && text != "" && m_viewer->getFrameZoomFactor() <= 50) {
-    text          = text[0];
-    int charWidth = metric.width(text, 1);
-    if (charWidth > nameRect.width()) nameRect.adjust(-2, 0, 4, 0);
-  }
+  int charWidth = metric.width(text, 1);
+  if ((charWidth * 2) > nameRect.width()) nameRect.adjust(-2, 0, 4, 0);
```
turtletooth: What do you think of something like this?
```
QFontMetrics metric(font);
if (text.size() && text != "" && m_viewer->getFrameZoomFactor() <= 50) {
	  if (m_viewer->getFrameZoomFactor() <= 20) text = text[0];
	  int charWidth = metric.width(text, 1);
	  if (charWidth > nameRect.width()) nameRect.adjust(-2, 0, 4, 0);
  }

  if (!sameLevel || prevCell.m_frameId != cell.m_frameId)
    p.drawText(nameRect, Qt::AlignLeft | Qt::AlignBottom, text);
```

I did two main things: 
- got rid of the elideText call - I think it is more useful to see part of a letter than a tilde(~).
- If the cell is a certain width, go back to just having one letter.  

This is just my preference.  But I thought I'd throw it out there.

manongjohn: Not sure I agree with taking out the elideText call entirely.

That tilde indicates there is more text that isn't displayed. Depending on where its clipped, it may look like there isn't any more text. At least with that visual indicator, you know to hover over it to get the full text if you need it.

turtletooth: I see your point.   I'm fine with it this way too.  I have reviewed this and it works fine.  If @shun-iwasawa wants to call Jenkins on this, this is ready.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @manongjohn for the modification, also than you @turtletooth for reviewing this!

PR #1619, opened 2017-11-29
Fix Rename Cell in Empty Column Crash

turtletooth:
If you try and rename a cell in an empty column, OpenToonz will crash.  This fixes it.

To test:
- Double click a cell
- Enter text
- Hit Enter


shun-iwasawa:
@turtletooth 
Confirmed the crash. Thank you for finding it!
Regarding the fix, I pushed an another approach which retains availability to insert a level into empty column. Such operation was available in the previous versions before introducing the note level.
Could you please take a look and check if it is OK for you?

Issue #1579, opened 2017-11-17
New touch controls incompatible with trackpad


artisteacher:
This may be specific to OS X (10.11.6 and 10.13.1) but the new touch controls make it impossible to make any changes/adjustments to drawings in the viewer for computers with a trackpad - unless using a drawing tablet or mouse. The tools do nothing - though sometimes clicking quickly three times in a row might trigger a fill or selection or text. It is possible that this is related to tablet use. After connecting and disconnecting a tablet I temporarily was able to use the tools in the viewer as expected. However that does not consistently resolve the issue. 

Also, the 1 finger panning keeps getting triggered unintentionally (since normal navigation is done with a single finger with the trackpad). The panning would actually be really great if there was a different trigger.

shun-iwasawa:
You means that you are using the trackpad as a normal pointing device like mouse?
Would you like to have an option for disabling the touch gesture ( just added in #1437 ) ?

artisteacher:
Yes, I am using the trackpad like a mouse. I’d rather keep the touch controls, have panning changed to multiple fingers and have a fix for the tools being disabled. If that’s not possible, then there should definitely be an option to disable the touch gestures.

artisteacher:
This may be related, but at this point I am only able to adjust or animate the plastic tool's skeleton with a mouse.

turtletooth:
Closing- fixed in the latest build.

PR #1621, opened 2017-11-30
touch control optional

shun-iwasawa:
#1579 
This PR will make the touch control optional in the preferences (added to "Drawing" category).
With this option being disabled `SceneViewer::gestureEvent()` and `SceneViewer::touchEvent()` will not be called and some controls will be caught by mouse event instead.
As there is a case that the touch control doesn't work properly with trackpad, I set this option to `false` (disabled) by default in OSX. 
Sorry @turtletooth for not notifying it earlier.

### DIFF
--- a/stuff/profiles/layouts/rooms/Default/menubar_template.xml
+++ b/stuff/profiles/layouts/rooms/Default/menubar_template.xml
@@ -75,6 +75,8 @@
     <command>MI_BringForward</command>
     <command>MI_SendBack</command>
     <command>MI_SendBackward</command>
+    <separator/>
+    <command>MI_TouchGestureControl</command>
   </menu>
   <menu title="Scan &amp;&amp; Cleanup">
     <command>MI_DefineScanner</command>
--- a/stuff/profiles/layouts/rooms/StudioGhibli/room3_menubar.xml
+++ b/stuff/profiles/layouts/rooms/StudioGhibli/room3_menubar.xml
@@ -74,6 +74,8 @@
     <command>MI_ResetShift</command>
     <separator/>
     <command>MI_RasterizePli</command>
+    <separator/>
+    <command>MI_TouchGestureControl</command>
   </menu>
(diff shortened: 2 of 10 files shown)


shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa Would you be willing to make this a command that can be toggled with a shortcut?  I can see instances where it would be nice to use touch to pan, but then use the selection tool, animate tool, or control point editor with touch.

shun-iwasawa:
OK, let me see.

artisteacher:
Thanks for resolving this so quickly! A shortcut would be great, I do really appreciate some of the touch gestures. Also, is it possible that it’s the one finger pan specifically that’s stopping the tools from being functional with a mac touch pad - since one finger is used to trigger normal actions on the trackpad? Could there be a preference where people could choose between 1 or 2 fingers for panning?

shun-iwasawa:
Moved the option from the preferences to toggle menu in the menu bar ( in Edit menu in the Default menu bar ). Now it can be assigned shortcut key.

Jenkins again

ernesttx:
Thank you for making this optional. I am using a Surface Book and the pen starts drawing and then about 2 seconds in, it switches to 'pan' the drawing. So, it is not specifically a touchpad issue. 
When I use the button on the pen, I can draw without any issues. But, as you can imagine, drawing this way is quite annoying. I'll test out and let you know. And just for the record, I commend everyone involved with development for doing a fantastic job of bringing features and fixes to this software. Thank you.

shun-iwasawa:
@ernesttx 
I disabled the touch control by default on Windows, in order to avoid occurrence of the problem on a new user.

shun-iwasawa:
Jenkins again

turtletooth:
Tested and LGTM.  This can be merged when Jenkins finishes.

turtletooth:
Thanks @shun-iwasawa for this!

Issue #1502, opened 2017-10-02
Default Style Editor Wheel very small


gab3d:
The default size of the Style Editor wheel on a 1920x1080 screen in the Basics Room is really too small to be useful.
![default_styleed_size](https://user-images.githubusercontent.com/18104091/31094389-6ae7a942-a78b-11e7-85bf-5d15d2806b41.png)

yes, perhaps the color wheel design is not the most suitable one for small spaces, but a big part of the problem seems to lay in the fact that both, the sliders and the bottom areas are bigger that they could be, and are not resizable.
![default_styleed_size2](https://user-images.githubusercontent.com/18104091/31094582-17fe872c-a78c-11e7-8e6e-6835b83cc9cd.png)

the bottom part is clearly the one that is wasting more space, in my opinion (it takes a lot of screen real estate for things that could be placed in a lot less room).

artisteacher:
I think the resizing was limited so that the text doesn’t get clipped.  I think #1397 will help with this issue.

gab3d:
@artisteacher, yes, it'd help, but there's still room for saving pixels here and there, like this exercise I just did in Photoshop, proves.

![compact_style_ed](https://user-images.githubusercontent.com/18104091/31103389-0972b1b2-a7ad-11e7-916a-08118fc3654a.png)

gab3d:
I've come again over this issue: The big screen space currently needed for the Style Editor.
Everytime I try to accomodate panels in "a better than default" way I stumble against the unreasonable big space this editor takes for being usable, and I see there are parts (as the bottom part, specially) that takes a LOT of space for things that realy should not.

In this mockup I've managed to reduce its footprint even more by moving the informative text to the title bar (as in other editors like the Visor or Xsheet currently in OT).

If any coder here could dedicate some working hours to this, it could make a big difference in everyday's usability of the UI.

![style_editor_compact](https://user-images.githubusercontent.com/18104091/33281857-6fb115f0-d385-11e7-8a4b-afd446fbdb25.jpg)

artisteacher:
If it’s possible, I think the settings toggle should go in the tab area to the very right. I think the @gab3d’s change for the auto/apply buttons would be a fantastic change. My work computer has a particularly small screen, so I definitely see the benefit. I wonder if the two swatches at the bottom could move into the end of the above row?

artisteacher:
While the wheel can still be very small, #1622 has really helped make the style editor usable on a small screen.

[...]

PR #1622, opened 2017-11-30
Reduce Footprint of Style Editor

turtletooth:
This is for #1502 

This moves the status to the title bar of the style editor - in keeping with other panels.
The apply button is moved next to the auto apply button.
The toggle buttons are moved to the bottom of the editor.

style_small


gab3d:
ohh... man
that's awesome! 🎉 
you made my day 😋

Agnyy:
I propose to do so.
1) the settings should be higher.
2) The bottom bar stole a lot of space, so it must be removed.

shun-iwasawa:
> The bottom bar stole a lot of space, so it must be removed.

The bottom bar is not always empty, but used for styles containing multiple color settings such as colored cleanup styles and some special vector styles.

gab3d:
@Agnyy 
I see two problems with what you propose:
1. Even if I "visually" like the location of the two settings buttons, the zone where you propose to place them does not fall inside the area what changes when you change tabs (that area is everything below the tabs area).
2. If you propose to remove the bottom bar (which I'm not against of) you must propose where to locate the various color swatches that usually show in this bottom bar. In your mockup you currently don't show this.

openanim:
Sorry but I don't see the benefit of this set up. It requires more clicks than before when you had direct access to different modes. The size problems are still there despite the layout of the space, the reason why I don't understand. Unless I have not understood everything.

gab3d:
@openanim it's true 1 more click is now required to toggle different slider sets or wheel.
On the other hand we need to consider how often things are used, to see what's the impact of each thing.
1. a panel which size is big and cannot be reduced without losing features.
2. a set of buttons to toggle the visualization of sliders/etc that is seldom used anyways.

To me the answer is clear: 
The big footprint penalty is suffered everyday, everytime a user needs to select a color and that panel occupies that much dpace of the UI.
The extra click needed to hide sliders would rarely be suffered, just by a user wanting to hide sone set of sliders (that, what orher reasin vould that user possibly have to want to hide a set of aliders, than to reduce the footprint of the panel?).
So... in all the scenarios I imagine, the reduced footprint is a winning situation, even at the cost of having to make 1 more click to hide the sliders fron the panel)

artisteacher:
Personally I agree with @gab3d that the extra click is worth the smaller footprint. I often have to use a computer with a small screen. Once I’ve adjusted the sliders to make the style editor fit better on that small screen, I leave the settings alone. At least for me, the extra click really wouldn’t impact my workflow.

32f3086a-853e-4445-a792-a55d47bdde5c

turtletooth:
I'm not a big fan of putting the color samples in the wheel area simply because if you change to a vector style, you are currently given an immediate visual cue that there are more colors to change.  If they are hidden in the wheel, a viewer might switch the vector style but not understand why it's not working right.

artisteacher:
In that case, what about putting the color samples to the side of the color swatches?

openanim:
Another problem for me, When we reduce the style editor panel, the cursor for the options is grayed out, we must readjust the panel again to access it.

adamearle:
Perhaps the option to choose between a square and circular wheel will also help reduce the size of the UI. the "hue value" slider could be easily added in instead of a color picker or a switch feature for fine-tuning.

Personally, a zoom feature into the color picker would be awesome so fine tunning was easier. normally when picking a color you start you search range rather large and then it progressively becomes
 smaller and smaller.

Just a thought.

Ad ;D

gab3d:
@openanim as I see it, the problem with the Options tab is deeper than what you expose. The bigest problem with the Options tab is that there's no good way of having it open at the same time you are switching brushes. This is specially evident when trying to access the MyPaint brushes, along with their options.
I've tried opening more than one Style editor, one showing the Options tab in front, another one showing the MyPaint brushes, and a third one showing the color wheel.
The problem with that setup is that:
1. it takes a LOT of screen space, as all of the versions of the editor show the bottom part (auto, apply buttons, swatches, etc)
2. when you try to make a new room with all these panels docked, and the open OT again, all of the panels have the same tab (Color I guess) selected.

So, is clear that there currently is no planned workflow in OT for working with all of these panels exposed at the same time.

To me it'd be good to rethink it a little bit, by allowing brush Options and Color picker to live in its own separate panels.

Also, perhaps in the long run, the brush Options panel should turn into a generic " Selected Item" Options panel, to hold options for whatever is currently selected in OT. But that would naturally take a lot of work to do and exceeds by far the discussion here.

adamearle:
Wonder if we start using something pie menu system or a pop-up window system in the viewport if this would help take care of the overuse of UI real estate. 

What is some more possible way that we could do this? krita color and brush pop-up well is one of the best I have ever used.

If I am using photoshop (not much these days because opensource) pretty much just work in full-screen mode and access everything via a hotkey ie f5 and f7

So how can we minimise the space? 
Floating windows?

Ad

turtletooth:
@openanim I fixed the resize issue.  Thanks for pointing it out.  I realize that the options are one extra click, but I think it is worth the space saved.  I would wager that most people set the options they prefer and just leave it.  I'm a color wheel and alpha person myself.

@artisteacher Thanks for the suggestion for the color samples, but there are some vector styles that have as many as 4 (I think) color samples, so I'm not sure there is a better place for them right now.

Either way.  This is as far as I have time to take this change for now.  It saves three rows worth of space- which I think will really help those with smaller screens.

gab3d:
@turtletooth 
Thank you so much!
Your work was indeed much effective and appreciated. 👍

gab3d:
Could this improvement be revised for inclusion?
Thanks!!

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @turtletooth for the improvement!

PR #1625, opened 2017-11-30
Fix Styles of the Message Center

shun-iwasawa:
This PR fixes the styles of Message Center panel in order to make the texts visible.
Also fixes the text color of suggestion list popup in the expression field.

The Message Center is relatively unknown to users for now, but I think it has some potential in the future..


shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa 
Going to submit a PR to change the word "infos" to "info" - even though it is plural, the word "infos" is awkward  and info is more easily understood.

PR #1626, opened 2017-11-30
Fix "Replace Parent Directory" Command Fails on the First Call

shun-iwasawa:

"Replace Parent Directory" always fails when you use the command for the first time after launching the software. It fails with an error message `Nothing to replace: no cells or columns selected` even if you surely select cells or columns before using this command.
It works properly from the second use of this command.

This problem occurs because the current selection is cleared on constructing the File Browser. In this PR I made a command handler specially used for this command and obtain the selection information before making the browser window.

### DIFF
--- a/toonz/sources/toonz/filebrowserpopup.cpp
+++ b/toonz/sources/toonz/filebrowserpopup.cpp
@@ -1859,29 +1859,63 @@ void LoadColorModelPopup::showEvent(QShowEvent *e) {
 //=============================================================================
 /*! replace the parent folder path of the levels in the selected cells
 */
+
+class OpenReplaceParentDirectoryPopupHandler final : public MenuItemHandler {
+  ReplaceParentDirectoryPopup *m_popup;
+
+public:
+  OpenReplaceParentDirectoryPopupHandler()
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
Jenkins

turtletooth:
I've tried to test this, but I don't actually know how the command "Replace Parent Directory" works.  The code seems fine, but if you'd like me to test the functionality- please let me know the steps to do.  I've looked in the online manual and parent shows up 5 times- but not in this context.

shun-iwasawa:
Imagine that all the levels used in some scene are stored in one folder.
There is a case that you manually move the folder to another place for organizing files. In such case, OT cannot find and load the levels anymore since they are saved as absolute paths in the scene. To re-link the levels you need to change the path in the level settings to locate the proper path. However, it may become troublesome if there are a lot of levels.

The "Replace Parent Directory" command will work to all the levels under the selected cells. You can change the paths of all of them by replacing the parent folder by the specified one at the same time.

turtletooth:
Thanks @shun-iwasawa - I've tested and confirmed the fix.  LGTM.

PR #1627, opened 2017-11-30
Clarify some wording

turtletooth:
This changes "infos" in the message center to "info" - informations is only a word when referring to legal filings.

Changes the dropdown label for the theme to say "theme" instead of style.  Theme is more commonly used for UI schemes.


shun-iwasawa:
LGTM Thank you @turtletooth for the modifications!
No need to run Jen.kins as it is just a changing of labels.

PR #1628, opened 2017-11-30
Tiff make call from cmake

Houkime:
Cmake runs make for TIFF if it is not present and is on Unixlike.
Reduces steps needed to compile from source.
Works for Arch (c)  
Untested for others.
(Well actually Arch users don't need that because of AUR, but other distros might need)


ideasman42:
Would prefer not to apply this, having CMake call into configure scripts makes it not very easy to debug build issues.

It's likely we will eventually use the systems libtiff too, so this wont be needed.

Houkime:
Well, using system's libtiff shouldn't be too hard for Linux builds. 
Will look into this shortly.

xenogenesi:
> Well, using system's libtiff shouldn't be too hard for Linux builds.

Actually may be a bit more complicate than expected.

I'm using pbuilder and pdebuild to build a deb package of OpenToonz for debian/sid (unstable), it works but I've to use the libtiff from the thirdparty/sdk directory, because it appears to have some custom patch, at least some symbol is missing using the system tiff:

    /build/opentoonz-1.2.1/toonz/sources/image/tif/tiio_tif.cpp:465:19: error: 'TIFFReadRGBATile_64' was not declared in this scope
    /build/opentoonz-1.2.1/toonz/sources/image/tif/tiio_tif.cpp:480:16: error: 'TIFFReadRGBAStrip_64' was not declared in this scope

Looks to me like the OpenToonz's libtiff add at least the file `tif_getimage_64.c` to handle 64bit per pixel images (where each rgba component get 16bit/64k values). I didn't look if there are more custom patches.

The latest version of libtiff for debian/unstable (at least with mirrors I'm using today) is `libtiff5-dev:amd64=4.0.9-6` and looks to me like is the last official release [gitlab/libtiff](https://gitlab.com/libtiff/libtiff)

RodneyBaker:
Closing as possibly unnecessary.
Can be reopened as needed.

PR #1642, opened 2017-12-03
Add project name to startup

### LINKED ISSUE #1553: [Feature Request] Improve Startup box: Scene association with projects
Consider that I've used OpenToonz for more than one project now, I actually realized that (due to my own bad habits when it comes to naming things) when I try to name scenes according to shot numberig in a storyboard for any given project, there are likely to be two different scenes with the same names for different projects. Trying to figure out which comes from which project can can be difficult, unless we just close the box and opened the scenes the old fashioned way. 

I was wondering if it's possible to improve this dialogue box so as to allow users to easily distinguish which scene belongs to which project? 

One idea may be to change the scenes shown in "Open Scene" depending on which scene is selected in "Choose Project", at least until a "Show All" checkbox is clicked. 

Another idea may be to simply "show" either besides or below each scene which projects they might belong to. 

This feature request may not be a priority, so it's okay to keep this in the back-burner and implement it when deciding on a larger, more necessary design / feature that may be discussed in the future when it comes to the startup box. I just thought I'd put this idea here either way, because I thought it may be helpful. 

Till then, I'm fine with myself just using better naming conventions. 

selection_849

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/51123720-feature-request-improve-startup-box-scene-association-with-projects?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

turtletooth:
fixes #1553 

This adds the project name to the end of the scene name in the startup popup.

scene name - project name

Note: there is already a tool tip if you hover.  This just makes it easier to quickly scan the list.

### DIFF
--- a/toonz/sources/toonz/startuppopup.cpp
+++ b/toonz/sources/toonz/startuppopup.cpp
@@ -416,7 +416,7 @@ void StartupPopup::refreshRecentScenes() {
               ->getProjectNameByScenePath(TFilePath(name))
               .getName());
       if (projectName.length() > 0)
-        justName             = justName + " (" + projectName + ")";
+        justName             = justName + " - " + projectName;
       m_recentNamesLabels[i] = new StartupLabel(justName, this, i);
       m_recentNamesLabels[i]->setToolTip(
           name.remove(0, name.indexOf(" ") + 1));  // remove "#. " prefix
--- a/toonz/sources/toonz/startuppopup.cpp
+++ b/toonz/sources/toonz/startuppopup.cpp
@@ -398,7 +398,8 @@ void StartupPopup::refreshRecentScenes() {
   removeAll(m_recentSceneLay);
 
   m_sceneNames.clear();
-  m_sceneNames = RecentFiles::instance()->getFilesNameList(RecentFiles::Scene);
+  m_sceneNames =
+      RecentFiles::instance()->getFilesNameList(RecentFiles::Scene, false);
   m_recentNamesLabels.clear();
   m_recentNamesLabels = QVector<StartupLabel *>(m_sceneNames.count());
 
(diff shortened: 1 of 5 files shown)


DarrenTAnims:
It'd be cool if the project name appeared in the top of the application, next to the scene name too, in the same format as in that dialog.

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -409,7 +410,13 @@ void StartupPopup::refreshRecentScenes() {
...
+      QString justName    = QString::fromStdString(TFilePath(name).getName());
+      QString projectName = QString::fromStdString(
+          TProjectManager::instance()
+              ->getProjectNameByScenePath(TFilePath(name))
+              .getName());
+      if (projectName.length() > 0)
+        justName             = justName + " - " + projectName;
       m_recentNamesLabels[i] = new StartupLabel(justName, this, i);
       m_recentNamesLabels[i]->setToolTip(
           name.remove(0, name.indexOf(" ") + 1));  // remove "#. " prefix
```
shun-iwasawa: This PR introduces the way to get the file name list without number prefixes. So you can use `name` for the tooltip instead of `name.remove(0, name.indexOf(" ") + 1)` .

toonz/sources/toonz/startuppopup.cpp
```diff
@@ -409,7 +410,13 @@ void StartupPopup::refreshRecentScenes() {
     int i = 0;
     for (QString name : m_sceneNames) {
       if (i > 9) break;  // box can hold 10 scenes
-      QString justName = QString::fromStdString(TFilePath(name).getName());
+      QString justName    = QString::fromStdString(TFilePath(name).getName());
+      QString projectName = QString::fromStdString(
+          TProjectManager::instance()
+              ->getProjectNameByScenePath(TFilePath(name))
+              .getName());
```
shun-iwasawa: I'm bit concern about that searching of project names for all 10 recent scene files may be time-consuming, especially if the files are stored in remote file server.
Instead, how about saving the project names along with the scene paths into the `RecentFiles.ini` for this purpose?

PR #1644, opened 2017-12-04
Timeline Lock On icon correction

manongjohn:
This PR corrects the timeline's Lock ON icon size to match the other timeline ON state icons.


turtletooth:
LGTM.  Thanks for the fix @manongjohn

PR #1645, opened 2017-12-04
Fix Actual Pixel Size and Zoom Factor Reporting

turtletooth:
traster

The above picture is a Toonz Raster layer at 1920 x 1080 resolution on a 1080p screen.  Notice the zoom factor says 100%, actual 100% display is reported as 225%.

The default return on getDpiFactor is currently 1 which worked well before the introduction of Stage::standardDpi.  

Turning on "Enable Actual Pixel View on Scene Editing Mode" in preferences fixes this for Toonz Raster and Raster levels, but the problem persists on vector levels.  I realize that vector levels have no actual dpi, but I think it is reasonable for a user to expect clicking actual pixel size to show the output size based on the camera settings.


shun-iwasawa:
Jenkins

shun-iwasawa:
It's more intuitive. Thank you @turtletooth ! LGTM

PR #1647, opened 2017-12-04
Fix Rename Cell in Empty Column Crash (modified)

shun-iwasawa:
This will fix the crash issue mentioned in #1619 but in another approach, retaining the availability to insert a level into empty column.
Also I moved the renaming process for note levels to separate function to clarify.
Thank you @turtletooth for finding the problem!


shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa 
Do you think that locked text columns shouldn't be changeable?  It wouldn't be a hard PR.

PR #1651, opened 2017-12-05
Fix Animate Tool Control with Tool Options

### LINKED ISSUE #1646: Click + drag to edit values for the Animate Tool only works on one column
I've run into a problem in OS X 10.11.6 and 10.12.1 where clicking and dragging on the label to edit values for the Animate tool only works on the first column where clicking and dragging is attempted. After changing the values on one column and then trying to change the values on a second column, the original column incorrectly accepts those changes and the new column ignores those changes to the value. Manually typing the value continues to work as expected.

shun-iwasawa:
fix #1646

In #1362 `PegbarChannelField` started to hold `TStageObjectValues m_before` to keep the values before dragging for undo.
However, since `m_before` is not initialized on each calling of `onChange()`, only the first object you changed will be stored in `TStageObjectValues::m_objectId` . (when `TStageObjectValues::updateValues()` is called, `m_objectId` will be set if it is not specified)

`m_before` is copied to `after`, and then the value is set to the wrong object when `after.applyValues()` is called.

I added a line to initialize the `TStageObjectValues` each time using `m_before`.

### DIFF
--- a/toonz/sources/tnztools/tooloptionscontrols.cpp
+++ b/toonz/sources/tnztools/tooloptionscontrols.cpp
@@ -1215,6 +1215,7 @@ void PegbarChannelField::onChange(TMeasuredValue *fld, bool addToUndo) {
   // and only for the first drag
   // This should always fire if addToUndo is true
   if (!m_firstMouseDrag) {
+    m_before = TStageObjectValues();
     m_before.setFrameHandle(m_frameHandle);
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa  for the fast fix.

PR #1640, opened 2017-12-03
Issue #1639 error:ordered comparison

damies13:
This is a fix for Issue #1639

PR #1663, opened 2017-12-08
Issue#1639 error ordered comparison try2

damies13:
This PR is essentially the same as PR #1640, just I tried to squash everything before creating the PR, not sure if I got this right? @shun-iwasawa keep which ever is better for you and we can close the other one.


shun-iwasawa:
The part you fixed was indeed mistake - it seems that all the process goes to `else` condition for now.
However, introducing this PR causes a problem as follows:

- Prepare two levels with palettes, one (palette `A`) has only two default styles (# 0 and # 1) and the other (palette `B`) has multiple styles like as follows:
image
- Select and copy styles # 4 and # 5 of the palette `B` and paste them to the palette `A` .

On my end it fails to paste styles properly.

The former condition seems to be the way to set the pasted styles the same style id as the copied ones.
However for now the palette not allows to have discrete style ids (like 0, 1, 4, 5) which causes malfunction. 
I think we need to investigate further regarding this fix.

damies13:
@shun-iwasawa ,

I've had a bit more of a look at the code here, the only place I could see a getStylePage() defined was at 372 of "toonz/sources/common/tvrender/tpalette.cpp" but it looks like (if I understood correctly) it can only ever return 0 or positive integers:

`//-------------------------------------------------------------------

TPalette::Page *TPalette::getStylePage(int styleId) const {
  if (0 <= styleId && styleId < getStyleCount())
    return m_styles[styleId].first;
  else
    return 0; 
}`

Perhaps the else condition in getStylePage() should return -1 and the line I changed should be left as it was?

I am not familiar enough with this function and don't know what else this might impact.... 

Dave.

damies13:
The other scenario I though of is if multiple styles are added, then 1 or more are deleted, then a call to `m_styles[styleId].first` on a deleted style might return a NULL or some error, but if thats the case it also has a logic error then as styleId could then be greater than getStyleCount().

shun-iwasawa:
@damies13 
I came to think that it would be more clear to maintain the current behavior for "Insert Paste" styles.
In #1743 I just removed the unused condition.

shun-iwasawa:
I'll close this as #1743 is merged. Please feel free to post any further ideas or opinions regarding this.

PR #1668, opened 2017-12-08
Moved xsheetChanged event and scene setDirty call outside loop

snapai:
In complex cutout animation scenes, these two events being called from within loops in the "Similar Drawing Substitution" commands were causing those commands to take as long as 90 seconds to complete (iterating over 200 frames and updating the scene & xsheet per-frame).

Moved the events outside all calls to the internal changeDrawing/setDrawing functions, so that they won't be caught up in the multi-frame loop. The scene and xsheet display only needs to be updated at the end of the command, not for each frame it alters.

This change drops the time for the command to complete to realtime, even in large/complex files.

### DIFF
--- a/toonz/sources/toonz/xsheetcmd.cpp
+++ b/toonz/sources/toonz/xsheetcmd.cpp
@@ -639,18 +639,16 @@ public:
         }
         if (found) {
           r++;
-          TApp::instance()->getCurrentXsheet()->notifyXsheetChanged();
-          TApp::instance()->getCurrentScene()->setDirtyFlag(true);
           continue;
         }
         changeDrawing(-m_direction, row, col);
-        TApp::instance()->getCurrentXsheet()->notifyXsheetChanged();
-		TApp::instance()->getCurrentScene()->setDirtyFlag(true);
         r++;
       }
       r = m_range.m_r0;
       c++;
     }
+	TApp::instance()->getCurrentXsheet()->notifyXsheetChanged();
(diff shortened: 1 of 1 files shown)


snapai:
Do I need to create an issue entry to go with this, since this already fixes the problem? I haven't done a pull request before. Whitespace probably needs fixed, the beautification & clang-format scripts wouldn't run for me.

manongjohn:
Nice fix. Probably a number of places that might be happening actuallly.

You don't need an issue for it since you found and fixed it.

Do you have the actual  clang executable on your machine? If you do,, you may need to update your PATH so it knows where it is when you run beautification.sh

snapai:
@manongjohn I don't, and the repository also no longer includes beautification.bat that's mentioned in the contributing doc - I apparently still have msys bash in my path somewhere so I was able to run beautification.sh but it just errored out since it couldn't find clang-format.

manongjohn:
Here is the exe. [clang-format.zip](https://github.com/opentoonz/opentoonz/files/1544202/tmp_11821-clang-format1065245104.zip)

Wherever you put it on your computer (c:\bin ?), make sure the windows path is updated.

snapai:
Done. 8a85d22 is whitespace changes only.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @snapai for the enhancement !

PR #1669, opened 2017-12-09
Timeline bug fixes and tweeks

manongjohn:
The PR fixes and tweeks the following Timeline issues:

1) Selected Keyframe move bug fix

When you Ctrl-Select cells with keyframes in them, the keyframe is selected and will move when the cell moves within the same column.  The original introduction of the timeline broke this feature by not allowing the cells to move due to the keyframes checking logic was incorrect.  This was not found until recently. This fix corrects it.

2) Timeline indicator tweeks

 - Timeline indicator in the 1st row is shifted to the right by 1 pixel.   Added logic to shift it back left for 1st row only
 - Timeline indicator was not being drawn over folded columns.  Added logic to corect
 - Timeline indicator was appearing in the 1st row when the bottom column was folded and did not move.  Corrected behavior.

3) Timeline border clipping fix

When there are fewer columns than what can be displayed in the xsheet area, Col1 is clipped at the bottom by 1 pixel due to the border.  It isn't obvious until you fold the column.  Adjusted logic to lower bottom of displayed columns by 1 pixel.

### DIFF
--- a/toonz/sources/toonz/keyframemover.cpp
+++ b/toonz/sources/toonz/keyframemover.cpp
@@ -330,10 +330,18 @@ void KeyframeMoverTool::rectSelect(int row, int col) {
 //-----------------------------------------------------------------------------
 
 bool KeyframeMoverTool::canMove(const TPoint &pos) {
-  TXsheet *xsh = getViewer()->getXsheet();
-  if (pos.x < 0) return false;
+  const Orientation *o = getViewer()->orientation();
+  TXsheet *xsh         = getViewer()->getXsheet();
 
-  int col      = pos.x;
+  TPoint usePos = pos;
+  if (!o->isVerticalTimeline()) {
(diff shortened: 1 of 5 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Dragging the ctrl-selected cells with keyframes in timeline view seems to have problem as follows:

The expected behavior would be that the ctrl-selected cells can be dragged only inside the same column as selected. 
However, if you drag the cells upward (with automatically inserting the empty columns), there is a case that the selection unexpectedly pasted without keyframes.

I think such problem can be another issue to be solved in the future.
I'll merge this anyway, as the problem is not much critical and this PR will fix other issues.

Wait for Jen.kins

shun-iwasawa:
LGTM Thank you @manongjohn for the fix.

PR #1687, opened 2017-12-15
Shift-Drag Keyframe fix

### LINKED ISSUE #1671: Keys flee being selected with accidental drag! (seen on timeline)
Selecting keys in multiple columns, where the second click accidentally includes a drag, causes even the unselected keys to "run away" effectively disappearing from the timeline. It also doesn't do this within an undo/redo wrapper, so the result is data loss that must be reverted.
keyflee

Steps to replicate:

1. Have multiple columns with concurrent keyframes on multiple frames.
2. Click one key in a same-frame line of keys.
3. Hold shift and click another key in the same line, with a drag event at the end of the click (ludicrously easy to do on a 2000+ lpi tablet), or click-and-drag again afterwards.
4. Watch your keys (even the unselected ones) fly away, never to return!
5. Notice how undo doesn't bring them back.

manongjohn:
This PR fixes #1671 

The logic for pasting keyframes in the flipped timeline was in the wrong spot.  Moved to where it needed to be so it only happened when actually pasting, and not when getting keyframes in general.

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -1583,7 +1583,7 @@ void TCellSelection::pasteCells() {
       // Retrieves all keyframe positions from mime data and translates them by
       // (r0,c0)
       std::set<TKeyframeSelection::Position> positions;
-      int newC0;
+      int newC0 = c0;
       if (viewer && !viewer->orientation()->isVerticalTimeline())
         newC0 = c0 - keyframeData->getColumnSpanCount() + 1;
       positions.insert(TKeyframeSelection::Position(r0, newC0));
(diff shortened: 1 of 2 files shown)


manongjohn:
@shun-iwasawa Seems something changed in the build environment.  One of the travis builds is failing on something I did not update and hasn't been updated or impacted by my change. Can you take a look?

shun-iwasawa:
@manongjohn Thank you for the modification. I'll check Travis.

shun-iwasawa:
Found that the [Trusty on Travis was updated last Tuesday](https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch) and some bug ( to be fixed after merging #1663 ) becomes an error under updated clang environment. 
I'll add a modification for temporally using the previous version until merging #1663 . Thanks!

shun-iwasawa:
I'll review this and #1703 next week. Thank you for your patience.

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed that the PR fixes the problem. I'll ignore Travis this time as Jen.kins succeeded on both Windows and OSX.

PR #1695, opened 2017-12-17
Hint NVIDIA and AMD GPU usage

manongjohn:
This PR adds logic to hint to NVIDIA and AMD cards that it should use the card's GPU.

OT currently does not automatically trigger GPU usage on NVIDIA and AMD cards. This causes OT to perform poorly on lower end machines because it uses CPU even though it may have a graphics card with a more powerful GPU.

Typically, NVIDIA and AMD cards have to be told that certain applications are to use its GPU.  OT is not one that is typically detected by default since it doesn't have a manufacturer's predefined profile in the card's settings. Users need to go into the card's properties and set up a profile for it in order to be used.

This change exports a setting that the cards scan for when OT starts up.  Once detected, it will force the use of the GPU unless an existing profile already exists.

NOTE: This may solve some performance problems with OT, but it does not resolve all issues due to lack of GPU usage.

Case in point...Currently without either an NVIDIA profile or this hint, when I run V 1.2, the viewers are blank.  Explicitly creating a NVIDIA profile fixes my problems, however, if I just run V 1.2 with this hint enabled, I can verify it is using GPU but I still have a blank viewer.

### DIFF
--- a/toonz/sources/toonz/main.cpp
+++ b/toonz/sources/toonz/main.cpp
@@ -230,6 +230,17 @@ project->setUseScenePath(TProject::Extras, false);
 }
 
 //-----------------------------------------------------------------------------
+#if defined(WIN32) || defined(_WIN32)
+extern "C" {
+__declspec(dllexport) DWORD NvOptimusEnablement                = 0x00000001;
+__declspec(dllexport) int AmdPowerXpressRequestHighPerformance = 1;
+}
+#else
+extern "C" {
+int NvOptimusEnablement                  = 1;
+int AmdPowerXpressRequestHighPerformance = 1;
+}
+#endif
 
 int main(int argc, char *argv[]) {
 #ifdef Q_OS_WIN


bobbybutterfly:
looking Forward to this:)
maybe it could solve the slow Playback Performance of the canvas. I guess, its GPU related, because
zooming out the canvas to just thumbnail size accelerates the speed enormously.
I have a i7 4970k Prozessor and a gtx 960 4gb.
My GPU-meter Shows barely no usage depite the stuttering Playback.

manongjohn:
@bobbybutterfly Are you able to get to the gtx's control panel and add the OT 1.2 executable to the 3d settings? Make sure it is set to use high-profile processor after adding it.

If you do this, you should start seeing gpu usage.

This change just makes it so you don't have to set this manually in your card's settings

bobbybutterfly:
@manongjohn i adjusted the settings. I had to add OT 1.2 manually, restarted. "high profile processor" was not available in the Program settings. No difference in Playbackspeed so far.
I figured out, that the slowdown was created by the soundfile (wave 8bit) which i use for lipsyncing my dialog.
After deleting the file, the slowdown was gone. But with a soundfile loaded, i have to zoom out the canvas very far to get a realtime lipsync sound/ animation match.

shun-iwasawa:
I couldn't see any enhancement with this PR because I already used the graphics card's GPU... But it looks fine and this fix may solve performance problem reported by some users. Thanks!

@manongjohn  I think the blank viewer problem is fixed in the current nightly (by stop using global share openGL context). 

Jenkins

shun-iwasawa:
Understood.

manongjohn:
@shun-iwasawa
I can't get this to work without that openGL context sharing enabled.

We can either...
1) Close this, for now, and wait until we can get the openGL context sharing working without causing the blank viewer.
or
2) Merge this knowing its harmless and not doing anything, so that it will work when we do finally switch over to openGL context sharing, or straight up openGL.

Ideally it would be great to have the openGL context sharing working and fix the real root cause as to why it is causing the blank viewer but I imagine trying to discover why and then fixing it may take a while.

Let me know which you prefer to do.

ASIDE: As far as why openGL context sharing might be breaking the viewer, I thought I read somewhere the openGL context sharing doesn't work well with threading without some work to keep track of stuff.  Could this somehow be causing the blank viewer?

manongjohn:
Closing for now since this change doesn't do anything.  If it ever becomes viable, I will reopen.

Oddly, I just tested a recent build in such a way as not to use my GPU and...I had no lag!!!  Not sure what changed to fix my lag.

RodneyBaker:
Regarding the vector lag...

I was noticing the other day that vector drawing was very fast and then later that day it was very slow.  

It appears that the difference is that during the slowdown the brushes were not the default vector brush but one of the custom brushes.  I haven't had time to investigate but I wan't to mention this as it might lead someone to a potential solution to unexplained slows downs while drawing vectors.  During the course of the day I had been exploring the various vector brushes and thought I had returned to the default brush but since there is a default brush (the one with a red diagonal line through it) for each of the vector brush sets I think I had one or two set right but not all of them.

PR #1707, opened 2017-12-24
Disable Auto-Stretch Frame

manongjohn:
This PR addresses a Google Forum feature request: [Autocreation preference needs some extra options](https://groups.google.com/forum/#!category-topic/opentoonz_en/feature-requests/6GhUKOlb4_k)

By default, if "Use Xsheet as Animation Sheet" setting is used, whenever drawing in a new frame, the prior frame, if there is one, is automatically stretched to meet the new frame. Additionally, the new frame is automatically stretched to meet the next frame, if there is one.

A new preference under Drawing has been added that is only enabled when "Use Xsheet as Animation Sheet" is selected.

image

When checked, the auto-stretching of prior and/or new frame does not happen.


Orphanlast:
You know, since we're on the topic of little things that are irritating when interacting with using the xsheet as an animation sheet. 

If you scrub through your timeline/xsheet, often times it'll revert you to usind your first style.  This is EXTREMELY irritating. And when you're on an empty frame, it doesn't show the style Pallet for the last selected level strip file on that column... and that's why it always goes back to the first style in that level's pallet.

shun-iwasawa:
IMO it would be more user-intuitive to make a text for the checkbox with positive sentence.
How about inverting the function of checkbox, having the text like `Enable auto-stretch frame` and being set to ON by default?

Sorry for my late response, I'll look into this.

RodneyBaker:
I like that idea Shun Iwasawa.
In general auto-stretch frame is the desired usage and so it makes sense that it would be the default.
Having the positive statement helps.

manongjohn:
Per recommendation, I've switched it to be Enable Auto-stretch, defaulted to ON.

shun-iwasawa:
Jenkins

shun-iwasawa:
It works fine. Thanks @manongjohn for the improvement ! LGTM

Issue #1620, opened 2017-11-29
My Paint + Plastic Tool (gif)


beeheemooth:
![ezgif-4-825684761c](https://user-images.githubusercontent.com/25724133/33402775-5cf4988c-d566-11e7-937b-a21de7398164.gif)

Applying the Plastic Tool to the MyPaint raster strokes, you can create amazing visual effects, clouds, smoke, cosmic nebulae and much more. Unfortunately, when you switch the Plastic Tool to Animate mode, there are cross-shaped artifacts.

beeheemooth:
![ezgif-4-f3e0837618](https://user-images.githubusercontent.com/25724133/33405378-c344933a-d570-11e7-9891-270ca752dd92.gif)

Another interesting fact is that switching the Preview change the position of crossed artifacts lines.

artisteacher:
I’m seeing the same thing in OS X. It’s interesting, going back to build skeleton hides the grid lines but they continue to display in preview. Changing the active tool or hiding the mesh has no impact. The plastic tool has so much potential but it needs the display reworked - particularly for Linux and OS X which also have the background transparency issue.

artisteacher:
Yes, this also appears in the final render.

beeheemooth:
This thing also appears when the color transparency changes
![ezgif-3-3c2f316a15](https://user-images.githubusercontent.com/25724133/34073818-0e37d2f2-e2ab-11e7-8581-b7131b8a7e30.gif)

shun-iwasawa:
Found that the unwanted line artifacts is due to the border of texture tiles used for plastic deformation.
Changing the maximum available tile size set [here](https://github.com/opentoonz/opentoonz/blob/792dc2b6b402d622292efe815bfaa2635c43eeb9/toonz/sources/tnzext/meshtexturizer.cpp#L249-L251) from 512 to larger number (let's say, 2048) will reduce the tile amount to 1 in most cases and will resolve the problem.

PR #1726, opened 2018-01-12
Fix the Line-shaped Artifacts on Semi-transparent Deformed Image with Plastic Tool

shun-iwasawa:
#1620 

This fix will enlarge the texture tile size from 512 to 2048 and minimize the occurrence of line-shaped artifacts which will appear at the border of the tiles.

**NOTE 1**
As a bad side effect of this fix, I changed the scene icon not to show the deformed image in order to avoid an unknown problem related to OpenGL context which I couldn't resolve. The icon will become non-deformed, just like a column icon of subxsheet.

**NOTE 2**
I set this PR to "WIP" state as I haven't test this fix on OSX yet. I'll post an update on next week.

### DIFF
--- a/toonz/sources/common/tfx/trenderer.cpp
+++ b/toonz/sources/common/tfx/trenderer.cpp
@@ -36,6 +36,12 @@
 #include <queue>
 #include <functional>
 
+#include <QOffscreenSurface>
+#include <QSurfaceFormat>
+#include <QOpenGLContext>
+#include <QThread>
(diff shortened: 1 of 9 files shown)


shun-iwasawa:
Jenkins

turtletooth:
Is this ready for review?

shun-iwasawa:
Yes, I confirmed that it works on OSX as well. Thanks.

turtletooth:

PR #1732, opened 2018-01-13
Render dialog no longer on top of all system windows.

snapai:
Removed Qt::WindowStaysOnTopHint - this was causing the render dialog to block windows of other applications. (the hint is applied system-wide, not just per-application). Also added a check to make sure that the dialog is set application-modal if not on Mac, which was likely the original intent of that flag.


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed it works fine. Thank you @snapai for the fix! LGTM

PR #1738, opened 2018-01-18
Fix Crash on Launching with Vector Fill Tool

### LINKED ISSUE #1737: "Old" Fill Tool crash (gif)
Another "old" crash. It is present in all versions of the program (1.2, OTX, Morevna)
Hope, it is easy to fix. 

Steps to reproduce:
1. Run OT
2. Select Fill Tool and change to "Line" mode.
3. Close program.
4. Run OT again.
5. Create new Vector Level.
6. Select Fill Tool.
7. OT will crash.

 If the Fill Tool is in any of the other modes (Areas, Lines & Areas), it does not crash.
If the scene or level is not the first after launching, the crash also does not occur, regardless of which mode the Fill Tool is in.

34687816-8e12e9b6-f4b8-11e7-9fb8-25d41f028db6

shun-iwasawa:
This will fix #1737 .

### DIFF
--- a/toonz/sources/tnztools/tooloptions.cpp
+++ b/toonz/sources/tnztools/tooloptions.cpp
@@ -910,7 +910,7 @@ ArrowToolOptionsBox::ArrowToolOptionsBox(
           SLOT(receiveMouseMove(QMouseEvent *)));
   connect(m_nsCenterLabel, SIGNAL(onMouseRelease(QMouseEvent *)),
           m_nsCenterField, SLOT(receiveMouseRelease(QMouseEvent *)));
-          
+
   if (globalKeyProp) {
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa This build fine on mac for Jenkins it looks like.  Would you like me to review this?

shun-iwasawa:
@turtletooth Yes please, thanks!
This crash was my fault. I missed null-pointer check when previously introducing a new option to the fill tool which is available only in toonz raster levels. Sorry for the trouble.

turtletooth:
Confirmed the fix.  It's in the new OTX already.  Thanks @shun-iwasawa for the fast fix.

PR #1740, opened 2018-01-19
Fix Plastic Tool not to Create the Black Contour 

### LINKED ISSUE #1249: Plastic Tool creates black contour
Plastic Tool creates black contour around mesh. The bug appears for both raster and vector layers, and  "premultiply" option (for raster +plastic tool) doesn't solve the problem.
ot_issue
how it appears - https://youtu.be/vaG4q1cW77Y

shun-iwasawa:
This will fix #1249 . 

The problem was related to the linear interpolation of the plastic-deformed texture.
At the border of texture, interpolation between "colored-opac" pixels and "black-transparent" pixels was made and caused color decaying in the interpolated pixels.
I fixed this by "expanding" color information at the border pixels before storing it to the texture. By doing this, interpolation will be done between  "colored-opac" pixels and "**colored**-transparent" pixels which will not decay color and obtain ideal result.

I also fixed the problem that the vector semi-transparent level becomes darken in the camstand view  when it is applied plastic-deformation.

fix_plastic_black_contour

### DIFF
--- a/toonz/sources/common/trop/toperators.cpp
+++ b/toonz/sources/common/trop/toperators.cpp
@@ -1547,7 +1547,7 @@ void TRop::depremultiply(const TRasterP &ras) {
       upPix  = upRow;
       endPix = upPix + ras32->getLx();
       while (upPix < endPix) {
-        depremult(*upPix);
+        if (upPix->m != 0) depremult(*upPix);
         ++upPix;
       }
       upRow += ras32->getWrap();
@@ -1563,7 +1563,7 @@ void TRop::depremultiply(const TRasterP &ras) {
         upPix  = upRow;
         endPix = upPix + ras64->getLx();
         while (upPix < endPix) {
-          depremult(*upPix);
+          if (upPix->m != 0) depremult(*upPix);
           ++upPix;
         }
         upRow += ras64->getWrap();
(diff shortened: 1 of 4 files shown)


shun-iwasawa:
Jenkins

turtletooth:
This is awesome @shun-iwasawa Thanks so much!
In all this, did you happen to see where pngs are being premultiplied where they shouldn't be?

turtletooth:

shun-iwasawa:
@turtletooth 

> did you happen to see where pngs are being premultiplied where they shouldn't be?

AFAIK, there is no premultiply process called in mid-flow. All the images in OT are loaded, handled and rendered with premultiplied-basis.

When you load the non-premultiplied level and the "premultiply" level settings option is applied, conversion is done [here](https://github.com/opentoonz/opentoonz/blob/34b8160c909246b105fe9e89277657ce8945890c/toonz/sources/common/trop/quickput.cpp#L556-L557) (for camstand view) and [here](https://github.com/opentoonz/opentoonz/blob/ef0db1c0bbbf0c13eab4d091d99915ebd3992d96/toonz/sources/toonzlib/tcolumnfx.cpp#L1115-L1119)  (for rendering). On rendering, after obtaining raster data of the column with the above conversion all the compositing is done on the basis that they are premultiplied.

So, IMO calling `TRop::depremultiply` before saving PNG would be reasonable.
When doing this PR I found that the same approach is taken - converting premultiplied levels to non-premultiplied raster before using it as texture data for the plastic tool.

turtletooth:
@shun This is ready to be merged.  Do you need to call Jenkins again on it?

shun-iwasawa:
Oh, I'll check what's happened. Thanks @turtletooth

shun-iwasawa:
Rebooted the server. Jenkins again

turtletooth:
LGTM.  Thanks @shun-iwasawa This is an exciting fix.

sfepa:
Excellent! Thanks a lot!

PR #1742, opened 2018-01-19
Allow Export of Individual Frames for Sprites

turtletooth:
Some game engines such as Godot make it easy to animate with individual sprites instead of spritesheets.  This creates a new option in the spritesheet export to export individual frames.  This gives the benefit of still trimming off the excess space, but then exporting as individual sprite files.


shun-iwasawa:
Jenkins

shun-iwasawa:
Thank you @turtletooth , it will bring benefit for game developers. LGTM

blurymind:
<3 Godot, Btw thank you for also making some tutorials on it. it's a brilliant engine and having someone who actually works on Opentoonz also use godot is very good for both communities

PR #1743, opened 2018-01-19
Fix #1639 by Removing Unused StyleId Check

### LINKED ISSUE #1639: styleselection.cpp error:ordered comparison palette->getStylePage(styleId)
Tried building from opentoonz/master today on OSX

Got the following error:

`
/Users/dave/Documents/Git/opentoonz/toonz/sources/toonzqt/styleselection.cpp:94:40: error: ordered comparison
      between pointer and zero ('TPalette::Page *' and 'int')
    if (palette->getStylePage(styleId) < 0) {
        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ^ ~
1 error generated.
`
Did a little searching for getStylePage() to see what the expected output of this function is (I was guessing that this might be a QT function that changed with the newer Qt version?) and found this link which gave me a clue how to fix it:

https://www.viva64.com/en/b/0389/ (v503)

I was actually a little annoyed at the author of the article, in the time it took to write that article criticising the OpenToonz Code, he could have forked it, made the necessary changes and submitted a PR.

shun-iwasawa:
This will fix #1639 

On pasting styles, there was wrong conditional inequality expression, comparing `TPalette::Page *` and zero.
It causes an error on recent Travis CI on Linux.

This bug has been existed from Toonz Harlequin. It seems to be originally intended to change pasting behavior according to the condition that the styleId of copied style is already in the paste destination palette. i.e. if there already are styles with the same styleId as the copied ones, then overwrite them.
But for now the condition expression is broken and the styles will always be inserted, regardless of the styleId.

@damies13 has been tackled with this issue to make it working properly. (in #1640 and #1663. Thanks @damies13 !)

However, I came to think that it would better to maintain the current behavior - pasted styles with "Insert Paste" command should always be inserted. It's more clear and we already have "Paste Color" and "Paste Color and Name" command as alternative.
So, this PR will remove unused 'if' condition at all. The pasting behavior will not change and will solve the Travis errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: cpp
 sudo: required
 dist: trusty
-group: deprecated-2017Q4 # temporary uses the previous version
 install:
     - if [[ $TRAVIS_OS_NAME == "osx" ]];   then bash ci-scripts/osx/travis-install.sh; fi
     - if [[ $TRAVIS_OS_NAME == "linux" ]]; then bash ci-scripts/linux/travis-install.sh; fi
--- a/toonz/sources/toonzqt/styleselection.cpp
+++ b/toonz/sources/toonzqt/styleselection.cpp
@@ -91,18 +91,13 @@ bool pasteStylesDataWithoutUndo(TPalette *palette, TPaletteHandle *pltHandle,
       break;
     }
 
-    if (palette->getStylePage(styleId) < 0) {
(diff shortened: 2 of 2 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1744, opened 2018-01-19
Use Precise Pen Position

shun-iwasawa:
This PR will enhance the brush stroke quality when using pen tablet for both the Toonz Raster / Raster levels.

I changed the mouse position to be handled in fractional number which can be obtained via pen tablet.
This enhancement can be found more obvious when you draw on the viewer with zoom scale lower than 100%.

I believe (hope) that this change will not slow the computation since the mouse position had already been converted to fractional number in many part of the software before introducing this PR.

The following image shows a comparison of strokes drawn on the left Viewer pane, with 50% zoom scale.

precise_pen_position

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -86,7 +86,7 @@ public:
   };
 
 public:
-  TPoint m_pos;  //!< Mouse position in window coordinates, bottom-left origin.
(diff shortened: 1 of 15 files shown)


shun-iwasawa:
Jenkins

turtletooth:
Tested this and it seems to work well.  I did not notice any noticeable lag.

turtletooth:
LGTM.  Thanks @shun-iwasawa

PR #1745, opened 2018-01-19
Fix Warnings on Launch

shun-iwasawa:
This PR will refactor sources and resource images in order to resolve all warnings appears on launching the software with console.

Here is a list of resolved warnings:

> QMetaObject::indexOfSignal: signal updateVisibility() from CommandBar redefined

Removed `signal updateVisibility()` from `XSheetToolbar` class definition as it is already defined in the base class `CommandBar`. Also removed `bool m_isCollapsible;` variable for the same reason.

> in XsheetGUI::XSheetToolbar
> QWidget::setMinimumSize: (XSheetToolbar/XsheetGUI::XSheetToolbar) Negative sizes (-5,30) are not possible

Made `XsheetViewer::positionSections()` to be executed only if the Xsheet panel is visible, since this function seems to be needed to be called only from showEvent and resizeEvent.

> QLayout: Attempting to add QLayout "" to StyleEditor "", which already has a layout

Made a new private function `StyleEditor::createVectorPage()` in order to make the ctor of `StyleEditor` cleaner. Made the sub-layouts to be constructed without parent widget pointer in order to avoid resetting of the parent widget. 

>PaletteViewerGUI::PaletteTabBar(0xc6d6d50, name="PaletteTabBar")  does not have a property named  "BottomAboveLineColor"
>DVGui::TabBar(0xc51b240, name="StyleEditorTabBar")  does not have a property named  "BottomAboveLineColor"

Removed `&:extend(.tab-container all);` from claimed objects in style sheets as they are not used.

> libpng warning: iCCP: known incorrect sRGB profile

Applied lossless file optimization to all resource PNG images in order to remove sRGB information (which may be added by PhotoShop). I used [FileOptimizer](https://sourceforge.net/projects/nikkhokkho/) by Javier Gutiérrez Chamorro. It could reduce the file size as well. :+1: 
Please note that the second commit ( `optimize png files` ) in this PR is dedicated to this.


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins again

masafumi-inoue:
LGTM

PR #1746, opened 2018-01-19
Fix png and spritesheet render (depremultiply)

### LINKED ISSUE #1230: PNG image is rendered with premultiplied alpha
According to the [PNG Specification](https://www.w3.org/TR/PNG/#6AlphaRepresentation) , it says that PNG should be straight alpha in normal.

For now OpenToonz renders PNG with premultiplied alpha. It causes problem especially with the loading option recently introduced in #1217.

turtletooth:
fixes #1230 

Depremultiplies png and spritesheets during render.

### DIFF
--- a/toonz/sources/common/timage_io/timage_io.cpp
+++ b/toonz/sources/common/timage_io/timage_io.cpp
@@ -631,6 +631,9 @@ void TImageWriter::save(const TImageP &img) {
     writer->open(file, info);
 
     ras->lock();
+    if (m_path.getType() == "png") {
+      TRop::depremultiply(ras);
+    }
     if (writer->getRowOrder() == Tiio::BOTTOM2TOP) {
       if (bpp == 1 || bpp == 8 || bpp == 24 || bpp == 32 || bpp == 16)
         for (int i = 0; i < ras->getLy(); i++)
--- a/toonz/sources/toonzlib/movierenderer.cpp
+++ b/toonz/sources/toonzlib/movierenderer.cpp
@@ -404,6 +404,9 @@ std::pair<bool, int> MovieRenderer::Imp::saveFrame(
 
     // Flush images
     try {
+      if (m_fp.getType() == "spritesheet") {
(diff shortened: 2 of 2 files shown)


jpturcotte:
Awesome! Thanks again, @turtletooth. Now I will sleep tight knowing my $15 bounty has found a great home!

gab3d:
clap clap clap

toonz/sources/common/timage_io/timage_io.cpp
```diff
@@ -631,6 +631,9 @@ void TImageWriter::save(const TImageP &img) {
     writer->open(file, info);
 
     ras->lock();
+    if (m_path.getType() == "png") {
+      TRop::depremultiply(ras);
+    }
```
shun-iwasawa: How about bringing this depremultiply process into the `PngWriter::writeLine()` function?

For now, there seems to be two problems as follows:
- If you activate `Open Flipbook after Rendering` preference option, rendered result shown in the flipbook looks strange as it is depremultiplied.
- For now the image is depremultiplied even if the `Alpha Channel` option of the PNG file format is off. In such case the rendered result becomes incorrect.

I think moving the depremultiply process into `PngWriter::writeLine()` can solve these problems and IMO it would be more reasonable to handle a format-specific process like this in each format's `Tiio::Writer` .

turtletooth: I tried playing with this and I can't get the correct results that way.  I just don't know enough about what I'm doing to get it right.

turtletooth: I've tried playing with this and I can't seem to get the right results this way.  I don't know enough about premultiplying to get it working.

shun-iwasawa: I'll try to modify this.

shun-iwasawa:
@turtletooth I posted a PR based on your fix. Can you please try #1844 ? Thanks!

PR #1753, opened 2018-01-24
Fix Interrupted Brush Stroke When Using Tablet

### LINKED ISSUE #1683: Interrupted brush stroke issues with tablet and 1.2
With the new tablet fix for OS X in 1.2, interrupted brush strokes still seem to have undesirable results. On a vector level, if the brush stroke is interrupted by an accidental button press or auto-save, the unfinished stroke is unable to be finished. After starting a new stroke, the incomplete stroke disappears and is replaced by the new stroke. On other level types, the interrupted stroke behaves like a completed stroke and remains visible. It would be better for auto-save and additional stylus input to be delayed until the brush stroke is finished. In the meantime, the accidental input can be minimized by being careful with the stylus grip (though that does not help the auto-save problem).

Possibly related, the greyed out/inaccessible menu problem #618 seems to be more of an issue than it’s been in a while.

Edit: it seems like auto-save does delay until the stroke is finished when using the trackpad (and mouse?) instead of a tablet. If auto-save interrupting something in progress is what actually triggers the inaccessible menu, delaying autosave until the stroke or etc. is complete could resolve the issue.

shun-iwasawa:
This will fix #1683 .

Replaced the cancelling stroke process by finishing of stroke when zero pressure is detected while tablet is moving.
Also refactored the source to handle pressure value as double (0.0-1.0).

It seems to solve the problem at least on Windows. I'll test on OSX from now.
Hope someone can check the modification on Linux.

**NOTE:**
To reproduce the issue, try using tablet with pressure as weak as possible. Some strokes were interrupted on my environment with Win7 + Wacom Intuos Art (CTH-690).

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -87,8 +87,8 @@ public:
 
 public:
   TPoint m_pos;  //!< Mouse position in window coordinates, bottom-left origin.
-  int m_pressure;  //!< Pressure of the tablet pen, or 255 for pure mouse
-                   //! events.
+  double m_pressure;  //!< Pressure of the tablet pen (0.0 - 1.0) , or 1.0 for
+                      //! pure mouse events.
   ModifierMask m_modifiersMask;  //!< Bitmask specifying key modifiers applying
                                  //! on the event.
 
@@ -100,7 +100,7 @@ public:
 
 public:
   TMouseEvent()
-      : m_pressure(255)
+      : m_pressure(1.0)
       , m_modifiersMask(NO_KEY)
       , m_buttons(Qt::NoButton)
       , m_button(Qt::NoButton)
--- a/toonz/sources/tnztools/brushtool.cpp
+++ b/toonz/sources/tnztools/brushtool.cpp
@@ -543,11 +543,10 @@ public:
 
 //=========================================================================================================
 
-double computeThickness(int pressure, const TDoublePairProperty &property,
+double computeThickness(double pressure, const TDoublePairProperty &property,
                         bool isPath) {
   if (isPath) return 0.0;
-  double p                    = pressure / 255.0;
-  double t                    = p * p * p;
+  double t                    = pressure * pressure * pressure;
   double thick0               = property.getValue().first;
   double thick1               = property.getValue().second;
   if (thick1 < 0.0001) thick0 = thick1 = 0.0;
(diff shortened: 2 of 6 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Oh, noticed that [Appveyor just updated Qt version to 5.9.3](https://github.com/appveyor/ci/issues/1949) yesterday!

artisteacher:
Will this also help with auto-save interruptions? Thanks so much for continuing to work on the tablet issues!

shun-iwasawa:
> Will this also help with auto-save interruptions?

I'll check it. Thank you for the notification!

shun-iwasawa:
Added calling `setToolBusy(true)` when using tablet. 
Jenkins again

shun-iwasawa:
Fixed conflicts. Jenkins again

masafumi-inoue:
LGTM

shun-iwasawa:
Found some malfunctions with tablet behavior on OSX. I'll fix it right away.

artisteacher:
The menu auto-save issue seems to be fixed! I do sometimes have to click twice when changing focus from the viewer to Xsheet or etc. (ie. to change cells or tools). If autosave runs just before starting a stroke, sometimes I have to start the stroke twice - this seems to be pretty rarely though & part of the focus issue.

Possibly related to this is #1796 - hovering doesn’t seem to be recognized by the tablet without light pressure. That could be a tablet setting issue but my settings are frozen due to an upgrade & I can’t check. I think it’s probably best that OT doesn’t recognize hovering in the viewer since it prevents that random scribbling, but it would still be helpful to have outside of the viewer.

So far, this seems a lot better!

RodneyBaker:
> If autosave runs just before starting a stroke

How often do you have Autosave set to save?

artisteacher:
I had it set to every minute to confirm that the greyed out menu problem was fixed. I usually keep autosave at somewhere around every 7 minutes (still fairly low but minimizes lost work after an unexpected crash).

RodneyBaker:
I had it set to every minute to confirm that the greyed out menu problem was fixed.

artisteacher:
I agree that developing the habit to regularly save manually is probably a better choice than auto-save. However, auto-save shouldn't break program functionality.

After updating again to the latest nightly build, I am getting unintended strokes directly after auto-save when the stylus is hovering & the disabled menu periodically.

PR #1760, opened 2018-01-25
Fix Vector Eraser Tool Crash

### LINKED ISSUE #1748: Vector Eraser Tool crash
I can repeatedly crash OT when erasing on a vector level as follows 

1) Draw 1 or more lines in a vector level
2) Switch to the Eraser Tool and select type "Normal"
3) Erase a part of any line
4) Switch Eraser Tool to Type "Rectangle" or "Freehand"
5) Attempt to erase  -> OT Crashes

This does not happen with "Polyline"

I am able to repeat this on OT 1.1.3 and 1.2 on a Windows.

shun-iwasawa:
This will fix #1748 and modified other things related to Vector Eraser Tool as follows:

- Modified line color and patter drawn with Rect / Polyline / Freehand shape + frame range option to be similar to Raster Eraser Tool.
- Modified that changing the current tool while making the polyline and back to the vector eraser tool again retains previous polyline. Now the previous polyline will be cleared when activating eraser tool.

### DIFF
--- a/toonz/sources/tnztools/vectorerasertool.cpp
+++ b/toonz/sources/tnztools/vectorerasertool.cpp
@@ -417,12 +417,12 @@ void EraserTool::draw() {
   if (vi) {
     bool blackBg = ToonzCheck::instance()->getChecks() & ToonzCheck::eBlackBg;
     if (m_eraseType.getValue() == RECT_ERASE) {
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

Issue #1761, opened 2018-01-25
Frame sequence import issues


MisterABK:
(I'm not sure if this is an issue per se. Please let me know if I'm submitting in the wrong place.)

Opentoonz seems to be quite limited when it comes to importing frame sequences. It throws a "Malformed Name" error when a sequence begins at zero: "imgseq_0000.tif". Also it ignores frame sequences which have too many leading zeroes: "imgseq_00001.tif" and treats them as separate images. Since After Effects and others do these things as default, it seems like Opentoonz should accommodate them.

Edit: It also doesn't seem to accept sequences where the number is not separated by an underscore: "imgseq0001.tif".

RodneyBaker:
It's definitely an issue but we need to lock down specific instances of where filenames are problematic.  Your description goes a long way toward doing that.

When you say 'It throws a "Mallformed Name" error... if you can provide specific information on that it will be appreciated.

This may be a duplicate report but if it is the other(s) have likely been lost to time so could be closed after merging with this report.

Concerning the Malforming of names... OpenToonz uses a specific regex (regular expression?) to accommodate naming (editable in Preferences). This should allow us to resolve this problem but unfortunately the problem persists.  There is a least one forum topic that talks about this in some detail to include a regex expression that should address your specific issue.

See:  https://groups.google.com/forum/#!topic/opentoonz_en/QraTfiOyKFI

RodneyBaker:
I'm guessing the regex code involved would be here:

https://github.com/opentoonz/opentoonz/search?utf8=%E2%9C%93&q=regex&type=

MisterABK:
The frame sequence was rendered outside of Opentoonz, because I'm rotoscoping video footage.
When I load a sequence beginning "_0000" I get an error which says: 
"imgseq:imgseq_0000.tif: Malformed frame name: Skipping frame."

I can then continue working with that sequence (presumably it starts from frame 0001). I get the same message every time open the scene. Hope that helps.

manongjohn:
This is the same as Issue #469 which was partially fixed to stop the crash with PR #967 which provides the malformed message you are getting instead of crashing outright.

I'm marking this as a duplicate and recommend closing.

MisterABK:
My report does contain additional observations - the problems with handling sequences with extra leading zeroes or no underscore - Is it worth recording those somewhere?

manongjohn:
Add a comment to 469 so if someone looks at that issue, they will see your additional information.

PR #1762, opened 2018-01-26
WIP - Start of Loading Image Sequences with 5 Numbers, Instead of Just 4

turtletooth:
This is the start of loading image sequences with one more place value than we currently load.  It loads and displays the images, but any modification to the images doesn't work yet.

I don't have time to continue this, but if anyone wants to pick it up, it's a start.

I'll be glad to answer questions about what is going on in the code.

If no one picks this up by March, I'll go ahead and close this to avoid cluttering up the PR list.


turtletooth:
If this gets done, this will take care of part of #1761

turtletooth:
@manongjohn Thanks for the new commit on your other PR.  I can close this now. . .

PR #1766, opened 2018-01-26
Fix Template Room Path

shun-iwasawa:
This will fix the problem as follows:

The rooms template cannot be loaded properly if `$TOONZPROFILES` is not set to `$TOONZROOT/profiles` .


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1776, opened 2018-01-29
Fix Vector Selection Move and Refactor Tool::KeyDown

### LINKED ISSUE #1765: Dragging object with Selection Tool jumps across their existing location
[This problem is reported in the forum.](https://groups.google.com/forum/#!msg/opentoonz_en/iR8pqR3fTvE/Wr1QpOAZAAAJ)

> Hi folks,
> 
> I'm using Morevna 1.2.0.2 on Debian Linux with 15G RAM and 15G swap.
> 
> I think I found a bug in OpenToonz. I've been seeing this since I started using the software, but just got around to making a video of it.
> 
> https://www.youtube.com/watch?v=JAZSKYhpgV4
> 
> In the first part of the video, I click-drag an object, and you can see that whenever it approaches the place I grabbed it from, the object jumps past that point. You can see the mouse get ahead of the object each time, and then the object suddenly snaps to catch up to the mouse. By experiment, I've determined that the object jumps across exactly the area that I'm trying to place it.
> 
> In the second part of the video, I click-drag the object to a distant location, and let it go. Now its location is elsewhere, so if I drag it again it will be coming from someplace distant. Then I click-drag it back to where it was, and while still dragging over the target area, you can see that the mouse no longer gets ahead of the object. This is because I'm no longer dragging the object over the more distant space where OpenToonz thinks it is.
> 
> I hope I just explained that clearly.
> 
> This is highly repeatable. It happens every time I try to do a micro-adjustment of an object's position using the Selection Tool.
> 
> The reason this bug is bad is because I often want to do micro-adjustments of objects, and I can't. If I use the workaround described here, it ceases to be a micro-adjustment, and just becomes guesswork. I can't home in on the placement I actually want.
> 
> Be well,
> Zack

This problem occurs not only in Morevna, but also in the master version.
I think I found a cause of this problem. See [my comment](https://groups.google.com/d/msg/opentoonz_en/iR8pqR3fTvE/biQSkSAdAAAJ) in the topic.

shun-iwasawa:
This fixes #1765 

Made Vector Selection Tool to be snapped at the original position, also made such snapping to be canceled if Ctrl key is pressed while dragging.
When you move objects with arrow keys, you can make precise move by pressing Ctrl.

Along with the fix, I refactored the function `Tool::keyDown()` to simply use `QKeyEvent` as an argument. It will stop using `TwConsts` keycodes and use Qt-based constants so that it can skip conversion from Qt-based to `TwConsts` which seems to be unneeded.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1780, opened 2018-01-31
Fix Palette Titlebar to be Updated When Loading Color Model 

shun-iwasawa:
This small will fix a problem as follows.

When loading a color model with "Overwrite the destination palette" option, level palette will be overwritten by the color model's palette. However, asterisk "*" ( a sign indicating the palette is changed ) is not shown on the palette's titlebar.


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
I'll merge this as the fix is quite simple and obvious.

Issue #1774, opened 2018-01-28
Light Table/White Box not showing up upon start up


benalsyght:
So I literally just downloaded Opentoonz 1.2.0 just a couple hours ago and I've been searching on YOutube and the web for answers but nothing it giving me any help. Upon start up, the light table/drawing pad/white box with the light blue circle around it (or whatever you want to call it) does not show up. All I get is a grey screen, yet when I draw on the screen, I can see the lines to right of me in the timeline. So how do I get the screen to show up? [Here's a link to a YouTube video to show what I mean.](https://youtu.be/BffeWKZD-F4) My computer is a Toshiba Satellite 2009 running on Windows 7. I don't know if that has any relevance to the software, but I feel like it's important.

Orphanlast:
I responded to your comment in my video, there you will find a download link to OpenToonz Morevna 1.2.0.2 and instructions on what to do before and after install

jpturcotte:
@shun-iwasawa These issues are multiplying. Can another official build be released ASAP, please? Imagine the number of people who have this problem, and instead of reporting it here just give up in frustration!

Orphanlast:
I'd report the "Opentoonz 1.2.0.3 Is Here" video. Not much to report, but still.

DarrenTea:
I could be wrong, it's hard to tell looking at the video, but as an observation, when you draw in the level strip, you're supposed to see a grey background and you only see the light table when you select in the xsheet (or timeline) and draw in there. Could this be your issue?

Personally, I'd prefer to see the lightbox when I draw in the level strip too. It makes it more visually appealing and helps with orientation when you've rotated your image.

manongjohn:
@benalsyght This is a known bug. You should install the OTX or Morevna edition instead.

Labeling as duplicate.

shun-iwasawa:
Yesterday I managed to get a computer which can reproduce the problem. ( It was my wife's private laptop. ) 
I'll tackle with this right away to release a hot-fix.

shun-iwasawa:
Interim report - Found that the problem seems to be NOT due to migration from `QGLWidget` to `QOpenGLWidget`, but caused by `QCoreApplication::setAttribute(Qt::AA_ShareOpenGLContexts);` added [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/main.cpp#L266) in order to make the opengl context handling easier. I confirmed that simply removing this line brings back the graphics to the viewer. 
[...]

PR #1785, opened 2018-02-02
Stop Using Global OpenGL Context

### LINKED ISSUE #1674: Blank Viewer/Combo Viewer after installing V1.2
A user on google groups originally reported this issue under V1.1.3 recently and saw the same issue after recently isntalling V1.2. (https://groups.google.com/forum/#!category-topic/opentoonz_en/yV8uVafZBxY)

I didn't have the problem with V1.1.3, but when I installed V1.2 today, I noticed the problem also.

We are both running Win7 64 bit.

However, it should be noted that, when I build V1.2 locally with the latest OT master code, the executable that is generated is Opentoonz_1.1.exe and it works normally, but if I renamed it to Opentoonz_1.2.exe the viewers are blank.

Unfortunately, if I rename the officially installed Opentoonz_1.2.exe back to Opentoonz_1.1.exe, it doesn't fix the issue.

Is something in the compiling process not accounting for the executable name change?  Are we missing some registry setting?

shun-iwasawa:
This will fix #1674 and #1774 
This is WIP as I still do not verify the fix.

### DIFF
--- a/toonz/sources/common/tfx/trenderer.cpp
+++ b/toonz/sources/common/tfx/trenderer.cpp
@@ -1424,8 +1424,6 @@ void TRendererImp::startRendering(
         QThread::currentThread() == qGuiApp->thread()) {
       rs.m_offScreenSurface.reset(new QOffscreenSurface());
       rs.m_offScreenSurface->setFormat(QSurfaceFormat::defaultFormat());
-      rs.m_offScreenSurface->setScreen(
-          QOpenGLContext::globalShareContext()->screen());
       rs.m_offScreenSurface->create();
     }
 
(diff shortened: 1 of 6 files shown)


shun-iwasawa:
It seems to work almost OK - except when rendering plastic deformed scenes with tcomposer.

shun-iwasawa:
Unfortunately I couldn't fix the problem with tcomposer so far. I think the problem can be separated to another issue, as rendering on GUI can be workaround of using tcomposer.

shun-iwasawa:
Jenkins

turtletooth:
@shun-iwasawa I tested this and it seems to work fine on my machine.  If you are ready to merge this, please do.  Thanks for working on this.

PR #1793, opened 2018-02-09
Color Calibration Feature using 3D LUT

shun-iwasawa:
This feature was demanded by Studio Ghibli.
(It was in Toonz Ghibli Edition but has been missing in OT until now.)

This PR will add a real-time color correction feature using 3 dimensional lookup table (3D LUT).

By applying 3D LUT file which is properly configured with the current display monitor, it will become possible to work on different color space, such as [DCI-P3](https://en.wikipedia.org/wiki/DCI-P3).

To use this feature, specify .3dl file path in Preferences > Interface category and restart the software.

- It will correct colors displayed on Viewer, Combo Viewer, Flipbook, Color Model, Palette, and Style Editor.
- For now, it only accepts .3dl file format. It must follow specification written [here](http://download.autodesk.com/us/systemdocs/help/2009/lustre_ext1/index.html?url=WSc4e151a45a3b785a24c3d9a411df9298473-7ffd.htm,topicNumber=d0e8061).
- For now, only Windows version can detect the current connected monitor. For other platforms only one 3D LUT file can be applied for any type of monitors.
- Even on Windows version, connecting multiple different monitors at the same time is not supported for now. It will only identify the first registered monitor. 
- PLEASE NOTE : For now using plastic deformation together with this feature causes display issue (related to texture, I guess) that the deformed image is not shown under the camera stand mode. I'll fix it soon.

3dlut

In the right pane of the above example, a "Sedona" LUT (downloaded from [here](https://luts.iwltbap.com/free-lut-download-color-grade-sedona/) and added some header to satisfy specification written [here](http://download.autodesk.com/us/systemdocs/help/2009/lustre_ext1/index.html?url=WSc4e151a45a3b785a24c3d9a411df9298473-7ffd.htm,topicNumber=d0e8061)) was applied to the original view on the left.

Reference : [GPU Gems 2 Chapter 24 : Using Lookup Tables to Accelerate Color Transformations](https://developer.nvidia.com/gpugems/GPUGems2/gpugems2_chapter24.html)

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -59,9 +59,6 @@ class ToolOptionsBox;
 
 class QMenu;
 
-// iwsw commented out
-// class Ghibli3DLutUtil;
-
 //===================================================================
 
 //*****************************************************************************************
@@ -655,8 +652,8 @@ public:
   /*-- Toolで画面の内外を判断するため --*/
   virtual TRectD getGeometry() const = 0;
 
-  // iwsw commented out
-  // virtual Ghibli3DLutUtil* get3DLutUtil(){ return 0; }
+  virtual void bindFBO() {}
+  virtual void releaseFBO() {}
 };
 
 #endif
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -235,6 +235,15 @@ public:
   void setInterfaceFontWeight(int weight);
   int getInterfaceFontWeight() { return m_interfaceFontWeight; }
 
+  // color calibration using 3DLUT
+  void enableColorCalibration(bool on);
+  bool isColorCalibrationEnabled() const { return m_colorCalibrationEnabled; }
+  void setColorCalibrationLutPath(QString monitorName, QString path);
+  QMap<QString, QString> &getColorCalibrationLutPathMap() {
+    return m_colorCalibrationLutPaths;
+  }
+  QString getColorCalibrationLutPath(QString &monitorName) const;
+
   // Visualization  tab
 
   void setShow0ThickLines(bool on);
@@ -651,6 +660,12 @@ private:
   PathAliasPriority m_pathAliasPriority;
 
   bool m_currentTimelineEnabled;
+
+  // color calibration using 3DLUT
+  bool m_colorCalibrationEnabled = true;  //とりあえず
+  // map of [monitor name]-[path to the lut file].
+  // for now non-Windows accepts only one lut path for all kinds of monitors
+  QMap<QString, QString> m_colorCalibrationLutPaths;
 
 private:
   Preferences();
--- a/toonz/sources/include/toonzqt/lutcalibrator.h
+++ b/toonz/sources/include/toonzqt/lutcalibrator.h
@@ -1,0 +1,78 @@
+#pragma once
+
+#ifndef LUT_CALIBRATOR_H
+#define LUT_CALIBRATOR_H
+
+#include "tcommon.h"
+#include "tpixelutils.h"
+
+#include <QOpenGLBuffer>
+#include <QMatrix4x4>
+#include <QOpenGLFunctions>
+
+#undef DVAPI
+#undef DVVAR
+#ifdef TOONZQT_EXPORTS
+#define DVAPI DV_EXPORT_API
+#define DVVAR DV_EXPORT_VAR
+#else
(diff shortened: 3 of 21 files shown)


shun-iwasawa:
Jenkins

gab3d:
this is a great addition!
thanks. :)

gab3d:
perhaps it'd be useful to have an on/off switch button for the correction, at the top of the visor, for quickly testing its effects.

(although considering the need to restart the program to activate it, I don't know if this is even currently possible)

shun-iwasawa:
Hmm, I couldn't fix the issue with plastic texture yet. :-(

@gab3d
Thank you for the comment. For now you need to reboot the software to activate/deactivate this option.
Well, I agree it would be better to have on/off button for testing the effect of LUT. However, once the proper LUT is set, the option will not be changed during production work.

Jenkins again

shun-iwasawa:
I'll post the problem with the plastic tool as another issue.

masafumi-inoue:
LGTM

shun-iwasawa:
Posted the issue.

RodneyBaker:
3D LUTs in OpenToonz!  Great addition.

I just went through the manual process of editing the header file of the free 3DLUT you linked to and it works great

Thanks!

RodneyBaker:
Compare that with F8700 (teal and blue) LUT:
f8700

RodneyBaker:
And the original with no LUT for reference:
no lut

RodneyBaker:
The known issue with opengl where viewports display as black screen still exists.
The known issue with plastic tool is one trigger.
When last checked editing a standard raster level can also cause the change.  Will verify again.

Somewhat related:  I've had mixed success with taking the various 3D LUTs I've used for this feature and using them for rendering in other programs.  My latest attempts seemed to have the colors reversing so more  exploration is needed.  The reason for this exploration is so that once a user gets the image looking as they desire (on any monitor but specifically on a calibrated monitor) they can then burn that look from the LUT into the image to allow for more general sharing with others who do not have the Luts or calibration.

RodneyBaker:
Post mortem  follow up...

I was struggling a bit to remember the 3dlut file header requirements but finally got it right.

The good news is that recent PRs fixing full screen operation appear to have similarly resolved those issue with 3Dluts.

I'm attaching a 3D lut for those that wish to try this out.
Keep in mind that the luts don't get baked into the image... they are just for the view of the monitor.
Long ago I did test with baking the luts into the image via Blackmagic Fusion but I didn't delve deeply.

Note:  The attached text file will need it's extension changed from .txt to .3dl

[OT3dlut.txt](https://github.com/opentoonz/opentoonz/files/3759076/OT3dlut.txt)

RodneyBaker:
As I don't know of any way to turn off the LUT view in OT (short of increasing the size of Level Strip icons...

A nice workflow for seeing the LUT's results side by side with a view without it is to use two instances of Opentoonz at the same time.

Here's one of another LUT I cobbled together than focuses on teals and blues.
It also adds noise... giving it a bit of a pencil line effect too.
(original drawing on right... same image with LUT display on left)
LUTdrawImages

RodneyBaker:
As a follow up...

I have been using the program DuMe to convert LUTs for use in Opentoonz.
I'd like to add a few sample LUTs to the v1.7 release and am looking into what the ideal exmple LUTs might be and making sure they are freely licensed to faciliate usage.

Here's some information on DuMe's LUT conversion:
https://github.com/OpenAnimationLibrary/DuME/blob/master/src-docs/docs/lut-converter.md

DuMe itself might make a nice (optional) third party program to pair up with Opentoonz in similar fashion as FFMPEG (which DuMe uses) and Rhubarb.  In my estimation it could use a bit of UI overhaul but over all works great.  DuMe is primarily a media encoder and could also fulfill the needs of a large number of production needs with its ability to cue up rendering.  DuMe's development does appear to have slowed considerably and it is still considered in Beta (although stated as good for production) so that is a consideration as well.  I'm quite certain they would appreciate DuMe being used more widely and as perhaps especially within the like minded open source Opentoonz community.

RodneyBaker:
Importantly above...
The Flame LUT format opens in Opentoonz without any editing of the LUT file.
Other formats will require manual editing of the LUT file as described above by @shun-iwasawa 

At present I recommend maintaining two LUT files; one .cube format for use in the program that bakes the LUT into an image and one .3dl format for viewing in Opentoonz.  Ideally this LUT set would also contain the source HALD image as well as an image that shows before and after results of applying the LUT.

PR #1801, opened 2018-02-13
Add an issue template to provide some framing for new issues.

romeojulietthotel:
Issue templates should do two things. _One_ is to act as a guide for the user that is reporting the issue and _two_ to provide some useful data for a busy maintainer. It is meant to start the conversation off with just the right amount of momentum and is not a burden to the reporter.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -13,6 +13,11 @@ Please use the SEARCH function before opening a new issue!
 ### --Steps to reproduce--
 <!-- Describe the steps to reproduce the problem, more steps are better. -->
 
+### --Actual Results--
+<!-- Explain what you actually got. -->
+
+### --Expected Results--
+<!-- Explain how the program should behave in the context you're using it. -->
 
 ### --System Information--
 + OpenToonz Version:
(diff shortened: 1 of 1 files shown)


jpturcotte:
Sorry for the duplicate PR! Might I suggest adding my proposed header and Prerequisites section, though?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,54 @@
...
+
+### --System Information--
++ OpenToonz Version:
+<!-- You can find the VERSION in Menu->Help->About OpenToonz -->
+
++ Operating System:
+<!-- (e.g., Windows10, macOS 10.12, Ubuntu 16.04, Debian Stretch) -->
+
++ RAM Size:
+<!-- (4GB, 8GB etc.) -->
```
shun-iwasawa: In my experience it was rare that the issue was proved to be due to the memory shortage so far.
How about making this item optional, or combining with the following `Graphics Tablet` item and rename like `Other Device Info` which is for users to write down any optional device information?

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,54 @@
...
+
++ Graphics Tablet:
+<!-- (Optional) (Write your MODEL, e.g., WACOM Intuos 2) -->
+
+### --Actual Results--
+<!-- Explain what you actually got. -->
+
+### --Expected Results--
+<!-- Explain how the program should behave in the context you're using it. -->
```
shun-iwasawa: How about moving `Actual Results` and `Expected Results` to just under `Step to reproduce` , since they are related?

romeojulietthotel: I had those last as sometimes they become quite long and the other info can provide a quick summary without having to look over long cut-n-pasted fields. It's a minor consideration and is easily changed back if cut-n-pastes happen often.

shun-iwasawa:
LGTM Thank you @romeojulietthotel !

PR #1810, opened 2018-02-16
New Fx: Barrel Distort Iwa Fx

shun-iwasawa:
This fx represents lens distortion effect.

Parameters are as follows: 
- **Center** : specifies the center of distortion.
- **Distortion** : specifies the amount of distortion. Distortion shape will become pincushion-shape with positive value, and barrel-shape with negative value.
image
- **Distortion Aspect Ratio** : specifies the aspect ratio of distortion.
  
- **Source Precision** : This fx will enlarge some part of the source image. With this parameter more than 1, input image will be computed with higher dpi in order to minimize alias caused by such enlargement.
  
- **Chromatic Aberration** : with this parameter non-zero, images will be computed with different distortion amount for each RGB channel which causes pseudo chromatic aberration effect.
  
- **Vignetting Amount, Vignetting Gamma, Vignetting Midpoint** : enables pseudo-[vignetting](https://en.wikipedia.org/wiki/Vignetting) effect which fades off (or lighten) toward the edges of image. 
image
- **Scale**: specifies global scaling. 
   **NOTE**: If you would like to make a size of the center of image unchanged on barrel distortion, set the scale value to `1.0 + distortion` . (e.g. if you set `distortion = -0.4`, then setting `scale = 0.6` can make the image center to be the same size as the original )

This Fx was requested by some Japanese animation studio.
The major part of specification is similar to [Lens Correction filter of PhotoShop](https://helpx.adobe.com/photoshop/using/correcting-image-distortion-noise.html).

Reference : [AW. Fitzgibbon "Simultaneous linear estimation of multiple view geometry and lens distortion", 2001](http://www.robots.ox.ac.uk/~vgg/publications/papers/fitzgibbon01b.pdf)

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1244,6 +1244,16 @@
   <item>"STD_iwa_TimeCodeFx.showBox"	"Show Box"</item>
   <item>"STD_iwa_TimeCodeFx.boxColor"	"Box Color"</item>
 
+  <item>"STD_iwa_BarrelDistortFx"	"Barrel Distort Iwa"</item>
+  <item>"STD_iwa_BarrelDistortFx.center" "Center"</item>
+  <item>"STD_iwa_BarrelDistortFx.distortion" "Distortion"</item>
+  <item>"STD_iwa_BarrelDistortFx.distortionAspect" "Distortion Aspect Ratio"</item>
+  <item>"STD_iwa_BarrelDistortFx.precision" "Source Precision"</item>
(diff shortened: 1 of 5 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Added the scale parameter which was requested by the studio.

Jenkins again

masafumi-inoue:
LGTM

PR #1811, opened 2018-02-16
Fix Jumping Stroke Behavior

### LINKED ISSUE #1781: Unwanted Stroke is Drawn when Using Tablet
This problem can be reproduced only with tablet.

- Select Brush tool
- Start drawing a stroke
- Press middle or right button without releasing the pen
- Release the pen
- Release middle or right button
- Try to start drawing another stroke from different position 

In my environment, unwanted line appears connecting from the end point of the first stroke to the start point of the second stroke. This problem occurs both on vector and raster levels.

line_problem
I'm not sure this problem is related to the problem mentioned in [this comment](https://github.com/opentoonz/opentoonz/pull/1659#issuecomment-350424961).

shun-iwasawa:
This will fix #1781

When the right or middle button is pressed while left dragging (i.e. while drawing stroke with brush tool), 
I added a procedure to finish the action for the left button (i.e. finish the stroke).

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -587,13 +587,19 @@ void SceneViewer::mousePressEvent(QMouseEvent *event) {
 //-----------------------------------------------------------------------------
 
 void SceneViewer::onPress(const TMouseEvent &event) {
+  if (m_mouseButton != Qt::NoButton) {
+    // if some button is pressed while another button being active,
+    // finish the action for the previous button first.
+    TMouseEvent preEvent = event;
+    preEvent.m_button    = m_mouseButton;
+    onRelease(preEvent);
+  }
+
   // evita i press ripetuti
   if (m_buttonClicked) return;
   m_buttonClicked = true;
   m_toolSwitched  = false;
   if (m_freezedStatus != NO_FREEZED) return;
-
-  if (m_mouseButton != Qt::NoButton) return;
 
   m_pos         = event.mousePos() * getDevPixRatio();
   m_mouseButton = event.button();


shun-iwasawa:
Jenkins

artisteacher:
Between this and #1753, should all of the OS X tablet issues be resolved? I can't wait to test these fixes out!

masafumi-inoue:
LGTM

artisteacher:
I'm still getting an occasional jumping stroke.

df-0:
Hi guys, I'm definitely experiencing this issue _a lot_ with OS X 10.14.3, Cintiq Pro and OpenToonz 1.30 (Morevna Edition) Built Match 13 2019 12:13:08, to the point where it's basically unusable (for me at least). It seems similar to the behaviour I've experienced with previous versions. It seems to happen for several different reasons, but one very reliable way to trigger it is draw a stroke and quickly lift the pen out of range of the tablet sensors. Presumably it doesn't register the mouse up or something, because when you touch again you get a big fat line. [Here's a video of it in action](http://sendvid.com/ed70jzv4). Would love to see this fixed, this is literally the only reason we're still stuck using Adobe Animate and don't use openToonz!

manongjohn:
Please try Official 1.3 and let us know if it happen with that.  The Morevna edition doesn't have a lot of the fixes that official 1.3 has.

df-0:
Ah, ok, will give it a go - thanks!

df-0:
wow, that actually seems to have fixed it - thanks! That's been an issue throughout 1.0, 1.1 and 1.2, great to see it fixed! [Thanks](https://imgur.com/E8R164y)!

DanielBlodgett:
So I'm using the new 1.3 and getting this same issue.  I've reinstalled even...twice.  This bug kinda makes this program worthless for me.  Is there another fix to this issue?

manongjohn:
Are you using 1.3 Official or Morevna version?

DanielBlodgett:
Using 1.3 official.  I'm using Windows 10, XP Pen 22hd.  I don't have this issue in: Photoshop, Clip Studio Paint, or Blender.

PR #1814, opened 2018-02-19
Disable 3dlut by Default

shun-iwasawa:
I forgot to set the default value for 3dlut option to `false` (disabled) . Sorry for trouble!


shun-iwasawa:
I'll merge as it is obvious.

PR #1823, opened 2018-02-22
Fix Crash on Undo of Raster Freehand Eraser

### LINKED ISSUE #1821: Crash with Freehand Erase and Undo
Reporting an odd bug here that I confirmed based on steps given by @beeheemooth.

Steps to repeat:

1. On the Raster Level, select the Eraser Tool and change its type to Freehand (this is done so that the next time the OT is started, the eraser is already in Freehand mode, and it does not have to be switched manually)
2. Restart OpenToonz.
3. Create a new raster level
4. Using MyPaint, draw something with a brush.
5. Choose Eraser Tool with HOTKEY (this is important!) and it should already be in Freehand mode.
6. Erase the fragment using this mode.
7. Undo this. The program crashes.

The crash will occur if we change step 7 to read.
7.  Go to WIndows>History (i.e. open the History Window)

So while we see this relates to Undo it doesn't specifically require using Undo keystrokes.

A workaround I observe is to insert a step between 4 and 5 where we create another (safe) event for OT to revert to via Undo.  Example:  Change Freehand to Normal and swipe at the screen to create an undo point.  No crash will occur when attempting the next undo or opening the History window.

Using Win 10 with latest nightly release as of 2/21/2018.

shun-iwasawa:
This will fix #1821 .

### DIFF
--- a/toonz/sources/tnztools/fullcolorerasertool.cpp
+++ b/toonz/sources/tnztools/fullcolorerasertool.cpp
@@ -428,6 +428,12 @@ void FullColorEraserTool::onActivate() {
   }
 
   m_brushPad = getBrushPad(m_size.getValue(), m_hardness.getValue() * 0.01);
+
+  // setting m_level in resetMulti() will take care when the tool is switched
+  // via shortcut ( i.e. the case skipping onEnter() )
+  resetMulti();
+  // clear previous polyline when the tool is activated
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1824, opened 2018-02-22
Enable to Skip the Version Check

shun-iwasawa:
For now OT checks the latest version on launch and displays the notification popup if the newer version is available.
However, especially during production, users do not always want to update the software in order to maintain compatibility.
This PR enables to choose whether to execute such version check. Disabling this option will avoid the trouble of having to cancel the popup on launch.

This option is available in `Preferences > Version Control` and also in the popup.

image

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -506,6 +506,12 @@ public:
   bool isAutomaticSVNFolderRefreshEnabled() const {
     return m_automaticSVNFolderRefreshEnabled;
   }
+
+  void enableLatestVersionCheck(bool on);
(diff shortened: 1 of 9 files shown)


shun-iwasawa:
Jenkins

RodneyBaker:
Yes, thank you!

DarrenTAnims:
Another small change could be to have a check for new version button in the toolbar (under a help menu). Without this, I'd need to run OT, tick the box in prefs, then close and reopen it. It'd be useful to just be able to manually check it whenever i'd like to.

masafumi-inoue:
The "check for new version" button can be another development in the future.
LGTM for this.

PR #1826, opened 2018-02-23
New Option for Match Line Feature : Merge Inks

shun-iwasawa:
This PR introduces a new option for Ink Usage property of Apply Match Line feature; Merge Inks.

For now, the existing options for the Ink Usage property are as follows:

- **Add Match Line Inks** : All match line inks will be added as new styles, to "match lines" page in the base palette.
- **Use Ink** : All lines in the match line image will be replaced by specified style of the base palette.

And, this is the new option:
- **Merge Inks** : If the base level has the same style as the match line ink (i.e. with the same index and the same color), the existing style will be used. Otherwise, a new style will be added to "match lines" page.

matchline_options

The above image shows how 3 options will work.
Please note that in the "Merge Inks" result, styles # 1 (black), # 2 (red) and # 3 (blue) are shared by the base and the match line images, since they are identical in both palettes. On the other hand, style # 4 (magenta) of the match line image is added to "match lines" page of the result palette since it does not exist in the base palette.

This option is useful for compositing match line image containing both black and self colored lines. You can merge palettes with keeping their colors unchanged.

This feature is requested by some Japanese animation studio.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1827, opened 2018-02-23
Make command bar icons consistent

artisteacher:
This is my first try at developing - please let me know if I did anything incorrectly.

The new level icons are now svg instead of png and updated to match the load level icon. Also, the time stretch icon is now the same height as the level icons.


turtletooth:
LGTM.  Thanks @artisteacher for the updates to the consistency.

Issue #1809, opened 2018-02-15
Extra periods in name stops image from loading


artisteacher:
In OS X, screenshots are named in the following format: “Screen Shot 2018-01-17 at 8.20.15 PM.png”. The placement of the 2 periods prevents OT from recognizing that the png is a png and so they will not load properly. Screenshots are visible in the browser but the icons do not display (just a white square with a zig-zag). After loading the screenshot, the file name displays as red in the scene cast and frame displays as missing in the Xsheet (no icon in the header, nothing visible in the viewer, and red text in the cell).

Screenshots that have been renamed without extra periods, display & load without issue. This is probably related to #1382 and is likely an issue for all OS with any files that have a similar name format (with extra periods).

romeojulietthotel:
Yes, the TFilePath is the problem:
https://github.com/opentoonz/opentoonz/issues/1382#issuecomment-320560810

That comment summarizes the problem correctly. The solution is also proposed but I don't know if that's the correct solution.

artisteacher:
Closing since this is a duplicate of #235.

Issue #1227, opened 2017-05-26
Crash when accessing alias folders in Browser


artisteacher:
When using the browser room to navigate to a file outside of the stuff folder (ie. Downloads or etc.), inside of My Computer are 2 paths. The first allows you to eventually navigate to the desired folder. If you try to access Macintosh HD, inside of Volumes (the 2nd path), there is an immediate crash. 
<img width="229" alt="screen shot 2017-05-25 at 8 57 33 pm" src="https://cloud.githubusercontent.com/assets/21159116/26476631/17fdf190-418e-11e7-9975-f0bdc3afaf84.png">

artisteacher:
I just realized that the Macintosh HD accessed through Volumes is an alias. Another alias folder “User Guides and Information” also causes a crash. Perhaps this is actually related to #984, though ffmpeg doesn’t look like an issue in the crash report.

artisteacher:
Could making alias folders/files greyed out/inaccessible prevent the crash?

artisteacher:
Maybe this is related...
[https://bugreports.qt.io/browse/QTBUG-23927](url)

PR #1833, opened 2018-02-27
Fix number+letter filename save and detection

### LINKED ISSUE #235: Cannot handle filepath including multiple "."
Loading `foo.bar.baz.png` results in getting blank image. 

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/33306998-cannot-handle-filepath-including-multiple?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1382: Missing mypaint icons
@blackwarthog thanks so much for the mypaint brushes! They are so fun to paint with. I've noticed a couple icons are missing in the style editor:
screen shot 2017-08-06 at 2 36 37 pm
screen shot 2017-08-06 at 2 36 49 pm
The brushes themselves seem to work just fine.

I've tried a clean reinstall, restarting the computer & etc. but the problem remains. It looks like the icons actually exist in the stuff folder's mypaint folders, so maybe it's just a path/naming issue?
screen shot 2017-08-06 at 2 45 54 pmscreen shot 2017-08-06 at 2 46 37 pm

I'm using the latest nightly build on OS X 10.12.6.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/47989132-missing-mypaint-icons?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1693: Opentoonz creates drawings with letters in name but doesnt reopen them
When autocreation is set to 'Use Xsheet as Animation Sheet' opentoonz can create drawings with names like 2a, 2b etc, but after saving and reopening the scene does not load them anymore. The drawings themselves are still present in the extras folder, opentoonz just doesnt recognise them. The only workaround is to renumber the level before saving and closing.

frames
image

I tried using regular expressions in Level Settings by File Format but it made no difference. This expression should have worked: `.+[0-9a-z]{4,5}\.tif` (#644)

Tested on windows 7 in OT 1.2 and OTX 12/5.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/52765229-opentoonz-creates-drawings-with-letters-in-name-but-doesnt-reopen-them?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1761: Frame sequence import issues

### LINKED ISSUE #1768: Image sequence loads multiple times when loaded from system file manager
Dragging and dropping a properly named image sequence (ie. with 25 images) from Finder (Mac OS) into a scene results in the sequence ended up loading into a column for each image (ie. 25 columns). In each column all the images displayed as a single level. If I had loaded the same sequence from OpenToonz's Browser, there would be a single column with all of the images in a single level. This is in 1.2, though it may happen in earlier versions.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/54262352-image-sequence-loads-multiple-times-when-loaded-from-system-file-manager?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

manongjohn:
This PR fixes #1693 

The original function for checking the sequence number format was not accounting for the optional letter at the end.  As such it treated these files in an odd way resulting in either the file being deleted while saving or loading as a separate level when brought into OT.

I corrected it to make sure the sequence number followed the format "####[a]" (4-digit number + optional letter). 

This PR also fixes #1382 

I added a new function to skip checking for sequence number in file names for specified file types.  Currently that is only for .myb files used by mypaint brushes and OT's .tpl, .plt and .tlv types..  Additional file types can be added as needed, when discovered

This PR also fixes #1761 

Added logic to better detect and keep track of sequence numbers in filenames. There can be 0-to-many leading zeros. Default for new files is still 4. It is still expected that sequence numbers are between '..' or '_.'. (i.e A.001.tif, A_001.tif - Note, these are considered 2 separate levels)

Additionally updated level strip to show frame letters.

This PR also fixes #1768 

Added a dialog box that pops up when it detects a file in the dragged/dropped file list (another file in the same level sequence) will cause a duplicate of the same level to be created. This gives the user the option to allow it or not. 

This PR also fixes #235, #1809 

This really wasn't an issue of multiple dots. The problem became what was between those dots.  OT only expected a frame number between the last 2 dots, but found regular text instead and didn't handle the filename properly from there.  It would take out the frame number for some operations and then couldn't find anything.

Corrected to only strip frame numbers from filenames in which the text between the last 2 dots is a valid frame number format.  Otherwise it retains the text.

### DIFF
--- a/toonz/sources/common/timage_io/tlevel_io.cpp
+++ b/toonz/sources/common/timage_io/tlevel_io.cpp
@@ -139,6 +139,11 @@ TLevelP TLevelReader::loadInfo() {
         m_frameFormat = TFrameId::FOUR_ZEROS;
       else
         m_frameFormat = TFrameId::UNDERSCORE_FOUR_ZEROS;
+    } else if (ws.rfind(L'0') == 1) {  // leads with any number of zeros
+      if (ws.rfind(L'_') == (int)wstring::npos)
+        m_frameFormat = TFrameId::CUSTOM_PAD;
+      else
+        m_frameFormat = TFrameId::UNDERSCORE_CUSTOM_PAD;
     } else {
       if (ws.rfind(L'_') == (int)wstring::npos)
         m_frameFormat = TFrameId::NO_PAD;
--- a/toonz/sources/common/tsystem/tfilepath.cpp
+++ b/toonz/sources/common/tsystem/tfilepath.cpp
@@ -811,7 +811,7 @@ TFilePath TFilePath::withFrame(const TFrameId &frame,
   if (!isFfmpegType() && checkForSeqNum(type) && isNumbers(str, k, j))
     hasValidFrameNum = true;
   std::string frameString;
-  if (frame.isNoFrame() || !hasValidFrameNum) {
+  if (frame.isNoFrame() || (!frame.isEmptyFrame() && !hasValidFrameNum)) {
     if (k != (int)std::wstring::npos) {
       std::wstring wstr = str.substr(k, j - k);
       std::string str2(wstr.begin(), wstr.end());
--- a/toonz/sources/common/tsystem/tfilepath.cpp
+++ b/toonz/sources/common/tsystem/tfilepath.cpp
@@ -580,11 +580,10 @@ std::wstring TFilePath::getWideName() const  // noDot! noSlash!
   i                = str.rfind(L".");
   if (i == (int)std::wstring::npos) return str;
   int j = str.substr(0, i).rfind(L".");
-  if (j != (int)std::wstring::npos)
-    i = j;
-  else if (m_underscoreFormatAllowed) {
-    j            = str.substr(0, i).rfind(L"_");
-    QString type = QString::fromStdString(getType()).toLower();
+  if (j != (int)std::wstring::npos) {
+    if (checkForSeqNum(type) && isNumbers(str, j, i)) i = j;
+  } else if (m_underscoreFormatAllowed) {
+    j = str.substr(0, i).rfind(L"_");
     if (j != (int)std::wstring::npos && checkForSeqNum(type) &&
         isNumbers(str, j, i))
       i = j;
@@ -649,7 +648,7 @@ TFilePath TFilePath::getParentDir() const  // noSlash!
 
 bool TFilePath::isLevelName() const {
   QString type = QString::fromStdString(getType()).toLower();
-  if (isFfmpegType() || type == "myb") return false;
+  if (isFfmpegType() || !checkForSeqNum(type)) return false;
   try {
     return getFrame() == TFrameId(TFrameId::EMPTY_FRAME);
   }
@@ -806,13 +805,21 @@ TFilePath TFilePath::withFrame(const TFrameId &frame,
       return TFilePath(m_path + ::to_wstring(ch + frame.expand(format)));
   }
 
-  std::string frameString;
-  if (frame.isNoFrame())
-    frameString = "";
-  else
-    frameString = ch + frame.expand(format);
-
   int k = str.substr(0, j).rfind(L'.');
+
+  bool hasValidFrameNum = false;
+  if (!isFfmpegType() && checkForSeqNum(type) && isNumbers(str, k, j))
+    hasValidFrameNum = true;
+  std::string frameString;
+  if (frame.isNoFrame() || !hasValidFrameNum) {
+    if (k != (int)std::wstring::npos) {
+      std::wstring wstr = str.substr(k, j - k);
+      std::string str2(wstr.begin(), wstr.end());
+      frameString = str2;
+    } else
+      frameString = "";
+  } else
+    frameString = ch + frame.expand(format);
 
   if (k != (int)std::wstring::npos)
     return TFilePath(m_path.substr(0, k + i + 1) + ::to_wstring(frameString) +
--- a/toonz/sources/common/tsystem/tfilepath.cpp
+++ b/toonz/sources/common/tsystem/tfilepath.cpp
@@ -73,7 +73,8 @@ bool isNumbers(std::wstring str, int fromSeg, int toSeg) {
 }
 
 bool checkForSeqNum(QString type) {
-  if (type == "myb") return false;
+  if (type == "myb" || type == "tlv" || type == "pli" || type == "tpl")
+    return false;
   return true;
 }
 };
@@ -792,7 +793,7 @@ TFilePath TFilePath::withFrame(const TFrameId &frame,
   int j          = str.rfind(L'.');
   const char *ch = ".";
   // Override format input because it may be wrong.
-  format = frame.getCurrentFormat();
+  if (!isFfmpegType() && !checkForSeqNum(type)) format = frame.getCurrentFormat();
   if (m_underscoreFormatAllowed && (format == TFrameId::UNDERSCORE_FOUR_ZEROS ||
                                     format == TFrameId::UNDERSCORE_NO_PAD ||
                                     format == TFrameId::UNDERSCORE_CUSTOM_PAD))
--- a/toonz/sources/common/tsystem/tfilepath.cpp
+++ b/toonz/sources/common/tsystem/tfilepath.cpp
@@ -793,7 +793,8 @@ TFilePath TFilePath::withFrame(const TFrameId &frame,
   int j          = str.rfind(L'.');
   const char *ch = ".";
   // Override format input because it may be wrong.
-  if (!isFfmpegType() && !checkForSeqNum(type)) format = frame.getCurrentFormat();
+  if (!isFfmpegType() && checkForSeqNum(type))
+    format = frame.getCurrentFormat();
   if (m_underscoreFormatAllowed && (format == TFrameId::UNDERSCORE_FOUR_ZEROS ||
                                     format == TFrameId::UNDERSCORE_NO_PAD ||
                                     format == TFrameId::UNDERSCORE_CUSTOM_PAD))
--- a/toonz/sources/common/tsystem/tfilepath.cpp
+++ b/toonz/sources/common/tsystem/tfilepath.cpp
@@ -616,7 +616,7 @@ std::wstring TFilePath::getLevelNameW() const {
   if (j == i || j - i == 1)  // prova.tif o prova..tif
     return str;
 
-  if (checkForSeqNum(type) && !isNumbers(str, i, j)) return str;
+  if (!checkForSeqNum(type) || !isNumbers(str, i, j)) return str;
   // prova.0001.tif
   return str.erase(i + 1, j - i - 1);
 }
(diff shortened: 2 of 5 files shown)


RodneyBaker:
Thank you for this fix and addressing the underlying problem!

artisteacher:
Any chance that this will impact #1768? Currently, if I load an image sequence by dragging & dropping it (without using the browser), the full sequence loads into a column for each individual image. Ie. An image sequence with 12 images will load as a complete image sequence in 12 columns.

manongjohn:
@

manongjohn:
@artisteacher No, not yet, anyway.

I have discovered you only need to drag 1 of the sequence files from the browser into OT and it will pull in all the rest by default.  It's not particularly intuitive in that respect.  Most people will assume you need to drag/drop all sequence files to load them so for every file it processes in this manner, it loads the entire set again.

I will see if I can fix this as well.

konero:
@manongjohn 

> I have discovered you only need to drag 1 of the sequence files from the browser into OT and it will pull in all the rest by default. It's not particularly intuitive in that respect. Most people will assume you need to drag/drop all sequence files to load them so for every file it processes in this manner, it loads the entire set again.

This is expected behavior IMO, most of the software I use from 3D to video editing do it this way.

manongjohn:
@konero Ok. 

I would still expect that in a selected file set, OT should still recognize that it loaded the file already and not try to load it again.

artisteacher:
If the existing behavior is industry standard, maybe there should be a dialog asking if the sequence should be loaded once or multiple times? Something similar to the behavior of load/import dialogue and preference.

konero:
@artisteacher I'm confused, your issue states if you drag in a single frame from a correctly named sequence of frames, each frame loads as separate columns, but this doesn't happen for me, they will load into one column. So, my level of 240 frames, loads into a single column with 240 cells if I drag in either the folder or the first frame, that's expected.

Is this a Mac issue? 😐 

So when you say...

> maybe there should be a dialog asking if the sequence should be loaded once or multiple times

I don't understand.

manongjohn:
@konero You may be used to dragging/dropping just 1 image of a 10 image seq to load all 10 images into 1 column. 

People who are not familiar with that will attempt to drag/drop all 10 at once. This creates 10 exact columns containing the full sequence of 10 images.

It happens in Windows too, not just macs.

artisteacher:
It's helpful to know that dragging/dropping a single image will load the entire sequence. I've never experienced that in other software & so had assumed I needed to drag/drop all of the images. Thanks @manongjohn for resolving so many loading issues! 

I just want to add, that while not directly related, #1809 and #1227 are both other file access issues.

RodneyBaker:
Yes, the drag/drop into OpenToonz (from outside of OpenToonz) and having folder contents that should be recognized as sequential load into separate columns is present on the PC as well.

I have this with a number of filenames but the one I format I've used for many years is 'image000.png' where the zeros are padding that tell the other programs renderer how to number the sequence.  

In addition to the problem of starting a sequence with zero OpenToonz is restrictive on what it recognizes as a sequence.

manongjohn:
@artisteacher Thanks.  #1809 is likely related to #235 which I am actively trying to fix now also.  If/when I get the fix working, I'll see if it fixes both.  #1227 appears to be just a Mac issue which I unfortunately cannot help with.

@RodneyBaker The current sequence number detection for OT is definitely overly restrictive expecting it to just be between 2 dots or underscore-dot and can only ever be 4 digits.  Anything beyond that, like a frame letter after the seq num, was treated as a non-sequenced file which explains why they might load into separate columns incorrectly. Hopefully the changes I've made will fix 90% of those issues.

I'm curious what sequence number formats other commonly used animating software generates and which ones we can't detect properly...yet.

RodneyBaker:
Thanks for all you have done regarding the issue. It's much appreciated.
The biggest issue I have with OT being overly restrictive is that the 8.3 filenaming convention is a very old one and one that can be expected to be found a lot out in the wild.  While most programs were upgraded long ago to allow for longer names and adding additional periods that isn't the safe route.

I looked for a RegExp that would allow other approaches but couldn't find anything that worked.
I investigated RegExp that might allow for 8.3 sequences to import properly but I think the trailing zero was the primary deal breaker in that case.  The following is what I show as being the default RegExp in OpenToonz for loading 'All' files:

.*[1-9].*

It's quite possible that I might have had success if I would not have had the trailing zeros so I should give the RegExp a try again.  One would think that *[1-9].* should work.

manongjohn:
Ok, that last bug was a pain in butt to find. :) With that, I think I've found and fix most open issues surrounding filename frame numbers or what OT would incorrectly think was a frame number and is not.  The only issue I didn't fix is the 0000 one.  Not touching that one.

So if the powers that be find some time to review, this PR is ready, barring someone finding another open issue I missed.

konero:
@manongjohn 

> You may be used to dragging/dropping just 1 image of a 10 image seq to load all 10 images into 1 column.

I was taught how to use Harlequin by a TV studio, it was the practice of the day.

> People who are not familiar with that will attempt to drag/drop all 10 at once. **This creates 10 exact columns containing the full sequence of 10 images**.

Ahhh, I understand now, thanks, I agree that'd be a pain. 🥂 

I meant before it was expected loading in a single frame loads in the entire sequence, I didn't mean to say the column issue was expected, just to be clear.

I'll get my coat. 🤣

turtletooth:
@manongjohn I think this PR breaks exporting.

manongjohn:
@turtletooth Will look onto.

manongjohn:
Fixed. Only impacted rendering to sequenced files like tif, png. Files like mp4, avi appear to be ok. Let me know otherwise.

shun-iwasawa:
I'll review this.

shun-iwasawa:
The second interim report:

- **Somewhat Critical** : The issue #235 seems not to be resolved. I try making a single image `too.many.dots.image.tif` . OT seems to misunderstand the image as a level with sequential frames and displayed as `too.many.dots..tif` in File Browser. OT fails to load such file properly.
- **Minor** : If the first frame of a level has the optional letter in the frame number, such level cannot be loaded in Flipbook ( by neither `View...` menubar command nor `Load / Append Images` command in the right-click menu of Flipbook ).
- **Minor** : A column generated as a result of `Convert to Vectors...` command fails to reproduce the cell order the source column if it contains a cell of which frame number has the optional letter.
- **Minor** : `Remove dot before frame number` option of `Convert...` command does not work properly if the level has the optional letter in the frame number.
- **Minor** : Some input fields for frame number do not accept "number+letter" format yet, such as `From frame` and `To frame` fields in `Auto Input Cell Number...` popup.

Again, I think minor problems can be solved later since it works OK on using in conventional way. I just list them as a reminder. Especially the latter two problems are related to the features which I implemented so I'll tackle with them after merging this.

I'll continue to test this.

shun-iwasawa:
Regarding the fix for #1768 :

>Added a dialog box that pops up when it detects a file in the dragged/dropped file list (another file in the same level sequence) will cause a duplicate of the same level to be created. This gives the user the option to allow it or not.

I'm wondering if there is a scenario users will choose "Allow Dups" option, since it seems to be hard for users to expect the result of duplicating columns by drag&dropping multiple frames in the same level...

IMO it's somewhat understandable if there would be a option to choose whether exposing subsequence frames (= only dragged/dropped frames) of the level into Xsheet, instead of whole frames.

manongjohn:
@shun-iwasawa Thanks for the thorough review.  I will work on fixing the critical issues you've noted so far.  Will review the minor ones and if I see if it's a simple enough fix to do with this PR.

shun-iwasawa:
This is the final report of my first review:

- **Minor** : Rendering spritesheet with "Individual" option will output sequential images with the names like `filename_number.png` . For now `number` has no padding and is stated from zero, like `character_0.png` , `character_1.png` , `character_2.png` , and so on. 
  From this PR the four-digits restriction will be abolished (That's great !). Thus these images will be recognized as one level. However due to the sequence starting from zero, the first frame cannot be loaded with warning popup.
  I'm not sure if the files starting from zero is necessary for game developers, but if it can be changed to starting-from-1 IMO it would be useful as they can load and preview the whole files in OT. 
- **Minor** : (NOTE - This is for my personal reminder) After introducing this feature completely, the preference option `Show "ABC" Appendix to the Frame Number in Xsheet Cell` can be deprecated.

manongjohn:
@shun-iwasawa 
I've fixed both critical issues you've noted.  I did look at the minors and agree those can be done in another PR so I won't be trying to fix any of them at this time.

My goal with this PR was just to make sure sequenced files and files with frame letters can be loaded/saved properly.

As for #1768, I'll defer to your guidance. If you feel the popup is unnecessary and should only ever load the sequence once, I can easily remove it.  If anything the popup serves more as a hint to the user who may be unaware you only need to drag 1 file.

shun-iwasawa:
@manongjohn Thank you for the fix, I'll check it.

> If anything the popup serves more as a hint to the user who may be unaware you only need to drag 1 file.

Understood. I agree it would be more user-friendly to have different response when dropping single frame vs multiple frames. Then let's leave the popup as-is.

Jenkins

shun-iwasawa:
Confirmed that the latest modification fixed the critical problems.
I'll post a new issue with the list of minor items soon.

Thank you @manongjohn for the big improvement! LGTM

jpturcotte:
This is tremendous work. Thank you very much, @manongjohn!

PR #1836, opened 2018-03-02
Fix Cell Dragging not to Duplicate Fxs with No Inputs

### LINKED ISSUE #1819: Bug expanding the .tnz scene file to the point it becomes unusable
I've been working on a small project. It has a sub-xsheet, some auto-animations (using the animate tool) and a few effects (glow and invisibility). The whole animation lasts 30 seconds or so. It got to the point where everytime I tried to do anything (like change rooms to the effects room or start the preview) OpenToonz would freeze. Sometimes it would come back after a few minutes, sometimes not. It's like I've hit some kind of wall with it, by adding the last couple of levels. It's not using many resources either: 17% cpu, total 28% memory (inc my other apps, mostly used by Chrome).

I don't understand the problem fully yet, but I keep backups as I work, so I rolled back a few versions to one that worked ok, then started working for 2 mins. Then, when I looked at the scene file (.tnz) it had grown from 620k to 18,119kb. Quite an increase. Looking at the difference between them showed that the new file had tons of new empty fx nodes toward the end.
<fxnode>
          <STD_fadeFx id='17782'/>
        </fxnode>
        <fxnode>
          <STD_fadeFx id='17783'/>
        </fxnode>
        <fxnode>
          <STD_fadeFx id='17784'/>
        </fxnode>
...

I'll keep looking to see if I can pin down the problem, but it seems that I can no longer edit this project.

shun-iwasawa:
This will fix #1819 .
The fx duplication occurs when the preferences option "Xsheet > Cell-dragging Behavior" is set to "Cells and Column Data".

With this option activated, dragging cells will move relevant fxs accompany with it.
The problem occurs because the algorithm for finding the relevant fxs did not expect the case that fx has input ports, but no fxs are connected to them. I fixed the algorithm for it.

Please note this does NOT fix the behavior that dragging cells always increment fxid of connected fxs. 
I think it would better to make this more adaptive, avoiding increment if it is not necessary.

### DIFF
--- a/toonz/sources/toonzqt/stageobjectsdata.cpp
+++ b/toonz/sources/toonzqt/stageobjectsdata.cpp
@@ -31,24 +31,42 @@
 
 namespace {
 
+enum FxCanGenerateState {
+  NoInput = 0,  // This fx has input port, but no connected fx - state will not
+                // transmit to parents and leave the decision up to another
+                // input fx
+  CanGenerate,  // This fx can generate only from fxsSet input
+  CanNotGenerate  // This fx contains input fx other than fxsSet - state will
+                  // transmit to parents
+};
+
 //! Returns whether ALL input chains from fx pass through fxsSet - ie if
 //! fxsSet can completely generate fx's input.
 //! \note Columns are assumed to be generated only if they are in fxsSet.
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1841, opened 2018-03-03
Remove unused vars

romeojulietthotel:
and some commented include statements.


shun-iwasawa:
Thank you for refactoring. 
The removed double byte glyph was "π" , so I think I'll add "pi" to the comment afterwards.
LGTM

romeojulietthotel:
Ha! Did not realize that was the Pi symbol. I recall checking quickly but didn't pay close enough attention.

Issue #984, opened 2017-01-12
Crash when using browser through load level


artisteacher:
On Mac OS X 10.11.6, with 1.1.1 and the latest nightly build. A student ran into an issue loading a level and trying to select a file from the desktop. As soon as the desktop folder is selected, a crash occurs. I'm not sure if this is something specific to her project or maybe the desktop files. Other students and myself have successfully loaded images from outside of the project folder without issue. Crash report is below.

[browsercrash.txt](https://github.com/opentoonz/opentoonz/files/702585/browsercrash.txt)

jabarrera:
A fast looking to the report it seems something related with **FFMPEG**. I have not seen into code, but maybe FFMEPG is used to get information from some type files (gif, mp4, etc...).

If the user have a gif type file in the Desktop, maybe it is corrupted and "break" FFMPEG and breaking Opentoonz.

If you found the file that is breaking it, you could post it here to look why its devilish behavior. : )

artisteacher:
I'll take a look. Thanks for the suggestion.

artisteacher:
So the faulty file is actually a shortcut to a gif.

![doggo--1](https://cloud.githubusercontent.com/assets/21159116/22064567/2a267a54-dd52-11e6-9ea5-ff0c33562f0d.gif)

artisteacher:
The shortcut/alias also causes a crash when rendering to the same parent folder.

turtletooth:
Added the labels for OSX and Linux.  I realize that Windows can have aliases, but it is less likely to be an issue on Windows.

RodneyBaker:
Is this still an issue?

Interestingly, in the past I've had a similar issue with Windows where to even point the browser at a directory that contained files in a sequence that were (according to OpenToonz criteria) malformed OT would crash.

The better part of a month ago I changed the settings in Preferences to remove any strange custom sequence formatting and I haven't had the issue since (that I can recall).  Then again however, I rarely use OT's browser because I had learned not to use it to avoid crashes.  I learned primarily to load images and image sequences via the Xsheet and/or (less frequently) drag/drop.

I guess what I'm saying is that the way OT expects to see image sequences does play a part here and these might also be 'aliases' that cause errors.

I will attempt to see if I can still reproduce the issue with crashing as that might help resolve the underlying issue in this report.

artisteacher:
I still get a crash when accessing folders/directories with an alias file through the File Browser. I'm no longer getting a crash when rendering to the same directory as an alias file.

PR #1842, opened 2018-03-03
Fix alias crashes

### LINKED ISSUE #1227: Crash when accessing alias folders in Browser

artisteacher:
This fixes #1227 and #984 by removing the alias directories and files (symlinks) from the File Browser for Mac and Linux OS.

### DIFF
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -449,8 +449,12 @@ void TSystem::readDirectory_Dir_ReadExe(TFilePathSet &dst,
 
   QStringList fil =
       QDir(toQString(path))
+#ifdef Q_OS_WIN
           .entryList(QDir::Dirs | QDir::NoDotAndDotDot | QDir::Readable);
-
+#else
+          .entryList(QDir::Dirs | QDir::NoDotAndDotDot | QDir::Readable | QDir::NoSymLinks);
+#endif
+  
   int i;
   for (i = 0; i < fil.size(); i++) {
     QString fi = fil.at(i);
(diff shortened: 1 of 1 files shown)


turtletooth:
Great work @artisteacher - Thanks for this.

manongjohn:
I wonder if this should be expanded to include LINUX systems for someone who might have somehow defined a symbolically linked directory on their system.  I'm guessing they would have the same issue.

artisteacher:
@manongjohn, would that just be a matter of switching the if else statements? If windows, keep original filters, else original filters + no symlinks...

manongjohn:
Yes. That should suffice.

shun-iwasawa:
@artisteacher  Thank you for reporting and tackling with this!

Sorry for responding this so lately, but I found that the same crash can be reproduced on Windows as well by clicking a shortcut item in the file tree.

[QTBUG-23927](https://bugreports.qt.io/browse/QTBUG-23927) which you mentioned in #1227 seems to be due to a bug in QuaZip according to the [comment](https://bugreports.qt.io/browse/QTBUG-23927?focusedCommentId=353000&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-353000). In the Description it also says 
> for normal symlinks this works perfectly .

Please let me investigate further on this and post an alternative PR, which will fix crash keeping SymLink available.

artisteacher:
No problem, thank you for looking into this. It would definitely be preferable to keep the symlinks available.

PR #1843, opened 2018-03-05
Fix Double Fields to Update Correctly When Input Values

shun-iwasawa:
To reproduce the problem:

- Create a vector level.
- Choose Brush tool.
- Try to set the `Smooth` parameter to `1.14` by manually typing in the field in Tool Option.

doublefield

On my environment, the value is changed to `1.13` as soon as the editing is finished.
This small shift may become critical, especially for the `Maximum Gap` parameter to be introduced with #1317 , since it will lose backward compatibility once the value is changed from the initial value of `1.15` .

This PR will fix such problem. It was due to the precision difference between the field and the slider in `DoubleValueField` .


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins again

masafumi-inoue:
LGTM

PR #1844, opened 2018-03-05
Fix png and spritesheet render (depremultiply) by turtletooth (modified)

### LINKED ISSUE #1230: PNG image is rendered with premultiplied alpha

shun-iwasawa:
This will fix #1230 , based on the investigations and efforts done in #1746 by @turtletooth .

### THE ISSUE
- In OpenToonz, images will be rendered with being premultiplied. On the other hand, [PNG Specification](https://www.w3.org/TR/PNG/#6AlphaRepresentation) says that PNG should be straight alpha. So images must be unpremultiplied on saving.

### FOR PNG FILES
- On saving PNG, color channels will be unpremultiplied in `PngWriter::writeLine()` . If the `Alpha Channel` format property is OFF, unpremultiply will not be executed.
- Modified Flipbook to consider the `premultiply` option in the level properties set in `Preferences > Loading > Level Settings by File Format` . It is ON for PNG format by default, so PNG will be premultiplied on loading in the Flipbook as well. This change will also modify the rendered result displayed in the Flipbook with the option `Preferences > Preview > Open Flipbook after Rendering` .

### FOR SPRITESHEET FILES
- I set an appropriate format (`QImage::Format_ARGB32_Premultiplied`) to QImage. By doing this QImage seems to be automatically unpremultiplied on saving to PNG.

If the issue will be settled with this PR, I would like all the bounty to be granted to @turtletooth for tackling this problem. I'm already paid for working on OpenToonz and this PR is based on his work.


shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa - this is a really valuable fix.  I really appreciate it.

PR #1848, opened 2018-03-07
Fix Saving PLI file's AutoClose Value

shun-iwasawa:
This PR will fix saving fractional portion of the Auto Close Tolerance value of Toonz Vector Level.
After merging #1317 this value will be used for identifying the file version. If this value is not equal to the default ( `1.15` ), PLI will be identified as new version (v120) which will have no backward compatibility.

In my environment of Win7, due to the value precision the default value of `m_autocloseTolerance` is set as `1.14999...` . So it should be rounded off before saving the fractional portion or it is saved as `1.14` .


shun-iwasawa:
Jenkins

Issue #1616, opened 2017-11-28
(WIP) Multishade brush tool - to take advantage of the opentoonz raster layer's indexed colors


blurymind:
Multishade tool- what is it the basic idea:
1.You set your cell shade gradients 

![gradientindexes](https://user-images.githubusercontent.com/6495061/33345468-5ff24e18-d484-11e7-9827-07d4256f3cd0.JPG)

2. You use a single brush tool to shade the entire layer
It's a brush tool available in cosmigo pro motion that makes it very easy to do cell shading without worrying about separating different colors. Here is a video:
https://youtu.be/iGFmj4TI3pQ?t=2m09s

every time you do a stroke,the tool picks the pixels and moves their color palette index forward or backward (depending if the user wants t lighten an area or darken it)
![multishadebrush](https://user-images.githubusercontent.com/6495061/33345376-1722e9b8-d484-11e7-9da6-45280a62fea9.gif)

Advantages
- no need to select colors while cell shading with a brush
- no need to carefully respect color islands while cell shading between different colored areas

created a bounty here:
https://www.bountysource.com/issues/52074694-multishade-brush-tool-to-take-advantage-of-the-opentoonz-raster-layer-s-indexed-colors

gab3d:
looks useful.

blurymind:
created a bounty here:
https://www.bountysource.com/issues/52074694-multishade-brush-tool-to-take-advantage-of-the-opentoonz-raster-layer-s-indexed-colors

RodneyBaker:
Definitely would be useful.  :)

jpturcotte:
@RodneyBaker Maybe that could work with separate layers? When you use the tool on a layer, it could apply to every visible layer underneath.

jpturcotte:
@blurymind Could you think of any way to make it non destructive? I don't know how the Undo function works in OpenToonz, but it seems quite limited. Also, don't MyPaint brushes have a lighten/darken function already? Maybe @morevnaproject has an idea. Anyway, this is all really interesting!

blurymind:
@jpturcotte well, the idea of having indexed colors allows you to change any of the colors you are using at any time. So there is no point in having different color shades or fills on different levels :/

The tool has undo and redo, it just doesnt work the way you described- what you request is not possible to do with color index shifting on pixels. It's not what is requested here

jpturcotte:
Great, I am looking forward to your bounty!

beeheemooth:
[...]

PR #1849, opened 2018-03-07
New option for Toonz Raster Brush Tool : Draw Order - Palette Order

shun-iwasawa:
This PR will introduce a new option to Toonz Raster Brush - ~~"Selective : Style Order".~~
**UPDATE:** This item is renamed to "Draw Order", with 3 options `Over All` , `Under All` and `Palette Order` .

We already have "Selective" option for the Toonz Raster Brush tool, which allows the drawing operation without affecting already drawn lines.
**UPDATE:** `Over All` and `Under All` options correspond to deactivate and activate the conventional "Selective" option, respectively.

The ~~"Style Order"~~ "Palette Order" option will be added to it. If the ~~"Style Order"~~ "Palette Order" is selected, stacking order of the stroke will be the same as the order of the styles in the palette. Therefore, style order in the palette will act as pseudo-layers.

Please note that the "style order" means the order in the palette page, regardless of the style index.
Which means that you can control the stacking order of the new stroke by moving the current style in the page before drawing.

This option is useful especially for drawing with switching between solid (black) ink and self-colored (red/blue) inks. You can draw solid lines always on top without any care of "Selective" option.

style_order

### DIFF
--- a/toonz/sources/include/toonz/rasterstrokegenerator.h
+++ b/toonz/sources/include/toonz/rasterstrokegenerator.h
@@ -35,8 +35,8 @@ class DVAPI RasterStrokeGenerator {
   int m_selectedStyle;
   bool m_keepAntiAlias;
   bool m_doAnArc;
-  bool m_useStyleOrder;  // Used in the Selective option of Brush Tool,
-                         // use style order to define line stacking order
+  bool m_isPaletteOrder;  // Used in the Draw Order option of Brush Tool,
+                          // use style order to define line stacking order
   QSet<int> m_aboveStyleIds;
 
   // Ricalcola i punti in un nuovo sistema di riferimento
@@ -51,14 +51,14 @@ public:
   RasterStrokeGenerator(const TRasterCM32P &raster, Tasks task,
                         ColorType colorType, int styleId, const TThickPoint &p,
                         bool selective, int selectedStyle, bool keepAntialias,
-                        bool useStyleOrder = false);
+                        bool isPaletteOrder = false);
   ~RasterStrokeGenerator();
   void setRaster(const TRasterCM32P &ras) { m_raster = ras; }
   void setStyle(int styleId) { m_styleId = styleId; }
   int getStyleId() const { return m_styleId; }
   bool isSelective() { return m_selective; }
 
-  bool useStyleOrder() { return m_useStyleOrder; }
+  bool isPaletteOrder() { return m_isPaletteOrder; }
   void setAboveStyleIds(QSet<int> &ids) { m_aboveStyleIds = ids; }
 
   // Inserisce un punto in "m_points"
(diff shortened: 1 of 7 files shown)


shun-iwasawa:
Jenkins

DarrenTAnims:
Quick question - Is it too late to suggest a change to the text shown to make it easier for new people to recognise and use it?
How about changing the title of the dropdown to "draw order".
And change the options to: "Under all", "Over all", "Palette order" .

gab3d:
@DarrenTAnims 
sounds good to me

masafumi-inoue:
LGTM

blurymind:
For a moment I thought that you have implemented this:
https://github.com/opentoonz/opentoonz/issues/1616
gradientindexes
multishadebrush

It's a brush mode that uses the palette order to change pixel colors.
It would be perfect for OT's raster brush!

shun-iwasawa:
@blurymind 
Yes, now I'm thinking of introducing something similar to this feature.
Not as the special brush, but as some kind of extension of palette data which specifies the relation of gradient between styles.

blurymind:
@shun-iwasawa This would be an incredible addition to OT and seems like a good fit for the brush tool - as a mode.
But if you are looking into desigining it into the palette data, I think that your implementation will probably be even better than the one in Pro Motion

They describe the multishade mode like this:

> Multi Shade
> Works the same as Shade but uses mor than a single gradient at a time. It only works for gradients that are set up for multi shade paint mode. Because you might want to have gradients that are not touched by the multi shade you must enable a gradient that may be used. Just go to the gradient editor and activate multi shade paint mode for the gradients that may be used for this paint mode.
> 
They also have a window to set an independent indexing order for multishade painting btw
https://www.cosmigo.com/promotion/docs/onlinehelp/GradientEditDlg.htm

blurymind:
@shun-iwasawa  oh god, I just realized that I have already posted about this here. Sorry for spamming you. I have totally forgotten

beeheemooth:
@shun-iwasawa I've got small but useful suggestion: to add also reverse order of the palettes Thus you can easily draw intertwining patterns.
ezgif-3-4b322d78ff

RodneyBaker:
Adding a postmortem note based on some feedback and testing.

The copy/paste process does not appear to respect palette order which is reasonable.
It just pastes the copied selection over the target.
It would be very useful to be able to paste UNDER the target.

I'm mainly thinking in terms of anime usage where a face might be on one level and a mouth on another and while merging and match lines is an option a quicker way can often be to simply copy/paste one into the other.  

In most cases a Merge TLV Levels will work here but in those cases column order is important.
It makes me think that there might need to be an option to Merge Up or Merge Down where to the user it's all pretty much the same but in the code the columns would flip stacking order.

PR #1851, opened 2018-03-08
Enable Vector Selection Right-click Menus in All Section Modes

shun-iwasawa:
For now, some right-click menu commands of Vector Selection Tool such as `Group`, `Bring to Front`, `Send Back`, etc. are available only when the `Mode` option is set to `Standard` .

This fix will modify it so that all commands can be used in any selection mode.
Please note that for now these commands are applied to the selected strokes only in the current frame even if you are selecting strokes across multiple frames by using `... on Selected Frames` or `... on Whole Level` mode.

Also I modified the undo text for `MoveGroupsUndo` , adding the type of movement.

image

### DIFF
--- a/toonz/sources/tnztools/imagegrouping.cpp
+++ b/toonz/sources/tnztools/imagegrouping.cpp
@@ -216,7 +216,15 @@ public:
 
   int getSize() const override { return sizeof(*this); }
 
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1345, opened 2017-07-26
WIP: EnumProperty: translation and icon

podsvirov:
Add support to translate EnumProperty items and
show icon.

Tested on `Active Axis` property for `Edit` tool.

Any feedback welcome!

PR #1863, opened 2018-03-12
EnumProperty: translation and icon by podsvirov (modified)

shun-iwasawa:
This PR is based on #1345 by @podsvirov which enables to translate and to show icons for `EnumProperty` items. (Thank you @podsvirov for your contribution!)

image

Please note that for adding translatable string use `TEnumProperty::setItemUIName()` NOT in ctor but in `updateTranslation()` of each tool, since the tool objects are constructed before applying the translation files.

To add icon for enum item, specify a file name in the second argument of `TEnumProperty::addValue()` . Then `Resources:\filename.svg` will be loaded for icon with 17x17 pixels size.

This PR also includes updated Japanese translation for all enum items.


shun-iwasawa:
Jenkins

beeheemooth:
@shun-iwasawa Sorry, but as I understand, the main idea of these icons was that they replace this drop-down list, and were directly on the Tool Option Bar. Or it can be configured in Preferences?

shun-iwasawa:
@beeheemooth 

> the main idea of these icons was that they replace this drop-down list, and were directly on the Tool Option Bar

Yes, AFAIK @Agnyy has already been tackling such modification and according to [his comment](https://github.com/opentoonz/opentoonz/pull/1345#issuecomment-318475571) this PR is necessary as the first step to continue his work.

artisteacher:
If this is the first step towards getting the animate modes directly visible in the tool option bar, I'm excited!

gab3d:
@shun-iwasawa 
great improvement for all non-English speaking users!
besides, if it bring us one step closer to having an easier animate tool axis selection, much better then!

thank you for your continuous commitment to make OT a better and more usable program, as well as for listening to users requests! :)

Orphanlast:
I think that if these icons appeared next to the animate tool when the animate tool is selected (Similar to how there's multiple types of brushes and stuff in Adobe Photoshop) that would look really sleek.

shun-iwasawa:
I found that OT already has icons representing "Position", "Scale" and "Rotation" which is displayed in the Selection Tool option:

Position : image / Scale :  image / Rotation :  image

I think it would better to standardize icons with the same meanings in order to gain consistency... Any thoughts?

turtletooth:
I think the ones you have at the top of this PR are better than what is already included.

gab3d:
I agree with turtletooth.

shun-iwasawa:
Conflicts resolved. Jenkins again

turtletooth:
Tested and this works well.  Thanks @shun-iwasawa

shun-iwasawa:

PR #1866, opened 2018-03-13
Fix Crash on Adding Batch Rendering Task from Network Drive

shun-iwasawa:
This problem was reported by some Japanese animation studio.

To reproduce (on local, on Windows) : 

- Make some folder to be shared.
  - Locate the folder you want to share and right click on it.
  - Choose `Share with...` and then select `Specific People`. Select `Everyone` .
- Map the folder to a network drive. 
  - For instance, map `\\localhost\foldername` to a drive `Z:`.
- Put some scene file in the folder. 
- Launch OpenToonz.
- On Tasks window,  click `Add Render Task` button and specify the scene in the drive `Z:` .
- OT crashes before the tasks being added to the list.

This crash is due to wrong translation of the network drive to UNC path.
The line 49 in `uncpath.cpp`, `puni->lpUniversalName` is LPWSTR (= pointer to wchar_t ). However `::to_string(puni->lpUniversalName)` invokes `to_string(void*)` instead of `to_string(std::wstring)` and returns wrong file name.

### DIFF
--- a/toonz/sources/common/tsystem/uncpath.cpp
+++ b/toonz/sources/common/tsystem/uncpath.cpp
@@ -46,7 +46,7 @@ TFilePath TSystem::toUNC(const TFilePath &fp) {
 
       switch (dwResult) {
       case NO_ERROR:
-        return TFilePath(::to_string(puni->lpUniversalName));
+        return TFilePath(::to_string(std::wstring(puni->lpUniversalName)));
 
       case ERROR_NOT_CONNECTED:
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
OK, [Appveyor updated Qt version to 5.9.4](https://www.appveyor.com/updates/) yesterday!

masafumi-inoue:
LGTM

PR #1867, opened 2018-03-13
Update Appveyor's Qt Version

shun-iwasawa:
[Appveyor updated its Qt version to 5.9.4 yesterday.](https://www.appveyor.com/updates/)
From this PR, I try using `C:\Qt\5.9` which will be always mapped to the current `C:\Qt\5.9.n` to maintain compatibility.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,6 +8,7 @@ configuration:
 - Release
 platform: x64
 clone_depth: 1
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Confirmed it passed Appveyor. I'll merge it by myself as the fix is quite obvious.

PR #1868, opened 2018-03-13
Fix "Replace Level..." Command Fails on the First Call

shun-iwasawa:
Please refer to the previous fix #1626 . 

>"Replace Parent Directory" always fails when you use the command for the first time after launching the software. It fails with an error message `Nothing to replace: no cells or columns selected` even if you surely select cells or columns before using this command.

It was found that the same problem occurs with "Replace Level..." command as well.

The problem was fixed just in the same manner.
I made a class template `OpenReplaceFilePopupHandler` in order to be used by both commands.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1875, opened 2018-03-17
Fix OSX Menu Items

### LINKED ISSUE #1874: Issue with exit group as menu item
In OS X, “Exit Group” is missing from the Edit menu and the quit menu item doesn’t work. I discovered that the problem is that menu items that include either exit or quit as part of the name are automatically placed as the quit menu item. Since exit group comes after the true quit menu item, it replaces the actual quit menu item (and then displays as if it were the quit command) but continues to function as the exit group command. Changing the name from “exit group” to “leave group” returns both menu items to working as intended. Is “leave” an acceptable substitute, or is there a better term?

artisteacher:
This fixes #1874 and part of #618. 

Exit group was automatically replacing quit since exit is part of the name. This made it look like quit was there but disabled and like there was no exit group menu item. In fact, there was no visible quit item and a wrongly named exit group item. Giving exit group norole, returns both quit and exit group to behaving as is intended.

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -1424,6 +1424,9 @@ QAction *MainWindow::createAction(const char *id, const QString &name,
   if (strcmp(id, MI_ShortcutPopup) == 0) {
     action->setMenuRole(QAction::NoRole);
   }
+  if (strcmp(id,MI_ExitGroup) == 0) {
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed the fix. Thank you @artisteacher for solving this! LGTM

Issue #1864, opened 2018-03-12
(Discussion) Slow Performance Issues


adamearle:
OpenToonz slow right down in vector and raster after less than ten drawings. Its so slow in fact I am thinking about switching back to paper because I can draw and animate way faster

How can we bring this viewer into the future?
How can we make it ready by the middle of the year?
What is it going to take?
How much is it going to cost?

Can we have a very real conversation about this instead of its alot of work and dismissing it. 
Perhaps we can make this a github Kanban so we can really start to nut this out and make it happen

Thanks guys 

Ad  :D

artisteacher:
Maybe if we can narrow down what is causing the lagginess, it will help the developers resolve the issue. 

What release are you using?
What dimensions/resolution are your levels?
Do you have the same issue with Toonz Raster levels?
Could you share an example of a typical vector level that slows things down?

Personally, I experience OT slowing down the most when I am working with loaded or imported video. The videos are usually just Quicktime screencaptures that I make into animated gifs for student tutorials. It would be amazing to not have this painful level of lag!

RodneyBaker:
I just went through Preferences and adjusted a lot of things with speed and optimation in mind.  I posted some of the results to that in the OpenToonz discord.  OT can work very fast so it's important to get into specifics and those will often rely on the settings on the local machine.  For other optimizations to work those should be sorted also.

@artisteacher has the right idea with detailing specific information about OS, OT release, resolutions etc.

I'd love to have an actual Scene file/Project that folks claim to have lagging.
Simplicity is the ultimate optimization and it's amazing how quickly simple projects can get out of hand.

adamearle:
You can see what I am looking at on the discord under "keyframing"

https://discord.gg/pCgQeET

adamearle:
also guys I am some answer
MP Raster and Vector are drawn at an A4 resolution 120 Dpi

What release are you using? at the moment I am testing the new OTX from turtle tooth and everything seems to be running swimmingly for now with a project rebuild operating at A5 resolution and about 120 DPI
I have changed the preview setting resample Balance to "Mitchelle" only because of thats what I use when I am rending in Maya

I have also checked "Apply to Shrink to Main Viewer" I am not too sure what this does but I am no am wrong to "assume" that it is reducing memory overhead while animating if the viewer is small on screen,

Do you have the same issue with Toonz Raster levels? I haven't trialed OT raster level and to be honest am not really interested in it.

[...]

PR #1877, opened 2018-03-19
Improve Brush Response

### LINKED ISSUE #1816: Brush Lag
I'm experiencing lag when drawing. I don't know since which version this began to happen as I rarely draw with OpenToonz, it didn't used to happen, I remember it was smooth, but I get these jagged corners when drawing curves and circles since at least version 1.2 and all nightly builds since.

It appears to happen with a mouse, so I don't think it's a tablet issue.

_**Note**: I have Windows flicks disabled and double-click set to off under Wacom preferences. OpenToonz is the only app affected_

**OpenToonz Version:** 1.2 Nightly ([93ee582](https://github.com/opentoonz/opentoonz/commit/93ee58256600d1a7c957d65e92961ae5b2cbeee7))
**OS:** Windows 10 Home (1709)
**RAM:** 16GB
**GPU:** Nvidia GTX 1050 Ti (latest)
**Tablet Device:** Wacom Intuos Pro L
**Tablet Driver:** 6.3.27-2 (latest)

tz_error

shun-iwasawa:
This is the ~~first~~ fix for #1864 , for enhancing Toonz Raster and Raster brushes.
UPDATE: This will also fix #1816 

This PR enables partial update of the viewer.
When using Raster / Toonz Raster brush, the viewer will update only a small region around the brush tip.
It will therefore reduce the computation time to redraw the scene.

- Note that it will NOT take effect with Toonz Vector brush, since the vector strokes are rendered in different way.
- Since I post this PR from my home, there seems to be some difference in behavior of `clang-format`. I actually modified only two lines as follows:
  - Added `setUpdateBehavior(QOpenGLWidget::PartialUpdate);` to ctor of `SceneVIewer`  .
  - Removed `m_clipRect.empty();` from `SceneViewer::GLInvalidateRect()` .

On my environment there seems to be no notable changes with Release build - as it already ran without delay. However, it dramatically enhanced the response of Debug build. Rendering time of the fullscreen viewer with 1920*1080 pixels while drawing with Toonz Raster brush was improved from 167.25 to 18.41 milliseconds per frame.

## UPDATE

Added other measures for enhancement as follows:

- Skip interpolation process if `Smooth` property is 0 . (Toonz Raster Brush and Toonz Vector Brush)
- Refactored source not to enter functions related only to the vector brush . (Toonz Raster Brush) 
- Removed multiple `check_framebuffer_status();` in `SceneViewer::paintGL()` which seem to be for debugging purpose. (All tools)
- Removed `copyFrontBufferToBackBuffer();` in `SceneViewer::paintGL()` . I'm not sure it may be needed in some specific condition / environment, but at least on my environment of Win7 it works OK without this function. And removing this saved considerable amount of computation time. (All tools)

Now the response becomes better and I can draw quite smoothly on my laptop :-)

### DIFF
--- a/toonz/sources/tnztools/brushtool.cpp
+++ b/toonz/sources/tnztools/brushtool.cpp
@@ -752,6 +752,19 @@ void SmoothStroke::generatePoints() {
   if (n == 0) {
     return;
   }
+
+  // if m_smooth = 0, then skip whole smoothing process
+  if (m_smooth == 0) {
+    for (int i = m_outputIndex; i < (int)m_outputPoints.size(); ++i) {
+      if (m_outputPoints[i] != m_rawPoints[i]) {
+        break;
+      }
+      ++m_outputIndex;
+    }
+    m_outputPoints = m_rawPoints;
+    return;
+  }
+
   std::vector<TThickPoint> smoothedPoints;
   // Add more stroke samples before applying the smoothing
   // This is because the raw inputs points are too few to support smooth result,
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
Jenkins

manongjohn:
Cool. Can't wait to see if this will help me without having to force my graphics card to work with OT, and if it doesn't, maybe the hint from my other PR might work now.

shun-iwasawa:
Included other measures. Please see the first comment for details.
I'm not sure removing `copyFrontBufferToBackBuffer();` will cause some bad side effect in another platform. I'll test on OSX later.

Jenkins again

norrinra:
I m too excited for this can I test this on macosx ? but I should build it I guess ? or wait for the pkg ?

adamearle:
i know that i have said this somewhere else. but because it works only in .tiff format i wonder if having the option to work in png will reduce the foot print. 

.PNG is half the file size of .tiff

I would expect the .png would help speed things up. Anything that makes OT faster and more stable would help OT be used more.

We all want to enjoy our work and not wrestle with it. Curious to know what you think about having a 
 .png option and how it maybe implemented.

Thanks again 

Ad

turtletooth:
@shun-iwasawa Have you had a chance to check this on Mac yet?  I will review this tonight or tomorrow.

shun-iwasawa:
@turtletooth 
Not yet, I'll check it right away. Thanks!

shun-iwasawa:
I checked briefly on OSX and it works OK. :+1:

shun-iwasawa:
@adamearle 
Saving raster level as PNG seems to be possible but we need to make OT to save the image dpi in it.

Note that PNG format does not store dpi, but [has the ability to store "pixels per meter" in unsigned integer value](https://www.w3.org/TR/2003/REC-PNG-20031110/#11pHYs) . Losing decimal part may cause failure in reproducing the level property on loading.

turtletooth:
This feels nice and snappy!  Thanks for the update.  I tried it on my main machine and my surface with good results.

norrinra:
That's a huge difference for me. Thank you so much ! I've been waiting for this for too long : ) 
There's is still a very very slight heaviness exists in the toonz raster comparing to raster. This very very slight lag doesn't exist on toonz premium. I don't know if it is worth to investigate though. Again thank you so much for looking into this.

beeheemooth:
Thanks guys, for all your work! You are a real superheroes !

norrinra:
Don't mind my previous message just downloaded the last version. It's just amazing. Working perfect. Not even slightest lag. Thank you so so much.

adamearle:
I haven't had the chance to test guys sorry and it won't be for another month it looks like it. love your work though.

PR #1904, opened 2018-04-04
Fix slow Toonz Raster brush on High DPI

turtletooth:
I don't know if anyone else has this issue, but when I use Toonz Raster levels on my Surface Pro, the brush stalls at the beginning of a stroke.  It only happens on Toonz Raster levels, but anything that I drew for the first second is smoothed even with 0 smoothing.

I'm not sure if this is the best way to solve it, but I just bypassed the smoothing code completely if the smooth factor was set to 0.  This seems to make a significant improvement.

I suspect that the high dpi of the screen is messing with the delay before the smoothing will start outputting a stroke, but I couldn't find where that might be happening.

This change at least allows me to use an unsmoothed brush with good performance.

### DIFF
--- a/toonz/sources/tnztools/brushtool.cpp
+++ b/toonz/sources/tnztools/brushtool.cpp
@@ -1264,11 +1264,14 @@ void BrushTool::leftButtonDown(const TPointD &pos, const TMouseEvent &e) {
         m_tileSaver->save(m_rasterTrack->getLastRect());
         m_rasterTrack->generateLastPieceOfStroke(m_pencil.getValue());
 
-        m_smoothStroke.beginStroke(m_smooth.getValue());
-        m_smoothStroke.addPoint(thickPoint);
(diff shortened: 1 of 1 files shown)


adamearle:
I can test the surface we have here can you send the link to the OT your using?

Ad

shun-iwasawa:
Jenkins

shun-iwasawa:
Unfortunately I couldn't reproduce the brush stall as you mentioned on my high-dpi devices (MobileStudioPro and MacBook) yet.
However I found a probable cause and fixed a bug in #1938 which will prevent calling the time-consuming function `GLInvalidateAll()` .

This PR will also result in preventing calling `GLInvalidateAll()` . Moreover there will be a benefit by skipping unnecessary call of `SmoothStroke` functions when smooth = 0.

LGTM thank you @turtletooth for the fix !

PR #1905, opened 2018-04-04
Make Touch Zoom More Responsive

turtletooth:
I've been using the touch controls quite a bit lately and it seems the zoom response is a little small compared to other programs.  I made the zoom response higher.

I also moved the zoom code above the rotation code.  I often and trying to zoom and it will trigger rotating instead.  This might catch at least a few of those cases.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -884,6 +884,31 @@ void SceneViewer::gestureEvent(QGestureEvent *e) {
       m_scaleFactor   = 0.0;
       m_rotationDelta = 0.0;
     } else {
+      if (changeFlags & QPinchGesture::ScaleFactorChanged) {
+        double scaleFactor = gesture->scaleFactor();
+        // the scale factor makes for too sensitive scaling
+        // divide the change in half
+        if (scaleFactor > 1) {
+          double decimalValue = scaleFactor - 1;
+          decimalValue /= 1.5;
+          scaleFactor = 1 + decimalValue;
+        } else if (scaleFactor < 1) {
+          double decimalValue = 1 - scaleFactor;
+          decimalValue /= 1.5;
+          scaleFactor = 1 - decimalValue;
+        }
+        if (!m_rotating && !m_zooming) {
+          double delta = scaleFactor - 1;
+          m_scaleFactor += delta;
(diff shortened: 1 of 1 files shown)


artisteacher:
Another issue with zooming (probably OS X specific), is that with trackpads, a two finger scrolling gesture is read as middle mouse wheel input. This triggers zooming in/out. By itself that's not a big deal but the standard (and expected) trackpad scrolling/panning behavior in OS X uses two fingers. Currently, changing panning to the standard two finger gesture results in a jerky, simultaneous zooming _and_ panning behavior. With touch controls enabled, mouse wheel scrolling should really be disabled - at least for OS X. 

On a related note, disabling the panning gesture altogether returns the tools to being functional in OS X (when touch is enabled).

bobbybutterfly:
two finger zooming is still a bit offset

artisteacher:
I’ve figured out how to get the tools to function while touch is enabled for devices that use trackpads and how to set two finger panning for trackpads only while leaving the 1 finger gesture for other touch devices. Some mouse functions were disabled specifically for tablets and all touch devices. That prevents touchpads from using the tools. It’s possible to target touchscreen devices and touchpad devices individually with QTouchDevice::Touchpad and QTouchDevice::TouchScreen. Since touchpads can be targeted individually, its also possible to disable the scrolling zoom behavior when touch is enabled for touchpads. 

I could make this fix limited to OS X, but other OS could potentially have the same issue with a trackpad. Would it be better to target all touchpads or just Mac touchpads?

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed that it works fine. Thank you @turtletooth for tuning!
LGTM 

By the way..
It's really a matter of preference, but for me I'd like to have one-finger panning more responsible.
( For now you need to scroll more than 100 pixels to trigger the panning mode. My lazy finger would like to make the threshold more shorter, say, 10 pixels. )

shun-iwasawa:
>It's really a matter of preference, but for me I'd like to have one-finger panning more responsible.
> ( For now you need to scroll more than 100 pixels to trigger the panning mode. My lazy finger would like to make the threshold more shorter, say, 10 pixels. )

**UPDATE**
I found that such slowness in reaction does exist only in OSX with TouchPad. On my MobieStudioPro with touch screen works fine in response for one-finger panning.

PR #1906, opened 2018-04-05
Improvements to Vector Snap

turtletooth:
This makes two changes to Vector snapping
You can now snap to the starting point of a brush stroke.
Holding alt while drawing will invert whether snapping is on or off.  This will help if you only need snapping occasionally or if you want to quickly disable it.

### DIFF
--- a/toonz/sources/tnztools/brushtool.cpp
+++ b/toonz/sources/tnztools/brushtool.cpp
@@ -1439,8 +1439,8 @@ void BrushTool::leftButtonDrag(const TPointD &pos, const TMouseEvent &e) {
     m_mousePos      = pos;
     m_lastSnapPoint = pos;
     m_foundLastSnap = false;
-    checkStrokeSnapping(false);
-    checkGuideSnapping(false);
+    checkStrokeSnapping(false, e.isAltPressed());
+    checkGuideSnapping(false, e.isAltPressed());
     m_brushPos = m_lastSnapPoint;
 
     if (e.isShiftPressed()) {
@@ -2023,8 +2023,8 @@ void BrushTool::mouseMove(const TPointD &pos, const TMouseEvent &e) {
       m_firstSnapPoint = pos;
       m_foundFirstSnap = false;
 
-      checkStrokeSnapping(true);
-      checkGuideSnapping(true);
+      checkStrokeSnapping(true, e.isAltPressed());
+      checkGuideSnapping(true, e.isAltPressed());
       m_brushPos = m_firstSnapPoint;
     }
   }
@@ -2044,11 +2044,13 @@ void BrushTool::mouseMove(const TPointD &pos, const TMouseEvent &e) {
 
 //-------------------------------------------------------------------------------------------------------------
 
-void BrushTool::checkStrokeSnapping(bool beforeMousePress) {
+void BrushTool::checkStrokeSnapping(bool beforeMousePress, bool invertCheck) {
   if (Preferences::instance()->getVectorSnappingTarget() == 1) return;
 
   TVectorImageP vi(getImage(false));
-  if (vi && m_snap.getValue()) {
+  bool checkSnap             = m_snap.getValue();
+  if (invertCheck) checkSnap = !checkSnap;
+  if (vi && checkSnap) {
     m_dragDraw          = true;
     double minDistance2 = m_minDistance2;
     if (beforeMousePress)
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Tested this out via AppVeyor and the snap to start point is working great!
Thank you!.

gab3d:
Would this feature allow to automatically close hand drawn strokes?
That would be useful!

RodneyBaker:
> Do you mean, the initial and ending points becoming one while editing the curve, as when you use the Tape Tool with the Join option activated?

Yes Sir.   This can depend on your stroke angle but adjusting the settings and getting a feel for how best to end the stroke can often produce near perfect results.  I'll say 'near perfect' because there are a lot of variables.  For instance, it might not look like the ends meet exactly if the brush stroke has any variation in thickness as the small end and large end will connect but look awkward.  But after playing with it for a little while you'll see for yourself.  If set to 'maximum' the lines will always connect but it help to release (i.e. stop drawing) at the perfect time as the ends will then find themselves.

RodneyBaker:
I do wonder if the "High" setting could be set to be more severely high.
It almost seems as if the setting is protecting me from myself.  ;)

In other words, if drawing slowly the accuracy is much higher (of course) but when trying to draw really fast, which is often required when creating arcs and circles the ends connect but not smoothly.

I expect some of this is just like drawing a circle or ellipse freehand on paper where practice makes perfect.

gab3d:
@RodneyBaker 
oh..!! that's really great news! :)
can't wait to test it...

turtletooth:
@gab3d It actually didn't connect the first and last points into one, until you mentioned it.  I just added that.  Good idea.

RodneyBaker:
Nice addition.
I just tested it and I like that it can join as that makes editing it later so much easier.

I do find that I have to slow down a little in drawing to make sure the highlight indicator fires before I release the stroke and sometimes it can be difficult to visually see that appear.  Is there are way to further emphasize it by making it brighter or thicker?  I think green is the right color but can be a little hard to see.  I do find that if I release the stroke (i.e. stop drawing) after I've taken the time to actually see the indicator the ends connect. If I just draw with the expectation that the ends will meet the ends will meet perfectly about 50% of the time.  Note here that the gaps closing is well above that percentage 75% of the time or better.  For testing purposes I'm quickly drawing vertical ellipses such as some artists might do to warm up before diving in to drawing so their lines are smooth and accurate.  This last part may be key in that I have the Accuracy and Smoothness settings set to around 75% each.  I'll test those settings as well and see if I can find a sweet spot.   

If drawing slowly none of this is much of an issue as the pen can be guided into place and when the indicator appear the stroke can be released.  I do find that at my normal speed of drawing I don't care to use any other setting than 'high' with the Snap feature.

This is a great addition to OT and.... I love that we can test things about via AppVeyor releases!

RodneyBaker:
@gab3d If you want to test this I'd be happy to walk you through the process of downloading the AppVeyor release.

RodneyBaker:
A nice part of joining the ends (as opposed to just placing) is that if the lines don't smoothly connect that can easily be remedied with the Ironing Tool.

RodneyBaker:
There is an interesting 'problem' with this vector snap in that it (by design) mostly concerns itself with snapping to line ends.  As such it loses that 'snap' influence inbetween the ends (in the middles).

A example of this would be to have a palette with three colors (the alpha white.. the black... and a red (or other color of preference other than black(
That third color is then used to draw some lines that will be used as guides for the vector snapping.

So ... draw something with the 'red'   grids... lines in perspective... squiggly thingies....
Then change to the black and use the red lines as a template/guide to draw other things.
While doing so... note how the influence of the snapping changes.
Try using the low setting.
Try it on Medium
Try it on High

In most cases I believe you will find that only the High setting allows for reasonable use
I noticed one exception to this where when drawing at a specific speed (fairly slow as I recall) I was able to get good snapping out of the lower settings.
But I found that to be rare.

So I find myself wondering if the extremes of low and high in these settings is extreme enough to allow for optimal use of low, medium and high settings.

Submitted for your consideration with grateful thanks and appreciation for this feature.

turtletooth:
Rodney- are you talking about tracing the red lines?  If so, that isn't really what the snap was intended to do.  It's only meant to serve as a guide for starting and stopping lines.

RodneyBaker:
Yes, Sorry Jeremy I was trying to suggest my exploration was well beyond your intended use.  I'm mostly just pointing out what the feature as currently implemented can now also do.

turtletooth:
Thanks Rodney.  I understand now.

@shun-iwasawa I made the change you requested.  Thanks for pointing that out.

openanim:
@turtletooth @RodneyBaker 

Asmall question about being able to join two strokes to make one when snapping. How to do it ? I tested, but I still get two separate strokes. Is there an option somewhere?

shun-iwasawa:
Jenkins again

RodneyBaker:
@openanim  I don't believe this feature allows for joining as currently implemented.  The only tool I know that will do that is the Tape tool.

PR #1907, opened 2018-04-06
Option for Automatically Generate Palette when Raster Image is Loaded as Color Model

shun-iwasawa:
### Background
- Color models are created for accurate color-referencing during production when painting characters and props. 
- If the color model is a Toonz raster or vector level (TLV or PLI file), it already comes with a palette, whose style names and page configuration will be preserved. If the color model is a full-color image or animation level, the palette is automatically generated by extracting colors from the image. (see also [the user manual](http://opentoonz.readthedocs.io/en/latest/painting_animation_levels.html#using-a-color-model))
- In Japanese animation production workflow, most of ink&paint work is done by using RETAS PaintMan and color models are created by using Adobe PhotoShop. During the migration of the softwares, there would be a case that ink&paint work must be done with OT by using the raster color model (such as PNG) made with PhotoShop.
- Before this PR, loading raster images as color model does automatically generate a palette, but the order of styles becomes too irregular so that users must manually organize style order after loading, which was very troublesome.

### Commonly Used Color Model in Japanese Animation Production
- Color model images used in Japanese animation production are like this : 

 This image is contained in [OpenToonz sample data](https://opentoonz.github.io/e/download/sample.html) . Note that they are generally;

- Have no anti-aliasing.
- Styles are put in rectangular chips surrounded by solid line with specific color.

### Description of This PR
Based on the above, this PR will add a new option when loading a raster image to the color model.
After specifying the file, the new dialog will open like this:

image

A group box titled `Picking Colors from Raster Image` is newly added in this PR.

- **Pick Type** : specifies a method to generate palette from the image. Note that the first two options ( 
`Pick Every Colors as Different Styles` and `Integrate Similar Colors as One Style` ) are conventional, but moved from the preferences to here. The third option `Pick Colors in Color Chip Grid` is newly added.
-  **Grid Line Color** : specifies a color used for border line of color chips.
- **Grid Line Width** : specifies line width of the color chips border.
- **Chip Order** : specifies order of picking color chips into the palette. 

For loading the above example, setting the Grid Line Color = `(R:255 G:0 B:4)` and Grid Line Width = `4` will work. The generated palette will be automatically organized like as follows:
image


shun-iwasawa:
Jenkins

gab3d:
@shun-iwasawa 
This feature seems to be very useful.
Thanks!

shun-iwasawa:

shun-iwasawa:

shun-iwasawa:

shun-iwasawa:

shun-iwasawa:

masafumi-inoue:
LGTM

RodneyBaker:
This is looking good to me in the wild (on Win10) and it is a great addition.

Informational
There are several workflows that may trip people up when they attempt to use this feature and that primarily stems from where OpenToonz must have a palette created first before a Color Model can be loaded.  This is evident in several places to include via File > Load Color Model  where an error will be generated stating, "Cannot load Color Model in current Palette" and via the color model window which will not allow Right Clicking in order to load or create a color model/image.  This can make it appear the Color Model feature as well as this enhancement (Auto Generate Palette) are not working.

One of the easiest ways to see this feature in action is to drag/drop the image desired as the color model into OpenToonz and then in the Color Model window Right Click and select "Use Current Frame".  HOWEVER, that method will bypass this new feature so the user may again miss out on this feature and also wonder why their palette don't match the color image in the color model window.

shun-iwasawa:
@RodneyBaker 
For now color model feature is not available for Raster levels.
The color model is supposed to be used for painting animation levels. And as you pointed, it is associated with palette. Therefore it can only be applied to levels with palette, i.e. Toonz Raster or Toonz Vector level types.

RodneyBaker:
Thanks for the clarification.
I had been leaning that way but ran off course even though you clearly stated the Toonz Raster/Vector level requirement in your first post.  Hopefully I'm back on track now.

PR #1909, opened 2018-04-08
Fixed redundant condition check in in initToonzEnv

cclark25:
Original code checks the condition (stuffDir == TFilePath()) twice, so I rewrote it so that it is only checked once, for efficiency. Also I translated the comment for this section of code to English to help future coders.


shun-iwasawa:
Jenkins

shun-iwasawa:
Thank you @cclark25 for refactoring ! LGTM

PR #1910, opened 2018-04-11
Fix Travis-OSX : Try Using the Latest Qt

shun-iwasawa:
This is for fixing the recent Travis-OSX errors.
Since [Homebrew was updated on April 9](https://brew.sh/2018/04/09/homebrew-1.6.0/) , old formula for obtaining Qt5.9.2 seems to become invalid with error like as follows:

```
Error: Calling 'depends_on :mysql' is disabled!
Use 'depends_on "mysql"' instead.
```
In order to resolve it, this PR will try using the latest formula which will obtain Qt 5.10.1 .


shun-iwasawa:
Confirmed Travis check has passed. I'll merge this.

PR #1914, opened 2018-04-13
Fix touch controls for touchpad and stylus issue

artisteacher:
This changes the touch controls so that they are fully usable on computers with touchpads #1579. This should not impact touch screen devices and will not change any behaviors when touch is not active.
 
- 2 finger panning is enabled so that 1 finger can navigate and use tools (1 finger panning continues for touch screen devices)
- Enables disabled mouse behaviors when touch is active so that tools can be used again
- Mouse wheel scrolling behavior is disabled (when touch controls are active) to prevent jerky, simultaneous panning/zooming during 2 finger scrolling

The mouse wheel scrolling change is necessary but currently does add one unexpected (but noncritical) behavior: 
- after switching from touch to no touch, you need to click in the viewer before zooming in/out the first time using the 2 finger scrolling behavior

This also adds a fix for some undesired behaviors when autosave interrupts stylus button input, including #618

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -100,6 +100,9 @@ void initToonzEvent(TMouseEvent &toonzEvent, QTabletEvent *event,
   toonzEvent.m_buttons  = event->buttons();
   toonzEvent.m_button   = event->button();
   toonzEvent.m_isTablet = true;
+  // this delays autosave during stylus button press until after the next 
+  // brush stroke - this minimizes problems from interruptions to tablet input
+  TApp::instance()->getCurrentTool()->setToolBusy(true);
 }
 
 //-----------------------------------------------------------------------------
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -404,7 +404,8 @@ void SceneViewer::onEnter() {
 void SceneViewer::mouseMoveEvent(QMouseEvent *event) {
   // if this is called just after tabletEvent, skip the execution
   if (m_tabletEvent) return;
-  if (m_gestureActive) {
+  // and touchscreens but not touchpads...
+  if (m_gestureActive && !QTouchDevice::TouchPad) {
     return;
   }
   // there are three cases to come here :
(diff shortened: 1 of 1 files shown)


turtletooth:
I can't check mac PRs, but I want to say that I love that you have jumped in and are helping make the experience on mac better.  Thanks a ton!

artisteacher:
Thanks @turtletooth! I'm happy to help, though I'm not a programmer by any means.  I understand CSS, not C++ so I'm relying on trial and error/lots of googling.

Technically this one is touchpad specific not Mac specific - though that is practically the same thing. Beyond confirming that this works on Macs/touchpad devices as expected, someone will need to make sure that this doesn't impact touchscreen devices. I'm pretty sure it won't, but I don't have a touch screen device to test with.

toonz/sources/toonz/sceneviewerevents.cpp
```diff
@@ -944,11 +954,15 @@ void SceneViewer::touchEvent(QTouchEvent *e, int type) {
     m_firstPanPoint = e->touchPoints().at(0).pos();
     m_undoPoint     = m_firstPanPoint;
   }
-  if (e->touchPoints().count() == 1 && m_touchActive) {
+  // touchpads must have 2 finger panning for tools and navigation to be functional
+  // on other devices, 1 finger panning is preferred
+  if ((e->touchPoints().count() == 2 && m_touchActive && QTouchDevice::TouchPad)
+      || (e->touchPoints().count() == 1 && m_touchActive && QTouchDevice::TouchScreen)) {
```
shun-iwasawa: You cannot use `QTouchDevice::TouchScreen` or `QTouchDevice::TouchPad` directly in the if statement for identifying the device type.
Please see the [Qt reference](http://doc.qt.io/qt-5/qtouchdevice.html#DeviceType-enum) - they are enum with values 0 (for TouchScreen) and 1 (for TouchPad) . 
To identify which device is used, you need to call the function `QTouchDevice::type()` ([reference](http://doc.qt.io/qt-5/qtouchdevice.html#type)) and compare the return value with these enums.

`QTouchDevice::type()` can be called only in the function `SceneViewer::touchEvent(QTouchEvent *e, int type)` like as follows:

```
if (e->device()->type() == QTouchDevice::TouchScreen) {
  // touch screen case
}
else {    // e->device()->type() == QTouchDevice::TouchPad
 // touch pad case
}
```

If you need to know the current device type outside of the `SceneViewer::touchEvent()` , the standard technique would be to keep the `QTouchDevice::DeviceType` value as the member variable of `SceneViewer` class.

I'll try to modify it and post another PR based on yours. Thanks!

PR #1920, opened 2018-04-18
Modifications on Studio Palette tree view

shun-iwasawa:
This PR will modify dragging behavior in the tree view of Studio Palette window as follows:

- Made a confirmation dialog to be opened before drag&move palette inside the tree view in order to prevent unintended moving files by mistake. Since the studio palette is usually shared among studio staffs, I think it is reasonable to take users attention on modifying it.
- Drag an item and drop on another palette item will now cause moving the selected item to the same folder as such palette. Before this PR, dropping on another palette caused nothing (but the operation added an useless undo).
- Item order in the tree view is changed for simplicity. Now all sub-folder items are put above palette items. Before this PR, items were all mixed up and put in alphabetical order.

This modification was requested by some Japanese animation studio.

### DIFF
--- a/toonz/sources/toonzlib/studiopalette.cpp
+++ b/toonz/sources/toonzlib/studiopalette.cpp
@@ -256,9 +256,10 @@ TPalette *StudioPalette::getPalette(const TFilePath &path,
 void StudioPalette::movePalette(const TFilePath &dstPath,
                                 const TFilePath &srcPath) {
   try {
-    TSystem::renameFile(dstPath, srcPath);
+    // do not allow overwrite palette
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Fixed that the studio palette could not be renamed.
Also fixed a bug that undo of moving palette fails to refresh the tree content properly.

Jenkins again

shun-iwasawa:
Made palettes not to be overwritten by renaming or dragging.
Jenkins again

masafumi-inoue:
LGTM

Issue #1876, opened 2018-03-19
Crash with incorrectly named text file (extension .GIF)  FFMPEG


RodneyBaker:
I think we almost have this one squashed but I ran into this odd one.

Attached is a text file with the word 'empty' in it.
Place this on your desktop or in a folder and OpenToonz will ignore it.
Rename it to filename.gif (or any filename with .gif extension) and it will cause OT to crash.

The primary way I tested was via xsheet Right Click > Load Level
While browsing move to the directory where the bad file resides.
Entering a directory where the file resides may not cause immediate crash but
Scroll down until
OpenToonz 'sees' the file in the listing.
Crash should occur at this point without every selecting a file.

You should not need the attached text file as any text file with .gif extension should suffice.
The file that did this on my system was apparently html file accidentally given the .gif extension.

Not sure if this is strictly FFMPEG related but it stands to reason that earlier fixes might have got at one part of the problem but there are a few lingering problems.

Confrimed with latest OTX (17 March) and 14 March official releases.

[crasherNot.txt](https://github.com/opentoonz/opentoonz/files/1823525/crasherNot.txt)

RodneyBaker:
Tested also with the majority of other image formats and most passed the test.  One other 

.MP4 extension also fails
.webm fails (I expected it to as I've not had much success using .webm format)

The following pass (no crash) If I don't explain further we might have an undocumented feature here (need to investigate!):

.BMP good
.PNG good
.TGA good
.spritesheet good (I expected this to fail but perhaps it passes because OT is not expecting that format to be one of a sequence of images?)  I think that is why it gets a pass.
.MP3  good (I expected it to crash.. perhaps again related to sequences)
.WAV good (same as .MP3)

artisteacher:
It looks like @shun-iwasawa's symlink fix also resolves this crash for gif, mp4, and webm files. I am getting a crash if I rename the text file to a jpeg.

konero:
Curious, why are you renaming a txt file gif?

RodneyBaker:
Renaming was the easiest way to reproduce and demonstrate the problem.

artisteacher:
I've found that students will occasionally rename files to a different type because they think that renaming will convert the file type.

RodneyBaker:
Closing this as the problem no longer occurs in 1.2.1 and more current releases.
If anyone confirms a related issue I'll gladly reopen.

Thanks!

PR #1930, opened 2018-04-25
Fix symlink crashes (alternative to #1842)

### LINKED ISSUE #1227: Crash when accessing alias folders in Browser

shun-iwasawa:
This fixes #1227 and #984 , with keeping the symbolic link available.
This is alternative to #1842.

Please note that OT can handle symbolic links (which can be created with `ln -s` command), but seems to have no capability to resolve macOS aliases (created by `File > Make Alias` command, or by pressing `Command-L`) due to the Qt bug ([QTBUG-23927](https://bugreports.qt.io/browse/QTBUG-23927)).
Actually I wasn't aware of the difference of them before investigating this.

### DIFF
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -411,7 +411,7 @@ void Ffmpeg::getFramesFromMovie(int frame) {
 
 QString Ffmpeg::cleanPathSymbols() {
   return m_path.getQString().remove(QRegExp(
-      QString::fromUtf8("[-`~!@#$%^&*()_E=|:;<>«»,.?/{}\'\"\\[\\]\\\\]")));
+      QString::fromUtf8("[-`~!@#$%^&*()_Â—+=|:;<>Â«Â»,.?/{}\'\"\\[\\]\\\\]")));
 }
 
 int Ffmpeg::getGifFrameCount() {
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -189,6 +189,9 @@ QString Ffmpeg::runFfprobe(QStringList args) {
   results += ffmpeg.readAllStandardOutput();
   int exitCode = ffmpeg.exitCode();
   ffmpeg.close();
+  // If the url cannot be opened or recognized as a multimedia file, ffprobe
+  // returns a positive exit code.
+  if (exitCode > 0) throw TImageException(m_path, "error reading info.");
   std::string strResults = results.toStdString();
   return results;
 }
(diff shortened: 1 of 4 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
@artisteacher 
The crash seemed to be due to recursive call of the function `FileBrowser::setFolder()`, happened when the symbolic link in the browser was clicked.
Can you please check if this PR would resolve the issue?

artisteacher:
@shun-iwasawa this works perfectly to fix the symlink folder crash #1227 but does not resolve the symlink file crash #984.

shun-iwasawa:
@artisteacher Thank you for testing. I'll look into it.

shun-iwasawa:
I've tested #984 and here are the results ;

### Test1 - Loading GIF alias
1. Drag & drop GIF file from the Stuff folder to the desktop. (As you pointed, the file was not moved but its alias was created)
1. Select the desktop folder and load the file (= the alias file with the same name as original) into OT.

**Result**: The file failed to be loaded properly. OT did not crash.

### Test2 - Loading TIF alias
1. Drag & drop TIF file from the Stuff folder to the desktop. (Same as Test1, alias was created)
1. Select the desktop folder and load the file into OT.

**Result**: Same as Test1. The file failed to be loaded properly. OT did not crash. 

### Test3 - Rendering files into the same folder as the alias
1. Keeping the GIF & TIF aliases in the desktop, render sequential TIF files into the desktop.  

**Result**: The files are rendered with no problem. OT did not crash.

So, to recap,  I couldn't reproduce the crash mentioned in #984 . The alias which is created via drag&drop the file from the stuff folder to desktop seems to be the same as the ones which can be created by `File > Make Alias` command. However, since it has the same extension as the original, OT misunderstood that they are supported file format.
( Actually, until checking this I didn't know that you need both `ffmpeg` and `ffprobe` to be kept in the same folder in order to use ffmpeg file I/O. )

@artisteacher Can you please let me know what I'm overlooking in order to reproduce the crash?

shun-iwasawa:
Ah! I could reproduce the crash. I missed to have ffprobe on Test1 yesterday.
I believe the crash occurs when the ffmpeg tried to open a file with unexpected content.
I'll try to fix this, probably adding some validation check.

shun-iwasawa:
Added an exception handling. Now OT just stops loading without crash for alias files.

Jenkins again

artisteacher:
This fixes the gif problem - and also most of #1876! However I can still get a crash with a jpg alias and a .txt file renamed as a jpg.
[jpg_crash.txt](https://github.com/opentoonz/opentoonz/files/1958148/jpg_crash.txt)

shun-iwasawa:
Jenkins again

masafumi-inoue:
LGTM

PR #1931, opened 2018-04-26
Allow for multilingual FX Help tooltips by jpturcotte (modified)

shun-iwasawa:
This PR is based on #1861 by @jpturcotte .

It seems that `TIStream::operator bool()` is overloaded for checking whether the file is successfully opened.
However, it doesn't work on Windows because `explicit basic_ifstream::basic_ifstream(_Filet *_File)` does not set `ios_base::failbit` in constructor even if `_File` is null.
This PR will fix such problem by setting `failbit` manually in the constructor of `Tifstream` in case it fails to open the file.

I also modified `Tofstream` in the same manner.

Regarding the multilingual Fx help files, I confirmed it works fine. (Thank you @jpturcotte for that!)
I think the default help files (to be put in `OpenToonz 1.2 stuff/doc/`) can be replaced by English ones as soon as they are ready, in order to coordinate with the default UI language.

### DIFF
--- a/toonz/sources/common/tsystem/tfilepath_io.cpp
+++ b/toonz/sources/common/tsystem/tfilepath_io.cpp
@@ -31,8 +31,12 @@ FILE *fopen(const TFilePath &fp, string mode) {
   return pFile;
 }
 
-Tifstream::Tifstream(const TFilePath &fp)
-    : ifstream(m_file = fopen(fp, "rb")) {}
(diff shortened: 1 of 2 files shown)


jpturcotte:
@shun-iwasawa Thank you for taking the time to review this! Once this is merged, I will make another PR to include the English docs, and make them the default option. Is this alright with you?

masafumi-inoue:
LGTM

PR #1937, opened 2018-04-28
Stylesheet Fixes

konero:
This will mainly fix a couple of issues with the stylesheets caused by previous pulls.

### Fixes:
- Tabs that were missing their bottom accent line
- Some icons seem to have been reprocessed at some point causing some to be moved from their pixel grid or resized, they should be sharp again now
- Removes icons that became deprecated due to #1863 
- Adds hover effect to checkbox text for all stylesheets
- Change window close button icon color so it's easier to see on Light
- Shrunk width of console tool buttons
- Style added for function curve editor for Light, similar to Blender

func_edit_comp

### Compiling
Some people asked how to edit the themes, so here is an instructions MD file, though I don't know what to do with it at this point. You can also view it on my forks [Wiki](https://github.com/konero/opentoonz/wiki/Stylesheets).

📁 [how_to_stylesheet.zip](https://github.com/opentoonz/opentoonz/files/1972859/how_to_stylesheet.zip)

### Try the QSS
📁 [qss.zip](https://github.com/opentoonz/opentoonz/files/1972855/qss.zip)

### Note
I added a new theme color initially but I removed it (again). Let's leave this PR as a bugfix and for now I'll share the lighter theme manually.


konero:
@artisteacher Aw, that's terrible! I don't believe it's an issue with stylesheets though, they don't really have a method of changing vertical alignment. The Mac interface fonts were Lucida Grande, Helvetica Neue and San Francisco, you could give those a try. Helvetica is different to Helvetica Neue.

One thing you could try, install **Source Sans Pro** (open-source font by Adobe) IMO the best interface font and set it as the font for OpenToonz, you can get it here:

**https://fonts.google.com/specimen/Source+Sans+Pro?selection.family=Open+Sans**

Something else you can try, download the stylesheet pack, which will force a text size or a specific font:

📁 **[ArtistTeacher.zip](https://github.com/opentoonz/opentoonz/files/1958285/ArtistTeacher.zip)**

But don't use these stylesheets for production, they will break anything that relies on text size and will likely make using Combo Boxes a pain, just for testing purposes. Since I don't own a Mac I won't be able to help further, sorry, but do post back any findings, or create an issue if not already. 👍

artisteacher:
Thanks! I have both Helvetica Neue and Lucida Grande on my computer by default. They both completely resolve the clipping problem. The graph editor ruler could still probably use a bit more padding at the top but the numbers are completely visible. The problem is that with a clean install, Helvetica **_not_** Helvetica Neue is the default font for the UI of OT on a Mac. It looks like that is set in main.cpp and should be a quick fix. Windows is set to Segoe UI & so won’t be impacted by a change. Should Helvetica Neue be used just for Macs or for all non-Windows OS?

gab3d:
@konero 
thank you for the function curve editor change for the Light theme! I was suffering it big time. :)

konero:
@artisteacher Nice!

I only had a quick look, I think in `preferences.cpp` at line 230 is where the default font is set, it's the feature developed by @turtletooth.

https://github.com/opentoonz/opentoonz/blob/63e8da9d531b0dc5e331858f93392d66fc5f5521/toonz/sources/toonzlib/preferences.cpp#L230

> Should Helvetica Neue be used just for Macs or for all non-Windows OS?

Only Mac ships with it by default (lucky) so maybe needs an ifdef in the place I linked above for MACOSX.

manongjohn:
@konero I recommend putting the stylesheet compiling instructions with the other instructions for building OT so those that need to update them for new features or even those who may want to create new themes know how to do it.

Maybe add some suggested instruction for creating a shareable theme that is not part of the installation?

konero:
@artisteacher I wouldn't worry about those, the preference variable seems to override everything else, if you're brave enough you can refactor and remove them to be honest.

@manongjohn I guess so. I'm not good at explaining stuff though, I'm a bit of a rambler. Maybe a Git Wiki page instead?

The theme is mostly variable where it matters. Anything that requires a color is variable, and those variables are defined with a separate theme file. In LESS, anything that comes last in the cascade is boss. Almost all variables use the variable `@bg` as their base color, we use color functions to change their appearance (darken, lighten, saturate, spin, etc).

For example, if `@manongjohn: red;` and `@artisteacher: lighten(@manongjohn, 50);` then `@artisteacher` is 50% lighter than what color `@manongjohn` chooses to be.

A visual example, since the color of a buttons/textfields background and border is linked to `@bg`, look what happens if we change `@bg` to a random color, and change `@hl-bg-color` to a different color to show what happens to highlighted elements.

tz_bl

Everything has a blue tint, because everything is attached to `@bg`. The button background color is going to be 13% lighter and textfield will be 10% darker.

To create a new theme file, you must choose a theme to derive from. Let's take a look at the Blue theme.

`@import '../Default';`

By importing Default, we've duplicated all the variables from the Default theme, it's now a clone.

`@bg: rgb(65, 67, 69);`

By setting the `@bg` variable, we've done an override since this comes after the import. It's weird I know.

`@text-color: rgb(214, 216, 221);`

We override the text color.

`@title-text-color-active: #43AEE5;`

We override the text color in the title bar when the window is active, making the text blue. And that's it for the Blue theme. If you change something in the Default theme, then it will obviously change the look of the Blue theme, its a child and parent scenario, like C++ classes.

Light is more complicated but it still derives from Default, because a lot of things needed to be inverted and the XSheet required more specific attention.

Is this the kind of thing you meant?

manongjohn:
@konero Yes, that is effectively what I meant.  

For new themes that would not be part of the official distribution, I would suggest that they create the .less file in their theme folder and inherit from the standard default.less location.

So for creating non-distributed themes, instructions might start off like so...
1) Create new theme folder:  C:\OpenToonz 1.2\stuff\config\qss\myTheme
2) Create or copy an existing theme as "myTheme.less" file in the new folder
3) Modify the file as follows:

**// out:**** ./myTheme.qss
// Include Default Theme
**@import** '../Default/less/Default';

Then when done. The last step would be to zip up the myTheme folder and share. :)

konero:
@manongjohn I wrote a wiki on my fork.

https://github.com/konero/opentoonz/wiki/Stylesheets

And the MD file is here:

📂 **[how_to_stylesheet.zip](https://github.com/opentoonz/opentoonz/files/1962589/how_to_stylesheet.zip)**

I don't really know what to do with it, you can copy paste it anywhere appropriate.

konero:
@shun-iwasawa Thanks for reviewing, I've corrected the problem. 😹

shun-iwasawa:
I will skip Jen.kins check since this PR does not change the source itself.
Again, thank you @konero . LGTM 

As @manongjohn suggested, how about adding your style sheet instructions to the main repository? It's informative and will be helpful for users who would like to modify the style by themselves.
You can just put the MD file into `doc` folder and add appropriate link from `README.md` then post a new PR. Please let me know if it'll be troublesome then I can do on behalf of you. Thanks!

konero:
@shun-iwasawa Thanks, I've created a new PR as requested.

PR #1938, opened 2018-04-28
Fix tabletEvent not to call onMove when the cursor position does not move

shun-iwasawa:
This will fix the following problem:
When using tablet, `TTool::leftButtonDrag()` is called continuously even if the cursor does not move.

There was a condition expression to prevent calling `onMove` if the cursor does not move from the previous position, but it didn't work properly since it compared `QPoint` (integer) and `QPointF` (decimal fraction).

I found this bug while reviewing #1904 and I think this PR will also relieve the slowness mentioned in #1904 .


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1942, opened 2018-05-01
Fix Shortcut Behavior on Tool Option Combobox 

shun-iwasawa:

For the property of which the first item in the combobox is "Normal" , pressing the shortcut key for another item (for instance, "Rectangular") could change back and forth between "Normal" and such item.
(That is to say, pressing "Rectangular" shortcut key when the property is already set to "Rectangular", the option will change to "Normal".)
However, after introducing #1863 this feature does not work on non-English UI.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1947, opened 2018-05-07
Timeline zoom control update

### LINKED ISSUE #1689: Zoom in/out slider do not hidden, When timeline is minimize
zoom_out_timeline_slider

manongjohn:
This PR fixes #1689

Additionally made the following changes:

1) Internally, moved the zoom slider controls into a panel rather than a direct draw to the xsheet viewer.  The panel can be used to add additional buttons later if desired. 

2) Implemented a suggestion by a user: click on zoom in/out icons and it will adjust the zoom level in that direction by a certain amount.  This functionality is naturally there when clicking on the black bar of the slider itself.  The slider will move towards the pointer with every click.  I essentially extended the logic to the in/out icons.

3) Added "Ctrl-click" of zoom in or out icon will zoom all the way in or out respectively.

4) Zoom in/out icons light up like a button when hovering over them.

### DIFF
--- a/toonz/sources/toonz/xsheetviewer.cpp
+++ b/toonz/sources/toonz/xsheetviewer.cpp
@@ -48,11 +48,12 @@ TEnv::IntVar FrameDisplayStyleInXsheetRowArea(
 namespace XsheetGUI {
 //-----------------------------------------------------------------------------
 
-const int ColumnWidth       = 74;
-const int RowHeight         = 20;
-const int SCROLLBAR_WIDTH   = 16;
-const int TOOLBAR_HEIGHT    = 30;
-const int ZOOM_SLIDER_WIDTH = 80;
+const int ColumnWidth     = 74;
+const int RowHeight       = 20;
+const int SCROLLBAR_WIDTH = 16;
+const int TOOLBAR_HEIGHT  = 30;
+const int ZOOM_FACTOR_MAX = 100;
+const int ZOOM_FACTOR_MIN = 20;
 }  // namespace XsheetGUI
 
 //=============================================================================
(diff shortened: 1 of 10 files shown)


toonz/sources/toonz/xsheetviewer.cpp
```diff
@@ -412,18 +404,11 @@ void XsheetViewer::positionSections() {
...
-        QRect(0, 0, XsheetGUI::ZOOM_SLIDER_WIDTH,
-              XsheetGUI::SCROLLBAR_WIDTH - 2)
-            .translated(m_columnScrollArea->rect().right() -
-                            XsheetGUI::ZOOM_SLIDER_WIDTH - 17,
-                        height() - XsheetGUI::SCROLLBAR_WIDTH + 1);
-    m_frameZoomSlider->setGeometry(sliderRect);
-  }
+
+  m_layerFooterPanel->setGeometry(0, m_columnScrollArea->height() + 50,
+                                  m_columnScrollArea->width(), 14);
```
shun-iwasawa: For now the slider seems to offset upward to the incorrect position when the xsheet toolbar is visible.

How about changing this line like as follows?:
```
  m_layerFooterPanel->setGeometry(0, m_columnScrollArea->geometry().bottom()+1,
                                  m_columnScrollArea->width(), 14);
```

shun-iwasawa:
Jenkins

gab3d:
@manongjohn 
Could the default Timeline zoom be set to halfway in the zoom range?

I tend to think one of the biggest benefits of the Timeline is being able to see more frames at a glance than with the Xsheet, but with the default zoom factor, in fact, we see fewer frames than with the Xsheet, which kind of goes against that potential benefit.

PR #1948, opened 2018-05-07
Fix Wheel Event to Be Disabled while Touch Gesture 

shun-iwasawa:
This will fix the problem mentioned by @artisteacher in the comment of #1905 as follows:

>Another issue with zooming (probably OS X specific), is that with trackpads, a two finger scrolling gesture is read as middle mouse wheel input. This triggers zooming in/out. By itself that's not a big deal but the standard (and expected) trackpad scrolling/panning behavior in OS X uses two fingers. Currently, changing panning to the standard two finger gesture results in a jerky, simultaneous zooming and panning behavior. With touch controls enabled, mouse wheel scrolling should really be disabled - at least for OS X.

I made the `wheelEvent()` not to zoom while touch gesture.


shun-iwasawa:
Jenkins

shun-iwasawa:
I'm so sorry for not notifying that! I'll review yours.

PR #1952, opened 2018-05-08
Fix MotionBlurIwaFx Fails on the First Frame

shun-iwasawa:

MotionBlurIwaFx fails to be rendered properly at the first frame in the rendering frame range. Especially when you use batch rendering with separating frame range into chunks, every first frames of the chunks shows incorrect result (i.e. the blur is not applied).


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1955, opened 2018-05-09
Add TNZ files to the Extension Whitelist of Checking Seqence

shun-iwasawa:
This quick patch will fix the following problem:

For now OT cannot load the scene with the name like `scene_44.tnz` .


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

Issue #1957, opened 2018-05-09
Column Copy - Paste - Undo bug


manongjohn:
Undo of pasted column can delete column next to it.

To reproduce:

1) Create a new scene with 5 columns and label them 'a' - 'e'
2) Select and copy column 'a' with Ctrl-c
3) Select columns 'c' and 'd' and then paste with Ctrl-v
      - The copied column 'a' is created next to column 'c'
4) Undo paste with Ctrl-z
      - The newly pasted column 'a' and existing column 'c' are removed.

This appears to be an old bug as I can duplicate in 1.1.3.

PR #1958, opened 2018-05-10
Timeline column creation fix and changes

manongjohn:
This PR accomplishes the following:

1) Inserting empty column on brand new scene doesn't work the 1st time

When starting a brand new scene, the timeline shows only 1 empty column. Inserting another empty column appears to do nothing. Any attempt after that to insert more empty columns will be okay.

On startup, even though there is a column showing, technically the first column isn't really there internally. Attempting to insert an empty column technically creates the first one making it appear as if nothing happened. 

I modified the command to create 2 new empty columns, 1 for the one that is displayed but not really there yet and 1 for the new one the user was trying to add.

2) Removed empty column creation when dragging columns/cells past first and last columns

Some users were reporting how easily and annoyingly it was to create a large amount of empty columns when dragging existing columns/cells.  I decided to remove this feature and replace it with...

3) Added Insert & Paste Above/After commands

By default, the _Insert_ (Ins) and _Paste_ (Ctrl-V) commands will create the new columns below (timeline) or left (xsheet) of a selected column.  This required remembering to select the next column over if you wanted to create a new column above/right of a target column.

I added an _Insert Above/After_ (Shift+Ins) and a _Paste Insert Above/After_ (Ctrl+Shift+V) command to create columns in the opposite direction of the target column.

Please note that I also changed the text of "Insert Paste" to "Paste Insert".


artisteacher:
That will be so helpful!

toonz/sources/toonz/xshcolumnviewer.cpp
```diff
@@ -2278,7 +2280,32 @@ void ColumnArea::contextMenuEvent(QContextMenuEvent *event) {
...
+    act3->setText(tr("&Paste Insert Below"));
+    act4->setText(tr("&Paste Insert Above"));
+  }
+
   menu.exec(event->globalPos());
+
+  act->setText(tr(actText.toLatin1().data()));
+  act2->setText(tr(act2Text.toLatin1().data()));
+  act3->setText(tr(act3Text.toLatin1().data()));
+  act4->setText(tr(act4Text.toLatin1().data()));
```
shun-iwasawa: I found that on Japanese language UI the command name in the menu bar becomes `??` after opening the column context menu.
IMO you don't need to call `tr()` here because the action texts are already translated on definition.
Rewriting the line to `act->setText(actText); ` seems to work.

shun-iwasawa:
I tried the `Paste Insert Above/After` commands with multiple conditions:

The conditions are combination of the following:
- How many columns is copied in the clipboard (single column / multiple columns).
- How many columns are selected when using the command (single column / multiple columns).

I found some problems as follows:

### Single column copied, single column selected 

Undo of the command fails to revert the scene.

### Multiple columns copied, single column selected

The result seems to be incorrect.
Undo of the command fails to revert the scene.

### Single/multiple columns copied, multiple columns selected

Actually I'm not sure what is the desired result under these conditions... At least, undo of the command fails to revert the scene.
And I found that the original `Paste Insert` command fails to undo as well. This must be posted as another issue.

manongjohn:
@shun-iwasawa Odd, I tried every single one of these scenarios many timesm strange combinations and they all behaved similar to how a normal Insert behaved, just in the opposite direction.  Undo generally worked for me, though I did discover an issue with the current insert-undo which I noted in #1957 

I may need you to give me specific steps to reproduce the issues you are seeing.

shun-iwasawa:
>  I did discover an issue with the current insert-undo which I noted in #1957

Oh, so sorry I didn't know that you already posted the issue for that. Thanks!

For the single column copied - single column selected condition, here is a gif for reproducing:
pasterinsertabove

The original `Paste Insert` command seems to work with this steps.

toonz/sources/toonz/xshcolumnviewer.cpp
manongjohn: Corrected per suggestion

manongjohn:
The paste undo issue you found is fixed. When pasting, I was adjusting the target columns to paste correctly, but didn't account for the adjustment for the undo.  Revised and simplified how I performed the paste operation which fixed the undo issue at the same time.

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed that now it works fine other than the case mentioned in #1957 .
Thank you @manongjohn ! LGTM

PR #1959, opened 2018-05-10
Change Default Folder Location to $scenefolder According to User Preference

shun-iwasawa:

The following changes will be applied if `Path Alias Priority` in the preferences is set to `Scene Folder Alias ($scenefolder)` :
- Initial folder of all file browser popups will be set to `$scenefolder` .
- When drawing on the empty cell with the `Autocreation` preference option is enabled, a new level will be created and saved in `$scenefolder` .
- Default output destination folder for new scenes will become `$scenefolder`, regardless of the project settings.

The above changes are made according to a basic policy that users who store their assets in the scene-folder based way should not take care of project folders.


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins again

shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1960, opened 2018-05-10
Cursor preferences

manongjohn:
This PR allows users to set some cursor preferences on drawing tools.  This currently only affects cursors on the canvas and not places like xsheet/timeline or toolbars, etc.

The preferences found under the Tools tab in the Preferences window and include the following:

- [x] Use Simple Cursors

- Some people find some of the icons annoying.  This converts them to a +

- [x] Use Left-Handed Cursors

- For those lefties out there, this flips most cursors horizontally.

- [x] Show Cursor Size Outlines

- This PR fixes #1857 providing the option to show or hide the circle surrounding the cursor which normally indicates the area of effect or size of the stroke.  This can be toggled easily via the context menu when right-clicking the canvas.  A keyboard shortcut can be assigned to this action as well.

PR #1962, opened 2018-05-11
Change eraser tool cursor icon

manongjohn:
This PR addresses a request to remove or replace the eraser cursor icon of the eraser tool.

The "eraser" icon is a little intrusive when you are trying to erase something that happens to be where the icon actually is.  I also find it a bit distracting.

I have the replaced the Toonz Vector and Toonz Raster cursor icons from an "eraser" to the "plus" sign, that the regular Raster eraser uses.  Please note, the eraser type text and the line/area text that is part of the Toonz Raster cursor icon has also been removed.

### DIFF
--- a/toonz/sources/tnztools/rastererasertool.cpp
+++ b/toonz/sources/tnztools/rastererasertool.cpp
@@ -709,21 +709,22 @@ void EraserTool::draw() {
 
 int EraserTool::getCursorId() const {
   int ret = ToolCursor::PenCursor;
+  /*
+    if (m_eraseType.getValue() == FREEHANDERASE)
+      ret = ret | ToolCursor::Ex_FreeHand;
+    else if (m_eraseType.getValue() == POLYLINEERASE)
+      ret = ret | ToolCursor::Ex_PolyLine;
+    else if (m_eraseType.getValue() == RECTERASE)
+      ret = ret | ToolCursor::Ex_Rectangle;
 
-  if (m_eraseType.getValue() == FREEHANDERASE)
-    ret = ret | ToolCursor::Ex_FreeHand;
-  else if (m_eraseType.getValue() == POLYLINEERASE)
-    ret = ret | ToolCursor::Ex_PolyLine;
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
> The "eraser" icon is a little intrusive when you are trying to erase something that happens to be where the icon actually is. I also find it a bit distracting.

Can you make this change optional ? Since I'm not sure if all users think so.
IMO cursor icon can be helpful for users to know what type/option of tool they are using without moving eyes from where you are working to the tool options bar.

manongjohn:
I can make it optional, though keep in mind the Raster eraser is not an "eraser" icon like the other ones and I personally like that better.  I feel it should be consistent across the board.

I toyed with the idea of just temporarily hiding it while actively erasing just in NORMAL mode (it's ok to keep it in other modes).  Perhaps that is a better compromise rather than creating yet another preference setting?

shun-iwasawa:
The inconsistency in cursor icon between the raster eraser and other ones seems to existed in Toonz Harlequin and I hadn't modified it since Ghibli staffs had little chance to use the raster eraser.

> Perhaps that is a better compromise rather than creating yet another preference setting?

I agree with you !

shun-iwasawa:
Well but now I'm bit afraid that the quick show/hide of icons according to the mouse action may become bothersome..

konero:
That would indeed be bothersome. In ClipStudio you can disable cursor icons if you use a Cintiq or choose a specific cursor icon for each tool type including its orientation for certain cursors, this can be done within its own preferences menu area, which would be OK.

clipcursor

I'm actually left-handed, so in most programs when I use a tablet, cursors are usually backwards for me. In Clip, I use a top-right facing triangle. 👍 

In DigiCel FlipBook, it's just two options. Simple Cursors means a cross, Left-handed means flip the cursor horizontally.

digicursor

manongjohn:
Preferences it is! 😊

I will cancel this PR and incorporate a cursor preference feature into PR #1960, probably something similar to the DigiCel Flipbook example.

Thanks @shun-iwasawa and @konero for the feedback.

PR #1971, opened 2018-05-15
Fix Slot Function Name

shun-iwasawa:
This small fix will modify slot function name to work the disconnection properly when deactivating the fill tool.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #1973, opened 2018-05-15
Fix Placement of Freeze Button in Preview Window

konero:
The freeze button is overlapping the minimize button in the preview window (**CTRL+R**). I assume it happens in the flipbook window too but I haven't seen the freeze button appear there yet to test, since the source file for fixing this is for the flipbook window.

This small fix will adjust its position.

**Here is the issue:**
preview_window_button_issue

**What this PR does:**
preview_window_button_issue_fix

### DIFF
--- a/toonz/sources/toonz/flipbook.cpp
+++ b/toonz/sources/toonz/flipbook.cpp
@@ -274,7 +274,7 @@ void FlipBook::addFreezeButtonToTitleBar() {
     m_freezeButton->setToolTip("Freeze");
     titleBar->add(QPoint(-64, 0), m_freezeButton);
     connect(m_freezeButton, SIGNAL(toggled(bool)), this, SLOT(freeze(bool)));
-    QPoint p(titleBar->width() - 62, 0);
+    QPoint p(titleBar->width() - 64, 0);
     m_freezeButton->move(p);
     m_freezeButton->show();
   }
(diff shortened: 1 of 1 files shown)


toonz/sources/toonz/flipbook.cpp
```diff
@@ -272,9 +272,9 @@ void FlipBook::addFreezeButtonToTitleBar() {
         titleBar, ":Resources/pane_freeze_off.svg",
         ":Resources/pane_freeze_over.svg", ":Resources/pane_freeze_on.svg");
     m_freezeButton->setToolTip("Freeze");
-    titleBar->add(QPoint(-55, 0), m_freezeButton);
+    titleBar->add(QPoint(-64, 0), m_freezeButton);
     connect(m_freezeButton, SIGNAL(toggled(bool)), this, SLOT(freeze(bool)));
-    QPoint p(titleBar->width() - 62, 1);
+    QPoint p(titleBar->width() - 62, 0);
```
shun-iwasawa: Thank you for finding and tackling with the offset bug.
Can you modify this line as follows?
`QPoint p(titleBar->width() - 64, 0);`

For now it seems that the freeze button slightly offset (2pixel) to the left when resizing the window.
It is because of the mismatch of the horizontal position set at here and line 275. Thanks!

konero: Now it makes sense, thanks @shun-iwasawa! Changes made.

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @konero for the fix!

Issue #1975, opened 2018-05-15
Edit In Place doesn't show main if the sub-xsheet's frame is not exposed in the main


manongjohn:
I'm submitting this issue on behalf of @Orphanlast who encountered this problem while working on one of his presentations.

Basically when working in a sub-xsheet with Edit In Place toggled ON, if you go to a frame that the sub-xsheet is not exposed on in the main, you stop seeing what's in the main.

Example to recreate:
1) Create 3 empty columns
2) On Col1-frame 1, draw something and extend to frame 10
3) On Col2 and Col3 - frame 1, draw something and extend to frame 5
4) Collapse Col2 and Col3 to create a sub-xsheet
5) Open the sub-xsheet and toggle Edit In Place to ON
6) Click on frame 5 - You can still see the drawing from Col1
7) Click on frame 6 - The drawing from Col1 disappears

This also happens if there is a gap in the exposed sub-xsheet frames in the main:
1) Close sub-xsheet to return to main xsheet
2) Delete the exposed sub-xsheet's frame 2
3) Open the sub-xsheet again
4) Click on frame 1 - You can still see the drawing from Col1
5) Click on frame 2 - The drawing from Col1 disappears
6) Click on frame 3 - You can see the drawing from Col1 again

We would expect if Edit In Place is on, we should still be able to see the main from inside the sub-xsheet, regardless if the sub-xsheet's corresponding frame is exposed in the main or not.

gab3d:
I think the frame from main displayed is the one the current subxsheet cell appears in. So this makes me wonder a couple of things:
- What frame from main to display when you are in an unexposed cell?
- What happens when the current subxsheet frame is exposed in multiple frames of main?

beeheemooth:
In my opinion this is the obvious behavior of sub-xsheet. If the window in the room is closed, it will not be possible to look either outside or into the room. The disappearance of visibility in places where sub-xheet frames are deleted can be useful for internal sub-xsheet animation, since it indicates the places where sub-xsheet does not have frames on the main (Xsheet)

manongjohn:
@gab3d Hmm, good point.  That could explain why the main disappears because it has no frame of reference from which to build an image.

I guess then it matters how you are using the sub-xsheet. If you assume the frames from the main and the sub-xsheet are in sync, then Edit In Place as described above would make sense.  However, that is not a requirement for using sub-xsheets since it can start anywhere in the main and the frames can be used like any normal level so it can be held, reordered, skipped.

[...]

PR #1980, opened 2018-05-16
Open sub-xsheet at selected sub-xsheet frame

manongjohn:
This PR enhances the behavior of the "Open Sub-xsheet" as discussed in #1975.

The following logic is used when opening a sub-xsheet"
- Opening via Column selection
  - The sub-xsheet will open at the sub-xsheet frame where the Current Frame Indicator is located.  If it is an empty cell, the sub-xsheet will open at the 1st exposed sub-xsheet frame.

- Opening via drop-down menu
  - If there is no active selection, the sub-xsheet will open at the sub-xsheet frame where the Current Frame Indicator is located.  If on an empty cell, it will remind you that you need to select a cell and not do anything.
  - If there is an active selection, the sub-xsheet will open at the 1st exposed sub-xsheet frame in the selection.

- Opening via right-click context menu
  - If there is no active selection, the sub-xsheet will open at the sub-xsheet frame exposed in the cell you right-clicked regardless if it is the Current Frame Indicator or not.  If the cell is empty, the option is not available.
  - If there is an active selection, the sub-xsheet will open at the 1st exposed sub-xsheet frame in the selection.

### DIFF
--- a/toonz/sources/toonz/subscenecommand.cpp
+++ b/toonz/sources/toonz/subscenecommand.cpp
@@ -976,29 +976,34 @@ void openSubXsheet() {
   TColumnSelection *columnSelection =
       dynamic_cast<TColumnSelection *>(TSelection::getCurrent());
 
-  bool ret          = false;
-  ToonzScene *scene = app->getCurrentScene()->getScene();
-  int row           = app->getCurrentFrame()->getFrame();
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
It works very nice. Thank you @manongjohn for the improvement!
LGTM

PR #1986, opened 2018-05-19
Update Styles for V121

konero:
This will update the stylesheets again, there is one aesthetic change but this was mainly done to fix one aesthetic bug that will be introduced by merging #1622 in the style editor intended for release of `1.2.1`.

**The Bug Fix**
- Because #1622  will remove the channel buttons it will leave a border line under the tab area since a top-border property is being applied to each `#PlainColorPageParts`, the change I made was to swap the border direction from top to bottom, although it doesn't solve it entirely, it's a much more reasonable location for the bug.

**Small Aesthetic Changes**
- Will add a hover effect to the text string of Radio Buttons.
- Changes the tab style in the FX property window to square since round looks glitched.
- Shrink the size of tool buttons in the Palette Toolbar, it's now one icon widths shorter.
- Fix `#DoublePairField` groove color to match `QSlider` in the Light theme.
- The render frame indicator color is now easier to see in the Light theme when you click Preview.

**Big Aesthetic Change**
- The Tree Branch icons have been changed to arrows. The reason I did this is because applying the current style is really difficult globally since every QTreeView is a different size, many areas display them broken like the Scene Cast, many windows don't display them at all and instead use the in-built style by Qt which can't be seen in the Light theme, such as the FX browser. This will significantly reduce the overhead of maintaining it too, reducing it from 28 images down to 4. If a folder contains no sub-folders then it will display no arrow.

**Note**
- The Tree Branch change needs to be tested on Mac specifically please.
**Download:** [qss.zip](https://github.com/opentoonz/opentoonz/files/2031634/qss.zip)


artisteacher:
I definitely prefer the arrows - the current tree branch icons look pretty outdated.  I’m not seeing the alternating line colors in the file tree like in the image above but otherwise they look & function perfectly in MacOS. 

A couple of small details, the color of the ruler is identical to the viewer background only in the light theme. The bottom bar of the timeline (zoom/horizontal scrolling) blends in completely with the empty timeline color in the light theme. In the other themes only the scrollbars have that problem. More definition would be beneficial.

konero:
@artisteacher Thanks for testing! If you go to `File > Load Folder`, that's the window I displayed. Not all windows use alternating rows.

I fixed the other issues you mentioned (ruler, timeline scroll area).

- [qss-fix.zip](https://github.com/opentoonz/opentoonz/files/2020288/qss-fix.zip)

shun-iwasawa:
Looks nice. Thank you @konero for the fix & update!
LGTM

PR #1659, opened 2017-12-07
Fix Tablet Behavior on OSX

shun-iwasawa:
(hopefully) fixed #756 

For now when using tablet on OSX, viewer seems to behave in a different way from Windows.
In particular, most of events such as `mousePressEvent`, `mouseMoveEvent`, `mouseReleaseEvent`, `contextMenuEvent`, `enterEvent` or `leaveEvent` seem NOT to be called properly. 
So, I tried to modify it by handling all of operations done in the above events in `tabletEvent` instead.
Also I fixed not to draw stroke if the tablet moves with 0 pressure.

Please note that the above changes are done only for OSX. I tried to keep the behavior in Windows and other OS unchanged.

I already notice that there are some problems remaining as follows:
- Instead of preventing the unwanted stroke, there is a case that a stroke cannot be drawn. Please just draw again in such case.
- Mouse cursor seems to fail to update properly when using tablet on OSX.

Anyway, I think I made it better than before.

After merging this, I would like to wait a few days before releasing the next version for verification since this will be quite a big change.

PR #1990, opened 2018-05-21
Timeline UI Update

### LINKED ISSUE #1688: marker interval is not sync when timeline is zoom out
Thank you first of all @manongjohn 

Marker Interval is not sync with bold number, When timeline is zoom out.
win7, OT 1.2

marker_interval

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/52759540-marker-interval-is-not-sync-when-timeline-is-zoom-out?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1985: Adding Clarity to Timeline View
Big fan of the timeline, but personally it's very difficult to use effectively due to the sheer amount of noise generated from the abundance of information held by each level. On and off for a couple months I've been looking at a re-design, but now I definitely won't have time to do it, so I'm posting my findings and assets here in-case somebody else looks into it.

Here's what I came up with:

timeline_design

Put this image into PaintDotNet or any editor with a pixel grid to determine the sizes for each area.

- Invert white text to black.
- The 'shift trace' and 'onion frames' indicators get a visual helper with a separator line.
- The interval blue marker is shortened to the time area.
- Remove the hold line, a hold will be obvious by the break line between two frames.
- Reduce vertical cel separator lines to small markers on the cel color blocks.
- New key icon to represent what is probably the industry standard for such.

Usually I would build in Python and then port to C++, I'm just not the biggest Qt fan so I didn't get very far, so no real code to share. Here's how I constructed the cel block style.

- **Each cel is 10x20 pixels at the smallest zoom factor.**
Here's how cels sit and how they're separated.

xsh_cel

Each cel has a 1 pixel black border on the bottom and on the right which are aligned with the separator dividing lines on the timeline grid.

Each cel also has 1 pixel white (15% alpha) on the top and on the left side which is what helps to distinguish each cel area from another with the outdent effect.

xsh_cel2

And that's pretty much it visual wise. The only icon is the loop key icon, the key icon itself can be drawn easily in Qt as a square and rotated, etc.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58575775-adding-clarity-to-timeline-view?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #2003: New Basics horizontal Room? (Called "Timeline"?)
With all the new improvements made lately to the Timeline, I think we should provide a good starting point for users wishing to use it.

Following is a variant of the Basics Room, tuned for use with the Timeline, instead of the Xsheet.

basicsroom horiz

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58785640-new-basics-horizontal-room-called-timeline?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

manongjohn:
This PR resolves #1985 

The timeline, especially when zoomed all the way out is very "noisy" with all the lines and symbols.  This PR attempts to clear some of that noise.  Changes are based on suggestions from @konero and further discussed in the PR.

The updated timeline display looks like this:

image

The following changes were made to the timeline display:
- The continuation line has been removed from all levels.
- The vertical frame separators have been removed between held frames.  A vertical separator only exists whenever there is a frame change.
- A scene's Marker Interval will only show when there is a vertical separator between 2 different frames.
- The frame marker dot no longer takes on text color. It is always black. It still turns red if source file cannot be found.
- The keyframe icon switches to a white diamond (blue when selected) when zoomed in less <= 50%.
- Horizontal lines were added in the frame header to show where the fixed and relative onion skin marks could be enabled.
- Vertical lines between frames in the frame header have been shortened to 3-6 pixels tall from the bottom.
- The small frame indicators below the numbers in the frame header have been removed.
- When activated, the space between the Start and End frame header markers have a different color than the rest of the header to show the scene boundaries.  Currently the color used is the same color for non-empty columns. [NOTE: Change was made to xsheet mode also]

Also included in this PR:
- Fixed bug with setting Marker Interval = 0 causing a crash.
- Fixed bug on loading a scene that made xsheet cell grid 1 px lower than it should be until it refreshed
- Changed the Xsheet/Timeline orientation button to have an icon instead of words;  Tooltips added to items in panel.
- Rearranged note panel icons
- Restored Preview Visibility button for Palette levels
- Windows menu item "Timeline" which opens directly to Timeline view
- PR resolves #2003 - A dedicated Timeline Room setup.  This has been requested by multiple people and is also being added in anticipation of the new Curve Animation feature currently in development.
image
- Allow turning off Marker Intervals when value = 0
- PR resolve #1688 - Matches zoomed out frame numbers to marker intervals in "Frame" mode only. Defaults to every 5 if marker intervals turned off. Other modes do not affected


bobbybutterfly:
@manongjohn great work!
i have a dream: selecting a keyframe in the timeline also selects the corresponding key/ curve in the function curve editor and viceversa

openanim:
I'm in love with the OT Xsheet but damn, this timeline is clean. Congratulations and thank you for your work.

I would like if it is not difficult to do, make the two lines of the onion skin optional. Once you know it, it's easy enough to find your way. Although I understand his integration. For the very purpose of this PR, this will be good imo. But, it's minor thing.

By the way, I tested this PR but although the drawing is the same, I have several vertical lines as before when I extend it.. But perhaps it is normal at this stage of development?

manongjohn:
@openanim Thanks. Can you post a screen shot of what you are seeing with the vertical lines?

openanim:
@manongjohn 
timeline
is it normal ?

manongjohn:
@openanim  No. Did you build it or use the appveyor artifact?

openanim:
@manongjohn 
I used appveyor artifact. the zip file archive. I unzip it and I launched OT nothing else.

manongjohn:
Strange. I just downloaded the artifact myself and it was fine for me.

Need someone else to test the artifact and see what results they get.

manongjohn:
@openanim looking closer at your screenshot, you are missing other stuff like interval markers, the highlighted area in the frames.

Oddly, if I zoom into frame 8, i can see the frame separators for frame 2 next to those other lines. Not sure why your display is so strange.

gab3d:
@manongjohn 
Very nice work! Congrats and thanks! :)

Two things com to my mind, regarding the timeline:

1. In my opinion the dots could be even smaller, so they:
  a. make the timeline less cluttered even.
  b. don't compete visually with the keyframes as much as they do now.

2. Could the Audio track height be made taller or user adjustable? ...or at least to toggle between two sizes via a [+] or [>] icon somewhere?
It's useful to be able to see the waveform peaks and valleys with more detail, when animating for sound.
(Also, the display of just the upper half of the waveform helps to save screen space).

RodneyBaker:
I just downloaded the AppVeyor artifact and am not getting the vertical lines.
Maybe it's a Mac issue?  Win10 here.

manongjohn:
@openanim Just realized your playback markers are inactive which is why that area is not lighter. I'm still baffled why you seem to be missing the blue marker intervals that display normally every 6 frames and instead have all these extra lines. Did you play with marker interval settings or something?

@gab3d Thanks. I'll play with the dot size and see how it looks. As for the sound column, unfortunately there isn't much I can easily do with that right now. The columns have to remain uniform and would need some coding to allow that to expand.

@RodneyBaker Thanks for testing. I can't imagine what could be causing the strange results on @openanim 's machine since the appveyor artifact was used and can only be used on windows machines.

manongjohn:
Hmm. Well that explains the markers. 😊 I was going to fix the logic that setting marker to 0 would turn it off. Not sure why 0 should default to 6, but its coded that way.

Need to figure out this extra line issue. I have a small wacom cintiq and it looked fine there for me.

openanim:
Here is a new screenshot, I think everything is good except the vertical lines?

timeline2

It's strange that it doesn't work on any of my tablets. But maybe this will be solved when the PR will be merged.

openanim:
With default intervals it's better to see timeline3

manongjohn:
Could it be some strange resolution issue?

manongjohn:
Do they disappear as you zoom out?

openanim:
You're right, it's a problem of resolution. By changing for 1920 x 1080 or 1920 x 1200 it works. But above it does not work anymore. 👍

konero:
Better, I can't run a build for awhile since my machine is now pinned down rendering for work. 😂 

When I built it in Py, I stuck to this schematic (I should have posted it in the issue) if it helps at all.

`CelBase(10x20) + CelMarker(1x3)*10 + CelBorder(1) = Cel`
`CelStartPoint(4x4) black <-- the black box at the start frame`

pyschem

Logic wise I didn't bother adding the drag handle as a visual component, here's what I did.

pyschem2

The drag handle is invisible and covers the entire cel block (except key frame data). The user has to hold the mouse button down for `> 1000ms` for it to function. The cel block will be highlighted and you can then drag it. 

However, if movement is detected at any point `<= 1000ms` then ignore it, to allow normal cel selection.

I'd say this is a common way to do it, the software I use seems to do it this way at least. A modifer key could also work here, when held down click dragging on the cel would result in no delay.

pyschem3

I know it displays information like lock status, but in fairness I look at the layer box area for that or expect the software to argue with me when I try to move something, "Hey! That's locked punk, get off."

Having it displayed everywhere and on everything seems like bad economics. If we had to make it clearer to users, then we can make the dot icon in the layer box an actual padlock.

manongjohn:
Ok. That explains what I am seeing. Must be the high res monitors. Zooming in on your images, normally the lines are 1 px wide when displayed....well on my screens anyway. Yours appear to be 2 px wide normally? However between held frames it only looks like 1 for you. Compare the blue marker at the end of your level with one in the middle of the level or even with the marker at the top in the number line.

The code normally shifts the cell left border +1px to create the separator. This PR doesn't add the 1px separator between held frames which could explain why in middle of your level they seem to be half a normal line width.

Problem is, how do I fix it for high res monitors?

konero:
> Problem is, how do I fix it for high res monitors?

IMO would it be easier making interval a bool and revert to old code? Then users like me who don't use it can turn it off, say, it uses `qproperty = verticalLineColor` instead.

manongjohn:
@openanim I made a change that I think/hope will fix the resolution issue you are encountering.  When you have time, please download the latest appveyor artifact and test it out again.

openanim:
I confirm everything works fine, thank you ! :)
timeline4

konero:
@manongjohn 

> I won't be putting the bezel effect on the timeline. Our UI is pretty flat looking and the bezel effect would make it stand out too much and possibly distracting. 

I'm not sure a flat look will add enough clarity, but I wait to be convinced.

For reference, here is the styles from different software. I'm hard pressed to find a flat type.

tz_time_concept

**Green** (Lightworks, Harmony, Animate CC, Avid ME, Premiere CC, After Effects CC)
**Purple** (Photoshop)
**Blue** (DaVinci Resolve).

manongjohn:
@konero The yellow one is pretty much what we have now.  That top part serves as our drag point.

openanim:
@manongjohn 

Personally, I also like the drag button. One thing I think should be possible also about this is to allow it to be used backwards to erase frames. We can do it a few seconds after adding frames, but past this delay, we can't anymore by dragging. I think that it would be an easy way to control the timing to allow to go back at any time. For exemple Tvpaint also has small wrists on each frame to allow this. What do you think in the future for timeline and xsheet mode ?

beeheemooth:
Guys. Please, make the active level color brighter, or make this color customizable in the Preferences. I really don't see it, I don't even see what is written. I have an old monitor. When I try to find the active level, I instinctively think: "where is this red hole between the layers?"

RodneyBaker:
That yellow for the selected column does improve on the red which, as others have reported, makes it seems as though there is an error at that location.

openanim:
@manongjohn 
Here is an example to illustrate this. Both ways are similar TVpaint and OT. Except that Opentoonz don't allow to go back to erase frames created after few seconds.
opentoonz drag frames
 below how Tvpaint do that.
tvpaint drag

manongjohn:
@openanim  When you are in a cell and you press and hold CTRL key, it will bring the cell drag-handle in the opposite direction for left dragging purposes.  It won't replace existing cells when you drag though but, you can fill in empty cells before it that way.

openanim:
Thank you @manongjohn  . I didn't know this nice trick to fill or create empty cells even if in my case it is already existing images to change the timing so it does not work. Just a reflection for the future. 👍

konero:
Be sure to fetch latest QSS.
👍 👍

openanim:
@manongjohn 

Sorry man, I think I didn't fully understand your explanation. When I tested the hold ctrl to use a second handle I didn't know that drag back required previously empty cells, to work. As I did on existing frames, nothing was happening. But now I understand better your first message, it seems to me about the selection.

I never realized until now that handle worked to add or delete frames by creating a selection beforehand. I'm sorry for the misinterpretation. Because now my request is already there thank you :) .

I still think that similar behavior to add frames would be logical also to remove without creating a selection beforehand.

snapai:
It came up on the discord chats that users are having a lot of trouble figuring out that the "xsheet"/"timeline" button is a button, and what it does - so we were discussing adding an icon. Here's an initial draft of some 20x20 SVG icons for the timeline/xsheet button. A single "two-way" icon, and a togglable icon for Xsheet and Timeline view.

I'm including a preview since I'm not sure how many of the features in the file will actually render when loaded as icons (specifically the dark background color setting in the file may need to be removed, and there is a clipping mask involved in each icon.)



[timeline_xsheet_icons.zip](https://github.com/opentoonz/opentoonz/files/2060679/timeline_xsheet_icons.zip)

gab3d:
I like the idea!

Especially as I'm foreseeing that the equivalent for "Timeline" in Spanish ("Línea de Tiempo") is too long for that button, so this could be a perfect solution. :)

I tend to like the two icons solution more, as it anticipates what it's gonna do when clicked.

snapai:
Here's a modified version- without clipping masks, lines aligned to the pixel grid, and with a black stroke to match the other icons. The original 20px are set in the preview in the header bar icons, I also scaled it to 32px (and expanded the top row of the timeline interface in a graphics editor, since it was crowded anyway) and placed it in place of the current button.

The partial-transparency is faked here with levels of gray.

image
[timeline_xsheet_icons.zip](https://github.com/opentoonz/opentoonz/files/2061817/timeline_xsheet_icons.zip)

konero:
Why not remove the button altogether and make the timeline its own window?

artisteacher:
I would hate to lose the ability to toggle back and forth.

konero:
What purpose does toggling serve out of curiosity? In all the years I've used Harmony I can't think of a reason I'd need to toggle back and forth, they're used in different phases of production.

Isn't it a pain when you have a timeline docked at bottom, switch to x-sheet view etc? You won't see the big picture unless it's docked at one side.

Just seems redundant to me, but if it's useful to anyone else then it's no issue.

artisteacher:
I think the current functionality is just more flexible.

Currently, the major advantage of the Xsheet over the Timeline (even when docked at the bottom), is that there is a visual way to identify the different columns. I like to have the timeline tall enough so that I can see the full column header of the Xsheet. That way, when I need to quickly identify a column, I can toggle into the Xsheet orientation & then back to the Timeline when I know what I am looking for. This is especially important when I am helping a student troubleshoot a problem with a level.

The Style Editor also has an orientation toggle with the double arrows. The function of the button might be less clear than the proposed button but is much smaller and would add to overall consistency. If the button is switched to being visual, I think it is important to have a Timeline/Xsheet label somewhere for new learners.
timeline
xsheet

snapai:
I toggle when I want to switch from considering a long span of time to considering multiple layers at once (assuming I'm working with <10 layers). Also if I want to see longer notes in the note layer(s).

(Unrelated to this PR, might be nice to have a notes panel that shows longer fields of all the notes for the current time - and an ability to output timed text files from these layers during the render - either as timed text files or a mux'd MP4 caption stream. Possibly also an FX that can take a notes layer as its input and render it like the timecode one. That'd be useful for adding captions or other timed text applications like presentations)

manongjohn:
I've added a menu item to the "Windows" menu to open directly to Timeline.  Something I had wanted to due originally when timeline was introduce but didn't know how....until today.  Easier than I thought it would be :)

I am still keeping the toggle button for now.  I had to rework the icons suggested by @snapai  (thanks btw!) because they didn't render well. I came up with these for now:

image

There are now tooltips when hovering over the buttons in this panel so they know what the icons are for.

gab3d:
perhaps the New Note icon width could be reduced a little?
at least to me, in your image by seeing it side by side with the new icons it seemed as much more "important".

manongjohn:
I changed the button type back to what it was so now it's shaded like before.  Here's what it looks like now.

image

artisteacher:
Does the height of the Xsheet button work ok when the column header is set to the Compact?

bobbybutterfly:
The Button does not sit well in the UI. The Button itself looks nice, but it does not match with the other buttons placement and size

manongjohn:
Played around with the button sizes to make them the same height and match widths (note and xsheet buttons). With the wider button,  I changed the button image as well.  Lastest and greatest:

image

bobbybutterfly:
@manongjohn Perfect!

jpturcotte:
@manongjohn I know it is probably stretching it, but is it safe enough to be included in v1.2.1? cc @shun-iwasawa

manongjohn:
Yes. It should be safe to include. It would be great if we could include it with 1.2.1.  

It is ready for review.   The PR description and images have been updated with the latest.

Agnyy:
I have such suggestions for the visual part.

1) Indicates the beginning and end of the track. (Similarly to Adobe Flash)
image

2) If the layer is empty then the circle is empty.
image

3) Arrows need to be changed. 
image 
Make them more neat and black (if on a light background) 
image
It can look like this.
image
or so.
image

4) Desaturate the head of the layers. And make a visual selection of the layer. (Now, there is nothing to catch on the eye and the dedicated layer is constantly lost) (like to Adobe Flash)
image

5) The selection of the frame should also be more noticeable.
image

6) The audio track can be differentiated.
image

7) dragging frames not for a strip.
image
The system that works now very often discourages new users. I want to move to a more understandable way.
Normal drag and drop, while holding the frame with the mouse button.
image
To drag multiple frames. You need to select them with the Shift button.

The same applies to the heads of the layers.
image


These comments are in my opinion one of the most necessary.

artisteacher:
I think there are advantages & disadvantages to how columns/layers and frames are selected and moved in the timeline. I really like the ability to click & drag to select multiple columns/layers or cells. I think it’s faster than the traditional shift-select method of selecting. On the down side, it isn’t a standard/expected behavior like shift-select is. Still, I would hate to lose it. 

As far as rearranging goes, it is often easier to grab the dragbar in the Xsheet than it is in the timeline. Again, dragbars aren’t really a standard behavior so new learners have no idea what they do or how to reorganize. Moving to a standard drag and drop behavior might be really beneficial, as long as it doesn’t interfere with other functionality.

I think the active layer/column could be made more obvious, though I wouldn’t completely desaturate the headers myself. Color cues are helpful.

konero:
A point.
- Clip Studio has a drag handle but it relates to a cel stack, it's not singular (per cel). To move each cel you click and drag it, here's a gif.

clip-time

manongjohn:
I'd have to refresh my memory, but shift/ctrl-drag activates additional actions in the drag that you may not want all the time. Would need time to evaluate what shift-drag does and possible workarounds if dragbars were removed.

Personally I like some consistancy in certain behaviors between timeline and xsheet. The timeline is technically just an xsheet on its side, afterall. Though the appearance is obviously different, how you manipulate things is the same between them. In one room you may prefer the Xsheet and would then need to learn a different set of actions to do the same thing in timeline. That would frustrate me more.

At this point, any more changes other than a bug fix or review change requests will likely prevent this from going into next release, if it can even make it at all. 

Granted, once it becomes more official with its own room, what is there would become OT's standard and taking things away tends to piss people off. I can see how changing things now vs later may be prefered.

As far as OT's look...we should have our own look, not copied versions of everyone else's. Most of that is subjective anyway. I want to be able to scan streams and videos and go "Hey, that's OT!".

Embrace the uniquenesses of OT. 😁

My 2 cents.

konero:
I've found some UI bugs so far in this PR, I'll document them soon.

I wonder if this should be postponed so a wider discussion can be had, I like some of the points raised by @Agnyy and I myself was a long-time Flash user, some Shin Chan and other anime made good use of it and the Dofus movie by Ankama Studio used it entirely. I know a lot of people rag on Adobe, but to their credit they set a lot of standards and terminology people learn in school which makes learning how to use something all the more easier.

When I worked full-time in UI/UX we would often look at what Adobe or Google was doing. Back when the first iPhone launched, we just copied Apple UI, these days though Google is winning the UI battle.

I'll have another crack at a design mock taking into consideration @manongjohn's preference for flat interfaces and @Agnyy's user-experience orientated conundrums.

konero:
Hmm, there's a lot to think about in terms of a re-design. I honestly don't have the time but here's a quick attempt combining @Agnyy and @manongjohn preferences.

agny-des

**Burger menu**:
- Display Frame
- Display Sec Frame
- Display 6sec Sheet
- Display 3sec Sheet
- **Separator**
- Switch to XSheet

**Plus Menu**:
- New Raster Level
- New Toonz Raster Level
- New Toonz Vector Level
- Etc.

**Minus Button**:
- Delete Level

**Arrow on Level**:
- Show Parameters/Objects/Etc

**Double Arrow Box**:
- Drag to Zoom

konero:
For the UI bugs currently, these are the ones I can notice.

**When the keyframe icon overlaps the start frame icon**:
bug1

**The horizontal separator line is offset**:
tml

manongjohn:
Will fix the overlap.

The horizontal offset was fixed with a commit I pushed last night. The offset issue only occured on initial scene load and would correct itself after the area was scrolled.

I had removed the blue numbers between the play area. Seemed a bit much with the hightlighted area but can easily put it back. Was thinking of doing that anyway.

artisteacher:
I really prefer keeping the Xsheet/Timeline toggle visible for easy access. I use it often enough that needing an extra step to access the command would be really inconvenient. With the command bar, I’m not sure how necessary the + and - buttons really are.

openanim:
We talk a lot about the integration of new users, but in my opinion the xsheet is a scary point for many new in the animation in short. This can do scary  to the news users, and that of OT although it is incredible does not escape the rule. And the introduction of timeline has probably helped a lot to integrate new users, I think the more it will remain consistent, at least in its design with the xsheet,  more the news users who are familiar with timiline will be able to use xsheet

manongjohn:
@openanim Not doing the classic/classic-revised for timeline this time. Need to draw a line somewhere. 😊

As for @konero's mockup...not a fan. Feels like we're copying someone else's style for what? Familiarity? As I have not used any other animation software, I have no attachments to these other "styles" as I am used to OT.

That said, I'm not against changing some stuff. This originally started as clearing the "noise" from the timeline but seems to be turning more into a redesign session.

AyameLee:
Hey, this seems like a terribly awkward exchange you guys are having. ☹️

Not really sure what problem this is meant to solve though, it actually looks more difficult to use since we can't see where individual cells are now on a hold. 😬 

But I don't use the timeline so make of that what you will.

Solidarity people. ❤️

artisteacher:
I believe it was originally about making the timeline more readable - especially when zoomed out. Which this PR does very successfully as is.

That is a good point about being able to easily identify the cell - though it’s not necessarily a bad thing to leave some strengths for the Xsheet over the Timeline.

There also definitely needs to be a certain level of consistency between the Xsheet & Timeline. Any big changes that impact both are probably something for a different PR - assuming there needs to be a change. After thinking some more, the current selection and drag behaviors seem more flexible and sometimes faster than what some other programs seem to use - even if they are less obvious. 

We’re never all going to agree on everything but I think it’s great that everyone is passionate about making OT the best it can be!

Agnyy:
@konero Said he was missing.

8) I think it would be appropriate to add small sticks from below. 

before
image

After
image


As I understand it, not all the points I proposed you liked. I would like you to write out the numbers that you liked and by a common vote we'll decide that the truth is still needed. And what about the unique type of OT and so on. That such logic ruined many free programs that refused to transfer effective solutions for fear that they are capturing competitors. As a result, it was hard for new users to start using the program, and old users stopped comparing the program with others and so the projects got stuck. I'm only looking for the most effective and fastest way to work in the program. I also taught OT for a long time to adults and children and I know what they did not like and then they went to the advantage of FLASH. Now I work well for a big cartoon studio and I know why there many did not like OT and why they do not want to use it. It turned out that out of 100 people only 15 can continue to experiment with the OT and then later and these people often abandon the program. (These are approximate figures). I try to take everything into account. I think we need to more often take into account the opinion of users of other programs and conduct surveys. I believe that OT will be the best program.

manongjohn:
My thoughts on the suggestions

1) Indicates the beginning and end of the track. (Similarly to Adobe Flash)

   Honestly, it is not needed. There is already a line separator and dot to denote the next frame.
   Just because Adobe Flash does it doesn't make it a standard or the best thing to do.  It just adds to the noise.

2) If the layer is empty then the circle is empty.

   Not sure I can easily do this, though the dot was just to indicate there is a frame there, not whether or not it has anything in it.
   Again, copying Adobe Flash. Is it really necessary or is this really a nice to have?  If it's a nice to have, I'd prefer not to have it and focus on developing other more important stuff.

3) Arrows need to be changed / Make them more neat and black (if on a light background)

   The arrows should probably remain white.  When the line or the ease handles are over a black frame marker they blend in and hard to see where they are.  In my opinion, they should be white regardless of theme since the keyframe marker is always white and the line is part of the keyframes in general.

4) Desaturate the head of the layers. And make a visual selection of the layer. (Now, there is nothing to catch on the eye and the dedicated layer is constantly lost) (like to Adobe Flash)

   I've already added a preference so you can change the color of the current layer to make it more readable and noticable per your preference. For me, cyan works best in default theme instead of the red we currently have.

   When I don't have exposed frames on a level in my timeline because they aren't in that frame range, I can easily tell what type of level it is by looking at the header's color.

   I'd prefer to keep this since that is how the xsheet is done but if there is overwhelming consenus to change it, that's fine.

5) The selection of the frame should also be more noticeable.

   Not sure what you are showing in the selection of the frame.  Anything done here should also be done in Xsheet for consistancy.  Please keep that in mind.

6) The audio track can be differentiated.

   I'm indifferent to this, though that probably means a new QSS setting for it and additional work.  This should likely affect Xsheet also for consistancy in appearance.

7) dragging frames not for a strip.

   I've already explained why we should keep the dragbars. We should be consistant with xsheet, therefore I do not think this should be done.

8) I think it would be appropriate to add small sticks from below.

   I already added that before I saw your message.


If I'm coming across like a jerk, then I truly apologize. But please don't make it sound like I'm ignoring others ideas and opinions. I am trying my best to keep an open mind to every comment and suggestion that is made even when I don't agree with them. I do have my own opinions and thoughts and will let it be known when I disagree. Dealing with UI changes is a very frustrating processing as everyone has their own preferences and tastes, their own ideas and opinions, their own experiences to draw upon and my job is to somehow turn that into something agreeable that I can do. I know we can't please everyone, even me.

I also want to make sure people can work efficiently in OT.  I don't necessarily agree these are "effective" solutions just because Adobe did it.  I don't see some of this stuff in TVPaint's UI and yet that is a very popular animation software from my understanding.

At any rate, pull the list together and vote on it however you want to do that.  I'll put this back in WIP since I don't know when we'll come to an agreement or when I'll finish changing it. It will not likely get into this upcoming release either way.

Agnyy:
I do not demand to do everything strictly by my request. I share my observations and surveys. You know that I do not know much English and maybe my comments sound very bold. But in fact, I'm just commenting. In any case, I know very well how hard and much it is to work. And I really respect work. I think that it makes no sense to translate this into WIP.  Much has already been done and well enough. And my comments were mainly directed to pay attention and hear the opinions of others. Yes, I very often cite the example of Adobe, since they are the most invested in the development of good UI and UX. Their products are many very simple to visual and easy to use. But I will correct in advance, I do not particularly like the company of Adobe, because for me they do little new things, but only squeeze money out of users. Once again, I reminded me of observations and just wanted to hear opinions. I think that the ideal solution will be to add extended settings for the timeline, so that everyone can adjust the scale to himself. If interested, I can write out a schedule of settings parameters in the timeline, which could allow the user to redesign it for themselves. This I think will be the best decision. Everyone will be happy.

Agnyy:
@konero  And what did you do your layout in? in which program? can you give. I would like to change something and watch more fun. 
image

konero:
No worries, no one's coming across like a jerk, it's good to brainstorm. 🤝

I respect you haven't used other animation software so things could have been explained in greater detail.
The timeline I was most referencing was Harmony as it's what most people will probably be working with today in TV, so yes "familiarity" could best describe it, I'm just looking at the common denominator, it's one of those tools you need to know how to use if you want to work in a lot of studios.

Here's a good primer that shows how powerful it is: https://www.youtube.com/watch?v=08WlmuLAjFA
And one for their XSheet: https://www.youtube.com/watch?v=42guwCUzoKQ

Granted they have a large developer team, and I never set out to go the whole hog anyway.

My own experience when I first used Harmony was how similar looking to Flash it was, since I was a Flash user at the time doing traditional animation it was easy-peasy to switch to.

But go with what feels right for OpenToonz.

har12_interface_general

konero:
@Agnyy I made it with [Inkscape](https://inkscape.org/).

You can download the SVG file here: [tztimeline.zip](https://github.com/opentoonz/opentoonz/files/2087881/tztimeline.zip)

I used many layers, so make sure to open the layers panel (CTRL+SHIFT+L).

beeheemooth:
What is the difference between these two images?

new canvas

(Correct answer: in the first one 10 frames are selected, and in the second - three frames.) ^__^

manongjohn:
@beeheemooth Thanks for pointing that bug. I never noticed that before 😁

It appears the colors are static for that and never defined a selected cell color for it. I will move colors for that to stylesheet and define a selected color for it.

VrKat:
comic2
comic3

openanim:
" 8. I think it would be appropriate to add small sticks from below."
https://user-images.githubusercontent.com/25137502/41198811-2b2cc64c-6c8f-11e8-9c7b-60fefcc22d53.png

https://user-images.githubusercontent.com/25137502/41198814-417a3006-6c8f-11e8-8705-47bc1d82a0b4.png

 @Agnyy 

I suggest you test the PR you will see that the cell is already recognizable. There is really no need to add this
For me it's a step back from this pr if you want to remove noise.

Here is the cell as it looks when selected.

timeline_1

In a pinch I would prefer a highlighting by a small colored area if it was really necessary to modify rather than these small lines that looks like the old timeline look of harmony. 

timeline_2

Personally, at one time I used harmony a lot, before switching for others. Honestly, I'm not as excited about the design of their timeline as you are, Even less on their xsheet. I find the workspace very rough, like a gas plant. Adobe in the text in less good. But I like some of their very powerful features including traditional markers that do not have OT. I much prefer TVpaint design level whether xsheet and timeline and both are very consistent until the thumbnail display.

Agnyy:
@konero Thanks for the layout! Спасибо!
And I have quite a lot of experience using programs. And of course I'm familiar with both ToonBoom and TVpaint.

I've heard many comments and then suggest for the beginning to do more cosmetic editing. The purpose of these changes will be simplification.


1) make the vertical lines lighter. 
Before
image
image
After.
image
image

2) Reduce the circle a little. 
image
This option looks good.
image

3) If you can not paint the arrows in black. then you can repaint them in white and remove the stroke!
image

4) Visual separation in the dragbars area.
image
                          а) You can either draw a strip from above.
                          b) or use one of the suggestions.
image

5) For the onion skin, remove the stroke and make the colors more clear.
image

6) Also remove the stroke. And make a quieter color.
image
image

7) Make the selected area of the animation more obvious.
image
@konero in his layout very well showed a clear selection of the animation loading.
image

Thank you all very much for hearing me and offering your ideas. I respect all of you!

artisteacher:
1. The after view looks a little too subtle 
2. The circle could certainly be a bit smaller, but keep the dragbar 
3. Does look better but may not work with every level color
4. Since the dragbar can select the entire level, I think it’s fine not having any division
5. The stroke may or may not be necessary but keep some sort of icon
6. If the stroke isn’t necessary for lighter themes, then it’s fine to remove it. I wouldn’t take away too much contrast though - as it is my students rarely notice those markers until I specifically point them out 
7. I think the visibility of the selected area for the animation has already been improved - though perhaps the change in background color could extend down the the empty cells as well. Also due to the difference in scale those pictures aren’t quite a fair comparison 🙃

openanim:
Personally, I do not understand why you absolutely stick to his sticks, there are many timelines that do not have and that are commonly used. But you must surely have your reasons, I have the impression that it increases the noise but if it helps to make things happen, why not. I also share the idea that both lines of the skin onion can be removed.

beeheemooth:
Guys, I'm satisfied with the color of the frames (except Note Level), but if you still decide to make selected frames of another color, I beg you, _**do not make them darker**_ !

manongjohn:
My thoughts and opinions

1) The color of these lines is changeable in the stylesheet BUT it is shared with the Xsheet as are all colors for xsheet/timeline. If the change looks nice on xsheet too then I just need the RGB values to use.

If they must be different, then i would have to add code. I would likely just lighten the defined color by some % for timeline mode only in this case.

2) I will attempt to shrink it down likely by 1px more which turned it into a square the last time I tested.

3) Without the outlines (follows text color actually) the white will bleed out into surrounding colors in the Light theme making them look fuzzy. I will keep to using text color for stroke (white in dark themes, black in light theme)

4) the bar when clicked auto selects all contigious cells for that level. Breaking that up visually will be confusing to people as they may expect it only for that cell and wonder why it is selecting more than it should.

As this is the behaviour also in xsheet, i do not think this should change

5) I prefer to keep the onion marker for consistancy with xsheet. Also keep in mind when Onion Skin is turned off there is a different current time indicator header. Plus when zoomed in, it will look strange having either a wide header with thin line, or a small marker (if it does not fill width of cell) in a wide cell.

I think i will experiment by making the line between the fixed and relative markers more transparent so it's a little more pleasing to look at while still hinting where to click, especially for mac users. 

6) Without the outline, the colors will bleed too much and make them fuzzy especially when not active (grey) in dark modes or active (white) in light theme. These markers should be clear, in my opinion. Again, i would prefer this match both in xsheet and timeline for consistancy.

7) Not quite sure I understand. The arrows are pointing to what you talked about in # 6.

I do rather like @artisteacher's idea of lightening empty cells more where the play range is active. This may require lightening the cell selection be even lighter. All this might be too much with the different levels of backgrounds and unintentionally add noise. Needs further exploration.

8) Regarding the frame markers between held cells.l: I added them to make it easier to find a cell in held frames. I'm 50/50 for it. I like it without but it does take the guesswork out of finding a cell you may want to click into for whatever reason. Definitely hard to count the cells also without it. 

I only have it at 1px high which did add back a little noise. 2px high made it noisy again and just looked odd. However, maybe lightening it similar to # 1 will make it better, like a subtle shadow, so it's not as visually distracting when you are not directly looking for it. Will experiment.

beeheemooth:
Is it possible to move all these disputable parameters for UI colors in Preferences? Or are they hardcoded?

konero:
Don't use RGB values in stylesheets if possible for border lines, use LESS color functions. I'm pretty sure it already uses a color function, ie `darken(@bg, 5%);`.

bobbybutterfly:
@beeheemooth  i agree!

For me, struggling with a slight red green blindness, its very hard to read the red letters of a selected column.
The possibility of changing some colorprefs of the UI would be a great relief

manongjohn:
@beeheemooth most of the colors for xsheet/timeline are controlled by themes and would not make sense as Preferences in my opinion. 

@konero good point. Will do that

@bobbybutterfly Included in this PR is the ability to select your own active column color which will help you.

modorov:
you turn op into flash

jpturcotte:
@modorov We do not need comments like that. If you have something to say, be clear, and argue your case. Thank you!

openanim:
Personally, I'm not really convinced because we have the same markers in white above but if it suits everyone, no problem. A check box in the view menu is not something that can be considered for this and onion skin lines ? In all cases, the timeline has been improved to the previous one, imo. Let's stay positive :)

artisteacher:
If hovering worked with a tablet on OS X, I'd say we could skip the guideline for onionskin. Since hovering does not work, it's a really, really helpful addition.

manongjohn:
@artisteacher Unfortunately, for now, I'm actually going to remove those guidelines.  It was making the UI look odd and I couldn't easily lighten them efficiently for the different themes with 1 setting.  However, with the play area different color between the play markers, the space above it with which to press for onion skin marks is becomes defined.  Just press closer to the top of the frames area for the fixed onion skins or closer to the play mark area for the relative ones.

Regarding the hovering issue, any idea which nightly that might have started in?  I'm a bit curious what is causing that.  Does hovering not work anywhere in OT for OS X users then or just in the Xsheet?

artisteacher:
That will work.

I'm not positive, but it might be #1659. I'll double check & make sure. It's definitely a problem beyond the Xsheet.

artisteacher:
Actually it looks like the hover issue goes at least as far back as 1.1.3. It seems that if a tablet is connected when OT opens, hovering works as expected. If the tablet is connected after, then hovering doesn’t work.

manongjohn:
Changes made per suggestion:
1) I programatically lowered the alpha on the non-marker separators between empty cells to blend in more making it less prominent.  I lowered the alpha just enough so it didn't cause the lines to disappear too much in the Light theme's empty cell background. The change does not affect xsheet mode.
2) Reduced circle by 1px.  If I reduce by 2x it turns into a square.  I think 1px is enough, but if you still want smaller then I'd probably reduce by 1px more and we'll have squares instead of circles.
3) I've modified the diamonds and the arrows to have a black outline and changed the color of the ease handles to always white.  This affects the Xsheet and it looks acceptable to me.
4) Not changing.
5) No changes made to onion skin icon.  I did decide to remove the lines for the fixed/relative onion skin sections.  It looks rather odd and does make it look busier than it needs to be.
6) Not changing.
7) Nothing done so far as I'm unclear what is being request.
As far as lightening up empty cells within the play range, after a quick mock up, I decided not to implement. 
8) Same as # 1

Image updated in body

Additional changes:
- Can turn off marker intervals with 0
- Zoomed out frame numbers match marker intervals in "Frame" mode only. Defaults to every 5 if marker intervals turned off. Other modes do not affected

Agnyy:
5. Personally, I liked stripes. 

image

I suggested removing only the outline of the circles.

image

image

I very much ask you. Can we suggest these changes in 1.1.2? Now it looks very good. The rest of the change is followed by a crown.

artisteacher:
Keep in mind there is a lot less contrast on the light theme - which is probably why there is an outline.
light1 light2
Also, at least in the current light theme, the markers are the same grey as the background.

manongjohn:
How's this (enabled/disabled states):

image

Agnyy:
cool! I like!

manongjohn:
PR body contains updated image of timeline showing both Default and Light themes.

beeheemooth:
Excellent! Yellow dots and lines look great! But the red dots on the background look a little dark. Can you make them brighter? In an extreme case, you can leave the current purple color, it will look good, but without strokes.

openanim:
Maybe you'll get it by choosing a lighter red in the onion skin menu?

modorov:
@jpturcotte, ok I turned not to you but the author of the post. timeline similar to flash is very small and spoils the eyesight

artisteacher:
@modorov - keep in mind that most of these changes are to help when the timeline is really zoomed out and are a huge improvement over the current overly crowded look. Fortunately, if you don’t like having things so small, you can zoom in on the timeline. I think this is a good blend of solutions for differing needs while allowing for a little more customization than we currently have available.

modorov:
@artisteacher and user comments also help make the product better.

jpturcotte:
@modorov 
> you turn op into flash

How is this comment helping make things better? Maybe you could try telling us the problems you have, and propose solutions instead.

RodneyBaker:
@jpturcotte It's easy to read into a comment but you may be raising this to a higher level than required..  I just read the comment for what it is worth... a note coming from someone with concerns that the current (or proposed) enhancements resemble that of Flash.   (that can be good or bad depending on any given users point of view).  Perhaps @modorov should have simply said, "I'd prefer not to move in the direction of Flash." and that would have been enough to weigh in on the subject.  

For long drawn out discussions on the pros and cons of UIs I'll recommend the OpenToonz forum where I'm sure lots of folks would love to add their thoughts into the mix.  

The Timeline is coming along nicely. 
Thanks for the hard work @manongjohn! :)

shun-iwasawa:
Jenkins

toonz/sources/toonz/xshnoteviewer.cpp
```diff
@@ -475,33 +474,42 @@ NoteArea::NoteArea(XsheetViewer *parent, Qt::WFlags flags)
 
   //-----
 
-  m_flipOrientationButton->setObjectName("PushButton_NoPadding");
+  m_flipOrientationButton->setObjectName("flipOrientationButton");
   m_flipOrientationButton->setFocusPolicy(Qt::FocusPolicy::NoFocus);
   m_flipOrientationButton->setFixedSize(QSize(70, 23));
+  m_flipOrientationButton->setIconSize(QSize(40, 20));
+  QIcon flipOrientationIcon = createQIcon("fliporientation");
```
shun-iwasawa: IMO this line should be just `QIcon flipOrientationIcon;` since you don't register the resource `fliporientation.svg`.
This line seems to cause alert message as follows on the debug console :  
`There qt.svg: Cannot open file ':Resources/fliporientation.svg', because: No such file or directory`

toonz/sources/toonz/xshnoteviewer.cpp
```diff
@@ -627,7 +627,15 @@ void NoteArea::updateButtons() {
 void NoteArea::flipOrientation() { m_viewer->flipOrientation(); }
 
 void NoteArea::onXsheetOrientationChanged(const Orientation *newOrientation) {
-  m_flipOrientationButton->setText(newOrientation->caption());
+  //  m_flipOrientationButton->setText(newOrientation->caption());
+
+  QIcon flipOrientationIcon = createQIcon("fliporientation");
```
shun-iwasawa: For the same reason, I think this line should be just `QIcon flipOrientationIcon;` as well.

Agnyy:
This add-on is very necessary to add in 1.2.1.  @shun-iwasawa

shun-iwasawa:
I agree with you @Agnyy . I briefly reviewed and found this PR has almost no problem. 
I'm considering to release RC2 as soon as this will be merged in order to have some extra days for testing recent changes & fixes.

manongjohn:
@shun-iwasawa Fixed. Thanks for reviewing and including in 1.2.1!

shun-iwasawa:
Jenkins again

shun-iwasawa:
LGTM Thank you @manongjohn for this improvement and thank you everyone for having the deep discussion!

PR #1953, opened 2018-05-08
Fix of Brush Tip Display

shun-iwasawa:
This PR fixes some problems regarding displaying and refreshing of the brush tip (= a red circle displayed when using Brush tool).

- Fixed #1950 . ( Hopefully it will work. Please note that I tested only on Windows yet and will remove `WIP` after checking on OSX. )
- Fixed #1899. A problem that the brush tip stuck at the center of the viewer and does not move along the cursor when the first use of the brush tool after launch. It will fix the similar problem in all other tools as well.
- Enhanced the brush tip response when moving the cursor without any mouse buttons being pressed.  ( I introduced partial GL update for `BrushTool::mouseMove()` as well. )

PR #1991, opened 2018-05-21
Additional fix of Brush Tip Display issue

blackwarthog:
Additional fix to PR #1953 

We still have bug in scene invalidation functionality. When we called SceneViewer::GLInvalidateAll() and after that SceneViewer::GLInvalidateRect(), then invalidation rectangle will reduced. This PR fixes it.

### DIFF
--- a/toonz/sources/toonz/sceneviewer.cpp
+++ b/toonz/sources/toonz/sceneviewer.cpp
@@ -472,6 +472,7 @@ SceneViewer::SceneViewer(ImageUtils::FullScreenWidget *parent)
     , m_eraserPointerOn(false)
     , m_backupTool("")
     , m_clipRect()
+    , m_dirty()
     , m_isPicking(false)
     , m_current3DDevice(NONE)
     , m_sideRasterPos()
@@ -1363,6 +1364,7 @@ static void drawFpsGraph(int t0, int t1) {
 //#define FPS_HISTOGRAM
 
 void SceneViewer::paintGL() {
+  m_dirty = false;
 #ifdef _DEBUG
   if (!check_framebuffer_status()) {
     /* QGLWidget の widget 生成/削除のタイミングで(platform によって?)
@@ -1689,6 +1691,7 @@ bool SceneViewer::is3DView() const {
 //-----------------------------------------------------------------------------
 
 void SceneViewer::invalidateAll() {
+  m_dirty = true;
   m_clipRect.empty();
   update();
   if (m_vRuler) m_vRuler->update();
@@ -1706,6 +1709,7 @@ void SceneViewer::navigatorPan(const QPoint &delta) {
 //-----------------------------------------------------------------------------
 
 void SceneViewer::GLInvalidateAll() {
+  m_dirty = true;
   m_clipRect.empty();
   update();
   if (m_vRuler) m_vRuler->update();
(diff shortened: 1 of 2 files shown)


toonz/sources/toonz/sceneviewer.cpp
```diff
@@ -1715,12 +1719,9 @@ void SceneViewer::GLInvalidateAll() {
...
 void SceneViewer::GLInvalidateRect(const TRectD &rect) {
-  // there is a case that this function is called more than once before
-  // paintGL() is called
-  if (!m_clipRect.isEmpty())
-    m_clipRect += rect;
-  else
-    m_clipRect = rect;
+  if (m_dirty) m_clipRect += rect;
+          else m_clipRect  = rect;
+  m_dirty = true;
```
shun-iwasawa: > When we called SceneViewer::GLInvalidateAll() and after that SceneViewer::GLInvalidateRect(), then invalidation rectangle will reduced. This PR fixes it.

I understand that such scenario will define the `m_clipRect` and will fail to update whole viewer. Good catch.

However I wonder whether this fix will resolve it.
In the `GLInvalidateRect()`, regardless of `m_dirty` value `m_clipRect` seems to be set to some value , which will activate clipping anyway.

Instead, I think you need to set `m_dirty` to true ONLY when `GLInvalidateAll()` is called, and if `m_dirty` is true `m_clipRect` should be kept empty in `GLInvalidateRect()` in order to update whole viewer.

blackwarthog: Calling GLInvalidateRect with empty rectangle should be equal to invalidate all. This is behavior of previous versions of GLInvalidateRect function.

I think the valid solution is:

```
void SceneViewer::GLInvalidateRect(const TRectD &rect) {
  if (!m_dirty || rect.isEmpty())
    m_clipRect = rect;
  else if (!m_clipRect.isEmpty())
    m_clipRect += rect;
  m_dirty = true;
  update();
  if (m_vRuler) m_vRuler->update();
  if (m_hRuler) m_hRuler->update();
}
```

Or should we change behavior GLInvalidateRect to consider empty rectangle as empty instead of infinite?

toonz/sources/toonz/sceneviewer.cpp
```diff
@@ -1715,17 +1719,15 @@ void SceneViewer::GLInvalidateAll() {
 //-----------------------------------------------------------------------------
 
 void SceneViewer::GLInvalidateRect(const TRectD &rect) {
-  // there is a case that this function is called more than once before
-  // paintGL() is called
-  if (!m_clipRect.isEmpty())
-    m_clipRect += rect;
-  else
+  if (!m_dirty || rect.isEmpty())
     m_clipRect = rect;
```
shun-iwasawa: As far as I understand, the purpose of this PR is to separate the following two cases:

- `GLInvalidateRect()` is solely called. (This is normal case)
- `GLInvalidateRect()` is called promptly after `GLInvalidateAll()` before `paintGL()` is executed.

In the former case, Viewer should be updated partially according to the clipping rectangular.
However, in the latter case, whole Viewer should be updated regardless of `GLInvalidateRect()`.
For now `m_clipRect` will be empty for both cases so we need to take measure to distinguish them.

> Or should we change behavior GLInvalidateRect to consider empty rectangle as empty instead of infinite?

Well, yes, in `GLInvalidateAll()` setting `m_clipRect` to `TConsts::infiniteRectD` instead of empty rectangular would be good idea. :+1:

blackwarthog: > Well, yes, in GLInvalidateAll() setting m_clipRect to TConsts::infiniteRectD instead of empty rectangular would be good idea. +1

It will very good, but i 'm afraid that in this case we shoud to investigate all cases when GLInvalidateRect called from the other code. Program may to call GLInvalidateRect with empty rectangle with meaning "invalidate all".

PR #1993, opened 2018-05-22
tfarm binaries into .app folder by damies13 #1141 (modified)

### LINKED ISSUE #1130: tfarmcontroller & tfarmserver on Mac OSX

shun-iwasawa:
This PR is based on #1141 by @damies13 and will resolve #1130 .

For unknown reasons setting `CMAKE_RUNTIME_OUTPUT_DIRECTORY` fails to build with current master.
So I took somewhat dirty solution to copy tfarm-related executables onto `OpenToonz_1.1.app/Contents/MacOS` after build.


masafumi-inoue:
LGTM

PR #1994, opened 2018-05-22
Remove Old Stuff Folder Check from OSX tfarm Executables

shun-iwasawa:
I found a problem while testing #1993 (former #1141) with #1609 by @damies13 , which will enable batch rendering and batch cleanup in OSX.

For now in `tcomposer` and `tcleanup` there are unnecessary operations handling old stuff folder path and it causes crash.
Confirmed that just removing the lines works fine.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2005, opened 2018-05-24
Fix brush using wrong initial thickness value

manongjohn:
This PR fixes a minor issue I accidentally encountered with the Vector Brush. 

The issue is when you click to start drawing with the Vector Brush with Pressure sensitivity turned ON and you start at max pressure (or using a mouse), it uses the Min Size value of the Toonz Raster Brush setting instead of the Toonz Vector Brush Setting.  It becomes obvous if the Toonz Raster Brush min size setting is higher than the Toonz Vector min size.

### DIFF
--- a/toonz/sources/tnztools/brushtool.cpp
+++ b/toonz/sources/tnztools/brushtool.cpp
@@ -1335,7 +1335,7 @@ void BrushTool::leftButtonDown(const TPointD &pos, const TMouseEvent &e) {
 
     /*--- ストロークの最初にMaxサイズの円が描かれてしまう不具合を防止する ---*/
     if (m_pressure.getValue() && e.m_pressure == 1.0)
-      thickness     = m_rasThickness.getValue().first;
+      thickness     = m_thickness.getValue().first;
     m_currThickness = thickness;
     m_smoothStroke.beginStroke(m_smooth.getValue());
 


openanim:
@manongjohn 
I don't know if it's related, but it's also a variation of unexpected thickness. I often notice it in the bottom of the viewer. I don't know if it's reproducible. The line becomes thicker down.
variation thickness

shun-iwasawa:
Jenkins

toonz/sources/tnztools/brushtool.cpp
```diff
@@ -1335,7 +1335,7 @@ void BrushTool::leftButtonDown(const TPointD &pos, const TMouseEvent &e) {
 
     /*--- ストロークの最初にMaxサイズの円が描かれてしまう不具合を防止する ---*/
     if (m_pressure.getValue() && e.m_pressure == 1.0)
-      thickness     = m_rasThickness.getValue().first;
+      thickness     = m_thickness.getValue().first;
```
shun-iwasawa: @manongjohn Thank you for tackling with this problem.
I myself inserted this line when I was in Ghibli, and it's obviously wrong. I couldn't realize it until today as Ghibli had mostly used Toonz Raster levels. 

Just one point I would like you to fix - it seems that this line seem to be 
`thickness     = m_thickness.getValue().first * 0.5;`
in order to get smooth stroke, since `computeThickness()` returns half of the thickness value.

manongjohn:
@shun-iwasawa  Fixed :)

shun-iwasawa:
@manongjohn :+1: 

@openanim Thank you for reporting. I could reproduce the problem and found that it is unrelated to this PR. I posted it as another issue #2017 .

shun-iwasawa:
Jenkins again

Issue #2012, opened 2018-05-25
CLI (tconverter) segfaults when converting Toonz Vector Levels (.pli)


morevnaproject:
### --Issue Summary--
When calling CLI tool "tconverter" segfaults when converting Toonz Vector Levels (.pli):

```
$ opentoonz --appimage-exec tconverter  ./sample-files/ep01/drawings/ep01-001/D.pli  /tmp/zzz/f.png
QCoreApplication::applicationDirPath: Please instantiate the QApplication object first

Loading D.pli
Level loaded
Conversion in progress: wait please...
QObject::connect: Cannot connect (null)::destroyed(QObject*) to QOffscreenSurface::screenDestroyed(QObject*)
Segmentation fault
```

Converting Toonz Raster Levels (tlv) works fine: 

```
$ opentoonz --appimage-exec tconverter  ./sample-files/ep01/drawings/ep01-001/C.tlv  /tmp/zzz/f.png
QCoreApplication::applicationDirPath: Please instantiate the QApplication object first

Loading C.tlv
Level loaded
Conversion in progress: wait please...
QCoreApplication::applicationDirPath: Please instantiate the QApplication object first

Conversion terminated!
```

Note: In the last example you can ignore "Conversion terminated!" message - files are rendered successfully. Looking at the source code of tconverter it is clear that there should be "Conversion successful!" instead of "Conversion terminated!" string.

### --Actual Results--
Segmentation fault

### --Expected Results--
A sequence of .png files created.

### --System Information--
+ OpenToonz Version:
1.2.0

+ Operating System:
Tested on OS Linux, for both native builds and appimages.
Needs to be checked on Windows/OSX.

morevnaproject:
The fix is requested by Melnitsa Animation studio - http://melnitsa.com/en/

RodneyBaker:
Closing this as likely resolved although unconfirmed.
Need Linux user feedback to this issue.

Recommend opening a new report with examples, video, project files in order to lock the issue down and target it for resolution.

Recommend referencing this issue in any newly opened report.

PR #2013, opened 2018-05-26
Fix tconvert tool for Linux (issue #2012)

blackwarthog:
Fix conversion of PLI-files for Linux (issue #2012)

By documentation for QOffscreenSurface:
Due to the fact that QOffscreenSurface is backed by a QWindow on some
platforms, cross-platform applications must ensure that create() is only
called on the main (GUI) thread. The QOffscreenSurface is then safe to
be used with makeCurrent() on other threads, but the initialization and
destruction must always happen on the main (GUI) thread.


turtletooth:
As the changes only affect Linux, I think this can be merged after conflicts are resolved.

PR #2024, opened 2018-05-30
Fix Tool Option to Be Aware of Another Enum Option's State

### LINKED ISSUE #2023: Tool Options are Not Aware of State of Another Enum Option on Non-English UI
### --Issue Summary--

After introducing #1863 tool option items fails to be aware of state change of other enum options.

### --Steps to reproduce--

This is one example:
- Choose Fill Tool
- Change `Mode` to `Lines` and then to `Areas`
- See `Onion Skin` checkbox becomes disabled when `Lines` is selected
- Change the current language to other than English and try the same procedure

### --Actual Results--

The enable/disable state of `Onion Skin` checkbox does not change.
Same problems occur in many part in Tool Option.

### --Expected Results--

They should update enable/disable state according to the current enum option value.

### --System Information--
+ OpenToonz Version: Current master branch

### --Cause of This Problem--
Looking [here in tooloptions.cpp](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/tnztools/tooloptions.cpp#L1666), OT calls `QComboBox::currentText()` for checking the current property state, but the return values are now already translated. They should be changed to see the current value of `TEnumProperty` directly.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59003758-tool-options-are-not-aware-of-state-of-another-enum-option-on-non-english-ui?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

shun-iwasawa:
This PR fixes #2023

Replaced all `ToolOptionCombo::currentText()` by checking the `TEnumProperty` values directly, since the item texts in the combo boxes are translated ones in non-English UI.

### DIFF
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -417,7 +417,7 @@ public:
   void updateStatus();
 
 protected slots:
-  void onShapeValueChanged();
+  void onShapeValueChanged(int);
   void onPencilModeToggled(bool);
   void onJoinStyleChanged(int);
 };
@@ -462,7 +462,7 @@ public:
   void updateStatus();
 
 protected slots:
-  void onColorModeChanged();
+  void onColorModeChanged(int);
 };
 
 //=============================================================================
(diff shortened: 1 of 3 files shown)


masafumi-inoue:
LGTM

Issue #1384, opened 2017-08-07
Missing libmypaint Linux build instructions


ideasman42:
Even after installing libmypaint I get the error that libmypaint is not found:

```
-- Checking for module 'libmypaint'
--   No package 'libmypaint' found
```

Could the Linux build instructions be updated?

ericandre615:
Same here. I tried libmypaint 1.3 and libmypaint2 it's in /usr/lib/. I had recently built opentoonz on another machine awhile back and it worked great. I've been looking around to see where it's looking for the lib, but haven't found anything yet.

jvault:
i

laher:
Thanks for the heads-up. I checked out the parent commit with `git checkout 14f0fd28bbeec42dfcebcebe59b858931eaeb881` and it built OK

retrotails:
If I change https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/CMakeLists.txt#L445
from this
`pkg_check_modules(MYPAINT_LIB REQUIRED libmypaint)`
to this
`pkg_check_modules(MYPAINT_LIB REQUIRED libmypaint-1.3)`
I can get it to compile fine on arch linux, though I still get
`/usr/bin/ld: warning: libgfortran.so.3, needed by /usr/lib/gcc/x86_64-pc-linux-gnu/7.2.0/../../../../lib/libsuperlu.so, not found (try using -rpath or -rpath-link)`
when trying to run the binary.
pretty sure superlu is out of date and needs rebuilt for libgfortran.so.4
if I manually copy over the old libgfortran.so.3 to /usr/lib it works

jvault:
Arch Linux.

Well, finely I can work this out. My solution was to manually move 
opentoonz/thirdparty/libmypaint/dist/32/include/libmypaint/mypaint-brush.h 
(along with the other scripts in the directory) to
opentoonz/toonz/sources/include
during building. That works is independent of global libmypaint installed.

Just as retrotails have mentioned to run the binary a copy of /usr/lib/libgfortran.so.4 must be renamed.

addslustertocollectibility:
https://community.mypaint.org/t/missing-libmypaint-pc-how-to-fix-it/379

here ya go!

ideasman42:
Committed change so this works in CMake (libmypaint package added to readme),
now other distros need to have their readme updated too.

ideasman42:
Closing, we can accept changes to build instruction PR's as they are submitted.

PR #2027, opened 2018-06-01
Instructions how to build OT with libmypaint dependency for Linux

blackwarthog:
issue #1384


turtletooth:
LGTM.  Thanks @blackwarthog

PR #2044, opened 2018-06-06
Fix Crash on Saving Curve in Function Editor

### LINKED ISSUE #1983: Crash using Save Curve In Function Editor
Finally getting around to reporting this one.
Sorry for the delay.

I confirm this process works in Toonz Premium but crashes in OpenToonz.

Steps to Reproduce.
- Animate something so that the Function Editor has data to work with such as moving a column North/Sount (N/S).
- Enter the Function Editor and select the animated item from the listing (N/S in this case).
- Right Click and select "Save Curve".
Normally this will save a .curve file out that contains the data associated with the curve/path of the animated object. 
OpenToonz will usually crash at this point.

In Toonz Premium the .curve file saves and reloads (via Load Curve) without issue.

Thanks in advance!
savecurvecrash

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58557564-crash-using-save-curve-in-function-editor?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

shun-iwasawa:
This will fix #1983 , the crash which seems to be due to transfer of the current selection.

I found that currently the `File` interpolation only accepts absolute paths to .dat or .txt file.
When you use a file browser popup from `...` button of the input field, it usually returns aliased path (with `+drawings`, `$scenefolder`, etc) which cannot access the file properly.
As a quick workaround, in this PR I also made the input field of the `File` interpolation not to use aliases.

### DIFF
--- a/toonz/sources/include/toonzqt/filefield.h
+++ b/toonz/sources/include/toonzqt/filefield.h
@@ -55,6 +55,7 @@ class DVAPI FileField : public QWidget {
   QString m_windowTitle;
   QString m_descriptionText;  // if the initial text is not path, set the string
                               // here and prevent browsing
+  bool m_codePath;
 
 protected:  // used in the child class for CleanupSettings
   QPushButton *m_fileBrowseButton;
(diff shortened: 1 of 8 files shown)


shun-iwasawa:
Jenkins

RodneyBaker:
I confirm that this fix works (via AppVeyor artifact) and allows saving and reloading of curve file data.

Thank you @shun-iwasawa !

masafumi-inoue:
LGTM

PR #2045, opened 2018-06-06
Add More Translatable Texts

### LINKED ISSUE #2043: Some Engligh strings still persist.
@shun-iwasawa 
Thank you so much for fixing my last PR. You're the best! :)

I've seen some trings that were suposed to be translated (I remember translating them for that last PR) but unfortunately in the UI, the English term still persists. Along with that, I've spotted some other ones that are currently untranslated and were not directly reported in #2020 .

**Untranslated**
- Xsheet header: "Table" string
- Animate Tool selection drop-down menu: "Table" string
- Function Editor tree: the "Table" string appears translated as "Table ([translation])". it seems as if the real name of the object is still Table, perhaps this is the cause of the above two cases?
- Convert (PLI to SVG) options: All strings

**Strings that are present in the PO file but not in use:**
- Fx Settings window: window title
- Camera options: "inch"
- Rotate Tool: "Rotate On Camera Center" is still untranslated.
- Histogram > Selected Color: R, G, B is still untranslated.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/59280494-some-engligh-strings-still-persist?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

shun-iwasawa:
This fixes #2043 

### Added

- Animate Tool selection drop-down menu: "Table" string
- Convert (PLI to SVG) options: All strings
- Fx Settings window: window title

### Fixed (i.e. you don't need to add new translation)

- Rotate Tool: "Rotate On Camera Center".

### Not translatable for some reasons

- Xsheet header: "Table" string
- Function Editor tree: the "Table" string appears translated as "Table ([translation])". 

They are showing an "object id" for table object rather than its name. The object id can be referred by expression and for now it accepts only non-translated strings.

- Camera options: "inch"

I need some time to make it translatable as it seems to be linked to OT's fundamental unit system.

---

@gab3d 
I couldn't find the text `Histogram > Selected Color: R, G, B` . Can you please show a snapshot of this UI?


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

gab3d:
Trying a reported bug regarding the "Apply Lip Sync Data to Column" I've catched another string that is not available for translation:
- "Drawing" string, used below each thumbnail in that window to identify the cell number from which it originates (Drawing 1, Drawing 2, etc.)

PR #2047, opened 2018-06-07
Fix Tool Shortcuts to Take Priority Over Frame Skip on Pressing Shift+Arrow Key 

### LINKED ISSUE #1998: Selection with Arrow keys+SHIFT  vs  Arrow Xsheet
I wanted to use the combination SHIFT + arrow to workaround this feature request here.

The offset after copy paste is 10 as @shun-iwasawa said, so it's quick to fix using shift arrow to move 10 pixels per direction instead of 1 pixel, 10 times per direction.

The problem is that the selection is lost also when using the combination shift + arrow up or down. It moves on the xsheet instead of the selection. Is there a way to be able to do this by making both work together ?

shun-iwasawa:
This will fix #1998 

Pressing Shift + Arrow key on the viewer triggers the frame skipping feature.
For now if such key combination is accepted by the current tool (such as when moving the selection with Selection Tool), such tool-specific action will take priority over the frame skipping.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -1178,68 +1178,65 @@ bool changeFrameSkippingHolds(QKeyEvent *e) {
 
 void SceneViewer::keyPressEvent(QKeyEvent *event) {
   if (m_freezedStatus != NO_FREEZED) return;
-
   int key = event->key();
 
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

openanim:
@shun-iwasawa 
 Thanks ! It works perfectly.

PR #2051, opened 2018-06-09
Pressure Sensitivy Loss Fix

### LINKED ISSUE #2017: Line Thickness Becomes Thicker When Drawing in the Bottom Part of the Viewer
### --Issue Summary--

Brush thickness becomes inconsistent in the bottom of the viewer.
Reported in [the comment in #2005](https://github.com/opentoonz/opentoonz/pull/2005#issuecomment-392387812) by @openanim , also confirmed with my MobileStudioPro13.
This problem happens with all three types of level.

### --Video or Image Reference--

GIF posted by @openanim  

### --Steps to reproduce--

- Select brush tool. Create and select any (Toonz Raster / Toonz Vector / Raster) type of level.
- Draw on the viewer with stylus.
- Compare the line drawn in the bottom of the viewer and in other part.

### --Actual Results--

When drawing in the bottom of the viewer, line becomes thicker as if it fails to recognize pen pressure.

### --Expected Results--

It should be drawn just like lines drawn in other part of the viewer. 

### --System Information--
+ OpenToonz Version:

latest mater (2018-5-28)

+ Operating System:

Win10
* Graphics Tablet:

MobileStudio Pro 13 (DTH-W1320)

---

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58910594-line-thickness-becomes-thicker-when-drawing-in-the-bottom-part-of-the-viewer?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

manongjohn:
This PR fixes a random pressure sensitivity loss I experience a lot since switching to a non-wacom tablet.  It should be noted, I do run into this issue on my older wacom tablet as we..

This was caused by the order of mouse/tablet events when drawing with a pen.  The current logic worked fine if the TabletPress event occurred before the MousePress event.  The MousePress event is designed to be ignored if the TabletPress happened first.

However, when a MousePress event occurs before the TabletPress event, usually randomly and for reasons unknown, the TabletPress event fails to overwrite the settings set up by the MousePress event and therefore fails to send pressure data to the brush tool.

I also ran into this issue when a TabletPress was somehow immediately followed by a MouseRelease event which resulted in the same issue.  For this case, I made it so the MouseRelease event is ignored if the tabletRelease event hasn't been called first.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -689,7 +689,11 @@ void SceneViewer::onPress(const TMouseEvent &event) {
 void SceneViewer::mouseReleaseEvent(QMouseEvent *event) {
   // if this is called just after tabletEvent, skip the execution
   if (m_tabletEvent) {
-    m_tabletEvent = false;
+    // mouseRelease should not clear flag if we are starting or in middle of
+    // stroke
+    // initiated by tableEvent. All other cases, it's ok to clear flag
+    if (m_tabletState == Released || m_tabletState == None)
+      m_tabletEvent = false;
     return;
   }
   // for touchscreens but not touchpads...
@@ -771,11 +775,16 @@ void SceneViewer::onRelease(const TMouseEvent &event) {
 
 quit:
   m_mouseButton = Qt::NoButton;
-  m_tabletState = None;
-  m_tabletMove  = false;
-  m_pressure    = 0;
+  // If m_tabletState is "Touched", we've been called by tabletPress event.
+  // Don't clear it out table state so the tablePress event will process
+  // correctly.
+  if (m_tabletState != Touched) m_tabletState = None;
+  m_tabletMove                                = false;
+  m_pressure                                  = 0;
   // Leave m_tabletEvent as-is in order to check whether the onRelease is called
   // from tabletEvent or not in mouseReleaseEvent.
+  if (m_tabletState == Released)  // only clear if tabletRelease event
+    m_tabletEvent = false;
 }
 
 //-----------------------------------------------------------------------------


openanim:
@manongjohn 
Thank you for attacking that too, unfortunately trying the PR, I see that the problem is still there. I do not know if anyone else can confirm.

Note a weird thing that I noticed, when I draw a line from the bottom up, the pressure is not taken into account. When I draw a stroke from up to down the pressure is taken into account.

openanim:
variation thickness_2
I tested that with this PR.

manongjohn:
Unfortunately, I can't seem to duplicate your specific issue. I had hoped this PR would fix yours too, but it seems your issue is different then.

I do find it interesting that the direction of your stroke makes a difference. Does bottom-to-top do that anywhere else on canvas or only at bottom of canvas? Does top-to-bottom at bottom to canvas draw normally or also thick?

openanim:
The problem seems specific to me at the bottom of the canvas the more we go down the more it is reproducible even outside the canvas.
When i draw a line from top to bottom even works at the bottom of the canvas. Draw from bottom to top works well if it is not at the bottom of the canvas. 

Sometimes, the pressure seemed to return to normal in a random way but I did not know how to reproduce it. But now it seems to me reproducible by drawing a line down outside the canvas.

Then, redraw at the bottom of the canvas, the pressure is taken into account regardless of the initial direction of the line.

See the end of gif to reproduce.

variation thickness_4

shun-iwasawa:
As far as I can see, the occurrence of the problem #2017 depends on not the direction of the stroke, but where you started to draw. If the stroke is started at the problematic area, `QEvent::TabletPress` is not captured and such stroke will not be recognized as being written with tablet.

Such problematic area are located bottom and right side of the viewer and its size varies depends on geometry of the viewer. For instance:

- If you put the viewer right side in the screen, the problematic area at right side gets wider.
- If you activate full screen view, the problem seems not to appear.
- If you put two viewers docked in vertical align, the whole area in the bottom viewer seems to become problematic.

@manongjohn 
Thank you for tackling with the problem! Although it will not be a resolution for #2017, I would like to review & merge this if it does fix the problem you had with xp-pen tablet. I'll check only if it works on my environment with wacom tablet as I don't have xp-pen. Is it OK for you?
Regarding #2017, I think you can reproduce the problem by setting the display scaling to higher value (such as 150%) so that the device pixel ratio becomes 2.

manongjohn:
@shun-iwasawa I had hoped it would help, but apparently not.  I still have to thank you because you did help me figure out my issue. :) For the record, I do experience this on my wacom 12WX as well but not as frequent so it's not necessarily tablet specific.


Please do review and merge.  I've updated the PR description to remove the reference so it won't close #2017.

I will change my scaling to a higher value and see if I can duplicate on both my tablets.

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed it works with no problem both on Win and OSX.
Thank you @manongjohn for the fix. LGTM

PR #2052, opened 2018-06-10
Chinese UI translation update for v1.2.1

jcome:


shun-iwasawa:
@jcome 太谢谢你了!
It seems that `image.ts` was edited but `image.qm` was not included in your commit. I can add it later on behalf of you. :+1:

PR #2080, opened 2018-06-15
Show actual fonts in font dropdowns

manongjohn:
This PR updates the Font dropdown lists to show actual fonts.

image

The following dropdowns have been converted to use the QFontComboBox class:
- Preferences (Interface tab shown above)
- Type tool

Additionally fixed a preexisting bug where the Style Combo box was not updating when a new font was selected.

### DIFF
--- a/toonz/sources/tnztools/tooloptionscontrols.cpp
+++ b/toonz/sources/tnztools/tooloptionscontrols.cpp
@@ -682,6 +682,86 @@ void ToolOptionCombo::doOnActivated(int index) {
 
 //=============================================================================
 
+ToolOptionFontCombo::ToolOptionFontCombo(TTool *tool, TEnumProperty *property,
+                                         ToolHandle *toolHandle)
+    : QFontComboBox()
+    , ToolOptionControl(tool, property->getName(), toolHandle)
+    , m_property(property) {
+  setMaximumWidth(250);
+  m_property->addListener(this);
+  setSizeAdjustPolicy(QFontComboBox::AdjustToContents);
+  connect(this, SIGNAL(activated(int)), this, SLOT(onActivated(int)));
+  // synchronize the state with the same widgets in other tool option bars
+  if (toolHandle)
+    connect(this, SIGNAL(activated(int)), toolHandle, SIGNAL(toolChanged()));
+}
+
+//-----------------------------------------------------------------------------
+
+void ToolOptionFontCombo::updateStatus() {
+  QString value = QString::fromStdWString(m_property->getValue());
+  int index     = findData(value);
+  if (index >= 0 && index != currentIndex()) setCurrentIndex(index);
+}
+
+//-----------------------------------------------------------------------------
+
+void ToolOptionFontCombo::onActivated(int index) {
+  const TEnumProperty::Range &range = m_property->getRange();
+  if (index < 0 || index >= (int)range.size()) return;
+
+  std::wstring item = range[index];
+  m_property->setValue(item);
+  notifyTool();
+}
+
+//-----------------------------------------------------------------------------
+
+void ToolOptionFontCombo::doShowPopup() {
+  if (Preferences::instance()->getDropdownShortcutsCycleOptions()) {
+    const TEnumProperty::Range &range           = m_property->getRange();
+    int theIndex                                = currentIndex() + 1;
+    if (theIndex >= (int)range.size()) theIndex = 0;
+    doOnActivated(theIndex);
+  } else {
+    if (isVisible()) showPopup();
+  }
+}
+
+//-----------------------------------------------------------------------------
+
+void ToolOptionFontCombo::doOnActivated(int index) {
+  if (m_toolHandle && m_toolHandle->getTool() != m_tool) return;
+  // active only if the belonging combo-viewer is visible
+  if (!isInVisibleViewer(this)) return;
+  bool cycleOptions =
+      Preferences::instance()->getDropdownShortcutsCycleOptions();
+  // Just move the index if the first item is not "Normal"
+  if (m_property->indexOf(L"Normal") != 0) {
+    onActivated(index);
+    setCurrentIndex(index);
+    // for updating the cursor
+    m_toolHandle->notifyToolChanged();
+    return;
+  }
+
+  // If the first item of this combo box is "Normal", enable shortcut key toggle
+  // can "back and forth" behavior.
+  if (currentIndex() == index) {
+    // estimating that the "Normal" option is located at the index 0
+    onActivated(0);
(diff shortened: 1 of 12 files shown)


RodneyBaker:
Thank you @manongjohn
Works great!

Many is the time I have had to locate a font style in OpenToonz by trial and error.
Font preview will be a great timesaver.

fontpreview

turtletooth:
This is awesome! I'm loving this.

artisteacher:
Love this! The only problem is my last font (Zapfino),  displays only the very top of the f:
954a4a60-4d3d-477e-9bf6-d5c674e7fee3

manongjohn:
Question for Mac users...

Do you know if the "Weight" dropdown in Preferences -> Interface actually does anything on a Mac? There are only 2 values in it: Regular (default) and Bold.

I know in the Type Tool, the "Style" dropdown is based on the font selected. It allows users to pick Bold, Italics or whatever is actually supported by the font. This dropdown is purposefully hidden for Mac users for some unknown reason.

The Weight preference isn't quite the same as Style since it only ever provides Regular and Bold regardless of what the font can really support. I'm wondering if it is even needed for Macs. If so, should the Type Tool have something similar? I wonder if I can add Italic and Bold Italic to it if it would even work.

artisteacher:
The weight field does work in MacOS. As far as the type tool goes, there is a style dropdown that always shows 2 options for every font - regular and bold. For some fonts, bold actually seems to apply an italics style. For others, a bold style is applied. For others, there is no change when the style is set to bold.

manongjohn:
@artisteacher Thanks for responding.

That behavior you are seeing with the Style dropdown is due to an existing bug I uncovered. The Style dropdown is stuck based on whatever font is shown at the time the toolbar option is created.  For users, it never changes when you select a different font.  Internally it does. The value from the dropdown doesn't always match internally so the Bold option may be mapping to something else.  It will be fixed with this PR.

The code suggests that the Style button shouldn't be shown for Mac users, yet you do see it.  It appears a directive is not being set at compile time.  I think I can safely remove this faulty logic that was supposed to hide it since it isn't doing that.  This means I can change the Weight dropdown in preferences to Style to match.

manongjohn:
I think its because you don't really have that font on your system? I know for some symbolic fonts like Chinese/Japanese/Arabic, if I tried to use that font, it will type normal characters for me instead of the actual font unless I'm doing something wrong.  In some cases it doesn't even show anything.

I don't think I can control what is in that font list and I'm not sure where it's pulling it from since it is a QT control.  The filter options provided for that control don't seem to allow me to filter out ones I can't render properly.

I see the same list of fonts in other application like Word and Krita (Krita uses the same exact QT font list also) and have the the same results when trying to type in that font.  I'd probably have to download the font pack for those and see if it corrects the issue.

manongjohn:
Hmm strange. Ok. Will take a look at that.

manongjohn:
@artisteacher Does the issue you noted with the mismatched font just start happening with a recent commit I did?  Does it start happening after a specific font? 

Trying to make sense if what is going on but can't tell duplicate it.  Might be something with macs?

As far as Zapfino font display issue, I need to download the font and test it out. Not sure I can do much about how that is being displayed in the dropdown list. That font seems abnormally huge to display effectively in the vertical space provided.

shun-iwasawa:
@manongjohn I confirmed that adding the following line could resolve the problem on OSX:

```
------------------ toonz/sources/common/tvrender/tfont_qt.cpp ------------------
index 1080fd1..385352e 100644
@@ -429,7 +429,8 @@ void TFontManager::getAllFamilies(vector<wstring> &families) const {
 
   QStringList::const_iterator it = qFamilies.begin();
   for (; it != qFamilies.end(); ++it) {
-    families.push_back(it->toStdWString());
+    if(!m_pimpl->m_qfontdb->isPrivateFamily(*it))
+      families.push_back(it->toStdWString());
   }
 }
```
The problem was due to the private font families which are counted in `TypeTool::loadFonts()` , but not included in the items of `ToolOptionFontCombo` .
Please refer to the doc of [QFontDatabase::isPrivateFamily()](http://doc.qt.io/qt-5/qfontdatabase.html#isPrivateFamily) for details.

manongjohn:
@shun-iwasawa Thank you for looking into the issue and finding the solution!

shun-iwasawa:
Jenkins

toonz/sources/tnztools/tooloptionscontrols.cpp
```diff
@@ -571,8 +571,12 @@ ToolOptionCombo::ToolOptionCombo(TTool *tool, TEnumProperty *property,
   setSizeAdjustPolicy(QComboBox::AdjustToContents);
   connect(this, SIGNAL(activated(int)), this, SLOT(onActivated(int)));
   // synchronize the state with the same widgets in other tool option bars
-  if (toolHandle)
+  if (toolHandle) {
     connect(this, SIGNAL(activated(int)), toolHandle, SIGNAL(toolChanged()));
+
+    connect(toolHandle, SIGNAL(toolComboBoxListChanged()), this,
+            SLOT(loadEntries()));
```
shun-iwasawa: I faced OT crash, when selecting Fill Tool and then selecting Type Tool.

I found that this connection is applied to all combo boxes for now, and once the signal `toolComboBoxList()` is emitted from Type Tool, all the combo boxes which are already constructed invoke `loadEntries()` , even if they are belonging to another tool.

Currently this connection seems to be needed only for the `TypeTool::m_typeFaceMenu` , so you should make it not to be called on other properties.

manongjohn: Fixed. I changed how and where that toolComboBoxListChanged signal is set up and processed so is can target a specific combobox on a given tool.

toonz/sources/tnztools/tooloptionscontrols.h
```diff
@@ -234,8 +234,6 @@ class ToolOptionFontCombo final : public QFontComboBox,
 public slots:
 
   void onActivated(int);
-  void doShowPopup();
-  void doOnActivated(int);
```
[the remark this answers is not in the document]
manongjohn: Sorry about that.  When I looked, I had somehow missed there was a TypeTool Font shortcut so didn't think that was needed and removed it.  I'll restore it.

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed that the shortcut keys work. Thank you @manongjohn for the improvement!
LGTM

RodneyBaker:
This addition will save us much time in finding the perfect font.
Gone are the days of slowly searching and trial and error.
Thank you @manongjohn !

PR #2084, opened 2018-06-18
Update Image DPI Field of Convert Popup Upon Opening

shun-iwasawa:
This PR fixes [a bug reported in the forum](https://groups.google.com/d/msg/opentoonz_en/vIoleLcQwoY/2NQmy6JnBgAJ) regarding the Convert Popup:

> It looks like in the RC that the Image DPI is only calculated upon changing into Image DPI mode - not upon opening the convert window which I think is the expected & desired behavior.


shun-iwasawa:
Jenkins

PR #2085, opened 2018-06-18
Fix Undo / Redo of Dragging Multiple Columns

### LINKED ISSUE #2083: Undo / Redo Failure of Dragging Multiple Columns
### Issue Summary

Undo / redo of dragging&moving multiple columns fails to work or even causes crash.

columndragcrash

### Steps to Reproduce

- Prepare a scene with multiple columns

**BUG#1**

- Set to Timeline view
- Select all columns
- Click the first column and drag up. It will make no change to the scene, but it's not a bug.
- Undo
- Redo causes crash

**BUG#2**

- Set any of Xsheet / Timeline view
- Select all columns
- Click the second or the third column and drag to the left (for xsheet view, drag down for timeline view). Again, it will make no change to the scene but it's not a bug.
- Undo unexpectedly inserts empty column

### Expected Results

In both cases, any operation should not be stacked in the History since it does not change the scene. 

### Actual Results

Some operation is stacked in the History and causes unwanted behaviors as written above. 

### System Information
* **OpenToonz Version**: current master
* **Operating System**: Win7

shun-iwasawa:
This fixes #2083 
Now dragging & moving columns will not stack an operation to the history if nothing is changed in the scene after the operation.

### DIFF
--- a/toonz/sources/toonz/xsheetdragtool.cpp
+++ b/toonz/sources/toonz/xsheetdragtool.cpp
@@ -1632,6 +1632,9 @@ public:
       col    = currEnd;
     int dCol = col - (m_lastCol - m_offset);
 
+    // ignore if the cursor moves in the drag-starting column
+    if (dCol == 0) return;
+
     int newBegin = *indices.begin() + dCol;
     int newEnd   = *indices.rbegin() + dCol;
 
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2087, opened 2018-06-18
Disable or Hide unusable tools

### LINKED ISSUE #343: Feature Request - context sensitive toolbar.
If you are working a level all the tools are always available. It would be better if the tools that are not usable are not shown or are greyed out depending on the level that is selected.

context

manongjohn:
This PR resolves #343 by introducing a new preference setting that hides tools in the toolbar if they do not work on the current active level.

You can go into Preferences to uncheck the "Show Disabled Tools In Toolbar" setting (on by default) if you want to hide unusable tools.

This can help narrow the toolset for new users so they do not try to use tools that are not meant for certain levels.

Based on my testing of each tool on the various levels, I used the following grid to determine which tools to hide in each level.

Tools for each   level | Raster | Sub-xsheet | Toonz Vector | Toonz Raster | Mesh | ZeraryFX | Sound | Note | Palette | Empty | Uneditable (i.e. PSD)
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
Animate Tool |   |   |   |   |   |   | Hide | Hide |   |   |  
Selection Tool |   | Hide |   |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Brush Tool |   | Hide |   |   | Hide | Hide | Hide | Hide | Hide |   | Hide
Geometric Tool |   | Hide |   |   | Hide | Hide | Hide | Hide | Hide |   | Hide
Type Tool | Hide | Hide |   |   | Hide | Hide | Hide | Hide | Hide |   | Hide
Fill Tool | Hide | Hide |   |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Paint Brush Tool | Hide | Hide | Hide |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Eraser Tool |   | Hide |   |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Tape Tool | Hide | Hide |   |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Finger Tool | Hide | Hide | Hide |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Style Picker Tool | Hide | Hide |   |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide
RGB Picker Tool |   | Hide |   |   | Hide | Hide | Hide | Hide | Hide | Hide |  
Ruler Tool |   |   |   |   |   |   |   |   |   |   |  
Control Point Editor Tool | Hide | Hide |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Pinch Tool | Hide | Hide |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Pump Tool | Hide | Hide |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Magnet Tool | Hide | Hide |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Bender Tool | Hide | Hide |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Iron Tool | Hide | Hide |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Cutter Tool | Hide | Hide |   | Hide | Hide | Hide | Hide | Hide | Hide | Hide | Hide
Skeleton Tool |   |   |   |   | Hide |   | Hide | Hide |   | Hide |  
Hook Tool |   | Hide |   |   | Hide | Hide | Hide | Hide | Hide | Hide |  
Tracker Tool |   | Hide |   |   | Hide | Hide | Hide | Hide | Hide | Hide |  
Plastic Tool |   |   |   |   |   | Hide | Hide | Hide | Hide | Hide | Hide
Zoom Tool |   |   |   |   |   |   |   |   |   |   |  
Rotate Tool |   |   |   |   |   |   |   |   |   |   |  
Hand Tool |   |   |   |   |   |   |   |   |   |   |  

Please let me know if there is anything that needs to change.

NOTE: When hiding tools, the displayed tool list will be based on the level you are in.
In the case of an empty cell, the tool set displayed will be based on the primary level in the column.
If the column is empty, only appropriate tools will be displayed.

Included in this is a fix to Type and Style Picker tools to show unavailable icon and message on regular Raster levels as they are not meant to be used on standard raster levels.


artisteacher:
I'm so excited to see this! This will be super helpful for new users & anyone with small screens. In a related feature request #343 there was some debate about whether it was better to completely remove the unusable menu items vs. having them greyed out. I guess completely removing them could be unhelpful for muscle memory? Personally, I'm on board.

Should the type tool work with raster levels? I can't get it to do anything in MacOS.

RodneyBaker:
I am anxious to see how this works as it seems a good direction to go.
I do have some concerns for some exceptions that I'm curious to see how it works in your implementation.

I'll wait for the Appveyor release to build first and give it a test before suggesting any specifics because you may have already addressed those concerns.

You are implementing very useful things here.  Keep up the great work!

manongjohn:
@artisteacher Ah yeah.  I am starting to recall that discussion now.  Personally, I'd rather not see the icons that don't make sense for a level.  It still potentially adds to the confusion if it remains.  I guess I can understand why some might prefer it grey'd out from a muscle memory standpoint.

Guess we'll see where this discussion goes.

I suppose, if this is a sticky subject, yet another Preference setting can be added to control that behavior. I'd like to avoid that if possible.

@RodneyBaker I know when I was testing out each of the tools on the various levels, I would find a particular tool option that would not work unless you selected another option in that tool.  In these cases, I let the button remain visible. I'd be curious to see what you find.  Definitely don't want to hide something if it can still be used somehow.

artisteacher:
I totally agree with you - I definitely prefer having the tools hidden if they don’t work for the level type.  It’ll be more understandable for my students & helpful for my small screen at school. If this does end up being a major concern for others, would it be easier to have a preference to “show disabled tools” as they are now (no greyed out state) - with hidden as the default state?

beeheemooth:
Will the tool icons stay in their places in the Toolbar, or will they move upwards, taking up empty spaces of hidden tools?

openanim:
Although I am a great adept of optimization, and surely this feature for some goes in this direction, I am not very favorable. I think in the long run it will hurt creativity. If we use only a certain level of level, we will get used to it. We may end up forgetting that the tool actually exists ...

This may be ok I imagine to some, but personally, I prefer the method that reminds you with a message that the tool does not work on this type of level or gray icons in the strictest sense. At least I know he's here. In the past, I have often tended to play only with vector levels on toonz, to the point of forgetting at times some possible features of other levels. Even if you do not like it, I find it quite impacting for it to be a preference.

beeheemooth:
I support @openanim about gray icons for inactive tools. Without icons will be too empty. Beginners can randomly select, for exemple, the audio level and search for tools for a long time. Let users know that there are tools, but they are not active. Also I think that moving icons from their usual places can confuse those users who are used to switching tools with the cursor.

artisteacher:
There’s also PSD’s - the animate tool, rgb picker tool, ruler, skeleton tool, zoom, rotate & hand tools are the only ones that work.

Like the type tool, the style picker doesn’t work on raster levels - at least on MacOS.

gab3d:
To be honest I have mixed feelings about this one.
- For one, I'm not a fan of UIs that dynamically change the visibility of options, cos I tend to automatically go to certain places to find what I need rapidly and that would make those places change all the time.
- On the other hand, I can understand the benefit from the screen real estate savings point of view. It's no doubt that the toolbar is too long to suit screens under 1080 pixels, so those users could find a benefit from this PR.

In all cases, I like the idea of disabling (i.e. greyout) invalid tools for the selected level type, so perhaps there could be a default state and an alternate one, one that just disables the tools and another one that hides them altogether.

manongjohn:
I'm putting in a preference to show disable tools.  This will be enabled by default to keep in line with what it is now.  The only difference is the tool buttons will actually be disabled and therefore grey'd out. This also means hovering over them will not highlight, further suggesting it is disabled.

If a user wants to hide disabled tools, they will need to go to Preference - Tools settings to uncheck the box.

artisteacher:
This works well! Thanks so much for this - it's definitely going to make my new classes in the Fall start out smoother! 

The greyed out state is obvious on the light theme but could be more noticeable on the default/dark themes - at least on a high dpi screen. 

Right now the first empty cell at the end of any given level displays the same tools as the level itself. Any additional empty cells after displays the empty tool set. That behavior is fine if autocreation is enabled. However if autocreation is set to "Use Xsheet as Animation Sheet", all of the empty cells are a potential frame in the level. I think in this case they should continue to show all of the tools available for the level type.

Also, just a reminder 🙂
Raster level - type tool and style picker should also be disabled 
Imported PSDs are locked from editing - all tools should be disabled except for animate, rgb picker, ruler, skeleton, zoom, rotate & hand tools

manongjohn:
@artisteacher Thanks for the feedback and suggestions will look into them.

Not sure I have control over how it greys out the image it is displaying in the button. The widget is doing that for me when I mark it as a disabled. I think the problem is that the buttons are relatively color free so there isn't a significant difference between the 2 states. The only thing I notice is the outline on the image disappears when disabled.

openanim:
After a little test the Pr, here is my feeling. Personally, I think it is also necessary that the basic mode is available in the preferences since it is the logical behavior available in most software. And it is also informative about the disabled mode even if we understand quickly that it is disabled because not available. I do not know how obvious this can be for a new user.

I notice that my shortcuts do not work in hidden or disabled states. You have to create a level first, which makes the autocreation function obselete. When you create 3 different levels, it is difficult to switch between them since in one case the shortcut is not active. You must wait to be on the desired level to activate your shortcut.

I hope my comments that may seem negative do not affect the great work you do, you are on all fronts of problems, thank you.

artisteacher:
I think it's really standard to have tools disabled when they are not usable.
@openanim what shortcuts are not working? 

Autocreation definitely has an unexpected impact. I would think the current behavior of this PR should be fine when autocreation is disabled. I've been thinking that empty cells should probably display the tools available for the default level type (when autocreation is enabled). It would make sense as each empty cell is potentially a new level.

What about this?

Autocreation = enabled: empty cells display tool set for default level type
Autocreation = use Xsheet as Animation sheet: 
- empty cells in a column/layer with no level display tools for the default level type
- empty cells in a column/layer with a level display tools for that level type

Autocreation = disabled: maintain current empty cell tool set

manongjohn:
@openanim Thanks for testing it out and uncovering the shortcut problem.  Constructive feedback is always welcome.

If the shortcut problem is going to be an issue,, which I suspect it will be with current users, the hard disabling of the buttons doesn't need to be done.  Just won't be able to glance at the toolbar and know which tools will work. The preference setting I added will then be the only way effectively show this by hiding unusable tools.

If we don't "hard disable", OT does a "soft disable" already.  If the tool was not meant to work when not in a specific level type, it will not allow you to use it.  It usually changes the icon and tells you why when you try. (Side note: I will fix Type and Style Picker tools to do this on Raster levels with this PR) 

With the soft disable in mind, the table I provided in the PR body is basically what OT is telling me is unusable under certain conditions and therefore a candidate to hide/disable.  Changing to show tools in an empty cell that is not meant to be used without a level will still not work.  It will still remain unusable unless I change that which is not within scope of this PR.
  
As far as empty cells go, autocreation is relevant only when you start drawing as far as I know.  It does not make sense with other tools which is why they are considered disabled.  The Brush, Geometry and Type tools are all active for empty cells for this reason.  Once the level is created, the toolbar should automatically update to the relevant tool set.

openanim:
@artisteacher , for exemple when I try to drawin an empty cell just after open OT I can not, because the tool is unavailable and my shortcut does not work.

RodneyBaker:
I believe @artisteacher has already identified the issue with having 'use xsheet as animaiton sheet' and where it turns off the brush tool so that initially the user can't draw directly onto the canvas via xsheet.  I would think you would be able to get that resolved by defaulting to having all Tools on unless specifically turned off.  If you are turning all Tools off by default unless specifically turned on... perhaps it should be the other way around because then when in doubt any unaccounted for process will err on the side of displaying all the tools.  This would seem to be what is needed when there isn't any level created yet... all tools should be visible whereas right now only 5 tools are.

artisteacher:
If it’s possible, it would be great to keep the hard disable when showing the tools - especially if that’s going to be the default behavior. 

It looks like the shortcut issue might be from the “hide certain tools” button. In the RC, when hide certain tools is enabled, the shortcuts still work for the hidden tools. In this PR, the shortcuts no longer work for the tools that are hidden through that button. Is that fixable? I wonder if that button is even needed anymore. Anyone who wants a smaller visible toolset can hide unused tools. 

Btw, I don’t think the empty cell/autocreation is a big deal as long as autocreation functions. It can just be a little jarring having the tool set keep contracting & expanding while cycling through a column that has periodic groups of empty frames. To see what I mean, turn “Use Xsheet as Animation Sheet” on & “enable auto-stretch frame” off. Then create a few keyframes with at least 2 empty cells between each one. Scrub or play the animation with the method of your choice. I still think hiding the tools is worth it though.

artisteacher:
@RodneyBaker’s suggestion of having all tools on by default until a level is active where some should be off is actually what I originally expected to see here.

manongjohn:
I might have messed up the logic a little with the expander button. It shouldn't disable if its a valid tool for the level regardless if expanded or not. Will review.

So what I am hearing is on an empty cell you don't really care that it wont work (it is soft disabled by the tool itself) but that the tool set remains visible based on the primary level of the column. This is not a slam. Just making sure I'm on the same page.  😊

RodneyBaker:
> So what I am hearing is on an empty cell you don't really care that it wont work (it is soft disabled by the tool itself) but that the tool set remains visible based on the primary level of the column. This is not a slam. Just making sure I'm on the same page. 

I'm not sure I follow all of that (I'm trying to catch up though!).   I need to read back about the hard and soft disables as I'm not entirely following that.

My understanding is that brushes *must* work even on empty cells in an exsheet in order for the 'use xsheet as animation sheet' to work.  I could actually live with the icon not being there on the toolbar if the shortcuts still worked but I'm sure others would suggest the icons should be there as well.

For educational purposes I'd say it should be good for all tool icons to be presented (and available for use if possible) because that gives the new users a view of all of the various tools that will be available.  As there does not appear to be any combination in your grid which has all Tools turned on at the same time perhaps an empty level would be the one where all tools appear (pending the creating of a level).

There is another alternative but I'm guessing some would not prefer it and that is to always launch OT with an active frame in a level (displayed on the canvas) of the default level type.  I only mention this because that approach would resolve at least one other issue currently encountered in that OT can't just be launched and the user start immediately drawing because there is no active level (on the canvas).   This surely has confused more than a few new users who might be expected to say, "Why can't I draw on the canvas."  This problem is easily overcome and very likely forgotten by users so we would need to observe brand new users that have never used OT in order to really nail that down.

artisteacher:
I’m able to draw on an empty cell without issue just after opening OT - without creating/loading a level or scene. Even with the tools hidden, the brush is currently available for empty cells. At least, that’s what I’m seeing.

Personally, yes, particularly when using Xsheet as animation sheet, I do think it can be beneficial to see the tools available for the level when empty cells are active.

RodneyBaker:
@artistteacher said:

> I’m able to draw on an empty cell without issue just after opening OT - without creating/loading a level or scene. Even with the tools hidden, the brush is currently available for empty cells. At least, that’s what I’m seeing.

I'm definitely not seeing that especially with the appveyor release of this PR.  Immediately after opening OT with a new scene (i.e. no previous levels established and a scene simply opened) the default tool is the Hand Tool and no brush icon is present nor does the shortcut key for the brush work.  So, there's definitely no way to draw as of right now (with the Appveyor artifact with this PR.  It may be that you are talking about the standard release.  In the standard release users can start drawing immediately *if* autocreation is turned on (which by default I do not believe is turned on).

manongjohn:
It must have been what I introduced in my last commit. I think I know what did it. Will definitly fix that. I'm glad you folks are able to test and find these issues. This is a huge help.

On that note, i am going to approach this PR as follow:

1) I will not hard disable any tools. It's causing too many issues. The tools will continue to determine if they are useable or not.
2) The toolset displayed will be controlled by preference setting. It will show all tools by default as it is now. Uncheck to hide unusable tools dynamically per table above.
3) When hiding tools, the displayed tool list will be based on the level you are in.
4) In the case of an empty cell, the tool set displayed will be based on the primary level in the column. 
5) If the column is empty, only appropriate tools will be displayed.
6) Will fix Type and Style Picker tools to show unavailable icon and message on regular Raster levels.

Let me know if this good or needs more refinement.

artisteacher:
Also, I just had an interesting thought. What if the “hide certain tools button” behavior could be changed to toggle between the full toolbar or the just the active tool set (with unusable tools hidden)? Ideally, the active tool set would be the default but it could quickly & easily be expanded for anyone who wanted it. That way, even if the tools aren’t disabled, new users would benefit without having a major impact on experienced users.

manongjohn:
In lieu of a preference setting, I like the idea of repurposing that button to hide/show unusable tools.

I have 2 somewhat minor issues with it though:
1) I do not understand why they chose the tools to hide that they chose. It is probably be ok to change it. Most tools that were hidden were for vector levels and would be handled anyway.
2) When the tool list is long (vector levels) but can't display all of them, that expander button disappears. On small screens this would be a problem. Not sure why the button disappears. Have to explore if I can find a way to overcome that.

artisteacher:
It kinda looks like the smaller tool set is really drawing focused.

artisteacher:
I definitely like that the toolbar doesn’t jump anymore when scrubbing/playing. However, if there are 2 levels of different types in a single column, the toolbar doesn’t update when the 2nd level becomes active. Expanding and contracting the toolbar does trigger the tools to display according to the new level. I think the behavior for a single level or multiple levels of the same type in a column is perfect. When there are multiple, different levels, is it possible for the toolbar to update when the level type changes but otherwise stay the same? Or perhaps the toolbar updates with the level changes and empty cells are based on whatever the previous level is in the column. 

I still like the idea of repurposing the show/hide tool buttons, if it’s actually feasible.

manongjohn:
@artisteacher  Thanks for testing it out.  I think I must have accidentally removed something since it should have changed if the level changed even when in the same column.  Will take a look at that, as well as the empty cell suggestion.

As for the hide/show button, I decided to leave that alone for now. I think the desired behavior of the hide button was to hide everything except the most basic set of tools. For Vector levels, only 1 tool (Paint brush) would be hidden if I changed the logic and may be a problem with existing users.

openanim:
New test, my shortcuts work well so far, I noticed anything abnormal in this mode about this for this moment.

- I had a brush issue, that was unisable on all levels, the first time but after opening and closing, the problem no longer occurred.

When I'm on an empty cell OT offers a minimal tool list, when I draw on that empty cell no matter what level it becomes the tool list remains unchanged. There is no refresh for the first, but for the next it's work. I do not know if it's intentional?

Small detail, but the order the tracker and the hook are reversed compared to the current version.

manongjohn:
@openanim Thanks for testing and for the feedback.

That unusable brush is an odd issue. Was that on a brand new scene or something? Closing and opening OT fixed it? I am not disabling tools so not sure if that is something with this PR or you've uncovered a new issue with something else. Will have to keep an eye out for it.

I'll see if I can do anything with that empty cell issue you noted.

Will reverse tracker and hook to match current.

openanim:
It was produced the very first time. But yes by closing and opening OT, I did not have this problem anymore.if it happens again I will try to reproduce.

artisteacher:
I actually had the unusable brush thing happen once, but there was no issue when I built with the latest changes. 

I’m also seeing that the tools do not update when using autocreation on an empty column. After switching out of the frame & back, the tool list does update properly. I see the same thing when copying and pasting a frame of one type below a frame of a different type. Otherwise, this is working really smoothly.

openanim:
On my side I did not have any problem either.

artisteacher:
Any chance this will be merged by Labor Day? 🙃

toonz/sources/toonz/toolbar.cpp
```diff
@@ -10,16 +10,121 @@
...
+  const char *toolName;
+  bool collapsable;
+  int displayLevels;
+  QAction *action;
+} buttonLayout[] = {
+    {T_Edit, false,
+     (ActivateLevel::Empty | ActivateLevel::Raster | ActivateLevel::Scan |
+      ActivateLevel::ToonzRaster | ActivateLevel::ToonzVector |
+      ActivateLevel::Child | ActivateLevel::Mesh | ActivateLevel::ZeraryFX |
+      ActivateLevel::Palette),
```
shun-iwasawa: You are defining here the level types in which each tool is enabled. 
Sorry for the late notice, but can `TTool::getTargetType()` be used instead of this for simplification?
Every tools are defined their target level types on construction and `TTool::getTargetType()` is actually used to enable/disable the tool. ( Please see [`TTool::updateEnabled()`](https://github.com/opentoonz/opentoonz/blob/25d3db05fd57aed51d666455ac8dbc44067a4cea/toonz/sources/tnztools/tool.cpp#L775-L939) )

manongjohn: I believe the reason I had chosen not to use this method was because it required constant switching to every tool to determine it's enabled status.  I also did't see an easy way to get a tool without causing it to actually switch to the tool I was checking.

I will look to see what I need to do it this way and make necessary changes to support it.

shun-iwasawa:
- Some available tools seems to be hidden from the toolbar when you select the level strip frame.
- For now the Plastic tool always seems to be hidden regardless of the level type.

manongjohn:
@turtletooth How did you restore the buttons? Might give me some idea of what happened.

turtletooth:
I had to close and reopen OT.

manongjohn:
The Palette Tool not being available is now fixed and the toolbar now interacts with the Level Strip correctly. As such, I've removed the [WIP] label in hopes this can be reviewed and included in the upcoming release.

If this PR will not be included in the next release for whatever reason, I am prepared to submit a separate PR which will include specific changes from this that I want to include.  Those changes are to to fix PR #2138 as well as fix the Type and StylePicker icons/blocking messages on Raster levels.

RodneyBaker:
Thought I'd give this a quick test and the only thing that popped out as not working is that when I create  each of the level types in the xsheet and then attempt to draw ala "use xsheet as animation sheet" both the vector and standard raster levels let me continue to select the next cell down and create a new frame but with Toonz raster levels I get the cursor change to Not Avialable symbol and of course at that point if I click on the canvas I get a warning that the brush tool is not compatible with a Toonz level... which it should be.

RodneyBaker:
The way to confirm my observation is to create all three level types in one column.
For our purpose we will put vector on top... toonz raster in the middle and standard raster at the bottom but any order is fine as it will illustrate the problem just as well.

Put some space (empty cells) in between the levels so that we can attempt to draw new frames into them.

Make sure "Use xsheet as animation sheet" is on in preferences

Note that we can draw/animate as normal if directly under the top-most level.
as soon as we attempt to do the same with either of the two level types below we will be prevented from creating the new frame (or using the brush tool).

If we drag/drop the top level's cells down below the other two... now we will be able to draw/animate under the second level type (because now it is at the top of the column compared to the other two level types.

But we won't be able to draw/animate under the level types below (including the vector level that was previously on top.

cant draw

RodneyBaker:
It would be good to have users test this release for awhile as a portable (and via full local builds) to get a sense of whether auto-hiding/disabling Tool icons is truly desireable.  For my part I'm good either way.  
Justification for changing is that I tend to have to use the arrow a lot to uncover hidden tools especially when using the touch pad on my laptop.  I also note that in it's current state I don't have much need for the Ruler Tool and yet it's there taking up space that another icon could be using.

(Potential) downsides:
Possibly harder to code for future additions to toolset.  Programmers will have to make sure they've allowed for the logic of where their new tool should appear and where it should be hidden.

We lose the sense of anticipation of what tools are there for future use when not able to see the icons even though they are not immediately needed or useful.

Neither of these potentials in any way negate the usefulness of the dynamic menu so the most important feedback we can get from users is whether it works well for them.  As such I look forward to seeing that feedback.

RodneyBaker:
Standard raster levels gained the ability to use the Fill Tool since this PR was first put forward.
As such Fill Tool needs to be added into the list of tools to display for raster levels.  I note that it is disabled via this PR.

RodneyBaker:
The RBG Picker Tool should also be turned on for (standard) Raster Levels.
This possibly relates to raster levels gaining the Fill Tool capability.

artisteacher:
The list itself may be out of date, but the function should be ok. The Fill Tool and RGB Picker Tools are turned on for Raster Levels in the OTX release.

I just want to put out there that the disabled tool version might be a good choice for the default preference. It's the happy medium between the current behavior and hiding the tools entirely.

manongjohn:
I just checked the artifact and the Fill tool and RGB Picker tool both work when Enable or Show options are used.

RodneyBaker:
Thanks for double checking.
After initially not working (and my initial report) they started working and I thought it better to sleep on it and look again.  Very strange.

RodneyBaker:
Merging as this feature is off by default and adjustable in Preferences.
I do have concerns about the long tail of maintenance associated with dynamic menus that cannot be adjusted directly by the user but that shouldn't hold up release of the feature.
The PR itself:  LGTM

PR #2091, opened 2018-06-20
Fix Loading Toonz4.x Level Files (.TZP)

shun-iwasawa:
This PR fixes the following problem reported by Studio Ghibli:

OT is capable to load levels created with old Toonz versions (Toonz4.x) by converting TZP to TLV. But for now OT seems to fail to load & convert the level's palette properly.

**UPDATE**

I modified `TFilePath::withNoFrame()` to return proper path. The originally intended behavior is like as follows:

```
TFilePath("C:\hoge..jpg").withNoFrame();       // C:\hoge.jpg
TFilePath("C:\fuga.0001.tif").withNoFrame();   // C:\fuga.tif
TFilePath("C:\hogera_001.png").withNoFrame();  // C:\hogera.png
```

This will also fix the wrong default file name appeared in the browser popup when clicking `Save Settings` button in the Cleanup Settings. For the raster level `A..tif`, the default file name of the cleanup settings is currently `A..cln` (with double dots) . It should be `A.cln`.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

Issue #2111, opened 2018-06-28
(WIP) Command Bar missing icons. [$45]


openanim:
### Issue Summary

The bar of commands is a great tool very practical, in the different wishes expressed in #1526 for its improvements, I personally think that the most urgent it is to give him the missing icons. I open a bounty for this purpose because there are many and we end up with a very long command bar with the names of the commands instead of icons. It would be great to be able to give it icons that fit in the current coherence.

openanim:
I opened the bounty https://www.bountysource.com/issues/60247847-command-bar-missing-icons

artisteacher:
As there are a lot of commands without icons, this will probably take a few rounds. What commands should have the highest priority? If everyone can share the commands they use in the command bar that are missing icons, that will give a good direction to start with - especially if there is overlap. 

Personally, I’d like to see:
- each 2
- each 3
- each 4
- clone
- delete
- remove unused levels
- insert multiple keys
- remove multiple keys

artisteacher:
I completely agree, if there are existing icons being used for the same commands elsewhere, they should definitely be used for the command bar.

openanim:
I agree too, it seems obvious to me to stay consistent.

the commands that I use the most or I’d like to see are

- enter group
- Exit group 
- ungroup
- group
- bring to front
- bring forward
- send backward
- send back
- zero thickness
- fill check
- gap chek
- remove vector overflow
- new note level
- add frames
- palette
- style editor
- schematic
- output settings

artisteacher:
On a related note, I really think “send back” really should be “send to back” for consistency & clarity.

RodneyBaker:
@artisteacher said:

> On a related note, I really think “send back” really should be “send to back” for consistency & clarity.

I am in agreement with you here as 'Send back' and 'Send to Back' might appear to be different things and the latter would convey what is happening more accurately.

'Send back' as a terminology actually means 'Send Backward' as in move the selection one step further away from the top toward the back of the stack.  But that's not the intention of the process that will be launched when 'Send Back' is triggered.
'Send to back' more obviously means push the current selection all the way to the end of the stack.
'Send Backward' does convey pushing the current selection toward the back.  It doesn't intrinsically carry with it the sense of moving only one step but that can easily be discovered by the user.

Having the 'Send to Back' terminology mirror that of 'Bring to Front' would also be appropriate.

Good idea.

artisteacher:
[...]

Issue #1526, opened 2017-10-17
Command Bar Refinements 


artisteacher:
Just a couple thoughts on the command bar/Xsheet toolbar:
- there are a couple commands that don’t seem to work at all when attempted from the command bar (ie. Fit to window, reset view, global key)
- some commands are greyed out when they are not possible (due to active level type, empty cell selection or etc) but others are not 
- when customizing the bar, it can be hard to find the desired command - having a search option (like when setting shortcuts) would be helpful 
- dragging & dropping is sometimes glitchy in OS X (when trying to drag a specific command, often a different command actually moves)
- some commands do have icons elsewhere in the program but the icons don’t display (field guide, fit to window, set key) - though in many cases the commands with existing icons are unlikely to be added to the command bar since they are usually already visible in the UI

Thanks @turtletooth for this great addition! I love the customization possible. Also, do you want icons for other commands? Here are some possibilities for time stretch, load level, and canvas size:
![6e7772d9-8524-4547-ab79-aa8718a86c9f](https://user-images.githubusercontent.com/21159116/31641417-73ec437c-b2b2-11e7-98fe-f9fbd6b6eae3.png)

Agnyy:
I want to make my suggestions.
1. Add a search.
2. Add the ability to upload icons.
3. The ability to resize as in Krita
![untitled-1](https://user-images.githubusercontent.com/25137502/31654300-4bc5065a-b32e-11e7-9714-3ce1c88731a2.jpg)

![2017-10-17-11-20-52](https://user-images.githubusercontent.com/25137502/31654305-5077fc20-b32e-11e7-9d2c-0ef6990ff91a.gif)

turtletooth:
@artisteacher Can I get the files you have for those icons?

artisteacher:
Here you go!
[Possible Icons.zip](https://github.com/opentoonz/opentoonz/files/1513767/Possible.Icons.zip)

artisteacher:
I added another one for the upcoming enabling/disabling touch command and made the stroke of a couple other potential icons consistent with the others.
[More icons.zip](https://github.com/opentoonz/opentoonz/files/1517720/More.icons.zip)

DoomDesigner:
![image](https://user-images.githubusercontent.com/32999433/33664831-0731e400-daa6-11e7-8ffa-a24d9d743b5e.png)

I think I need to add icons (Play, Pause and others). Simply personally, I like the location of the icons from the bottom.
[...]

PR #2116, opened 2018-06-29
(WIP) Missing icons

areckx:
This is for PR #2111 

 I started to add some of the missing icons. I started with the tool modifiers, the seem to work as intended. It seems the `const char *iconName` will need to be added to the other actions in order to use the original functions. I found the additional icons in issue #1526 but I will need to create more or modify existing for the menu actions, etc.


areckx:
I added `const char *iconName`  to what (I think) is all of the functions needing icons. I am using `"broken"` for everything so I can either remove or replace with a relevant icon and keep track of where I am. Expect commits to come in over the next few days.

areckx:
it might be the first commit shun-iwasawa  mentioned that the setWindowFlag is used in QT5.9 but apparently that line can be removed entirely. I will try building without it. 

This also may be relevant. https://stackoverflow.com/questions/33743493/why-visual-studio-2015-cant-run-exe-file-ucrtbased-dll

openanim:
Hi,
Thank you for your interest and have already started a solution, however I would like to be certain that others in the community have not yet started work on it, I think especially artisteacher. This is important for the consistency of the visual identity of the icons already available. Of course it is quite possible to modify along the way. Do you have any suggestions ?
Btw, I did not get this error message.

areckx:
If others have already made icons I can add them in. It doesn't seem like there are any in progress so I'd like to head from them because if at all possible it would be better to use custom icons to match the spirit of OT

areckx:
> Of course it is quite possible to modify along the way. Do you have any suggestions ?

While the first step is to make sure all the tools and options have at least the option of having an icon, there should be an option to use either text _or_ the icon with for each entry. Right now all the icons have to be hardcoded in (it seems) and a custom icon loader would be pretty useful.

artisteacher:
I’ve been brainstorming some potential options for a few of the icons. Now that school is out, I can actually get started. I do think it’ll be important to propose the icons in the issue first & get feedback before actually adding them in so that there can be revision if it’s necessary. Also, it will take some serious time to make icons for all of the commands. It would be great to make a few lists to help prioritize. What commands are the highest priority? What commands would be helpful but are secondary? Are there commands that are low priority?

openanim:
This seems like good ideas, but for the second point how to prioritize? Should we post here and on the forum the different commands that everyone uses the most ? Then group and synthesize those that come back more ? or arbitrarily? I imagine that depending on the workflow of each it can be different.

artisteacher:
I think we should post both here & in the forum & then synthesize.

areckx:
Well what should I do? I feel like I'm not going to be able to make any income but I found bountysource and it looked like there was hope for me. I thought this project would be a good challenge for me and I would be able to get some money to pay the bills. I've worked on this for several days and it looks like it's for nothing...

Agnyy:
@areckx Hey. I have some projects for OT that I need to implement, but there is no programmer. I can offer you a part-time job. For convenience, I will give an advance payment. Here is my Email: Agniyya@gmail.com

openanim:
@areckx 

First of all, no one said that you could not have this bounty. It should not turn to drama unnecessarily. However, it must meet the expectations initially defined namely the creation of icons consistent with the current design. Do not see it, a reproach, but it would have been nice to set the expectations again, before you commit, I think.

I'm not a programmer, so I do not have any notion of the time, the difficulty, or any other parameters that are needed to put that in place. Although I suspect that it will take time, given the large number of commands, I know that the creation was not going to happen overnight. Even targeting the most commonly used.but you seem in a hurry.

Honestly, I was thinking first of all about the creation of icons with this request as suggested by @artisteacher. But, it turns out that it takes a lot of work to implement the code you did. On top of that, I would especially like the approval of the current main OT developers  (@shun-iwasawa  @manongjohn @turtletooth ) to know if this will be accepted and necessary since it is they who decide. Basically, to find out if your work is consistent/coherent since the goal is that it be incorporated into OT, later. If everything is OK then you will have the bounty.

And if necessary, I'll create another bounty for creating Icons to integrate with your work if approved because the goal is not to go through a database of icons.

areckx:
Sorry I'm just starting to feel like I'm not going to be able to earn a living. It's really nerve-wracking.

711GM:
as I understand it we need to prepare a list of icons and draw and then vote for them? If so, why have not we created this list yet and started working?

artisteacher:
@711GM, please let us know the commands you most want icons for in #2111.

RodneyBaker:
This PR appears to have been abandoned.

RodneyBaker:
There is still a $45 bounty assigned to this one.
@openanim I think that is all funded by you so the decision to close this PR is yours.  If you do decide to pull support make sure you transfer your bounty. Also, IMPORTANTLY while it may not apply in your specific case as the bounty hasn't been awarded and yet unclaimed, in those cases fees may apply after 90 days.  As such it will be good to read up on recent change to policy by BountySource.

openanim:
Well, i dont have possibility to close or not this PR (be free to do it)). As said before, this was pending review. I do not know where things are right now. I had an agreement with areckx for this PR and I was planning to create another bounty for the other complement that had started artisteacher as said. But no news since that day. I suppose an abandonment too.

RodneyBaker:
Will close pending renewed activity.

openanim:
About that ... Alex told me that he could not work on Opentoonz anymore, and he moved on to something else.

RodneyBaker:
Thanks for the update @openanim  and good luck to Alex as he moves on.

Perhaps we can examine and summarize where this effort left off so we can better identify what still needs to be done.

RodneyBaker:
Reopening on the chance this might make it in to the pending release.
Somewhat doubtful and lower priority but... hope springs eternal.

turtletooth:
Closing.  I think we need to look at proposed icons and work them in as they get support.

PR #2140, opened 2018-07-04
[Hiatus] Recent FXs list

jpturcotte:
This PR will add a new category to the FX contextual menus containing a list of ten (10) recently used FXs from the most to the least recent. This list is contained in a user-modifiable file. Currently, it loads the file if present at the start of a session, but does not create or save it. Also, the tree view when clicking on the buttons is not supported. I am sure the logic and code are wonky, especially since it makes extensive use of pointers, which I am not familiar with. There are some unused stubs for a favourite list too. I have not touched this since Easter, hence why I am opening a WIP PR in the hope of getting some reviews, and hopefully some motivation back. Thank you!

From an idea by @DarrenTAnims and @RodneyBaker.


DarrenTAnims:
If it doesn't create or save it, how does the file come into being? Is it just user created? If so, how does it modify the most used-ness?

artisteacher:
I’m assuming that the eventual goal is for OT to automatically create & maintain the list.

jpturcotte:
@DarrenTAnims It is as @artisteacher said. Unfortunately, a lot of things that used to work around Easter now cause OT to crash, which is not good. Basically, it has to be rewritten.

jpturcotte:
Yes, I do not think I will get back to it this year. Maybe you could label it? I thought the WIP in the title was enough, but I guess WIVSP would be closer to reality. ;)

RodneyBaker:
Hmmm... yes, we need a good designator for things not headed for the next release.
WIP is a good one to indicate the obvious (that it's a work in progress) but doesn't convey whether it is being targeted for pending release.  

Tagged as 'Hiatus' which has a description of "On hold / Not pending release".

RodneyBaker:
Closing.  There is no anticipated work being done on this PR.

PR #2151, opened 2018-07-09
Fix Crash When Using Vector Brush Frame Range with Thickness 0

### LINKED ISSUE #2118: Crash when using Vector Brush Frame Range with thickness 0 (gif)
That crash happens only when you use the Brush Tool with a thickness 0

1. Draw first stroke
2. Go to next frame (or more)
3. Click (mouse or stylus)

ezgif-2-879638ac1e

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/60341721-crash-when-using-vector-brush-frame-range-with-thickness-0-gif?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

shun-iwasawa:
This will fix #2118 
`TVectorImage::addStroke(TStroke *stroke, bool discardPoints = true)` discards stroke without adding when `dscardPoints` is set to true and the bounding box of `stroke` is empty, which can be happened with a point with thickness 0.

### DIFF
--- a/toonz/sources/tnztools/brushtool.cpp
+++ b/toonz/sources/tnztools/brushtool.cpp
@@ -1746,8 +1746,8 @@ bool BrushTool::doFrameRangeStrokes(TFrameId firstFrameId, TStroke *firstStroke,
     swapped = true;
   }
 
-  firstImage->addStroke(first);
-  lastImage->addStroke(last);
+  firstImage->addStroke(first, false);
+  lastImage->addStroke(last, false);
   assert(firstFrameId <= lastFrameId);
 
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2152, opened 2018-07-09
Enable Windows Ink Support

### LINKED ISSUE #2149: Pen pressure not recognized in laptop-tablet
I have a laptop-tablet (Thinkpad Yoga 260 - Win10) that has a built in digitizer and supports pen pressure. I discovered the pen pressure does not work in OT like it does in other software like Krita or MS Paint3D.

Upon further investigation, laptop-tablet digitizers like this use Windows Ink to handle pen input but QT only supports Wintab.

Krita recently included support for switching between Wintab and Windows Ink to address this issue in 3.3. It was later adopted recently by Drawpile as well.

I am recommending OT also adopt this fix.

This will also help graphics tablets that also support Windows Ink, though typically it can be disabled in the tablet to get around this issue.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/60623290-pen-pressure-not-recognized-in-laptop-tablet?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

manongjohn:
This PR resolves #2149 

A new `Tablet Settings` preference tab has been added which give the user the option to `Enable Windows Ink Support`.  Future tablet related settings can be added here.  For now, the new tab is currently only available on Windows systems that have Windows Ink.

NOTE: Logic for handling the Windows Ink support was taken from Krita open source code and adapted to OT.


shun-iwasawa:
> NOTE: Logic for handling the Windows Ink support was taken from Krita open source code and adapted to OT.

I'm concerned about that there may be a license conflict since Krita seems to be licensed under GNU GPL.

manongjohn:
Please forgive my naive-ness when it comes to licenses.  I definitely don't want to cause any issues.  If there is a way we can use it, please advise, otherwise I will probably have to just close it.  Let me know.

shun-iwasawa:
( Please note in advance, I still don't know match about OSS license compatibility... I could be wrong or overlook something. )
Unfortunately, we can't merge a PR based on GPL source as it is not license-compatible with OpenToonz under Modified BSD License.
As far as I can see, there may be several ways you can take to solve the Windows Ink issue. 

- Look for another source which is published under compatible license with OT and refer to it.
- Ask original author of the reference source to grant you the source with a license other than GPL.
- Make a module including the GPL source as separate binary and make OT to externally use it. (Just like LZO compress module used by OT)
- Drink alcohol, completely forget what you see in the GPL source and write an original source from scratch.
- [Qt seems to be tackling with Windows Ink support](https://bugreports.qt.io/browse/QTBUG-60437). So just wait for it.

Anyhow, I think it would better to close or at least state as WIP until it is cleared up. I'm sorry.

manongjohn:
Did some checking and found this on https://www.gnu.org/licenses/license-list.html

> Modified BSD license (#ModifiedBSD)
>This is the original BSD license, modified by removal of the advertising clause. It is a lax, permissive non-copyleft free software license, compatible with the GNU GPL.

Doesn't this mean it is compatible?  I'd just have to follow whatever requirement is needed to satisfy both licenses.

shun-iwasawa:
Please see [FAQ : What does it mean to say a license is “compatible with the GPL?”](https://www.gnu.org/licenses/gpl-faq.html.en#WhatDoesCompatMean)

> It means that the other license and the GNU GPL are compatible; you can combine code released under the other license with code released **under the GNU GPL** in one larger program.

So, AFAIK, if you combine the GPL code the software will be no longer Modified BSD.

manongjohn:
I saw that and disagree with how you are reading that. 

Based on how that is worded, "in one larger program" you can "**combine** code released under the other license **with** code released under the GNU GPL". There is no directionality. It doesn't say GNU GPL supersedes any other license. It is compatible because inclusion of both licenses under 1 program means the requirements for each license agreement can both be equally met.

Please see: https://en.m.wikipedia.org/wiki/Multi-licensing

jpturcotte:
I am no lawyer, but as far as I know, if you add GPL-licensed code, you will have to provide the whole source code as GPL as well.

shun-iwasawa:
Please forgive my misunderstanding in my previous post. It seems to be - as you pointed - just talking about that the sources under both licenses can be combined in one program.

Let me try to explain more precisely:

- The [Krita's source](https://github.com/KDE/krita/blob/3bea0c74384071c2887e4a7037f6ca56644241d3/libs/ui/input/wintab/kis_tablet_support_win8.cpp) you referenced is distributed under GPL.
  - As far as we know, it's not dual-licensed. So GPL is only the license we can choose.
  - According to the comment in the source `either version 2 of the License, or (at your option) any later version`, I'll refer to [GPL version 2](https://www.gnu.org/licenses/gpl-2.0.html).
- Due to the condition 2-b , your source in this PR should also be distributed under GPL.
- And once the GPL source is included in the software (no matter it is partially or entirely), the "derivative work" (i.e. OpenToonz and other softwares which will be derived from OT's source code in the future) must  distribute source code on distributing the software, according to the condition 3.

Modified BSD license is more relaxed - for example, if you make some commercial software based on OpenToonz source, you can sell it without publishing its source code. Introducing GPL source will restrict such possibility. 
Such aspect of GPL is sometimes slanged as [Viral license](https://en.wikipedia.org/wiki/Viral_license).

manongjohn:
@shun-iwasawa and @jpturcotte Thanks for the info and insight. As I read more, including the actual GPL, I could see why it could be an issue since GPL does have a lot of other conditions we'd have to adhere.  It is what it is...*sighs*

I actually don't know how this was coded and only know it somehow accesses Windows Ink APIs.  Aside from quickly removing debugging statements and replacing a few macros we didn't have to get it to compile, all I did was call the new class from the main.  I'm sure if I drank a lot of alcohol (which I will do regardless) to forget whatever I might have seen,  I could come up with my own solution.  In the interest of minimizing potential conflicts, I will not pursue it.

Just really sucks that there is an obvious fix for this but we can't use it.  Only good thing is I can still use it for my own personal use.

I can only hope QT does something soon about it for everyone else.

shun-iwasawa:
@manongjohn 
Thank you for your understanding. I feel sorry for not being able to be of assistance for including your fix.
Please let me know if you ever come to Tokyo - I'll be willing to help you to vanish your memory with Japanese sake.

turtletooth:
I reached out on the Krita IRC channel to find out who the developer was who implemented Windows Ink support in Krita and sent him an email inquiring if he would be willing to share his code under a different license.

manongjohn:
@shun-iwasawa I had actually reached out to the developer of the code I had wanted to adopt before @turtletooth mentioned he sent out a request via the Krita IRC channel.  We got the following e-mail back from the developer:

> Dear all,
> 
> I have been told by multiple people that OpenToonz developers would
> like to make use of the Windows Pointer Input Message (Windows Ink on
> Windows 8 or above) support code files
> (libs/ui/input/wintab/kis_tablet_support_win8.h and
> libs/ui/input/wintab/kis_tablet_support_win8.cpp) that I wrote for
> Krita, but couldn't due to Krita being GPLv2+ and OpenToonz being
> licensed in BSD 3-Clause, Currently Drawpile has already adopted these
> code with no issues since Drawpile is GPLv3.
> 
> I would be fine with allowing OpenToonz to use the current version of
> the code in BSD 3-Clause licence, but this would seem short-sighted
> since future changes of the code in Krita wouldn't make it into
> OpenToonz without further permissions, and it would get more
> complicated if other developers make changes.
> 
> Therefore, I am considering to re-license these two files within the
> Krita repo as BSD 2-Clause (BSD 3-Clause is not allowed according to
> KDE policies), and request that in the future other developers who
> make changes to these two files within Krita should agree to license
> their changes as BSD 2-Clause. A problem with this approach is that if
> Drawpile, OpenToonz or other derivatives make other functional changes
> under a more restrictive licence (e.g. GPL or even BSD 3-Clause), they
> can't be backported to Krita without explicit permission. It would
> also mean that even closed-source software can use code from these two
> files, though I don't have problems with this. Drawpile and others can
> still continue to use the current or older versions of the files (and
> derivatives of it) as GPLv2+, though it would get a bit complicated if
> they wish to include any future changes.
> 
> The process should be as simple as me making a commit to change the
> license comment blocks in these two files.
> 
> Are there any technical issues with this approach, or did I show any
> misunderstanding in the previous paragraphs?
> 
> I would like to have this done in no more than a week, so if there are
> any concerns please raise them as soon as possible.
> 
> Best Regards,
> Alvin Wong
> 
> CC: OpenToonz develoeprs John Dancel and Jeremy Bullock and Drawpile
> developer Calle Laakkonen

Is this acceptable?

shun-iwasawa:
Thank you @manongjohn and @turtletooth for making contact with the author. Also I'm grateful for Mr. Alvin Wong's kind proposal for re-licensing.
Yes, it can be included if the reference source is re-licensed to BSD 2-Clause.

Since I'm off work until next Tuesday, I'll post later in detail.

jpturcotte:
@manongjohn & @turtletooth You guys rock!

shun-iwasawa:
@jpturcotte 
Yes, I agree that it would better to redistribute this source keeping the license unchanged from BSD 2-Clause in order to enable backporting in the future.

@manongjohn 
Now we clarified the copyright and license statement.
Once the license of original source will be changed, please transcribe its copyright notice and the license body at the very beginning of your code, modifying the annotation like as follows:

```
This file is based on the Windows Pointer Input Message support code files by Alvin Wong. 
Notwithstanding the license specified in this repository, this file is redistributed under 
BSD 2-Clause license written below in order to keep backporting available.

All contributions by Alvin Wong:
Copyright (c) 2017 Alvin Wong <alvinhochun@gmail.com>
All rights reserved.

All other contributions:
Copyright (c) 2018, the respective contributors.
All rights reserved.
 
Each contributor holds copyright over their respective contributions.
The project versioning (Git) records all such contribution source information.

(followed by BSD 2-Clause license body)
- - - 
```

Your copyright (regarding your modification done in this code) will be covered in the second copyright notice. In case someone would like to backport or redistribute this code in the future, the name of copyright holders can be retrieved from Git.

Also, please add the license file to `stuff/doc/LICENSE/` with the original copyright notice and the license body in order to meet the 2nd clause of the license of the original code.

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -1211,6 +1219,14 @@ PreferencesPopup::PreferencesPopup()
     , m_inksOnly(0)
     , m_blanksCount(0)
     , m_blankColor(0) {
+  bool showTabletSettings = false;
+
+#ifdef _WIN32
+  showTabletSettings = KisTabletSupportWin8::isAvailable();
+#else
+  showTabletSettings = true;
```
shun-iwasawa: Should it be `showTabletSettings = false` for non-windows environment?

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -1538,6 +1554,16 @@ PreferencesPopup::PreferencesPopup()
       new QLabel(tr("* Changes will take effect the next time you run Toonz"));
   note_version->setStyleSheet("font-size: 10px; font: italic;");
 
+  //--- Tablet Settings ------------------------------
+  if (showTabletSettings) categoryList->addItem(tr("Tablet Settings"));
+
+  m_enableWinInk =
+      new DVGui::CheckBox(tr("Enable Windows Ink Support* (EXPERIMENTAL)"));
```
shun-iwasawa: For now you construct the checkbox `m_enableWinInk` regardless of the `showTabletSettings` value.
If the `showTabletSettings` is false, the checkbox will be created without parent widget and thus will not be deleted along with a destruction of the `PreferencesPopup`. 

How about constructing `m_enableWinInk` only if the `showTabletSettings` is true?

stuff/doc/LICENSE/LICENSE_win8.txt
```diff
@@ -0,0 +1,23 @@
+Windows Pointer Input Message support code files from Krita
```
shun-iwasawa: It may be a a matter of preference, but can you change the name of the license file from `LICENSE_win8.txt` to something more illustrative one?
Well, let's say, how about `LICENSE_kistabletsupportwin8.txt` or `LICENSE_winink.txt`?

shun-iwasawa:
Jenkins

shun-iwasawa:
I confirmed that this option also resolves #2017 . Great!
Thank you @manongjohn and Mr. Alvin Wong for this option. LGTM

bobbybutterfly:
@manongjohn thanks for this
but on my Surface Pro 2017 with Surface pen (or Wacom Bamboo ink pen) it has the following flaw:

with Windows ink enabled, the pen now draws like it should, without moving the canvas at the same time, but:

switching from the brush tool to the Hand tool (or Animate tool) and back to the brush tool the pen moves the canvas around while drawing. This behavior stops, when i move the pen away from the screen for lets say 20centimeters. Moving pen back to the screen now draws the brush without moving the canvas.

PR #2153, opened 2018-07-10
fix license link

pocketberserker:


shun-iwasawa:
Thank you @pocketberserker for the fix! LGTM

PR #2173, opened 2018-07-20
Fix UNC Path Conversion

shun-iwasawa:
While checking the Toonz Farm feature, I happened to find the problem that the scene cannot be rendered with tcomposer if the scene file is specified with UNC path.

This PR fixes such problem, which is due to missing of the slash between the local path to the shared directory and the latter half of the path.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2175, opened 2018-07-20
Schematic navigation enhancement

manongjohn:
This PR improves the navigation of the Schematic viewer as follows:

1) Adds 3 new buttons: Selection (default/normal), Zoom and, Hand (pan)

image

2) Handles touch gesture

   1 finger = Pan
   2 finger pinch = Zoom
   1 finger double tap = fit to window


artisteacher:
This is great! While there are a lot of shortcuts for navigating the schematic, it will be really helpful to have visible navigation options. I spent an embarrassingly long time navigating just through zooming in & out because I was unaware of the shortcuts myself. I’m most excited by the touch controls - I always try to use touch gestures in the schematic after using them in the viewer.

openanim:
it's something that happens to me too, I'm excited about it too, I can not wait to use the new schematic in a nighly build

manongjohn:
@shun-iwasawa For some reason the Appveyor build isn't kicking off.  It's not showing in the list of checks which I submit/update a PR.

artisteacher:
Just tried out the touch gestures on MacOS + touchpad. There are a few issues with the touchpad behaviors, but once they are fixed this will be a really nice addition.

Right now the default middle-mouse-wheel scrolling is interfering with the touchpad 2 finger panning behavior. The middle mouse wheel scrolling needs to be turned off for touchpad devices when touch gesture is active (like it is for the Viewer). Schematicviewer.cpp lines 324-331 should be adjusted to something like this:
screen shot 2018-07-20 at 2 39 56 pm

The 1 finger double tap to reset the view is not working - right now it's set to work only for touchscreen devices. You should be able to completely remove "&& m_touchDevice == QTouchDevice::TouchScreen" from line 305 of schematicviewer.cpp:
screen shot 2018-07-20 at 3 50 01 pm

The 2 finger pinch zooming behavior is significantly more jerky than it is in the viewer - it seems to be panning at the same time. Adding "m_panning = false" to m_zooming in schematicviewer.cpp, seems to help a lot:
screen shot 2018-07-20 at 3 48 44 pm

Sometimes the touch zooming behavior seems to get locked to the diagonal instead of based on the cursor location. I've run into that before when using the shortcuts to zoom in/out so the glitch is probably related to zooming in the schematic, not the touch controls.

On a related note, the buttons work great. Ideally, I do think the cursor icon should probably change to the hand when panning and to the magnifying glass when zooming (when these new navigation buttons are active).

manongjohn:
@artisteacher Thanks for the awesome feedback! 

Will fix the 3 issues you found.  I'm not sure I understand what you mean it gets "locked to the diagonal".

I tried to get the schematic to use the cursor icons that are used in the scene viewer. Have to figure out how to link the cursor manager code to this code so it can utilize it.

openanim:
@manongjohn 

BTW, normally to insert a node into a link we can do Alt click and drag it into the link, and to extract a node from a link, Alt click and drag it away from the link. But the behavior no longer works, to put the link correclty. However, as long as we have not permanently dropped the node, the behavior still seems to be there. Is it possible to bring it back ? it looks like a bug.

openanim:
Small question, some panels share the same way of navigation as the schematic, like the color model, the flipbook. Do these enhancement  also concern them or the link is not the same ?

manongjohn:
@openanim This change is specifically for Schematic only.

openanim:
@manongjohn 

In the last version of the last NB, I have a display issue about the color of the selected column (that of the preferences).

Now, the color is also linked with the shematic. But when you modify it, the effect is immediate in the shematic view but not in the xsheet / timeline. It takes a while for it to take effect.

manongjohn:
@openanim I'm not experiencing that issue.  Is it just that it doesn't update the xsheet/timeline until you focus on that pane?  If you are on the xsheet/timeline and change the selected column color, does it change the schematic right away or not until you focus on that pane?

openanim:
I have not been able to reproduce it these last times but that happened the first time I launched, the last NB.If I notice it again, I'll let you know.

artisteacher:
The updated cursors are a nice touch. 

As you can see in the following gif, zoom is still really jumpy/floaty with a touchpad.
zoom

Also there's an unexpected problem with the double tap to reset the view - it interferes with opening the camera or fx settings. Once you are able to open the settings (after excessive double tapping/clicking), the double tap behavior to reset zoom no longer works in either the schematic or viewer. At that point the 2 finger panning reverts to the mouse wheel zooming behavior in both the schematic and viewer. 2 finger pinch to zoom is the only touch behavior to work correctly but only in the viewer. The 2 finger zoom in the schematic partially works but gets completely locked along the diagonal. The only thing that helps once this starts is restarting OT.

manongjohn:
@artisteacher 

> As you can see in the following gif, zoom is still really jumpy/floaty with a touchpad.

The touchpad zooming can be done 2 ways, 2-fingers staying together and moving or 2-finger pinch.  I notice it is only jumpy when I use 2-finger pinch but smooth the other way.  Is that what you are seeing?  It's the same for the viewer for me.  The one that is jumpy is likely using the mouse wheel version of zooming. Its jumpy because each mouse wheel click generates 1 event and a larger change in value vs the other method which generates several gesture events with smaller changes making the zoom smoother.  i will verify. If this is the case, I won't be able to do much about it.

I still don't get panning either way in either viewer.  This only happens in schematic for you? If so, I'll try and see what viewer is doing differently. I patterned schematic's logic after it but had to make a few adjustments.

> Also there's an unexpected problem with the double tap to reset the view - it interferes with opening the camera or fx settings. 

A mouse/pen double tap would still work as normal on a node. I'll see if I can fix this.  Where the touch/gesture code is at I have to see if I can easily tell if I'm over some object or not otherwise it might be a negative feature of touch gestures or i remove this double tap feature.

> Once you are able to open the settings (after excessive double tapping/clicking), the double tap behavior to reset zoom no longer works in either the schematic or viewer. At that point the 2 finger panning reverts to the mouse wheel zooming behavior in both the schematic and viewer. 2 finger pinch to zoom is the only touch behavior to work correctly but only in the viewer. The 2 finger zoom in the schematic partially works but gets completely locked along the diagonal. The only thing that helps once this starts is restarting OT.

Not sure what is going on here. I can't reproduce.

artisteacher:
I’ve been using the same pinch gesture in both the viewer and schematic to zoom. It is very occasionally a little jumpy in the viewer but usually quite smooth. Again very occasionally I might trigger a panning motion when trying to zoom in the viewer if my fingers slide a bit when pinching but it is very irregular. 

I’m not sure what you mean by the zoom gesture with 2 fingers together, unless you mean the touchpad scrolling behavior which zooms by default when gestures are turned off.  That is currently significantly smoother for zooming in the schematic - however the pinch gesture zooming is usually equally smooth in the viewer. 

It’s like OT stops recognizing certain touch gestures and overrides them with mouse behaviors. Maybe it’s caused by an input overload? I’ve occasionally had this happen when troubleshooting touch gestures for touchpad in the viewer. If the double tap to reset view is fixable, that’s fantastic. Otherwise removing the gesture does seem to prevent the strange behaviors.

jpturcotte:
@openanim Follow these steps: https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_test_prs.md.

manongjohn:
The Appveyor build hasn't been triggering for some odd reason so there isn't one currently for this PR right now.

artisteacher:
@openanim - since I only have access to Macs, I can't use appveyor. I've been building myself following the build instructions which can be accessed [here](https://github.com/opentoonz/opentoonz/blob/master/README.md) if you click on the link for building locally on your operating system. 

@manongjohn - that does help with the jumpy part of the zoom but the panning while zooming is still a big problem. Adding the following to line 668 (below `m_zooming = true`) in SchematicViewer.cpp does stop the panning once zoom is triggered:
`m_panning = false`
 
Interestingly, if you change line 626 from
`if (m_scaleFactor > .2 || m_scaleFactor < -.2) {`
to
`if ((m_scaleFactor > .2 || m_scaleFactor < -.2) && !m_panning) {`
the pinch behavior causes panning instead of zooming. 

I am also getting a lot of the unexpected diagonal zooming behavior I've mentioned before - especially when the cursor is over parts of the background. Maybe the problem is to do with how the center is calculated or something? That does seem to be a difference from the code for the viewer.

manongjohn:
@artisteacher I just reimplemented the sceneviewer's zoom/pan touch and gesture logic. It should be nearly identical to the sceneviewer.  I hope it resolves some of the zoom/pan issues you were seeing. Can you see if this helped?  I am still working out the double-tap reset issue.

artisteacher:
This is definitely a lot a better. It's still not quite perfect but I think I'll need to test this out when I'm somewhere with less humidity just in case that is making an impact.

manongjohn:
@shun-iwasawa Not sure what changed, but the OSX travis build keeps failing on something I did not change. Still no Appveyor build.

shun-iwasawa:
@manongjohn 
I've restarted the Travis build. 
As for Appveyor I still could not fix it. It seems to run the build for each PR, but fails to report the result back for unknown reasons... Sorry for the trouble.

manongjohn:
@shun-iwasawa i had restarted earlier but it still failed. Looks like it failed again for the same reason.

shun-iwasawa:
Hmm, it seems to fail with libtiff. [Travis started to use Xcode9.4 on July 31st](https://changelog.travis-ci.com/default-macos-image-is-now-xcode-9-4-67738) and it may be related to this problem. I'll try to check by forcing the older version of Xcode in Travis.

Anyway, I think I could fix AppVeyor. It seems to be back to normal. :-)

manongjohn:
Thanks!

artisteacher:
Double tap is working great - both to open nodes and to reset view. 

Zooming is behaving strangely again. Sometimes the zoom is jerky and a bit floaty, sometimes it is smooth. It's a little hard to predict but seems to be based on the part of the window where the cursor is located (ie. slightly to the left or right of center - among other locations). Having the cursor over the background might make the jerky zoom a little more likely. There seems to be a problem with the zoom point - especially when the schematic window is not expanded to take up the full program window. In that case, zooming seems completely stuck on the diagonal in that case (similar to the gif I shared earlier). When the schematic is expanded, zooming is still off from the location of the cursor but not necessarily stuck to the diagonal.

On the topic of zoom in expanded/contracted views of the schematic, I've never liked how the schematic view does not update the window focus based on the window size. When the nodes are centered in full screen, they are often completely out of view when the schematic is out of full screen. I wish that the focus of the schematic stayed the same in both views - ie. if the nodes are centered & fit the window in fullscreen I wish they would also be centered & fit the window out of fullscreen.

artisteacher:
2 finger pinching on a touchpad - I don’t have access to a touchscreen. Maybe @shun-iwasawa might have a suggestion?

artisteacher:
I believe it's the [force touch](https://developer.apple.com/macos/force-touch/) capability of mac touchpads that allows touch gestures to register. Is there a way to track the touch events?

By 2 finger slide, do you mean the 2 finger scrolling gesture? Right now in both the viewer and schematic, the zoom function is disabled during a 2 finger scroll that is recognized as a touch event (necessary for panning to work while leaving navigation possible).

With the changes, zooming is still jerky and the zoom center is definitely offset from the cursor location. Zooming in the viewer with a two finger pinch is smooth & centered on the cursor location.

Thanks for sticking with this! I wish I could be more of a help.

manongjohn:
Ah. Didn't know about that force touch ability in macs touchpads.

@shun-iwasawa As you have access to both windows and macs (hopefully one with a touchpad?) can you take a look to see what I might be missing to fix the behavior @artisteacher is experiencing with the 2 finger pinch zoom? The logic is nearly identical to  the scene viewer's logic. Am I missing something?

shun-iwasawa:
@manongjohn 
I could reproduce the center offset with 2 finger pinch zoom using touchpad. I'll look into this.

toonz/sources/toonzqt/schematicviewer.cpp
```diff
@@ -432,6 +544,198 @@ void SchematicSceneViewer::showEvent(QShowEvent *se) {
...
+  m_gestureActive = false;
+  if (QGesture *swipe = e->gesture(Qt::SwipeGesture)) {
+    m_gestureActive = true;
+  } else if (QGesture *pan = e->gesture(Qt::PanGesture)) {
+    m_gestureActive = true;
+  }
+  if (QGesture *pinch = e->gesture(Qt::PinchGesture)) {
+    QPinchGesture *gesture = static_cast<QPinchGesture *>(pinch);
+    QPinchGesture::ChangeFlags changeFlags = gesture->changeFlags();
+    QPoint firstCenter = mapFromGlobal(gesture->centerPoint().toPoint());
```
shun-iwasawa: I think this is a probable cause of the offset problem.
You are obtaining `firstCenter` by converting `gesture->centerPoint()` with `mapFromGlobal()` , but `gesture->centerPoint()` seems to be already in widget coordinate, not in the global screen coordinate.

toonz/sources/toonzqt/schematicviewer.cpp
```diff
@@ -432,6 +544,198 @@ void SchematicSceneViewer::showEvent(QShowEvent *se) {
...
+        }
+        if (!m_zooming) {
+          double delta = scaleFactor - 1;
+          m_scaleFactor += delta;
+          if (m_scaleFactor > .2 || m_scaleFactor < -.2) {
+            m_zooming = true;
+          }
+        }
+        if (m_zooming) {
+          changeScale(firstCenter * getDevPixRatio(), scaleFactor);
```
shun-iwasawa: Well, calling `setMouseTracking(true)` in ctor and modifying this line to
`changeScale(m_oldWinPos, scaleFactor);` 
seems to resolve the offset. 

I think I need to investigate further, anyway.

toonz/sources/toonzqt/schematicviewer.cpp
manongjohn: @shun-iwasawa Something I just noticed on my Windows machine in regards to pinch zooming on my **touchscreen** in both the Scene and Schematic viewers:

When **not** using mapFromGlobal() while getting the firstCenter, if I pinch zoom directly in the center of the either Viewer, it always causes the zoom to shift to/from the left.  I would have to be pinching left of center to zoom without causing a shift left or right.  However if I change the logic to use mapFromGlobal(), pinch zooming at the center does not cause any shifting.

This seems like a minor existing gesture bug to me.  I'd be curious to know if any others with a touchscreen see this pinching behavior also.

With or without mapFromGlobal(), **touchpad** pinching on my Windows machine works exactly the same and correctly, zooming relative to where the pointer is.

Would removing mapFromGlobal() fix the offset on the Mac?  If so, I can put a compile macro around that statement to use mapFromGlobal() for Windows compiles only.

manongjohn:
@shun-iwasawa There were 2 issues @artisteacher was seeing with this PR on a Mac

1) Zooming was off center.
2) Zooming was jerky at times.

The off center issue you were able to see. Did you also notice the jerky zooming?

turtletooth:
@manongjohn I'd say submit a PR without the macro and let's test it.  @artisteacher Do you have a way to test PRs before they are merged?

I'm actually creating a couple PRs tonight, so I don't mind making it if you are busy.

artisteacher:
@turtletooth I can test some things before they are merged, but there’s definitely something off right now in my build environment #2292 - I’m wondering if the build instructions for MacOS are out of date for Mojave.

shun-iwasawa:
There seems to be a difference in return value of `QGesture::centerPoint()` between when using **touchscreen** and **touchpad** .

I tested pinch zoom in detail with two devices, one with the **touchpad** on OSX and the other is **touchpad** on Windows. (For Windows I used a touch feature of wacom pen tablet  CTH-690)

### Pinch Zoom with Touchpad on OSX

- Confirmed that `QGesture::centerPoint()` returns in widget coordinate.
- It looks a bit jerky compared to the zoom on viewer.

### Pinch Zoom with Touchpad on Windows

- Found that the zoom operation is called from `SchematicSceneViewer::wheelEvent()` , not from `SchematicSceneViewer::gestureEvent` . 

Then I'll test with **touchscreen** on Windows.

manongjohn:
@shun-iwasawa I just noticed the update on your comment.  I never got a notice you had an update.  Thanks for following up!  I'll submit it with compile directives to use mapFromGlobal() for Windows builds only.

manongjohn:
@shun-iwasawa No prob. I revised the logic to use mapFromGlobal() if it detects a touchscreen instead of based on build directives.  It still works properly on my system with this change.  Please let me know how it works with Macs when you get a chance.

artisteacher:
Thanks @shun-iwasawa and @manongjohn for continuing to pursue this!

The zoom center still seems to be working differently in the schematic compared to the viewer. I'm also getting a weird redraw issue when zooming in the schematic. Here's a video comparing the experience in the viewer to the schematic using a touchpad & two finger pinch to zoom.
[zooming.mov.zip](https://github.com/opentoonz/opentoonz/files/2517331/zooming.mov.zip)

manongjohn:
So...I managed to get this PR built on my old Macbook pro running 10.9.5 (Mavericks) using QT 5.6.3 in hopes that I could troubleshoot the issues being encountered.

What i found was the 2 finger touchpad zoom was smooth unlike the video. I also didn't have any issues with the Hand tool.

Not sure if the issues encountered are with the OS version, QT version, both or something else. Unfortunately I don't have access to a newer macbook pro.

Was hoping to see this in the upcoming release, but don't want to merge something that will have problems. ☹

openanim:
@manongjohn @artisteacher @shun-iwasawa 

Question : no chance to see this PR includes for the next version? The possibility to disable or activate the multitouch is it not enough in case of problem ?

manongjohn:
I would really like to include this in the next release if at all possible.

Problem is I cannot easily debug and fix the issues being experienced on the macOS side. I don't have a macbook that is experiencing this problem. ☹

manongjohn:
After building this PR on a Macbook running macOS 10.11.6 using XCode 8.2.1 and QT 5.11.3, I still find no jittery/jumpy scrolling with 2 finger touchpad movement and no issue with nodes disappearing when using hand tool.

I am removing the WIP label so it can be reviewed.

artisteacher:
I have Xcode 10.1, and Cmake 3.12.3 on MacOS 10.14. I'm updating to QT 5.11.3 and will start a new build after the update to see if anything changes.

The difference in behavior we are seeing may be from differences in the hardware. My old 2010 MacBook pro (which is no longer usable) had a touchpad that you could physically press down on. On my 2017 macbook pro (and any mac touchpad from mid-2015 on), the touchpad doesn't shift at all when you press down on it but haptic feedback creates the illusion that it does. I'm pretty sure the newer touchpads are more sensitive and have more capability. I do have access to an older MacBook pro at school that I could test a portable build on. It should be running 10.11.6 too.

manongjohn:
@artisteacher I have an update coming shortly to fix a potential issue I noticed in which schematic zooming was not taking into account high DPI monitors when processing certain events that would likely be used by touchpads. It might explain why it is not jittery on older macbooks but it will be on newer macbooks if the resolution is higher.  

One way to maybe check this is to temporarily lower your resolution to 1280 x 800 and see if that corrects it?

artisteacher:
I do have a high res monitor - I'll give that a try.

artisteacher:
So I opened the app in low resolution and it was a much smoother and reliable experience. The problem with the incorrect zoom center is resolved. Zooming is smooth. The only remaining issue is that the disappearing nodes still happens with the hand tool.

Just an fyi, if you have a high resolution screen & want to open an app in low resolution, you can do so without changing the computer's resolution  (right-click on the app in Finder and select "Get Info"). If you check the low resolution option, the app will open in low res whenever it is opened.
screen shot 2019-01-06 at 9 25 34 pm

manongjohn:
Oh I guess it is resolution related. Thanks for verifying. I'm curious if the change I submitted earlier fixes it without having to go to a lower resolution.

manongjohn:
I'm curious...does running in lower resolution mode also help with the mouse over issues, like the onion skin mark not highlighting, you were seeing?

artisteacher:
Just updated the build to include your latest commit and it works just as well in high resolution.  Again the hand tool is still a problem but otherwise everything is great.

manongjohn:
Great! Thanks again for verifying the fix.  1 down, 1 to go. :)

Looking at the Hand Tool now to see if anything stands out.

artisteacher:
Hand tool works perfectly now!

manongjohn:
@artisteacher Thanks for all your help in getting this figured out!

openanim:
Thanks !

toonz/sources/toonzqt/schematicviewer.cpp
```diff
@@ -432,6 +553,200 @@ void SchematicSceneViewer::showEvent(QShowEvent *se) {
...
+    qDebug() << "[event] Gesture";
+    break;
+  default:
+    break;
+  }
+  */
+
+  if (CommandManager::instance()
+          ->getAction(MI_TouchGestureControl)
+          ->isChecked()) {
```
shun-iwasawa: I faced a crash at here when quitting the software. It is because actions may be deleted before closing this widget.
How about modifying the source just like I have done to `sceneviewerevents.cpp` ? Please see [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/sceneviewerevents.cpp#L1105-L1108).

shun-iwasawa:
@manongjohn 

I tried the PR with my MobileStudioPro (with high-dpi monitor) again.
Unfortunately, there seems to be an offset of zoom center on wheel zoom and unwanted scaling on panning.
I think you can reproduce the problem on your environment with normal monitor by setting `Change the size of text, apps, and other items` in Display settings to larger value (say, 200%).

schematic_mouse_offset

I'll investigate it again after reviewing other remaining PRs.

artisteacher:
After spending more time, I'm also seeing the offset zoom center - just in high resolution. However, it's still a lot better than it was earlier on. The weird diagonal thing doesn't happen. I don't have a mouse to test the wheel zooming problem.

manongjohn:
I am able to duplicate that offset issue with 175% (most I can do on my current machine). Will investigate.

toonz/sources/toonzqt/schematicviewer.cpp
manongjohn: Done. Will be included in my next commit after fixing offset issue.

manongjohn:
Got it working correctly on my Win10 laptop-tablet with pen, mouse, touchpad and touchscreen using display at 125% (default) and 175%.  Possible I broke macOS again. Please let me know how it goes.

artisteacher:
Yeah, it does break the touch gestures on MacOS. The hand tool makes the nodes disappear again (just in high resolution). Touch zooming is mostly off center (both high and low resolution). The zoom movement jerks unexpectedly on occasion. The zoom center is off for the zoom button (just in high resolution so far). Tap gesture to reset the view still works perfectly.

manongjohn:
@artisteacher When you get chance, can you check to see if my last commit fixed the hand tool?

artisteacher:
So the hand tool still only works in low resolution. Even more concerning, now the nodes disappear when zooming in both high and low resolution. Sometimes you can see flickers of the nodes if you continue zooming. Double tap to reset view brings them back.

manongjohn:
@artisteacher Thanks. Will continue to look at why it worked before and why its broken again.

Not sure when you rebuilt but i did commit a change that might fix offset.  Let me know how that goes.

artisteacher:
I rebuilt after. Thanks for your dedication in seeing this through!

manongjohn:
For the zooming and panning results in your recent tests, I'm assuming the current issues are the same whether you are using gesture or the button?

When you get a moment, can you fill the following chart out indicating results of  zooming and panning

Low Res - Touchpad - gesture
Low Res - Touchpad - button
Low Res - Mouse - button
Low Res - Tablet Pen- button
High Res - Touchpad - gesture
High Res - Touchpad - button
High Res - Mouse - button
High Res - Tablet Pen - button

artisteacher:
Low Res - Touchpad - gesture: panning=fine, zooming=nodes disappear/sometimes flicker
Low Res - Touchpad - button: panning=fine, zooming=nodes disappear/sometimes flicker
Low Res - Mouse - button:
Low Res - Tablet Pen- button: panning=nothing happens, zooming=nothing happens
High Res - Touchpad - gesture: panning=fine, zooming=nodes disappear/sometimes flicker
High Res - Touchpad - button: panning=nodes disappear, zooming=nodes disappear/sometimes flicker
High Res - Mouse - button:
High Res - Tablet Pen - button: panning=nothing happens, zooming=nothing happens

After playing around, I think that the panning/zooming is way too fast/sensitive. If the nodes are zoomed out enough and I zoom very, very slowly, the nodes don't disappear but do zoom in a jerky manner (jumping from one position to the next). The zoom in this case seems locked to the diagonal again. Similarly with panning, if the nodes are zoomed out enough and I use the button to pan as slowly as possible, the nodes don't disappear - though they are almost off screen.

I don't have a mouse...

artisteacher:
Zooming has the same problem but button panning is working completely now (both with touchpad & stylus - high & low resolution).

artisteacher:
Zooming is working with low resolution.

artisteacher:
The nodes stay visible in high resolution zooming! The zoom center seems off though.

manongjohn:
Thanks! Was hoping to get back to that point at least :)  Now to figure out why it's off...

manongjohn:
@artisteacher  Can you do me a favor and in `schematicviewer.cpp`, please replace the `changeScale()` function with this:

> void SchematicSceneViewer::changeScale(const QPoint &winPos,
                                       qreal scaleFactor) {
  QPointF startScenePos = mapToScene(winPos);
  QMatrix scale = QMatrix().scale(scaleFactor, scaleFactor);
  setMatrix(scale, true);
  QPointF endScenePos = mapToScene(winPos);
  QPointF delta = endScenePos - startScenePos;
  translate(delta.x(), delta.y());
}

and let me know the effect on touchpad zooming.

artisteacher:
It's better but not always perfect. Part of the time the zoom center is correct but part of the time it is still off center. I'm not sure exactly what conditions result in the incorrect center - but it could be related to fingers sliding slightly while making the pinch gesture.  When the zoom center is working correctly, panning can't get triggered while zooming. When the zoom center is off, then panning can occur while zooming.

artisteacher:
It definitely doesn't on my low resolution screen - which I use more frequently with OpenToonz. I don't think it happens in high resolution either, but I should spend a little more time testing to be sure.

artisteacher:
So it can happen in the Viewer with a high resolution screen.

manongjohn:
Thanks for verifying.

Since a zoom blocks a pan, if a zoom follows a pan, I'll cancel the pan and hopefully that will stop the offset issue....unless someone disagrees with this action ofcourse.

artisteacher:
It's almost there, but instead of sliding around, sometimes the focus jumps slightly off course instead of staying focused on the cursor. 
[slightly_off.mp4.zip](https://github.com/opentoonz/opentoonz/files/2774910/slightly_off.mp4.zip)

shun-iwasawa:
@manongjohn @artisteacher 
Thank you very much for the continued working on this! Can you please advice if it seems to ready to be reviewed & merged or seems to need some time for brushing up?

artisteacher:
It is getting close. High res panning with the button still has the offset position issue and the slight jumping still happens while zooming in high resolution.

shun-iwasawa:
Understood.
I'm so sorry but I hadn't enough time to investigate this PR to complete the remaining problems.
I'll postpone including this enhancement to the future version.

manongjohn:
@shun-iwasawa No problem.  

Since we have more time, I am working on adding gesture controls to ImageViewer (color model, flipbooks), PlaneViewer (Antialias, Adjust Thickness, etc) and SwatchViewer (fx nodes).  Would you prefer I put all those in another PR or should I merge them into this one and test them all at once?

artisteacher:
@manongjohn  - all of those additional gesture controls will be fantastic! The viewer from the "create mesh" window for the plastic tool would also benefit from touch gestures.

manongjohn:
The create mesh viewer falls under the PlaneViewer. I verified it today actually. I think the 3 main viewers types I listed are the ones used throughout OT.

On these other viewers, i also added right-click context menus with Reset View and Fit to Window commands.

Almost done. Just working out 1 issue with panning on SwatchViewer. Then the only question left is...will they work without issue on macOS? 😁

artisteacher:
Great! Here's hoping! When those viewers open, they are almost always centered on the wrong area - not where you actually have a drawing. Could they open by default set to "Fit to Window"?

manongjohn:
@artisteacher Question regarding the offset for panning with button in high res:  Is the offset seen when using touchpad, tablet pen or both?

artisteacher:
It's mostly the touchpad. It looks like there can be a very slight offset with the pan button but is practically unnoticable. While I was checking, I found that if I am using a stylus and then use the touchpad and after switch back to the stylus, the buttons don't work for pan or zoom.

manongjohn:
I can't seem to reproduce that issue. I think i need more details on what you are doing with stylus, then with touchpad, then with stylus again.

artisteacher:
If I zoom and pan with the buttons using the stylus, then use the touchpad to zoom and pan, when I try to use the zoom/pan buttons again with the stylus, they have no impact.

RodneyBaker:
This PR could use some testers... especially those using Macs.
Thanks!

manongjohn:
I would like to merge this in the next few days.  Are there any objections or particular reasons I shouldn't?

If you would like to test on macOS or Linux but don't have time to build, i suggest using the OTX release.

manongjohn:
I've not heard anything so I'm going to go ahead and merge this.

PR #2188, opened 2018-07-24
Fix Display of Motion Path

### LINKED ISSUE #2187: Can't See a Motion Path While Drawing
### Issue Summary
This problem is [reported in the forum] (https://groups.google.com/forum/#!msg/opentoonz_en/B9JsVCt7xPs/tqgeI11pCAAJ).

@pnelson2 reported;
> When I draw a motion path it doesn't show the line until I take my pen off the digitizing tablet, so I can't see it while I'm drawing it!

### Steps to Reproduce
Confirmed that the stroke always disappears when you draw the motion path which is connected to some level.

### System Information
* **OpenToonz Version**: Latest
* **Operating System**: Win7

shun-iwasawa:
This will fix #2187


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2191, opened 2018-07-25
Revert Alt-Dragging Feature to Fx Schematic

### LINKED ISSUE #2181: Insert or extract node with alt does not work properly.
### Issue Summary

I open a specific issue for this bug.

normally  to insert a node into a link we can do Alt click and drag it into the  link, and to extract a node from a link, Alt click and drag it away from  the link. But the behavior no longer works, to put the link correclty.  However, as long as we have not permanently dropped the node, the  behavior still seems to be there. Is it possible to bring it back ? it  looks like a bug.

alt_node don t works

### Steps to Reproduce

hold the alt key and try to incorporate the nodes automatically, it does not work.

### System Information
* **OpenToonz Version**: 1.21
* **Operating System**: w10

shun-iwasawa:
This will resolve #2181 .
The useful feature of Fx Schematic existed in Toonz Harlequin is reverted. Which is as follows, according to the [User Manual](http://opentoonz.readthedocs.io/en/latest/applying_special_fx.html#editing-the-fx-schematic):

> **To insert an FX node into a link:**
> Alt-click and drag it onto the link.
> 
> **To extract an FX node from a link:**
> Alt-click and drag it away from the link.

Also I refactored the source to skip process which seems to be unnecessary.

### DIFF
--- a/toonz/sources/toonzqt/fxschematicscene.cpp
+++ b/toonz/sources/toonzqt/fxschematicscene.cpp
@@ -1499,6 +1499,7 @@ void FxSchematicScene::mousePressEvent(QGraphicsSceneMouseEvent *me) {
   FxsData fxsData;
   fxsData.setFxs(m_selection->getFxs(), m_selection->getLinks(),
                  m_selection->getColumnIndexes(), m_xshHandle->getXsheet());
+  // m_isConnected indicates that the all selected nodes are connected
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2192, opened 2018-07-25
Remove Unnecessary Schematic Update

shun-iwasawa:
After introduction of #2034 , Fx Schematic no longer switches between normal/icon view on zooming and thus it does not need to rebuild its content.
This PR refactors the source, removing such unnecessary process.


shun-iwasawa:
I'll test AppVeyor with this PR

masafumi-inoue:
LGTM

PR #2194, opened 2018-07-25
ignore Thumbs.db

otakuto:
It is not necessary file.


shun-iwasawa:
Thank you @otakuto for cleaning up the repository! LGTM

PR #2195, opened 2018-07-25
using std::swap instead of tswap

otakuto:


shun-iwasawa:
Jenkins

shun-iwasawa:
As far as I could see, `std::swap` works with no problem in any of parts of this modification.
Thank you @otakuto for refactoring. LGTM

PR #2196, opened 2018-07-25
Stage Schematic Multiple Connection Feature

shun-iwasawa:

When building a stage schematic tree, it often happens that multiple columns need to connect as a child of one pegbar or another column. In such case they had to connect columns one by one, which was quite troublesome.

This PR enables you to connect multiple nodes to the parent node at the same time, as shown in below:

stage_schematic_multiple_connect

Please note that this feature is available only when you start dragging from the parent (blue-colored) port of stage schematic node.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2197, opened 2018-07-25
use std::unique_ptr instead of std::auto_ptr

otakuto:
clang says

```
opentoonz/toonz/sources/toonz/penciltestpopup.cpp:1700:14:
warning: 'template<class> class std::auto_ptr' is deprecated [-Wdeprecated-declarations]  
    const std::auto_ptr<NameBuilder> nameBuilder(NameBuilder::getBuilder(L""));
```

I replaced std::auto_ptr to std::unique_ptr.


otakuto:
Jenkins

PR #2199, opened 2018-07-26
Inputting Numbers in XSheet Cell Can Use the Level in the Following Cells

shun-iwasawa:

When inputting the xsheet by using the rename cell field which pops up by double-clicking on the xsheet cell, typing the number had no effect if there was no occupied cell in the above.
I changed its behavior to search the following cells if all the above cells are vacant. It will be useful when you would like to input a cell on top of the column.

rename_cell

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -803,6 +803,10 @@ void RenameCellField::renameCell() {
       // if there is no level at the current cell, take the level from the
       // previous frames
       // (when editing not empty column)
+      if (xsheet->isColumnEmpty(c)) {
+        cells.append(TXshCell());
+        continue;
+      }
 
       TXshCell cell;
       int tmpRow = m_row;
@@ -811,6 +815,19 @@ void RenameCellField::renameCell() {
         if (!cell.isEmpty() || tmpRow == 0) break;
         tmpRow--;
       }
+
+      // if the level is not found in the previous frames, then search in the
+      // following frames
+      if (cell.isEmpty()) {
+        tmpRow       = m_row + 1;
+        int maxFrame = xsheet->getFrameCount();
+        while (1) {
+          cell = xsheet->getCell(tmpRow, c);
+          if (!cell.isEmpty() || tmpRow >= maxFrame) break;
+          tmpRow++;
+        }
+      }
+
       TXshLevel *xl = cell.m_level.getPointer();
       if (!xl || (xl->getType() == OVL_XSHLEVEL &&
                   xl->getPath().getFrame() == TFrameId::NO_FRAME)) {


shun-iwasawa:
Jenkins

gab3d:
While at this...
Could it be made that the two Enter keys function a bit differently?
- Numpad Enter: Focus passes on to next cell (as shown above).
- Main keyboard Enter: Ends number entering and focus returns to main program.

turtletooth:
@gab3d I don't think that would work for laptop keyboards.

gab3d:
@turtletooth
Valid point, but perhaaps it could be the other way round,  i.e.:
- Main keyboard Enter: Focus passes on to next cell (as shown above).
- Numpad Enter: Focus is kept in the same cell, to try other numbers right there.

That would preserve the current normal behaviour available in most cases and still offer the other alternative...

Note: Perhaps the alternate behaviour could be triggered by Shift + Enter or Ctrl + Enter instead. I think it'd still be a useful thing to have.

shun-iwasawa:
@gab3d 

> Numpad Enter: Focus is kept in the same cell, to try other numbers right there.

Sorry but I don't quite understand what this behavior is.
Is it different from deleting current input number with BackSpace key then type another number?

Anyway, I'm reluctant to assign different behaviors to two Enter keys as it may cause confusion. As you suggested, adding some modifier key would be better.

gab3d:
@shun-iwasawa 
no, it's not different from deleting with backspace and entering another number, but is essentially much more practical than doing that. Let me explain myself:

Current behaviour:
1. Double click on a cell and enter another number.
2. Press Enter to accept the new number and show the corresponding drawing. The focus goes down to the next cell. (Suppose the entered number showed an unintended drawing)
3. Select back the previous cell by any means.
4. Go to point 2 and repeat until the number you enter shows the expected drawing.
5. Press Esc to exit.

Proposed behaviour:
1. Double click on a cell and enter another number.
2. Press Shift + Enter (or whatever) to accept the new number and show the corresponding drawing. The focus remains at the current cell, and the newly entered number remains selected ready to be changed again if necessary.  (Suppose the entered number showed an unintended drawing)
3. Go to point 2 and repeat until the number you enter shows the expected drawing. 
4. Press Esc to exit.

(This behaviour I'm describing is a function I've used a lot in Maya, and it can save a lot of clicks and also help remain focused when trying to find the right value for any parameter). In Maya it works with the main and numpad Enter keys, as I initially asked for, but I think perhaps it could be useful too using a modifier key to trigger it.

Here we're talking of cell number editing, but it could really be useful for any number input field in OT.

shun-iwasawa:
@gab3d 
Thank you for clarification. Now I think I understand your demand. 
I hadn't imagined that this field is used for "finding" the right drawing in the level by trial and error as you mentioned, because we usually use this field to transcribe xsheet content from paper to OT.
Anyway, it can be discussed as a separate feature request.

shun-iwasawa:
I'll use this PR for checking Appveyor.

masafumi-inoue:
LGTM

PR #2208, opened 2018-08-02
Use Xcode 8.3 on Travis CI

shun-iwasawa:
This PR is to check if the recent Travis CI failures are due to the Travis CI update on July 31st, [the default macOS image was changed to Xcode 9.4](https://changelog.travis-ci.com/default-macos-image-is-now-xcode-9-4-67738).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,3 +15,4 @@ matrix:
         - os: linux
           compiler: clang
         - os: osx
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
OK, it works.
Then let's keep using the older version of Xcode until we eventually fix the problem.

Issue #1423, opened 2017-08-21
"Test tween command" and inbetween pop-up [$200]


ghost:
It 's a request to save time in the tweening process workflow. (already posted on the forum https://groups.google.com/forum/#!category-topic/opentoonz_en/feature-requests/JFhxb9NvPeo)

Currently in opentoonz the step to create a tweening is done by selecting the two keys in a range after adding empty inbetween frames by insert command.

It would be a huge time saver if we could preview this step in a single command (for exemple available on xsheet toolbar)

a button to simulate the process to create a tweening (it looks like a script ?) and use it as corrector 

without having to do it manually as long as we are not on a finalization step, by drawing only the keys from beginning.

The goal is to be able to check if our strokes/drawing goes in the right direction without finishing it completely.

The recent guided drawing feature is a great addition, and I think combining with this feature, we can have a total control in this type of process.

Ideally, we should preview with any key in the level strip or xsheet.

And select in a second button the number of inbetweening frames taken into account something between 1 and 12 for example. I can give more example if necessary.

https://1drv.ms/i/s!AspsYffPSwbdh9wtEIBJK5JDc4hDTw

The idea comes from cacani  "Quickflip" : https://www.youtube.com/watch?v=1LIrPyaG03o

I don't know the feasibility rate but opentoonz has many elements in common to do it.

If other people are interested please help me to promote it

PS : Sorry about the bad English I hope this is understandable, 

I will submit this to bountysource

openanim:
I create the bounty here : https://www.bountysource.com/issues/48479745-check-strokes-command

Note: Take into consideration with this bounty that I have no idea how difficult it is to implant this.

morevnaproject:
Hello @openanim! Thank you very much for initiative and providing a bounty! I will do my best to spread a word about it. ^__^

openanim:
thanks @morevnaproject , I appreciate it :)

blurymind:
This would be a huge time saver for me! One thing to note is setting the number of inbetweens and the type of tweening needs to be set somewhere for the quickflip preview - somewhere convenient.
Cacani has the advantage of having a dedicated gui element for onion skin settings and another one for tweening
![quickflip](https://user-images.githubusercontent.com/6495061/29961040-fc826080-8ef5-11e7-8fa4-de3b1ffd1bb8.PNG)
Another feature that would make this feature absolutely priceless is if had a quick command to generate a new inbetween from the quickflip after we are happy with it.

[...]

PR #2212, opened 2018-08-03
Enhancement of the Shift and Trace Feature

shun-iwasawa:
This PR will enhance the Shift & Trace feature, based on demands from some Japanese animation studio.

- The frames to be shown for "ghosts" can now be specified by user. A special gadget appears at the onion skin handle both in the xsheet and in the level strip. (Previously, only neighbor frames were shown.)
- Enabled this feature when the level strip frame is selected as well. (Previously, this feature was available only when the xsheet frame is selected.)
- On the Edit Shift mode, the control boxes are now colored (with the onion skin colors) in order to make it easier to know which "ghost" drawing is grasped. 
- On the Edit Shift mode, users now can modify scale of the "ghost". Holding shift+drag can modify vertical and horizontal scales independently.
- Transparency of the "ghost" now can be modified with "Preferences > Onion Skin > Paper Thickness"
- Holding F1, F2 or F3 key on the Viewer will display only the corresponding "ghost" or the current drawing. This feature can "flip" your drawing for checking the movement.

shiftandtrace

### DIFF
--- a/toonz/sources/include/orientation.h
+++ b/toonz/sources/include/orientation.h
@@ -118,7 +118,9 @@ enum class PredefinedRect {
   ZOOM_OUT_AREA,
   ZOOM_OUT,
   LAYER_FOOTER_PANEL,
-  PREVIEW_FRAME_AREA
+  PREVIEW_FRAME_AREA,
+  SHIFTTRACE_DOT,
+  SHIFTTRACE_DOT_AREA
 };
 enum class PredefinedLine {
   LOCKED,              //! dotted vertical line when cell is locked
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -669,6 +669,26 @@ protected slots:
   void updateRealTimePickLabel(const int, const int, const int);
 };
 
+//=============================================================================
+//
+// ShiftTraceToolOptionBox
+// shown only when "Edit Shift" mode is active
+//
+//=============================================================================
+
+class ShiftTraceToolOptionBox final : public ToolOptionsBox {
+  Q_OBJECT
+  QPushButton *m_resetPrevGhostBtn;
+  QPushButton *m_resetAfterGhostBtn;
+  void resetGhost(int index);
+
+public:
+  ShiftTraceToolOptionBox(QWidget *parent = 0);
+protected slots:
+  void onResetPrevGhostBtnPressed();
+  void onResetAfterGhostBtnPressed();
+};
+
 //-----------------------------------------------------------------------------
 
 class DVAPI ToolOptions final : public QFrame {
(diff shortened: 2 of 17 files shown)


shun-iwasawa:
Jenkins

711GM:
Amazing!

openanim:
If our function keys are already used for other commands, will it be possible to assign other shortcuts for the flip features ?

matitanimata:
this is awesome!

shun-iwasawa:
@DarrenTAnims 
Thanks, I read your post in the forum. If I extend the flipping feature to multiple frames (more than 3), I think I'll assign function keys to each onion skin frame, so that users can specify which frames to flip. Anyway, I'll ask user's feedback with this feature and see how that feature is demanded. 

@openanim 
No. For now F1, F2 and F3 keys are hard-coded and will override existing commands only when the "Shift & Trace" mode is active and the Viewer is focused. ( I just fixed the source and now the overriding works properly. )
Well, I think these shortcuts are not suitable for making configurable via "Configure Shortcuts" like other commands, because they are only available under the "Shift & Trace" mode and assigned keys should be contiguous. 

Jenkins again

Orphanlast:
I would like to put my two cents in. This is a dream come true. I've been talking to Matitanimata (Franko Bianch) about shift and trace improvements and NOT needing to play musical chairs with your frames to get shift and trace to work sounds like a DREAM COME TRUE. 

It would be cool to see shift and trace menues in the properties menue in the same region as the Onion Skin information. This way you can customize shift and trace "skins" to be a specific color. This way you never confuse a shift and trace "skin" for an onion skin

DarrenTAnims:
Hi @shun-iwasawa . Thanks for replying back. Tbh, if you just had flipping on 3 frames (outside of shift and trace), that'd be a huge step forwards for me. That's mostly what I need. And then maybe later extend it to 4, 5 or whatever. However, I  must say (because I forgot earlier) that this change to the Shift and Trace is awesome. I'm really looking forward to seeing it.

openanim:
@shun-iwasawa.

It works perfectly now, thank you!

I also think that DarrenTAnims' request would be a nice addition,
In this case, if we can specify the frames we want as you suggested it will be great.
By the way, the requested "flip" feature also looks like this #1423  but as part of tweening. And especially with Blurymind suggestion's he made to his last post at the end of the thread. I do not know how this can be related. But maybe it could help in the future to get it, too. 

Just another thought, would it be difficult to use onion skin color preferences for previous and next frames ? This would also facilitate the inbetween for very close poses in shift and trace mode. Something like a box "activate colors of images". Or two boxes, one for the color of the previous image and the other for the color of the next image that is related to the possible preferences.

Anyway, thanks for this enhancement !

hhsyya:
I'm frustrated that Shift & Trace is not available on the lower level while I'm working on the upper level. Shift & Trace can only be used left and right

PR #2240, opened 2018-08-15
Shift and Trace Improvement

shun-iwasawa:
This PR improves the shift and trace feature as follows:

- When the scene frame is selected, non-current column will also display images in correspondent reference frames when flipping with F1, F2 and F3 keys. ( Previously only the current column was flipped and the non-current columns stayed unchanged. )
- The "Edit Shift" tool options now includes radio buttons for selecting a target drawing to move. 
- The "Edit Shift" tool options now displays color chips synchronized with onion skin colors.
- The "Edit Shift" tool options' reset buttons will be disabled if there is no shift.
   image

- Added a new preference option for displaying the reference drawings being filtered with the onion skin colors. ( `Onion Skin category > Use Onion Skin Colors for Reference Drawings of Shift and Trace` )
- Added icons for `Shift and Trace` , `Edit Shift` , and `Reset Shift` commands like an image below:
image

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -607,7 +607,8 @@ public:
   //! onionskins)
   //! (window coordinate, pixels, bottom-left origin)
   virtual int posToRow(const TPointD &p, double distance,
-                       bool includeInvisible = true) const = 0;
+                       bool includeInvisible  = true,
+                       bool currentColumnOnly = false) const = 0;
 
   //! return pos in pixel, bottom-left origin
   virtual TPointD worldToPos(const TPointD &worldPos) const = 0;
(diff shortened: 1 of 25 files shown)


shun-iwasawa:
Jenkins

openanim:
Perfect ! Thanks !

gab3d:
Really nice improvements, kudos to you!

A minor suggestion: 
The Reset Shift icon perhaps could include a "more standard" concept to express "reset", such as something similar to the depicted ones in the following image:

reset-options

konero:
Personally I prefer the icon Shun made, looks the most similar to the shift and trace (general reset) icon in Clip Studio so was familiar to me at least.

Very nice enhancement by the way.

masafumi-inoue:
LGTM

PR #2241, opened 2018-08-16
Schematic UI Modification

shun-iwasawa:
This PR modifies schematic UI based on feedbacks from Japanese animation studio.
I understand UI is always a matter of preference and hope most of users will accept this.

- The preview and the camstand visibility toggle buttons become gray with yellow / orange border line when inactive. Previously the toggle buttons became semi-transparent when inactive and they said it was quite hard to recognize.
- Column nodes become gray with column-colored border line when they are reference (= not rendered) columns, just like the color for reference columns in xsheet.
- The default value of `Preferences > Xsheet > Current Column Color` is changed from red to yellow, because red color may get misunderstood as some error. As you can see in the image below, yellow text color seems not to be eye-friendly under the Light style ( or when using Classic column header layout for xsheet ) but I  prioritized an appearance of UI with default settings (i.e. Default style + Classic-revised column header).

schematic_color_change


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2264, opened 2018-08-28
Fix Updating Animated Guided Drawing

### LINKED ISSUE #2251: Animated "Guided Drawing" option needs to be removed.
With the new update to the viewer only updating pixels that have changed, the animated guided drawing option needs to be removed.  The static arrows still work, but the animated lines don't update properly.

I would remove it myself, but I can't get my build environment to work.

shun-iwasawa:
This will fix #2251

If the animated guided drawing appears on the viewer, the bounding box of such guide strokes will be added to the clipping area of the viewer.
I think this will (hopefully) not cause an appreciable slowness when not using this mode...


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

Issue #2262, opened 2018-08-27
The handle for moving the frames in timeline view is hard to click


turtletooth:
This may be just me, but I have a hard time clicking the handle for moving frames when in timeline view.  It's a pretty small target especially when using a pen vs a mouse.  @manongjohn Is there any way to get a couple more pixels of height in there to improve the target?

z-uo:
I agree, with a tablet, it is very small. Keyframe are also very small, I tend to take my mouse to move these.

Is there a need for a handle, maybe the whole cell can be draggable?

konero:
@turtletooth I recommended the same thing awhile back, though can't remember what the problem was.

manongjohn:
@konero probably me being stubborn and wanting to keep similar behaviors between xsheet and timeline. 😁

Whatever change is made will probably also impact the Column Header drag bar.

gab3d:
@turtletooth I'd have no problem at all with such a system, but I suspect the japanesse animation studios currently using OT would have to give their agreement as well for that to be included.

artisteacher:
Would that interfere with renaming or inputting frame numbers?

phillycheese007:
what would being able to click directly on the cell to move interfere with? if it's just a matter of consistency i would place that at lower priority than ease of use. a little info in the docs to take care of any confusion

artisteacher:
I think that would interfere with click & drag to select multiple cells.

konero:
The only feature it would interfere with is **Enable to Input Cells without Double Clicking**, found under the Xsheet settings in the Preferences panel, I think.

I wonder if this could be solved with a cursor change? For example when you hover the handle the cursor changes to the hand icon, this is what happens in Clip Studio.

![csp_cursor](https://user-images.githubusercontent.com/19820721/44947107-e8df1580-adff-11e8-9c45-cda41e00fc5b.gif)

Clip Studio's handle is 7px.

gab3d:
@z-uo  I agree. This should be fairly intuitive in most cases.
- if the desireds cells are not selected the action (click+drag) would allow you to select them.
- if they are already selected the same action (click+drag) o one of them would allow you to move them. I see this second step logical, because you already have the desired cells selected anyway.

the solution proposed above by @konero (changing the cursor icon) would give the user the perfect feedback to know what to expect in his/her next action.

artisteacher:
Changing the cursor is a great idea.

Issue #2261, opened 2018-08-27
cant make a right clic in the timeline between 2 keyframe


z-uo:
### Issue Summary
In the timeline, I can't make a right clic in the cell between 2 keyframe (with the tiny white line) except on the upper part of the cell (watch the red zone on the screenshot below)
in the xsheet the forbiden zone is only around the white line ( which is less a problem)

### Screenshot or Video Reference
![capturerc](https://user-images.githubusercontent.com/652668/44657122-12f99780-a9fc-11e8-9651-1a733f2a3964.PNG)

manongjohn:
I'll see if I can address this with #2262.

The issue here is the context menu for the keyframe line takes precedence over the cell's context menu.  When zoomed out that much, the area of keyframe context menu activation takes up most of the cell.

I might need to combine context menus to make it easier to access either the cell or the keyframes.

z-uo:
But there are no context menu for the keyframe line? 
And yes with the timeline zoomed in there is only a little space that is not clickable.

PR #2268, opened 2018-09-01
Timeline height and keyframe area adjustments

manongjohn:
This PR addresses #2261 and #2262 by the following:

- Increased the height of the timeline layers

image

With the increased height, I also increased the height of the dragbar so the height matches the width of the xsheet dragbar.  Hopefully this will make it a bit easier to target with a pen.

**NOTE**: This does not mean this is the end of cell/column dragging conversation! It's merely to help for now in case it takes a while to find an agreeable dragging method.

Additionally, the increased height actually makes the timeline a bit easier on the eyes to look at.

- Narrowed the activation area for key frame selection

I narrowed the active selection area around where the keyframe icons and white lines are. Coupled with the increased height, this should allow for slightly better click/right-click access to target the cell rather than the keyframe icons/lines with a pen.

- Update: Added thumbnail into timeline

With the increased height, it made for a better area to display a thumbnail

image

With the addition on thumbnail
1) Sound icon moved to thumbnail area
2) Filter color icon moved to thumbnail area

Additionally, made a Config button dropdown just for the Sound volume control, removing it from the thumbnail/name areas.  This change was made for all layouts except the "Classic" view since there is no config button. It should be noted that volume setting < 100 will change the Camera Stand Visibility icon similar to changing Opacity on other columns.

image

### DIFF
--- a/toonz/sources/include/orientation.h
+++ b/toonz/sources/include/orientation.h
@@ -120,7 +120,11 @@ enum class PredefinedRect {
   LAYER_FOOTER_PANEL,
   PREVIEW_FRAME_AREA,
   SHIFTTRACE_DOT,
-  SHIFTTRACE_DOT_AREA
+  SHIFTTRACE_DOT_AREA,
+  PANEL_EYE,
+  PANEL_PREVIEW_LAYER,
+  PANEL_LOCK,
+  PANEL_LAYER_NAME
 };
 enum class PredefinedLine {
   LOCKED,              //! dotted vertical line when cell is locked
--- a/toonz/sources/toonz/layerheaderpanel.cpp
+++ b/toonz/sources/toonz/layerheaderpanel.cpp
@@ -77,9 +77,9 @@ void LayerHeaderPanel::paintEvent(QPaintEvent *event) {
                      ? m_viewer->getLayerHeaderLockOverImage()
                      : m_viewer->getLayerHeaderLockImage());
 
-  drawIcon(p, PredefinedRect::EYE, boost::none, preview);
-  drawIcon(p, PredefinedRect::PREVIEW_LAYER, boost::none, camstand);
-  drawIcon(p, PredefinedRect::LOCK, boost::none, lock);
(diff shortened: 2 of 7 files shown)


shun-iwasawa:
Jenkins

manongjohn:
@shun-iwasawa Sorry for the last minute changes if you started reviewing this.  Unless you have some change requests, I don't think I have any further changes for this PR.

@turtletooth I'm always glad to hear the timeline is being used and has brought a positive working experience to OT. 

@artisteacher It's funny, I didn't realize how much I missed the thumbnail until after I added it to the timeline and working on some animations. Definitely makes identifying a column a little quicker than trying to read names all the time. :)

turtletooth:
I tested this tonight.  Works really well.  It's nice to have a thumbnail for the columns again.  There's part of me that hates to lose some of the vertical space for more columns, but it's probably a worthwhile trade off in order to have the thumbnail - especially for someone like me who forgets to label his columns.

As far as dragging, with the way that OT shows its keyframes, using the header dragging method is still probably the best.  

BTW - I don't know if this is practical, but would it be possible for columns to show the name of the level on the column instead of Col1, Col2. . . ?

manongjohn:
@turtletooth I know that suggestion or something similar has been requested and debated before, with no real agreeable solution.

Thinking about it a little more, though, I guess we could add a preference to default the Column name to just display the filename of the 1st level in the column, similar to how the thumbnail is chosen, but not actually name the column that.  If it is an empty column, it will remain Col#.  If the user has overridden the column name, that will display instead. 

This would basically just be a visual replacement. Any logic that references its name would still get Col# or whatever the user named the actual column.

It would likely need to also change in the Schematic Viewers, Function Editor and anywhere else Col# is displayed for consistency and visual reference. I would likely handle this in a separate PR if we want to go ahead and do this.

shun-iwasawa:
Jenkins

shun-iwasawa:
Thank you @manongjohn for the improvement ! LGTM

PR #2277, opened 2018-09-06
use std::unique_ptr instead of tcg::unique_ptr

otakuto:
Use std::unique_ptr instead of tcg::unique_ptr, and delete unique_ptr.h file.


shun-iwasawa:
Confirmed that it works fine. Thank you @otakuto ! LGTM

PR #2299, opened 2018-09-19
Load Palette's Global Name on Loading Level

### LINKED ISSUE #2298: Studio palette loses its link-ability once it is loaded on the scene and overwritten
### Issue Summary
Once you load studio palette into the scene by using load level command and overwrite it, the palette loses its link-ability and it turns into normal (non-studio) palette, containing styles with invalid links.

### Steps to Reproduce

1. Create a new studio palette in the studio palette window, under the "Project Palette" folder. (A palette item with a green icon is created. The green icon means it is a studio palette. )
1. Load the palette into the scene with load level command.
1. Select the palette level and modify it in the palette.
1. Overwrite the palette.
1. Open **another** studio palette window and see the icon of the palette item. 

### Expected Results
The studio palette should remains studio palette (with green icon) even if it is overwritten.

### Actual Results
The palette turns into normal palette with orange icon and loses its link-ability.
Showing the styles in the list view fails to display the original path of the link.

### System Information
* **Operating System**: Win7

shun-iwasawa:
This will fix #2298 
`TIStream::operator>>(TPersist *&v)` seems to have no ability to load attribute value of the current tag.

### DIFF
--- a/toonz/sources/image/tzl/tiio_tzl.cpp
+++ b/toonz/sources/image/tzl/tiio_tzl.cpp
@@ -1409,8 +1409,15 @@ void TLevelReaderTzl::readPalette() {
   TPalette *palette = 0;
 
   if (is && fs.doesExist()) {
-    is >> p;
-    palette = dynamic_cast<TPalette *>(p);
+    std::string tagName;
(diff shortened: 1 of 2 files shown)


masafumi-inoue:
LGTM

PR #2312, opened 2018-10-02
Search in the Special Folder Trees on Moving Folders in the File Browser

shun-iwasawa:
This PR will modify a behavior of the folder tree view on the left side in the file browser.
When accessing some folder which is not located under the project folders, if such folder is located under `My Document` , `Desktop` , or `Library` folder, the current node in the tree view will move to under the corresponding tree.
Before this change, accessing such folders always move the current node under the `My Computer` tree.

specialfolder


shun-iwasawa:
Jenkins

gab3d:
nice touch! :)

RodneyBaker:
It's often the little thing like this that make a big difference.
If I'm reading this right this update will resolve a frustration I have had with quickly finding files.    Looking good!

masafumi-inoue:
LGTM

PR #2314, opened 2018-10-02
New Feature : Adding Clapperboard

shun-iwasawa:
This PR will introduce a new feature, adding clapperboard.
The clapperboard is a single image showing the information of the scene. It will be added to the head of the rendered footage, generally in order to record and convey the information to subsequent workflows.
This feature is requested by some Japanese animation studio.

courage_with_board

### How to Add the Clapperboard

- Open the output settings.
- Select the output format to movie type, such as mp4.
- Open "Other Settings" group box.
- Toggle "Add Clapperboard" checkbox.
- Click "Edit Clapperboard..." button. The clapperboard settings dialog opens.
- Set the "Duration (frames)" to more than 0.

### How to Edit the Clapperboard

image

The clapperboard consists of the "board items".
In the bottom-right of the dialog there is a list view of all items.
Each item has a information type. Most of them will retrieve the correspondent information from the current scene automatically, such as duration of the scene, scene file path, current date, etc.
Settings of the selected item is displayed on the top-right of the dialog.

The settings can be saved as a preset in order to reuse among scenes.
It can also be stored in the project's default settings, by using "Save Default Settings" command.

For details, please see [the YouTube video](https://youtu.be/ot-6Yd3Bx1c).

### Warning!
Adding the clapperboard will lose compatibility of the scene file with the older versions of OpenToonz. Setting the duration to 0 will remove the clapperboard data from the scene at all so that you can maintain the compatibility.

### DIFF
--- a/stuff/library/clapperboards/preset_sample.clapperboard
+++ b/stuff/library/clapperboards/preset_sample.clapperboard
@@ -1,6 +1,6 @@
 <clapperboardSettingsPreset>
   <duration>
-    10 
+    24 
   </duration>
   <boardItems>
     <item>
--- a/toonz/sources/toonz/boardsettingspopup.cpp
+++ b/toonz/sources/toonz/boardsettingspopup.cpp
@@ -1067,7 +1067,7 @@ void BoardSettingsPopup::onSavePreset() {
 SaveBoardPresetFilePopup::SaveBoardPresetFilePopup()
     : GenericSaveFilePopup(tr("Save Clapperboard Settings As Preset")) {
   m_browser->enableGlobalSelection(false);
-  setFilterTypes(QStringList("board"));
+  setFilterTypes(QStringList("clapperboard"));
 }
 
 void SaveBoardPresetFilePopup::showEvent(QShowEvent* e) {
@@ -1080,7 +1080,7 @@ void SaveBoardPresetFilePopup::showEvent(QShowEvent* e) {
 LoadBoardPresetFilePopup::LoadBoardPresetFilePopup()
     : GenericLoadFilePopup(tr("Load Clapperboard Settings Preset")) {
   m_browser->enableGlobalSelection(false);
-  setFilterTypes(QStringList("board"));
+  setFilterTypes(QStringList("clapperboard"));
 }
 
 void LoadBoardPresetFilePopup::showEvent(QShowEvent* e) {


RodneyBaker:
Awesomely awesome.  You are just full of surprises!

RodneyBaker:
Reporting back
- No problem with saving and loading clapperboard presets.
- I must assume the example files will be provided later after full merge of the PR.

The one problem then is the Image Item:
I can't seem to get an image to appear testing with a number of different file formats.  I will guess this could related to images that require ffmpeg and I'll test against that.  I assume this has to be a single image (not a sequence) and I ran into a problem where the loader/browser was only identifying image sequences so that distracted me.   I'll try with BMP etc.

shun-iwasawa:
@RodneyBaker 
Thank you very much for a quick feedback on this.

I overlooked the possibility of specifying the sequential image path for the Image item. I'll try to fix it, probably by picking the first frame of the sequence.

Regarding the preset example, please put the `clapperboards` folder uploaded below under `OpenToonz 1.2 stuff\library` when you test via Appveyor build.
[clapperboards.zip](https://github.com/opentoonz/opentoonz/files/2440688/clapperboards.zip)

FYI, on dragging item rectangle you can press Alt key for resizing symmetrically and press Shift key for aligning on discrete positions.

RodneyBaker:
All appears to be working now.  Thank you!
[x] No crashes.
[x] Everything working.

A few random thoughts outside the immediate scope of this feature:
Boards
I love the format of the (text based) board files themselves.  Wow.Wow.Wow.  :)
And... you've now implemented editable text capability in OpenToonz.  Powerful.
I don't know if we (as users) realize how incredible and useful this is.

The Clapperboard obviously places the board at the front of a shot/sequence but I look forward to a time when we might be able to target the insertion point of a board at any point in the sequence.  This seems to be a natural progression for this feature and so for now I'll just admire the framework you've set in place.  The interim workaround that enables this already is to simply render the ClapperBoard sequences as separate scenes and composite them together in a master scene.  So we can do this now already.   We just need to plan ahead and execute accordingly.   

Dockable Output Settings
This might be a good time to mention this as you've just plussed up the usefulness of Output Settings.
I often find myself wanting to dock the Output Settings popup and keep it handy in a room. 
This obviously is low priority but a workflow enhancement.

RodneyBaker:
An option to render one of these boards out even if a Scene is empty would be useful.
Here we run afoul of OpenToonz current (and logical) error that when we attempt to render an empty seen we are reminded that there isn't anything in the scene.  A means to bypass that restriction would be required in order to set up Scenes with only Clapperboards as their content.

Workaround:  Create/expose one empty frame in the scene to allow it to render.

gab3d:
As @Rodney suggests (but not directly mentions), this is indeed the basis for a video Titler inside OT.

If we could create a "Title Level" and be able to see it in the Scene Cast window, we could put that (single frame) level in the Xsheet wherever we need to, and make it be exposed for as much frames as we need. That'd be just great! ✌️

shun-iwasawa:
@RodneyBaker 
> Conversely, it's a lot easier to search and segregate for 'clapperboard' than 'board' so perhaps clapperboard is to be preferred.

Indeed. I'll change the extension of the preset file from `.board` to `.clapperboard` . Thanks!

For now the clapperboard can be rendered only with the movie formats. I'm thinking of expanding it to the sequential image files in the future PR. For the sequential images, the clapperboard will be saved always one frame and the file name will be with user specified suffix instead of frame number.
For instance, when rendering a sequence `A.0001.png` , `A.0002.png` ... , the clapperboard will be saved with name like `A_clapperboard.png` .

@gab3d 
>If we could create a "Title Level" and be able to see it in the Scene Cast window, we could put that (single frame) level in the Xsheet wherever we need to, and make it be exposed for as much frames as we need. 

I'm thinking of developing a new fx which can render text... It would be better to have an option to get the text from a user-specified Note Level column somehow.

manongjohn:
@shun-iwasawa 

> I'm thinking of developing a new fx which can render text... It would be better to have an option to get the text from a user-specified Note Level column somehow.

Funny that you mention that.  I actually started slowly working on a "subtitle" Fx that would use the Note Levels as a source of the text.  I have it somewhat working but it has some issues and still needs a bit of work.

shun-iwasawa:
@manongjohn Wow, great!

RodneyBaker:
@shun-iwasawa  @manongjohn 

This recent release into open source of  OpenTimelineIO by Pixar may be of interest regarding some of your commentary here related to notes, subtitles, etc.

The scope is rather large compared to consideration in OpenToonz but it'd be good to keep this on the radar.

github:
https://github.com/PixarAnimationStudios/OpenTimelineIO

Google Group:
https://groups.google.com/forum/#!forum/open-timeline-io

artisteacher:
I had a crash on MacOS when trying to use fast render to mp4 with clapperboard on. Second try it went through ok. If I can find reproducible steps, I'll open an issue.
[clapperboard_render_crash.txt](https://github.com/opentoonz/opentoonz/files/2484444/clapperboard_render_crash.txt)

gab3d:
Just for reference...
I've added a manual section for the Clapperboard in the manual, [here](https://opentoonz.readthedocs.io/en/latest/rendering_the_animation.html#using-the-clapperboard).
I've used Shun's original description and screenshot as the basis.
If there's anything that could be improved, I'm all ears...

I've also reshuffled a bit the options of the Render Settings, to be displayed in the manual in the same order as in the UI.

PR #2316, opened 2018-10-04
Fix Output Settings Preset

### LINKED ISSUE #2315: Output Settings Preset Fails to Load Properly
### Issue Summary
Loading the output settings preset by using the combo box located at the top-right corner of the output settings popup fails to load the resample balance and the channel width.

### Steps to Reproduce
- Open the output settings popup.
- Save the current settings as the preset.
- Alter the settings.
- Load the preset.

### Expected Results
Settings should be the same as the one recorded in the preset.

### Actual Results
The resample balance and the channel width fail to be loaded. 

### System Information
* **Operating System**: win7

shun-iwasawa:
This will fix #2315 .

Also fixed the preset to save the resample balance and the channel width values always in English (like `resquality = Standard` ) regardless of the current interface language, in order to make it available among users with different UI languages. Before this fix the preset was saved in the current UI language (like `resquality = "標準"` in Japanese UI).
Loading such values from the preset is now done in two steps, firstly trying to find from the English keywords and secondly from the current UI texts. The second search was kept in order to maintain backward compatibility with the older versions of OT.

### DIFF
--- a/toonz/sources/toonz/outputsettingspopup.cpp
+++ b/toonz/sources/toonz/outputsettingspopup.cpp
@@ -43,12 +43,13 @@
 #include <QMainWindow>
 #include <QFrame>
 #include <QMessageBox>
+#include <QMap>
 
 //-----------------------------------------------------------------------------
 namespace {
 
 enum ResampleOption {
-  c_standard,
+  c_standard = 0,
   c_improved,
   c_high,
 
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
Jenkins

RodneyBaker:
I'm not sure if everything is working with the Output Presets.
According to the Preset Tooltip:

outputpreset

I'm not seeing the 'Save in' setting get saved.

Also, I understand that by design currently the Filename doesn't get saved in the preset although I'm not sure why it wouldn't be.  I can see cases where having it saved as well as not having it saved would both be useful.

RodneyBaker:
Update to last.

The "Save In" or Project Folder field for the Preset does appear to get saved sometimes although rarely from what I can tell.

It seems like the location is being saved in some place that overwrites changes to the project location.
Once changed it stays the same for all presets and is not stored in the Preset file.
The exception to this appears to be when the location is the output directory of the sandbox.
That output entry appears to be the only one that gets added to the present file on disk.

Extract from file:

  <savein_projectfolder>
    "+outputs"
  </savein_projectfolder>

PR #2325, opened 2018-10-12
Change Temporary File Name Suffix

### LINKED ISSUE #2306: Importing an image renames it incorrectly
### Issue Summary
Drag an image that isn't in your project's folder from a Windows Explorer window into OpenToonz (I can't remember if I dropped it onto the timeline, xsheet or the table) and it'll ask if you want to import or load. Choose load. The image appears to load correctly and you can work with it, but if you restart OpenToonz, the image column text is shown in red and the image isn't visible on the table.

The reason this happens is the image copied in the Extras folder has been renamed with an appended underscore.
E.g. If you drag in: **"image.png"**
The .tnz file, it references: **"+extras\\image.png"**
but on disk, in the Extras folder is an image named: **image_.png**

You can fix it by closing OpenToonz, renaming the files (just delete the underscore) then restarting OpenToonz.

Note, this doesn't happen every time. I just tried it with 1 more file, to upload a video reference of the issue and and it worked fine.

image

### System Information
I'm on Windows10 using the latest OTX version.

shun-iwasawa:
This will hopefully fix #2306 .

When saving the raster images, OT firstly save them into temporary files.
The temporary names has some suffix (like `image__1.png` for temporary file on saving `image.png` ) and it seemed to be misunderstood as sequential files under the current file name rules. 

So I changed the suffix like `image_ottmp1.png` in order to avoid such misunderstanding.

### DIFF
--- a/toonz/sources/toonzlib/levelupdater.cpp
+++ b/toonz/sources/toonzlib/levelupdater.cpp
@@ -342,7 +342,9 @@ TFilePath LevelUpdater::getNewTemporaryFilePath(const TFilePath &fp) {
   int count = 1;
 
   for (;;) {
-    fp2 = fp.withName(fp.getWideName() + L"__" + std::to_wstring(count++));
+    // changed the temporary name as the previous naming (like
+    // "filename__1.png") had been misteken as sequential images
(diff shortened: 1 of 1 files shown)


masafumi-inoue:
LGTM

PR #2330, opened 2018-10-15
Fix Color Model Problem on High DPI Monitor

shun-iwasawa:

OT fails to properly pickup the style at the clicked position in the color model window if it runs on the high dpi monitor (like Retina display) environment.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2334, opened 2018-10-15
Check Folder Access on Saving Level

shun-iwasawa:

When overwriting level, if by any chance the destination path cannot be accessed (for instance, because of unanticipated disconnection from file server), then stop the saving procedure with showing the warning popup in order to allow users to save it to another available location.
Before this change, such exception caused removing of image cache without any notification to users.


shun-iwasawa:
Jenkins

RodneyBaker:
A welcome addition.  Thanks!

PR #2340, opened 2018-10-23
Made small pump tool sizes actually do something.

turtletooth:
Right now the pump tool doesn't allow for precise tapering on small objects.  This line:

`double toolSize         = std::max(m_toolSize.getValue(), 5.0);
`
takes whichever is bigger of 5 or the pump toolsize.  So anything below 5 is treated as 5.  This makes it impossible to tweak the taper of small lines.  The tool is just so big it changes the width of the whole line.  I ran into this issue in my last animation.

### DIFF
--- a/toonz/sources/tnztools/pumptool.cpp
+++ b/toonz/sources/tnztools/pumptool.cpp
@@ -546,7 +546,7 @@ void PumpTool::onLeave() {
 //*****************************************************************************
 
 double PumpTool::actionRadius(double strokeLength) {
-  double toolSize         = std::max(m_toolSize.getValue(), 1.0);
+  double toolSize         = m_toolSize.getValue();
   double toolPercent      = toolSize * 0.01;
   double interpolationVal = pow(toolPercent, 5);
   double indipendentValue = 7.0 * toolSize;
--- a/toonz/sources/tnztools/pumptool.cpp
+++ b/toonz/sources/tnztools/pumptool.cpp
@@ -546,7 +546,7 @@ void PumpTool::onLeave() {
 //*****************************************************************************
 
 double PumpTool::actionRadius(double strokeLength) {
-  double toolSize         = std::max(m_toolSize.getValue(), 5.0);
+  double toolSize         = std::max(m_toolSize.getValue(), 1.0);
   double toolPercent      = toolSize * 0.01;
   double interpolationVal = pow(toolPercent, 5);
   double indipendentValue = 7.0 * toolSize;


RodneyBaker:
LGTM  Looks very good!

It's great to have you back Jeremy!

The control now is pretty impressive.  I even sculpted this almost circle with the Pump Tool:

pumptoolcircle

RodneyBaker:
What @openanim is referring to might almost be considered a bug in the Pump Tool's performance.
The Pump Tool can drive the thickness of a line down to zero thickness, which is very useful.
Where the problem occurs is that as soon as any part of the line is made thicker the entire line gets at least a value of 1 whereas ideally it would remain locked at a thickness of zero until that area is also influenced otherwise.

Allowing that thickness to remain zero and not pop to 1 unit of thickness would allow for drawing styles where colors fill an area but the area appears to have gaps.

This likely is another PR but to replicate this.
- Draw a line and increase it's line thickness (with the pump Tool or via manual entry.
- Use Pump Tool icon to decrease thickness to zero (which OT will readily do)
- Attempt to make only one part of the line's thickness larger while leaving another part at zero thickness
- Note that if any part of the line has thickness the entire line has at least a minimal thickness greater than zero.

Thanks!

RodneyBaker:
For testing purposes...
Anyone on Windows wishing to use this update can download the Appveyor release.
Direct link:

https://ci.appveyor.com/api/buildjobs/iuv4je8fkgv6hgi9/artifacts/toonz%2Fx64%2FOpenToonz_1.2.zip

To run:
Download and extract contents of the zip file.
The program will use your current Stuff directory.

openanim:
@RodneyBaker  Yes, and I'll also add that it will dramatically improve the workflow of vector techniques by acting as a hide tool and preserve interpolation process:

- more clarity
- faster
- easier
- less headache

for tweening, rigging/cut out of bodyturns, headturns, totally reusable.

If you wish to re-use your interpolation. The workaround that you proposed is not usable in most of case, it would be much too long. The destructive tools in OT like the eraser or the cutter tool work with the rule give and take, in  tweening process, what you remove must be done with an equivalence relationship. for small adjustments, it may be appropriate otherwise, not.
If we get the pump tool as suggested, we can mostly overcome theses weaknesses.

RodneyBaker:
> The workaround that you proposed is not usable in most of case, it would be much too long. The destructive tools in OT like the eraser or the cutter tool work with the rule give and take, in tweening process, what you remove must be done with an equivalence relationship. for small adjustments, it may be appropriate otherwise, not.

There's a reason they are called a workaround.
The workaround usually goes the long way around.  ;)

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @turtletooth for the modification!

@openanim @RodneyBaker For now rendering of the vector stroke seems to be interrupted only if all the control points are with zero thickness. ( [Here](https://github.com/opentoonz/opentoonz/blob/18a7c631a5941db6a89e34cb2aefe47e12953727/toonz/sources/common/tvrender/tglregions.cpp#L462-L467) in the source. ) And for now vector strokes are rendered in two ways - drawing outline of the stroke to obtain antialias and  filling the center of the stroke. ( [Here](https://github.com/opentoonz/opentoonz/blob/25d3db05fd57aed51d666455ac8dbc44067a4cea/toonz/sources/common/tvrender/tsimplecolorstyles.cpp#L466-L470) in the source. ) The problem is that the outline is rendered over the entire stroke, regardless of the thickness of the control point.
Anyway, as @RodneyBaker pointed, it can be a separate issue to be solved in another PR.

PR #2342, opened 2018-10-23
Duplicate in timeline

ghost:
This will allow the duplicate command to be used in the xsheet.  It only works if "Sync Level Drawing Number Changes with XSheet" is enabled.

With this you can hit D and duplicate a drawing from the xsheet or timeline and get a new copy automatically placed in the xsheet.  

I made it to work the way F6 does in Adobe Flash/Animate:
If you are on a drawing frame and the next frame is empty, it will insert a copy of the drawing in the next frame.
If you are on an empty frame, it will check if there is a drawing in the previous frame, and use that.
If you are in the middle of a held frame (e.g. drawing 2 is held for 15 frames), it will insert a copy on the frame that is selected as well as all sequential frames in the hold.

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -3254,6 +3254,7 @@ void CellArea::createCellMenu(QMenu &menu, bool isCellSelected, TXshCell cell) {
 
     menu.addAction(cmdManager->getAction(MI_Clear));
     menu.addAction(cmdManager->getAction(MI_Insert));
+    menu.addAction(cmdManager->getAction(MI_Duplicate));
     menu.addSeparator();
 
     TXshSimpleLevel *sl = TApp::instance()->getCurrentLevel()->getSimpleLevel();


RodneyBaker:
Works as advertised.
I really like how it's adding to the Level/Level Strip and not just extending an exposure.

Note:  I was thrown for a little while because I thought I had "Sync Level Drawing Number Changes with XSheet" turned on in Preferences when in fact I did not.  Turning that option on got it going as described.

LGTM

RodneyBaker:
There is one anomaly I observe.
Let's say I draw one image and duplicate that.
Then duplicate that.
Then go back to frame 2 and hit D.
Nothing happens.
If I extend the frame down one in xsheet then hit duplicate I recover and the frame is properly duplicated in it's correct place.

This isn't a problem but I guess my question is why some cells/frames don't duplicate.

turtletooth:
Cells will only duplicate if they are part of a held frame or if there is an empty frame.  It doesn't push frames over to duplicate a frame.

RodneyBaker:
Got it.  Thanks Jeremy!

711GM:
Add more F5 and F7. You are welcome! This is very useful for those who come after Flash. I was at first without these buttons was completely difficult.

RodneyBaker:
LGTM on WIndows.  This needs someone to validate on Mac and Linux.

shun-iwasawa:
Jenkins

toonz/sources/toonz/cellselection.cpp
```diff
@@ -1849,6 +1906,96 @@ void TCellSelection::pasteKeyframesInto() {
 
 //-----------------------------------------------------------------------------
 
+void TCellSelection::duplicateFrame() {
+  if (!Preferences::instance()->isSyncLevelRenumberWithXsheetEnabled()) return;
```
shun-iwasawa: How about opening some warning popup notifying that the preference option should be changed to use this command ?

RodneyBaker:
Often it's the little things that really help.
This will be a great workflow enhancement.

duplicateframes

Thanks much!

DarrenTAnims:
This is a great improvement, but just a question: Would it help flash users if the default shortcut was F6 ?

PR #2343, opened 2018-10-24
Allow Increase Step and Decrease Step to work even in selection is lost.

turtletooth:
Right now, we can't increase or decrease the step in the xsheet if the technical selection is lost.  There is still an active column and active frame, but just not a selection.  This PR makes that work even if there is only an active frame.  It also allows for decreasing the step even without having multiple frames selected.  It will detect if there are additional held frames and decrease the step using those.


shun-iwasawa:
I found a problem that `DecreaseStepUndo` is always stacked even if the Decrease Step command did not change the xsheet. It can be a separate issue.
LGTM Thank you @turtletooth for the modification !

PR #2344, opened 2018-10-24
Default and setting Interpolations

manongjohn:
This PR does the following with interpolations:

1) Adds more Default Interpolation options in Preferences

Requested by a few users, I added Constant, Exponential, Expression and File to the list of Default 
 Interpolations to choose from

image

2) Add ability to change Interpolation between 2 keyframes from xsheet/timeline

Added options to the context menu brought up when right-clicking the white line between 2 keyframes allowing you to change the type of interpolation being used. The current interpolation will not be shown.

image

In either case, use of Expression and File will still require going into the Function Editor to provide more information.

### DIFF
--- a/toonz/sources/toonz/xsheetcmd.cpp
+++ b/toonz/sources/toonz/xsheetcmd.cpp
@@ -1687,6 +1687,338 @@ public:
 
 } ResetArrowCommand;
 
+//-----------------------------------------------------------------------------
+
+class UseLinearInterpolation final : public MenuItemHandler {
+public:
+  UseLinearInterpolation() : MenuItemHandler(MI_UseLinearInterpolation) {}
+
+  void execute() override {
+    TApp *app    = TApp::instance();
+    TXsheet *xsh = app->getCurrentXsheet()->getXsheet();
+    int row      = app->getCurrentFrame()->getFrame();
+
+    TStageObjectId objectId = app->getCurrentObject()->getObjectId();
+    TStageObject *pegbar    = xsh->getStageObject(objectId);
+    if (!pegbar) return;
+
+    int r0, r1;
+    double ease0, ease1;
+
+    pegbar->getKeyframeSpan(row, r0, ease0, r1, ease1);
+
+    KeyFrameHandleCommandUndo *undo =
+        new KeyFrameHandleCommandUndo(objectId, r0, r1);
+
+    TStageObject::Keyframe k0 = pegbar->getKeyframe(r0);
+    TStageObject::Keyframe k1 = pegbar->getKeyframe(r1);
+
+    for (int i = 0; i < TStageObject::T_ChannelCount; i++) {
+      k0.m_channels[i].m_type     = TDoubleKeyframe::Linear;
+      k1.m_channels[i].m_prevType = TDoubleKeyframe::Linear;
+    }
+    pegbar->setKeyframeWithoutUndo(r0, k0);
+    pegbar->setKeyframeWithoutUndo(r1, k1);
+
+    TUndoManager::manager()->add(undo);
+
+    TApp::instance()->getCurrentScene()->setDirtyFlag(true);
+    TApp::instance()->getCurrentObject()->notifyObjectIdChanged(false);
+  }
+
+} UseLinearInterpolation;
+
+//-----------------------------------------------------------------------------
+
+class UseSpeedInOutInterpolation final : public MenuItemHandler {
+public:
+  UseSpeedInOutInterpolation()
+      : MenuItemHandler(MI_UseSpeedInOutInterpolation) {}
+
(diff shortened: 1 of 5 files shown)


711GM:
Let's make such keys. image So do in After Effects.

RodneyBaker:
LGTM
This will make a lot of OpenToonz users happy.

I'd say the Expression and File defaults are a WIP for the present.  Perhaps during the next round of updates a default value (for expression) and file reference (for File) might be supplied for those options.  They are very useful and powerful.

I'm tempted to suggest that (in the future round of enhancement) the default expression might simply be the number zero (0).  What this would do is lock everything in place in that  channel (until/unless the expression was manually changed later.  It could be very useful in that everything placed on the canvas and keyed while the default was set to expression would automatically be locked in place. The way to be free from this constraint would be to simply delete the key.

Similarly, for File Interpolation an example text file could be referenced that contained a few handy curve values (to be determined) to allow for forward movemet, reverse movement, stepping, ramp up quickly, full rotations (i.e. values in increments of 360), etc.

Thank you much for this addition!

toonz/sources/toonz/xsheetcmd.cpp
```diff
@@ -1687,6 +1687,338 @@ class ResetArrowCommand final : public MenuItemHandler {
 
 } ResetArrowCommand;
 
+//-----------------------------------------------------------------------------
+
+class UseLinearInterpolation final : public MenuItemHandler {
```
shun-iwasawa: It seems that the classes `UseXXXXXInterpolation` are almost the same, except that the command id and the interpolation types are different.
How about unify them like as follows for simplification? : 

```
// Unify commands for all types of interpolation
class SetInterpolation final : public MenuItemHandler {
  TDoubleKeyframe::Type m_type;
public:
  SetInterpolation(CommandId cmdId, TDoubleKeyframe::Type type) :
    MenuItemHandler(cmdId), m_type(type) {}

  void execute() override {
    TApp *app = TApp::instance();
    TXsheet *xsh = app->getCurrentXsheet()->getXsheet();
    int row = app->getCurrentFrame()->getFrame();

    TStageObjectId objectId = app->getCurrentObject()->getObjectId();
    TStageObject *pegbar = xsh->getStageObject(objectId);
    if (!pegbar) return;

    int r0, r1;
    double ease0, ease1;

    pegbar->getKeyframeSpan(row, r0, ease0, r1, ease1);

    KeyFrameHandleCommandUndo *undo =
      new KeyFrameHandleCommandUndo(objectId, r0, r1);

    TStageObject::Keyframe k0 = pegbar->getKeyframe(r0);
    TStageObject::Keyframe k1 = pegbar->getKeyframe(r1);

    for (int i = 0; i < TStageObject::T_ChannelCount; i++) {
      k0.m_channels[i].m_type = m_type;
      k1.m_channels[i].m_prevType = m_type;
    }
    pegbar->setKeyframeWithoutUndo(r0, k0);
    pegbar->setKeyframeWithoutUndo(r1, k1);

    TUndoManager::manager()->add(undo);

    TApp::instance()->getCurrentScene()->setDirtyFlag(true);
    TApp::instance()->getCurrentObject()->notifyObjectIdChanged(false);
  }

} UseLinearInterpolation(MI_UseLinearInterpolation, TDoubleKeyframe::Linear)
, UseSpeedInOutInterpolation(MI_UseSpeedInOutInterpolation, TDoubleKeyframe::SpeedInOut)
, UseEaseInOutInterpolation(MI_UseEaseInOutInterpolation, TDoubleKeyframe::EaseInOut)
, UseEaseInOutPctInterpolation(MI_UseEaseInOutPctInterpolation, TDoubleKeyframe::EaseInOutPercentage)
, UseExponentialInterpolation(MI_UseExponentialInterpolation, TDoubleKeyframe::Exponential)
, UseExpressionInterpolation(MI_UseExpressionInterpolation, TDoubleKeyframe::Expression)
, UseFileInterpolation(MI_UseFileInterpolation, TDoubleKeyframe::File)
, UseConstantInterpolation(MI_UseConstantInterpolation, TDoubleKeyframe::Constant);
```

manongjohn: Done.  Thanks for providing the replacement logic.

gab3d:
I'm not really sure about including Expression and File as default options. 
I think it may lead to confussion specially if no input field is shown below to input default values for those, right there in the Preferences panel.

Anyway, what people were really asking for was just to include Constant as an option.

shun-iwasawa:
Jenkins

shun-iwasawa:
Regarding whether to hide the Expression and the File interpolation options, let's leave a decision to the future user discussion after releasing RC version.
LGTM at the moment. Thank you @manongjohn for the addition!

PR #2352, opened 2018-10-31
Fix Toonz Raster Tape Tool Crash

shun-iwasawa:
This PR fixes several problems related to the style number field of Tape Tool option as follows:

- Under non-English UI and with no level being selected, OT crashes when selecting the tape tool.
- If no level is selected, OT crashes when typing number in the style number field.
- Under non-English language UI, the style number field does not update when it is set to `current` .
- You cannot type the word `current` in the style number field.

**UPDATE:**
Please note that the problematic `Style index` field appears only with the tape tool for Toonz raster levels. @turtletooth thank you for pointing that!

### DIFF
--- a/toonz/sources/include/toonzqt/styleindexlineedit.h
+++ b/toonz/sources/include/toonzqt/styleindexlineedit.h
@@ -25,6 +25,8 @@ class TPaletteHandle;
 namespace DVGui {
 
 class DVAPI StyleIndexLineEdit : public LineEdit {
+  Q_OBJECT
+
   TPaletteHandle *m_pltHandle;
 
 public:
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
Jenkins

turtletooth:
> If no level is selected, OT crashes when typing number in the style number field.
@shun-iwasawa I can't find where to test entering the style number in the tape tool options.

shun-iwasawa:
@turtletooth Modified the PR comment. Thank you for pointing that!

turtletooth:
LGTM.  Seems to work fine.  Thanks @shun-iwasawa for the fix.

PR #2362, opened 2018-11-04
Fix restoring raster tool crash

### LINKED ISSUE #2361: Raster level crashes when switching back to brush
### Issue Summary
A Discord user reported an issue that when she was drawing in the raster level, switched to pan tool temporarily (holding space key) and then releasing to draw again would crash OT.

I was able to confirm this only happens on Raster levels while using a pen.

### Steps to Reproduce
1) Open a new Raster level in OT
2) Draw something
3) Press and hold Hand tool shortcut (space)
4) Move canvas around
5) While still having the pen actively touching the screen, release the shortcut to return to brush tool
6) Continue drawing again without lifting the pen --> No longer draws and crashes

manongjohn:
This PR fixes #2361 

When returning to the Raster bush from temporary usage of Hand Tool (holding tool shortcut) while still pressing with the pen, the Raster brush is no longer initialized properly so the drag event tries to work with an uninitialized tool and crashes when you try to continue moving.

Similar to Toonz Raster (where it does not crash), I modified to block the drag event from processing if not properly initialized.

### DIFF
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -350,6 +350,8 @@ void FullColorBrushTool::leftButtonDrag(const TPointD &pos,
   TRasterImageP ri        = (TRasterImageP)getImage(true);
   if (!ri) return;
 
+  if (!m_toonz_brush) return;
+
   TRasterP ras      = ri->getRaster();
   TPointD rasCenter = ras->getCenterD();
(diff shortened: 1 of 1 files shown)


turtletooth:
LGTM.  Thanks @manongjohn

turtletooth:
@shun-iwasawa It's a pretty simple change.  Would you like to call Jenkins on this?

shun-iwasawa:
Jenkins

PR #2365, opened 2018-11-06
Fix Hide/Lock shortcut bug

### LINKED ISSUE #2323: Hiding a column with hotkeys does not block the action of the tools (gif).
### Issue Summary

Hiding a column with hotkeys does not block the action of the tools, and sometimes confuses when they should be applied.

ezgif-1-120c1dbc7621

### Steps to Reproduce

Try to switch the level display with hotkeys (Hide Selected, Show All). It does matter if the cursor is over the canvas, or over the Xsheet.

### Expected Results

The tools must be blocked
### Actual Results

The tools is not blocked
### System Information

manongjohn:
This PR fixes #2323

When the Hide/Show and Lock/Unlock shortcuts are used, this forces the current tool to be checked to see if it can still used on the canvas or not.

### DIFF
--- a/toonz/sources/toonz/columncommand.cpp
+++ b/toonz/sources/toonz/columncommand.cpp
@@ -45,6 +45,8 @@
 #include "tstroke.h"
 #include "tundo.h"
 
+#include "tools/toolhandle.h"
+
 // Qt includes
 #include <QApplication>
 #include <QClipboard>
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed the fix, thank you @manongjohn ! LGTM
Currently OT updates the tool state when the mouse cursor enters the viewer. It seems to be not a good choice considering that the shortcut key can alter many aspects of the software and the scene data with staying the mouse cursor in the viewer as written in the issue #2323.

PR #2366, opened 2018-11-06
Remove mysettings.ini

shun-iwasawa:
This PR will remove `mysettings.ini` from user profiles for simplification.

Settings formerly stored in `mysettings.ini` are now moved to the following:

- Window geometry of the floating sheet / curve window of the Function Editor is stored in `popup.ini` .
- Chip size of the palette is stored in the room settings file ( like `room1.ini` ), with an id `viewtype`.
- Visible parts (toolbar / tool options bar / console) information of the comboviewer is also stored in the room settings file, with an id `visibleParts` .

This PR also unify the behavior of the Level Strip window between `Default` and `StudioGhibli` room layouts. Now the navigator (red rectangle appears in the current frame in the Level Strip) always shows the visible region of the focused viewer even if in `StudioGhibli` room layout.


shun-iwasawa:
Jenkins

shun-iwasawa:
Resolved conflicts. Jenkins again

PR #2369, opened 2018-11-06
Fix raster crash after unhiding column

### LINKED ISSUE #2311: Crash on hidden column
### Issue Summary

When inadvertently I draw on a hidden column, after several strokes or sometimes only one is enough, OT crash.

### Steps to Reproduce

- Create severals columns
- Hide 1 column
- then try to draw something on hidden column

### Expected Results

no crash

OT Last NB

Windows 10 8g

manongjohn:
This PR fixes #2311

When attempting to draw immediately after unhiding a column, a button-up event is encountered but the raster bush is not initialized properly so it crashes when trying to access certain variables.

Similar to Toonz Raster (where it does not crash), I modified to block the button-up event from processing if not properly initialized.


RodneyBaker:
That's done it!  Thanks for sticking with this one all the way to the finish line.

shun-iwasawa:

PR #2374, opened 2018-11-08
Fix wrong line cap loading

### LINKED ISSUE #2373: Wrong line end caps used when reloading scene
### Issue Summary
Sometimes when loading a scene, the vector lines with rounded end caps are restored with flat end caps.

This happens when lines are drawn after a geometry tool object is drawn and the geometry tool's caps do not match the brush tool caps.

### Steps to Reproduce
1) Create a new vector level
2) Select Brush tool and make sure end cap is rounded
3) Draw a line
4) Switch to Geometry tool and make sure end cap is flat
5) Draw a circle
6) Select Brush tool
7) Draw a line
8) Save scene and close OT
9) Open OT and load the scene

The second line will be drawn with the flat end caps

NOTE: This may be the cause for #1269.

manongjohn:
This PR fixes #2373 

This is actually a saving issue. When saving the vector strokes to the file, it writes tagging information showing the next stroke has non-standard outline information, but does not write instructions to reset it back afterwards.  Without resetting it, as the file is loaded the non-standard outline information gets applied to every line afterwards until it either encounters another non-standard outline tag or finishes processing all the strokes.

When non-standard outline information is detected,  this writes out tagging information to reset the outline information.

### DIFF
--- a/toonz/sources/image/pli/tiio_pli.cpp
+++ b/toonz/sources/image/pli/tiio_pli.cpp
@@ -457,6 +457,7 @@ TImageWriterPli::TImageWriterPli(const TFilePath &f, const TFrameId &frameId,
 static void putStroke(TStroke *stroke, int &currStyleId,
                       std::vector<PliObjectTag *> &tags) {
   double maxThickness = 0;
+  bool nonStdOutline  = false;
   assert(stroke);
 
   int chunkCount = stroke->getChunkCount();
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed fix. Thank you @manongjohn ! LGTM

PR #2375, opened 2018-11-09
Fix Current Frame on Load Level

### LINKED ISSUE #2135: Automatic switching between layers.
### Issue Summary
Very annoying when you upload images and are already going to animate and then you will find out that you are now col2 instead of col1 this is very confusing. Can it be removed?  

  https://youtu.be/3-0wQDzHNgY

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/60459875-automatic-switching-between-layers?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

manongjohn:
This PR fixes #2135

When loading a level file into the xsheet/timeline, the current frame will switch to the first frame of the newly loaded level.  This is true whether the file was drag/dropped or loaded through menu items.

Note: When dragging/dropping multiple levels at once, the current frame will be set to the first frame of the **last** level loaded into the xsheet/timeline.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -2433,10 +2433,6 @@ int IoCmd::loadResources(LoadResourceArguments &args, bool updateRecentFile,
         // increment the number of loaded resources
         ++loadedCount;
 
-        // move the current column to right
-        col0++;
-        app->getCurrentColumn()->setColumnIndex(col0);
-
         // load the image data of all frames to cache at the beginning
         if (cachingBehavior != ON_DEMAND) {
           TXshSimpleLevel *simpleLevel = xl->getSimpleLevel();
@@ -2449,10 +2445,7 @@ int IoCmd::loadResources(LoadResourceArguments &args, bool updateRecentFile,
     }
   }
 
-  if (loadedCount) {
-    app->getCurrentFrame()->setFrameIndex(row0);
-    app->getCurrentColumn()->setColumnIndex(col0 - 1);
-  }
+  if (loadedCount) app->getCurrentFrame()->setFrameIndex(row0);
 
   sb->data().m_loadedCount += loadedCount;
   sb->data().m_hasSoundLevel = sb->data().m_hasSoundLevel || isSoundLevel;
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
@manongjohn 
Thank you for tackling with this issue.
And so sorry for my late comment on it - but actually I included this behavior when I developed Toonz Ghibli Edition. As far as I can remember it was because the old version of Toonz always put the loaded level at the current row/col even if the cell is occupied by another level. For now the problem has been modified and the patch seems to be unnecessary. 

Please try removing [lines 2436 - 2439 of iocommand.cpp](https://github.com/opentoonz/opentoonz/blob/2986f2e399afd94c15e85e7b5f36eeef7eeb9f40/toonz/sources/toonz/iocommand.cpp#L2436-L2439) like as follows and check if it can get the same behavior as your PR. Sorry for the trouble!

```
---------------------- toonz/sources/toonz/iocommand.cpp ----------------------
index 2ac4e62..cbdd37c 100644
@@ -2433,10 +2433,6 @@ int IoCmd::loadResources(LoadResourceArguments &args, bool updateRecentFile,
         // increment the number of loaded resources
         ++loadedCount;
 
-        // move the current column to right
-        col0++;
-        app->getCurrentColumn()->setColumnIndex(col0);
-
         // load the image data of all frames to cache at the beginning
         if (cachingBehavior != ON_DEMAND) {
           TXshSimpleLevel *simpleLevel = xl->getSimpleLevel();
```

manongjohn:
@shun-iwasawa No problem. 

I just compared my change's behavior vs your suggestion.  The only behavioral difference between my change and yours is that mine will change the current frame to the frame where the last file was dropped.

I will update the PR to switch my change to yours and include the current frame change.

shun-iwasawa:
Jenkins again

shun-iwasawa:

PR #2382, opened 2018-11-16
Fix processing touchEvents after tablet events

### LINKED ISSUE #2220: pen behavior Windows ink
Condition after implementing Windows ink (build 2018-08-07)
on my Windows 10 Surface Pro 2017 i5 8GB with Surface pen (or Wacom Bamboo ink pen) it has the following flaw:

with Windows ink enabled, the pen now draws like it should, without moving the canvas at the same time, but:
switching from the brush tool to the Hand tool (or Animate tool) and back to the brush tool the pen moves the canvas around while drawing. This behavior stops, when i move the pen away from the screen for lets say 20centimeters. Moving pen back to the screen now draws the brush without moving the canvas.

manongjohn:
This PR fixes #2220 

Sometimes when quickly switching to and using a tool with a pen, the tool winds up moving the canvas rather than what the tool should be doing, like drawing or selecting. This is because sometimes a TouchEvent immediately follows a TabletPress event and takes over operation acting like the Hand tool.

Similar to mousePress events, if a begin TouchEvent event is encountered while a Tablet event is active, the TouchEvent will be ignored.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -966,6 +966,8 @@ void SceneViewer::gestureEvent(QGestureEvent *e) {
 
 void SceneViewer::touchEvent(QTouchEvent *e, int type) {
   if (type == QEvent::TouchBegin) {
+    if (m_tabletEvent) return;
+
     m_touchActive   = true;
     m_firstPanPoint = e->touchPoints().at(0).pos();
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Although I still could not reproduce the issue on my environment, I confirmed that at least this fail-safe will not cause any bad side effect. Thank you @manongjohn for the fix! LGTM

PR #2389, opened 2018-11-22
Fix restoring eraser tool crash

manongjohn:
This is similar to PR #2362 but with the eraser tool.

When returning to the Raster Eraser from temporary usage of Hand Tool (holding tool shortcut) while still pressing with the pen, the Raster eraser is no longer initialized properly so the drag event tries to work with an uninitialized tool and crashes when you try to continue moving.

Similar to Toonz Raster Eraser (where it does not crash), I modified to block the drag event from processing if not properly initialized.


shun-iwasawa:
Jenkins

shun-iwasawa:

PR #2392, opened 2018-11-28
Fix moving note cells crash

manongjohn:
This fixes a crash that can happen if you attempt to select all the cells on a Note Level and attempt to drag it to an empty column (prevented) and back again to where it was


shun-iwasawa:
@manongjohn 
Unfortunately I couldn't reproduce the crash on the current master branch yet... 

move_note_level

Can you please let me know if I'm missing something?

manongjohn:
@shun-iwasawa I can still reproduce this on 2 different machines (Win7 and Win10) using nightly build 2018-12-20.

artisteacher:
I'm not getting a crash in MacOS.

manongjohn:
Ok.  I figured out this happens if you have Xsheet -> Cell Dragging Behavior = Cells and Column Data.

If it is set to Cells only, this does not happen.

shun-iwasawa:
LGTM

PR #2393, opened 2018-11-28
New Fx: Text Fx Iwa

shun-iwasawa:
This PR will add a new Fx for rendering text.

image

Here is a brief description of parameters of this Fx:

### Source
Specifies the text source. You can choose one from the below:

- **Nearby Note Column** : Get the text from note level column put on the directly left side of the Text Iwa Fx Column in the Xsheet (For Timeline view, the Note Level layer just under the Fx Column is used). If the column does not contain Note Level or the cell is empty, text will not be rendered.
image

- **Specified Note Column** : It's similar to the "Nearby Note Column" option, but the Note Level Column can be specified with the **Column Index** field below.

- **Input Text** : The text typed in the "Text" field below will be used (as shown in the image on top of this post).

### Column Index
Specifies the source Note Level column index. Only used in "Specified Note Column" Source mode.

### Text
Specifies the source text. Only used in "Input Text" Source mode.

### Horizontal Align
Specifies the text align (Left / Right / Center / Justify).

### Center, Width and Height
Specifies the text box. Just like the "Text" item of the clapperboard feature, the font size will be automatically adjusted in order to fit the box.

### Font
Specifies the font family, style and maximum size. Note that the size will take effect only if the automatically-calculated font size (in order to fit the text box) is bigger than the size value. (Therefore, it specifies the **maximum** font size.)

### Text Color
Specifies the text color. This will also be used for the border line.

### Box Color
Specifies the box background color.

### Show Border
When activated, the border line will be rendered around the text bounding box. Please note that the text box (specified by the user) and the text bounding box (computed from the rendered text) may become different, especially when the maximum font size is set to small value.

------

### Known Issue
When you are using a Note Level column for the source text, the change of the note text will not trigger the preview regeneration. You need to regenerate preview manually from the context menu of the viewer.

------

I hope this Fx will be utilized for making subtitle, credit or any other text objects.
@manongjohn I understand that you are also planning of (or already tackling with) developing this sort of Fx. Please feel free to make suggestions. Thank you!

### DIFF
--- a/stuff/profiles/layouts/fxs/fxs.lst
+++ b/stuff/profiles/layouts/fxs/fxs.lst
@@ -146,7 +146,8 @@
     STD_inopnCloudsFx
     STD_particlesFx 
     STD_iwa_TiledParticlesFx
-    STD_iwa_TimeCodeFx
+    STD_iwa_PNPerspectiveFx
+    STD_iwa_TextFx
   </Render>
   <Stylize>
     STD_colorEmbossFx
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1255,6 +1255,19 @@
   <item>"STD_iwa_BarrelDistortFx.vignetteMidpoint" "Vignetting Midpoint"</item>
   <item>"STD_iwa_BarrelDistortFx.scale" "Scale"</item>
 
+  <item>"STD_iwa_TextFx"	"Text Iwa"</item>
+  <item>"STD_iwa_TextFx.targetType" "Source"</item>
+  <item>"STD_iwa_TextFx.columnIndex" "Column Index"</item>
+  <item>"STD_iwa_TextFx.hAlign" "Holizontal Align"</item>
+  <item>"STD_iwa_TextFx.text" "Text"</item>
+  <item>"STD_iwa_TextFx.center" "Center"</item>
+  <item>"STD_iwa_TextFx.width" "Width"</item>
+  <item>"STD_iwa_TextFx.height" "Height"</item>
+  <item>"STD_iwa_TextFx.font" "Font"</item>
+  <item>"STD_iwa_TextFx.textColor" "Text Color"</item>
+  <item>"STD_iwa_TextFx.boxColor" "Box Color"</item>
+  <item>"STD_iwa_TextFx.showBorder" "Show Border"</item>
+
  <!------------------------------ Tiled Particles Iwa ------------------------------------------->
 
   <item>STD_iwa_TiledParticlesFx "Tiled Particles Iwa" </item>
--- a/stuff/profiles/layouts/fxs/STD_iwa_TextFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_iwa_TextFx.xml
@@ -1,0 +1,17 @@
+<fxlayout>
+	<page name="Text Iwa">
+ 	<vbox>
+		<control>targetType</control>
+		<control>columnIndex</control>
+		<control>text</control>
+		<control>hAlign</control>
+		<control>center</control>
+		<control>width</control>
+		<control>height</control>
+		<control>font</control>
+		<control>textColor</control>
+		<control>boxColor</control>
+		<control>showBorder</control>
+  	</vbox>
+  	</page>
+</fxlayout>
--- a/toonz/sources/common/tparam/tnotanimatableparam.cpp
+++ b/toonz/sources/common/tparam/tnotanimatableparam.cpp
@@ -8,6 +8,7 @@ PERSIST_IDENTIFIER(TBoolParam, "boolParam")
 PERSIST_IDENTIFIER(TFilePathParam, "filePathParam")
 PERSIST_IDENTIFIER(TStringParam, "stringParam")
 PERSIST_IDENTIFIER(TNADoubleParam, "naDoubleParam")
+PERSIST_IDENTIFIER(TFontParam, "fontParam")
 // PERSIST_IDENTIFIER(TIntEnumParam, "intEnumParam")
 
 TPersistDeclarationT<TEnumParam> TEnumParam::m_declaration("intEnumParam");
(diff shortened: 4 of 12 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
I forgot to update CMakeFiles.txt. Jenkins again

RodneyBaker:
Wonderful!  I'm not able to test the appveyor as this change adds content to the stuff directory... 
BUT... very excited about this.

manongjohn:
@shun-iwasawa This is great. I'm glad you didn't wait for my version to finish.  Not sure when I was going to get back to it.  Plus, this is probably a better implementation than I was doing, though I was not far from the changes you made. :)

The one difference I have was I added a dropdown to my fx properties page that was populated with Note Level column names which the user needed to select.  The one problem with this was if the Note Level column moved, the fx wouldn't update automatically since i was storing the Col # .  Your version will likely have the same issue if the user chose the Specified Note Column entry.  I would probably find the Nearby Note Column would be a more flexible option.

Just something I was thinking about recently was potentially adding a column option for Note Levels to show the column as a node in the Schematic viewer. Then fxs, like a subtitle/text fx could be easily attached to it.  Not sure how involved that would be or if that would make sense to do.  Potentially something that can be done in the future and then this fx could be retro fitted to work with it.

RodneyBaker:
Got the Appveyor to work by copying down the files needed.
Looking very good!
reachforthestars

RodneyBaker:
Two things thus far that don't quite appear to work as desired...

1.  Notelevels don't accept input of the letter z
This very likely is due to shortcut assignments (perhaps outside of OT?)

2.  After updating text in a notelevel the text on canvas doesn't update.  Saving refreshes this and the new text is displayed.

artisteacher:
It’s going to be so great having a way to include editable text! 

I don’t know if this is a problem on Windows, or just MacOS but in general I’ve found working with FX/preview to be more unstable than other areas of OT. Sometimes just opening the fx settings causes a crash. Various actions while preview is active or just turning on preview can cause a crash. It’s probably a harder issue to solve as the crashing can be inconsistent so there’s likely not time to resolve this before the next release. However as more and more helpful fx are added, it will be more & more important to improve stability of fx and preview.

shun-iwasawa:
Jenkins again

RodneyBaker:
Shun,
The main issue I have with this addition... which from what I can tell is working as you've specified... is that **we may need a way to bypass it**.  If it's not too much trouble can there be a checkbox to turn on/off or bypass?  It looks fine to have it on by default.

My main concern arises from formats that support having an alpha channel that can also not have an alpha.  I believe the primary example of that might be GIF that can be either still image or a set of sequential animated gifs.  As gifs don't have an alpha per se but tend to key out a specific color to achieve transparency it can be problematic to change that transparency.  It can also be very helpful and this PR should assist in creating animated gifs with transparency by dictating a specific color.  Nice!

At any rate, this being something in Preferences it might be appropriate to have an on/off switch.
Thanks for your work and consideration.

masafumi-inoue:
LGTM

RodneyBaker:
Shun's documentation in PDF format:
[Text FX.pdf](https://github.com/opentoonz/opentoonz/files/2694072/Text.FX.pdf)

RodneyBaker:
> You can temporary prevent it by removing the shortcut key for "Delete Key" command (which is Z key by default).

I can't find a shortcut for "Delete Key" in the shortcut listing.
I see an entry for "Delete" with a DEL assigned as the key.
I also don't see that key assigned in the INI file.
At any rate, I don't see any way to turn that off.
I was trying to use "Opentoonz" for as a text example so perhaps that needs to be revisited? ;)

artisteacher:
This is great! Now that the note level can be used as a source for the text, it would be great to have access to the some of the context menu for the note level for adjusting steps & etc.

artisteacher:
Should the timecode fx be gone now? I'm seeing a second occurrence of PN Perspective in it's place now...
screen shot 2018-12-21 at 10 23 08 am

RodneyBaker:
Good catch @artisteacher 
An error seems to have crept into /stuff/profiles/layouts/fxs/fxs.lst 

I'm trying to copy/paste the line in now for easy reference

stuff/profiles/layouts/fxs/fxs.lst
```diff
@@ -146,7 +146,8 @@
     STD_inopnCloudsFx
     STD_particlesFx 
     STD_iwa_TiledParticlesFx
-    STD_iwa_TimeCodeFx
+    STD_iwa_PNPerspectiveFx
```
RodneyBaker: Should still be STD_iwa_TimeCodeFx

shun-iwasawa: Thank you @artisteacher @RodneyBaker , I'll fix it right away.

RodneyBaker:
The shortcut key in question for Z appears to be 'Set Key'.
I confirm that assigning another shortcut key to 'Set Key' is a good workaround (although unfortunate for those that use lots of keyframing).

Perhaps we can explicitly escape shortcut keys while in a text field.
What is strange to me is that other single letter characters used with shortcuts/hotkeys don't appear to produce the same problem.

manongjohn:
Shortcuts that affect cells will likely have this problem when editting a cell to rename them or entering note level text.

I had argued in the past why OT shortcuts shouldn't be allowed when in cell edit mode, except standard text editing shortcuts like copy/cut/paste. It was ultimately dismissed because it would impact certain workflows currently used by some Japanese studios.

We will likely see other related reports as note levels become more useful.

RodneyBaker:
Just a quick note...

It had been awhile since I had set up a TextIwa FX and when I did the text was not appearing.
It just so happened that the Font I had selected was 'script'.
Apparently some of the Fonts listed will not work.
If I had previously known this... I had forgotten.
The fix/workaround is to select a different/appropriate Font.

I'm still enjoying this feature! :)

gab3d:
@shun-iwasawa 
...just noticed the UI says **Holizontal Align**, instead of **Horizontal Align** 😄

PR #2395, opened 2018-11-28
Fix empty column property issues

### LINKED ISSUE #2202: Problem with hidden column after deletion of frames on that column
### Issue Summary
Every time I create a new raster level (has not been tested with other levels) and then I make the column that contains that raster level invisible, then delete it. I cannot use the column again. I have to delete it and start over. Not a big issue but maybe a bug. 

opentoonz 1

### Steps to Reproduce
Create a new raster level in any column. Write something on the raster level. With the raster level still there, make the column preview disabled and make the camera view invisible by clicking the eyeball and the camera icons in the column header. Delete the raster level. Try to create a raster level again on the same column...unable. 

The raster level CAN be created by right clicking and using create raster level function. This does not work for auto-creation function when you touch the screen with the pen-tip. 

### System Information
* **OpenToonz Version**:  1.2.1   july 28 2018 build
* **CPU**: i7
* **Memory**: 8 Gig
* **Graphics Card**: intel intergrated
* **Graphics Tablet**: Wacom Cintiq 13

manongjohn:
This PR addresses some issues with Column properties (preview, visibility, opacity, color filter, etc) not being set/reset correctly when deleting or moving all cells in a column to an empty column.

1) This fixes #2202 

When a column becomes empty as a result of deleting or cutting all the cells within it, all column properties are reset to baseline.

NOTE: Undo of cell delete/cut will not restore the prior column properties.

2) Fixes issue with color filter disappearing when moving to a new empty column

When moving all cells to a new empty column, opacity and other button statuses and properties are kept, but the color filter is removed.  It will now copy the color filter to the empty column as well.  Impacts all Level column types and Mesh columns.

### DIFF
--- a/toonz/sources/include/toonz/txshcolumn.h
+++ b/toonz/sources/include/toonz/txshcolumn.h
@@ -265,6 +265,8 @@ Set column color tag to \b colorTag.
   TPixel32 getFilterColor();
   static QPair<QString, TPixel32> getFilterInfo(FilterColor key);
   static void initColorFilters();
+
+  void resetColumnProperties();
 };
 
 #ifdef _WIN32
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -119,6 +119,7 @@ void deleteCellsWithoutUndo(int &r0, int &c0, int &r1, int &c1) {
       xsh->clearCells(r0, c, r1 - r0 + 1);
       TXshColumn *column = xsh->getColumn(c);
       if (column && column->isEmpty()) {
+        column->resetColumnProperties();
         TFx *fx = column->getFx();
         if (fx) {
           int i;
@@ -144,6 +145,7 @@ void cutCellsWithoutUndo(int &r0, int &c0, int &r1, int &c1) {
     xsh->removeCells(r0, c, r1 - r0 + 1);
     TXshColumn *column = xsh->getColumn(c);
     if (column && column->isEmpty()) {
+      column->resetColumnProperties();
       TFx *fx = column->getFx();
       if (!fx) continue;
       int i;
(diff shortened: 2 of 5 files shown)


artisteacher:
@manongjohn thanks so much for working to perfect the small details that may seem unimportant on their own but together really add to a more polished, professional feel to OT. 

For number 2, if the column is moved with the column header, I’m seeing the color filter move too. However when moving the column by selecting all the cells with the side dragbar & moving that way, the color filter doesn’t move with the cells. I guess it might  make sense as you are moving only the cells and not the column header directly...

This is loosely related, but if the column is at the right edge of the program window (Xsheet view), the dropdown with the filter/opacity settings is mostly offscreen/inaccessible.

manongjohn:
@artisteacher For # 2, consider dragging all cells by the cell dragbar is essentially the same as moving the column with the column dragbar.

The current behavior already does this with the Preview, Camera and Opacity settinga so it made sense, to me at least, to fix it to also moves the color filte instead of losing it.

artisteacher:
I agree, I’m just not seeing that and thought it could be deliberate. Are you sure that those settings already stick with the column when moving all of the cells with the side dragbar? I’m not getting that behavior in MacOS - with or without your changes. I only see the settings stick when moving the column header.

manongjohn:
Yep.  Here is what is currently happening in the official OT version when you drag from cell dragbar:

cellcolmove

This doesn't happen on a Mac?

artisteacher:
I’m using a nightly build from November 5th & I definitely do not get that behavior on MacOS.  I’ll check at school tomorrow to see if anything is different with 1.2.1.

artisteacher:
No change with 1.2.1 - none of the column header settings stick with the moving cells in MacOS when moving with the cell dragbar.

shun-iwasawa:
> NOTE: Undo of cell delete/cut will not restore the prior column properties.

It would be better to make the column properties to be restored on undo, but it can be a separate future issue since currently all the operations changing column properties are not undo-able.

@artisteacher 
Please check if `Preferences > Xsheet > Cell-dragging Behavior` is set to `Cells and Column Data` just like the same condition as #2392 .

Jenkins

shun-iwasawa:
Jenkins again

shun-iwasawa:

PR #2396, opened 2018-11-28
Fix eraser mode switch crash

### LINKED ISSUE #2391: Crash switching Eraser Tool types in Vector Levels
### Issue Summary

If you erase with the Eraser Tool and at the same time hold the button that switches the types "Freehand" and "Normal" (i use F5 for this), then a crash occurs.

ezgif-5-72561d49df79

### Steps to Reproduce

1. Assign hotkeys to switch between "Normal" and "Freehand" types (for me is F5) Since the same button switches between two modes, by holding the key you start their cyclic selection. (Intuitively, it seems to me that while holding a button I want to temporarily switch to another type)

2. If you try to erase a vector drawing while holding the type switching hotkey, a crash will occur.

### Expected Results

Holding down the type toggle button should not switch types cyclically. I think that the same behavior would be appropriate here as when switching tools (when you press a button, you switch to the next type, while holding the button, temporarily switch to the next type) Well, the “hold button to switch types” looks more like a function request, I would like to hear other opinions on this issue.
### Actual Results

Crash
### System Information

manongjohn:
The PR fixes #2391

While actively erasing, if the user switches from Freehand/Rectangle/Polyline to Normal using a shortcut, internally the eraser tool is not set up properly to do an Undo in Normal mode and causes a crash when it tries to reference internal variables for it.

This fix now checks to see if it's in Normal mode and not properly setup for Normal undo and immediately initiates a new Normal mode setup (left button click).

### DIFF
--- a/toonz/sources/tnztools/vectorerasertool.cpp
+++ b/toonz/sources/tnztools/vectorerasertool.cpp
@@ -834,6 +834,7 @@ void EraserTool::leftButtonDrag(const TPointD &pos, const TMouseEvent &e) {
     invalidate();
     return;
   } else if (m_eraseType.getValue() == NORMAL_ERASE) {
+    if (!m_undo) leftButtonDown(pos, e);
     if (TVectorImageP vi = image) erase(vi, pos);
   } else if (m_eraseType.getValue() == FREEHAND_ERASE) {
     freehandDrag(pos);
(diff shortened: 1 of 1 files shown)


RodneyBaker:
LGTM Can't confirm on Mac/Linux

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed the crash resolved. Thank you @manongjohn LGTM

PR #2398, opened 2018-11-30
Fix incorrect keyframe cell pasting and redo

### LINKED ISSUE #2189: Enable redo after a copy key
### Issue Summary

I have not found a similar subject but it is possibly a duplicate.

There is a problem when trying to use redo after copying keys.

### Steps to Reproduce

1- Create a column with one or more levels
2- Create one or several keys
3- copy the keys
4- paste the keys on the levels below or on another column
5 - use undo
6- use redo.

It does not work.

W10 OT 1.2.1

### LINKED ISSUE #2242: Keyframes not pasted correctly
### Steps to Reproduce
Create 2 column with cells
Make some keyframe on the middle layer
Select all with Ctrl to select cell and keyframe
Ctrl C
select an empty frame Ctrl V
The frame are pasted on the right column 
But the keyframe are not pasted on the middle column but on the first column

### Expected Results
Keyframe should be pasted on the middle column

tempsnip

manongjohn:
This PR fixes #2242 

Added logic to offset where the keyframes need to be pasted based on the cell selection area that was copied.

This PR also fixes #2189

Internally it was clearing information on undo so it couldn't complete the redo properly. Corrected the logic so it would keep the information it needed.

### DIFF
--- a/toonz/sources/toonz/cellkeyframeselection.cpp
+++ b/toonz/sources/toonz/cellkeyframeselection.cpp
@@ -71,7 +71,9 @@ void TCellKeyframeSelection::copyCellsKeyframes() {
     QClipboard *clipboard       = QApplication::clipboard();
     TXsheet *xsh                = m_xsheetHandle->getXsheet();
     TKeyframeData *keyframeData = new TKeyframeData();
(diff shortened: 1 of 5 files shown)


shun-iwasawa:
It works fine. Thank you @manongjohn ! LGTM

PR #2405, opened 2018-12-06
Fix Z=0 Camera Preview crash

### LINKED ISSUE #2372: Crash when Z camera position = 0 in Preview
### Issue Summary
OpenToonz crashes in preview mode when Z coordinate of camera reach bellow 0. 
edit: crash occurs when **Camera Stand View** and **Preview mode** are activated 

https://1drv.ms/v/s!AtMv1SrHrjT38m-OYR9dbAZbItzl

### Steps to Reproduce
Select Animate tool. Choose Camera from drop down list. Choose Position option from drop down list. Activate Preview mode. Select **Camera Stand View**.  Scroll camera Z parameter to negative value. 

### Expected Results
Camera to move into negative space

### Actual Results
OpenToonz crashes

### System Information
* **OpenToonz Version**: 1.2.1 (built jun 28)
* **Operating System**: Windows 10 Pro
* **CPU**: i7 2630QM
* **Memory**: 20 gb
* **Graphics Card**: NVidia Quadro 1000m

manongjohn:
This PR fixes #2372 

NOTE: This issue not only impacts previewing, but it also causes a crash while rendering output.

Fixed it by adding simple logic in places it would crash to check if the area being rendered is a single point.  If so, it will stop trying to render the frame.

If a better solution presents itself please use that.

### DIFF
--- a/toonz/sources/common/tgeometry/tgeometry.cpp
+++ b/toonz/sources/common/tgeometry/tgeometry.cpp
@@ -58,14 +58,18 @@ TAffine TAffine::inv() const {
   if (a12 == 0.0 && a21 == 0.0) {
     assert(a11 != 0.0);
     assert(a22 != 0.0);
-    double inv_a11 = 1.0 / a11;
-    double inv_a22 = 1.0 / a22;
+    double inv_a11 =
+        (a11 == 0.0 ? std::numeric_limits<double>::max() / 2 : 1.0 / a11);
+    double inv_a22 =
+        (a22 == 0.0 ? std::numeric_limits<double>::max() / 2 : 1.0 / a22);
     return TAffine(inv_a11, 0, -a13 * inv_a11, 0, inv_a22, -a23 * inv_a22);
   } else if (a11 == 0.0 && a22 == 0.0) {
     assert(a12 != 0.0);
     assert(a21 != 0.0);
-    double inv_a21 = 1.0 / a21;
-    double inv_a12 = 1.0 / a12;
+    double inv_a21 =
+        (a21 == 0.0 ? std::numeric_limits<double>::max() / 2 : 1.0 / a21);
+    double inv_a12 =
+        (a12 == 0.0 ? std::numeric_limits<double>::max() / 2 : 1.0 / a12);
     return TAffine(0, inv_a21, -a23 * inv_a21, inv_a12, 0, -a13 * inv_a12);
   } else {
     double d = 1. / det();
--- a/toonz/sources/common/tgeometry/tgeometry.cpp
+++ b/toonz/sources/common/tgeometry/tgeometry.cpp
@@ -56,20 +56,22 @@ TAffine TAffine::operator*=(const TAffine &b) { return *this = *this * b; }
 //--------------------------------------------------------------------------------------------------
 TAffine TAffine::inv() const {
   if (a12 == 0.0 && a21 == 0.0) {
-    assert(a11 != 0.0);
-    assert(a22 != 0.0);
     double inv_a11 =
-        (a11 == 0.0 ? std::numeric_limits<double>::max() / 2 : 1.0 / a11);
+        (a11 == 0.0 ? std::numeric_limits<double>::max() / (1 << 16)
+                    : 1.0 / a11);
     double inv_a22 =
-        (a22 == 0.0 ? std::numeric_limits<double>::max() / 2 : 1.0 / a22);
+        (a22 == 0.0 ? std::numeric_limits<double>::max() / (1 << 16)
+                    : 1.0 / a22);
     return TAffine(inv_a11, 0, -a13 * inv_a11, 0, inv_a22, -a23 * inv_a22);
   } else if (a11 == 0.0 && a22 == 0.0) {
     assert(a12 != 0.0);
     assert(a21 != 0.0);
     double inv_a21 =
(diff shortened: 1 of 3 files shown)


toonz/sources/tnzbase/trasterfx.cpp
```diff
@@ -834,6 +836,10 @@ void TRasterFx::compute(TTile &tile, double frame,
   if (tile.m_pos != newTilePos) {
     /*-- RenderSettingsのaffine行列に位置ずれを足しこむ --*/
     TRenderSettings newInfo(info);
+    // Image size is a 0 point.  Do nothing
+    if (!(newInfo.m_cameraBox.x0 + newInfo.m_cameraBox.x1) &&
+        !(newInfo.m_cameraBox.y0 + newInfo.m_cameraBox.y1))
+      return;
```
shun-iwasawa: Please try just putting any level and preview the scene. This condition seems to always prevent the scene to be rendered.

manongjohn: Odd.  The scene I usually test stuff with worked fine, but with a new scene it doesn't. Will l look further into this unless there is a better way to fix this.

manongjohn: My new scene issue appears to be something else.  I think I had some odd setting that was even impacting my official OT release.  Preview made my scene black, no matter what I did.  At any rate, after reinstalling OT using a nightly to recreate my stuff folder, I tried creating a new level and turned on preview.  I am not seeing any issue.   It's working as I expect it to, unless my expectation is wrong. 

Can you send me a scene, a video or exact steps to reproduce what you are seeing so I can see what you are referring to and try to duplicate?

shun-iwasawa: It seems that you need to make the camera dpi to be slightly different from one of the toonz raster / raster level you are putting like as follows:

image

Having the different dpi between the camera and the level makes tile position to be with fraction part. And it is the condition to enter this part of the source.

shun-iwasawa:
@manongjohn 
I think I come up with an alternative plan to avoid the crash which is simply force the camera z not to be zero. I'll post another PR for your review.

shun-iwasawa:
Can you please take a look at #2450 ? Thanks!

manongjohn:
@shun-iwasawa I'm going to revert the changes on trasterfx.cpp back to what it was before and instead focus on fixing the calculation issue, the underlying cause of this.

The underlying issue is in `TAffine::inv()` doing a Divide by 0 issue which results in invalid numbers and causes crashes depending on how the affine is being used.  Definitely not the 1st time I've seen this.

For now, if the calculation is trying to divide by 0, I will force the result to 1/2 of max double which is overkill but leaves some room for growth.  If there is a more appropriate max value I should set, let me know.

shun-iwasawa:
Jenkins

toonz/sources/common/tgeometry/tgeometry.cpp
```diff
@@ -58,14 +58,18 @@ TAffine TAffine::inv() const {
   if (a12 == 0.0 && a21 == 0.0) {
     assert(a11 != 0.0);
     assert(a22 != 0.0);
-    double inv_a11 = 1.0 / a11;
-    double inv_a22 = 1.0 / a22;
+    double inv_a11 =
+        (a11 == 0.0 ? std::numeric_limits<double>::max() / 2 : 1.0 / a11);
+    double inv_a22 =
+        (a22 == 0.0 ? std::numeric_limits<double>::max() / 2 : 1.0 / a22);
     return TAffine(inv_a11, 0, -a13 * inv_a11, 0, inv_a22, -a23 * inv_a22);
```
shun-iwasawa: Sorry if I am wrong, but when `inv_a11` is set to `std::numeric_limits<double>::max() / 2` , does the value `-a13 * inv_a11` easily overflow when `a13` is bigger than 2.0 ?

toonz/sources/common/tgeometry/tgeometry.cpp
```diff
@@ -58,14 +58,18 @@ TAffine TAffine::inv() const {
   if (a12 == 0.0 && a21 == 0.0) {
     assert(a11 != 0.0);
     assert(a22 != 0.0);
```
shun-iwasawa: IMO these assertions should be removed since for now this function should work with the values `a11` and `a22` being 0.0.

shun-iwasawa:
It seems that the problem occurs when the camera Z is set to not only 0 but also near-zero value.
Please try the following:

1. Draw and create a vector level.
1. Set the camera Z to 0.00001 .
1. Toggle Preview button.
1. Zoom in the viewer with mouse wheel (to approx. 150%).

In my environment OT crashes. 

Regarding the probable cause, please see [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/common/trop/quickput.cpp#L1257-L1274).
In the function `doQuickPutNoFilter()` in `quickput.cpp` , `a11` and `a22` of the result of `TAffine::inv()` are multiplied by `(1<<PADN)`  ( `PADN` is set to 16 so `1<<PADN` is 65536 ).  Therefore, in `TAffine::inv()` , it seems that `a11` and `a22` should be smaller than `std::numeric_limits<double>::max() / (1 << 16)` in order to prevent the values from overflowing.

shun-iwasawa:
Tested and found that the fix seems to resolve the crash. Thank you @manongjohn ! LGTM

Issue #356, opened 2016-05-24
UTF-8 encoded .psd metadata cannot be loaded correctly


kogaki:
Related to #309 

#309 was fixed for "SJIS" encoded `.psd` files, but there remains problems for other encodings, especially for utf-8.

kogaki:
But.. UTF-8 is a little bit harder to treat  because it encode characters to various size (2~4 Bytes).

jpturcotte:
Is this still an issue?

RodneyBaker:
Another query of whether this is still an issue.
Especially given fairly recent changes relating to PSD format.
The assumption is that it still is a valid issue so confirmation will be appreciated to keep it on the radar.

PR #2415, opened 2018-12-12
Enable To Reload PSD Levels

shun-iwasawa:
This PR enables Reload command on PSD levels.
I tried simply abolishing the restriction of the command and couldn't find any problem so far.


RodneyBaker:
Shun,
Do you know if this impacts issue #356 in any way?
I will guess not but wanted to check as that issue is still tagged as open.

masafumi-inoue:
LGTM

PR #2416, opened 2018-12-12
Fix Renaming PSD Cell

shun-iwasawa:

Xsheet cells containing PSD levels cannot be renamed by just typing a frame number like a cell containing other types of sequential raster level.


shun-iwasawa:
Jenkins

RodneyBaker:
Assuming I'm not misunderstanding.... this is working as advertised.
I can type numbers into xsheet cells and change the frames created by the PSD import. 
Looks like I need to start using PSD format more often.  Thanks!

masafumi-inoue:
LGTM

PR #2417, opened 2018-12-13
Fix replace fx nodes

### LINKED ISSUE #2177: Schematic:  Crash using Replace FX
### Issue Summary

When I use replace command on a node that has no input, OT crash.

replace no whitout input crash ot

And change the value of cast shadow fx, also crash OT.

### Steps to Reproduce

Use replace command on node whitout input

w10 cintiq companion HD graphics OT 1.21

manongjohn:
This PR fixes #2177 

The logic for replacing fx nodes incorrectly creates a normal fx as a ZeraryColumn fx. When this happens, accessing certain fx information can cause a crash due to bad pointers.

It should be noted that when replacing a normal fx with a ZeraryColumn fx, the ZeraryColumn fx is created as a normal fx incorrectly.

Corrected the logic such that if replacing a ZeraryColumn fx with a normal fx, the normal fx is created as a normal fx.  Conversely, if replacing a normal fx with a ZeraryColumn fx, a new ZeraryColumn fx is created

### DIFF
--- a/toonz/sources/toonzlib/fxcommand.cpp
+++ b/toonz/sources/toonzlib/fxcommand.cpp
@@ -1152,7 +1152,13 @@ void ReplaceFxUndo::initialize() {
     fx = zcfx->getZeraryFx();
   }
 
-  if (has_fx_column(fx)) {
+  TZeraryColumnFx *zcrepfx = dynamic_cast<TZeraryColumnFx *>(repFx);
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Works as advertised.  Thanks @manongjohn!

shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @manongjohn for the fix!!

PR #2418, opened 2018-12-14
Matte Color for Saving Transparent Raster Levels in Non Alpha-enabled Image Format

shun-iwasawa:
This PR will add a new preferences parameter "Matte Color", to be used as background color when **overwriting** raster levels with transparent pixels in non alpha-enabled image format like JPG or BMP. This situation can occur when loading JPG image, erasing it with the Eraser tool and then overwriting it with "Save Level" command. 
The "Matte Color" control is added in the "Saving" category which is newly created.
The default value of "Matte Color" is set to white. If you would like to have black background (which will obtain the same result as before), please make sure to change it to black.

Please note that the Matte Color will **NOT** be applied when **rendering** the transparent scene to non alpha-enabled format. Instead, in such case the Camera BG Color in the scene settings is used, temporarily being converted to full-opac regardless of the alpha channel value.
Currently OT set the solid Camera BG Color only when rendering JPG, AVI, BMP, MP4, WEBM, or GIF.
This PR also fixes this operation to be applied other non-alpha enabled formats as well, such as TIF format with setting the "Bits Per Pixel" option to "24 (RGB)".

Also please note that the Matte Color will **NOT** be applied when **exporting** the transparent image to non alpha-enabled format with "Export Level" command. Instead, in such case the "Export Options > Background Color" in the Export Level dialog is used.

### DIFF
--- a/toonz/sources/common/tiio/tiio_bmp.cpp
+++ b/toonz/sources/common/tiio/tiio_bmp.cpp
@@ -598,6 +598,9 @@ public:
   void flush() override { fflush(m_chan); }
 
   void writeLine(char *buffer) override;
+
+  // for now opentoonz does not support bmp format with alpha channel
+  bool writeAlphaSupported() const override { return false; }
 };
 
 //---------------------------------------------------------
--- a/toonz/sources/common/tiio/tiio_jpg.cpp
+++ b/toonz/sources/common/tiio/tiio_jpg.cpp
@@ -256,6 +256,9 @@ public:
     }
     jpeg_write_scanlines(&m_cinfo, m_buffer, 1);
   }
+
+  // jpeg format does not support alpha channel
+  bool writeAlphaSupported() const override { return false; }
 };
 
 //----
--- a/toonz/sources/common/timage_io/timage_io.cpp
+++ b/toonz/sources/common/timage_io/timage_io.cpp
@@ -11,6 +11,7 @@
 #include "trasterimage.h"
 #include "tiio.h"
 #include "tfilepath_io.h"
+#include "tpixelutils.h"
 
 // boost includes
 #include <boost/range.hpp>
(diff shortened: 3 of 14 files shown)


shun-iwasawa:
Jenkins

RodneyBaker:
I've been using an undocumented 'feature' to render gifs with transparency in OT.
Interstingly, I was recently thinking that it was likely a change could take that capability away.

The way to test this (and see some other results of this current PR as well is to use the Multiple render option in Output settings and render the flow.   This renders each column of the xsheet separately.
It is there that OT renders out gif sequences with transparency.

So... might case is rather niche. ;)

artisteacher:
I left the default settings, except for multiple rendering as @RodneyBaker described & I’m getting a separate animated gif for each column with a transparent background.

shun-iwasawa:
Unfortunately I couldn't manage to render the transparent GIF yet.
@RodneyBaker Sorry for the trouble, but can you please check if the the "feature" is still available in the Appveyor build of this PR? If yes, then we can introduce this at ease.

RodneyBaker:
Thanks Shun.  I'll look into this and where there are any disconnects record/document so they are easier to understand and diagnose.  I don't anticipate any issues with merging this PR but and I think the flexibility of the options available will allow for continued use of gif transparency using the undocumented feature.

RodneyBaker:
Shun, the undocumented feature still appears to work so that isn't an obstacle to merging this feature.
Thanks!  Depending on the program that is used I see different things but that is likely something I just hadn't noticed before and isn't an issue.  The important thing is that I can still render to gif and other format such as spritesheet with transparency.

Green light from me! :)

shun-iwasawa:
Fine! So let's investigate the mysterious GIF behavior in the future as another challenge :-)

RodneyBaker:
Yes.  I'm confident we can resolve any remaining gif transparency issues and even extend that to other formats such as APNG.  In the process those of us scratching at the surface of Opentoonz source code will learn a whole lot in the process. :-)

masafumi-inoue:
LGTM

RodneyBaker:
I'm starting to think you might have snuck the gif transparency into this PR.

texttesta

RodneyBaker:
Interestingly, the only file format I'm locking up on is GIF.
Here's spritesheet (with transparency!):
texttest_textiwa01

RodneyBaker:
Perhaps a false alarm.  The problem seems to have resolved itself after I changed the render target to a new filename and set the ffmpeg timeout back to its default.  (increasing it again later didn't produce any issues).  
texttestz_textiwa01

RodneyBaker:
I'm happy to report this addition appears to have resolved #769.

PR #2419, opened 2018-12-14
Improve JPEG Quality

shun-iwasawa:
This PR will improve JPEG quality especially with higher value of "Quality" option.

Before this change, OT has been always using 4:1:1 [chroma subsampling](https://en.wikipedia.org/wiki/Chroma_subsampling) when saving JPEG image. It has better compression rate but is poor in quality even if you set the Quality to 100. 
So I made the chroma subsampling type to be with better quality ( 4:1:1 -> 4:2:2 -> 4:4:4 ) along with the "Quality" option value. It will enlarge the file size, but will enables users to get the nice result.

jpeg_chroma_subsampling

In the above image results with Quality=25 are exactly the same and the difference is most obvious in the results with Quality=75. Please note that the chromatic noises around the black line are reduced.

### DIFF
--- a/toonz/sources/common/tiio/tiio_jpg.cpp
+++ b/toonz/sources/common/tiio/tiio_jpg.cpp
@@ -214,12 +214,29 @@ public:
 
     if (!m_properties) m_properties = new Tiio::JpgWriterProperties();
 
-    jpeg_set_quality(
-        &m_cinfo,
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2425, opened 2018-12-17
Fix loading macrofx preset issue

### LINKED ISSUE #2424: Failure to read pointParam when instancing certain Macro FX
### Issue Summary
Attempting to instance a certain Macro FX causes OT to crash.
Attached are two similar MacroFX that exhibit this issue.
[macroFx.zip](https://github.com/opentoonz/opentoonz/files/2683833/macroFx.zip)

image

EDIT:
This is an image from an unrelated issue, you can see the original version of the Outline Macro that presented no issue.
image

### Steps to Reproduce
1. Create the nodes and link them up as shown in the reference image.
2. Select all of the nodes shown in the red box and create a Macro from them.
3. Save the Macro, the name doesn't matter.
4. Attempt to create an instance of the Macro FX in the Schematic window.

### Expected Results
The Macro FX is instanced in the Schematic and can be used.

### Actual Results
OpenToonz crashes as soon as the Macro is selected from the drop-down list.

### System Information
1.2.1 (Dec 15 2018 Appveyor build)
Windows 10
I7 5820k
16GB DDR 4
GTX 980

manongjohn:
This PR fixes #2424

The `skipChild()` procedure does not handle Open-Close tags (ie `<tagname />`) properly like the pointParam of a Curves Fx.  Corrected logic to account for this.

### DIFF
--- a/toonz/sources/common/tfx/tfx.cpp
+++ b/toonz/sources/common/tfx/tfx.cpp
@@ -61,6 +61,7 @@ void skipChild(TIStream &is) {
       if (is.isBeginEndTag()) is.matchTag(tagName);
       is.closeChild();
     }
+    if (is.isBeginEndTag()) is.matchTag(tagName);
   }
 }
 


shun-iwasawa:
Jenkins

shun-iwasawa:

WolfInABowl:
It appears that this issue has reemerged; testing on the latest official build and @manongjohn 's latest OTX, I can reproduce the original crash with the Curves node.

manongjohn:
Strange. I used the original files from the issue to retest this and it crashes in a different place than what I remembered.  Even removing this fix still doesn't crash where I remember it.  Perhaps I fixed a different issue I encountered looking into this without realizing it and this one was never really fixed.

Anyway...A new PR to fix this is forthcoming.

PR #2364, opened 2018-11-06
Fix raster paste crash

manongjohn:
This PR fixes #1730 

When attempting to paste a raster selection while the current focus is the scene viewer canvas, if there isn't an image in the frame already, a new blank image will be created for the frame so it will have something to paste into.

Note: When autocreation is disabled, no blank image will be created and the paste will not do anything.

PR #2428, opened 2018-12-18
Update Tool State on Level Switch

shun-iwasawa:
This will fix the problem after merging #2364 as follows:

- Create a raster or toonz raster level with one frame
- Draw with the brush or geometric tool
- Use the shortcut "S" to change to the select tool
- Click & drag to select a rectangle around the drawing
- Ctrl-C to copy
- Left-click to exit the selection & then use the right or down arrow key TWO TIMES to navigate to the two frames down.
- Ctrl-V to paste
- Level is created, but the pasted selection does not appear in the viewer.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2431, opened 2018-12-19
Fix Pressure Lost

shun-iwasawa:
This will fix the following problem, related with the fix #2369 .

When you are using tablet, the pressure is unexpectedly lost after you do one of the following operations:
- Try to draw on the hidden column and close the warning popup.
- Drag the vertical or horizontal marker which is for comparing the previewed frame with the snapshot.

Actually, this modification seems to avoid the crash reported in #2311 as well.


masafumi-inoue:
LGTM

PR #2433, opened 2018-12-19
Use range-based for instead of foreach

otakuto:


shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM Thank you @otakuto for refactoring!

PR #2441, opened 2018-12-22
Fix plastic deformer fx crash

### LINKED ISSUE #2440: Crash when Motion Blur FX is applied to level deformed by Plastic Tool

ezgif-1-6d1462d59ee3

### Steps to Reproduce

1. Create an animation level with plastic deformation
2. Apply Motion Blur FX to level
3. Turn on Preview

### System Information

manongjohn:
This PR fixes #2440 

This patch fixes an issue where the off screen rendering surface for plastic deformed levels was not previously created causing a crash when trying to use it.

This will create it immediately if not found.

### DIFF
--- a/toonz/sources/toonzlib/plasticdeformerfx.cpp
+++ b/toonz/sources/toonzlib/plasticdeformerfx.cpp
@@ -20,6 +20,7 @@
 
 // TnzBase includes
 #include "trenderer.h"
+#include "trasterfx.h"
 
 // TnzCore includes
 #include "tgl.h"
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
@manongjohn 
The modification you made in this PR will create `QOffscreenSurface` outside of the main (GUI) thread when rendering / previewing the scene. 
According to the [Qt documentation](http://doc.qt.io/qt-5/qoffscreensurface.html#details) : 

> Note: Due to the fact that QOffscreenSurface is backed by a QWindow on some platforms, cross-platform applications must ensure that create() is only called on the main (GUI) thread. The QOffscreenSurface is then safe to be used with makeCurrent() on other threads, but the initialization and destruction must always happen on the main (GUI) thread.

The cause of the problem seems to be that the alias of Motion Blur Fx does not include the ones of the upstream nodes. I posted an alternative solution. Can you please take a look at #2446 ? Thanks!

manongjohn:
@shun-iwasawa In light of your response, I will cancel this PR since my fix is being done outside the main thread.

My only concern about #2446 is it only affects 1 specific fx or a few based on inheritance of the base class you fixed it in. What other fx may have this problem now but we haven't seen a crash report for yet? New fx may also run into this issue if the developer doesn't account for plastic deformed levels and add the logic you added.

shun-iwasawa:
@manongjohn 
For this time, the fix is done only on the Motion Blur Fx which was only problematic as far as I could find.
Fxs which have input port should return the alias including aliases of upstream nodes.
In most cases you don't need to re-implement `getAlias()` since `TRasterFx::getAlias()` will take care of it.
The case of the Motion Blur Fx, it needs to be re-implemented since the result may change according to not only the fx parameter but also the level's movement. In the function `MotionBlurFx::getAlias()` the returned alias includes frame number so that the fx is computed for every frame.

Issue #2439, opened 2018-12-21
Build and stability issues in pre-RC nightly builds


RodneyBaker:
@shun-iwasawa
Collecting a number of reported issues with recent (nightly) releases

RodneyBaker:
- [ ] Timecode Iwa FX appears to be missing from latest nightly releases

RodneyBaker:
- [ ] SVG fixes don't seem to have taken effect

RodneyBaker:
I'm not experiencing the load/quit/crash problems with latest nightly release.  Those appear to only be issues with a few nightlies prior.  

The two issues most pressing now with regard to recent updates then appear to be the missing Timecode FX IWA (known typo in the code) and the recent SVG updates that don't appear to resolve the SVG issues.  The latter may be because those specific PRs merged weren't designed to address the specific problems still being seen.

RodneyBaker:
Closing as all observed issues are no longer present in the latest nightly build.
Nicely squashed! :)

PR #2444, opened 2018-12-25
Revert Fx List

shun-iwasawa:
This PR will revert the fx list to fix the problem introduced when #2393 was merged.
The problem is also listed in #2439.


shun-iwasawa:
I'll merge this by myself as it is obvious. Sorry for the trouble!

PR #2446, opened 2018-12-27
Fix Motion Blur Fx Alias

### LINKED ISSUE #2440: Crash when Motion Blur FX is applied to level deformed by Plastic Tool

shun-iwasawa:
This PR fixes #2440 as an alternative solution of #2441 .
This PR also fixes the following problem with the Motion Blur Fx:

When you apply the Motion Blur Fx and preview the scene, changes done to the upstream nodes of the fx (such as changing the color of styles of the blurred level, changing parameters of the upstream nodes, etc.) does not trigger the regeneration of preview.

The "alias" of the fx is a text data which identifies the status of rendering frame in order to determine if it can reuse the cache or it needs to be re-rendered. The cause of the both problem was that the alias of the Motion Blur Fx does not include the ones of the upstream nodes.

### DIFF
--- a/toonz/sources/stdfx/motionblurfx.cpp
+++ b/toonz/sources/stdfx/motionblurfx.cpp
@@ -446,10 +446,26 @@ public:
 
   std::string getAlias(double frame,
                        const TRenderSettings &info) const override {
+    std::string alias = getFxType();
+    alias += "[";
+
+    // alias of the effects related to the input ports separated by commas
+    // a port that is not connected to an alias blank (empty string)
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2447, opened 2018-12-27
Add Warning Popup for Duplicate Command on Xsheet

shun-iwasawa:
This PR will add a warning popup to notify user that the "Sync Level Strip Drawing Number Changes with the XSheet" preference option should be enabled before using the duplicate command on xsheet / timeline, which is introduced by @turtletooth in #2342 .

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -1911,7 +1911,13 @@ void TCellSelection::pasteKeyframesInto() {
 //-----------------------------------------------------------------------------
 
 void TCellSelection::duplicateFrame() {
-  if (!Preferences::instance()->isSyncLevelRenumberWithXsheetEnabled()) return;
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

gab3d:
Great! :)

Just an improvement that comes to my mind:
Would it be possible to include in the warning dialog itself a means to activate that option right away?
I think, usability wise, it would eliminate the somewhat tedious process of:
1. Closing the warning
2. Going to File > Preferences to open the preferences window
3. Looking there for the option in the several categories on the left (the warning dialog text doesn't states where exactly in the Preferences it resides that option).
4. Actually activate the option.
5. Close the preferences.
6. Going back to xsheet to duplicate again.

On the other hand, if the warning dialog could let the user activate the option right there, it would also let the user perform the action when clicking the OK button, and continue working without any disturbance.

Thanks!!

PR #2452, opened 2018-12-30
Autohold drawing substitutions

manongjohn:
This PR modifies the behavior of drawing subtitution with Autocreation = Use Xsheet as Animation Sheet
and Enable Auto-stretch frame = checked.

When drawings are substituted with Q or W shortcuts, the current behavior is to just replace the target cells but any held cells afterwards remain the original drawing.

This change will now update the following held cells also, in keeping with the auto-hold behavior that occurs when adding new drawings in the middle of held cells.

autoholddrawing

### DIFF
--- a/toonz/sources/toonz/xsheetcmd.cpp
+++ b/toonz/sources/toonz/xsheetcmd.cpp
@@ -866,6 +866,7 @@ void DrawingSubtitutionUndo::setDrawing(const TFrameId &fid, int row, int col,
 //-----------------------------------------------------------------------------
 
 static void drawingSubstituion(int dir) {
+  Preferences *pref         = Preferences::instance();
   TCellSelection *selection = dynamic_cast<TCellSelection *>(
       TTool::getApplication()->getCurrentSelection()->getSelection());
   TCellSelection::Range range;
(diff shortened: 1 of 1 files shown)


gab3d:
Makes so much sense!
Thanks

turtletooth:
@manongjohn I'm not sure I understand this one.  If you hold alt and do q or w, it already will change all held frames.  Is this different than that?

manongjohn:
Didn't realize Alt-Q/W does that.  Yet another not so obvious action to perform.

At any rate, alt-q/w is good if you are not using Xsheet as Animation Sheet.

This change basically does holds automatically if you are using Xsheet as Animation Sheet so you don't have to do additional key presses, though my implementation might be redundant. I'll check to see if i can reuse the Alt logic.

turtletooth:
The benefit of having two commands is that there are sometimes a person may want to only substitute one frame of a hold, or substitute all subsequent frames in a hold.  I'd like to not lose the ability to substitute one frame at a time - that can be useful for lip sync work.

gab3d:
@turtletooth 
I didn't know the Alt option existed either.
Better said, I've always knew the command existed but didn't deciphered what its name meant. 😄 
It seems to me that name is not so good: **Similar Drawing Substitution Forward/Backwards**, why "Similar"?

bobbybutterfly:
Please try keyframe Adjustments by yourself. Its aweful in OT, its perfect in Toonboom.
Simply move your keys without messing up all other keys

bobbybutterfly:
rober05012019184357

artisteacher:
So Toonboom Harmony's drawing replacement is based upon the drawings being visible in the Sheet/Timeline? I really like using W/Q since it can toggle through all of the drawings in the level, whether or not they are already exposed. It also seems like fewer steps. I do like seeing the change in focus in the drawing substitution window though. I wonder if that could be extended to the Level Strip?

artisteacher:
@bobbybutterfly - oh, I see. I guess the current shortcuts are only helpful if you are using your keyboard. Using the keyboard seems easier to me, but I don't use a mouse or tablet regularly.  If you are using a tablet or etc, your suggestion could be more helpful.

turtletooth:
@manongjohn I'm building a new OTX and came across this PR again.  I'd like to find a way to compromise on this functionality.  I use xsheet as animation sheet and don't want to lose the ability to just do a one frame substitution if that is all that's needed.  For lip sync work, that comes in handy.

manongjohn:
I suppose I can just cancel this since ALT-Q/W do the same thing, though ATL-Q/W doesn't work on selected cells across multiple columns like Q/W or my changes allow.  Nor does it fill empty selected cells.

It just seems odd that other cell operations autohold but this action doesn't when using Xsheet as Animation Sheet.  I'd almost suggest that maybe Q/W and ALT-Q/W should swap how it works depending on the Autocreation mode selected, but I think that would be too inconsistant in command behavior and may be confusing.

I will augment ALT-Q/W to work more like Q/W in terms on how it impacts empty cells and multiple columns since it seems overly restricted for not apparent reason.

PR #2465, opened 2019-01-07
Update Chinese UI translation for v1.3

jcome:


Savraska2:
@jcome It seems you forgot to attach tnztools.qm and toonzlib.qm files.

shun-iwasawa:
@jcome 
谢谢你的翻译！LGTM

PR #2469, opened 2019-01-08
Fix vector replacement crash

### LINKED ISSUE #2466: Crash attempting to convert FX to Vector etc.
There are two easy ways to crash OT with the new TextIwa FX.

Steps:
1. Instance a new TextIwa node in the FX Schematic
2. With the newly created level selected, choose either 'Vectors to Toonz Raster' or 'Replace Vectors with Simplified Vectors' from the 'Level' drop-down
3. OT will crash when either command is chosen

**Edit**:
These crashes are also present on any levels created by FX e.g. ParticleFX, Clouds, ColorCard, TimeCodeIwa, any gradient FX, etc.

Some checks need to be put in place to prevent these commands being run on levels that have been created by FX.

Any node with this color can be used:
image

Checks are already in place in other circumstances as if the above commands are run on a non-Vector level this message appears:
image

It seems that these checks are not run in the case that the selected level was an FX-created level.

manongjohn:
This PR fixes #2466 

Corrects logic to pop up a message if actions are not performed on a valid level.

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -2632,7 +2632,7 @@ bool TCellSelection::areOnlyVectorCellsSelected() {
   }
 
   TXshSimpleLevel *sourceSl = firstCell.getSimpleLevel();
-  if (sourceSl->getType() != PLI_XSHLEVEL) {
+  if (!sourceSl || sourceSl->getType() != PLI_XSHLEVEL) {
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed this PR fixes the crash. Thank you @manongjohn ! LGTM

Issue #1016, opened 2017-01-31
FX Swatch Viewer Size Issues


artisteacher:
I've run into an issue where the swatch viewer (in some FX settings), keeps resetting so that it is hidden/minimized. In the particle settings, any change in the swatch viewer (ie. changing frames) results in the viewer minimizing. This has been ongoing and continues in the latest nightly build in OS X 10.11.6 and 10.12.3.
![fxglitch1](https://cloud.githubusercontent.com/assets/21159116/22476961/4b00451a-e7b2-11e6-8ebb-a05a62c7be79.gif)
In FX where there is only the swatch viewer and no adjustable settings, the swatch viewer opens in a strange minimized position.
![fxglitch2](https://cloud.githubusercontent.com/assets/21159116/22481338/ddf6ace6-e7c2-11e6-928b-5d41417b6a81.gif)

mcjadski:
When I tried the problem with the Particles FX on OSX 10.11, I got almost the same result, except that my screen is bigger, so you can see that it's not actually minimizing:  

![swatch_minimize](https://cloud.githubusercontent.com/assets/25377429/22489006/c68ee498-e7da-11e6-9980-4702335816fa.gif)

When you go to a different frame (and when you first open the properties window), the big, gray region above the Swatch Viewer grows vertically, and seems to push the Swatch Viewer down below what is viewable.  You have to minimize and then maximize the Swatch Viewer before you can get a full view of it again.  

Also, when I try adjusting the properties window size, the window immediately adjusts itself so that the Swatch Viewer is in full view.  But then if I go to a different frame, the window expands vertically to take up the full height of the screen (minus top and bottom bars).  (The Swatch Viewer is in full view still; the window just refuses to stay the size I wanted it to be.)  

If I then go to a different frame a second time, the Swatch Viewer is once again pushed out of view.  

As for the second problem, it's even worse for me when I first open the Color Dodge effect:  

![tiny_window](https://cloud.githubusercontent.com/assets/25377429/22489140/6214a402-e7db-11e6-8347-ea41be68f3af.jpg)

But then if I resize the window, exit it, and reopen it again, it looks the way it does for artisteacher.  

I also tried these on Win 7.  I didn't have the same problem with the Particles properties window.  However, when I did the ColorDodge effect, the window starts way too small:  

[...]

PR #2479, opened 2019-01-14
Fx setting popup enhancements

manongjohn:
This PR accomplishes the following:

1) Addresses some of the issues reported in #1016

Ensure a minimum fx settings popup size of 380x50.
Not resize of the popup to original dimensions when switching frames.

This will not remember popup dimensions between fx nodes.

2) When the fx settings is open, clicking on nodes will immediately update the popup to display the node's settings, if any.

This does not apply to Xsheet and Output fx nodes.

3) Double clicking column nodes will open fx settings popup for consistancy and quick access to preview swatch.

4) Fx Settings popup is now a dockable window.


manongjohn:
@shun-iwasawa Yes, it is mimiced after Harlequin. It was a request by a user, which made more sense to me than double-clicking, especially if you are doing a lot of work in the fx schematic and moving between nodes a lot.

I will back out that change if requested unless another option is presented.

gab3d:
@manongjohn 
just wanted to say I like FX Settings panel being dockable! :tada: 
thanks!!

gab3d:
Which makes me think... :thinking: 
...if only the styles Settings tab could reuse this same dockable panel, it´d be just amazing!

Currently, the workflow when painting with MyPaint brushes suffers considerably because brush Settings are just a tab in the Style Editor, so you have to be permanently switching between the Color, Raster and Settings tabs... or open several Style Editor windws and have each one in a different tab (that sort of works, but have one big problem, when you reopen OT all the style editors open in the Color tab by default.

I think that if styles Settings could be made to use the same dockable panel you made for FX Settings (as just another tab in it), then that panel could start to work as a kind of "general Settings panel", for whatever is currently selected (being either the current brush).

Well, it´s just an idea... I don't really know if it's even possible to do so.

manongjohn:
@gab3d I will say almost anything is possible. :) It's already in tab so putting it in its own pane shouldn't be too difficult, I would think. I agree, a separate Style Settings pane would be nice.  Definitely something to think about. :)

shun-iwasawa:
I've checked Ghibli staffs and confirmed that there seems to be no problem for them with this PR, except the following two points as follows:
- When you firstly open the Fx Settings it always become at top-left corner of the screen. It should be opened at the center of the screen just like other panels.
- When you click and activate the floating Schematic panel, it comes on top of the floating Fx Settings. The Fx Settings should always be on top of the Schematic.

In order to resolve them, how about modifying the source like as follows?:
```
----------------------- toonz/sources/toonz/tpanels.cpp -----------------------
index 77b4d77..4b5b21b 100644
@@ -1393,10 +1393,12 @@ public:
 
   TPanel *createPanel(QWidget *parent) override {
     FxSettingsPanel *panel = new FxSettingsPanel(parent);
+    panel->move(qApp->desktop()->screenGeometry(panel).center());
     panel->setObjectName(getPanelType());
     panel->setWindowTitle(QObject::tr("Fx Settings"));
     panel->setMinimumSize(390, 85);
     panel->allowMultipleInstances(false);
+    panel->setStaysOnTop(true);
     return panel;
   }
 

--------------------- toonz/sources/toonzqt/docklayout.cpp ---------------------
index ffb0b69..3228635 100644
@@ -845,6 +845,10 @@ bool DockLayout::undockItem(DockWidget *item) {
   // Usando Tool questo non accade. In questo caso, i placeholder devono essere
   // disattivati...
 
+  // make the panel to stay on top of the others. see FxSettingsPanel
+  if (item->staysOnTop())
+    item->setWindowFlags(item->windowFlags() | Qt::WindowStaysOnTopHint);
+
   item->setFloatingAppearance();
   item->m_floating = true;
 

---------------------- toonz/sources/toonzqt/docklayout.h ----------------------
index 2b0a69f..ef9d9df 100644
@@ -186,6 +186,8 @@ protected:
   // Maximization
   bool m_maximized;
 
+  bool m_staysOnTop = false;
+
 private:
   QPoint m_dragInitialPos;
   QPoint m_dragMouseInitialPos;
@@ -269,6 +271,9 @@ public:
   // reimpremented in FlipbookPanel
   virtual void onDock(bool docked) {}
 
+  bool staysOnTop() { return m_staysOnTop; }
+  void setStaysOnTop(bool on = true) { m_staysOnTop = on; }
+
 private:
   // Event handling
   // Basic events

-------------------- toonz/sources/toonzqt/tdockwindows.cpp --------------------
index b18abd8..3b16f94 100644
@@ -262,6 +262,7 @@ int TDockWidget::isResizeGrip(QPoint p) {
 void TDockWidget::setFloating(bool status) {
   if (status) {
     setWindowFlags(Qt::Tool | Qt::FramelessWindowHint);
+    if (staysOnTop()) setWindowFlags(windowFlags() | Qt::WindowStaysOnTopHint);
     if (!m_floating) parentLayout()->undockItem(this);
   }
 }
```

For the former problem I added a line in `FxSettingsFactory::createPanel()` for force-moving the panel to the center. For the latter request, I added a new property to the `DockWidget` which will make the widget always stay on top when floating.

manongjohn:
@shun-iwasawa Thank you and the Ghibli staff for allowing this change to proceed forward as is. I will gladly submit the recommended changes shortly.

manongjohn:
@shun-iwasawa I made the changes and was quickly testing it out.  The only thing I noticed was that now the fx settings window stays in front of every window, even when OT is hidden behind other windows.  Not sure that's a good thing. It does minimize when OT is minimized.

In the current release, the fx settings window moves behind the floating Schematic panel when clicked on so it's not a new behavior.

shun-iwasawa:
> In the current release, the fx settings window moves behind the floating Schematic panel when clicked on so it's not a new behavior.

.. Ah, you are right!
So sorry for the mess, but please forget the latter request regarding the stacking order of the panels.

gab3d:
@shun-iwasawa @manongjohn 
>     * When you firstly open the Fx Settings it always become at top-left corner of the screen. It should be opened at the center of the screen just like other panels.

...speaking of which, at least in my windows 10 notebook, almost all of the option windows (Scene, Preview, Output, Camera, etc. settings) open non centered when I do a fresh install of OT. From my experience, they all tend to open offcentered to the bottom-right of the screen). 
A different case is the Memo window that by default opens so far to the top-left of the screen that even the window title is off screen and you have to resize it, to force it to relocate itself wholly inside the screen; the you can move it to wherever you want it to open and it remembers that new position correctly.

manongjohn:
@shun-iwasawa Last question. Would it be acceptible if I convert from RightClickMenuAction "Edit FX ..." to MenuWindowsAction "FX Editor" so it can be added it to Windows Menu dropdown?  For the end user, it will mostly appear as a label change on the right-click menus where they appear now with access to it from the Windows menu.

manongjohn:
Oddly, without the force move to center, the initial opening of the FX Settings will create it in the upper-left corner (0,0).  However, if you double click any fx node, it will either move/open in center on it's own.  Seems like only the initial creation that is the problem.

I'm not sure why it does this on the initial creation for this panel but doesn't do it on other panels.  It might be because I only allow 1 instance of a floating FX Settings to be created, where as all other panels always creates another copy and might move it automatically in that logic.

shun-iwasawa:
@manongjohn 

> Would it be acceptible if I convert from RightClickMenuAction "Edit FX ..." to MenuWindowsAction "FX Editor" so it can be added it to Windows Menu dropdown? 

No problem :+1: . (So sorry for my late reply!)

Jenkins

turtletooth:
Thanks @manongjohn for the great feature!

openanim:
I was wondering how this could impact other workflows if the fx editor panels were linked and share data ?
 
In order to be able to use the panel alone and another panel for settings in small screens.

Currently a modification on one does not operate on the other in the settings. But there may be advantages to it being independent ?

But the strange thing is that the data between the panels changes at the same time but the display of the image does not reflect these changes between them.

panel fx

PR #2483, opened 2019-01-18
Fix style picker and type tools invalid status

manongjohn:
This PR fixes the issue where the Style Picker and the Type tools are invalid for certain levels but does not change icon or bring up popup message that it can't be used on non vector and non-toonz raster level types.

NOTE: These changes were moved out of PR #2087

### DIFF
--- a/toonz/sources/tnztools/stylepickertool.cpp
+++ b/toonz/sources/tnztools/stylepickertool.cpp
@@ -51,7 +51,7 @@ StylePickerTool::StylePickerTool()
   m_colorType.addValue(LINES);
   m_colorType.addValue(ALL);
   m_colorType.setId("Mode");
-  bind(TTool::CommonLevels);
+  bind(TTool::VectorImage | TTool::ToonzImage);
 
   m_prop.bind(m_passivePick);
   m_passivePick.setId("PassivePick");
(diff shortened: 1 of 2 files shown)


toonz/sources/tnztools/stylepickertool.cpp
```diff
@@ -213,6 +213,17 @@ void StylePickerTool::mouseMove(const TPointD &pos, const TMouseEvent &e) {
 
 int StylePickerTool::getCursorId() const {
   int ret;
+
+  TImageP img;
+  if (!m_active)
+    img = touchImage();
```
shun-iwasawa: This line seems to cause creating a new level when clicking an empty cell in Xsheet.

manongjohn: Correcting.

shun-iwasawa:
Jenkins

toonz/sources/tnztools/typetool.cpp
```diff
@@ -856,7 +856,11 @@ void TypeTool::updateTextBox() {
 void TypeTool::updateMouseCursor(const TPointD &pos) {
   int oldCursor = m_cursorId;
 
-  if (!m_validFonts)
+  TImageP img      = getImage(false);
+  TVectorImageP vi = img;
+  TToonzImageP ti  = img;
+
+  if (!m_validFonts || (!vi && !ti))
```
shun-iwasawa: For the Type tool, it seems that the modification done at line 425 is sufficient.
This tool is available on empty target but the condition here sets the cursor id to `CURSOR_NO` when selecting the empty cell.

shun-iwasawa:
Also found that for now the cursor id `ToolCursor::CURSOR_NO` is not assigned to any cursor image. I'll fix it right away.

shun-iwasawa:
Jenkins again

PR #2485, opened 2019-01-19
linux data: misc fixes

eszlari:


shun-iwasawa:
Jenkins

shun-iwasawa:
I haven't Linux environment but confirmed that there is no problem in other platforms. Thank you @eszlari for the fix! LGTM

PR #2486, opened 2019-01-19
Fix toonz raster deactivate crash

### LINKED ISSUE #2484: Toonz Raster crash selecting frame in 2nd level of a column
Getting a crash in 1.3RC when selecting a blank frame on Toonz Raster level when it is the second level in the same column

To recreate:

Using pen
Autocreation mode =  Xsheet As Animation Sheet with Auto-stretch enabled

1) Open new scene and select Brush tool
2) Select Frame 1 on xsheet and draw to create a Vector level
3) Select Frame 13 in xsheet and create a new Toonz Raster level and draw something
4) Select Frame 15 in xsheet and draw something
7) Select Frame 16 or 17 -> Crash 
    If it doesn't crash right away, draw something then try selecting another Frame. 1 or 2 away from the previous.

This only crashes when the 2nd level is the Toonz Raster level and when using a pen.

### System Information
* **OpenToonz Version**: 1.3RC. Also seen in 1/2/19 Nightly, but does not happen in official release
* **Operating System**: Win7 w/ XP-Pen tablet and Win10 laptop-tablet

manongjohn:
This PR fixes #2484 

The issue occurs only when drawing with a tablet in columns with multiple levels.  The 1st time you select a new empty cell and draw, it's fine. However, if you attempt to select another empty cell right away, it will crash because of an internal tool switch (vector -> toonz raster in this case) that occurs but does not properly end the toonz raster brush stroke. It crashes because it tries to access an image that doesn't exist on the empty cell.

Added logic to do nothing in the deactivation if there is no image, as well as enhanced the logic to properly handle to tool switch.

The logic I added uses the existing `m_toolSwitched` variable that tracks tool switching on the main viewer.  When it detects a tool switch in a drag event, it will call `leftButtonDown()` on the tool to properly initiate the tool and then clear m_toolSwitched.  This will allow the `leftButtonUp()` function to be called properly when drawing stops.

There are 2 additional effects of this change when there are multiple levels in 1 column:
1) Currently when drawing with a mouse, when you click an empty cell, it will generate the frame but you cannot continue drawing. You would need to stop the click-hold and start again to draw.

Now when you draw with a mouse on an empty cell, you can continue drawing right away, similar to how it works with the pen.

2) Currently the new cell and initial drawing (with pen) are not properly recorded and cannot be undone/redone.

Now that the tool is being properly ended, changes are properly recorded for undo/redo operations.

### DIFF
--- a/toonz/sources/tnztools/toonzrasterbrushtool.cpp
+++ b/toonz/sources/tnztools/toonzrasterbrushtool.cpp
@@ -1133,6 +1133,7 @@ void ToonzRasterBrushTool::onDeactivate() {
   if (m_tileSaver) {
     bool isValid = m_enabled && m_active;
     m_enabled    = false;
+    m_active     = false;
     if (isValid) {
       finishRasterBrush(m_mousePos,
                         1); /*-- 最後のストロークの筆圧は1とする --*/
@@ -1474,6 +1475,7 @@ void ToonzRasterBrushTool::leftButtonUp(const TPointD &pos,
                                         const TMouseEvent &e) {
   bool isValid = m_enabled && m_active;
   m_enabled    = false;
+  m_active     = false;
   if (!isValid) {
     return;
   }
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed the crash and the fix. Thank you @manongjohn ! LGTM

Issue #40, opened 2016-03-26
Compilation notes for OSX


justinclift:
For anyone else trying to compile on OSX, please note the OSX compiling instructions (so far) are missing a few lines (eg libusb dependency).

Some initial commits here to add further bits, which might help others:

https://github.com/justinclift/opentoonz/commits/addqt5

Emasoft:
Thank you for improving the OSX instructions! Can you also make a shell script to do all that?

justinclift:
At first thought, that should be doable. :smile:

One way to find out for sure though... :wink:

scribblemaniac:
I'm having some difficulties getting it to compile on my mac, but when I figure it out I will try to make a brew formula for opentoonz which will make it very easy to build on mac. A shell script and some documentation would still be useful for developers though.

justinclift:
@scribblemaniac Do the adjustments here help with your OSX compiling?

&nbsp; &nbsp; https://github.com/justinclift/opentoonz/commit/b7cf4cdc1ae8b966f3783c6617635e2b73a1a264

With the Homebrew formula, that's a good thought.  It might need to wait until some of the dependency problems are fixed (eg the customised tiff 4.0.3 code), but then again... probably better to get it ready sooner rather than later. :smile:

I was going to put some time into it... but I'm kind of tired and not really feeling into it atm.  :innocent:  If you get the Homebrew formula happening instead, that would be great. :grin:

scribblemaniac:
@justinclift Actually, I was already using your branch! Thank you very much as that probably saved me quite a bit of trouble. Unfortunately the problems I am currently having are related to libtiff, probably just conflicts with the system-installed 4.0.6 version which I need for building other programs.

justinclift:
Ahhh, yeah.  It definitely requires the bundled tiff-4.0.3.  If you can somehow force it to detect that, you should be fine. :smile:

justinclift:
The tiff detection bits are here, though you've probably already found them and started trying things out:

* https://github.com/opentoonz/opentoonz/blob/master/toonz/cmake/FindTIFF.cmake
* https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/CMakeLists.txt#L144-L145

You might be able to do something super simple like just hard code the path value into the CMakeLists.txt file.  Not sure though, as I'm not very clueful with CMake atm.

Emasoft:
OpenToonz has been just added to homebrew-cask. 
To install on OS X:
[...]

Issue #2496, opened 2019-01-25
OSX 10.14.2 local build failure


yehjaemin:
### Edit 1
I found some closed issues from a couple years ago that go over this problem but their fixes don't work for me. Specifically, #40 Compilation notes for OSX and #1566 Undefined symbols for architecture x86_64 mention running make clean and changing toonz/cmake/FindTIFF.cmake line 22 from libtiff.a to libtiff.dylib, which seemed to work for them but isn't working on my current version of OS X.

### Edit 2 - Working
After hard deleting the build directory and remaking it, I was able to successfully run make! It seems that the edit to FindTIFF.cmake did the trick. If no one minds, I'd like to submit a pull request that changes the file.

### Issue Summary
When building on Mac Mojave 10.14.2 using the local build instructions, I run into an error on the final make step. Everything works fine beforehand, although I noticed that when installing Qt that I had to check the custom installation boxes for Sources and Qt Script.

### Screenshot or Video Reference
This is the output of calling make in the opentoonz/toonz/build folder. I think some of the build steps are missing because this is the second time trying to build.
```
[  7%] Built target tnzcore
[  7%] Linking CXX shared library libtnzbase.dylib
[ 10%] Built target tnzbase
[ 10%] Linking CXX shared library libtnzext.dylib
[ 12%] Built target tnzext
[ 12%] Linking CXX shared library libtoonzlib.dylib
[ 38%] Built target toonzlib
[ 38%] Linking CXX shared library libtfarm.dylib
[ 39%] Built target tfarm
[ 39%] Linking CXX shared library libtnzstdfx.dylib
[ 48%] Built target tnzstdfx
[ 48%] Linking CXX shared library libsound.dylib
[ 48%] Built target sound
[ 48%] Linking CXX shared library libcolorfx.dylib
[ 50%] Built target colorfx
[ 50%] Linking CXX shared library libimage.dylib
Undefined symbols for architecture x86_64:
  "_lzma_code", referenced from:
      _LZMADecode in libtiff.a(tif_lzma.o)
      _LZMAPostEncode in libtiff.a(tif_lzma.o)
      _LZMAEncode in libtiff.a(tif_lzma.o)
  "_lzma_end", referenced from:
      _LZMASetupDecode in libtiff.a(tif_lzma.o)
      _LZMASetupEncode in libtiff.a(tif_lzma.o)
      _LZMACleanup in libtiff.a(tif_lzma.o)
[...]

Issue #1566, opened 2017-11-08
Undefined symbols for architecture x86_64


damies13:
I have been getting this for a while... it stopped me from working on the issues that I was working on, so to confirm it wasn't me I switched back to the master branch and I get the same issue when trying to compile for OSX.

I had hoped it would be resolved quickly but after a month i'm still getting this when trying to compile from master even after getting the latest updates for master. I found that it has also been reported before as issue #40 (Compilation notes for OSX #40) back in 2016. I tried the resolutions listed in that issue, but it didn't seem to help.
I have since upgraded xcode and it's command line tools (in case that was the cause) and have also tried to use the "ci-scripts/osx/travis-install.sh" and "ci-scripts/osx/travis-build.sh" scripts but I still get the same error (below):

[ 51%] Linking CXX shared library libimage.dylib
Undefined symbols for architecture x86_64:
  "_lzma_code", referenced from:
      _LZMADecode in libtiff.a(tif_lzma.o)
      _LZMAPostEncode in libtiff.a(tif_lzma.o)
      _LZMAEncode in libtiff.a(tif_lzma.o)
  "_lzma_end", referenced from:
      _LZMASetupDecode in libtiff.a(tif_lzma.o)
      _LZMASetupEncode in libtiff.a(tif_lzma.o)
      _LZMACleanup in libtiff.a(tif_lzma.o)
  "_lzma_lzma_preset", referenced from:
      _TIFFInitLZMA in libtiff.a(tif_lzma.o)
      _LZMAVSetField in libtiff.a(tif_lzma.o)
  "_lzma_memusage", referenced from:
      _LZMADecode in libtiff.a(tif_lzma.o)
  "_lzma_stream_decoder", referenced from:
      _LZMAPreDecode in libtiff.a(tif_lzma.o)
      _LZMADecode in libtiff.a(tif_lzma.o)
  "_lzma_stream_encoder", referenced from:
      _LZMAVSetField in libtiff.a(tif_lzma.o)
      _LZMAPreEncode in libtiff.a(tif_lzma.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [image/libimage.dylib] Error 1
make[1]: *** [image/CMakeFiles/image.dir/all] Error 2
make: *** [all] Error 2

Is anyone else getting this when trying to build for OSX? Any suggestions?

[...]

PR #2497, opened 2019-01-25
Small typo fixes and clarifications in OS X build instructions

yehjaemin:
I noticed some small typos in the local build readme for Mac, so I created a pull request to fix them.

PR #2499, opened 2019-01-27
Fix for OS X build error

yehjaemin:
Referenced in closed issues #40, #1566, and #2496, the build process outputs an error message due to toonz/cmake/FindTIFF.cmake referencing libtiff.a instead of libtiff.dylib. This causes the build to crash. I applied the change mentioned in the threads and tested it on a fresh clone opentoonz repo and was able to successfully build the OpenToonz app executable (but I'm not sure if this change causes any problems for Windows/Linux).


yehjaemin:
I just checked the results of the Travis CI build and the changes work only for xcode. I'd be happy to revert the changes and edit the OS X build instructions to mention the libtiff.dylib fix.

yehjaemin:
Thanks @manongjohn! I think this commit might solve the different platforms problem.

yehjaemin:
@artisteacher The problem I had with the whole libimage.dylib error message wasn't from not having things installed but from FindTIFF.cmake not searching for the right file extension on mac environments. This PR is still pending, but you could test it yourself in the meantime by copying my commits into your file.

artisteacher:
The problem is that without these commits, I can get OT to build and open - though there are issues. With these commits, I get that error when trying to open.

yehjaemin:
@artisteacher I didn't experiment with Qt versions ( I just installed the one the guide told me to), but I did notice that installing "Sources" and "Qt Script" were necessary for me. I proposed it as a change to local build instruction in PR #2497.

artisteacher:
QT version doesn't seem to make a difference on my end. I definitely have both sources & qt script installed. Is there something else I could be missing?

manongjohn:
Try going to your 'build' directory, then delete the 'toonz' directory and then 'make' again and see if that helps. Changes to certain things outside of OT code does not cause some things to recompile when rebuilding OT and will continue to use old settings.

Btw, when you changed Qt version, i'm assuming you updated the scripts accordingly? If you did, you will definitely need to delete the toonz directory and reissue the CMAKE_PATH_DIR step.

artisteacher:
I made a completely new build - so I don't think that was the problem. I took a look and I don't actually have libtiff.5.dylib installed where OT is looking for it. I found a copy in another program I have and moved a copy to usr/local/lib  - which allows me to open the build. Is there a step I could have missed or done incorrectly that should have installed libtiff.5.dylib? Or is there a missing step in the directions?

artisteacher:
Alright, using homebrew to install libtiff makes this work. @yehjaemin, this PR is going to need an addition to the build instructions - so that libtiff is installed with the other dependent packages (Step 1).

shun-iwasawa:
@yehjaemin 
Please confirm that you did [the step 3 in the instruction](https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_build_macosx.md#3-build-tiff-in-thirdparty) .
It should generate `libtiff.a` under `opentoonz/thirdparty/tiff-4.0.3/libtiff/.libs` .

Libtiff used in OpenToonz is [slightly modified](https://github.com/opentoonz/opentoonz/blob/master/thirdparty/tiff-4.0.3/libtiff/tif_getimage_64.c) from the one you can install from Homebrew.

artisteacher:
I’ve been able to build successfully both ways. However following the normal build instructions, I always get a crash when trying to save. I don’t get a crash when saving with this pr.

shun-iwasawa:
@artisteacher 

> I always get a crash when trying to save.

You meant that you get a crash when you save TIF raster levels?
Can you please check if the crash occurs only with your build or also with the release version of OT ?
I tried saving raster levels and scenes in my environment of macOS Sierra and I couldn't reproduce the crash yet.

artisteacher:
I have not had issues saving with the released version - just local builds. I always use the "Save All" command, so I wan't sure if it was the scene or a specific level type causing the crash.

yehjaemin:
@shun-iwasawa Hi, sorry for the late response, my car broke down so I've been devoting my free time lately to getting a replacement.

I did step 3 as specified in the instructions, but I also included @artisteacher's libtiff edit because I assumed it didn't work for him. Installing without homebrewing libtiff worked fine for me.

RodneyBaker:
This PR needs to be looked at again due to nerger of PR #2600.
Perhaps it is no longer needed?

manongjohn:
I recommend closing this. Since we use a custom libtiff.a which is part of our normal build process, we shouldn't be using the homebrew version or there could be unintended side effects.

If the crashing problem noted here is still an issue, a new issue should be opened to track it. However, it seems related to local building and currently not in the official version. Is it possible it's due to some changes in libraries when compiling on macOS versions after Sierra (10.12)?

RodneyBaker:
Closing.
@yehjaemin Let us know if you have any concerns.

PR #2501, opened 2019-01-28
Viewer context menu additions

manongjohn:
This PR adds the following options to the Viewer's (canvas) context menu

1) Reset Rotation
User request to be able to reset the rotation back to normal without having to adjust zoom or flip.

2) Reset Zoom
User request to be able to reset the zoom back to normal without having to adjust rotation or flip.

NOTE: This is different from `Reset View`. Reset View reverts zoom, rotation and flip all back to normal which is not always ideal

3) Reset Position
Reset the screen back to center without having to adjust rotation, zoom, flip.

4) Flip Viewer Horizontally and Flip Viewer Vertically

Just so it's easily accessible without a keyboard

5) Enable the following visualization commands on Command Bar:
- Reset View
- Reset Zoom
- Fit to Window
- Actual Pixel Size
- Flip Viewer Horizontally
- Flip Viewer Vertically
- Reset Rotation
- Reset Position

6) Added appropriate reset button to tool option bar for Zoom, Rotate and Hand tools.

### DIFF
--- a/toonz/sources/tnztools/tooloptions.cpp
+++ b/toonz/sources/tnztools/tooloptions.cpp
@@ -2658,22 +2658,8 @@ ZoomToolOptionsBox::ZoomToolOptionsBox(QWidget *parent, TTool *tool,
   button->addAction(resetZoomAction);
   connect(button, SIGNAL(clicked()), resetZoomAction, SLOT(trigger()));
 
-  TPropertyGroup *props = tool->getProperties(0);
-  assert(props->getPropertyCount() > 0);
-
-  ToolOptionControlBuilder builder(this, tool, pltHandle, toolHandle);
-  if (tool && tool->getProperties(0)) tool->getProperties(0)->accept(builder);
-
   m_layout->addStretch(1);
   m_layout->addWidget(button, 0);
-}
-
-//-----------------------------------------------------------------------------
-
-void ZoomToolOptionsBox::updateStatus() {
-  QMap<std::string, ToolOptionControl *>::iterator it;
-  for (it = m_controls.begin(); it != m_controls.end(); it++)
-    it.value()->updateStatus();
 }
 
 //=============================================================================
@@ -2695,22 +2681,8 @@ RotateToolOptionsBox::RotateToolOptionsBox(QWidget *parent, TTool *tool,
   button->addAction(resetRotationAction);
   connect(button, SIGNAL(clicked()), resetRotationAction, SLOT(trigger()));
 
-  TPropertyGroup *props = tool->getProperties(0);
-  assert(props->getPropertyCount() > 0);
-
-  ToolOptionControlBuilder builder(this, tool, pltHandle, toolHandle);
-  if (tool && tool->getProperties(0)) tool->getProperties(0)->accept(builder);
-
   m_layout->addStretch(1);
   m_layout->addWidget(button, 0);
-}
-
-//-----------------------------------------------------------------------------
-
-void RotateToolOptionsBox::updateStatus() {
-  QMap<std::string, ToolOptionControl *>::iterator it;
-  for (it = m_controls.begin(); it != m_controls.end(); it++)
-    it.value()->updateStatus();
 }
 
 //=============================================================================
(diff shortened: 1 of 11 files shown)


artisteacher:
If the options were added there, perhaps there would be a reset button - perhaps similar to a refresh button - that could have the list of reset options in a drop down. Otherwise there could be a lot of buttons.

manongjohn:
Technically these can be added as buttons to the customizable Command Bar. I normally keep a command bar just above my viewer with buttons for various drawing actions I use a lot (undo/redo, copy/paste/delete, grouping, etc). I would add these as well  but, for some reason, these buttons never get enabled for me when I add it.  I plan on looking into why it doesn't and try to fix that,

manongjohn:
Actually, one can argue that these would be more useful on the title bar of Viewer/Combo Viewers near the other buttons like Freeze, etc.  Would need icons for these options though...not that I'm saying I would implement this...might look into it though :)

openanim:
@manongjohn Thank you for these helpful commands.

Some suggestions if possible for additional views :
- fit to the drawings / (contents of the column)
- fit to the selection (when we using selection tool or control point editor tool, view will target the selection)
- Reset Zoom should also to fit the drawing.

artisteacher:
I can make some icons if they will be needed - just let me know which ones.

gab3d:
@manongjohn thanks for considering it :)
> Actually, one can argue that these would be more useful on the title bar of Viewer/Combo Viewers near the other buttons like Freeze, etc.

This could also be useful (and certainly very noticeable), but perhaps at the cost of some extra cluttering in the UI.


@artisteacher 
> If the options were added there, perhaps there would be a reset button - perhaps similar to a refresh button - that could have the list of reset options in a drop down. Otherwise there could be a lot of buttons.

I don't fully understand where in the UI you are proposing to add that dropdown list.
What I meant was to add one **Reset** button inside each of the Zoom, Rotate and Hand tools' toolbar (which are currently mostly unpopulated). This would hardly make the UI any heavier and it would ease the learning curve of the program a bit.

toonz/sources/toonz/sceneviewercontextmenu.cpp
```diff
@@ -81,8 +81,18 @@ SceneViewerContextMenu::SceneViewerContextMenu(SceneViewer *parent)
...
+  // flip horizontally
+  action = commandManager->createAction(V_FlipX, this);
+  addAction(action);
+  ret = ret && parent->connect(action, SIGNAL(triggered()), SLOT(flipX()));
+
+  // flip vertically
+  action = commandManager->createAction(V_FlipY, this);
+  addAction(action);
+  ret = ret && parent->connect(action, SIGNAL(triggered()), SLOT(flipY()));
+
```
shun-iwasawa: I'm a bit reluctant to put many commands on the top layer of the context menu.
How about putting flip horizontally / vertically commands into "Flip View" submenu, and putting all "Reset xxx" commands into "Reset View" submenu like as follows?:
image

toonz/sources/tnztools/tooloptions.cpp
```diff
@@ -2639,6 +2640,116 @@ void ShiftTraceToolOptionBox::onAfterRadioBtnClicked() {
...
+  QPushButton *button = new QPushButton(tr("Reset Zoom"));
+  button->setFixedHeight(20);
+  button->addAction(resetZoomAction);
+  connect(button, SIGNAL(clicked()), resetZoomAction, SLOT(trigger()));
+
+  TPropertyGroup *props = tool->getProperties(0);
+  assert(props->getPropertyCount() > 0);
+
+  ToolOptionControlBuilder builder(this, tool, pltHandle, toolHandle);
+  if (tool && tool->getProperties(0)) tool->getProperties(0)->accept(builder);
```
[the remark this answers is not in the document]
manongjohn: Ok. Will remove those from all three classes.

toonz/sources/toonz/sceneviewercontextmenu.cpp
manongjohn: I will condense into submenus as requested.

If you are ok with the proposed changes in #2540,  I can remove the Flip Viewer commands in this context menu as they will be in a more easily accessible position.

shun-iwasawa:
Jenkins again

shun-iwasawa:
LGTM Thank you @manongjohn for the addition! 
Let's consider whether to keep or remove the Flip commands after introducing #2540.

PR #2509, opened 2019-02-05
Fix Output Glitch When the Channel Width Mismatches the Bit Per Pixel

### LINKED ISSUE #2508: Output Image Corrupts When the Channel Width Mismatches the Bit Per Pixel
### Issue Summary

When rendering images, if the `Channel Width` value in the output settings does not match with the `Bit Per Sample` value in format properties of some format (such as TIF), the output image corrupts as shown in the image below.

This issue newly appears in V1.3, after introducing the Matte Color option ( #2418 ).

image

### Steps to Reproduce

1. Open output settings
1. Set the `Channel Width` to 16bit
1. Set the output format to TIF
1. Open the format option and set the `Bit Per Pixel` value to 24(RGB)
1. Render

### Expected Results

The image ( with 3 x 8bit channels ) should be rendered normally.

### Actual Results

The rendered image corrupts.

### System Information
* **OpenToonz Version**: 1.3
* **Operating System**: Win7

shun-iwasawa:
This PR fixes #2508

### DIFF
--- a/toonz/sources/common/timage_io/timage_io.cpp
+++ b/toonz/sources/common/timage_io/timage_io.cpp
@@ -634,14 +634,14 @@ void TImageWriter::save(const TImageP &img) {
     // add background colors for non alpha-enabled image types
     if ((ras32 || ras64) && !writer->writeAlphaSupported() &&
         TImageWriter::getBackgroundColor() != TPixel::Black) {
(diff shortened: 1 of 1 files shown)


masafumi-inoue:
LGTM

Issue #2303, opened 2018-09-26
Pli custom styles cause crash on switch to vector tab


artisteacher:
### Issue Summary

I'm getting a regular crash when switching to the vector tab in the style editor if it is the **first time** switching to the vector tab **and a raster brush** is selected in the level palette. 
### Screenshot or Video Reference

[vector_tab_crash.txt](https://github.com/opentoonz/opentoonz/files/2421187/vector_tab_crash.txt)

### Steps to Reproduce

1. Open OpenToonz and  open a scene with at least one raster or toonz raster level
2. Select a raster (or toonz raster) frame in the Xsheet or Timeline
3. With the brush active, select a raster (mypaint) brush style in the level palette (create one if needed)
4. With a raster brush and raster (or toonz raster level) still active, switch the style editor to the vector tab = crash

### Expected Results

No crash when switching tabs.
### Actual Results

A crash if the vector tab has not been opened already. If a non-mypaint brush is active in the level palette there is no crash. If the vector tab has already been open there is no crash.
### System Information
* **OpenToonz Version**: Nightly build & ongoing for some time
* **Operating System**: MacOS 10.13.5 & 10.14
* **CPU**: 
* **Memory**: 
* **Graphics Card**: 
* **Graphics Tablet**:

artisteacher:
The crash is sometimes harder to trigger with 10.14. It is most consistent if the selected raster style is changed to a different raster brush before switching to the vector tab.

artisteacher:
I’m still trying to narrow this down but think it is likely related to a problem .pli in the custom styles folder - similar to #201.

artisteacher:
If I temporarily remove all of the png files from the custom styles folder, it’s easier to trigger this crash on demand, though it is still not an every time thing. 

Steps to reproduce:
1. Remove png files from custom styles folder 
2. Create a new project & scene (I’m using an alternate project location - Documents)
3. Create a Raster level & set the brush to the charcoal brush on the raster tab
4. Draw with the brush & save scene 
5. Switch to the vector tab
6. If there is no crash, close & reopen OT & the scene
7. Click on the raster style in the level palette & then select the vector tab
8. Repeat steps 6 & 7 as necessary 
In case it makes a difference, I’m using a touchpad, not a mouse.

Sometimes the crash happens without needing to repeat the last steps, sometimes it happens every other time, sometimes it happens several times in a row. At the very best, after 7 times or so, there is always a crash.

[...]

PR #2513, opened 2019-02-07
Crash fix for issue #2303

yehjaemin:
Some portions of the CustomStyleManager::StyleLoaderTask::run() method
caused OT to crash when switching to the vector tab. I commented these
portions out (3 in total) and labeled them as problem, problem
continued, etc. I'm not sure how much functionality these edits took
out, but after some testing I documented what I noticed in my comments
in issue #2303.

### DIFF
--- a/toonz/sources/toonzlib/stylemanager.cpp
+++ b/toonz/sources/toonzlib/stylemanager.cpp
@@ -90,6 +90,7 @@ void CustomStyleManager::StyleLoaderTask::run() {
     TRasterImageP rimg = img;
 
     TRaster32P ras;
+
     if (vimg) {
       assert(level->getPalette());
       TPalette *vPalette = level->getPalette();
@@ -97,9 +98,13 @@ void CustomStyleManager::StyleLoaderTask::run() {
       vimg->setPalette(vPalette);
 
       TOfflineGL *glContext = 0;
+// -------------------
+// PROBLEM STARTS HERE
+// -------------------
+/*
       glContext             = TOfflineGL::getStock(chipSize);
-
       glContext->clear(TPixel32::White);
+*/
       TRectD bbox = img->getBBox();
       double scx  = 0.8 * chipSize.lx / bbox.getLx();
       double scy  = 0.8 * chipSize.ly / bbox.getLy();
@@ -111,17 +116,21 @@ void CustomStyleManager::StyleLoaderTask::run() {
           TTranslation(dx, dy) * TScale(sc) *
           TTranslation(-0.5 * (bbox.x0 + bbox.x1), -0.5 * (bbox.y0 + bbox.y1));
       TVectorRenderData rd(aff, chipSize, vPalette, 0, true);
-
+// -----------------
+// PROBLEM CONTINUED
+// -----------------
+/*
       glContext->draw(img, rd);
 
       // No need to clone! The received raster already is a copy of the
       // context's buffer
       ras = glContext->getRaster();  //->clone();
+*/
     } else if (rimg) {
       TDimension size = rimg->getRaster()->getSize();
-      if (size == chipSize)
+      if (size == chipSize) {
         ras = rimg->getRaster()->clone();  // Yep, this may be necessary
-      else {
+      } else {
         TRaster32P rout(chipSize);
 
         TRop::resample(rout, rimg->getRaster(),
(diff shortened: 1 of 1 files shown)


artisteacher:
I can confirm that this stops the crash when building on a mac and switching to the vector tab. Thanks @yehjaemin for finding the problem areas. It would be amazing if this could fix the crash without losing access to the trail styles and vector brushes.

manongjohn:
I don't believe this is the correct fix.  Commenting out the offending code makes Trail and Vector Brushes no longer available. 

The real error, I'm guessing, is that on some Macs, the OfflineGL context isn't being created before it is used when attempting to render these icons and prepare these brushes for use later.

yehjaemin:
@manongjohn I spent some time trying to fix the line
```
glContext = TOfflineGL::getStock(chipSize);
```
in particular. This lead me to tofflinegl.cpp and qtofflinegl.cpp, where I found the ```QTOfflineGL::createContext``` method. I tried (without the comments) changing a code block from
```
  m_surface = std::make_shared<QOffscreenSurface>();
  m_surface->setFormat(format);
  m_surface->create();

  m_context = std::make_shared<QOpenGLContext>();
  m_context->setFormat(format);
  m_context->create();
  m_context->makeCurrent(m_surface.get());
```
to
```
  m_context = std::make_shared<QOpenGLContext>();
  m_context->setFormat(format);
  m_context->create();

  m_surface = std::make_shared<QOffscreenSurface>();
  m_surface->setFormat(format);
  m_surface->create();

  m_context->makeCurrent(m_surface.get());
```
and couldn't get the crash to go away. Also, after going through the threads on on the QT debugger, I found that the thread that calls CustomStyleManager::StyleLoaderTask::run() sometimes crashes on in this order:

 run() -> ... -> TOfflineGL::getStock -> ... -> QTOfflineGL::createContext

and sometimes in this order:

run() -> ... -> TLevelReader::loadinfo

I through the inconsistency could be a sign that the code might have been working fine, but the threads weren't interacting properly. Do you think there's any validity to this?

manongjohn:
This is likely something that @shun-iwasawa will need to look into as I'm not well versed with the OpenGLContext and OffScreenSurface stuff.  Maybe something to do with how macOS threading is working in more recent OS versions?

I'm curious to know if this may be an issue with macOS and/or whatever version of QT you are building with.  I don't know if it's just a weird correlation, but it seems since El Capitan and later macOS versions, there have been oddities in how OT and/or QT behaves.  Stuff that hasn't been touched, necessarily, has started breaking or not working correctly, like pen hovering.  

I tried to reproduce this issue with El Capitan but couldn't, unless I was not doing it right. Could this issue be somehow specific to High Sierra or later? 

Out of curiosity, can you build with an older QT version, like 5.9.7 that Windows currently uses and see if the problem is still there?  I suspect it will be.

I'd be interested to know if installing older versions of OT where we didn't see this problem before, is now a problem on recent versions of macOS.

yehjaemin:
@manongjohn I was using the latest Qt version (5.11 i think) and OSX 10.14.2 Mojave. I think OpenGL was deprecated on Mac in favor of Metal somewhat recently, although I'm not sure how big a problem that really is. I agree that it seems like the os updates on macs are exacerbating the bugs, but at the same time OpenToonz 1.2 works fine on Mojave (doesn't crash on switching to the vector tab, lists all generated/trail/vector brush presets). So in the end I think the bugs are originating from changes to the source code.

I'll try testing with older versions of QT over the weekend and report back in a day or two.

artisteacher:
@yehjaemin - my students and I have had crashing when switching to the vector tab before the 1.3 release both with Mojave and Sierra but it has never consistently repeatable.

yehjaemin:
@artisteacher sorry about the delayed response - I think it might be more likely that it is a multithreading error if it's been an inconsistent issue across multiple systems. I'm still not sure but thanks for the information.

artisteacher:
@yehjaemin - it's great to hear from you! I wonder if that's what causes a lot of the sporadic crashes that are hard to track?

yehjaemin:
@artisteacher maybe, but I don't have enough knowledge on the whole codebase to be able to say for sure (sorry!) I remember thinking it was strange that when I was reproducing the crash on debug mode in Qt, the thread logs kept changing from case to case. I don't think a single UI bug on its own would behave like that, but I'm not sure.

shun-iwasawa:
@yehjaemin @artisteacher 
Thank you very much for tackling with the issue. I pushed #2634 as an alternative to this PR. Please take a look. Thanks!

PR #2514, opened 2019-02-07
Backup scene file

manongjohn:
Currently there is a level-only backup, but that preference is OFF by default.  Considering OT's instability, being OFF by default is not a good thing. In my opinion, it should be ON by default.

Additionally, the scene file is not backed up and it really should do that when saving also.  I've seen 1 too many crashes by users that resulted in level(s) and scene files being corrupted and lots of work lost.

This PR will somewhat address these issue with the following changes:

1) Modified the backup Preference in the General tab to say `Backup Scene and Animation Levels when Saving`.
2) Set the backup preference setting to ON by default.
3) A simple copy of the `.tnz` file to a  `_backup.tnz` file, similar to how levels are backed up, will be created when saving scenes.

NOTE: I have intentionally changed how backup preference is saved in the preference file, making the old setting obsolete. Therefore, once upgraded to a version that includes this, backups will be created by default until they turn it off.

### DIFF
--- a/toonz/sources/toonzlib/toonzscene.cpp
+++ b/toonz/sources/toonzlib/toonzscene.cpp
@@ -171,8 +171,26 @@ void deleteUntitledScene(const TFilePath &fp) {
 //-----------------------------------------------------------------------------
 
 static void saveBackup(TFilePath path) {
+  int totalBackups = Preferences::instance()->getBackupKeepCount();
+  totalBackups -= 1;
+  TFilePath backup = path.withType(path.getType() + ".bak");
+  TFilePath prevBackup =
+      path.withType(path.getType() + ".bak" + std::to_string(totalBackups));
+  while (--totalBackups >= 0) {
+    std::string bakExt =
+        ".bak" + (totalBackups > 0 ? std::to_string(totalBackups) : "");
+    backup = path.withType(path.getType() + bakExt);
+    if (!TSystem::doesExistFileOrLevel(backup)) continue;
+    try {
+      if (TSystem::doesExistFileOrLevel(prevBackup))
+        TSystem::removeFileOrLevel_throw(prevBackup);
+      TSystem::copyFileOrLevel_throw(prevBackup, backup);
+    } catch (...) {
+    }
+    prevBackup = backup;
+  }
+
   try {
-    TFilePath backup = path.withType(path.getType() + ".bak");
     if (TSystem::doesExistFileOrLevel(backup))
       TSystem::removeFileOrLevel_throw(backup);
     TSystem::copyFileOrLevel_throw(backup, path);
--- a/toonz/sources/toonzlib/toonzscene.cpp
+++ b/toonz/sources/toonzlib/toonzscene.cpp
@@ -180,12 +180,11 @@ static void saveBackup(TFilePath path) {
     std::string bakExt =
         ".bak" + (totalBackups > 0 ? std::to_string(totalBackups) : "");
     backup = path.withType(path.getType() + bakExt);
-    if (!TSystem::doesExistFileOrLevel(backup)) continue;
-    try {
-      if (TSystem::doesExistFileOrLevel(prevBackup))
-        TSystem::removeFileOrLevel_throw(prevBackup);
-      TSystem::copyFileOrLevel_throw(prevBackup, backup);
-    } catch (...) {
+    if (TSystem::doesExistFileOrLevel(backup)) {
+      try {
+        TSystem::copyFileOrLevel_throw(prevBackup, backup);
+      } catch (...) {
+      }
     }
     prevBackup = backup;
   }
--- a/toonz/sources/toonzlib/toonzscene.cpp
+++ b/toonz/sources/toonzlib/toonzscene.cpp
@@ -169,71 +169,16 @@ void deleteUntitledScene(const TFilePath &fp) {
 }
 
 //-----------------------------------------------------------------------------
-/*-- TODO: オプション化して復活させるか、検討のこと --*/
-/*
-void saveBackup(const TFilePath &fp)
-{
-  if(!TFileStatus(fp).doesExist()) return;
-        wstring sceneName = fp.getWideName();
-  TFilePath bckDir = fp.getParentDir() + "backups" + sceneName;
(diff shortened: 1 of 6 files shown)


artisteacher:
This is so, so needed. Thank you!!!

turtletooth:
Thanks for doing this.  What do you think about keeping the last 5 or so backups?  That might help if something goes really wrong.  I haven't actually had corrupted scenes, so I don't know if that would be useful or not.

konero:
Does Toonz show backup files in the browser window? The projects I worked on usually had hundreds of scene files in the scenes folder, so that'd be a pain. _I don't use the backup feature, so I don't know._

The software I've used in the past with auto backups used a unique file extension. Something like `filename.tnz.bak` or `filename.tnz~` so they didn't display in browser windows when opening files and in the Windows Explorer window they'd just have a blank white document icon since they weren't associated with any program to open with, making it easier to find a file and less prone to opening a backup file by mistake.

manongjohn:
Right now, to keep space usage low I'd only want to keep the most recent.  If we think it's a good idea, I can make a Preference setting of how many backups to keep, defaulting to 1.

Level files only backup whenever there is an actual change to the level, but currently scene files will backup every time there is a save operation (current behavior).

Current backup naming convension simply adds `_backup` to the name (i.e level.pli -> level_backup.pli, scene.tnz -> scene_backup.tnz).  Because they will still have typical OT extensions, they will show up as normal in browser windows.  I personally would also not want to see them in OT browsers.

I am also not fond of adding `_backup` to the filename.  I would prefer to add an additional extension `.bkp`, `.bkup' or something like that (i.e level.pli.bkup).  I confirmed that would prevent them from showing up in OT's browser windows.

Unless someone has a really good reason to keep the current backup naming convention, I can change it to add an additional backup extension instead.

artisteacher:
How would you restore from a backup if they don't show up in the browser window?  Would we just rename the backup in Finder (or etc.) to remove the backup extension?

While transitioning to 1.3, I've had a couple students whose video files have disappeared on re-opening the scene (for rotoscoping). We've made duplicates of the backups & then renamed them to the original name to restore.

manongjohn:
Yes. Users would just need to remove the backup extension and then they can load it.

artisteacher:
Eventually, it would be great for that step to be automated. When OT loads a scene and recognizes a level is missing, if there is a backup level, a pop-up window should ask to restore from the backup & automatically remove the backup extension.

If additional backups are added at some point, maybe they should be moved to a backups folder.

toonz/sources/toonzlib/toonzscene.cpp
```diff
@@ -169,71 +169,34 @@ void deleteUntitledScene(const TFilePath &fp) {
...
+  int totalBackups = Preferences::instance()->getBackupKeepCount();
+  totalBackups -= 1;
+  TFilePath backup = path.withType(path.getType() + ".bak");
+  TFilePath prevBackup =
+      path.withType(path.getType() + ".bak" + std::to_string(totalBackups));
+  while (--totalBackups >= 0) {
+    std::string bakExt =
+        ".bak" + (totalBackups > 0 ? std::to_string(totalBackups) : "");
+    backup = path.withType(path.getType() + bakExt);
+    if (!TSystem::doesExistFileOrLevel(backup)) continue;
```
shun-iwasawa: If you set the number of backups to 5, currently only `.bak` and `.bak4` are saved.
I think you need to call `prevBackup = backup;` before calling `continue;` here. 

The same goes for the function in `txshsimplelevel.cpp` .

toonz/sources/toonzlib/toonzscene.cpp
```diff
@@ -169,71 +169,34 @@ void deleteUntitledScene(const TFilePath &fp) {
...
+    try {
+      if (TSystem::doesExistFileOrLevel(prevBackup))
+        TSystem::removeFileOrLevel_throw(prevBackup);
```
shun-iwasawa: I think you don't need to check existence of `prevBackup` and remove it here, because the subsequent function `TSystem::copyFileOrLevel_throw` will do the overwrite copy.

The same goes for the function in `txshsimplelevel.cpp` .

toonz/sources/toonzlib/txshsimplelevel.cpp
```diff
@@ -1374,8 +1374,26 @@ void TXshSimpleLevel::save() {
...
+    } catch (...) {
+    }
+    prevBackup = backup;
+  }
+
   try {
-    TFilePath backup = path.withName(path.getName() + "_backup");
     if (TSystem::doesExistFileOrLevel(backup))
       TSystem::removeFileOrLevel_throw(backup);
     TSystem::copyFileOrLevel_throw(backup, path);
```
shun-iwasawa: Currently it fails to save the backup properly when you save the raster levels (like `A..tif` ).

Since the backup extension has changed to `.bak` , OT seems to fail copying the level here as the target level name becomes with the single period ( like `A..tif.bak` ), which is recognized as single image.

manongjohn: The individual frames are still saving but on new levels being saved for the 1st time, it creates an odd file because of the sequence range.

This appears to be an existing bug. In the current version, I observed the files would be saved as A_backup.0001.tif, A.0001.tif, A.0002.tif, etc. If you delete A_backup.0001.tif and make changes to the frames then save, A_backup.0001.tif does not get created.

I will fix this bug with this release.

toonz/sources/toonzlib/toonzscene.cpp
manongjohn: That logic mimics the logic that was already there in the original saveBackup logic. I can certainly remove it.

toonz/sources/toonzlib/txshsimplelevel.cpp
manongjohn: I just discovered why this does not work for sequenced files. The change in actual extension causes OT to no longer acknowledge jt as sequenced files.  I'll need to figure out the best way to deal with this.

shun-iwasawa:
LGTM Thank you @manongjohn !

PR #2517, opened 2019-02-10
Fix Toonz Raster brush style update crash

### LINKED ISSUE #2516: Crash loading palette into Toonz Raster
### Issue Summary
OT crashes when you load another palette over an existing palette that has a Mypaint brush as the current selected style and you attempt to use it immediately.

### Steps to Reproduce
1) Open OT and create a new scene
2) Add a Toonz Raster Level
3) Select Brush tool
4) Select Color 1 and change to a MyPaint Brush style (I used Brush)
5) Open Studio Palette and select Global Palette -> basic
6) Right-click `basic` palette and `Load into Current Palette`
7) Move pointer directly from Studio Palette to Canvas and move around or attempt to draw-> Crash

This does not happen with Vector or standard Raster levels.

### System Information
* **OpenToonz Version**:  1.3

manongjohn:
This PR fixes #2516.

When updating the current style, if it thinks it's a MyPaint brush but actually is not, it will complete a color style change, internally, and allow it to continue as a normal Toonz Raster brush.


shun-iwasawa:
Jenkins

shun-iwasawa:

PR #2518, opened 2019-02-11
Palette switching behavior changes

manongjohn:
This PR changes the behavior when switching palettes as follows:

1) Tracks and restores current style when switching levels

User request to restore the last style used on a level palette when switching away and back to that level.  This is done at a palette level, not based on column.

2) Switch to style page of current style when restored

Goes along with the above.  You are switched to the style page where the restore current style is found.

3) Auto switch to new page in palette

When creating a new page in a palette, you are automatically switched to the blank page after it is created.

4) Auto switch to drag-dropped merged palette

When dragging a palette between a Studio Palette and a Level Palette to merge them, you will be automatically switched to the page of the newly dropped palette. In the case of a dropped palette with multiple pages, you are switched to the first page added.

This currently does not apply to the "Merge into Current Palette" right-click context menu in the Studio Palette Viewer.


toonz/sources/toonzlib/tpalettehandle.cpp
```diff
@@ -130,6 +130,13 @@ bool TPaletteHandle::disconnectBroadcasts(const QObject *receiver) {
 //-----------------------------------------------------------------------------
 
 void TPaletteHandle::setPalette(TPalette *palette, int styleIndex) {
+  if (palette) {
+    if (styleIndex < 0)
+      styleIndex = palette->getCurrentStyleId();
```
shun-iwasawa: Currently, it seems that the current style can still be reproduced even if it is removed. Please try the following steps:
- Create a new level and create a style # 2 (or, copy the style # 1 and insert-paste on the same page) in the palette. 
- Make sure that the current style index is # 2.
- Switch the current level to another one.
- Undo (= style # 2 is removed from the palette).
- Switch the current level back to the previous one.

How about adding lines to check if the style is removed like as follows ?:

```
 void TPaletteHandle::setPalette(TPalette *palette, int styleIndex) {
   if (palette) {
-    if (styleIndex < 0)
+    if (styleIndex < 0) {
       styleIndex = palette->getCurrentStyleId();
-    else
+      if (!palette->getStylePage(styleIndex)) {  // in case the style is deleted
+        styleIndex = 1;
+        palette->setCurrentStyleId(1);
+      }
+    } else
       palette->setCurrentStyleId(styleIndex);
   }
```

shun-iwasawa:
Jenkins

shun-iwasawa:
Thank you @manongjohn for the useful modification! LGTM

PR #2507, opened 2019-02-03
Show project name in title bars and startup menu

manongjohn:
This PR displays the Project name next to the Scene name in the following areas

1) Title bar

image

2) Viewer (shown below), Combo Viewer, Xsheet

image

3) Startup popup - recent list

image

NOTE: For existing recent history, the Project name will not display since it doesn't know where it is yet.  Once the scene is opened, it will capture the project name for displaying in the future.

PR #2520, opened 2019-02-13
Startup popup changes

manongjohn:
This PR attempts to clarify some issues/confusion with the startup popup when first opening OT. There have been a few complaints and confused users as to the relationship between the different boxes and why the Import or Change Project popup unexpectedly appears when opening a recent scene

Please note these changes are built on top of PR #2507 which can be cancelled if this is merged first.

The changes include the following

image

1) Changed `Choose Project` to `Current Project`

The moment you open OT you are always in a project, and this field always reflects the project you were last and currently in.  Changing it here is the same as changing it in File -> Project Settings..., immediately causing you to switch projects without realizing it.

2) Changed `Open Scene` to `Recent Scenes`

3) When opening a recent scene, you will automatically switch to the Project the selected scene belongs to, instead of the Project currently selected in the `Current Project`.

Current behavior always attempts to open the recent scene against the Current Project selected.  In most cases, when going back to what you were working on previously, it will be the same.  When they are not is when you get that message to Import or Switch Projects which can lead to some confusion.

4) **REMOVED 6/4:** Added checkbox `Import into Current Project` to be used when you want to open and import a recent scene  into the selected `Current Project`.

This checkbox is OFF by default when OT starts up and its setting is not saved when OT is shut down.

This setting does not apply to scenes loaded via the `Open Another Scene...` button.


RodneyBaker:
The appveyor is looking good.
Very useful addition.

turtletooth:
I'm not sure that No 4 is needed, and I'm afraid it adds unneeded complexity.  I don't see it being a common enough task to include it in the startup popup.  Thoughts?

manongjohn:
I'm ok with or without it. It's there because you can do this now, somewhat inadvertantly if not paying attention. The option is just to provide a little more control over when that happens.

artisteacher:
On the one hand, it's usually not a common task and there are a couple ways to import scenes. Visually, the popup may be a bit more balanced without the option. For new users/students, simpler may be better.

On the other hand, in those rare circumstances you do want to import a recent scene into a different project, it would be handy having the option more visible.  Removing the option would add more steps for anyone who does want to import a recent scene when they first open OT.

turtletooth:
I understand the rare use case, but with it being such a rare use case, I think the simplicity of the startup popup wins at least in my opinion.  I also agree that aesthetically it is also more balanced without it.

turtletooth:
The code looks fine for this.  I would still prefer for the import into current project option to be removed.  I feel that one of the downfalls of open source software can be that everything can be an option instead of having some design choices made for the sake of simplicity.  If something shows up in the recent scenes list, it is already associated in their mind with the project they were working on.  They aren't likely to get upset or be surprised when the scene loads in that project.

manongjohn:
I will remove the checkbox to help minimize any confusion this might bring up on the startup dialog.

turtletooth:
LGTM.  Thanks @manongjohn for the update.

PR #2521, opened 2019-02-15
Subcamera Option for Camera Capture

shun-iwasawa:
This PR will add a new option to the camera capture feature; Subcamera option.
It can specify a sub-region to be captured in order to minimize file size and to adjust capturing area easily.

image

- The subcamera option is available from `Subcamera` button next to the resolution combo box and the gear button on top of the camera capture popup.
- Once activated, magenta frame appears on the video pane which can be dragged to adjust.
- When the subcamera option is active, captured images will be placed on the xsheet with their dpi being automatically adjusted so that the images fit to the current camera frame (* See the note below).

*PLEASE NOTE:* As an exception, if the current unit is set to pixel the captured images dpi will not be adjusted and always saved with the same dpi as the current camera.

Also this PR modifies the camera capture feature in the points as follows:

- Altered capturing process (from `QCameraImageCapture` based to `QAbstractVideoSurface` based) so that the temporary image is no longer saved.
- Enhanced color adjusting feature by introducing multi-thread processing.


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins again

turtletooth:
This is awesome!  Thanks @shun-iwasawa 
Without the temp image being saved to disk, do you know if this reduces some of the lag that happened on slower machines?

shun-iwasawa:
@turtletooth 

> Without the temp image being saved to disk, do you know if this reduces some of the lag that happened on slower machines?

Yes, but unfortunately the speed-up was not much as I expected. 
In my test environment I realized that the capturing speed seems to be rather depended on the camera specification (i.e. shutter speed and processing speed of the camera).
With slower machines the speed-up may be more perceivable.

turtletooth:
Thanks for the feedback.  I think that at least some users will be glad that the camera capture isn't pounding on their hard drive anymore.

turtletooth:
I'm going to test this tomorrow.

konero:
@shun-iwasawa Did you miss the `*.less` file changes?

shun-iwasawa:
Thank you @konero for the notification of missing LESS file ! I added it.
Jenkins again

masafumi-inoue:
LGTM

PR #2534, opened 2019-02-27
Fix autosave pen pressure loss

manongjohn:
This fixes an issue reported by a Windows user where the pen pressure sensitivity is sometimes lost after an autosave.  This can happen when quickly drawing around the time autosave happens.

It's a bit difficult to recreate this issue repeatedly as it requires precise timing and a bit of luck but I was able to do it enough times to find the issue. Basically when quickly drawing marks, if an autosave happens just prior to a `TabletRelease` event, the logic keeping track of the tablet state gets locked into a `Touched` state from a series of unexpected tablet events that follow. This state causes the viewer to ignore pressure information, and default to drawing with a mouse.  Once in this bad state, the only way past it is to open a new viewer or restart OT.

Added a quick check for this scenario and force it to move to the next tablet state so subsequent events can process correctly.  This fix was only made for non-macOS systems for now.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -279,6 +279,8 @@ void SceneViewer::tabletEvent(QTabletEvent *e) {
         initToonzEvent(mouseEvent, e, height(), m_pressure, getDevPixRatio());
         m_tabletState = Touched;
         onPress(mouseEvent);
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Although I couldn't reproduce the proposed problem yet, I'll merge this as it seems to have no bad side effect and seems to work fine.

PR #2536, opened 2019-02-28
Fix Updating Scene Settings Popup Checkbox

shun-iwasawa:

When you load scene, `Enable Column Color Filter and Transparency for Rendering` checkbox in the scene settings popup does not update with the loaded scene property.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2537, opened 2019-03-01
Fix Flipbook Color Mask When Viewing TLV Files

shun-iwasawa:

- When viewing TLV files by using `View...` command, `Red/Green/Blue Channel in Grayscale` buttons and `Alpha Channel` button in the Flipbook console does not take effect.

Please note that the same problem occurs on viewing PLI files as well but not resolved in this PR.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

Issue #2523, opened 2019-02-17
Discussion: Updating default preferences for new users


manongjohn:
Watching a number of users on Twitch and other platforms struggling to learn to use OT for the first time, I tend to see some common issues that can be addressed if the default Preference setting or some OT window property's default was different.

I'm starting this discussion and list with the eventuality of submitting a PR to change the defaults with new users in mind.

I am not looking to create new preferences but modify existing default settings that will mostly affect new users and may affect some existing users if they've never changed that particular setting before.

Suggested changes:

- [ ] Preference: General -> Backup Animation Levels when Saving: Enabled
  - I have another PR that will already change this as well as backup Scenes at the same time
- [ ] Preference: Import/Export -> ffmpeg Timeout: 600
- [ ] Preference: Drawing -> Default Level Type: Toonz Raster Level
  - This will likely be debated, but with the implementation of MyPaint brushes on Toonz Raster levels, I think this might make a better starting experience for new users.
- [ ] Preference: Drawing ->  Autocreation: Use Xsheet as Animation Sheet
  - Suggested because people get confused why all of a sudden a new level is created when drawing
- [ ] Preference: Xsheet -> Cell-dragging Behaviour:  Cells and Column Data
- [ ] Preference: Xsheet -> Show Toolbar in the XSheet: Enabled
- [ ] Preference: Xsheet -> Use Arrow Key to Shift Cell Selecion: Enabled
- [ ] Preference: Animation -> Default Interpolation:  Ease In/Ease Out

Please feel free to suggest any other default preference or window option settings changes you feel will benefit new users.

RodneyBaker:
General thoughts/feedback:
The primary concern I have with Preferences in general is that all should be sticky so that once changed they will remain in that state until changed by the user or overwritten during install (as optional by user during install)

> Preference: General -> Backup Animation Levels when Saving: Enabled
> - another PR that will already change this as well as backup Scenes at the same time

I like the direction here (as outlined in #2514 and can see where having this on by default
I'll add a comment in that PR relating a few concerns about the backup process itself.

>   Preference: Import/Export -> ffmpeg Timeout: 600

[...]

PR #2540, opened 2019-03-01
Add view modifier buttons to viewer control panel

manongjohn:
This PR adds the following buttons to the Viewer and ComboViewer control panel.

image

- Zoom In
- Zoom Out
- Flip Viewer Horizontally
- Flip Viewer Vertically
- Reset Viewer

These can be hidden by unchecking `Viewer Controls` as needed:

image


konero:
I don't like that placement to be honest. I think they're going to get in the way since they sit in a high traffic zone (cam/preview) but are rarely used commands themselves. I use all the other buttons a lot, so I can't help but feel this is adding clutter.

Would adding them to the top menu (under View?) not solve any issue? For example like _View > Canvas > All canvas commands here_?

I strongly feel they'd be much better in the Command Bar though, that way it's to the users preference but it's where you usually find them (I have it there in Clip Studio, for example).

csp_v

RodneyBaker:
Just weighing in with a contrary opinion here to say it looks great in that position @manongjohn and I think the placement there would be very useful.

@konero  Not trying to step on any toes here.  On a scale of 1 to 10... how strongly do you feel about that location?  If more than 7... heck... if you say 6... I'll defer.  :)

manongjohn:
The problem is no ones know the command currently exists. I caught a stream where the user wished she could flip the screen. I was not on to let them know how. 

This was an attempt to make it more obvious in a place where other viewer change buttons are present. 

As the Command bar does not currently exist in any room layout adding it there does not help. In addition to that, those buttons don't work when on a Command bar without some code changes (which I have in another PR)

Also in my other PR, I added these as right click menu option but will likely remove them if this goes in, but I still think it is better to have it visually  and easily available.

gab3d:
...and what about placing them to the left of the Safe Area button? It'd be a much more less used area, so it'd contemplate Konero's objections too.

manongjohn:
I put them next to Freeze button more as a logical grouping of buttons that affect what is being viewed.

If it is more acceptible to the left of the Safe Area, despite the perceived added clutter, then I can easily move it.

I'll give time for others to weigh in.

konero:
@RodneyBaker I strongly dislike the location unfortunately, I hope I don't come across too negative though, I'd feel the same if Toon Boom did it, those viewer buttons are very high traffic, you're hitting them all the time, this is like adding the print commands amongst them, they're very rarely used. 🤷‍♀️ 

tb1

Toonz is a software that needs to be highly customizable in its interface like Harmony or Clip Studio, different departments require different workspaces, in-fact if you ever purchased Harlequin, you'd have to choose from a myriad of different installation setups depending on the department. If I set up some workstations for compositors, these commands are completely useless for example. But in the Command Bar, it would allow us to pick and choose what's required.

@manongjohn If anything, this PR highlights the inadequacy of Toonz Quick Actions abilities if you say some commands don't work in the Command Bar, I had no idea.

Then it should be re-invented. Do it how Clip Studio does it, or Corel Painter. A single window that holds a drop box list of custom panels.

csp_aa

Right-clicking a shortcut allows the user to set their own icon or BG color. If there's no icon available, then just display a default icon (like CSP), also allow the user to pick between "icon only", "text only" and "icon and text" layouts like in TVPaint and to display them as "1 row", "2 rows", etc.

csp_aa2

handy-panel

I still think this issue is solved by adding the commands into the View menu where they should be in the first place. If a user can't find it there then that's their fault.

artisteacher:
I think it’s a great idea to make these functions more visible, but like @konero I’m not crazy about the location. The command bar would be a logical choice, but several view options don’t seem to work from there at the moment. If the view commands would work & the preference changes being discussed in #2523 were implemented, then the command bar would be fine. 

Another option could be the menu at the bottom of the viewer. Function-wise, that might make the most sense.  In practice, it might not work since that menu is pretty over populated already.

manongjohn:
In #2501, I fixed all the Viewer related actions to work from the Command Bar. 

Right now, the issue I have with the Command Bar is that not every action has an icon, so the Command Bar fills up fast, especially on smaller screens.  But that is another issue. :)

So putting the icons left of the Safe Area button is still not acceptable so people won't have to adjust their muscle memory for the existing buttons?  This seems to be more of the issue, in my opinion, than anything else because if these buttons were there when OT was first developed or made public, people would have just acknowledged it was part of the interface and gotten accustomed to it being there.

artisteacher:
There's a start to the missing commands icons in #2111...

konero:
A quick concept, I wanted to see what it might look like on the left as it was a divided opinion in the office. Seems OK with smaller icons, tighter spacing. 🤔

v_r_1

Zoom text field | Zoom in | Zoom out | Flip H | Flip V | Reset View | And some spacers.

Click the zoom number to bring it into a text field (right-aligned) with all text preselected etc.

v_r_2

manongjohn:
If it makes more sense to put these on the viewer control panel, I'm ok with that as well.

Just as long as it's visable and easily accessible somewhere that makes sense.

Zoom text field seems a bit much. I'd be fine with just the Zoom In/Out buttons for quick zooming operations. With the zoom controls more accessible like this, it would probably make the Zoom tool somewhat obsolete.

konero:
Fair enough, here's the icons if you want to use them, they're all 20x20dp, with 2x2dp padding leaving 18x18dp for live area. On the viewer, each button is spaced 1px apart, each group is spaced 10px apart to give enough dead zone between types.

pane_preview11

**Download: [NaviPaneIcons.zip](https://github.com/opentoonz/opentoonz/files/2927137/NaviPaneIcons.zip)**

gab3d:
@manongjohn 
Zoom tool still lets the user zoom to the current mouse position, this buttons won't.

bobbybutterfly:
would be nice to have the horizontal flip button in the scale menu of the animate tool too! Turn your character to the left or right with one toggle button. 
Would be easier then setting a "-" in the H field

artisteacher:
If toggle buttons were added to the animate tool, they really should be to the select tool as well for consistency. 

Personally, I feel like the animate tool doesn't have the space for more options in the tool option bar - that would be at least 2 more buttons (horizontal & vertical). Try looking at the tool options when the Animate tool is in All mode! The Select tool isn't any better on a small screen since it can't scroll. 

I think adding a - doesn't take significantly more time/effort than a toggle button would. There is a downside though - new users usually don't realize that negative numbers can flip the direction.

konero:
Gave this is a test, the zoom ratio is off somewhat, it's not the same as the shortcut keys. You should be able to get to 50% from 100% in one or two presses, same for 200%, whereas these buttons took ten or so.

manongjohn:
@konero Thanks for testing it out. I've modified it to use the same zoom in/out logic as the shortcut.

konero:
@manongjohn Works perfect. 👏

RodneyBaker:
@konero Did all of your concerns get taken care of with this?
That looks to be the case but want to be sure.

konero:
@RodneyBaker @manongjohn Sure, this is an interface change so you might want to wait for Shun.

I thought you wasn't a programmer though so how are you reviewing these pull requests for merge?

RodneyBaker:
Based on @shun-iwasawa 's clarification of changes in #2746 that can be merged specifically:

> Altering UI which cannot be controlled user profiles (like adding buttons to the panel or adding commands to the context menu) : Basically it's fine to be merged, but please be careful if it will not disturb the conventional usage.

The question:  What if anything with this PR disturbs conventional usage?
The answer to this will let us know if we can proceed with merging and if not what to highlight for review by @shun-iwasawa.

It is my understanding that these new icons are all to the left of the original icons which still remain in their places.  As such my assessment is that this PR does not disturb conventional usage.

But... what say you all?

hayoko:
They're useful, but they should be made optional, similar to a navigation window you find in other painting software, they're usually hide-able.

Can you also fix the title text sliding under the icons? These icons are transparent so the title text fits under them making both icons and text difficult to read.

manongjohn:
To address some of the concerns and reservations over placement of the new buttons, I will move them down to bar at the bottom of viewers after `Set Key`.  It will be under button group `Viewer Controls` which you can hide it if you don't want it.

artisteacher:
I like the button placement - though I wish that fewer of those controls were enabled. If it were up to me, I'd leave only framerate, playback, set key, and viewer controls on by default.

manongjohn:
Any objections to merging this?

RodneyBaker:
It looks to me like you've addressed all concerns.

PR #2543, opened 2019-03-04
Change Preference defaults

### LINKED ISSUE #2523: Discussion: Updating default preferences for new users

manongjohn:
This PR resolves #2523 

Per the initial post and subsequent discussion, the following Preference defaults are being changed as follows:

### General
- Backup Animation Levels when Saving: **Enabled**

### Interface
- Icon Size: **80 x 45**
  - The default room layouts have been refreshed and Level Strip updated to reflect the smaller size.
  - Added logic to scale down Inbetween icon as needed + added Tool Tip "Auto Inbetween"

### Import/Export
- ffmpeg Timeout: **600**

### Drawing
- Default Level Type: **Toonz Raster Level**
- Autocreation: **Use Xsheet as Animation Sheet**
- Down Arrow at End of Level Strip Creates a New Frame: **Enabled**

### Tools
- Dropdown Shortcuts: **Cycle through the available options**

### Xsheet
- Cell-dragging Behaviour: **Cells and Column Data**
- Use Arrow Key to Shift Cell Selection: **Enabled**
- Show Toolbar in the XSheet: **Enabled**
- Sync Level Strip Drawing Number Changes with the Xsheet: **Enabled**

### Animation
- Default Interpolation: **Linear**

### Onion Skin
- Use Onion Skin Colors for Reference Drawings of Shift and Trace: **Enabled**

Upgrade notice: Existing users that have changed an impacted preference at least 1x will not be affected by that default change because their choice would have already been written to the preference.ini file which takes precedence.  A change in the default will only apply to an existing user if the user never changed the preference in any way.

For Testing purposes, here is the updated C:\OpenToonz stuff\profiles\layouts\rooms\Default directory: [Default.zip](https://github.com/opentoonz/opentoonz/files/3588331/Default.zip)


gab3d:
Nice!

my only doubt is about Auto-stretch Frame being Disabled...

manongjohn:
I did that to simulate the current behavior of `Enabled` setting but without creating new levels every time they skip a few frames.

artisteacher:
I really prefer having auto-stretch frame disabled. It's more flexible with frame by frame techniques. It's also faster using the "fill in empty cells" command than it is selecting & deleting unwanted cells.

gab3d:
ah... ok. I understand your logic of changing the least possible and that's fine...

Though, now that I'm testing the behaviour, I see two problems: one with the **Use Xsheet as Animation Sheet** mode in general, and another one specifically when **Auto-stretch frame** is Disabled.

Let's consider we have a level's drawing **1** exposed in frames 1 to 10 of the Xsheet, and the user goes to frame 4 and draws something new in that same column:

**Use Xsheet as Animation Sheet** mode problem:
If a level drawing with the same frame number (4 in this example) already exists, OT creates a new level drawing numbered **4a**, **4b**, **4c**, etc. 
This is not a bug, per se... but it's something I predict would confuse users if we don't give them an initial warning about what's about to happen. (It's usual that some programs warn users the first time they take a potentially confusing/dangerous action in a session; then they can use the function normally for the rest of the session without any warnings, also on most programs this warning can be toggled on/off by advanced users in the Preferences).

**Auto-stretch frame** Disabled state problem:
If a level drawing with the same frame number (4 in this example) didn't exist previously, drawing **4** of the level is created and exposed at frame 4 of the Xsheet, but the frames following that one will still have drawing **1** of the level exposed (as previously) until frame 10. 
This is strange... at least it's surely not what I would expect, and is something that doesn't happen when the **Auto-stretch frame** option is Enabled.
Again, i do not consider this a bug, but an unpredicted behaviour.

Perhaps we could make OT to change the Xsheet cells following a new created drawing to expose that new drawing, also when **Auto-stretch Frame** is Disabled.
That way we would still get a different behaviour between the Enabled or Disabled states, regarding what happens with the Xsheet cells *before* the newly drawn frames...


It'd be great if others also jump in and speak their minds on this one, perhaps there's some important scenarios I'm not considering here?

artisteacher:
> Use Xsheet as Animation Sheet mode problem

Instead of a warning, perhaps a dialog like the following:
Frame 4 already exists - do you want to overwrite the existing frame or create frame 4a? 

This might be best if it were similar to the load/import option dialog/preference. It would need a checkbox to always do the selected option - which could be changed in preferences.

> Perhaps we could make OT to change the Xsheet cells following a new created drawing to expose that new drawing, also when Auto-stretch Frame is Disabled.

As long as cells with different frame numbers wouldn't be changed too, it wouldn't impact my workflow.  It would definitely be important to hear from more people using these settings though. I could see it being annoying both ways, depending on what you are used to. As a workaround, shortcuts W & Q work well to change the exposure of individual or multiple cells.

@manongjohn - I'm really excited for these changes! I'll have a lot less setup to do with my students when we start the course or update OT mid-way through.

manongjohn:
@gab3d These are known behaviors/issue of this particular Autocreation mode which should probably be dealt with separately.

For the purposes of this PR, the auto-stretch settings seems to have people who like it on all the time and those who don't.  I guess the question is more of a new user's expectation, especially if they come from other animation software.  What's the default behavior for new users in other animation software? Is the expectation that when a new frame is dropped/replaced it autostretches?  If so, then we should keep it enabled by default to be consistant with others.

gab3d:
@manongjohn 
I agree this fixes should be dealt with in another PR.

One thing I like about the Auto-stretch being Enabled is that you can play the WIP animation and always see the frames standing until the next one shows. When Disabled,  you can only see drawings where you've actually drew them, and the rest is blank. So you get sort of a flashing effect.

But I agree another important thing is to see what the most used alternatives do, and try to fit into that too as much as possible.

manongjohn:
It appears Krita autoholds between their key frames. If anyone has any experience with other animation software and how it works by default, please chime in.

Personally, I would prefer autohold/stretch turned on as a new user.  For evaluating and learning OT initially, I think it makes it quicker to start drawing, animating and using onion skin.

artisteacher:
In Adobe Animate, it also auto holds between keyframes. If you add a new keyframe in the middle of another keyframe, the frames afterwards match the new keyframe instead of the first.

gab3d:
yep, to me it seems the most intuitive thing to expect, if I were doing my first steps with the program.

artisteacher:
Have you compensated for the smaller icon size with the Inbetween buttons? If I manually set the preference for the icon using the dimensions listed and try to inbetween, I get:
Screen Shot 2019-03-10 at 1 52 25 PM

turtletooth:
I'd like to request that rewind after playback also be changed.  Most folks would probably want to continue where they left off instead of going back to the beginning.

manongjohn:
I fixed the Inbetween image by scaling as needed. Also added a tool tip for it.

Still a bit divide about Rewind after Playback.  What do other animation software do after playback? In Krita, it resets back to whatever frame you started playback on, which I like but woulds handle separately as a feature request.

turtletooth:
Rewind after playback may be more useful for Stop Motion like I've been doing.  So we can leave it as is.

manongjohn:
As most of this is just changing defaults, this should be fairly easy to test and then merge.

To test on Windows:
1) Download the AppVeyor Release Artifact and open it somewhere.
2) Backup your stuff folder!!!!
   - If you don't want to use your existing stuff folder, copy the existing stuff folder to the same directory as the AppVeyor OpenToonz.exe and rename the stuff folder 'portablestuff'.
3) Remove/Rename the following folder: ...stuff\profiles\layouts\rooms\Default
4) Replace the Default folder with the contents of the Default.zip from the PR description
5) Remove/Rename the following file: ...stuff\profiles\layouts\settings._userid_\preferences.ini
6) Remove/Rename the following folder: ...stuff\profiles\layouts\personal\Default._userid_
7) Start up the AppVeyor OpenToonz version.
8) Restore your stuff folder from the backup, if you used your main stuff folder.

RodneyBaker:
I want to add something here regarding Preference defaults that likely shouldn't be changed but comes up often enough that it should be good to add it for informational purpose here.

When moving from the Cam Stand view to Level Strip new users often wonder where their canvas disappeared to because of the change from white to dark grey.  In testing I was trying to find an off white color that was different enough from the canvas to stand out and yet didn't disorient the user as much as the larger shift from white to grey.   Setting the RGB colors of the Viewer's background to something like 225, 225, 225 works although will be too bright for most who are staring at the screen all day (perhaps headache inducing even!).  As such I'm not recommending a change to the default here so much as adding this note as a reminder that users do have an option to adjust the viewer's background so the transition from Level Strip to Cam Stand view works for them better.

That change can be made via Preferences > Colors > Viewer BG Color
This is the first color setting at the top of the page.

ViewerBGcolor

RodneyBaker:
Testing this now and so far so good.

- [x] General > Backup Animation Levels when Saving: Enabled
- [x] Interface > Icon Size: 80 x 45
Note:  There appears to be excess space to the right of the icons in the Level Strip.  The size of the Level Strip window has to stay the same width as the Style Editor below and so it expands the right side, leaving the empty space.  It works as is but I will likely change it to better fit my laptop.

- [x] The default room layouts have been refreshed and Level Strip updated to reflect the smaller size.
- [x] Added logic to scale down Inbetween icon as needed + added Tool Tip "Auto Inbetween"
- {x} Import/Export > ffmpeg Timeout: 600
600 has been my personal default of late so I'll have to double check to see that this was updated from the replaced files and not held over from previous settings.  Looks good though.
Aside:  I know this would be a different PR but looking at that empty FFMPEG address space above reminds me that I'd dearly love to see a default value for an address placed there.  Either a ghosted address or some recommended defaults such as C:\ffmpeg\bin (for windows) would be optimal.  @jpturcotte was working on a universal address for ffmpeg that would generally work for PC, Mac and Linux but it didn't get all the way to the finish line.

- [x] Drawing > Default Level Type: Toonz Raster Level
- [x] Drawing > Autocreation: Use Xsheet as Animation Sheet
- [x] Drawing > Down Arrow at End of Level Strip Creates a New Frame: Enabled
- [x] Tools > Dropdown Shortcuts: Cycle through the available options
I usually don't have this on so am looking forward to working with it.

- [x] Xsheet > Cell-dragging Behaviour: Cells and Column Data
- [x] Use Arrow Key to Shift Cell Selection: Enabled
- [x] Show Toolbar in the XSheet: Enabled
I usually have this off because I don't take the time to turn it on.  Thanks for defaulting it on.

- [x] Xsheet > Sync Level Strip Drawing Number Changes with the Xsheet: Enabled
- [x] Animation > Default Interpolation: Linear
- [x] Onion Skin > Use Onion Skin Colors for Reference Drawings of Shift and Trace: Enabled

LGTM
If you say this is read to be merged I'll be happy to do it.

gab3d:
what do you guys think of moving the backup options to the Saving category?
it'd mean less clutter in the General one:

prefs-general
prefs-saving

manongjohn:
@gab3d It makes sense to me to move the backup settings to the `Saving` tab.  I would probably look there 1st...I actually did that the other day and was confused why it wasn't there.

I won't make that change in this PR but will do that in another PR at some point.  Maybe I should start a "Preferences Reorganization" discussion :)

As far as this PR is concerned, most of the changes are value changes (actual preferences) or pane adjustments (room files).  There is only a tiny bit of code to deal with the Inbetween icon scaling and adding a Tool tip over it.  I'm good with this merging as is unless there is an objection to it.

turtletooth:
LGTM.  Thanks @manongjohn

PR #2544, opened 2019-03-04
Fix Transparent MOV Output

### LINKED ISSUE #2530: Can't render MOV with transparent background
This problem was reported to my e-mail from one of the community members of the forum.

> I am having problem with exporting in mov format- I cant export my videos with a transparent background anymore. It used to work before 1.3. I have attached some images of my settings for more info.

  

It is probably be related to the changes done for adding a "matte color" preference option in #2418 .

### System Information
* **OpenToonz Version**: V1.3

shun-iwasawa:
This will fix #2530 

MOV may have alpha channel under some settings (Millions of Colors+ color depth). 
Currently a solid background is always inserted, which makes impossible to render transparent MOV.

I tried to make OT to detect the MOV settings and adaptively switch the behavior of the background insertion, but ended in vain :-( ... So I just omitted every MOV from applying solid background as a quick fix.

Please note that after merging this PR you will need to set the alpha channel to 255 in `Scene Settings > Camera BG Color` in order to make the background color visible, or the rendered MOV will have black background at the transparent region if it does not support alpha channel.

### DIFF
--- a/toonz/sources/toonz/rendercommand.cpp
+++ b/toonz/sources/toonz/rendercommand.cpp
@@ -515,7 +515,11 @@ void RenderCommand::rasterRender(bool isPreview) {
   // fixes background colors for non alpha-enabled movie types (eventually
   // transparent gif would be good)
   currBgColor.m = 255;
-  if (isMovieType(ext)) {
+  // Mov may have alpha channel under some settings (Millions of Colors+ color
+  // depth). I tried to make OT to detect the mov settings and adaptively switch
+  // the behavior, but ended in vain :-(
+  // So I just omitted every mov from applying solid background as a quick fix.
+  if (isMovieType(ext) && ext != "mov") {
     scene->getProperties()->setBgColor(currBgColor);
   }
   // for non alpha-enabled images (like jpg), background color will be inserted


shun-iwasawa:
Jenkins

artisteacher:
I'd give this a try but we can't render or import .mov files on macOS - which is weird since .mov is a mac format.

manongjohn:
Even though you are on a Mac, don't you still need to install Quicktime for .mov and some other mac formats? I think you can still download and install it, though Mojave will probably be the last version it will be available for I believe.

You don't need the Quicktime app, but probably need the Quicktime Essentials from that installer.

artisteacher:
Quicktime is one of the default apps that comes on a Mac, so I've never had to install it.

shun-iwasawa:
Unfortunately, [QuickTime Mov output is currently unavailable in macOS version](https://github.com/opentoonz/opentoonz/blob/25d3db05fd57aed51d666455ac8dbc44067a4cea/toonz/sources/image/mov/tiio_mov_proxy.cpp#L54) of OT.
To enable this, it seems that some OT modules just for outputting Mov files are needed to be built in 32bit. ( `opentoonz/toonz/sources/image/mov/tiio_movM.cpp` seems to be for Mov output and [it will be compiled under 32bit build only](https://github.com/opentoonz/opentoonz/blob/25d3db05fd57aed51d666455ac8dbc44067a4cea/toonz/sources/image/mov/tiio_movM.cpp#L3). ) Or you may just be able to build in 64bit using 64bit QTKit, but I haven't tried either of them.

masafumi-inoue:
LGTM

Issue #2028, opened 2018-06-01
Link to Online Documentation in Help Menu?


gab3d:
@manongjohn 
...now that you're into adding menu options. ;) ;)
How about adding a link to the OT's online documentation to the Help Menu?
It seems easy to do and it'd certainly increase the exposure of the online documentation to users, especially new ones.

DarrenTAnims:
While you're at it, it could be worth adding a link to this site for reporting bugs and the other site for suggesting feature requests. That'd certainly get more coming through.
(a little ironic as this site is for bugs and this is a feature request, but thought I'd tack it on as I've just noticed this one)

artisteacher:
I'm not sure I would specifically add a link for feature requests - that might create the expectation that they will be fulfilled. However a link to the forum for help (in addition to the online documentation) could be useful.

This is a little off topic, but in addition to these links, I think the menus in general could use some reorganization. The File menu in particular is excessively long. It would be great to have some submenus to make it easier to find certain commands.

DarrenTAnims:
I agree artisteacher. And I super-strongly agree about re-organising the menus. I struggle to find items, especially in the File menu and Windows menu. They all need grouping better.

RodneyBaker:
Closing as a feature request slipping to 2019 development cycle.
Ideal online locations might be the primary opentoonz site and/or readthedocs.

Note that several Digital Video products have links to local and online html and pdf resources.
One thing to consider is how to best account for links that might break should that resource be removed.. as seen in one of those products.

Recommend testing to see if linking to a local html file which in turn links to the online locations might be ideal as then the user could update and customize the links as desired.

gab3d:
@shun-iwasawa 
could we please add this in the next version of OT?
I think it's really needed...

this is just an example that happened today in Facebook:
![Sin título](https://user-images.githubusercontent.com/18104091/63614553-bc487e80-c5b9-11e9-88ff-5f9eade4a6dd.png)

gab3d:
@RodneyBaker 
could you please reopen it?

manongjohn:
Adding link to on-line help is actually part of PR #2545.  It needs to be reviewed and merged.

gab3d:
great!

RodneyBaker:
Agree that documentation link from the Help menu is important.
If @shun-iwasawa isn't opposed to me merging the other PR i'll be glad to do it.
The issue may be that too many changes are included in #2545.
[...]

PR #2545, opened 2019-03-04
Reorganize Default menubar

### LINKED ISSUE #2524: Discussion: Cleaning up menus
There has been some discussion on Discord about how the current state of our menu's organization can be improved upon.

I'm opening this issue to discuss how best to reorganize the main menus so they make sense to new users without being overwhelming. Eventually I'll turn it into a PR.

manongjohn:
This PR resolves #2524 

I've implemented the menubar layout @konero has provided that includes feedback from others. I feel it works and flows pretty well without making it so different that existing users will have a hard time finding things.

newmenubar

I've made additional changes not shown in his gif:
- Moved `New Note Level` from the `Xsheet` menu to the `Level` -> `New` submenu with the other levels
- Added `Toggle Edit In Place` to `Xsheet` menu
- Added `Apply Lip Sync Data to Column` to `Xsheet` menu
- Moved `Startup popup...` to `Windows` menu
- Moved `Touch Gesture Control` to `Touch/Tablet Settings`  tab in Preferences
  - Existing toggle button still exists and can be added to any bar as desired.
- Added `Online Manual...` to `Help` Menu
  - This opens to http://opentoonz.readthedocs.io in the user's default browser window.
  - Assigned shortcut key: F1

Upgrade notice:  For existing Windows users, this change will only be effective when they choose to `Reset to Default Rooms`.  This is because the  menubar layout is stored in configuration files after opening OT for the 1st time which takes precedence over what is coded.  For all macOS and Linux users, this change will take immediate effect.


artisteacher:
This will be really helpful!

Is there any way to get a link to the manual in the Help menu?

gab3d:
> Is there any way to get a link to the manual in the Help menu?

yes, please.

gab3d:
I have a some constructive comments about this new menu layout, please take them as personal preferences/points of view. I'm perfectly well if nobody agrees on any of them: 😄 

- File menu: I see the Import/Export sub-menus as unnecesary with so few options there, because:
  - **Import:**
    - It has only one option into it: **Import Magpie File...**, which could be moved next to where **Apply Lip Sync Data to Columns** is, to group similar functions together, which is a good thing. It can be renamed to **Apply Magpie Lip Sync File...** because it essentially does the same as the other lipsync option (the only big difference is that it asks for the file in first place, and then opens the options window)
  - **Export:**
    - I think the **Fast Render to MP4** option fits much better in the Render menu. It's basically a render function...
    - **Export Soundtrack**, well... I know it has the word "export" in its name, but is also basically rendering audio, so it could fit in the Render menu as well. Perhaps **Render Soundtrack** could also be easily understandable and could make it feel better in the Render menu?

- Level menu:
    - I don't see **Antialiasing** to be any kind of "Tonal Correction" function, as to be inside that sub-menu.

- Xsheet menu:
  - Perhaps there could be a **New** submenu at the top there? It'd maintain a certain consistency between the Level menu structure and this one. And there are several options that could easily fit there: **New Note Level**, **New FX**, **New Output**
  - **New Note Level**, could be in the Level > New sub-menu instead of here. I guess it was placed here because it is not a standard independent level (in the sense that it's not accessible via the Scene Cast to be used in other Xsheets), but... it is still a level, isn't it?

- Play menu:
    - Honestly, do we need it? I mean, if it is to keep a visual reference of what shortcuts are assigned to these commands, well... I think those could be included in the viewer play buttons tooltips, as it currently happens with the tools in the toolbar. Following the same logic we would need a Tools menu, where all tools are listed too. What'd be the sense in having that? I would better make this Play menu a sub-menu of the Xsheet one, so it is not that present in the UI.

- Render menu:
  - Why do **Link Flipbooks** option is present here and not in the View menu, where all the rest of the different viewer modes and options are in?

- Windows menu:
  - I really like the Workspaces submenu. Should we rename them from "Rooms" to "Workspaces"? 

- Help menu:
  - A link to the OpenToonz Manual is really needed here. (I know it has already been said, but just to put all together here).

manongjohn:
@gab3d 

> New Note Level, could be in the Level > New sub-menu instead of here. I guess it was placed here because it is not a standard independent level (in the sense that it's not accessible via the Scene Cast to be used in other Xsheets), but... it is still a level, isn't it?

I already moved it there.  If you used the animated gif as reference, it was the original one @konero provided.  I didn't create a new one but noted 3 changes below it, the 1st one being the New Note Level.

What's the link to the OpenToonz Manual? (You'd think I should know this).  I'll look at adding a button called "Online Manual" to open a webpage directly to it.

manongjohn:
Before I push the commit, is this placement ok or should it be above the About Opentoonz?

image

Actually, I think About OpenToonz should really be at the bottom of the menu, not the top.  The Startup Popup and the Touch Gesture Controls still seems odd to have under the Help menu.

konero:
Oops, About usually goes at the bottom, my bad. Manual could go at the top, in other software you'll usually find something like "Help Topics" at the top.

artisteacher:
I agree that it’s not really an expected place to have the touch gestures nor startup popup. I’m not sure if the best locations though. 

Someone proposed having the startup pop-up under Windows, which isn’t a bad spot for it. The File menu is probably the other logical place.

konero:
> I agree that it’s not really an expected place to have the touch gestures nor startup popup. I’m not sure if the best locations though.

Personally I'd remove touch gestures completely, put it into Preferences under the tablet tab.

> Someone proposed having the startup pop-up under Windows, which isn’t a bad spot for it. The File menu is probably the other logical place.

I just copied where Harmony had theirs, my guess is that it's a window not everybody wants to use, which is why it's optional and put out of the way.

manongjohn:
The gestures control is handled as an environmental preference which is handled differently than what's in the Preferences dialog. But I will see about placing it in the `Interface`  or `Tablet Settings` tab in Preferences, whichever makes more sense.  Currently the `Tablet Settings` tab is only visible when it detects a system that has Windows Ink on it so it's not there for macOS/Linux users, but i can fix that.  

Maybe we can put the `Startup Popup` after `Record Audio` in the Windows menu to keep it low on the list since it isn't used much?

@gab3d `Online Manual...` sounds good.  Assigning F1 to it is a good idea.

gab3d:
I've got no problem with **Startup Popup...** in the Windows menu, sounds logical to me.

konero:
> File menu: I see the Import/Export sub-menus as unnecesary with so few options there...

It'll be good to have an established directory for similar commands in the future.

> I don't see Antialiasing to be any kind of "Tonal Correction" function, as to be inside that sub-menu.

I agree, I guess it could go between "Level Settings..." and "Info..."?

> I think the Fast Render to MP4 option fits much better in the Render menu. It's basically a render function...

My colleague hated that function being so close to the Render command. He's got more experience than me at rendering so I just went with the flow here, putting it out of the way.

> New Note Level, could be in the Level > New sub-menu instead of [Xsheet]. **I guess it was placed here because it is not a [disk action]**... but... it is still a level, isn't it?

It's as you guessed, it doesn't bother me where it's placed but I wouldn't combine them if it was commercial software.

> Play menu: Honestly, do we need it? I would better make this Play menu a sub-menu of the Xsheet one, so it is not that present in the UI.

Can do, I got no problem with that suggestion. I put it there because some InknPaint users hide the console bar in their room (small Cintiq displays) and just needed easier access.

> Render menu: Why do Link Flipbooks option is present here and not in the View menu, where all the rest of the different viewer modes and options are in?

I just did what my colleague suggested; that it was nice having the playback/render stuff in their own menu. It's how the room is laid out in the Ghibli configuration too, (habit)?

manongjohn:
Can "Tonal Correction" label be changed to something else to keep Antialias and possible other future level adjustment tools in that sub-menu?  I can move Antialias if a suitable label can't be found.

The placement of the Fast Render MP4 does seem rather odd to me as well.  I also would prefer to see that in the Render menu.  I'll leave it where it is for now, I guess.  I don't hate it there nearly as much as your colleague hates it being in the Render menu apparently. :)

Since Note levels are now more useful with the `Text Iwa` fx, it think it deserves to be with the other levels. I know some have requested exporting Note levels to text file. Perhaps in the future, Note levels can be directly exported to or imported from a file.

I can see the Play menu being useful for those who may not use keyboard shortcuts and may want to hide the viewer console to gain more screen real estate.  Moving it to a sub-menu in Xsheet would probably make me want to keep the viewer console though. 1 click to play/pause is ideal. 2 clicks, doable. 3 would be more trouble than it's worth. If it's going to be a sub-menu, I'd rather not have a Play menu.

The Link Flipbooks makes more sense to me to be under the Play menu as it's a setting for playback in a way.  If you rendered something and loaded into a flipbook, then made an adjustment and rendered again into another flipbook, you could link the flipbooks playback to compare to see which you like better.  I guess that may be why it's under Render?  At any rate, I'm fine with it being under Render, Play or View.

artisteacher:
What about “Adjustments” instead of tonal correction?

RodneyBaker:
Just a note concerning the 'Fast Render to MP4' feature...

Jeremy Bullock (@turtletooth) added that feature specifically to the File menu for new users/students so they could easily find it.  

Jeremy isn't very active here in Opentoonz these days but it would be good to get his feedback before moving the location.

konero:
> What about “Adjustments” instead of tonal correction?

That can work, or "Adjust". 🤔 

> The placement of the Fast Render MP4 does seem rather odd to me as well. I also would prefer to see that in the Render menu..

I couldn't figure where to put it, in the Preferences window the settings for this feature are listed under "Import/Export" so I went with that. If it gets moved maybe add a separator if it goes next to "Render" again, that could give enough dead space to stop miss-clicks. Since the file gets overwritten with no confirmation I thought it was a good candidate for a sub-menu though.

> I can see the Play menu being useful for those who may not use keyboard shortcuts and may want to hide the viewer console to gain more screen real estate. **Moving it to a sub-menu in Xsheet would probably make me want to keep the viewer console though**.

Agreed. Toon Boom Harmony has a "Play" menu at the root of the menu bar and Clip Studio Paint has theirs hidden in the Animation menu (Animation > Move frame).

> Since Note levels are now more useful with the Text Iwa fx, it think it deserves to be with the other levels.

Sounds good.

> **The Link Flipbooks makes more sense to me to be under the Play menu** as it's a setting for playback in a way. If you rendered something and loaded into a flipbook, then made an adjustment and rendered again into another flipbook, you could link the flipbooks playback to compare to see which you like better.

That actually sounds better. 👍

artisteacher:
On related note, this has been discussed at some point previously, but "Send Back" should really be renamed to "Send to Back" to be more consistent with "Send to Front". It would also be more clear what the command does.

manongjohn:
@artisteacher Sounds good. Can you add that to your label correction PR?

Savraska2:
It would be great if the string with the link to the manual would be available for translation, so that clicking on the menu would open the manual in the appropriate language.

manongjohn:
Done.

konero:
Just noticed 'Reframe With Empty Inbetweens...' should be under the 'Reframe' group (under Cells) which is where it shows up in the rest of the UI.

Sorry. 🤦‍♂️

turtletooth:
I was looking at this tonight.  All in all it seems like a pretty solid reorg.

I think that the import and export menus under the file menu shouldn't be submenus since they only have one item in them.  

I also think the touch control button should be brought back.  I think it could go at the bottom of the edit menu.  Having to go all the way into preferences is pretty cumbersome and makes finding that ability less likely.

Thanks for your work on this @manongjohn

artisteacher:
Or the touch control button could be added to the command bar. As long as it's enabled by default, that will allow easy access.

After spending some time with the new menu, overall I really like the reorganization. Having the render options in their own section takes some getting used to - but that's partially from still using the old menu at school.

I would argue that both "Toggle Maximize Panel" and "Toggle Main Window's Full Screen Mode" might fit better under the "View" menu. They do both refer to windows, but to me at least, they seem more as view options. 

Lastly, this isn't really a preference, but I wish "Lock Room Panes" was enabled by default. New users are constantly pulling out windows and messing up their rooms by mistake. Really, this should probably be the default either way, but with the lock option slightly more hidden, I think it will be even more important.

shun-iwasawa:
Jenkins

shun-iwasawa:
The new menubar layout looks tidy and accessible. Thank you very much @manongjohn for the improvement! Also thank you @konero and other members for proposals and discussions. LGTM

gab3d:
@manongjohn  @shun-iwasawa 
Are the Help > "Online Manual..." option and it's related *target link* translatable?
I can't find those strings in the .ts files...

manongjohn:
I didn't run the update.sh scripts for any languages to capture new strings that need to be translated. Figured that will happen closer to next release.

gab3d:
ah... ok, my mistake. 
(As I've found a lot of new strings untranslated I assumed the new menus were also already included in those).
thanks.

RodneyBaker:
Also resolves #2028

PR #2546, opened 2019-03-04
Fix Update of Dedicated CPUs Field in Output Settings

shun-iwasawa:

When loading a scene, `Dedicated CPUs` field in the Output Settings does not update properly.


shun-iwasawa:
Jenkins

masafumi-inoue:
LGTM

PR #2554, opened 2019-03-10
Fix bad Fx connection crash

### LINKED ISSUE #2552: Crash when dragging to switch ports on Particles node
Found by Gab3d
https://github.com/opentoonz/opentoonz/issues/2550#issuecomment-471331016

Repro Steps:
1. Create a Particles node
2. Create a node of any type and connect it to the Particle node's Texture input
3. Click and drag the Texture port to the Source port - this is supposed to switch the inputs

The connection line will be frozen in place and shortly after, OT will crash

Visual example:
Swizzle Crash

manongjohn:
This PR fixes #2552 

This crash is due to accessing elements on null pointers which are somehow generated as a result of an issue identified with port swapping.  Added checks to deal with the null pointers that it was not expecting to see.

The underlying issue caused by the port swapping will still need to be looked at and fixed. These checks should be there anyway in case there are other scenarios that cause the null pointers.

### DIFF
--- a/toonz/sources/toonzqt/fxselection.cpp
+++ b/toonz/sources/toonzqt/fxselection.cpp
@@ -417,6 +417,7 @@ Link FxSelection::getBoundingFxs(SchematicLink *link) {
 Link FxSelection::getBoundingFxs(SchematicPort *inputPort,
                                  SchematicPort *outputPort) {
   Link boundingFxs;
+  if (!inputPort || !outputPort) return boundingFxs;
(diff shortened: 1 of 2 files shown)


artisteacher:
I think I've been running into this a lot recently! Thanks for the fix.

shun-iwasawa:
Jenkins

shun-iwasawa:
Confirmed that this PR prevents crash. 
Thank you @manongjohn for adding the fail-safe! LGTM

PR #2571, opened 2019-03-21
Fix timeline key-scrolling

manongjohn:
This PR fixes the following scrolling issues when using keys to navigate the timeline.

1) When you use Down-arrow to reach the bottom of the timeline, the current frame is selected but the bottom level is not scrolled into view requiring the user to use the scroll bar to see it.

Added logic to account for the height of the bottom level when scrolling to the last column.

2) Using END key to get to last frame was jumping to a frame way past the last exposed frame.

Added logic to factor in timeline zoom factor to correctly calculate the last exposed frame.


manongjohn:
Looks like QT version for OSX builds changed again. Now at 5.12.2.  I'll submit a PR to fix

turtletooth:
Restarted the build.

shun-iwasawa:
Jenkins

shun-iwasawa:
It works fine. Thank you @manongjohn for the fix! LGTM

PR #2574, opened 2019-03-22
Add Camera Column to Xsheet

manongjohn:
This PR adds, or rather re-adds, a Camera Column to the Xsheet/Timeline.

At several users requests, I have re-enabled the Camera Column that existed at one point before Toonz was open sourced. The Camera Column can still be seen in the Toonz Tab product line.

image

Camera Column notes:
- Always appears in Column 1 (left-most in xsheet, bottom in timeline).
- It cannot be moved nor anything moved into or to the left of it.
- It is shown by default
- The Camera Column is hidden when 'Xsheet -> Show Keyframes on Cell Area` is unchecked.
- Can be hidden completely with Preference `Xsheet -> Show Camera Column` , or column header context menu option `Show/Hide Camera Column`.
- Camera column name is now synced with the Schematic node and can be renamed.
- Can be folded and locked like other columns
- Some cell/column operations will work when the Camera column is selected. 
  - Most operations (i.e cut, copy, etc) will either ignore the camera column or block the operation, especially if selected with cells of another level type.
  - Logic added where needed to prevent a crash due to selecting the camera column 
- Can change the camera displayed in xsheet by selecting the camera in the Config button dropdown
  - Switching cameras is tracked for undo/redo operations to ensure proper undo/redo operations for keyframes on that column
  - When a camera is selected in the Animation tool, the Xsheet will update to that camera automatically.
- Column header takes on the same color used for cameras in the Stage Schematic
  - Color changes based on active/inactive status of the camera selected
  - Colors controlled by stylesheet along with other colors.


Agnyy:
Very cool! But I propose to fill the column with the camera color. Otherwise, this column looks empty.
image
image

manongjohn:
@agnyy Thought crossed my mind to use the same color as the one used in the Stage Schematic.  Technically the cells are empty since there is really nothing to expose, it's mostly used for keyframing as far as i can tell.

I will play around with it to see how it would look.

blurymind:
Wish we had a better looking icon :)

manongjohn:
@blurymind I agree! 

If anyone feels the urge to give me a new icon to make it more personalized for OT, I would gladly incorporate it.  Ideally it will be 74x44px in SVG format.

artisteacher:
This will definitely be handy - now it will just be the table and style keyframes that are not visible in the Xsheet. 

If you have multiple cameras setup, will they all display in the front/bottom of the Xsheet/Timeline? 

 I wish the camera column could be narrower in the Xsheet. Since the camera column is pretty much just for keyframes, it doesn't need nearly as much space as the other columns. Another helpful addition could be the ability to set the active camera in the column header - from the gear icon perhaps.

For the icon, should it be based on the new camera icon in the Stage, or the camera view icon?

manongjohn:
Oh I should note this only shows 1 camera.  Currently it only shows the main camera whether it is active or not..

I do like the idea of selecting which camera though.  I can certainly look into that now though.

I wish I could see/manipulate Fx keyframes also.  All these other potential columns will need to be done another time though.

I'm going to leave the width as it is for now as it would require a number of changes to account for different sized columns. Perhaps if/when we add these other columns for other keyframes we can revisit the topic.

As for the icon, I don't think it needs to look like any other camera icon found in OT. It's a thumbnail so I suppose we can be a bit more freer about what it looks like.  I'll let others weigh in, but I'm open to whatever makes it looks nicer than what it is now. :)

gab3d:
I also think it'd be handy to see those other keyframes in the xsheet, including pegbars also.
And definitely, it'd be great to save screen real estate by making such columns to be as thinner as possible in the xsheet.

Perhaps there could be a toggle button in each Schematic node to control whether this particular node is visible in the xsheet or not. That way the user would set which things would like to see exposed in the xsheet and what others is better no have them "hidden".

I agree that such a change would rather be subject of another PR, though.

phillycheese007:
this is amazing....
i personally dont mind camera icon much.  Maybe a bit more simplified...
I am however drooling at this possibility        ......what about a different shaped icon?
If the trajectory of OT is adding more access in xsheet, and u talking bout one day FX getting in there..and pegs hopefully on my end etc......

how about this or this jist:
Drawing level icons are rectangles
camera level icons are triangles
Fx level icons are a star shape
etc...........

currently everything being against the same color rectangular background does not provide as much visual differentiation as could be done by incorporating different shapes

turtletooth:
@manongjohn Thanks so much for this.  Since this column only shows keyframes, do you think it would be possible for the column to be more narrow when in xsheet mode?

manongjohn:
@turtletooth I wasn't planning on addressing that with this PR as it will require a number of changes to handle it that i'm not ready to deal with. For now this is similar to Tab and can be hidden/folded to keep it small when not in use.

I will plan for it when I add the ability to show keyframe-only columns for other things like fx, pegbars, etc.

With that in mind, how wide should it be and what would the header look like?

Agnyy:
I think that it is necessary to do so that they would not get confused when there are too many layers.
Untitled-2

manongjohn:
I'm still a bit divided on whether the Camera Column cells should be colored or not.  I can see how it would help, but at the same rate, we aren't really exposing anything in those cells, so technically they are empty.

Thoughts?

Keep in mind, if/when we get keyframe-only columns for other things, they will likely be handled the same way.

I do like the darkened color version. I'd have to add more entries into the stylesheet to support it.  I tried to darken it programatically based on the original color but trying to find a fixed % to darken colors that works well across the 4 themes is problematic.  A fixed % could be problematic in the future also.

artisteacher:
I also prefer the darkened version. I get that nothing is exposed in the Xsheet - but it still feels off having no color there while having it in the column header. 

What about using rgba colors instead of hex? That way the node and column header could be full opacity (ie.  rgba(29, 101, 157, 1)) and the xsheet would have whatever alpha value worked best for the color (ie. rgba(29, 101, 157, .2)).

melieconiek:
Spotted a few bugs with the PR.
- Pasting over a keyframe on any column shifts the keyframes in the camera column forward by one frame. The normal behavior is to overwrite the keyframes on that frame.

_no-stream

- Selecting a region of keyframes by holding down shift has some strange behavior on the camera column. Sometimes it will select keyframes outside your selection--sometimes it will skip over keyframes--sometimes both--sometimes it works correctly. ¯\_(ツ)_/¯

_no-stream_m1

manongjohn:
@melieconiek 

Thanks for testing.  It appears that 1st issue is actually an existing issue in the current version, but it is not obvious because the Camera keys aren't visible on the xsheet/timeline.  You can see it happening if you watch it in the Function Editor.

I will fix it with this PR.

ave4:
Do not judge strictly. My first experience drawing svg graphics.
[camera2.svg.zip](https://github.com/opentoonz/opentoonz/files/3319597/camera2.svg.zip)

RodneyBaker:
Two issues that seem to still be open related to this enhancement are:
1.  Refining the camera icon
2.  Limiting the size of the Camera column

In the case of the first, it seems to me that scale is the primary concern as a smaller icon would be preferred.  With the large column we can see why it was removed in Studio Ghibli version.

In the case of the second, perhaps simply folding the camera column by default might be an option?  Then the user could simply click to expand.  The downside would be that this partially defeats the purpose of presenting the camera in the xsheet/timeline to begin with.  As discussed having the column just wide enough to show keyframes might be ideal.

manongjohn:
> Two issues that seem to still be open related to this enhancement are:
> 
> 1. Refining the camera icon
> 2. Limiting the size of the Camera column

Neither of those issues should hold up the review/merge of this PR.

The icon can be changed at any time.  For now we'll use what was there originally unless someone gives me a different one to use.

I will not be handling the size of the column with this PR.  Folding by default is problematic because, per my note in the description, if you fold a column, you can't open the scene in older versions of OT or OT variants without this PR in it.  I would recommend users Hide Camera Column when they don't need it, but it shouldn't be hidden by default on initial installation.

RodneyBaker:
Tagged as Pending Review.
As a compromise I'll suggest the feature be turned off by default pending the addition of the other PRs that refine the UI.  Rationale is that the column does interfere especially in rooms/UI locations where space is limited.  At the time when those PRs are added then the default can be changed to have it on.

gab3d:
@manongjohn
I'm in favor of having the Camera column shown by default, but at the same time I guess it'd be a good idea to give users an obvious way to get rid of it whenever they need. Because, let's face it, it takes space when you dont need it.

So, couldn't we have a "Show Camera Column" button, close to the "New Note" button in the Xsheet, so that with just a click users could show/hide this column? (and in the future perhaps other similar ones, as pegbars or fx columns)

just an idea...

manongjohn:
I added a `Show/Hide Xsheet Camera Column` option to the Column Header context menus which basically updates the preference setting.  So once it's hidden it stays hidden for all sessions until it's toggled back on.  Users also have the option to add it to the Xsheet/Command bars.

RodneyBaker:
Outstanding.  Works as advertised.
Thanks!

gab3d:
@manongjohn 
> I added a `Show/Hide Xsheet Camera Column` option to the Column Header context menus which basically updates the preference setting. So once it's hidden it stays hidden for all sessions until it's toggled back on. Users also have the option to add it to the Xsheet/Command bars.

yeah, I think the function is useful.
it's just that will not be immediately obvious to any user that the functionality of hiding that column exists, because it's basically hidden in a right click menu in a specific screen area. 
I was looking for something more prominent, for any user to say "hmm... what this button does?", then click it, and see the camera column disappearing/reappearing. 😉 
I've suggested putting a button in close to the New Note button, but could also be in the Xsheet Toolbar I think. (Anything that's visible at first sight and improves awareness of the feature, basically).

ave4:
I drew svg icon:
[camera2.svg.zip](https://github.com/opentoonz/opentoonz/files/3435563/camera2.svg.zip)

manongjohn:
@ave4 Thanks.  I just realized you posted this a while ago but somehow I missed it.  Sorry about that.

Here is what it would look like once it is implemented:
image

It's a bit prominent. Maybe shrink the camera icon itself down a bit?  Keep the 74x44px dimensions though.

turtletooth:
I think it would be better to keep the style of the other icons in OpenToonz.  We already have a camera icon in the viewer.  Why not using something very similar to that?  Most of the icons are flat and simple, so as to not be distracting.

manongjohn:
Are you referring to the Camera View icon?  It's a bit too simple for me, yet I also don't want something too elaborate and distracting.  A simple black/white type icon would be good and perhaps not so 3 dimensional to be more consistent with the look of our other icons and images.

i do rather like that the svg doesn't have a background so the column color comes through.

gab3d:
I agree that that kind of 3D icons look rather distracting (and oldish, if I may say).

Why not reuse something from another free project?
For example, here's a nice icon from Blender's camera options panel:


...just an idea.

artisteacher:
camera

konero:
A modern take on @artisteacher's design.

cam_big

**Download:** [cam_big.zip](https://github.com/opentoonz/opentoonz/files/3437667/cam_big.zip)

You could also swap the viewer icons if continuity becomes an issue. 🤷‍♂ 

cam_small

**Download:** [pane_cam_icons.zip](https://github.com/opentoonz/opentoonz/files/3437668/pane_cam_icons.zip)

Feel free to tweak/reject these, no biggie.

gab3d:
@artisteacher personally I don't think it's a good quality icon, sorry.

@konero this one is a good quality icon.
personally I'd go for a more classic camera, camcorders are "old" but not "classic", at least to me. Besides, video always was percieved as low quality compared to film, do we want to associate OT with low quality stuff?

artisteacher:
@gab3d - no worries, it was just a quick draft. It definitely needs some polishing.

I agree that @konero's icon is superior. I think it gives a more modern feel than the classic camera, rather than low quality.  Of course the most modern video camera is just a phone 🙃

In terms of consistency, most of the other icons have a black outline. Is it ok not to have one here? Visually, a flat filled icon with no stroke is the most pleasing and feels modern. I'd love to go in this direction with new icons - especially if they can be color coded in the theme.  Otherwise, it can be challenging finding colors that work across all themes. 

Right now the existing note and audio icons don't feel very cohesive. The note icon is thick black lines. The audio icon has a fill with a very narrow stroke. It would be best if all 3 icons could feel unified.

konero:
@gab3d Classic reel style.

cam_big

**Download:** [cam_big_reel.zip](https://github.com/opentoonz/opentoonz/files/3438202/cam_big_reel.zip)

And same for viewer.

cam_prev2

**Download:** [pane_cam_icons_reel.zip](https://github.com/opentoonz/opentoonz/files/3438204/pane_cam_icons_reel.zip)

gab3d:
@konero 
I like this one! :)

gab3d:
@artisteacher 

> Right now the existing note and audio icons don't feel very cohesive. The note icon is thick black lines. The audio icon has a fill with a very narrow stroke. It would be best if all 3 icons could feel unified.

yep, btw... the Note Level icon kinda sucks to me 😸 
I think the direction of this last camera icon could be taken as a guide to create new icons for sound and note levels (and pretty much any icon that could show in the Xsheet header).

also, as @manongjohn pointed out:
> i do rather like that the svg doesn't have a background so the column color comes through.

I also think that having a transparent background is the best for these kind of icons, as they better integrate with the UI, no matter what theme is used.

manongjohn:
Was testing out the pane_cam_*.svg icons and had to modify the width/height from `100%` to `20` in order display right.

The FX Schematic window has the following camera icon:

image

It's already an SVG, though would need a camera_over.svg version. Should we use this one instead or replace it with the new one?

RodneyBaker:
Yes, definitely use that icon @manongjohn.
That also looks to be about the right size for when/if the camera column (in xsheet) is made smaller in width.

gab3d:
Personally I like Konero's icon a little more.
If there's no technical issues on using it, that is.

The other one is not bad either, though more simplistic.

konero:
Either is fine with me.

Did I mess up the view box? I'm still getting used to Affinity export, sorry. I modified the viewer icons slightly to reduce the prominence of the viewfinder (on the camera) if you decide to use these at all.

pane_cam_on

**Download:** [pane_cam_icons_v2.zip](https://github.com/opentoonz/opentoonz/files/3439390/pane_cam_icons_v2.zip)

And if you want to use the other one here's some pane icons for that.

pane_cam_on2

**Download:** [pane_fx_icons.zip](https://github.com/opentoonz/opentoonz/files/3439409/pane_fx_icons.zip)

The pane is 20x20, so the FX icon might not work if you're trying to be economical with assets. I don't know why it is the way it is I just rolled with it. You could try exporting the icons again without any canvas padding, just crop the icon shape itself and set a container area in Qt for specific sizes with center-center positioning.

manongjohn:
@konero I tried the pane_fx_icons.zip in place of your new ones to match the Stage Schematic camera icon. Although it renders nicely, I don't like the look of it. With it being just a little larger, it oddly looks more childish. Kind of reminds me of Mickey Mouse too.  I wouldn't want it as a camera column thumbnail.

The new icon you created looks much nicer.  It looks so much nicer that, if you don't mind, can you provide a 17x17 off/over version of it to replace the Stage Schematic icon? 😁

konero:
🤣

I made two sets, one is a clone of the file above except with a shrunk document area (it's a tight fit for pixel alignment), I also provided a slightly edited version (smaller) just in-case it looks wonky.

[clone | edit]

camera_off_1 camera_off_2


**Download:** [fx_cam_icon_clone.zip](https://github.com/opentoonz/opentoonz/files/3439563/fx_cam_icon_clone.zip)
**Download:** [fx_cam_icon_small.zip](https://github.com/opentoonz/opentoonz/files/3439576/fx_cam_icon_small.zip)

_Pink BG is just to show canvas area, there's no BG in the SVGs._

gab3d:
I agree with @manongjohn on the icons.

@konero  personally I like the icon on the left a little more. 
regardless of this, both are good to me.

manongjohn:
I used the small icons.  Here's what they all look like:

image

gab3d:
nice! :)

artisteacher:
I think the large icon looks great. The small ones keep reminding me of wind-up toys.

gab3d:
> 
> 
> I think the large icon looks great. The small ones keep reminding me of wind-up toys.

it reminds me to a Sumerian horse rig, but that must be just me 😄 
sumerianhorserig

...or perhaps they mounted cameras on the back of the horses to film their battles ..hmmm! GoPro precursors 🤔

konero:
Only other design I can come up with is based on RED cameras, which are usually just a box with a lens and a handle, at least until you start adding extensions and rails.

cam_big

view2

**Download:** [red_camera_icons.zip](https://github.com/opentoonz/opentoonz/files/3443052/red_camera_icons.zip)
_(xsheet, viewer, fx)_

I'll be travelling for awhile so I might not be able to edit more, but I'm fine with any icon you guys pick.

gab3d:
@konero 
it's nice.
still looks like a 80s camcorder to me, though.

turtletooth:
Why not use the current small icon for the viewer pane and the new big icon for the column? The viewer pane icon area is so small that the simplicity of the current icon works well.

gab3d:
just for the record, I'm perfectly fine with what @manongjohn presented in his last post.
I think that, although there would always be room for touch-ups and improvements, what it is there is well designed, immediately recognizable (i.e. "iconic") and, finally, way better than what we had previously.

RodneyBaker:
Given @shun-iwasawa checklist for merging PRs I'll make the recommendation that this feature be turned off by default and the PR merged.

Rationale:  Those working in the xsheet have developed a solid and specific workflow that does not include the camera column.  Further, the camera column was deprecated for use in Studio Ghibli as unnecessary.  Having the camera column present by default will therefore interfere with that established workflow.

I know this is likely not the preferred option by many but it is the ideal approach based on Shun's direction for merging new PRs.

At some point in the future perhaps the Camera column can be turned on by default after everyone has had time to assimilate it into their workflow.

In the end having the Camera Column is much needed and the sooner it can be added the better.  Keeping it out of the way by default is a reasonable approach that satisfies all users to the maximum extent possible.

What say you all?

manongjohn:
Actually have a question regarding the Insert/Remove Scene Frame and the Camera column.

Currently if you keyframe the camera and then Insert a scene frame before it somewhere on the xsheet/timeline, it does not shift the camera keyframes like it does all other levels.

It's not obvious right now because you can't tell where the camera keyframes are. 

I would consider this a bug. I have a fix to include the camera column regardless if hidden or not, but wanted to make sure this isn't some intended behavior.

RodneyBaker:
Very good question.

I'm not sure about the intention but at first glance it would seem ideal to have the keyframe move upon inserting a new frame. It is useful to have the key be sticky and stay with the current frame as we wouldn't want shots to reframe themselves after we spent time getting the camera view on that frame perfected and then realize we need to adjust the timing.  Given that I'd think the keys should likely follow the frame they were exposed on as if fully assigned to it.  The user would then move it manually as needed.  

Conversely, where the timing is locked down at the outset and it is known where the camera is required to be on any given frame we wouldn't want the camera to shift from that position as we insert that much needed extra frame to say... sync dialogue with lipsync or other timing change.

gab3d:
- Regarding the 'off' state of the camera column visibility, I'm ok with it.
Although, I'd like to have a button in the Xsheet top-left corner with a small camera icon in it, so that users could see there is the possibility of having such a camera column shown in the Xsheet, without having to learn it from somewhere.


- Regarding inserted frames, I think it seems natural to also have camera keys shifted if the whole animation will have one more frame inserted in it.

melieconiek:
I'd prefer that the camera keyframes moved along with the other columns. I don't think you can lock the camera column like you can others, though--that's the usual strategy if you don't want exposures/keyframes to move along with the rest of the xsheet.

shun-iwasawa:
@RodneyBaker

> Given @shun-iwasawa checklist for merging PRs I'll make the recommendation that this feature be turned off by default and the PR merged. Reference #2746.

As I wrote `Altering default user profiles (preferences, panel layout and shortcut) : Fine to be merged.` in #2746 , it's fine to make this feature turned ON by default if you think that it would be helpful for new users. I think I can manage our environment here in Ghibli.
What I wanted to suggest in #2746 was, that the software should leave the option to act as the same as ever, if there may be users who will not need the new feature in their workflow. 

However, regarding this PR, I believe the preference option `Show Camera Column` should be a sub-option of `Show Keyframes on Cell Area` , because it seems that the camera column is only be utilized along with the keyframe UI.

RodneyBaker:
@shun-iwasawa I like the idea of having the Camera column be a sub-option of the Keyframe UI option.
That would likely resolve any potential issues of interference with workflows that only include exposures and don't use keyframing at all.

RodneyBaker:
LGTM
I'll delay merging so that others can test and comment but this is ready to be merged now.

Thank you @manongjohn.

gab3d:
briefly tested, seems to work ok.
very nice being able to select to see any camera from header! 👍 

...not sure about the *black* background for the header isn't too disturbing, perhaps the default grey would just be ok? 🤔

RodneyBaker:
I like the black column background as it let's the user immediately know the camera column is a different sort of column than the other columns.  Of course the camera icon itself does that as well.

manongjohn:
If you are seeing a black background on the column header it's because your stuff directory does not have the updated stylesheets to color the column in blue (see images in PR). The bad part testing Appveyor releases.

RodneyBaker:
Good point @manongjohn 
That takes care of that concern.  :)

Any others?

gab3d:
@manongjohn 
ah... good to know.
anyways it doesn't change much in the end ...I guess black or blue will stand out about the same, so... 
with the camera icon there in the header I don't think another color is needed to make it noticeable, but well... it's just my preference.

and thank you for this great feature!!

RodneyBaker:
I suppose for completeness sake we could ask:

Is that camera column color exposed via the QSS where it could be hacked should a user highly desire a change or is it hard coded?

manongjohn:
Color can be changed in the qss files.

RodneyBaker:
> Color can be changed in the qss files.    

Outstanding.

RodneyBaker:
LGTM

shun-iwasawa:
@manongjohn 
Are you already tackling with narrowing the camera column in xsheet view?
If not, would it be all right if I worked for it?
Here is a draft proposal of UI:
narrow_camera_column

manongjohn:
I am not working on it. Thanks for addressing that request.

Issue #1327, opened 2017-07-13
Some default shortcuts are not compatible with mac laptops 


artisteacher:
My macbook's keyboard does not have a numberpad and my only delete key is the same as the backspace key on other keyboards. This means that with the default shortcuts, I can't use delete. Any shortcuts with a + (such as zoom in or fullscreen) also aren't possible (my + key is part of the = key).

 While I've saved my own shortcuts, I think that the default shortcuts should work for mac laptops. In the same way that the command and control keys are treated as the same key, backspace and delete could be treated as the same key. The same for the numberpad's + key and the = key.

artisteacher:
For anyone else on a Mac laptop, using fn-delete works like delete should without changing the shortcut. 

I could create a custom set of Mac laptop keyboard shortcuts to share as a workaround, but it would still be lovely to get backspace/delete treated as the same key and for +/= to be treated as the same key.

MrBlockCat:
I agree something should be done about that, having a MacBook with a compact keyboard myself, though of course it'd be nice to figure out where and how Control is converted to Command and Alt to Option on macOS so that that same code could be used for Insert to (?), Backspace to Delete, and '+' to Shift + '='!

MrBlockCat:
That said, what should the Ins key be replaced with? We’d need to figure that out both for a long term (macOS detection and conversion of keys) and short term (macOS preset) fix of this issue

PR #2575, opened 2019-03-22
Add shortcuts for MyPaint Brush toolbar options

manongjohn:
User request for shortcuts for the following MyPaint brush toolbar options:

1. `Eraser` - No default shortcut assigned.
2. `Lock Alpha` - No default shortcut assigned.
3. `Pressure` - Uses default shortcut Shift+P
4. `Size` - Uses default shortcuts U/I.  Increments value by 0.06

### DIFF
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -145,6 +145,7 @@ FullColorBrushTool::FullColorBrushTool(std::string name)
   m_preset.setId("BrushPreset");
   m_modifierEraser.setId("RasterEraser");
   m_modifierLockAlpha.setId("LockAlpha");
+  m_pressure.setId("PressureSensitivity");
 
   m_brushTimer.start();
 }
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -144,6 +144,7 @@ FullColorBrushTool::FullColorBrushTool(std::string name)
 
   m_preset.setId("BrushPreset");
   m_modifierEraser.setId("RasterEraser");
+  m_modifierLockAlpha.setId("LockAlpha");
 
   m_brushTimer.start();
 }
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -143,6 +143,7 @@ FullColorBrushTool::FullColorBrushTool(std::string name)
   m_prop.bind(m_preset);
 
   m_preset.setId("BrushPreset");
+  m_modifierEraser.setId("RasterEraser");
 
   m_brushTimer.start();
 }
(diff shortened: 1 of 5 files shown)


artisteacher:
Nice! On a related note, the lock alpha option could also use a shortcut.

manongjohn:
@artisteacher Sounds good. I can modify this PR to include it.

Dare I ask.....are there any other "useful" toolbar option shortcuts that are needed?

Please try not to say all of them :)

artisteacher:
Great! The only one that I really miss is the ability to change the size of the raster brush. Right now it ignores all of the brush size shortcuts - (alt-Ctrl, U, I, H, J).

On a related note, but maybe not the scope of this PR, I'd love to see the ability to set a tool's shortcut to one of it's tool options. Ie. pressing G switches to the Geometric tool. Continuing to press G again would let me toggle through the different shapes. Similarly A switches to the Animate tool. Continuing to press A could toggle through the modes - to scale, rotate, or etc. It would be fewer shortcuts to remember...

Similarly, at some point I'd love to see some resolution for #1327, but that's probably out of scope too.

gab3d:
@artisteacher 
I agree. 
Although the basis of a good workflow is there with the possibility of assigning shortcuts for Animate Tool's **Active Axis:** option, and the **Mode:** and **Type:** options of various other tools, there are still some shortcomings such as the **Mode** shortcut not allowing to cycle the Skeleton or Plastic tools' Mode dropdown menu options.
Achieving that would make assigning a shortcut to Tool Modifiers > Mode much more useful.

Still, I think that having the possibility of using the own tool shortcut to change the main tool "mode" (whatever that is called for that tool) would be extremely handy.

manongjohn:
> Great! The only one that I really miss is the ability to change the size of the raster brush. **Right now it ignores all of the brush size shortcuts - (alt-Ctrl, U, I, H, J).**

So the MyPaint Brush size only has 1 brush size control. Do you want the shortcut to reuse H/J (min), U/I (max) or add a different set of shortcuts?  Where does the alt-Ctrl come into play with these shortcuts?

FYI, I also enabled the existing Pressure shortcut to work with the MyPaint Pressure toolbar option.

manongjohn:
Actually, apparently in other tools, the Size with only 1 control uses shortcut U/I (max), so I'll make sure it uses the same for MyPaint brush size.

artisteacher:
I'd reuse the existing shortcuts - probably U/I since the maximum brush size is used when drawing without a stylus. As far as alt-ctrl (alt-cmd on mac) #685 goes:

> Holding Ctrl + Alt and mouse up/down changes the min brush size. Left/right changes the max brush size.

manongjohn:
I've enabled the shortcut for brush size but it currently increments by a factor of 1.0.  On other controls, this maybe ok since they tend to range from 0-100, but for Mypaint brushes, they go from -3 to 3.  What increments should this go by? 0.1? 0.25? 0.5?

artisteacher:
Maybe .25?

manongjohn:
0.25 seemed too much of a jump. 0.1 wasn't too bad, but I decided to keep consistent with the other size options requiring 99 keypresses to go from 100 to 1, therefore the step is 0.06.   It's actually not that bad since you can hold the button down til you get close to the size you want.

Aside: I still don't get why the mypaint brush size ranges from -3 to 3.

artisteacher:
I don’t get the range myself. It would be nice to have consistency in the brush sizes.

gab3d:
AFAIK the MyPaint brushes size is intended to be changed inside the Settings of each selected brush (i.e. **Radius** parameter), and not in the OT brush. In this way each MP brush can have its own independent size.

turtletooth:
LGTM.  Thanks @manongjohn for the update.

PR #2576, opened 2019-03-23
Allow Similiar Drawing Substitution with selections

manongjohn:
This PR augments the behavior of Similar Draw Substitution commands (ALT-Q/W) to work against a selection of cells.

Currently if a group of cells are selected on 1 or more columns, only the cell with the current focus is used for the command.

Added logic to behave similar to the regular Draw Substitution commnds (Q/W) such that it will perform the command on all cells within the selection area, including the feature of filling empty cells within the selection area.


manongjohn:
Oops. Didnt mean to publish this under my temp branch.  Closing an resubmitting.

PR #2577, opened 2019-03-23
Allow Similar Drawing Substitution with selections

manongjohn:
This PR augments the behavior of Similar Draw Substitution commands (ALT-Q/W) to work against a selection of cells.

Currently if a group of cells are selected on 1 or more columns, only the cell with the current focus is used for the command.

Added logic to behave similar to the regular Draw Substitution commnds (Q/W) such that it will perform the command on all cells within the selection area, including the feature of filling empty cells within the selection area.

### DIFF
--- a/toonz/sources/toonz/xsheetcmd.cpp
+++ b/toonz/sources/toonz/xsheetcmd.cpp
@@ -741,7 +741,8 @@ public:
           frameCount++;
           nextCell = xsh->getCell((r + frameCount), c);
         }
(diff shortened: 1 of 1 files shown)


bobbybutterfly:
@manongjohn  does this mean, i would be able to move a bunch of selected cells in the timeline simultaneously in both directions?

shun-iwasawa:
Jenkins

shun-iwasawa:
It works fine. Thank you @manongjohn for the modification! LGTM

PR #2599, opened 2019-04-20
Fix command for Running OpenToonz in how_to_build_linux.md

ken-taguma:
Separate commands to 2 lines because there is 2 commands in one line at Running OpenToonz section.


turtletooth:
LGTM  Thansk @ken-taguma

PR #2607, opened 2019-05-03
Allow sequenced files starting with 0

manongjohn:
This PR allows sequenced files starting with frame 0 to load.

Apparently some fix to sequenced file detection and handling actually fixed the issue that would cause OT to crash whenever it tried to process a sequenced file starting with 0.

After discovering this I disabled the malformed exception that was thrown when a file started with frame 0. This allows level frame 0 to load successfully.

image

Also, I modified the logic of renaming a level cell to 0. Normally when Autocreation mode is not `Use Xsheet as Animation Sheet` and you double click a filled cell then enter 0, the cell is deleted.  I enhanced the logic to delete the cell only if the level does not have a frame 0, otherwise it will replace the cell with level frame 0.

### DIFF
--- a/toonz/sources/common/tsystem/tfilepath.cpp
+++ b/toonz/sources/common/tsystem/tfilepath.cpp
@@ -687,12 +687,12 @@ TFrameId TFilePath::getFrame() const {
   }
   char letter                  = '\0';
   if (iswalpha(str[k])) letter = str[k++] + ('a' - L'a');
-
-  if (number == 0 || k < i)  // || letter!='\0')
-    throw TMalformedFrameException(
-        *this,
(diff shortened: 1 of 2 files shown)


RodneyBaker:
@shun-iwasawa This trailing zero in an image filename continues to be an issue.
Any help with reviewing and/or merging this PR will be appreciated.
Thanks!

RodneyBaker:
LGTM 
This one is very welcome by me as it addresses the very frustrating issue of Opentoonz interruptions whenever it encounters a file with trailing zero.

RodneyBaker:
Also fixes #469

PR #2608, opened 2019-05-03
Update Travis QT Version to 5.12.3

manongjohn:
Homebrew updated QT version to 5.12.3.

Added logic to automatically detect the version of QT installed by homebrew so we shouldn't have to keep updating the script every time homebrew decides to update QT version.


shun-iwasawa:
Thank you @manongjohn it's nice & clever addition ! LGTM

PR #2612, opened 2019-05-07
Fix shift/ctrl + right-click empty column crash

manongjohn:
This PR fixes a crash reported by a user on Discord.

If you select an empty column after the last filled column in the Xsheet and then use `SHIFT` or `CTRL` + right-click on another empty column, OT will crash. This issue only occurs with the virtual columns that follows the last filled column in the Xsheet. 

Cause of the crash was because it was trying to access column information for a column that doesn't really exist.  Added logic to prevent the crash.


shun-iwasawa:
Jenkins

shun-iwasawa:

PR #2613, opened 2019-05-07
Fix timeline multi-cell dragging into empty columns

manongjohn:
This PR fixes an issue reported by a Discord user.

The issue occurred when selecting multiple cells across multiple columns in the Timeline and attempting to drag it to empty cells in adjacent columns. The selection box would turn red and not allow the cells to move into the new position.

I found the bigger the cell selection area was, the more likely the cells could not be dragged into empty cells in adjacent columns.

Removed incorrect timeline-only logic that was erroneously limiting what columns the selected cells could be dropped in.


shun-iwasawa:
Jenkins

shun-iwasawa:

PR #2614, opened 2019-05-10
Fix "Convert to Vectors" command for TLV with MyPaint Styles

shun-iwasawa:
This PR will fix `Convert to Vectors` command to preview / work properly when the target level has a palette containing MyPaint styles.

Toonz Raster Level may have palette including MyPaint styles, which cannot be rendered in vector levels.
In such case `TMyPaintBrushStyle` will be replaced by `TSolidColorStyle` with the same color in the converted result.

### DIFF
--- a/toonz/sources/toonz/vectorizerpopup.cpp
+++ b/toonz/sources/toonz/vectorizerpopup.cpp
@@ -190,6 +190,17 @@ void getSelectedFids(std::vector<TFrameId> &fids, TXshSimpleLevel *level,
   for (fst = fidsSet.begin(); fst != fsEnd; ++fst) fids.push_back(*fst);
 }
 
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
Jenkins

RodneyBaker:
I encountered a crash when attempting to test this although I don't think it directly relates to this PR.
It may be more related to Mypaint issues of the past where transitioning from normal brush to Mypaint brush has problems on some systems.
Will test more.

RodneyBaker:
After more testing all appears to be working well.
LGTM.

RodneyBaker:
After more testing.. .LGTM.
To reiterate what @shun-iwasawa 
Merging.

PR #2558, opened 2019-03-12
Fix updating Style Editor when switching between level palette and studio palette

shun-iwasawa:
This PR will fix #2557

Issue #2557, opened 2019-03-12
Style Editor does not refresh when switching between level palette and studio palette


shun-iwasawa:
### Issue Summary

The Style Editor does not update when click and switch between the level palette and the studio palette under some condition.

### Steps to Reproduce

1. Load some level.
1. Open Palette, Studio Palette and Style Editor. 
1. Select some palette in the studio palette tree.
1. Click the style # 1 in the level palette.
1. Click the style # 2 in the studio palette.
1. Click the style # 1 in the level palette again and see if the Style Editor refreshes.

### Expected Results

The Style Editor should refresh to the style # 1 of the level palette.

### Actual Results

The Style Editor does not refresh, the style # 2 of the studio palette is still displayed.
You need to click another style to refresh the Style Editor.

### System Information
* **OpenToonz Version**: latest 
* **Operating System**: Win7

PR #2615, opened 2019-05-10
Fix updating Style Editor when switching between level palette and studio palette by Style Picker tool

shun-iwasawa:
This PR is related to #2558 .
This resolves the problem that the Style Editor does not refresh when switching between level palette and studio palette by picking style with Style Picker tool.
Please note that the difference from #2557 is that the trigger of switching style is not by directly clicking the style in the palette but by picking with Style Picker tool.

### DIFF
--- a/toonz/sources/include/toonz/tapplication.h
+++ b/toonz/sources/include/toonz/tapplication.h
@@ -49,7 +49,8 @@ public:
   // Current Palette (PaletteController) methods
   virtual TColorStyle *getCurrentLevelStyle() const = 0;
   virtual int getCurrentLevelStyleIndex() const     = 0;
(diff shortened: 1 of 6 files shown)


shun-iwasawa:
Jenkins

turtletooth:
Looks like your Jenkins stalled out.

shun-iwasawa:
@turtletooth Thank you for the notification. It was due to planned outage in our office last weekend.
Jenkins again

shun-iwasawa:
Jenkins

shun-iwasawa:
I'll merge this by myself as it is bugfix and will (hopefully) not suffer exiting usage.

PR #2621, opened 2019-05-20
File Browser Enhancement for Windows 

shun-iwasawa:
This PR is for enhancing the response of the folder tree view in File Browser by applying the modifications as follows:

-  `DvDirModelFileFolderNode::hasChildren()` will no longer use `QDir::count()` (equivalent to `QDir::entryList().count()` ) which will create a list of all items in the directory. Instead, it will use `QDirIterator` and only check existence of the first item.

- Only in Windows environment `DvDirModelFileFolderNode::getChildrenNames` will use WIN32 API instread of `QDir::entryList()` as it seems to be time consuming in some case such as when accessing to network drives.

**NOTICE**
As a drawback, the latter change will disable to show shortcut files (.lnk) in the folder tree view. You still can access the shortcuts from the item view on the right side of the File Browser.

This PR is WIP as I would like to verify the effect of enhancement in some animation studio, which has been bothered by the slowness.

### DIFF
--- a/toonz/sources/common/tsound/tsound_nt.cpp
+++ b/toonz/sources/common/tsound/tsound_nt.cpp
@@ -1915,7 +1915,7 @@ std::string getMixerLineName(DWORD lineID) {
   assert(false);
   return "";
 #else
-  return std::string(mxl.szName);
+  return QString::fromWCharArray(mxl.szName).toStdString();
 #endif
 }
 
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -565,7 +565,22 @@ void TSystem::readDirectory(TFilePathSet &dst, const QDir &dir,
   WIN32_FIND_DATA find_dir_data;
   QString dir_search_path = dir.absolutePath() + "\\*";
   QDir::Filters filter    = dir.filter();
-  auto addEntry           = [&]() {
+
+  // store name filters
+  bool hasNameFilter = false;
+  QList<QRegExp> nameFilters;
+  for (const QString &nameFilter : dir.nameFilters()) {
+    if (nameFilter == "*") {
+      hasNameFilter = false;
+      break;
+    }
+    QRegExp regExp(nameFilter);
+    regExp.setPatternSyntax(QRegExp::Wildcard);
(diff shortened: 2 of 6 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Changed `DvDirModelFileFolderNode::getChildrenNames()` to use `TSystem::readDirectory` which is now calling WIN32 API instead of `QDir::entryList()` .
Jenkins again

RodneyBaker:
Loading and Browsing appears to be much faster on Windows with this PR.
Looking good thus far.

manongjohn:
@shun-iwasawa 

We seemed to have uncovered a critical bug where files in the extras folder are deleted unintentionally.  It seems to be coming from this PR.

Steps to reproduce:
1) Open a brand new scene
2) Create a new Raster level 'A'
3) Verify the A.0001.tif was created in the extras folder
4) Create another new Raster level 'B'
5) A.0001.tif is replaced by B.0001.tif in the extras folder
To futher show the gravity of the problem...
6) Manually Save Level A and B from above.
7) Verify A.0001.tif and B.0001.tif exist in the extras folder
8) Create another new Raster Level 'C'
9) Both A and B have been deleted and C remains.

shun-iwasawa:
@manongjohn
Thank you for the quick notification of the problem. I’ll check it once I arrive at my office.

shun-iwasawa:
Confirmed the bug and found the probable cause. I'll fix this right away.

shun-iwasawa:
Modified `TSystem::readDirectory()` to consider the name filters. I believe this can avoid the unintended file deletion. Again, thank you @manongjohn for finding the critical problem! 

Jenkins again

shun-iwasawa:
Did anyone test the latest version of this PR? I'd like to include this if it seems to work properly.

RodneyBaker:
@shun-iwasawa, I've tested the latest appveyor and ran through the steps @manongjohn outlined for inadvertent file deletion.  I can confirm I have seen no adverse effects to include deletion of files.
I do think it would be good to have @manongjohn check as he might catch something I've overlooked.

The Browser listing appears to still be as fast as I recall.

RodneyBaker:
Merging as this appears to be working solidly.

PR #2622, opened 2019-05-20
Fix Placing Macro Fx Node

shun-iwasawa:
This PR will fix the following problem in Fx Schematic:

When creating & placing a Macro Fx node by using `Add Fx` command in the context menu of a fx node, the Macro Fx node is placed not next to the selected node (just like when adding a normal fx), but near the XSheet node.


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
I'll merge this by myself as it is bugfix and will (hopefully) not suffer exiting usage.

Issue #1242, opened 2017-05-31
Crash When Editing Lineart using Control Point Editor tool


tghart:
Hello, I have Problem

This problem always happen when i editing lineart using control point editor tool . Opentoonz will be crash and My project always lost :( , I Use Opentoonz 1.1.2 (Built May 26 2017).

For Example :
![giphy 3](https://cloud.githubusercontent.com/assets/25517443/26627258/0789a242-4624-11e7-8491-ba90677bea4b.gif)

Thanks you

turtletooth:
Would you mind sharing the level file this happens on?

beeheemooth:
Just make a vector level, select the Line Tool, and click.
![ezgif-1-8de99790be](https://cloud.githubusercontent.com/assets/25724133/26651389/31bb60ba-4655-11e7-8f4d-e2c464b1d423.gif)

tghart:
this my level file
https://drive.google.com/open?id=0ByaxsBi8eeBDOGRZQmNibHZHU0U

turtletooth:
@beeheemooth I can't reproduce your example.

beeheemooth:
Strange. This is from the current opentoonz_nightlies.
There used to be no such problem. Probably at me the program was incorrectly established.

shun-iwasawa:
@tghart Thank you for reporting the problem. It seems that you uploaded the scene file, not the level file. Can you please re-upload the file `+drawings\BZ.pli`?

@beeheemooth I tried to reproduce the crash with the current nightly build but I couldn't, neither...

tghart:
@shun-iwasawa This https://drive.google.com/open?id=0ByaxsBi8eeBDNUJJQ1J2eHpOWVE

beeheemooth:
Hm.. I installed 2017-06-01 05:42:24 UTC opentoonz_nightlies, but the crash still happens.
To reproduce the error, I do the following:
1. Run OpenToonz.
2. Create Vector Level.
3. Select Geometry Tool > Line.
4. Click on canvas. ( and release, no drag )
An error occurs stably only when the layer is empty. If it already contains a vector, then there is no error.
If you can not reproduce the error, tell me, is it possible that this is due to any problems in the system? (windows 8.1)

shun-iwasawa:
I could reproduce the crash with the sample data from @tghart .

RodneyBaker:
Not sure of the status of this issue.
Any updates?

RodneyBaker:
[...]

PR #2624, opened 2019-05-23
Fix Saving Sub Xsheet Including $scenefolder Path

shun-iwasawa:
This PR will fix the behavior of `Save Scene As` and `Save Sub-xsheet As` commands, especially for a scene containing a level of which the path uses `$scenefolder` alias.

- Made the saved level's path to be automatically re-coded with the new scene folder path, if the level is in  or under the new scene folder.
- When saving sub-xsheet, the current scene will not be switched to the saved one after saving. So I made the level paths to be reverted after saving after executing the `Save Sub-xsheet As` command.


shun-iwasawa:
Jenkins

RodneyBaker:
LGTM
Merging.

Aside:  This may address an issue encountered in #1242 (which was closed as it could not be reproduced).

PR #2627, opened 2019-05-25
Fix macOS portable detection

manongjohn:
This PR addresses the following:

1) Can't detect portablestuff folder
Normally if you rename OpenToonz stuff folder to `portablestuff` and put it in the same directory as the OpenToonz.exe (win)/OpenToonz.app (macOS), OT can be placed on thumb drivers or moved to other locations without having to do an actual installation.

macOS 10.12 (Sierra) put in a change that makes applications added to the system from other sources (I'm guessing not from Apple store) to "translocate" before running. When this happens, the application is temporarily moved/copied to a new directory just before running.  Since it is not running in the same directory as the portablestuff directory it stops working as a portable and will look for the traditional stuff folder.

For macOS versions, this PR adds an additional check for a portablestuff folder inside the OpenToonz.app directory. If found in the same directory as OpenToonz.app or inside it, it will be treated as a portable.

2) Renaming the OpenToonz app causes OT not to work

This is a general problem with renaming the application in macOS.  Code currently expects the name of the app to be "OpenToonz" and builds path based on it. If the user renames it for whatever reason (say they want to keep different versions), OT cannot find the contents of "OpenToonz.app".

When app starts, it captures the name of the actual program from argv[0] and stores it for later use.

### DIFF
--- a/toonz/sources/common/tapptools/tenv.cpp
+++ b/toonz/sources/common/tapptools/tenv.cpp
@@ -223,6 +223,22 @@ public:
         TFilePath(m_workingDirectory + "\\portablestuff\\");
     TFileStatus portableStatus(portableCheck);
     m_isPortable = portableStatus.doesExist();
+
+#ifdef MACOSX
+    // macOS 10.12 (Sierra) translocates applications before running them
+    // depending on how it was installed. This separates the app from the
+    // portablestuff folder and we don't know where it is so we stop treating it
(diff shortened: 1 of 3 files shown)


artisteacher:
I'm so excited about this!

RodneyBaker:
Verified all that I could for this one with Mac users reporting all appears to work.
Merging.

PR #2631, opened 2019-05-31
Skip long $scenefolder levels list in the warning dialog

shun-iwasawa:
When you save some scene containing levels using `$scenefolder` path to other folder using "Save Scene As" command, the warning popup appears notifying that such level will lose link.
For now the level list in the popup shows the list of all levels using `$scenefolder` path. If the number of such level is too large, the list can push out the boundary of the popup so that it runs out the screen.

This PR will make the list to be skipped if the number of the levels is more than 10 in order to prevent such problem like as follows:

image


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

turtletooth:
LGTM.  Thanks @shun-iwasawa for the update.

PR #2632, opened 2019-06-01
Fix copyDir for macOS

manongjohn:
This PR fixes an issue with the TSystem::copyDir() function that affect macOS (and likely Linux).

Currently when it copies the files from a source directory, instead of creating a destination directory with the file in it. it creates a file with the format "destination\file"

Ex:  /rootdir/sourceDir/sourceFile -> /rootdir/destintationDir\sourceFile

Replaced the logic that created the new path using a direct string concatenation with "\\" as the directory separator. It now uses TFilePath concatenation and then extracting the full string allowing existing logic to determine the correct path separator to use.

### DIFF
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -298,9 +298,10 @@ void TSystem::copyDir(const TFilePath &dst, const TFilePath &src) {
       TFilePath srcDir = TFilePath(fi.filePath().toStdString());
       TFilePath dstDir = dst + srcDir.getName();
       copyDir(dstDir, srcDir);
-    } else
-      QFile::copy(fi.filePath(),
-                  toQString(dst) + QString("\\") + fi.fileName());
+    } else {
+      TFilePath srcFi = dst + TFilePath(fi.fileName());
+      QFile::copy(fi.filePath(), toQString(srcFi));
+    }
   }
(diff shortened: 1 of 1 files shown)


RodneyBaker:
This PR could use a good testing by someone that builds on Mac OS.

RodneyBaker:
@manongjohn It looks like we'll need to use your validation of this PR working on a Mac to press ahead with merging.  Let us know how you prefer to proceed.

artisteacher:
I'll build this over the weekend. I'm just not sure how best to test it...

manongjohn:
I discovered this issue when I added OTX's ability to detect for existing installations of OpenToonz and copy the various settings/preference files and folders to the portablestuff folder so you could easily start up OTX with your current settings and not have to set it up all over again.

Reviewing official code, it does not look like users could easily run into this particular code. There is some obscure path and specific conditions to get there, something with a directory name ending in "_files", but this might have been related to the Tab product and/or versions prior to OT's release.  I cannot seem to find the right way to trigger this piece of code naturally in OT.

Since the fix is in OTX and is actively being used for anyone that has copied installed settings into the portable without issue, I will go ahead and merge this in case someone adds a new feature that may use this code in the future.

artisteacher:
I love that OTX asks to copy preferences and rooms! I wish the normal installation was more like that.

PR #2635, opened 2019-06-04
Canon DSLR and Stop Motion Support

turtletooth:
This will add support for Canon DSLR cameras for stop motion animation.

It creates a new pane to control stop motion:
stopmo

Instead of stop motion happening in a popup, live view and onion skin are overlayed on top of the main viewer, making it easier to watch your progress as you go.

This is still a WIP - there are still plenty of features to add.  As of the initial commit, live view works as does capturing live view images to the xsheet.  

@shun-iwasawa I put this up here to let you know what direction it is going and to get the ball rolling on having the Dwango lawyers look at the licenses of the included libraries.

There are two new libraries introduced with this PR:
1. The Canon SDK.  This is proprietary and I don't think should be hosted here on GitHub publicly.  This may break travis or appveyor.  However, a private repo could be used as the source of their builds.  

2.  libjpeg-turbo This is necessary for live view to run at smooth speeds.  Standard libjpeg is much slower.  Perhaps we could replace the current libjpeg with turbo.


gab3d:
Thank you @turtletooth !
As part of a studio that also does stop-motion animation, I'm eager to see where this development could lead us to.

turtletooth:
I just want to throw out a little bit more of my thoughts about where this is going to get some feedback.

DSLR cameras produce both live view images and final images.  The live view images don't usually accurately reflect depth of field, and have much lower image quality among other limitations.  The benefit of live view images is that they are small.

Software like DragonFrame captures both live view and final image on capture pressed and you can toggle back and forth.  This allows you to see the final image, but perhaps with less performance, or you can use the live view images with snappy performance.

Originally I was going to do the same thing.  And expose the two types on two different columns.  

However, I loaded a sequence from DragonFrame into OpenToonz using full size images and the performance was slow.  (This was 120 6MB jpgs.)  So using full size images isn't a great choice.

libjpeg-turbo can scale images pretty quickly.  So I'm currently thinking of not keeping the live view images at all, saving the full size photos to a subfolder, and loading a scaled down final image into OpenToonz on capture.  I could make keeping the live view image an option, but I don't think it would be necessary.

Does anyone have feedback on this workflow?

artisteacher:
When I do a pixilation unit with my students, we've had issues with really, really, really slow performance if the camera's quality settings/image size are too high. I've had much better results when using the smallest image size instead. I can definitely see how large/high quality full size images would not be a reasonable option at the moment.

I'm curious, would the scaled down image still be what is displayed after finishing the capture setting? Or is that what you mean by the live view image? 

Also, thanks so much for getting this started @turtletooth. I am so ridiculously excited about this!

turtletooth:
> I'm curious, would the scaled down image still be what is displayed after finishing the capture setting? Or is that what you mean by the live view image?

Yes, the final full size image would only be saved to a subfolder, but never actually loaded into an OT level.  The image that would be loaded to a level would be a scaled down version of the final image, scaled down to the current scene camera size ( or slightly larger, I'll see where the balance is between performance and resolution).

You could then render from OT, or import the larger images into a video editor or back into OT with slower playback.

RodneyBaker:
DSLR support in Opentoonz ... much appreciated!

I started to write my thoughts but I'm not sure how relevant at this point.
I'm confident what you are seeing will send you in the right direction.

The end image is the goal but we would hate to lose live preview.

turtletooth:
Oh - If I made it seem like we would lose live preview, that's not what I meant to say.
The live view images would still be used for live preview and onion skinning.  I just wouldn't create a level for them.  DragonFrame creates what OT would call two levels - one for lower quality images and one for higher quality levels.

RodneyBaker:
In that case, all is good.  Very good.

artisteacher:
> The image that would be loaded to a level would be a scaled down version of the final image, scaled down to the current scene camera size

Sounds good to me.

> DragonFrame creates what OT would call two levels - one for lower quality images and one for higher quality levels.

Personally, I don't see any need for 2 levels.

gab3d:
@turtletooth 
I see your approach as the most reasonable one, given the performance issues you describe.

Perhaps if OT had level *Proxies* feature (as in After Effects, etc.). i.e. the lowres version could be a **Proxy** of the original captured level footage so, if needed, the user could at any point toggle off the Proxy to see the real fullres footage instead, and then toggle it back on to see the lowres again. 
This could also be used to edit the footage in lowres and then render using the fullres one, by using a global Scene Setting:: **Disable proxies in render**, or perhaps an option could also exist in each Level Settings window: **Disable level proxy in render**, for a more granular control of the feature.
The **Proxy Path** for a level could be edited (set, changed, removed) via the Level Settings window, much like there is currently a *Scan Path* for keeping track of the original scanned artwork of a Toonz Raster level. There should also exist an easy way to toggle the proxy of a level on/off, perhaps by right clicking it in the Xsheet and/or a visible switch in the Scene Cast or Level Strip panes.

Not having this in place, I guess we can always use Replace Level to switch it to fullres if needed.

All in all, great work Jeremy!
Thank you!

turtletooth:
I've gone ahead and made it work the way described above:

A lower resolution proxy file is generated on image import and placed in the timeline.
The full resolution image is saved in a subfolder.

As it stands, the lower res proxy still looks really good.  Unless you are planning on panning and zooming a lot, they will actually be good enough for many cases.  They are still always at least 1920 x 1080.

I've coded it to always be at least as wide in resolution as the scene camera.  I should make sure that it doesn't do this for really large scene cameras, or the memory usage would skyrocket.  

We definitely need the proxy images.  The uncompressed images coming from my camera are around 80mb in size.  Imagine even 50 of those in a scene -- 4gb of memory usage there.  

Right now performance still feels pretty snappy and the live view/onion skinning is pretty fast. 

Next I need to code controlling iso, aperture and shutter speed from OT.

shun-iwasawa:
@turtletooth 
First of all, thank you for tackling with the DSLR support. It must be very useful.
I'm not yet sure whether the libraries are suitable for including into OT. Please let me have some time for checking the licenses of them.

Regarding the image quality, FWIW OT has `subsampling` option in the level settings which allow you to load simplified images when displayed in the work area in order to have a faster visualization and playback. The full-sampled image will be loaded when you preview or render the scene.
Please refer to [here](https://opentoonz.readthedocs.io/en/latest/working_in_xsheet.html#editing-level-settings) for details.
(This is off topic, but I think the term "subsampling" is the opposite to what this value actually does... This value will make the sampling discrete.)

artisteacher:
Is subsampling supposed to change the size of the level's frames in comparison to the camera? Changing the subsampling to 2 for a level makes it look twice as big in the camera, bringing it out of scale with everything else.

gab3d:
@artisteacher 
it's not.
I've just tested it and it behaves normally on my machine (Windows 10).
Perhaps if you have a Mac could be something related to the High DPI display? 🤔

gab3d:
@shun-iwasawa 
Correct me if I'm wrong, but while the use of the *Subsampling* option will alleviate the slow display issue, there would still be a high use of RAM, since you'd be working with the fullres frames all the time.

shun-iwasawa:
@artisteacher Weird. I've tested with my MacBook and the subsampling option seems to work OK.

@gab3d 
When you set the subsampling value to more than 1, shrinked raster data will be cached while you just put and play it on the viewer with camstand mode (therefore, it saves RAM). Once you try to draw on the image or to render it, full sampled raster data will be loaded (consuming larger RAM at this moment).

gab3d:
@shun-iwasawa 
Ohh... I see.
Good to know, thanks for the explanation!

BTW speaking of *drawing over raster images*... Could it be possible to load a level in "Read Only" mode?
So that one could load it in OT without the risk of accidentally overwriting it? 
I mean, many times one could load a level just for applying some effects to it as part of the composition process, but you don't want changes made to it being written to the original files, if you inadvertently used a paint operation there and then the Save All command. 
It would not be that different from the Lock icon we currently have for Columns, but at the "Level" level, so no matter where a locked level is used in a scene, you know it could not be touched.

turtletooth:
Controlling the camera settings works now:

settings

artisteacher:
@shun-iwasawa - it seems that image levels work just fine with subsampling. However any loaded mp4 level, seems to change size in the camera with subsampling on.

artisteacher:
Camera settings are looking good! Will there be the ability to control focus?

shun-iwasawa:
@artisteacher 
Thank you for the info. Now I could reproduce the problem and posted a new issue #2637 for it.

turtletooth:
@artisteacher Since stop motion is usually done in manual focus mode, I added the ability to use the Canon zoom mode in live view to magnify a certain area to check for focus.  
I made a button that you can pick a point on the screen to focus on.

turtletooth:
@shun-iwasawa Any news on if this will be able to be added to OT with the added libraries?  I don't mind maintaining a separate branch if needed.  This addition would make recruiting new developers harder as they would have to register for and be accepted by Canon to use the Canon SDK.  I'm not sure many new developers will want to jump through that extra hoop to build the software.

@artisteacher If I do need to maintain this separately, I still plan on building a Mac version, so don't worry.

turtletooth:
@artisteacher I added focus controls so we aren't locked to manual focus.

turtletooth:
@shun-iwasawa After thinking about it, I think it might be better for this to be its own fork.  I don't want to make it harder for new developers to come on board or to make to much a hassle for the current build system.

artisteacher:
Personally I prefer manual focus & encourage my students to use it. It is good to have options though.  Will this make it into OTX?

gab3d:
@turtletooth 
I think, the prospect of needing to have two different programs installed, to be able to use the stop-motion part of it, pretty much negates the benefit of incorporating the functionality into a bigger program like OT. 
I mean, there are various options of smaller stop-motion capture programs out there, and the resulting sequences can be loaded into other bigger programs for post work, so why would anyone have another full copy of OT just to be able to do the capture. 🤔 I know it's not that bad, but... let's face it, it doesn't have much sense, from a user point of view.

So I'd try to include this into the main OT version, as much as it's possible...

gab3d:
@turtletooth 
Now that we can control some aspects of the camera from within the program... would it be possible to animate things such as the focus distance over time, so that we can "program" a certain camera action that involves a change in focus (or pretty much any other camera property)?

konero:
@turtletooth Why Canon SDK? Have you seen [digiCamControl](http://www.digicamcontrol.com/)? You can use it as a library, it's free under the MIT licence and supports Canon/Nikon/Sony.

**Homepage**: http://www.digicamcontrol.com/
**Library Doc**: http://www.digicamcontrol.com/doc/development/lib
**Example**: https://github.com/dukus/digiCamControl/tree/master/CameraControl.Devices.Example

Would that not work better?

turtletooth:
There was a comment here about digicamcontrol that was removed.  I looked at digicamcontrol, but it is Windows only and still uses the Canon SDK under the hood.

turtletooth:
@gab3d I understand what you are saying about wanting this all to be in one program.  However, if this is included in OpenToonz, we are not legally allowed to distribute the files needed by the Canon SDK to build the program.  We can distribute the DLLs to run the program, but not the header files or libs needed for building.  So EVERY dev would need to apply for the Canon SDK and wait to be accepted.  

Since 2D animation is the main focus of the software, there are many potential devs who probably wouldn't be willing to jump through the hoops.  I probably wouldn't have when I was first starting on OpenToonz.  There is a good chance I would have seen that requirement and moved on.

This would also require an overhaul of the build system.  Travis and AppVeyor currently build from the public GitHub repo, but without the required files (which we can't put in a public repo), the builds would always fail (like mine here have done every time).

If I need to make my own fork, I will commit to keeping up to date.

konero:
@turtletooth Was trying to think of alternatives since the Canon thing sounds like a pain, only other I know of is libgphoto2 but, no MSVC, [unless you've got nothing to do](https://github.com/microsoft/vcpkg/issues/5075).

You could separate DSLR support as a plugin and maintain it in an official repo at the root of OpenToonz like GTS, that way you'll only need to sign up if you want to develop DSLR? Like the OpenCV dll for the optional plugins.

turtletooth:
@konero I looked at libgphoto2- (darktable uses that) and it can work on Windows, but it actually broke my camera drivers, so I'm hesitant to use that again.

I'm not sure that I can make this work as a plugin.

turtletooth:
I'm going to close this, since I plan on adding Nikon and Sony support down the line, and that will only make things even more complicated.  I'm really close though and will have trial builds available soon.

turtletooth:
I put an alpha build here: https://github.com/turtletooth/OpenToonzPortable/releases/tag/0.0.0.2
It is Windows only for now.
If you are using a webcam and your resolutions give black bars, you may want to check "Use Direct Show . . " in the options tab.
The big focus was adding DSLR support.  Webcams work, but I think there is room for optimization.

turtletooth:
@shun-iwasawa I plan on closing this again.  I just wanted to give you a chance to see what all was going into this PR.  I am using three new libraries (opencv, libjpeg-turbo, and the Canon SDK).  OpenCV gives greater control over webcam options and performance than the current option of using the built in qt stuff.

Most of the new files and lines of code are from opencv.  

The performance of the capture works pretty well and the options given are pretty feature complete.  I still don't expect this to get merged into the main OT, but I wanted to give you a chance to see before it goes off the radar completely.

Here's a look at the features:
mainpage

page2

page3

All preview/onion skin images are placed directly on the viewer.

turtletooth:
On a side note, one of the benefits of using OpenCV for the webcam stuff is that it also could lead to bundling the Dwango plugins.  Those build with the latest version of OpenCV just fine and and some nice effect options for users.

artisteacher:
Wow - that's a lot of options/settings! Are the arrows on each side of the zoom buttons intended to zoom in/out by different degrees?

turtletooth:
Yep.

artisteacher:
It looks like you've thought of everything. I can't wait to give this a try!

gab3d:
As I see it, if these features could ultimately be bundled with OT, they could give the software a whole new purpose and user base.

(I know all the hard part about it, I'm just saying that I think is worth the effort to make it possible.)

Great work @turtletooth !!

turtletooth:
Thanks folks.  @shun-iwasawa Is there a way to conditionally include libs or files for linking only with certain flags?

turtletooth:
@shun-iwasawa I made stop motion to be an optional build setting in cmake.  It is off by default, but this allows the code to be included in OpenToonz.

I also removed OpenCV and libjpeg-turbo from the PR.  

libjpeg-turbo does have an interface is interchangeable with libjpeg, so if we ever wanted to use that instead, I think it would be an easy swap.

I also think that including opencv could be beneficial in multiple ways.  It would allow at least the webcam side of the stop motion stuff to be included by default.  It would also allow the Dwango plugins to be included by default if desired.

As I mentioned, it is possible to enable at least the webcam side of the stop motion stuff without including the Canon stuff, and it wouldn't be too hard of a change.

turtletooth:
Hey look!  It didn't break the build system!

shun-iwasawa:
@turtletooth Sorry for my late reply.

>I made stop motion to be an optional build setting in cmake. It is off by default, but this allows the code to be included in OpenToonz.

Thank you for the modification. I agree making the feature optional with settings of cmake would be a smart way. 
I'm requesting to the customer service representative of canon for the sdk. I'll try this once it will be available. (And oh, I also need to borrow a canon camera from someone..!)

turtletooth:
I added control over some additional camera settings: Image Quality and Picture Style.  Also organized some options:

newSettings

newOptions

artisteacher:
Any news?

manongjohn:
I will include in the Sept OTX build sometime next week.

manongjohn:
@turtletooth Please resolve the conflicts in the PR.

shun-iwasawa:
@turtletooth 
Sorry for the late notice, but the current license agreement (Japanese version) of EDSDK which I agreed with Canon Marketing Japan Inc. seems to allow me to redistribute the software using the sdk to end users ONLY in Japan.

Does your version of the agreement allows worldwide redistribution? Or may I need to ask them for amendment of the agreement?

manongjohn:
Sadly I realize I cannot include this in the Sept release of OTX for a few reasons:

1) I don't have the Canon SDK to build with currently.
2) If I had the SDK, I can't put it up in my github repo which I would need to do since I do my builds remotely through AppVeyor and Travis. 

I would need to switch to local builds and push final builds to the release page which I am not prepared to do right now. 

Sorry folks. This is going to have to wait. :(

artisteacher:
Too bad!

RodneyBaker:
@turtletooth Should "Canon DSLR" be included in the title of this PR due to the licensing restriction to place a little emphasis on that focus of this PR?

turtletooth:
@RodneyBaker I can do that.

turtletooth:
@shun-iwasawa Since this is disabled by default, what are your thoughts on merging this in once the commits are resolved.  It doesn't affect the software if it is disabled in CMake.

shun-iwasawa:
@turtletooth  Sorry for my late response. Please give me some time for last-minute reviewing. Thanks!

toonz/sources/toonz/xshrowviewer.cpp
```diff
@@ -830,6 +849,15 @@ void RowArea::paintEvent(QPaintEvent *event) {
     // current frame
     drawCurrentRowGadget(p, r0, r1);
 
+#ifdef WITH_STOPMOTION
+  StopMotion *stopMotion = StopMotion::instance();
+  if (stopMotion->getPlaceOnXSheet() &&
+      (stopMotion->m_usingWebcam || stopMotion->m_active)) {
+    drawStopMotionCameraIndicator(p);
```
shun-iwasawa: I'm not sure with the DSLR case, but when using webcams IMO the indicator can be hidden if the live view is deactivated.

artisteacher:
> One point I would like to ask - can you include some instructions in ./doc/how_to_build_win.md

This won't be Windows only, I hope...

turtletooth:
Requested changes made.

turtletooth:
> This won't be Windows only, I hope...
Right now, I have no way of building on a Mac, so hopefully someone with a Mac will be able to update the build instructions on the Mac side.

shun-iwasawa:
This is to let you know that the legal check for signing CANON SDK license agreement is in process in Dwango. Thank you for your patience.

turtletooth:
I just realized that there are going to be significant issues with the level strip commands (renumber, duplicate, 2s, 3s, etc).  Using DSLR creates 3 separate sets of images- the main image, the live view image which is used for onion skin and lining up, and a lower resolution version of the main image,  Anytime the level is changed with one of those commands, all 3 sets of images are going to need to be updated.  I can't take that on right now with various things happening in my life.  

That leaves a couple options:
- Make the level a read-only level in regards to those commands
- Leave it as is with the possibility of those commands causing issues
- Ditch this PR
- Find someone willing to tackle the level strip commands once this is merged.

RodneyBaker:
I think I'd vote for #1, make the level read only for those commands.

As for the other options:

> Leave it as is with the possibility of those commands causing issues
Definitely not ideal

> Find someone willing to tackle the level strip commands once this is merged.
Would work well -after- the read only option is pursued first.

> Ditch this PR
What you talking about WIllis?
(I.e. definitely not)

Sorry to hear that life is weighing down on you.
If I can help in any way... send me a PM.  ( I think you still have my email)
Hang in there!

gab3d:
@turtletooth 
Option 1 is the simplest one (and certainly better that 2 or 3, at least for the time being).

artisteacher:
Definitely option 1, at least for now.

shun-iwasawa:
@turtletooth

Thank you so much for waiting. Dwango approved to sign the Canon SDK license agreement.

There is a question I'd like to ask.

In order to abide by the terms in the agreement, they are thinking of slightly revising OpenToonz terms of use.
In Japanese version of Canon SDK license agreement, there are disclaimer terms saying like as follows: 
"In case a digital camera breaks down while using it along with OpenToonz, repair charge may not follow a fee structure based on the warranty of the camera. Dwango shall inform users the above in their responsibility."

Is there similar terms in the US version of Canon SDK license agreement? If yes, can you please copy & paste (or send me via email) the correspondent sentence for reference?
Thanks!

turtletooth:
@shun-iwasawa 

I figured it would be best if I give you the whole developer agreement and let the lawyers take a look at it.

[canon.txt](https://github.com/opentoonz/opentoonz/files/3984234/canon.txt)

turtletooth:
There is an additional file in the download as well.
[readme.txt](https://github.com/opentoonz/opentoonz/files/3984262/readme.txt)

shun-iwasawa:
@turtletooth 
Thank you for uploading US version of the agreement. It seems to be quite different from Japanese one. I think we will only sign Japanese version and revise OT terms of use based on it.

Is this PR ready to be merged?
I think I found some points to modify (like adding proper license files to `stuff/doc/LICENSE` for thirdparty libraries and making OpenCV version flexible) but I can post the change as another PR afterwards.

shun-iwasawa:
LGTM Thank you @turtletooth for the great addition!

Please note that this feature won't appear in the nightly build until after updating the terms of use, followed by modifying the installer script and Jen.kins script.

gab3d:
Thank you @shun-iwasawa for interceding for this to happen, it's great for OT to be able to expand its territory to "adjacent" fields like stop-motion, in this case.

And a BIG thank you to @turtletooth for making such a great contribution to the community!
Kudos to you!!

shun-iwasawa:
This is now available in the Nightly Build.

prokoudine:
Hello. Quick question: what was the reasoning behind going with proprietary Canon SDK rather than using [GPhoto](http://www.gphoto.org/) that is free and open source and provides API for remote capturing, supporting vendors far beyond just Canon? That library is in use by multiple free/libre photography applications that work on both Windows, macOS, and Linux. Thank you!

turtletooth:
@prokoudine Please see the Windows related section here: https://www.darktable.org/about/faq/
libgphoto2 is a pain to make work with Windows.  Having to use winusb drivers can break other software - I've done it.  I have looked extensively for a solution that uses open source solutions but nothing is as seamless as using the Canon SDK.

prokoudine:
@turtletooth Fair enough! Thanks for the explanation and happy holidays :)

gab3d:
@manongjohn 
I just realized I have a small problem with this feature:
- in the Windows menu, its entry (Stop-motion Controls) disappears after the first execution of the program (that is, the first time it's there, from then on it doesn't show anymore).

I managed to make it reappear by using **Reset to Default Rooms** option, and restarting the program, but, then again, the next time I execute it it's gone... 🤔 


I'm running the last nightly ("freshly" installed 10 minutes ago), on Windows 10.

artisteacher:
Will this be enabled for macOS anytime soon?

turtletooth:
@artisteacher I have a license to Dragonframe that I am no longer using.  I would like to transfer my Dragonframe account to you.  Would you have any interest in a Dragonframe license?

morevnaproject:
@turtletooth I am in contact with a community of 100+ teachers, who teach stop-motion animation for children in Russia. They are surely interested to get a free tool as replacement to commercial Dragon Frame. Almost all of them use Windows. So, they asked me how to use it.

So, I've got a Windows installation and Canon EOS 450D. Tested latest nightly build and got the following problem - when I hit "Capture" button nothing appears on XSheet. Am I missing something? (live preview works fine)

canon450d

Thank you!

morevnaproject:
Here is a video how it works for me - https://youtu.be/OZ17pas4EE0

morevnaproject:
I found that if I press capture button **on the camera itself**, then it makes a photo and image appears on timeline in OpenToonz. ^__^

turtletooth:
It looks like your camera is in bulb mode. Can you set the shutter speed to something like 1 second and see what happens? That would explain why the screen never comes back from black screen.  It is waiting for a signal that an image is ready.

morevnaproject:
No, changing shutter speed doesn't helps.

But I discovered another thing - I can manually press "shot" button on camera and then frame appears in OpenToonz on timeline.

Here in this video at 8:38 (I use remote control connected to camera instead of pressing "shot" button on camera itself) - https://youtu.be/JBYUOx0NyF8?t=518

turtletooth:
I just tested on the latest nightly build and it works for me.  Is the Capture button in OpenToonz triggering the shutter on your camera?

turtletooth:
Some possible things to consider:
It may be an issue with the 450D camera model
It may be a Windows 7 thing

Those are probably not the issue, but I'm just trying to brainstorm possible solutions.

morevnaproject:
> Is the Capture button in OpenToonz triggering the shutter on your camera?

Camera makes some sound when "Capture" button pressed (can be heard on video at 9:11 - https://youtu.be/JBYUOx0NyF8?t=551). And onionskin overlay is updated. But the sound is not the same as shutter one.

I have asked some other people to test and give their feedback.

turtletooth:
@morevnaproject I think I figured out what is happening.  Older cameras need to use a deprecated command to take a picture.  Do you still have the camera to help me troubleshoot?

morevnaproject:
Hi @turtletooth! Apologies for delayed reply. ^__^
Yes, I still have the camera and ready to help!

Issue #2384, opened 2018-11-17
Project breaks if parentfolder (project root folder) renamed


Ayyyg:
We have animation class and my teacher is really frustrated that many projects sent by student is broken, It will show "not possible to load scene because it does not belong to any project." error message when trying to load it

after I mess around a while I found that if your folder name is not same as _otprj.xml prefix, the project will be broken and you can't open any scene in this project

so the issue is if a user renamed the project folder after he created project, the scenes can't be load and will show "not possible to load scene because it does not belong to any project."

I think this is really a bad design because no one will expect project can be broken when you renamed project folder name, we have some many student don't realize this and this made my teacher considering to switch software

artisteacher:
I’ve run into this too. I avoid renaming in general- whether it is a level or project since it always causes path issues. 

OpenToonz should recognize when you try to load a project where the project name doesn’t match the directory name. This should prompt a warning dialog “project name does not match the project directory - do you want to replace the project name with the current directory name?” 

On a somewhat related note about renaming levels, when renaming in the Browser, it would be really helpful to have the option to update the level’s path/name in the active scene so that they use the new name.

DarrenTAnims:
I was hit by this originally, until I realised that you just need to rename the project's xml file. Now it's not a big deal to me. I rename projects all the time. It's just a shame that these need to be in sync and I can't (from an external point of view) see why they need to be named the same.

PR #2638, opened 2019-06-09
Add project outside project root to drop-downs

manongjohn:
This PR addresses a minor issue where if you load a project that doesn't fall under the Project Root folders you've configured the Project doesn't appear in the drop-downs as a selectable project. An example of this is if someone shares a project with you and you drop it in a Downloads folder or some other folder you don't normally use for new projects.

I added logic to check if the current scene is not under a configured project root folder and temporarily adds the project name to the Start Up and Project Settings popup drop-down. This give the ability to see the settings in the Project Settings dialog as well start in that project on startup.

Additionally I fixed an issue where the "+directoryname" (i.e. +drawing, +extra) notation is used for level paths but the folder is not configured in the Project settings file.  Assuming there are no folders defined, it will build the path relative to the project folder as normal.

### DIFF
--- a/toonz/sources/include/toonz/tproject.h
+++ b/toonz/sources/include/toonz/tproject.h
@@ -146,6 +146,7 @@ public:
   TFilePath projectNameToProjectPath(const TFilePath &projectName);
   TFilePath projectFolderToProjectPath(const TFilePath &projectFolder);
   TFilePath getProjectPathByName(const TFilePath &projectName);
+  TFilePath getProjectPathByProjectFolder(const TFilePath &projectFolder);
 
   TProjectP loadSceneProject(const TFilePath &scenePath);
   void getFolderNames(std::vector<std::string> &names);
--- a/toonz/sources/toonz/projectpopup.cpp
+++ b/toonz/sources/toonz/projectpopup.cpp
@@ -482,9 +482,18 @@ ProjectSettingsPopup::ProjectSettingsPopup() : ProjectPopup(false) {
 void ProjectSettingsPopup::onChooseProjectChanged(int index) {
   TFilePath projectFp = m_projectPaths[index];
 
-  TProjectManager::instance()->setCurrentProjectPath(projectFp);
+  TProjectManager *pm = TProjectManager::instance();
+  pm->setCurrentProjectPath(projectFp);
+
   TProject *projectP =
       TProjectManager::instance()->getCurrentProject().getPointer();
+
+  // In case the project file was upgraded to current version, save it now
+  if (projectP->getProjectPath() != projectFp) {
+    m_projectPaths[index] = projectP->getProjectPath();
+    projectP->save();
+  }
+
   updateFieldsFromProject(projectP);
   IoCmd::saveSceneIfNeeded("Change project");
   IoCmd::newScene();
--- a/toonz/sources/toonz/projectpopup.cpp
+++ b/toonz/sources/toonz/projectpopup.cpp
@@ -368,7 +368,14 @@ void ProjectPopup::updateChooseProjectCombo() {
       }
     }
   }
-
+  // Add in project of current project if outside known Project root folders
+  TProjectP currentProject   = TProjectManager::instance()->getCurrentProject();
+  TFilePath currentProjectFP = currentProject->getProjectPath();
+  if (m_projectPaths.indexOf(currentProjectFP) == -1) {
+    m_projectPaths.push_back(currentProjectFP);
+    m_chooseProjectCombo->addItem(
+        QString::fromStdString(currentProject->getName().getName()));
+  }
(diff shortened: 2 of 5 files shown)


artisteacher:
This sounds helpful! On a related note, if a project folder is renamed (without the name of the xml being updated), it is no longer recognized as a project #2384.

manongjohn:
@artisteacher
Yeah, I am aware of this and have thought of fixing it with this PR but not entirely sure why it expects it to be named the same.

Perhaps @shun-iwasawa can shed some light on this?

Can there be multiple project files in 1 project folder and you only want to load the one for the scene you are opening? Is there harm in loading whatever project files it finds in the project folder even if not named the same?

I mean I can easily augment the logic to first look for a named project file and load it if found. If not found, find the first project file and use that instead. But i guess I'd need to understand the ramifications of doing that.

shun-iwasawa:
@manongjohn 
Thank you for the modification, basically I'll go along with this PR as it seems to enhance the flexibility of the project management.

> Can there be multiple project files in 1 project folder and you only want to load the one for the scene you are opening?

When you open the scenes from the project created with Toonz Harlequin and then modify the project settings, the project settings files with different suffix ( like `myproject_otprj.xml` and `myproject_prj63ml.xml` ) can be in the same folder. In such case OT already has a logic to prioritize the newer one.

I also basically agree with adding the logic of finding project settings as you suggested.  But I'm concerning some points as follows:
- We need to decide which will be handled as the project name if the folder name and the project settings name are different. If you have the project settings like `[Project Root]/foo/bar_otprj.xml` , which will be the project name,  `foo` or `bar` ?
- We need to decide the priority of the project settings between the one with the same name and the one with the newer suffix. If there are two project settings files in the same folder `[Project Root]/foo/bar_otprj.xml` and `[Project Root]/foo/foo_prj63ml.xml` , which setting will be used ?
- Changing the folder name still may break the link in some cases. The logic will resolve the issue only if you are setting all aliases (`+drawing`, `+scenes`, etc.) under the project root. If you set the alias to the location outside of the project root, you need to update all `scenes.xml` with the new path.

artisteacher:
> If you have the project settings like [Project Root]/foo/bar_otprj.xml , which will be the project name, foo or bar ?

Personally, I would go with the name of the project folder. If you've changed the folder name, you probably intend to rename the project.

> If you set the alias to the location outside of the project root, you need to update all scenes.xml with the new path.

What about having a pop-up dialog when trying to open a scene from such a project - ie. Asset locations in this scene reference an old project name and may fail - then have a few options:
- update all paths in this scene to use the current project name 
- update paths for all scenes in this project 
- load anyway

manongjohn:
>> If you have the project settings like [Project Root]/foo/bar_otprj.xml , which will be the project name, foo or bar ?

> Personally, I would go with the name of the project folder. If you've changed the folder name, you probably intend to rename the project.

Not sure that I agree with being able to change the project name outside of OT by merely renaming the folder which is essentially what would happen if folder name took priority.  Having said that, I prefer to keep the file name as the source of the project name.

> We need to decide the priority of the project settings between the one with the same name and the one with the newer suffix. If there are two project settings files in the same folder [Project Root]/foo/bar_otprj.xml and [Project Root]/foo/foo_prj63ml.xml , which setting will be used ?

I would say the project settings should be pulled based on suffix order regardless of project name on the file.  Obviously we want the one that matches with the newest suffix.  If there isn't a match and there is an _otprj.xml in the directory, then we know the project was last opened/created in OT. This means that was the name of the project when it was opened in OT.

> Changing the folder name still may break the link in some cases. The logic will resolve the issue only if you are setting all aliases (+drawing, +scenes, etc.) under the project root. If you set the alias to the location outside of the project root, you need to update all scenes.xml with the new path.

Most scene.xml files I see have the project folder being relative to the current folder it is in ("..") .  In what scenario can these files be updated to point to an absolute path within the same project folder or even to another project elsewhere?
 
A warning message with options to correct paths might be good, though, if the folder name and the file name are not made to match, you'll always get that warning message even if you've corrected path's in files already.

I think OT should ignore the folder name from a project naming point of view and just use the project file for everything project related.  If a file path is broken as a result of a folder rename, OT should be made to catch it and treat it like a broken level path. Depending on what is broken, some things can already be fixed by the user within OT.

manongjohn:
After I coded it to pull project name from the project file, I realized it was rather confusing looking for a project in the browser window (OT or standard) because it was displaying folder name not the project name inside it.

For this reason, I've changed my mind and think it's better to pull project name from the folder.  It's also a lot less code :)

I guess, if we ever allowed a compressed project file (similar to Flash .FLA file), then it would make more sent for the name to come from the file.

RodneyBaker:
LGTM

PR #2644, opened 2019-06-18
Short Play Feature

turtletooth:
This feature adds the ability to play the last x number of frames instead of playing the entire sequence.  The number of frames is configurable in the preferences.

This was created because sometimes a user might just want to see how the most recent additions are looking instead of watching all of their frames.


gab3d:
...and especially useful for stop-motion capture. 
I like the idea.

pculli:
I'm not sure if this is connected, but sometimes it is also useful to play, say, 3 frames before the actual frame and 3 frames after, specially when you're working on inbetweens

RodneyBaker:
Tested the appveyor and I'm not sure if that has all the required elements.
I also may not fully understand the scope and intent of the feature.
It seems to me that upon invoking the Alt + P shortcut the Short Play works once but then reverts to playing the full Play Range thereafter.

RodneyBaker:
I'm missing something about this feature/PR so perhaps others can guide me through.
I can see the shortcut Alt+P but in using it (and in setting different values in Preferences) I don't see anything approaching a short play in the viewer (or as seen via playback in the xsheet).

I tested this also on the latest OTX released recently and I don't see anything there either.

It's probably something simple but also might relate to conflicts identified since August 2, whereas the OTX build was more recent than that.

manongjohn:
Brief look at the code, it appears you need more than 8 frames to use this?


Fyi, if there are conflicts on a PR, you shouldn't force it to rebuild. Appveyor will not build PRs with conflicts. I'm a little surprised Travis allows a build.

RodneyBaker:
My refresh was primarily intended to allow your recent Travis fixes to resolve the remaining conflicts with Travis and apparently that did.  I was thinking that appveyor checks had all passed but my memory is likely faulty there.  If not closing/reopening... which automatically calls Jenkins... someone would have to call Jenkins in order to trigger that rebuild.

manongjohn:
Resolving conflicts doesn't work that way. Any changes to master that is causing conflicts has to be manually pulled into the branch that this PR is based off of. From there the conflicts can be resolved and commited to the branch. So don't bother force rebuilding as its not going to work.

Opening/reclosing does not reissue Jenkins builds. A new commit is needed by some developers to do that.

turtletooth:
Fixed conflicts and a couple bugs.  This command will only run if there are at least as many frames in the xsheet as in the short play preferences.

RodneyBaker:
Ref:  #2885 as one of the bugs identified and addressed via other issue/PR.

turtletooth:
Updated to reflect #2885

manongjohn:
This PR needs to be updated to reflect the preference changes that were recently merged.

RodneyBaker:
I haven't pressed forward in checking this PR because I'm not quite seeing how the feature works.  While I do note the requirement to have more frames than the number set in Preferences... the 'short play' that I'm seeing appears to rely more on the in/out markers in the Xsheet than anything set in Preferences.

I'll give it another go.  :)

manongjohn:
I tried testing this new feature and found 2 problems
1. I cannot seem to change the value in Preferences.  When I type in the box, it won't accept any input except 0. If I somehow get a different number in there (clicked in Blank Frames then clicked back), it changes it to -1
2. I added the `Short Play` to my Xsheet Tool bar. When using the default `8` frames, it goes back to the beginning of my scene and plays from there.

toonz/sources/toonzlib/preferences.cpp
```diff
@@ -519,7 +519,7 @@ void Preferences::definePreferenceItems() {
   define(blanksCount, "blanksCount", QMetaType::Int, 0, 0, 1000);
   define(blankColor, "blankColor", QMetaType::QColor, QColor(Qt::white));
   define(rewindAfterPlayback, "rewindAfterPlayback", QMetaType::Bool, true);
-  define(getShortPlayFrameCount, "shortPlayFrameCount", QMetaType::Int, 8);
+  define(shortPlayFrameCount, "shortPlayFrameCount", QMetaType::Int, 8);
```
shun-iwasawa: You need to the minimum and the maximum values (to 5th and 6th arguments) to make this value works.
Without the range, the value will be treated as enum controlled by the combo box.

turtletooth: Thanks @shun-iwasawa 
This is my first time dealing with the reworked preferences system.  
I like what you did there, just never had to add anything to it yet.

turtletooth:
OK.  I think I finally got it figured out.

turtletooth:
@shun-iwasawa I realize that we are at a feature freeze right now, but with the new addition of stop motion into the workflow, this would be a useful feature to slip into the new release.

RodneyBaker:
Very close Jeremy.
I still get the unfortunate return/resetting to frame 1 after using Short Play that @manongjohn identified.  Shortplay appears to work well other than that.

Aside:   If value of 8 is minimum for shortplay recommend users not be able to enter numbers less than that minimum into the Preferences. Alternatively, perhaps appended text at that location in Preferences to state that minimum.

RodneyBaker:
Aha!  The resetting to frame 1 is due to 'Rewind after Playback' being on in Preferences.
With that toggled off Short Play works as anticipated.

Will merge with expectation that additional enhancements (and perhaps even a dedicated icon in the viewer?) might be added in the future.

Thank you @turtletooth.

RodneyBaker:
Adding this as a post mortem/note to self:

We need to remember to add Short Play to the Play menu as part of the next UI Scrub.

Play ShortPlay

manongjohn:
This still doesn't seem to be working.  I just tried it on the artifact, the nightly and even in the upcoming OTX version and it always goes back to frame 1 and plays the entire scene even thought the default is 8 frames, and I'm on frame 46.  I tried changing it to 10 and it still did it.  

Am I missing something?

artisteacher:
I'm not seeing it work either on macOS - it always plays from frame 1 and keeps going past the 8 frames.

RodneyBaker:
If 'Rewind after Playback' in Preferences > Preview (above the Short Play entry) is on then the return to Frame 1 occurs.  If set to off Short Play plays the frames as expected.

As users likely won't anticipate that turning off 'Rewind after Playback will resolve this oddity I'd say Short Play should either ignore that setting or (if on) treat that as rewinding to the Short Play initial frame reference.

manongjohn:
I turned off 'Rewind after Playback' and it still didn't work

manongjohn:
Ah..I just figured it why it wasn't working.  If you have Start and Stop markers enabled, it always plays the range.  You have to turn off Start/Stop markers for the Short Play to work.

I think it should work regardless of Start/Stop marker status

gab3d:
I agree

PR #2645, opened 2019-06-19
Stop Width Change on Film Strip on Resize

turtletooth:
This temporarily sets the film strips size to a fixed width when the window is being resized (this includes maximized).  This allows the width of the film strip to be restored after closing.

Resizing the film strip is usually the first thing I do when I open OpenToonz.

### DIFF
--- a/toonz/sources/toonzqt/docklayout.cpp
+++ b/toonz/sources/toonzqt/docklayout.cpp
@@ -423,10 +423,26 @@ void DockLayout::applyTransform(const QTransform &transform) {
 //------------------------------------------------------
 
 void DockLayout::redistribute() {
+  std::vector<QWidget *> widgets;
   if (!m_regions.empty()) {
     // Recompute extremal region sizes
     // NOTA: Sarebbe da fare solo se un certo flag lo richiede; altrimenti tipo
     // per resize events e' inutile...
+
+    // let's force the width of the film strip not to change
+
+    for (int i = 0; i < m_items.size(); i++) {
+      if (m_items.at(i)->widget() != 0) {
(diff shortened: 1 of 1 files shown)


turtletooth:
This partially takes care of #1490 
When maximizing the window, it equally distributes the new area among all the panels based on their original size.  This only takes care of the film strip.

Another solution is on maximize to set the unmaximized size to also be the size of the monitor.  This would break expected unmaximizing behavior, but I would wager that the frustration of resizing panels is worth it.

manongjohn:
This has been working pretty well for me.  Any chance of adding the Style Editor to this?  2 things I keep having to adjust in the Timeline and Drawing rooms are the Level Strip and the Style Editor.

RodneyBaker:
LGTM.  This is an important addition as the resizing of the Level Strip has been problematic for a long time.

Recommend merging this PR and then adding the similar functionality to the Style Editor as a separate PR.

hayoko:
@RodneyBaker The "PR Pending Review / Merge" tag is unnecessary and verbose, that's quite literally what a "Pull Request" is already, waiting for a code reviewer or two.

I'd just delete it...

RodneyBaker:
I agree @hayoko although we may need a mechanism to assist @shun-iwasawa in knowing when initial testing and review has been completed.  Many PRs are simply awaiting merger and it can be hard to know which haven't seen any testing versus which are simply awaiting a final look and merge.

I'll gladly remove all PR tags as I agree they are mostly unnecessary or other can remove them as they see fit.

RodneyBaker:
Merging as this is a straightforward change.

PR #2646, opened 2019-06-19
Fill Tool for Raster Level

shun-iwasawa:
This PR will enable Fill tool to Raster level.

Please note that currently I implemented only "Normal" type fill. Other types (i.e. Rect, Polyline and Freehand types) are not available yet. Also note that the "Autopaint" option of styles does not take effect yet. Both of them are future development.

The algorithm to define the area to be filled are different between which color of pixel you are clicked:
- If you click at the transparent pixel, alpha channel value will be used for flood fill.
- If you click at the solid or semi-transparent pixel, RGB channel values will be used for flood fill.

### DIFF
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -476,6 +476,21 @@ protected slots:
 
 //=============================================================================
 //
+// FullColorFillToolOptionsBox
+//
+//=============================================================================
+
+class FullColorFillToolOptionsBox final : public ToolOptionsBox {
+  Q_OBJECT
+
+public:
+  FullColorFillToolOptionsBox(QWidget *parent, TTool *tool,
+                              TPaletteHandle *pltHandle,
+                              ToolHandle *toolHandle);
+};
+
+//=============================================================================
+//
 // FillToolOptionsBox
 //
 //=============================================================================
--- a/toonz/sources/include/toonz/fill.h
+++ b/toonz/sources/include/toonz/fill.h
@@ -81,6 +81,9 @@ void DVAPI rectFillInk(const TRasterCM32P &ras, const TRect &r, int color);
 void DVAPI fillautoInks(TRasterCM32P &r, TRect &rect,
                         const TRasterCM32P &rbefore, TPalette *plt);
 
+void DVAPI fullColorFill(const TRaster32P &ras, const FillParameters &params,
+                         TTileSaverFullColor *saver = 0);
+
 //=============================================================================
 //! The class AreaFiller allows to fill a raster area, delimited by rect or
 //! spline.
--- a/toonz/sources/include/toonz/ttileset.h
+++ b/toonz/sources/include/toonz/ttileset.h
@@ -146,6 +146,7 @@ public:
   void add(const TRasterP &ras, TRect rect) override;
 
   const Tile *getTile(int index) const;
+  Tile *editTile(int index) const;
 
   TTileSetFullColor *clone() const override;
 };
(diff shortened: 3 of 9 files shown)


shun-iwasawa:
Jenkins

beeheemooth:
This is very good news! Is the "Frame Range" option for the Raster Fill Tool planned? Thanks!

RodneyBaker:
Alpha channel flood fill.  Most outstanding.  
I'm not sure I've seen that in any software.

Thanks Shun for Flood Fill in raster levels.
That's a very welcome change for Opentoonz.
I dare say it may be the most requested feature to date.

gab3d:
great news @shun-iwasawa !
thank you!

turtletooth:
@shun-iwasawa Thanks a ton for this.  I was trying this tonight and found that it often left an empty row of pixels at the top or bottom (or both) of the shape being filled.  I know that raster fill is a tricky issue, so I don't know if it can be avoided.  The rest of the fill looks great.

artisteacher:
I also saw empty rows at the top and/or bottom - changing the min/max fill depth often resolved the problem. Similarly to Toonz Raster levels, the fill is a bit glitchy when dealing with translucency. That's probably to be expected.
Screen Shot 2019-06-20 at 8 16 10 PM
Working with opaque outlines, the fill works really well!
Screen Shot 2019-06-20 at 8 31 31 PM

shun-iwasawa:
@turtletooth 

>it often left an empty row of pixels at the top or bottom (or both) of the shape being filled

Thank you for the feedback. I modified the behavior of the fill depth value to be the same as both horizontal and vertical direction expansion. Now with the fill depth = 0 will make the fill not protrude left and right half-tone pixels.
Also I made the default values of fill depth from [min=0 max=10] to [min =4 max=12]. (Note that the min value will be used on regular click and the max value will be used on shift+click.) 

Jenkins again

artisteacher:
This is even better! It's really nice & flexible being able to use the modifier to switch to the max value as needed. At some point it could be nice having a shortcut to adjust the min/max values.

artisteacher:
Also, the fill depth is now working much better on the raster levels than the toonz raster levels for translucent styles. I think the toonz raster level fills too much when the minimum fill value is set to 0. 
Screen Shot 2019-06-21 at 11 07 08 PM
The left circle is a made with the blend+paint brush on a raster level with a minimum fill depth of .19. The right circle uses the same brush on a toonz raster level with a minimum fill depth of 0.

turtletooth:
Tested and it is much better.   Thanks @shun-iwasawa This is an amazing addition.
Going to merge this so people can try it out and report on it.  LGTM.

RodneyBaker:
I just downloaded the nightly and tried it out.  

Outstanding!

I haven't tested thoroughly but first impression is that I need to crank up the minimum Fill Depth (I currently have it set to 15 (matching the Fill Depth Max) and the fill floods beautifully.

Thanks very much for this feature @shun-iwasawa !

artisteacher:
@RodneyBaker - so far I've found that the highest fill depth value works really well for filling opaque outlines.  So I've been using shift-click to use the maximum depth value to fill when the outline is opaque. Filling translucent outlines usually works best at a much lower value. So far somewhere between 0 and 2 has worked best for me depending on the brush making the outline. So I keep the minimum fill value very low and adjust as needed - using a normal left-click to fill in this case.

Here's an image with strokes from most of the raster brushes:
nofill

As a test, here's the same image with the background filled in with black:
Fill depth 0:

Fill depth 2:

Fill depth 4:

Fill depth 15:


Something to consider is what types of brushes will people be most likely to use for outlines? If opaque brushes will be used in the majority of cases, perhaps the modifier would be best used for the minimum value instead. This might also be more consistent with the brush tool which uses the maximum size when there's no stylus.

PR #2647, opened 2019-06-21
Fix copy-paste keyframe interpolation

### LINKED ISSUE #2636: Copy-pasted keys always have linear interpolation.

ezgif-1-c339aaea074d

### Steps to Reproduce

1. Change the type of key interpolation in Preferences to any type (except "Linear").
2. Make a simple animation (using Animate Tool).
3. Copy the first key and paste it into the last frame.
### Expected Results

The key (in theory) should be copied with interpolation.
### Actual Results

 Interpolation will always be linear.
### System Information
* **OpenToonz Version**: 1.3

manongjohn:
The PR fixes #2636.

When copying and pasting keyframes, the following logic is used to handle interpolations:

- When pasting a new keyframe or the 1st new keyframe in a set for the column
  - If pasted after the last existing keyframe in the column, the default interpolation will be created based on preference setting.
  - If pasted between existing keyframes, the new keyframe's previous interpolation value is synced with the previous existing keyframe.

- When pasting a new keyframe or the last new keyframe in a set for the column
  - If pasted before the first existing keyframe in the column, the default interpolation will be created based on preference setting.
  - If pasted between existing keyframes, the new keyframe's interpolation is synced with the next existing keyframe's previous interpolation value.

As usual, when copying multiple keys, the existing interpolation between the copied keys is maintained when pasted.

### DIFF
--- a/toonz/sources/toonz/keyframedata.cpp
+++ b/toonz/sources/toonz/keyframedata.cpp
@@ -10,6 +10,7 @@
 #include "toonz/tstageobjecttree.h"
 #include "toonz/tstageobjectkeyframe.h"
 #include "toonz/txshcolumn.h"
+#include "toonz/preferences.h"
 
 #include <assert.h>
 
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins again

shun-iwasawa:
LGTM thank you @manongjohn for the fix!

PR #2650, opened 2019-06-25
Allow viewer arrow keys to shift cell selection

manongjohn:
The PR extends the `Use Arrow Keys to Shift Cell Selection` preference to the arrow key bindings of the Viewer.

Assuming you have `Use Arrow Keys to Shift Cell Selection` enabled...currently,  when you select a cell in the xsheet and move about with the arrow keys, the cell selection moves with the current frame.  However, if you are in the viewer and use the arrow keys to move left or right, the xsheet's current frame moves, but the cell selection remains where it is. 

Updated the viewer logic to shift the cell selection with the frame selection when this preference is enabled.

Scenario in which this is useful:
Say you have Frames 1-10 filled. You click on Frame 2 to select it in the xsheet and draw.  Now you decided you want to add a new frame between Frame 4 and 5. You use arrow keys to get to Frame 5 and hit Insert key.  In the current version, Frames 2 - 10 shift because the cell selection didn't move.  With this PR, the selected cell moves from Frame 2 to Frame 5 with the arrow keys. Now when you hit Insert, it will shift Frames 5-10 down exposing an empty frame where you want it.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -15,6 +15,7 @@
 #include "ruler.h"
 #include "comboviewerpane.h"
 #include "locatorpopup.h"
+#include "cellselection.h"
 
 // TnzQt includes
 #include "toonzqt/tselectionhandle.h"
@@ -1331,6 +1332,7 @@ void SceneViewer::keyPressEvent(QKeyEvent *event) {
     if (changeFrameSkippingHolds(event)) return;
 
     TFrameHandle *fh = TApp::instance()->getCurrentFrame();
+    int origFrame    = fh->getFrame();
 
     if (key == Qt::Key_Up || key == Qt::Key_Left)
       fh->prevFrame();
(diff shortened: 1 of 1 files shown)


gab3d:
I think it's a good thing! 👍 

Talking about viewer and arrow keys...
Currently when **Select** tool is active the arrow keys allow for position tweaking (which is pretty useful), but when using the **Animate** tool there's no such function (that I know of).
Could it be possible to have some sort of Shift+arrows, Ctrl+arrows (...whatever) to have such a thing available too? It's very useful when adjusting column positions by small amounts. 
🤔

artisteacher:
The other thing about the arrow keys, when the XSheet is in focus, you can change columns with the arrow keys and also change frames. With the viewer in focus, you can only change between frames and never between columns. Maybe this is intentional to prevent accidentally modifying the wrong column, since the Xsheet or level strip might not be visible? Personally I'd like to be able to change columns from the viewer with the arrow keys too.

RodneyBaker:
@manongjohn 
As this PR was initiated four months ago it predates your addition of appveyors deploying the stuff directory.  If you can call jenkins (or otherwise trigger a rebuild) it will be appreciated.  Thanks!

manongjohn:
I can't trigger Jenkins out of the blue. Only shun-iwasawa can do that. I can only trigger Jenkins when I actually commit something to a PR.  

Honestly, not even sure what Jenkins is used for on a non-merged PR. We don't see those results. We can't even see why it sometimes fails even though Appveyor and Travis are successful. I find it useless to me other than to know Jenkins builds were not affected.

At any rate, this PR has no change to stuff folder so there is no real reason to trigger an Appveyor rebuild if that's all you were looking for it to have but I triggered an Appveyor build anyway.

manongjohn:
Hmm. Interesting. A Jenk.ns build started after my post about it.  I guess I can trigger it by just posting the word?  Thought it was only on commits? *shrugs*

RodneyBaker:
Assuming @artisteacher's thoughts are best suited for later improvement.
Current PR... LGTM.

PR #2653, opened 2019-06-29
Fix tabletEvent state on release

manongjohn:
This PR fixes an issue with internal tracking of the tabletEvent state not being cleared properly when a tabletRelease event is encountered.

The tabletEvent state not being cleared properly may lead to odd behavior if a mouseRelease event is encountered afterwards.

I suspect this might be a cause of a "Jumping stroke behavior" someone re-reported recently but it is hard to duplicate. The issue may be specific to how a person's tablet is generating tablet/mouse release events so not many people may experience it or may experience it randomly without realizing it.


shun-iwasawa:
Jenkins

shun-iwasawa:
Although I couldn't reproduced the odd behavior with tablet, the fix seems to be reasonable.

PR #2655, opened 2019-07-01
Display Cycled Values in Function Editor Spread Sheet 

shun-iwasawa:
This PR will modify Function Editor's spread sheet, making it to display values outside of the keyframe range if the parameter is cycled.

- Please note that currently OT has two cycle options: one can be set by clicking the "cycle arrow" button in the xsheet ( = cycle option of the stage object ) , and the other can be set via the context menu of function curves ( = cycle option of the parameter ).
- The stage object's cycle option literally cycles values with no offset. Applied to all transformation parameters of the cycled object. Cycle range will be set to contain all key frames of all channels of the object.
- The parameter's cycle option includes value offset so that the curve connects smoothly. Applied to each parameter. 
- The stage object's cycle option has a priority to the parameter's one, if the both options are activated.

So, In this PR I modified the following:

- Displayed the cycled values from the last key frame to the end frame of the scene. You can distinguish the type of cycle option by color of zigzag line (white for stage object's cycle and light orange for parameter's cycle, in the Default style).
- Enabled to set the parameter's cycle option from the context menu of spread sheet.

Currently, the stage object's cycle option is rather "ignored" in Function Editor. The followings are not included in this PR but I'm thinking of modifying in the future, if they will be accepted by users:
If the stage object's cycle option is activated,
- Display the cycled value in the value field of the Function Editor toolbar.
- Display the cycled shape in the Function Editor graph viewer.
- Return the cycled value when the parameter is referred by an expression.

### DIFF
--- a/toonz/sources/include/toonz/doubleparamcmd.h
+++ b/toonz/sources/include/toonz/doubleparamcmd.h
@@ -18,6 +18,7 @@
 #endif
 
 class KeyframesUndo;
+class TSceneHandle;
 
 class DVAPI KeyframeSetter {
   TDoubleParamP m_param;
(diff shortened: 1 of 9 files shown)


shun-iwasawa:
Jenkins

RodneyBaker:
LGTM
Merging.

PR #2658, opened 2019-07-02
Allow to compile OpenToonz for windows 32bit (via mingw)

blackwarthog:
Disable (ifdef) some options incompatible with mingw 32bit for MinGW/GCC compiler only


RodneyBaker:
All checks passed and the code LGTM.
This PR needs a code review to push it on down the line...

RodneyBaker:
Due to the If/Then nature of this PR it should be fine to move forward.
Recommend merging **immediately after** the v1.4 release.
In this way the 32bit Morevna release and subsequent builds (as needed) can take advantage while the official release will not contain the enhancement as it is not required for the official release.

RodneyBaker:
Thanks @blackwarthog 
Looking forward to the next 32bit release of Opentoonz! :)

PR #2665, opened 2019-07-09
Fix errors building with Qt 5.13

manongjohn:
This PR fixes Travis macOS build errors due to building with Qt 5.13.


shun-iwasawa:
LGTM Thank you @manongjohn for the fix !

Issue #1739, opened 2018-01-18
Audio not updating properly in Viewer


artisteacher:
Sometimes when an audio track is removed from a scene, the audio continues to play when scrubbing or playing the animation in the Viewer. Removing the unused levels from the scene has no impact.

Similarly, when editing the audio track directly in OpenToonz with cutting or deleting, the audio does not play/scrub accurately in the Viewer - the audio plays/scrubs as if it had not been edited.

In both cases, switching to another scene and back or restarting OT fixes the incorrect playback. 
This is in OS X 10.11.6 and OT 1.2.

artisteacher:
Someone else discovered this, but turning the visibility of the layer/column off/on updates the playback.

PR #2666, opened 2019-07-10
Fix sound not playing after dragging the sound level

shun-iwasawa:
Sound fails to play once you move the sound level in the xsheet.

To reproduce:

1. Load a sound level to column 1.
1. Click the play button on the viewer console and make sure the sound plays with no problem.
1. Select whole sound level cells by clicking the side bar of the xsheet cell.
1. Drag the sound level cells to column 2. (*)
1. Click the play button on the viewer console. The sound will not play.

(*) Please note you need to drag the _cells_ , not the sound columns in order to reproduce the problem.

Dragging the sound cells can create the empty `TXshSoundColumn` . The function `searchAudioColumn()` should ignore it or OT fails to mix the sound properly.


shun-iwasawa:
Jenkins

artisteacher:
Will this help with #1739 at all? At least in macOS, scrubbing/playing only updates after reloading a scene if an audio track is edited directly in OpenToonz by deleting/cutting or etc.

turtletooth:
LGTM.

PR #2668, opened 2019-07-14
Fix Crash on Disabling Pixel Unit in Russian Interface

### LINKED ISSUE #2667: Crash programs when unchecking
In the Russian interface, clicking on the tick pixels only, the program crashes, it is impossible to remove the tick, as soon as I change the interface to English, everything works. This is in Windows 10, in others I have not checked it yet.
caed809b8e85b718

shun-iwasawa:
This PR will fix #2667 .
Texts in the unit combo box may be translated so the old unit item cannot always be found with the function `findText()` .


shun-iwasawa:
Jenkins

turtletooth:
LGTM.

PR #2679, opened 2019-07-30
Simplify the Selective Option of Paint Brush Tool Behavior 

shun-iwasawa:
This PR will simplify the behavior of "Selective" option of Paint Brush tool.

This will resolve a problem with following steps:
- Select some Toonz Raster Level.
- Select Paint Brush tool.
- In tool options, set the mode to "Lines". (The selective check box will be disabled.)
- Toggle the selective option to ON via shortcut key.
- Try drawing on the level. The tool will unexpectedly paint areas as if it is in "Areas" mode.

For now the mode is automatically switched from "Lines" to "Areas", when the selective option is activated. It seems to be rather confusing since the "Mode" combo box in the tool option does not reflect such change.

In this PR I omitted such behavior for simplicity. Toggling the Selective option will only toggle the selective option.

### DIFF
--- a/toonz/sources/tnztools/paintbrushtool.cpp
+++ b/toonz/sources/tnztools/paintbrushtool.cpp
@@ -401,18 +401,11 @@ bool PaintBrushTool::onPropertyChanged(std::string propertyName) {
 
   // Selective
   else if (propertyName == m_onlyEmptyAreas.getName()) {
-    if (m_onlyEmptyAreas.getValue() && m_colorType.getValue() == LINES) {
-      m_colorType.setValue(AREAS);
-      PaintBrushColorType = ::to_string(m_colorType.getValue());
-    }
     PaintBrushSelective = (int)(m_onlyEmptyAreas.getValue());
   }
 
   // Areas, Lines etc.
   else if (propertyName == m_colorType.getName()) {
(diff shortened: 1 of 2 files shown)


gab3d:
...as a temporary solution it could work, in the long run the correct thing to do would be to make it impossible to activate the Selective option, through the shortcut, while the Mode is in Lines.

shun-iwasawa:
@gab3d 
Thank you for sour suggestion. I made the selective checkbox hidden on Lines mode to clarify that the option is not related to the lines mode.

Jenkins

RodneyBaker:
I confirm that all checks have passed and the Selection option no longer appears in (raster) line mode.
Will merge as this is a very straightforward change.

gab3d:
> Thank you for your suggestion. I made the selective checkbox hidden on Lines mode to clarify that the option is not related to the lines mode.

It's great to have it hidden in Lines mode!

If it could be avoided that the shortcut worked while in Lines mode it'd perfect, since as it is now you can still press F7, and when going back to Areas mode the Selective option would be changed.
(of course I know nothing about the inner workings of doing that, so if it's not easy to achieve it just don't, it's not super important either. 👍 )

PR #2681, opened 2019-07-31
Camera Capture enhancement : Enable to input the frame field without focus

shun-iwasawa:
This PR will enable to input frame number without clicking and moving focus on the frame number field. While opening the camera capture popup, typing numkey will immediately move focus to the frame field.
This enhancement was demanded by Studio Ghibli. Unlike the inbetween animation, the key animation drawings are not with sequential numbers. It has been bothersome for key animators to click the frame field in advance of editing the frame numbers. 

This PR will also modify the following issues:
- Prevent the frame field to be empty.
- Fixed the style sheets so that the level name, frame number and the capture button are with large sized text.


shun-iwasawa:
Jenkins

shun-iwasawa:
Forgive me but I'll merge this by myself in order to test this enhancement in the studio.
If by any change it causes any malfunction I'll address it.

PR #2685, opened 2019-08-01
Modify saving and restoring panels / popups geometries.

shun-iwasawa:
This PR modifies two points regarding saving and restoring panels / popups geometries as follows:

- The last opened floating panel geometry and state are saved on quiting the software. It will be reproduced when you open the floating panel next time. Please note this feature is not talking about restoring the floating panel state on launch - it is already available without this PR.
-  Modified saving behavior of popup geometry, not to save repeatedly while dragging the popup. Now the popup geometry will be saved only when being hidden or deleted. It will fix non-smooth movement when dragging / resizing popups.

### DIFF
--- a/toonz/sources/include/toonzqt/dvdialog.h
+++ b/toonz/sources/include/toonzqt/dvdialog.h
@@ -170,7 +170,6 @@ protected:
 
 class DVAPI Dialog : public QDialog {
   Q_OBJECT
-  static QSettings *m_settings;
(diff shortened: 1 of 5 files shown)


shun-iwasawa:
Jenkins

shun-iwasawa:
Forgive me but I'll merge this by myself in order to test this enhancement in the studio.
If by any chance it causes any malfunction I'll address it.

Issue #2148, opened 2018-07-08
Crash:  Ctrl-z and x while moving a line


ble-h:
### Issue Summary

OpenToonz crashes when I try to remove a vector line using ctrl-x or ctrl-z while moving it. Ctrl-z also doesn’t undo changes to a line properly.

### Steps to Reproduce

Select a vector line using the selection tool. Move it and press ctrl-x while moving it. The line should disappear but OpenToonz will crash instead.

Draw a vector line and then select it using the selection tool. Move it and press ctrl-z while moving it. The line should disappear but OpenToonz will crash instead.

Erase one end of a vector line and then select it using the selection tool. Move it and press ctrl-z while moving it. The line should look like it did before it was partially erased but it will instead look different.

### System Information
* **OpenToonz Version**: 1.2.1
* **Operating System**: Windows 10

RodneyBaker:
@niccoi 
I can confirm the move and delete simultaneously issue here on 1.2.1 Win 10

Interesting problem.
Apparently I've never tried to move and delete a vector line at the same time.
The immediate workaround would appear to be that of delaying the delete operation until the movement is complete.

beeheemooth:
Move line, Ctrl+C, Paste Insert (Ctrl+V), Undo (all this while moving) = dot with lots point

beeheemooth:
the same principle
![ezgif-2-8c35d77978](https://user-images.githubusercontent.com/25724133/42424866-c4595d00-831c-11e8-8280-63fefcfa3dcb.gif)

PR #2686, opened 2019-08-01
(WIP) Fix crash when pressing Ctrl+X while moving a line (#2148)

martinvanzijl:
Fixes the crash that occurs when pressing Ctrl+X while moving a line (#2148). The selected line will now be cut as expected upon pressing Ctrl+X.

NOTE: I have not solved the Ctrl+Z issue yet.


shun-iwasawa:
Jenkins

manongjohn:
I can confirm this fixes the issue when you use Ctrl+X (cut), but I can duplicate this crash when using DEL (delete) or even Ctrl+Z (undo) to undo the stroke.  Basically anything that causes the stroke to disappear while in the middle of moving the selection will cause this to crash.

shun-iwasawa:
I think triggering command with shortcut key while dragging with tool tend to cause an unexpected result.
We should probably think of rather general resolution, like filtering shortcut events if the current tool is in dragging or in busy state, than adding a patch to every command on an ad hoc basis.

martinvanzijl:
I made some further progress on this. It now also handles the scenario where you press Delete while moving the line, use Undo/Redo shortcuts (Ctrl+Z) or open a dialog (e.g. with Ctrl+N). 

However, it may be easier to simply disable these shortcuts in "SceneViewer::event" while the tool is busy with a mouse-drag. Any votes?

RodneyBaker:
@martinvanzijl  Something appears to have gone very wrong here.
365 commits?  700+ documents changed?

manongjohn:
Seems the commit history on your branched got messed up somehow. Not sure what you did to cause it. 

I've done this on occasion when I pull in the latest OT master into my working branch and somehow remerge existing commits again.

It's also causing conflicts so this PR wont build. Needs to be fixed.

martinvanzijl:
@RodneyBaker @manongjohn Yes, I thought it seemed odd...

I think I'll make a new PR altogether and see if I can copy over the conversation history.

martinvanzijl:
Closing this PR since I somehow messed up the commit history.

New PR is #2912.

PR #2500, opened 2019-01-27
Image viewer gesture controls

manongjohn:
This PR adds gesture controls to the following image viewer classes used throughout the application:

- ImageViewer (Flipbook, Color Model, Cleanup, Preview)
- PlaneViewer (Export Level, Create Mesh, Convert to Vectors, Adjust Levels, Adjust Thickness, Antialias, Binarize, Brightness and Contrast, Color Fade)
- SwatchViewer (Fx Settings)

Each class of viewer was modified as follows:

1) Added gesture controls
- 1 finger pan
- 2 finger pinch zoom
- Double-tap to Fit to Window

2) Added right-click context menu and shortcut support for `Reset View` and `Fit to Window` (Plane/Swatch viewers only)

3) The first time the dialog viewer is created the 1st displayed image will be made to "Fit to Window".

PR #2696, opened 2019-08-05
Limit flipbook zoom to 100% on initial loading

shun-iwasawa:
This PR will modify the flipbook behavior introduced in #2500 as follows:

> The first time the dialog viewer is created the 1st displayed image will be made to "Fit to Window".

I modified the flipbook to limit the zoom ratio to 100% when the panel size is large enough, so that the image is shown in actual pixel size without jaggies due to resampling.


shun-iwasawa:
Jenkins

RodneyBaker:
Works quite well.
Merging.

PR #2697, opened 2019-08-05
Fix crash with visualization commands

shun-iwasawa:
This PR will fix the following problems regarding the visualization commands ( `Reset View, Reset Zoom, Fit to Window, Actual Pixel Size, Flip Viewer Horizontally, Flip Viewer Vertically, Reset Rotation and Reset Position` ).

- OT crashes when typing shortcut key assigned any of the above visualization commands, if no Viewer is displayed. (i.e. `TApp::instance()->getActiveViewer()` returns nullptr.)
- Shortcut key for Reset View command becomes not available on Schematic or Flipbook.

To resolve this, I separated the menu items for command bar from the ones for `ShortcutZoomer`.

Also I inherited the following `ShortcutZoomer` commands and made them to work in each panel;
- `resetZoom(), resetRotation() and resetPosition()` in SceneViewer
- `resetZoom()` in Flipbook(ImageViewer) and
- `resetZoom() and fit()` in Schematic. 

In order to keep consistency, `Reset View` will now totally reset the view matrix and `Reset Zoom` will only reset the zoom ratio, keeping the position unchanged.

### DIFF
--- a/toonz/sources/include/toonzqt/imageutils.h
+++ b/toonz/sources/include/toonzqt/imageutils.h
@@ -101,7 +101,7 @@ void DVAPI convert(
         false /*-- ConvertPopup
                  での指定に合わせて、[レベル名].[フレーム番号].[拡張子]のうち、
                                                                                           [レベル名]と[フレーム番号]の間のドットを消す。 --*/
-    );        //!< Converts a saved level to fullcolor, and saves the result.
+);            //!< Converts a saved level to fullcolor, and saves the result.
 
 void DVAPI convertNaa2Tlv(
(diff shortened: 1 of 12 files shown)


shun-iwasawa:
Jenkins

beeheemooth:
Will this PR solve this problem?
ezgif-1-144685b98eca

shun-iwasawa:
@beeheemooth 
Oh, I didn't know that problem and I don't think this PR will solve it.
OT should prevent enlarging at some magnification factor without crash before users find Higgs boson.

RodneyBaker:
All checks passed and LGTM.
Merging.

PR #2698, opened 2019-08-05
Add "Always Overwrite" button to Camera Capture

shun-iwasawa:
This PR will add 'Always Overwrite in This Scene' to the confirmation dialog which opens when the capturing image will overwrite an existing file.

camcap_dialog

The button will avoid opening the popup while working on the current scene, so that users can re-capture drawings efficiently. The popup will open again if you switch the scene.


shun-iwasawa:
Jenkins

RodneyBaker:
All checks passed but I wasn't able to test as some updates to the UI likely rely on files in the stuff directory.

Will need someone else to validate from a full build/install.

cc:  @manongjohn @turtletooth @masafumi-inoue

shun-iwasawa:
@RodneyBaker 
Hmm, I believe that the button is hard-coded and does not rely on any file in the stuff directory... in my environment I could find the button in Appveyor artifact.

Sorry for being selfish but I'll merge this by myself in order to test this enhancement in the studio.
If perchance it causes any malfunction I'll address it as soon as possible.

PR #2700, opened 2019-08-07
Fix Color Model fails to display if it is minimized on loading the image

shun-iwasawa:

When switching the current level, the corresponding color model is automatically loaded if it is set.
On loading the color model image, if the color model panel is too shrinked to be visible, it fails to display the image until calling "Reset View" command.

I added a line to skip doing "fit to the window" if the panel size is empty.


shun-iwasawa:
Jenkins

shun-iwasawa:
I'll merge this by myself as it is quite obvious bugfix.

PR #2717, opened 2019-08-20
Revert text align of function editor sheet

shun-iwasawa:
This PR will revert text alignment of numerical sheet in the function editor from right to left, as requested by some animation studio.


shun-iwasawa:
Jenkins

gab3d:
@shun-iwasawa
sorry to hijack this to ask a question, but could it be possible to have a toggle somewhere to filter out the inbetween values in the spreadsheet?
to me spreadsheet always look so crowded, even if there's just a few keyframes there.

moreover, in all other software that I know of, the dopesheet (i.e. OT function editor's spreadsheet) only shows keyframes, so it'd be more familiar to people coming from other lands too.
thanks!

shun-iwasawa:
I'll merge this by myself as it is quite obvious.

@gab3d Thank you for the suggestion, I'll try implementing it.

PR #2719, opened 2019-08-21
Option to hide inbetween values in the Function Editor spreadsheet

shun-iwasawa:
This PR will add an option to hide inbetween values in the Function Editor spreadsheet, as requested [here](https://github.com/opentoonz/opentoonz/pull/2717#issuecomment-523070571).

hide_inbtwn_values

- The visibility can be switched via context menu of the spreadsheet area.
- The property will be stored in the layout settings of each user's profile, so that the panel will reproduce it on the next launch.
- When the inbetween values are hidden, cells in the middle of step will become semi-transparent.
- I think it would be nice to show the interpolation type in the segment somehow, but I still do not  come up with a good way to do it... I left it as a future issue.

In this PR I also organized the context menu of the spreadsheet, putting interpolation options and step options to sub-menus, `Change Interpolation` and `Change Step`.

### DIFF
--- a/toonz/sources/include/toonzqt/functionsheet.h
+++ b/toonz/sources/include/toonzqt/functionsheet.h
@@ -96,6 +96,7 @@ private slots:
 
 class FunctionSheet final : public SpreadsheetViewer {
   Q_OBJECT
+  bool m_showIbtwnValue = true;
 
 public:
   FunctionSheet(QWidget *parent = 0, bool isFloating = false);
@@ -128,6 +129,12 @@ public:
   // parameter of specified column. Returns nullptr for
   // fx parameter columns.
   TStageObject *getStageObject(int column);
+
+  bool isIbtwnValueVisible() { return m_showIbtwnValue; }
+  void setIbtwnValueVisible(bool visible) {
+    m_showIbtwnValue = visible;
+    update();
+  }
 
 protected:
   void showEvent(QShowEvent *e) override;
--- a/toonz/sources/toonzqt/functionsheet.cpp
+++ b/toonz/sources/toonzqt/functionsheet.cpp
@@ -109,8 +109,8 @@ public:
   }
 
   void drag(int row, int col, QMouseEvent *e) override {
-    int d                             = row - m_oldRow;
-    m_oldRow                          = row;
+    int d    = row - m_oldRow;
+    m_oldRow = row;
     if (d + m_firstKeyframeRow < 0) d = -m_firstKeyframeRow;
     m_firstKeyframeRow += d;
     for (int i = 0; i < (int)m_setters.size(); i++)
@@ -166,10 +166,10 @@ public:
   void drag(int row, int col, QMouseEvent *e) override {
     if (row < 0) row = 0;
     if (col < 0) col = 0;
-    int r0           = qMin(row, m_firstRow);
-    int r1           = qMax(row, m_firstRow);
-    int c0           = qMin(col, m_firstCol);
-    int c1           = qMax(col, m_firstCol);
+    int r0 = qMin(row, m_firstRow);
+    int r1 = qMax(row, m_firstRow);
+    int c0 = qMin(col, m_firstCol);
+    int c1 = qMax(col, m_firstCol);
     QRect selectedCells(c0, r0, c1 - c0 + 1, r1 - r0 + 1);
     m_sheet->selectCells(selectedCells);
   }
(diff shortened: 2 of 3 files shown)


shun-iwasawa:
Jenkins

DarrenTAnims:
That's a great change. Nice.

One thing I've always thought would be useful, would be to have a context menu option (like the one "Show only Animated Values") would be to be able to hide any columns with only 1 value in it; the setup value. So often I've set a position & scale, etc, but then animated 1 other value (like the horizontal position) and it'd be great to only show the column that has more than 1 value in, so I can see the actual animated values and not the columns with  setup values in. I hope that makes sense. It's kinda hard to explain, but it would make the function editor even more useful.

RodneyBaker:
Darren,
If you aren't a regular user of the Channel/Function Curve Editor you may want to look into that as that view is expressly designed for those needing to easily see what channels have been animated and which only have static keyframes.

gab3d:
@shun-iwasawa ohh!! that was faast! :)  ...and thanks so much for listening!!


@RodneyBaker I think Darren knows the usage of the Function Editor well, what he's asking for is a mode where it'd be required 2 keys in a channel to be shown in the speradsheet as "an animated channel".

To speak frankly, what is strange about OT is that you cannot change a transformation value without creating a keyframe for it and that if you happen to delete that keyframe that channel loses the value that the key had; that whole situation leads to unnecessary keyframing when a user just wants to change a parameter and leave it static for the rest of the animation. 

So I can see that what Darren is asking for, points towards alleviating that issue through hiding one-key-channels and showing channels which have 2 or more keys... makes sense to me. 
Although I can't help but think the root of the problem lays elsewhere...

RodneyBaker:
Darren was clear in his request and it's a valid thought.
My general thought is that as animators gain experience things such as the curve editor and views such as those are a natural progression into complexity.  The tools are there but often overlooked and underutilized. I could have stated that better.  Still hopefully can. ;)

There is a little method to my madness also in suggesting use of Opentoonz's deeper animation toolset; what gets used tends to get developed and enhanced.  I perceive the Curve Editor is highly underutilized and yet it's very reason for existence is to facilitate (as simply as possible) working with keyframes and channels (animated or otherwise).

I wasn't trying to hijack the topic and have now done a little bit of that.  
So back to your point here:

> ...what is strange about OT is that you cannot change a transformation value without creating a keyframe for it and that if you happen to delete that keyframe that channel loses the value that the key had; that whole situation leads to unnecessary keyframing when a user just wants to change a parameter and leave it static for the rest of the animation.

Not overly strange but I follow you.
It seems logical that anything beyond default settings would likely generate a keyframe.
Still there is some room for examining the nature of keyframes themselves and I can easily see where a default setting might be set and that 'key' not automatically designated as a 'keyframe'.  It could be a bit like Offsets/Pivots in some programs.  

Conjecture:
It seems to me that any setting on frame zero would adjust the default values while NOT being designated a key.  A difficulty in that approach might be in how Opentoonz takes into account a frame zero (or negatively numbered frames for that matter as well).  I don't believe OT does this.  If it does iIt might store data/values there but just not expose the values to the user in any directly meaningful way.   (This is a bit like Groups in vector levels.  We can create Groups and order Groups from front to back but we don't get much of a view of them otherwise).  If Opentoonz did allow for a staging frame the rest of the program could use values stored there as the designated default values and still not consider them a keyframe.  As such they wouldn't appear in any 'Show Animated Only' listing as they wouldn't meet the required criteria.

gab3d:
@RodneyBaker 
I said *strange* because, all other animation software I've used, lets you modify transformation values without forcing you to add a keyframe, and also, when you delete the last keyframe of an animated parameter, the value that survives that deletion is precisely the value of that last key (and not returning to a default value).

In OT, on the other hand, it seems that a transformation parameter can only contain its "default" value if it doesn't have a keyframe to store another value in it, and if you delete that key, it returns to the default value ...what (to me) was totally unexpected at first.

Strangely enough, that don't seem to be the case with effects, where you do are able change a parameter value without automatically generating keyframes. 🦄 
I think it'd show consistency to get that behavior with object transformations as well, besides being more of an "industry standard".

DarrenTAnims:
Hi all. Thanks @gab3d. Yes I do use the function editor a lot and the curve editor. That's why this feature would help me so much. And to be totally clear, it's not a column with more than 1 key, but a column with more than 1 DIFFERENT key. It's possible (by clicking the key button) to have a key added for each editable animated value, but they can all have the same value. They're not animated values. It's only if you change them that I'm interested in editing them most of the time.

PR #2722, opened 2019-08-22
Fix crash on loading linked macro fx

### LINKED ISSUE #2721: Crash on loading linked macro fx
### Issue Summary

OT crashes on loading linked macro fx under certain condition. 

### Steps to Reproduce

1. Create some fx and create a new macro fx.
macro1
1. Create linked fx
1. Connect them like as follows:
macro2
1. Save the scene as `scene1.tnz`.
1. In the same scene, swap the connection like as follows:
macro3
1. Save the scene as `scene2.tnz`.
1. Try to load the scenes `scene1` and then `scene2`.

### Expected Results

Both scenes should be loaded properly.

### Actual Results

OT crashes when loading one of them.
It is random which scene becomes problematic. (Actually it depends on pointer address of the two macro fx instances when saving the scene)

### System Information
* **Operating System**: Win10

shun-iwasawa:
This will fix #2721 
The crash occurs when linking connected macro fx from the leaf to the root.
The leaf fx may be loaded during loading the root fx tree, in such case the root fx is not ready to be linked.

I modified it in order to prevent the crash, delaying the link procedure until the the root fx is loaded.

### DIFF
--- a/toonz/sources/common/tfx/tmacrofx.cpp
+++ b/toonz/sources/common/tfx/tmacrofx.cpp
@@ -220,9 +220,9 @@ TFx *TMacroFx::clone(bool recursive) const {
     assert(fx);
     clones[i] = fx->clone(false);
     assert(table.count(fx) == 0);
-    table[fx]                                = i;
+    table[fx] = i;
     if (fx == m_root.getPointer()) rootIndex = i;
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
Jenkins

PR #2723, opened 2019-08-23
Fix viewer pan disabled on toggling "Compare to Snapshot"

shun-iwasawa:

1. Assign some shortcut key to "Compare to Snapshot".
1. Type the shortcut key while the viewer is in the camera stand mode.
1. Try to pan the viewer with middle-button dragging. (It won't work.)

I made the "Compare to Snapshot" to override the panning only when the viewer is in preview mode.
Also I made the shortcut key of flip console buttons not to work if they are disabled.


shun-iwasawa:
Jenkins

shun-iwasawa:
I'll merge this by myself as it is quite obvious fix.

PR #2734, opened 2019-08-27
Add next/prev keyframe shortcuts

manongjohn:
Requested on Google forum, this PR adds shortcuts for keyframe navigation controls, Next Key and Prev Key.

The behavior of the shortcut keys changes based on the panel/window you are focused on, coinciding with how the keyframe navigation controls works on that panel:

When focused on...

- Scene/Combo Viewers - Moves between keyframes on the current level in Xsheet/Timeline
- Function Editor - Moves between keyframes in the Function Editor
- Level/Studio Palettes - Moves between color keyframes in either Xsheet/Timeline or Level Strip.
- Fx Settings - Moves between Fx keyframes

All other panels without a keyframe navigation control will use the Scene/Combo Viewer logic.

### DIFF
--- a/stuff/profiles/layouts/rooms/Default/menubar_template.xml
+++ b/stuff/profiles/layouts/rooms/Default/menubar_template.xml
@@ -230,6 +230,8 @@
         <separator/>
         <command>MI_PrevDrawing</command>
         <command>MI_NextDrawing</command>
+        <command>MI_PrevKeyframe</command>
+        <command>MI_NextKeyframe</command>
         <separator/>
         <command>MI_Link</command>
     </menu>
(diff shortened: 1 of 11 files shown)


beeheemooth:
@manongjohn Thank you very much for this feature, and for listening to user requests !

gab3d:
Sorry to ask, but will there be default shortcuts assigned to them?
It would be nice.

What do other people think?

manongjohn:
I was using CTRL+, (prev) and CTRL+. (Next). Basically the Prev/Next Drawing shortcuts with CTRL.

What do other software use, if they have similar keys?

gab3d:
In all main applications use a variant of the Previous/Next Frame shortcuts.
It seems good to me. 👍

RodneyBaker:
LGTM  
Very useful addition.
Thanks @manongjohn

PR #2735, opened 2019-08-27
Add Desktop and Documents location in file browser for Linux

martinvanzijl:
**Problem**: I am using Linux (Ubuntu). In the File Browser, the _Desktop_ and _My Documents_ items don't work. See below:

desktop-location-demo-before

**Solution**: I pointed them to the default paths for _Desktop_ and _My Documents_ on Linux (if the OS is not Windows or Mac). See below:

desktop-location-demo-after

The code applies to any OS besides Windows or Mac. It works for me on Ubuntu, but should also be tested on other Linux distributions, and any other OS used for OpenToonz.


martinvanzijl:
@manongjohn Thanks for your comment. Yes, I really should have used QDir::home(). I'll aim to make a PR to correct that later today.

PR #2742, opened 2019-08-29
Add "Type" option to gradient fxs

shun-iwasawa:
This PR will add a new option "Type" to existing gradient fxs; LinearGradient, MultiLinearGradient, RadialGradient and MultiRadialGradient.

The "Type" parameter will define the type of color interpolation in each period of gradient.
You can choose the type from four options; Ease In-Out, Linear, Ease In and Ease Out.
The default values are "Ease In-Out" for LinearGradient and MultiLinearGradient, and "Linear" for RadialGradient and MultiRadialGradient, in order to keep consistency with older versions.

The gradient fxs are often used for generating reference image for another fx (such as Depth reference for BokehRefIwaFx). This option had been demanded in order to enhance control for that purpose.

gradient

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -240,6 +240,7 @@
   <item>"STD_linearGradientFx.wave_phase"		"Phase"		</item>
   <item>"STD_linearGradientFx.color1"		"Color 1"		</item>
   <item>"STD_linearGradientFx.color2"		"Color 2"		</item>
+  <item>"STD_linearGradientFx.curveType"		"Type"		</item>
 
   <item>"STD_linearWaveFx"		"Linear Wave"	</item>
   <item>"STD_linearWaveFx.period"		"Period"		</item>
@@ -267,6 +268,7 @@
   <item>"STD_multiLinearGradientFx.wave_frequency"	"Frequency"		</item>
   <item>"STD_multiLinearGradientFx.wave_phase"	"Phase"			</item>
   <item>"STD_multiLinearGradientFx.colors"		"Colors"			</item>
+  <item>"STD_multiLinearGradientFx.curveType"		"Type"		</item>
 
   <item>"STD_localBlurFx"		"Local Blur"		</item>
   <item>"STD_localBlurFx.value"		"Intensity"		</item>
@@ -324,6 +326,7 @@
   <item>"STD_radialGradientFx.innerperiod"		"Inner Size"		</item>
   <item>"STD_radialGradientFx.color1"		"Color 1"		</item>
   <item>"STD_radialGradientFx.color2"		"Color 2"		</item>
+  <item>"STD_radialGradientFx.curveType"		"Type"		</item>
 
 
   <item>"STD_randomWaveFx"		"Random Wave"	</item>
@@ -358,6 +361,7 @@
   <item>"STD_multiRadialGradientFx.count"	"Quantity"		</item>
   <item>"STD_multiRadialGradientFx.cycle"	"Cycle"			</item>
   <item>"STD_multiRadialGradientFx.colors"	"Colors"			</item>
+  <item>"STD_multiRadialGradientFx.curveType"		"Type"		</item>
 
   <item>"STD_raylitFx"			"Raylit"	</item>
   <item>"STD_raylitFx.p"		"Center X"	</item>
--- a/stuff/profiles/layouts/fxs/STD_linearGradientFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_linearGradientFx.xml
@@ -4,6 +4,7 @@
      <control>color1</control>
      <control>color2</control>
      <rgb_link_button>color1 color2</rgb_link_button>
+     <control>curveType</control>
 <separator label="Distortion Wave"/>
      <control>wave_amplitude</control>
      <control>wave_frequency</control>
--- a/stuff/profiles/layouts/fxs/STD_multiLinearGradientFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_multiLinearGradientFx.xml
@@ -4,6 +4,7 @@
      <control>count</control>
      <control>cycle</control>
      <control>colors</control>
+     <control>curveType</control>
 <separator label="Distortion Wave"/>
      <control>wave_amplitude</control>
      <control>wave_frequency</control>
--- a/stuff/profiles/layouts/fxs/STD_multiRadialGradientFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_multiRadialGradientFx.xml
@@ -4,6 +4,7 @@
     <control>count</control>
     <control>cycle</control>
     <control>colors</control>
+     <control>curveType</control>
   </page>
   
 </fxlayout>
--- a/stuff/profiles/layouts/fxs/STD_radialGradientFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_radialGradientFx.xml
@@ -5,6 +5,7 @@
     <control>period</control>
     <control>color2</control>
 	<rgb_link_button>color1 color2</rgb_link_button>
+    <control>curveType</control>
   </page>
   
 </fxlayout>
--- a/toonz/sources/stdfx/gradients.cpp
+++ b/toonz/sources/stdfx/gradients.cpp
@@ -13,7 +13,8 @@ namespace {
 template <class T>
 void doComputeRadialT(TRasterPT<T> ras, TPointD posTrasf,
                       const TSpectrumT<T> &spectrum, double period,
-                      double count, double cycle, const TAffine &aff) {
+                      double count, double cycle, const TAffine &aff,
+                      double inner = 0.0, GradientCurveType type = Linear) {
   int j;
   double maxRadius = period * count;
   double freq      = 1.0 / period;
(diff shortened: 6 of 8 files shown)


shun-iwasawa:
Jenkins

RodneyBaker:
LGTM although I wasn't able to test thoroughly via appveyor due to changes in UI and files in Stuff directory.

gab3d:
@shun-iwasawa 
I got a comment about something that's depicted here, but not only applied here but throughout all of OT (Inbetweener, keyframe interpolations, etc.).

The concepts of *Ease-In* and *Ease-Out* seem to be reversed in OT with respect to what is standard in all other graphics programs, where:
- **Ease-Out** represent a slow and increasing acceleration of change, from an initial starting point.
- **Ease-In**, represent a slow and decreasing acceleration of change, up to an ending point.

...and of course this adds a potential source of problems with users from a diverse origin.

[Here in this video](https://www.youtube.com/watch?v=UVox_2py9yU) is a quick sample of what I'm saying, referring to After Effects.

RodneyBaker:
Gab,
I read your comment in the other report and thought I'd respond but didn't want to do that before looking more deeply into the matter.  I still haven't delved deeply so hesitate to respond even now.
  
There are some differences between how interpolations work in Opentoonz.... for instance some programs apply the ease only to the keyframe selected and Opentoonz appears to apply the results to both (this is because it is "Ease In/Ease Out" and not "Ease In" and Ease Out".  Perhaps that is the majority of what you are seeing in your results?

Can you show some examples in Opentoonz where the results of the interpolation are opposite?

gab3d:
@RodneyBaker 
yes, I know OT considers the whole segment for assigning the interpolation, and not just a particular keyframe as most software do, but what is really reversed in OT is the *concept* of what should represent an "ease-in/out", and that what will confuse users of most programs. We use the same words, with the exact opposite meaning. 🤔 

The solution is really pretty simple: just inverting the labels in every place where they appear will do.
(In fact in the Spanish translation I've already done so long time ago, but English spoken users also deserve not to be confused I think, and that's why I'm bringing this to the table here).

manongjohn:
Are you sure about your definitions?  Perhaps AE has it backwards.

Doing a quick search on-line regarding Ease In/Out, I found this:

> In classic animation, the term for motion that **starts slowly and accelerates is "slow in**," and for motion that **starts quickly and decelerates is "slow out."** The terminology most commonly used on the web for these are “ease in” and “ease out,” respectively.

There is also the Ease-in-out, which is what our interpolations appear to be?

This definition seems to go along with with OpenToonz.

gab3d:
I'm 100% sure of what I'm saying, if not I even dare to disturb with the matter. 😸 
This is one of those things that is so clear (to me) that it's even hard to even have to "prove" it, but I totally get it if you people are not so sure.

I've been a professional user/instructor of a lot of animation related programs over the last 20 years (being myself certified instructor of the likes of Autodesk's Maya and 3ds Max, Adobe's After Effects / Premiere) and also used heavily things like NewTek's Lightwave, TV Paint, Blender and whatnot... and believe me, the concept everywhere is opposed to what we have in OT. 
Even in programs that do not directly mention it by name (like ToonBoom) you see every tutorial about it and *ease-out* is a progressive acceleration and *ease-in* deceleration.

[Evidence #1](https://youtu.be/DRapx5NQx24?t=213)
[Evidence #2](https://www.youtube.com/watch?v=yQ-NC0bHTYs)
[Evidence #3](https://www.youtube.com/watch?v=78HGPRO9k9A) 
[Evidence #4](https://youtu.be/Rez--KqP9ds?t=177) (though, I just found Blender have the icons inverted too, I'll report it there too. Still, the names and the functions are correct).

I hope this helps...

manongjohn:
Here is what I noticed about the videos...

Everyone explained Ease In by itself and Ease Out by itself as you did, but when explaining Ease In/Out, they explained it the opposite of the individual definitions.  So everyone is wrong?

Keep in mind that the other software have Ease In and Ease Out as individual options with Ease In/Out as the 3rd choice. OT only has Ease In/Out so it matches how everyone was explaining it.

So why does everyone use that label even if the combined action is different than the individual actions? I think it's more common to hear and easier to say In-Out than it is to say Out-In...well in English anyway.

I think it's fine as is with that understanding. I don't think it needs to change.

RodneyBaker:
I wrote stuff... but @manongjohn said it better.
I can see why you think the term are reversed although I don't think swapping the terms to 'Ease Out/Ease In' would be appropriate.  It's the entire process that is called ease in/ease out.  That's the  accepted terminology.

Should developers pursue such a thing, we could have an individual set of 'Ease In' and 'Ease Out' interpolations that only apply to the current or nearest keyframe. They likely would need to break/peak the curve gong through the control point so that it doesn't effect the angle of entry on the opposite side of the control point though.  If it didn't then we have someone else wondering why they got an Ease In over there when they applied an Ease Out at the selected location.

gab3d:
Well I think there are two important things to consider here:
1. actual meaning of the terms. (they have to mean something, right?)
2. commonplace or standard use. (go with the flow or against it)



(1) 
What means to Ease-out or Ease-in? 
To me, at least, in this context of motion or action:
- "Ease", always refered to a *slow action*, as opposed to a Fast one.
- "Out" or "In" always indicated the direction of that action: being "In" an action that is incoming to a place, and "Out" an action that is outgoing of a place.

As interpolations in the end always refer to a series of frames between two keys, it's natural that the In and Out terms be there for making some reference to those keys. In that sense, I don't see the logic in that the refer to the inverse of what I described above, do you see it? Please make me see it...
I'm really open, but I don't see another way of using these terms for the opposite.

So (again to me) naturally, to Ease Out must mean to leave a position in a slow fashion and, conversely, to Ease In to come a to a position in a slow fashion.

As to the naturality of saying "Ease In/Out" I think is more of something about the language constructions that we tend to use, and not having to do with something specific to this field in itself: i.e. one always goes "in and out" and not "out and in", as we always say "up and down" and "black and white" and not the opposite; to me they're just things we have incorporated deeply in our minds, but don't mean nothing regarding the actual meaning of what we want to express (i.e. we don't want to express that black goes *before* white, do we?) less anything related to something *that* specific as animation interpolation, as to assign to the fact that we *feel better* saying "In and Out" should mean that "In" should refer to the start of the motion and "Out" to the end of it, because we just feel better in saying that in that order for other more basic/instinctive reasons.


(2) 
As for standardized use of the terms, in the context of animation, I've only saw them used as I see them used in the main application I've used over the years.
All of the Adobe software line (including After Effects, Premiere, Animator CC, etc.), all of the Autodesk line (Maya, 3ds Max, MotionBuilder, etc.), and then there is software that does not 
call it Ease in/out, because they just use Bezier tangents for solving this.
So, from an animation software point of view the majority of the main used programs use Ease Out/In the way I described it.
Because I found that Blender is using it the other way round I investigated a little more, and I see that it uses it just in a specific context of Quadratic interpolation, and digging a little bit I've come to realize there are online charts of mathematical easing functions that does exactly what blender does, so they must have used the, to incorporate a battery of premade easing functions.
[Here you can see one of these online charts.](https://kodi.wiki/view/Tweeners),
the problem I see from that is that the people that built that functions and names are not people related with animation in any way, but just with math...  perhaps in a sense of mind it has all the logic to name them as they did, but from an animator perspective I think it has no sense.

Concluding:
Again, all is open to debate, but that's my point of view, for a bottom line:
"Let's do our best to favor what's more of a standard, whenever possible" that in the end will make people's life easier on a day to day basis.

As a counter example, let me expose this case out: 
Blender spent years arguing in favor of its totally strange right clicking selection system, and forcing users to pass through that weird anti-standard way of doing things, just to finally go with the flow and make left click selection as everyone uses. The had pretty good "logic" reasons behind to defend the right click system, but in the end, to complicate the users' lives just for standing strong with a theoretical logic point is a bad decision. 
Users tend to use more the tools that they feel good and natural with, and they don't care much about more elaborated stand points. Whenever you as a developer let a user re-utilize an previously learned concept you gain that users heart, just because nobody wants to be constantly dealing with new unnecessary things to do what he/she already does in another place, if something is different in OpenToonz with respect to a standard it has to be because it is so much better that it's worth the pain to learn the difference! And I'm afraid, that is certainly not the case, my friends, with the Ease In/Out names. 🤔

manongjohn:
Searching online and just looking even at images of easing options in other software, you can see quite a number of images label it as Ease In/Out. I have yet to see it the other way. To me that suggests that is the more widely accepted naming convention for that curve.

Anyway. I'm not disagreeing that it's not logicallly backwards. I understand your points.

For the record, it confused me too when i tried using it after watching people using Ease In and Ease Out, independantly in other software. It literally just made sense to me after this discussion started, so thanks for that. 😁

At the end of the day, its just a label, not an action changing feature like right clicking vs some other action. It's just a visual queue like an icon to tell you to select that option so you can get to the curve to adjust the speed of the object between key frames. 

If this were a vote (wish we had something like that) my vote is to keep it as is.

gab3d:
well, yes... everyone refers to it as "Ease In/Out" when they speak of applying both to a segment, but in fact, when you go to see which one is either one you relaize Out is at the start of motion and ,In at the end of it.

gab3d:
for the record, even in After Effects it says "Ease In/Out" while at the sane time you've already seen for yourself in the video link I posted above, that it calls Out to the beginning and In to the end.

shun-iwasawa:
Thank you @gab3d , I think you are right in the context of animation production. I read Richard William's "Animator's Survival Kit" and found that he says Ease Out from the beginning key drawing and Ease In to the end key drawing.

OT's opposite wording was just inherited from Toonz Harlequin.
I imagine that engineers in Digital Video tended to view the movement from a standpoint of a "segment" between keys. I was thinking the same - a parameter comes Easing In at the beginning of the segment and goes Easing Out at the end of the segment. It's interesting.

I'm sorry to say that currently I myself can't take time for fixing it. But I'll not interrupt if someone would like to correct it with enthusiasm.
And I'll keep the gradient type names of this PR unchanged until the correction for the sake of wording consistency in the software.

gab3d:
@shun-iwasawa 
Thanks for commenting... appreciate.

It's a really easy fix I think (just invert some *Ease In* and *Ease Out* labels; the ones that say both together: *Ease In/Out* should be kept as they are). So let's hope someone could find some time to do it. 🙏

PR #2749, opened 2019-09-02
Fix Toonz Raster MyPaint brush center

manongjohn:
MyPaint brush cursor is no longer centered for Toonz Raster levels.

cursor_offset

Corrected by not applying the center adjustment logic if it's a MyPaint brush.


RodneyBaker:
LGTM

PR #2750, opened 2019-09-03
Enable system path arguments in tcomposer

shun-iwasawa:
This PR enables system path arguments in tcomposer, just like the same way as OpenToonz.
By adding the argument (like `-TOONZROOT "[path_to_another_stuff_folder]"` ) you can override the registry values set in the rendering server node machines. It can ensure that the scene will be rendered under the same environment as the client machine.

### DIFF
--- a/toonz/sources/tcomposer/tcomposer.cpp
+++ b/toonz/sources/tcomposer/tcomposer.cpp
@@ -400,7 +400,7 @@ static std::pair<int, int> generateMovie(ToonzScene *scene, const TFilePath &fp,
   r0 = r0 - 1;
   r1 = r1 - 1;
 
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
@turtletooth
As you can see in this PR Jen.kins will run automatically once the PR is posted.
For a security reason, it is white-list based. Currently members of opentoonz organization and some developers (you, @manongjohn and @blackwarthog) are in the list for testing the feature. I'll add other authors to the list as needed. Thanks!

shun-iwasawa:
Sorry but I'll merge this by myself to verify the change in the studio. If there are any problem, I'll handle it right away.

PR #2503, opened 2019-01-29
Feature to allow right-click delete of guides

yehjaemin:
I thought it would be nice to be able to delete custom ruler guides by
right-clicking them, in addition to dragging them off the ruler. I have tested this commit on OSX 10.14.2.

PR #2751, opened 2019-09-04
Feature to allow right-click delete of guides by yehjaemin (modified)

shun-iwasawa:
This PR is based on #2503 by @yehjaemin , with small modifications as follows:
- Added a condition to limit the cursor distance for deleting ruler guides.
- Modified tooltip to show the instruction for it.


RodneyBaker:
LGTM

PR #2753, opened 2019-09-05
Use std::max and std::min instead of qMax and qMin by otakuto

shun-iwasawa:
This PR is based on #2596 by @otakuto , with resolving conflicts.


turtletooth:
LGTM.  Thanks @shun-iwasawa

PR #2754, opened 2019-09-06
Change Step and Interpolation Type of Selected Multiple Keyframes in Function Editor

shun-iwasawa:
This PR will enhance the behavior of `Change Step` and `Change Interpolation` submenu commands in the context menu of Function Editor spread sheet. Now they will not only modify the clicked key frame segment, but also modify all segments overlapping the selection at the same time.

- When you right click outside of the selected cells, the selection will set to a single cell at the clicked position in advance of opening the context menu. (This behavior is just like the one in the Xsheet)
- `Change Step` commands can now trigger from the shortcut keys. Shortcut keys for the Xsheet commands `Reset Step` , `Step 2`, `Step 3` , and `Step 4` are now available in Function Editor as well.

### DIFF
--- a/toonz/sources/include/toonzqt/functionselection.h
+++ b/toonz/sources/include/toonzqt/functionselection.h
@@ -109,6 +109,28 @@ public:
   void doDelete();
   void insertCells();
 
+  // if inclusive == true, consider all segments overlapping the selection
+  void setStep(int, bool inclusive = true);
+  void setStep1() { setStep(1); }
+  void setStep2() { setStep(2); }
+  void setStep3() { setStep(3); }
+  void setStep4() { setStep(4); }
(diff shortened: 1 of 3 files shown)


RodneyBaker:
I tested this and everything I tested appeared to work fine with the possible exception of using shortcut keys to adjust and reset the Step.  However, I don't think I fully understand the options and workflow described so look forward to others testing.

shun-iwasawa:
@RodneyBaker
The effective target of shortcut keys for adjusting step commands depends on the current selection (whether selecting Xsheet cells or Function sheet cells).

Sorry but I'll merge this by myself to verify the change in the studio. If there are any problem, I'll handle it right away.

PR #2756, opened 2019-09-06
Use QDir::home() instead of getenv("HOME") in file browser

martinvanzijl:
Use `QDir::home()` instead of `getenv("HOME")` when getting the "My Documents" and "Desktop" folders in the File Browser. This is more robust and makes the code easier to read.

This is following up from PR #2735.

### DIFF
--- a/toonz/sources/toonz/filebrowsermodel.cpp
+++ b/toonz/sources/toonz/filebrowsermodel.cpp
@@ -51,13 +51,10 @@ TFilePath getMyDocumentsPath() {
   return TFilePath((const char *)[documentsDirectory
       cStringUsingEncoding:NSASCIIStringEncoding]);
 #else
-  std::string path(getenv("HOME"));
-  if(path.empty()) return TFilePath();
-  path += "/Documents";
-  QString pathAsQString = QString::fromStdString(path);
-  QDir dir(pathAsQString);
+  QDir dir = QDir::home();
+  dir.cd("Documents");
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@manongjohn You are familiar with this one.
Can you give this a review and push toward merger?

toonz/sources/toonz/filebrowsermodel.cpp
```diff
@@ -51,13 +51,10 @@ TFilePath getMyDocumentsPath() {
...
       cStringUsingEncoding:NSASCIIStringEncoding]);
 #else
-  std::string path(getenv("HOME"));
-  if(path.empty()) return TFilePath();
-  path += "/Documents";
-  QString pathAsQString = QString::fromStdString(path);
-  QDir dir(pathAsQString);
+  QDir dir = QDir::home();
+  dir.cd("Documents");
   if(!dir.exists()) return TFilePath();
```
manongjohn: If `dir.cd("Documents")` fails to change directory, I think dir.exists() will still be true because it's still in a valid directory.
I think this can be just written as follows:

```
QDir dir = QDir::home();
if(!dir.cd("Documents")) return TFilePath();
``` 

Same for the one below.

martinvanzijl: @manongjohn Thanks, I've made the change as suggested.

manongjohn:
LGTM

PR #2761, opened 2019-09-08
Add English Fx Help files

manongjohn:
Thanks to the time and efforts of @PeterLemon, we now have English versions of the Fx Help files which he has graciously agreed to incorporate into our releases.

This PR does the following:
1. Moves the Japanese versions to `stuff\doc\日本語` which will be used when a user selects `日本語` as the language preference.
2. Puts the English version  in `stuff\doc` as the default
3. Stores the English LaTeX files used to generate the English PDFs in `sources\translation_docs\english` for future use and reference.
     NOTE: Japanese LaTex files also added for future use under `source\translation_docs\japanese` but the existing PDFs were not generated by them.


RodneyBaker:
Outstanding.
Thanks for following up on this one for us @manongjohn 
Very much appreciated.

gab3d:
@manongjohn 
Great, thanks! 

I will start by doing a revision of some of these English docs, as two days ago when I used the texts in *particlesFx.html* to complete some parameter descriptions of the Particles manual section, I noticed the English text was less than stellar in some parts.

@RodneyBaker thanks for testing too, and for that tip about special characters!

shun-iwasawa:
First of all, thank you @PeterLemon for translation and making LaTex version of the fx documents. It must not only be helpful for users but also be useful for future maintainers.
Also thank you @manongjohn to convert them to PDF and post this. 

Regarding adding the license and copyright name of @PeterLemon , IMO it should be rewrited, given the fact that that his translation seem to be more a derivative work based on the Japanese version, than a work done by him from scratch.

I think there may be two options:

1. Rewriting the license document as follows:
    - Change the name of document from `LICENSE_latex.txt` to `LICENSE_english_fx_doc.txt` or something appropriate so that users do not misunderstand that this document is the license of LaTex system itself.
    - Insert a line at the top of the document like `English translation and LaTex files of Fx documents`, in order to describe which part of the software is subject to this license.
    - Insert copyright holders of original version of the documents and then add Peter's name with his contribution like as follows:
    ````
    All contributions by DWANGO:
    Copyright (c) 2016 - 2019, DWANGO Co., Ltd.
    All rights reserved.
    
    Initial English translation and LaTex convertion:
    Copyright (c) 2019, Peter Lemon 
    All rights reserved.
    
    All other contributions:
    Copyright (c) 2016 - 2019, the respective contributors.
    All rights reserved.
      
    Each contributor holds copyright over their respective contributions.
    The project versioning (Git) records all such contribution source information. 
    ````

1. Remove the license document and write the Peter Lemon's name as the copywrite holder in the commit message of this PR, so that we can trace contributors via Git afterwards. In this way his work and himself will be regarded as one of the `All other contributions` and `the respective contributors` respectively, in OpenToonz license `LICENSE.txt` .

If @PeterLemon allows, IMO the latter way would be simpler than the former.

PeterLemon:
Hi shun-iwasawa,

Yes you are correct that my translation is a a derivative work based on the Japanese version.
My workflow was I 1st created the Japanese version of the documents in LaTeX with media in the "images" directory, then I translated them into English directly from those Japanese LaTeX files, creating the English LaTeX documents.
I then made the English HTML docs derived from the Japanese HTML docs with media in the "img" directory.

I give you full permission to do whatever you want regarding how you want to deal with the license, as I created these docs to help the OpenToonz community.

I hope this helps resolve the license problem =D

manongjohn:
@shun-iwasawa 

As suggested, I will I will perform #2, removing the license information, recommit everything with @PeterLemon 's tag (and name), and force update the PR's commit history.

Sorry for the trouble. Just wanted to make sure everyone was covered in an appropriate way.

manongjohn:
@shun-iwasawa  Ok. I corrected the commit and removed the license.  If the commit message is inadequate, please provide the verbage I should use and I'll redo it again.

Thanks.

shun-iwasawa:
Thank you @PeterLemon for kind consideration. Also thank you @manongjohn for quick response.
LGTM

RodneyBaker:
Adding a post mortem note as something to look at in the future.

Additional documentation in more than one language does add to the file size of the installation/download.  Currently this is just over 3MB for Japanese files and under 1MB for English files.  The difference in size in this case is likely due to increased compression of PDF files in the more current English documents.  As additional FX documents are added the size of install will correspondingly increase.

Investigate having the installer/updater give the user an option to install a language pack where they would select desired languages and the installer would fetch those in the installation process.  If no language pack content was selected the user is opting for no FX documentation thus ensuring minimal file size of the installation with regard to FX documentation.

Current nightly installer (Windows) as of this post is 40.156MB.
Unpacked/installed this creates a Opentoonz stuff directory of 22MB and Program Files/Opentoonz size of 107MB.

manongjohn:
I agree that we shouldn't include every available translation of Fx Help files with the core release.  The only ones that should really be there are the English ones because they are the default.  I would go so far as to even remove the Japanese ones from the core release and make them available during installation and/or some other means.

In my opinion, additional languages should be a part of the installer as optional add-on packages.  The user can choose which language(s) they want to install. Then it's just a matter of the installer unzipping the files to the appropriate directory. 

If there was a common site for shared OT content like language packs, additional Fx macros, custom brushes, etc, a link could added to the Help menu as "Additional resources..." or something like that.

gab3d:
I agree that adding several additional languages to those help files would increase unnecessarily the package size. 
I'm currently halfway into an in-depth revision of the English files, but after that I will produce Spanish files too. Eventually, I will add all of these effects to their respective Manual section too.

So personally, the ideal way to handle every language without a file size penalty at all, would be that those Help buttons in the FX Settings panel would link directly to the appropriate section for that effect in the online Manual. That way every language could be there with zero cost.

hayoko:
Not every software does it that way though, most bundle it up into one package, like Krita, Affinity Serif and such since on workstations you may get people comfortable using software in different languages. It's only 1mb I wouldn't worry about it, Krita's locale is nearly 40mb and I'm still not bothered.

gab3d:
@shun-iwasawa @manongjohn @RodneyBaker whomever...

Here are the English help docs for the effects, fully revised/improved. (pdf and tex version, as well as html for those that were originally in that format).
If anyone could put them into place it'd be great.
👍 

[english docs.zip](https://github.com/opentoonz/opentoonz/files/3632292/english.docs.zip)
[english docs (tex).zip](https://github.com/opentoonz/opentoonz/files/3632293/english.docs.tex.zip)

RodneyBaker:
Thanks @gab3d 
I did a quick combine of those into one PDF file and posted it online.
The document should be able to receive comments etc if anyone feels the need.

[LINK](https://documentcloud.adobe.com/link/review?uri=urn%3Aaaid%3Ascds%3AUS%3Af67937f3-6e2d-4959-ab6c-f514984f7fbd)

gab3d:
Just a reminder.
It'd be good to integrate these into OT before 1.4 gets released.

gab3d:
Just in case...
Another remainder that these new revised docs haven't been merged yet and we're about to hit 1.4 milestone.

Thanks!

RodneyBaker:
@gab3d It might be good to open a new Issue to put a specific spotlight on what needs to be changed/updated here as a cursory view of this PR just indicates that the help files were merged.

cc:  @shun-iwasawa

gab3d:
OK, if that makes sense, I'll do just that.
But in essence I've revised all of the FX Help files for improving the meaning/wording/layout of the previous English translation, that was many times confusing.

As an example, let's take the **PerspectiveDistortIwa** help, Overview section:
**Previous:**

> This effect is for generating the movement of the ground at the time of Follow (Front Moves Fast, Back Moves Slow = Motion Parallax).
Since the deformed trapezoid is stretching the picture to the left & right, there is no vertical distortion.
In other words, the contraction in the vertical direction by the depth must be incorporated in the pre-input image.
Also, if you want to apply this effect to the background to parallel move, the movement of the background must be folded into the SubXsheet.

**Revised:**

> Allows to generate a motion parallax movement effect on the Source image, where things in
front move faster than things in the back.
The image width is stretched to the sides, to transform it into a trapezoid, but it will not
be stretched vertically.
In other words, the contraction in the vertical direction caused by depth, should be
incorporated beforehand in the image.
Also, if you want to apply this effect to a moving background, the movement of the background
must be collapsed into a SubXsheet.


Or the **CrossDissolveIno** effect help, **Clipping Mask** option description (also present in a lot of other effects):
**Previous:**

> When ON,
Material will not exist in the location (Back) (Alpha Value Zero), and remains clear.
The default setting is OFF.

**Revised:**
> When ON, it will take the Back port alpha information to define the alpha of the
resulting image, leaving areas where the Alpha value is zero as transparent.
The default value is ON.

This kind of things, more or less, is what you can find in a lot of the descriptions in these files.

gab3d:
ok, I've reuploaded them in #3015

PR #2762, opened 2019-09-09
Add Ability to Execute 'Merge Levels' Command Without Grouping Strokes

melieconiek:
Adds the 'Merge Levels Without Groups' command.

The new command works like the existing 'Merge Levels' command but will not group the resulting merge by level. Useful if you want to edit/fill the resulting drawing together. Like this:
a

The existing 'Merge Levels' command is better if you want to preserve the stack order of the merged levels and their fills (say if you were flattening a cutout animation into a frame-by-frame animation.) Llike this:
b

_I've never created a pull request before, so do let me know if I've made any mistakes._


toonz/sources/toonz/mainwindow.cpp
```diff
@@ -1739,6 +1739,9 @@ void MainWindow::defineActions() {
   QAction *mergeLevelsAction =
       createMenuXsheetAction(MI_MergeColumns, tr("&Merge Levels"), "");
   mergeLevelsAction->setIcon(QIcon(":Resources/merge.svg"));
+  QAction *mergeLevelsWOGroupsAction =
+      createMenuXsheetAction(MI_MergeColumnsWOGroups, tr("&Merge Levels Without Groups"), "");
+  mergeLevelsWOGroupsAction->setIcon(QIcon(":Resources/merge.svg"));
```
manongjohn: You are using the same icon that the original Merge Levels command uses.

This could be confusing when added to tool bars as they won't know which one it is at a glance. I'd keep the new command's icon blank for now.

melieconiek: I agree. I'll remove the icon.

melieconiek:
I'm not totally confident in the name for the command, either. I want what they both do to be as clear as possible. I'm a bit reluctant to rename the original command.

beeheemooth:
Wooohooo !!! Finally! I have been waiting for this for two and a half years.(This was my first request #1036) Thank you very much! @melieconiek

doko-desuka:
If a new icon is needed:  
https://github.com/doko-desuka/temp/blob/master/merge_wogroups.svg

RodneyBaker:
At this point it looks like the icon isn't needed.
Any further discussion before this PR is merged?

manongjohn:
Are we ok with the name? Is it clear enough that people will understand what it is? I don't think there is another that wont be wordy, but does anyone have other suggestions to consider?

doko-desuka:
> I don't think there is another that wont be wordy, but does anyone have other suggestions to consider?

At the cost of having to do some rewrites in code, a suggestion is to not make it a new command, but a setting onto the existing command. This way no new name is needed.    

Merging levels is an operation, whether to place the drawings of each level in groups is a parameter of that operation.

melieconiek:
@doko-desuka I've had instances where I wanted to do both these operations in the same project. You could, of course, go into the options and change it as needed. But in that sense, I thought it was better that they were two, distinct operations. This would be an issue for keybinding the commands too. Thanks for the icon, too--I'll try slotting it in tomorrow, see what it looks like. :)

I'm more concerned that it's clear what the new command does differently. @manongjohn I'd considered renaming the old command, but I don't know that that's a good idea `Merge Levels into Groups` & `Merge Levels without Groups`.

doko-desuka:
@melieconiek what I've seen other programs do in these cases is showing a popup dialog. Something like a 3-button dialog, with title "Merge Levels" and buttons "Grouped", "Ungrouped", "Cancel".

There's the MsgBox function for this, from DVDialog.h, you get an `int` as the choice:
- (Header) https://github.com/opentoonz/opentoonz/blob/7a6807b001eb0d25b3924045a2b8acdf27eb8ae3/toonz/sources/include/toonzqt/dvdialog.h#L60-L66
- (Implementation) https://github.com/opentoonz/opentoonz/blob/7a6807b001eb0d25b3924045a2b8acdf27eb8ae3/toonz/sources/toonzqt/dvdialog.cpp#L1095-L1143
- (Usage) https://github.com/opentoonz/opentoonz/blob/25d3db05fd57aed51d666455ac8dbc44067a4cea/toonz/sources/toonz/levelcreatepopup.cpp#L498-L499

manongjohn:
I'd be ok with either option, but I probably would lean towards the dialog box myself.

melieconiek:
I do think the dialog box would make the most sense if we wanted the commands merged--and as long as I could confirm it quickly it would be alright. Anything that slows us down is a concern.

I'd prefer to keep them separate, though, so they can both be triggered quickly--and if we'd be ok with renaming both commands (kinda like splitting them) to `Merge Levels Into Groups` & `Merge Levels Without Groups` I think that would make it much less confusing.

But I'll have to go with whatever the consensus is here.

manongjohn:
Adding the dialog box inherently changes the behavior of the command. Might be something @shun-iwasawa needs to comment on?

shun-iwasawa:
Thank you @melieconiek for tackling on this.
I’m not sure if users would like to choose both options as separate commands or they tend to use mostly one of them depending on the workflow, but I think there could be an “intermediate” option as follows:
Open popup on merging vector levels, asking whether to group or not. And the popup has another checkbox with a label “always choose this option in the current scene” . Activating this checkbox will make the command to act with the previously selected option without asking. Switching the scene will reset the preference and the popup opens again.

manongjohn:
The "always use the option" negates the abilty to use the other option when needed though. That would mean you could not use the other option on the scene without going to preferences (where it needs to be added) to turn it back on.

beeheemooth:
I want to emphasize that the problem of merging several levels at the moment is not the grouping itself, but the fact that this grouping occurs hierarchically. See picture below. The movement of elements after the merger of several levels becomes impossible due to the fact that grouping occurs hierarchically, that is, after each new merger, groups are formed that include the result of previous mergers.
group
@melieconiek I hope that your function when merging levels does not destroy the groups available in the drawings, but simply transfers them to the main level.

melieconiek:
If the goal is to make this as simple and easy to understand as possible, I don't think adding dialog boxes and additional options is the right way to go. As long as the commands have clear, unambiguous names, it should be fine.

@beeheemooth Groups from the source levels are preserved. :) The hierarchical issue with the existing command happens when more than two levels are merged--what you described, each level getting their own group, would make a lot more sense if you wanted the resulting merge grouped by level--iIt would be a lot easier to edit. I'd consider that a separate issue for now--but definitely worth looking into.

melieconiek:
Tried renaming the cmds and using @doko-desuka's icon. What do you think of this?
image

manongjohn:
How about just this:

Merge Levels (Grouped)
Merge Levels (Ungrouped)

melieconiek:
So it's more clear *what's* being grouped and there's no actual desolving of groups going on, how about:
Merge Levels (Grouped by Level)
Merge Levels (No Groups)

manongjohn:
Seems a little redundant to say Group by Levels. No Groups sounds better, though

melieconiek:
Alright, I give that a try a bit later.

gab3d:
Merge Levels (Grouped)
Merge Levels (Ungrouped)

I think this names are both: clear and short, so I like them.

melieconiek:
I changed the command names to `Merge Levels (Grouped)` and `Merge Levels (No Groups)`. I also dropped in @doko-desuka's icon. I'm not sure if there's something I need to do with crediting that--let me know.

image

shun-iwasawa:
I'm little concerned about that adding `(Grouped)` to the original command may confuse some users since the conventional command `Merge Levels` can also be used for merging raster levels and `(Grouped)` suffix is only meaningful with vector levels.

melieconiek:
@shun-iwasawa Understood, I had not thought of that. Do we think it would work if we just named the new command `Merge Levels (No Groups)` and kept the old command as is? Do we need to come up with another solution?

shun-iwasawa:
@ melieconiek

> If the goal is to make this as simple and easy to understand as possible, I don't think adding dialog boxes and additional options is the right way to go.

I agree that it would make the software useful to make frequently used command available by just pressing shortcut key, without opening popup.
I'm quite unfamiliar with vector based workflow, but I have roughly expected this command is not called so often (say, called once a 5 minutes at a maximum) because it will basically be used to pair of drawings, not to strokes. So, on that condition, IMO separating the behaviors with popup would be reasonable. 
If my understanding is mistaken, please let me know.

shun-iwasawa:
>  Do we think it would work if we just named the new command Merge Levels (No Groups) and kept the old command as is? 

Yes, this is fine in case you would like to separate the commands, as the conventional command will be kept as-is.

melieconiek:
> So, on that condition, IMO separating the behaviors with popup would be reasonable.
If my understanding is mistaken, please let me know.

@shun-iwasawa No, you are right, it is not unreasonable. We could add a dialog box that shows up only when merging vector levels, asking the user if they would like the resulting merge grouped by level. After thinking about it, I think you are right, this is probably the best way to go.

Can we consider adding a keybindable command for bypassing the dialog box, but not adding it to the default menu? Are there other commands like this? Is this best avoided?

Sorry for the back and forth, I just want this to be done correctly. Thanks for the patience.

melieconiek:
I will look into the dialog box--I was not thinking about the command being used for merging raster levels. The use of a dialog box is consistent with other commands and makes more sense given it's multiple uses. I'll set it up to only show if the levels being merged are vector levels.

shun-iwasawa:
> Can we consider adding a keybindable command for bypassing the dialog box, but not adding it to the default menu? Are there other commands like this? Is this best avoided?

FYI, there are keybindable commands which let users to switch the tool with specific mode with one key. (Created [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/mainwindow.cpp#L2254-L2263) in the source).
These commands were demanded by ink&paint staffs as they switch the tools and modes very often and would like to save the effort of typing two keys; one for switching tools and the other is for switching modes.

melieconiek:
I've added a dialog box asking "Group strokes by vector levels?" when merging vector levels--no dialog box is presented if merging raster levels.

I've also added two commands that bypass the dialog box:`Merge Levels (Do Not group Vector Levels)` and `Merge Levels (Group Vector Levels)`. I did not add them to the default menubar--they're for anyone who might want to keybind them.

toonz/sources/toonz/menubar.cpp
```diff
@@ -1253,6 +1253,7 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   addMenuItem(xsheetMenu, MI_MergeCmapped);
   xsheetMenu->addSeparator();
   addMenuItem(xsheetMenu, MI_MergeColumns);
+  addMenuItem(xsheetMenu, MI_MergeColumnsWOGroups);
```
manongjohn: > I've also added two commands that bypass the dialog box:`Merge Levels (Do Not group Vector Levels)` and `Merge Levels (Group Vector Levels)`. I did not add them to the default menubar--they're for anyone who might want to keybind them.

This makes the command show on the menubar by default. Will definitely show on macOS version.
 Should this be removed?

manongjohn:
As I'm testing this out, I'm wondering if it makes sense to change the wording on these merge specific commands to `Merge Vector Levels` and make them only work on Vector levels.  The new commands affects grouping which we can't create on Raster levels (correct me if i am wrong). The existing `Merge Levels` commands can still work on either Vector or Raster as they currently are for backwards compatibility. 

If we change it just for Vectors, maybe then we can change labels to something like `Merge Vector Levels (Group Strokes)` and `Merge Vector Levels (Individual Strokes)`?

Aside: I really wish we could have 1 merge command that was smart enough to determine whether it was doing a Vector, Toonz Raster or Raster level merge and act accordingly.  All these extra merge commands seem rather silly.

melieconiek:
Thanks for taking a look at this, @manongjohn . Yeah, I'll double check the menubar, I probably forgot to remove it.

I can use the same check for the dialog box on the `Merge Levels` command to prevent the vector level commands from being used on a raster level. I think that makes sense.

I agree about the merge commands. The `Apply Match Lines` command is different, but otherwise, all three level types should probably be merged into a single command. That'd have to be discussed, though since it we'd probably want to change the toonz raster level command to consistent with the other two. (The Toonz raster level merge command currently merges into a new file instead of merging into the bottom most level in the selection.)

There are a few other issues with the how the merge commands work, but I think that's outside the scope of this PR. It'd be a lot of work to fix.

melieconiek:
I removed the dialog bypass command entirely. They can be added in another PR later.

Should be functional. Please take a look. :)

RodneyBaker:
That's the stuff.  I like the simplicity in that.
Much appreciated!

There is certainly room for improvement via other PRs in the future for Merge but this resolves the main issue with Groups.

shun-iwasawa:
Jenkins

RodneyBaker:
@manongjohn Can you run Jenkins again on this PR.
Travis and Appveyor checks did not pass and yet no conflicts are identified so perhaps it was a temporary glitch.

shun-iwasawa:
Jenkins

RodneyBaker:
@shun-iwasawa That looks to have got the final issue resolved.
Thanks!

PR #2763, opened 2019-09-10
Support for MyPaint brush format version 2

manongjohn:
The current implementation of MyPaint brushes in OpenToonz expects the brush files (.myb) to be in Version 3 format. If a user tries to install Version 2 formatted brushes in OT or even in MyPaint directly (which OT can see), OpenToonz will load them but the brushes won't work.

To support Version 2 formatted brushes, I added a rather down and dirty method to detect and convert the contents of the file to Version 3 format, before sending it to the MyPaint brush engine.

### DIFF
--- a/toonz/sources/toonzlib/mypaintbrushstyle.cpp
+++ b/toonz/sources/toonzlib/mypaintbrushstyle.cpp
@@ -13,13 +13,15 @@
 
 #include "toonz/mypaintbrushstyle.h"
 
+#include <QDebug>
+
+#include <sstream>
 
 //*************************************************************************************
 //    TMyPaintBrushStyle  implementation
 //*************************************************************************************
 
-TMyPaintBrushStyle::TMyPaintBrushStyle()
-  { }
+TMyPaintBrushStyle::TMyPaintBrushStyle() {}
 
 //-----------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Is there a way to ignore types other than those you've expressly allowed?
For instance I think v4 mypaint brushes will crash Opentoonz if encountered.

manongjohn:
I have not seen V4 brushes, only V2 and V3. Can you attach one so I can see how it is formatted?

RodneyBaker:
I edited my previous post to add the link to the brush I was testing.

manongjohn:
The brushes from the link you posted are actually V3 formatted. I don't do anything with V3 formats and allow the mypaint engine to take care of being able to handle the elements or not.

I tried to search for V4 - V6 formats and have not found any that weren't in V3 format (it says "version 3" in the metadata).  I thought I ran across 1 example of a format that looked different from V3, but cannot seem to find it anymore. 
 
I just retriggered the Release to rebuild. It should rebuild with the updated libmypaint DLL from my other PR since it was merged.  Might fix any crash you were experiencing.

RodneyBaker:
Wow.  Those Dieterle brushes not only don't crash... they now work!  
I'll give this a little while for others to test and provide feedback but from my testing this PR is ready to be merged.

Somewhat related:
I do have one concern relating to Mypaint brushes and that is the long pause after initially clicking on the raster tab.  I've gotten use to holding my breath as that is where a crash will generally occur (if at all).  Obviously it takes some time to populate the Style Editor with all the mypaint icons so the assumption is that is the majority of delay.  I'm not sure what an be done to lessen that delay but any tests in that specific area would be appreciated.  While not preferable a message letting the user know that mypaint brushes are being loaded would at least let the user know that the delay isn't accidental.  Hopefully that makes sense.  Perhaps others don't see that same delay?

RodneyBaker:
I'm not sure what @blackwarthog's schedule is like these days so review by him make takes some time.
Open call for testing and feedback and that shouldn't take long as very little has changed:

Code/file changed:  [LINK](https://github.com/opentoonz/opentoonz/pull/2763/files)
Appveyor release:  [LINK](https://ci.appveyor.com/api/buildjobs/trnqw954vv0u6x7w/artifacts/toonz%2Fx64%2FOpenToonz.zip)

RodneyBaker:
There is a mypaint brush pack that works with Opentoonz but that dropped out of circulation (bad links via the mypaint brush pack site).  It's still available via the wayback machine for anyone that wishes to use them.  The brushes are called Tone v2.

Link:  https://web.archive.org/web/20130920094325/http://download.gna.org/mypaint/brushpacks/tone_v2.zip

RodneyBaker:
Yet another dead link via the mypaint brush page yieids 280+ more brushes from the very early days of MyPaint.  Also recovered via wayback machine:

https://web.archive.org/web/20140326182322/http://download.gna.org/mypaint/brushpacks/brushes-08.zip

RodneyBaker:
The Dirty Brush set works just fine:

http://goo.gl/tMfpZT

RodneyBaker:
After testing 500+ brushes multiple times without a single crash...
Merging.

PR #2744, opened 2019-08-30
Fix Flipbook Histogram Problems

shun-iwasawa:
This PR fixes several problems related to the Histogram and color picking feature in Flipbook / Color Model.

- Fixed Histogram to display correct values when applying color collection with 3DLUT.
- Fixed problem that the color picking fails when viewing Toonz Raster levels.
- Fixed the Color Model offsets the picking position for a half pixel.

PR #2766, opened 2019-09-10
Fix picking color of the snapshot image on Flipbook

shun-iwasawa:
This PR fixes the following problem with Flipbook, which appeared after introducing #2744 :
- When applying color collection with 3DLUT, OT fails to pick the color on the snapshot image. Currently it always picks the color on the current image.


RodneyBaker:
LGTM

PR #2767, opened 2019-09-10
Ensure dragged curve handles stay inside graph (#2480)

### LINKED ISSUE #2480: It's possible to lose the Curves node's handles
Gif example:
curves issue

As you can see, it is possible to move the Curves node handles into an area that is unclickable; this results in the user being unable to retrieve the handle.

A suggestion for fixing this issue would be to add cursor-key controls to selected handles and holding shift would increase the distance at which the handles move.
Alternatively, it would make sense to lock the handles within the clickable area, though this comparatively limits their movement range.

martinvanzijl:
Fixes #2480. When the curve handles are dragged, they will now stay within the bounds of the graph.

Demo below:

curve-handle-demo

### DIFF
--- a/toonz/sources/toonzqt/tonecurvefield.cpp
+++ b/toonz/sources/toonzqt/tonecurvefield.cpp
@@ -636,7 +636,7 @@ void ChennelCurveEditor::paintEvent(QPaintEvent *e) {
 void ChennelCurveEditor::mouseMoveEvent(QMouseEvent *e) {
   if (m_mouseButton == Qt::LeftButton && m_currentControlPointIndex != -1) {
     QPoint pos   = e->pos();
-    QPointF posF = QPointF(pos.x(), pos.y());
+    QPointF posF = checkPoint(QPointF(pos.x(), pos.y()));
     moveCurrentControlPoint(posF - m_points.at(m_currentControlPointIndex));
   }
 }


manongjohn:
The change for this was very simple and effective.

My only gripe (small) about not being able to move the handle beyond the box is that you need extra points and some trial and error to get a curve (right) to look similar to what it looked like before (left).

image

The other option would be to allow the handle beyond the box, but when let go, move the handle to the inside of the box without altering the curve. Likely a bit of work to get that right.

In general, this can impact the ease in creating intense arcs. If there isn't a huge concern about it, this can be merged.

RodneyBaker:
I had noticed the problem @manongjohn mentions on my first look and it concerned me too.
As this PR does restrict the ability to overshoot the values of the curves it will be good to allow others to weigh in on the change.  

Is there any reasonable way we could add a checkbox right inside the windows along with the Linear checkbox to give users the option to constrain the curve handles or allow them to move freely?

martinvanzijl:
Yes, I think I can add a checkbox for that.

I'm also working on an alternative solution where the mouse movement is not limited, but you can use the arrow keys to select a point and move it back onto the graph area. I'll try to post a demo later today.

manongjohn:
Something to keep in mind, to get the handle back in the current release, you need only to select the initial point and then DEL to delete it. It resets the start or end point instead of deleting.

martinvanzijl:
@manongjohn Thanks for the tip. It looks like that works for the first and last point, but any points you add in the middle with Ctrl+Click will be deleted entirely?

Anyway, here is the arrow key idea I mentioned: #2799.

I'm happy for this PR to be closed, since it would sacrifice some functionality.

RodneyBaker:
Closing this in anticipation of merging the alternative (#2799).

PR #2770, opened 2019-09-14
Remember last brush preset used

### LINKED ISSUE #2757: Brush Tool - Last used settings not saved on preset selection.
### Issue Summary
The last used brush settings are not saved when updated by selecting a brush preset--only when the brush setting controls are manipulated. Affects all three level types.

### Steps to Reproduce
Manipulate one of the controls on the brush tool's option bar. Select a brush preset from the dropdown on the tool option bar (something that will load different settings.) Restart OpenToonz.

### Expected Results
The brush tool should retain the brush settings that were loaded from the preset.

### Actual Results
The brush tool settings revert to the last time the brush settings controls were manipulated on the tool option bar.

### System Information
* **OpenToonz Version**: Latest Nightly 
* **Operating System**: Windows

manongjohn:
This PR fixes #2757 

Modified the logic to save the preset value for later use.   When loading the brush tool, if the preset is valued, it will load the settings for it instead of the `<custom>` settings stored in environment variables.

Also made it so when switching to a preset, the `<custom>` settings are not changed so if the user switches back to `<custom>`, the prior values are restored.  The settings are overridden when changing a setting on a preset, making the new settings the `<custom>` settings.

Also fixed an issue when changing from `<custom>` to a preset.  Although the preset values are loaded which switching, the `Preset:` dropdown reverts back to `<custom>`.  This typically happens only when the custom and preset values for a single control sliders differ, like Accuracy on the Vector brush options.

### DIFF
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -578,6 +578,8 @@ void FullColorBrushTool::setWorkAndBackupImages() {
 //------------------------------------------------------------------
 
 bool FullColorBrushTool::onPropertyChanged(std::string propertyName) {
+  if (m_propertyUpdating) return true;
+
   updateCurrentStyle();
 
   if (propertyName == "Preset:") {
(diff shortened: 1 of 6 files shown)


melieconiek:
Very nice! This preserves the custom settings too, so an excellent solution, thank you. :)

RodneyBaker:
Confirm this fixes #2087 
And the new addition of maintaining custom settings addresses my earlier concern.

Aside:  In the future as a user led development perhaps we can collect some useful/optimized presets and have them preloaded with Opentoonz along with the default.  

Thanks @manongjohn !
Merging.

PR #2771, opened 2019-09-15
Fix default shortcut assignments

### LINKED ISSUE #2738: Shortcut Ctrl + F doubly assigned so won't work for either process
# Issue Summary
The Shortcut key Control + F appears to be assigned to both New FX and New Output commands at the same time preventing either option from executing.

### Steps to Reproduce
Enter FX Schematic and use Control + F shortcut
Note that nothing happens.

### Expected Results
Control + F shortcut should execute either the New FX or the New Output command
I am not sure which is preferred but New FX is likely used more often.

### Actual Results
Neither the new FX dialog popup nor a new Output node are created.

If either New FX or New Output is assigned another shortcut... such as Ctrl + Space + F... both shortcuts will work.

manongjohn:
This PR fixes #2738.

Updated the default shortcuts in code as follows:
   - New Output: Reassigned from `Ctrl+F` to `Alt+O` to fix duplication of shortcut issue.
   - Capture (Line Test code not used): Unassigned from `Space` which would conflict with Hand Tool if it was ever enabled.

Other changes in code were due to running beautification on it.

I additionally reviewed and updated `defopentoonz.ini`, used for reloading OT default shortcuts, to add missing commands and ensure the defaults match what is coded.

For the `otadobe.ini`, `otharmony.ini` and `otretas.ini` I added missing commands as unassigned and sorted the entries so all the .ini files were consistently formatted.


RodneyBaker:
LGTM
I take it the changing of things such as "%3A" to " :" (colon) are what is being updated via beautification.

Thanks @manongjohn.

manongjohn:
Beautification only works on stuff in source folder. I manually replaced the %xx stuff with the actual character after confirming it would not impact loading the file.

RodneyBaker:
Thanks for the info.  Much appreciated.
That was the only thing in the code that stood out to me as something worth investigating.

PR #2773, opened 2019-09-16
Fix pasting into Type Tool edit box

manongjohn:
This PR fixes the ability to paste text into the Type Tool edit box.

As the edit box is not a true textbox control, I added logic to detect the `Paste` command and use the text contents of the clipboard as the source of the text to render instead of what was key'd in. All `\n` characters are automatically converted to `\r` so hard returns are displayed correctly.

### DIFF
--- a/toonz/sources/tnztools/typetool.cpp
+++ b/toonz/sources/tnztools/typetool.cpp
@@ -35,6 +35,9 @@
 
 // For Qt translation support
 #include <QCoreApplication>
+#include <QClipboard>
+#include <QApplication>
+#include <QMimeData>
 
 //#include "tw/message.h"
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
LGTM / Works as advertised.  Thanks @manongjohn 
Open call for review and feedback.

Here's the Appveyor download for testing:  [LINK](https://ci.appveyor.com/api/buildjobs/kqt2gxee690wpct3/artifacts/toonz%2Fx64%2FOpenToonz.zip)

melieconiek:
Excellent work! Thank you. :D

I was experiencing a bug with clicking to place the cursor on multi-line text, but it turns out that's in the nightly as well--so everything seems to be working as advertised. :)

RodneyBaker:
Often it's the little things that count and this little PR update pluses up the little known 'Text on a Path' feature in Opentoonz also.

Text on a Path in Opentoonz

PR #2774, opened 2019-09-17
Make control points easier to click (#1722)

### LINKED ISSUE #1722: Control points hard to click on
There appears to be a problem clicking on control points and handles in vector drawings, where sometimes the point gets selected, and sometimes not. I noticed it in Linux, but it seems to exist in Windows too. Here's a video of my experience:

https://www.youtube.com/watch?v=dVpiMIPsRpI

On Google Groups I said:
"It's just me clicking over and over again on a bunch of control points and handles. In a couple cases, I admit I just mess up and click in the wrong place. And in the majority of other cases, there's no problem. But in a significant minority of cases, the problem occurs. And while my clicking may not be absolutely centered on the control points, the control points are always highlighted when I click on them, which to me suggests that clicking on them should work."

And ManongJohn said:
"I can replicate this in Windows as well. After some careful and controlled testing, it seems that after the white box expands as you hover over it, when you click it on it to grab it, the logic doesn't account for the expanded box's dimension.  If you manage to click in the larger box, but outside the original tiny box that is theoretically centered underneath, it will either grab a line (if you are close to one) or nothing and start a selection drag.  The closer you get to the center of the expanded box, the more likely you will grab the box and not the line."

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/53535185-control-points-hard-to-click-on?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

martinvanzijl:
Fixes #1722. When a control point is highlighted (expanded), clicking anywhere on it will select and drag it as expected.

Demo below:

control-point-demo

For this fix, I updated the calculation for `maxDist2` in `ControlPointEditorTool::leftButtonDown` to match `ControlPointEditorTool::drawControlPoint`.

NOTE: The second post on on the bug report (https://github.com/opentoonz/opentoonz/issues/1722#issuecomment-356170985) mentions a problem with selecting a vector by clicking on it's fill. Should that be logged as a separate issue?

### DIFF
--- a/toonz/sources/tnztools/controlpointeditortool.cpp
+++ b/toonz/sources/tnztools/controlpointeditortool.cpp
@@ -429,7 +429,8 @@ void ControlPointEditorTool::mouseMove(const TPointD &pos,
 void ControlPointEditorTool::leftButtonDown(const TPointD &pos,
                                             const TMouseEvent &e) {
   m_pos           = pos;
-  double maxDist  = 5 * getPixelSize();
+  double pix      = getPixelSize() * 2.0f;
+  double maxDist  = 5 * pix;
   double maxDist2 = maxDist * maxDist;
   double dist2    = 0;
   int pointIndex;


RodneyBaker:
> NOTE: The second post on on the bug report (#1722 (comment)) mentions a problem with selecting a vector by clicking on it's fill. Should that be logged as a separate issue?

Yes.  I believe so.
@artisteacher Can you open a separate issue for the fill/shape selection you mention in that post?  Then a separate PR can specifically address that problem.

artisteacher:
Sure - #2775

RodneyBaker:
This fix seems to fall somewhere between a bugfix and an enhancement.

> Enhancement / Speed-up : Fine to be merged.

For my part I'm not seeming much of a difference here on my Windows laptop so others feedback will be appreciated.

Open call for testing and feedback prior to merger.

To view the changed code:  [LINK](https://github.com/opentoonz/opentoonz/pull/2774/files)
To test out the appveyor build:  [Link to zipfile](https://ci.appveyor.com/api/buildjobs/id33nu248h8qb74m/artifacts/toonz%2Fx64%2FOpenToonz.zip)

RodneyBaker:
Something we should consider with changes such as these...

For values such as these that could be placed into Preferences where users could fine tune them but are not what is the best way to highlight that area of code so that it can be easily found for future reference.  

This doesn't necessarily apply just to this change but other similar changes as well.  It doesn't seem reasonable that every variable in the code should be user adjustable via Preferences.

Would adding a comment in the code at that location to indicate it is the place to adjust control point selection sensitivity suffice?

Thoughts?

manongjohn:
For testing this, it's easiest to see with a mouse and having the current release and this one open a the same time.  In the current release, if you click inside the box, closer to the edges of the box, the more likely you will hit the line or click out of the selection in the current release.  With this change, I have yet to miss the box, even when I'm exactly on the edge.

LGTM. Thanks for tackling this.

PR #2776, opened 2019-09-17
Cache folder replacement & Clear cache command

shun-iwasawa:
This PR will change the location of cache folder from user-specified path to standard location on the local filesystem, for sake of simplicity and consistency with other applications.
- The system path `TOONZCACHEROOT` will no longer be used.
- Instead, the cache folder will always be set as the location of `QStandardPaths::CacheLocation` . The location differs depending on operating systems like as follows:
    - Windows : `C:/Users/<USER>/AppData/Local/<APPNAME>/cache`
    - macOS : `~/Library/Caches/<APPNAME>`
    - Linux : `~/.cache/<APPNAME>`

On my Win10 environment `<APPNAME>` becomes `<OrganizationName>/<ApplicationName>` ~~like `C:\Users\shun-iwasawa\AppData\Local\OpenToonz\OpenToonz 1.3\cache` .~~

**UPDATE :**   In this PR I removed version number from the `<ApplicationName>` . So the cache location on Windows will be like; `C:\Users\shun-iwasawa\AppData\Local\OpenToonz\OpenToonz\cache` .

This PR also adds a new command : `Clear Cache Folder` for removing all unused data in the cache folder. Cache data is usually cleared on appropriate timing, but it may remain upon software crash.

I would like to hear any comments or suggestions, especially if there are any possible incoveniences as the change will somewhat limit the potential to set the location of data.

### DIFF
--- a/toonz/sources/toonz/menubar.cpp
+++ b/toonz/sources/toonz/menubar.cpp
@@ -1117,6 +1117,8 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   addMenuItem(fileMenu, MI_Preferences);
   addMenuItem(fileMenu, MI_ShortcutPopup);
   fileMenu->addSeparator();
+  addMenuItem(fileMenu, MI_ClearCacheFolder);
+  fileMenu->addSeparator();
   addMenuItem(fileMenu, MI_Quit);
 
   // Menu' EDIT
@@ -1433,7 +1435,6 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   addMenuItem(helpMenu, MI_OpenOnlineManual);
   helpMenu->addSeparator();
   addMenuItem(helpMenu, MI_About);
-  addMenuItem(helpMenu, MI_ClearCacheFolder);
 
 // addMenuItem(fileMenu, MI_TestAnimation);
 // fileMenu->addSeparator();
--- a/toonz/sources/toonz/menubar.cpp
+++ b/toonz/sources/toonz/menubar.cpp
@@ -117,7 +117,7 @@ void RoomTabWidget::mouseReleaseEvent(QMouseEvent *event) {
 
 //-----------------------------------------------------------------------------
 /*! Set a text field with focus in event position to edit tab name.
-*/
+ */
 void RoomTabWidget::mouseDoubleClickEvent(QMouseEvent *event) {
   if (m_isLocked) return;
   int index = tabAt(event->pos());
@@ -1433,6 +1433,7 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   addMenuItem(helpMenu, MI_OpenOnlineManual);
   helpMenu->addSeparator();
   addMenuItem(helpMenu, MI_About);
+  addMenuItem(helpMenu, MI_ClearCacheFolder);
 
 // addMenuItem(fileMenu, MI_TestAnimation);
 // fileMenu->addSeparator();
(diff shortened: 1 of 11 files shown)


RodneyBaker:
I like this change.  Among the other benefits such as standardization and optimization I appreciate how this removes the cache data from the stuff directory.  Although not much cache data accumulates there it's good to remove unnecessary files from that location  This might not be the case if cache data was intentionally stored in order to use it later.  This would be my only area of concern but in such cases it would likely be better to move the cache data to more permanent storage anyway.

This surely will speed up use of portable versions of Opentoonz as well as cache won't need to write to the USB, network drive or other location.

If there is any downside that comes to mind it might be the user now potentially not knowing where the cache file is located.  Up until now it's easy to know because it's the first directory seen every time we open the stuff directory.  This however is a trivial concern because most who would look for cache files will likely be able to track down where they are stored readily.  To mitigate this potential deficiency recommend a note be added to state the location of cache files in Opentoonz documentation.

gab3d:
@shun-iwasawa 

I like the new location of cache files, it certainly seems more like a standard cache folder.

that said, in general I see removing options as potentially limiting for use cases not foreseen at the removal decision time.

also, if that same change could be achieved without removing the possibility of manually redirecting cache location if needed, then I don't see the benefit of hardwiring it to a static location.
(most probably there are some issues, and if that's the case that's alright, but I don't know it now)

toonz/sources/toonz/menubar.cpp
```diff
@@ -1433,6 +1433,7 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   addMenuItem(helpMenu, MI_OpenOnlineManual);
   helpMenu->addSeparator();
   addMenuItem(helpMenu, MI_About);
+  addMenuItem(helpMenu, MI_ClearCacheFolder);
```
manongjohn: In `menubar_template.xml`, you have `Clear Cache Folder` above `About`, but reversed here.   I would put it above `About` and add a separator between them. Same in the template.

Seems like a rather odd placement for this command. I'd probably put this under the `File` menu myself.

shun-iwasawa: Thank you @manongjohn for reviewing this.
Sure, I'll move the command under the `File` menu.

manongjohn:
Regarding the Windows path `...\Local\OpenToonz\OpenToonz 1.3\cache`, can we get rid of the `OpenToonz 1.3` subfolder?  

When we upgrade to 1.4 or later, you will no longer be able to clear that version's cache and if you haven't cleared the cache for a long time, there could be quite a lot in it.

I know it's what Qt is returning for CacheLocation which is based on Organization name, Application name or both.  I see you are setting the Application name with this. Is it really necessary?

What do you think of some sort of completion dialog notifying that the clearing of the cache is done.

shun-iwasawa:
@manongjohn 
How about removing version number from the application name? 
As far as I think right this second there would be no bad side effect by altering it.
Then the cache folder for Windows will become like `C:\Users\shun-iwasawa\AppData\Local\OpenToonz\OpenToonz\cache` .

I think it would better to keep `<OrganizationName>/<ApplicationName>` structure considering that we may use the same location from another software in the future.

> What do you think of some sort of completion dialog notifying that the clearing of the cache is done.

Other commands in OT basically have no completion dialog. For now I think it is enough to notify users only when there is an error.

manongjohn:
Removing just the version number would be good enough. Just trying to stay away from version numbered folders.

I'm ok if you don't want to add a completion dialog. Was just a thought.

manongjohn:
Tested without issue on macOS (Mojave - virtual) and Windows10.  Unfortunately my virtual Linux build environment is messed up and needs to be recreated so I can't confirm the PR.  It can be fixed later if there is a problem.

Anyway. LGTM.  1 less thing in the stuff folder that doesn't need to be. Thank you.

PR #2779, opened 2019-09-18
Fix Fx Settings Geometry Restoration

shun-iwasawa:
- Fx settings fails to resize according to the current fx parameters when opening for the first time after launch.

To reproduce:
1. Open Schematic, create some fx (say, `Blur Fx`).
1. Double click the fx node. Fx settings opens.
1. Close the Fx settings.
1. Restart OpenToonz.
1. Create some fx which is different from the previous one (say, `Bokeh Iwa Fx`).
1. Double click the fx node. 
1. Fx settings opens with improper size. 

The problem is due to the restoration process of the panel geometry. I modified it to reproduce only the position for Fx Settings panel.

### DIFF
--- a/toonz/sources/toonz/pane.h
+++ b/toonz/sources/toonz/pane.h
@@ -232,7 +232,7 @@ public:
       return false;
   };
 
-  void restoreFloatingPanelState();
+  virtual void restoreFloatingPanelState();
 
 protected:
   void paintEvent(QPaintEvent *) override;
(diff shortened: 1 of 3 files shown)


artisteacher:
Thanks for the fix! On a related note, some of #1016 is still an issue.

shun-iwasawa:
I'll merge this by myself as it is quite obvious.

@artisteacher #1016 seems to be macOS specific. I'll try to investigate when I could find time at home.

PR #2780, opened 2019-09-18
Fix Animate Tool Availability

shun-iwasawa:
This PR fixes Animate Tool being incorrectly disabled with operations as follows:
1. Load or create a level in a column 1.
1. Hide the column 1.
1. Select Animate Tool.
1. In tool options, switch the current object to "Camera1" .
1. Move the cursor to the viewer and see if the tool is available.

I modified the animate tool's availability to be defined not by the _current column_'s state, but by the _current object_.

### DIFF
--- a/toonz/sources/tnztools/tool.cpp
+++ b/toonz/sources/tnztools/tool.cpp
@@ -603,7 +603,7 @@ void TTool::notifyImageChanged() {
 //-----------------------------------------------------------------------------
 
 /*! Notify change of image in \b fid: update icon and notify level change.
-*/
+ */
 void TTool::notifyImageChanged(const TFrameId &fid) {
   onImageChanged();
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
It took me a moment to parse the problem here as not the column but the camera/object.
Confirm the primary issue and will test this more deeply once the Camera column PR is in circulation.

I do see that all checks have passed successfully and yet there is a report of one conflcting file:  toonz/sources/tnztools/tool.cpp

shun-iwasawa:
@RodneyBaker 
Thank you for testing this PR.
Since the animation tool changes the transformation of the object, the validation of the tool should depend on the current object, not on the current column.
I believe that the latest master still has the same issue and this PR will resolve it.

Sorry but I'll merge this by myself to verify the fix in the studio. If there is any problem I'll tackle with it.

PR #2786, opened 2019-09-19
Add Swap Colors Button to Gradient Fxs

shun-iwasawa:
This PR will add a button to Fx settings which will swap gradient colors.

Related to the addition, this PR also includes the modifications as follows:
- Made the button labels ( `Copy RGB: ** > **` and `Swap ** and **` ) translatable.
- Made the button actions undo-able.

### DIFF
--- a/toonz/sources/include/toonzqt/paramfield.h
+++ b/toonz/sources/include/toonzqt/paramfield.h
@@ -53,7 +53,7 @@ class ColorField;
 class SpectrumField;
 class ToneCurveField;
 class CheckBox;
-}
+}  // namespace DVGui
 
 //=============================================================================
 /*! \brief ParamField.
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Works great.  Thanks also for the code commenting.

gab3d:
@shun-iwasawa 
>     * Made the button labels ( `Copy RGB: ** > **` and `Swap ** and **` ) translatable.

great!
so, just a question: Can FX settings UI be translatable?
...I was under the impression that it wasn't the case. 🤔

shun-iwasawa:
@gab3d 

> Can FX settings UI be translatable?

Not yet. They are currently set in `$StuffFolder/config/current.txt` .
I agree it would be better to be translatable with Qt Linguist like other UI texts (I don't want to make `current_es.txt` or `current_ja.txt` for that!).
Even if it becomes translatable, I think we'll need to prepare a way to obtain original English names because currently they are used in expression.

gab3d:
I understand, yes. It'd be the right way to do it.
Thanks!

PR #2789, opened 2019-09-20
Fix filmstrip disabled tools

### LINKED ISSUE #2783: Tools disabled if the camera column is selected
When users are selecting a frame in Level Strip, tools should be still enabled even if the camera column is selected in Xsheet.

manongjohn:
This PR fixes #2783.

Modified to
- Ignore Camera Column checks while in Level Strip.
- Ensure the Vector and Toonz Raster brushes are enabled while in Level Strip if the currentl column isn'ta valid column or a spline (if vector) to begin with.


shun-iwasawa:

PR #2790, opened 2019-09-20
Remove selection requirement to toggle Camera Column

### LINKED ISSUE #2785: Show Camera Column command cannot be triggered without column selection
Since the command only changes the preference and has no relation with the column selection, IMO it should be able to trigger with shortcut key regardless of the current selection type.

manongjohn:
This PR resolves #2785

Moved the logic for toggling the Camera Column from cell selection to xsheet command.

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -1758,7 +1758,7 @@ void MainWindow::defineActions() {
                          "Alt+L");
   createRightClickMenuAction(MI_ToggleXSheetToolbar,
                              tr("Toggle XSheet Toolbar"), "");
-  createMenuXsheetAction(MI_ToggleXsheetCameraColumn,
+  createRightClickMenuAction(MI_ToggleXsheetCameraColumn,
                          tr("Show/Hide Xsheet Camera Column"), "");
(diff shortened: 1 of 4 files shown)


toonz/sources/toonz/mainwindow.cpp
```diff
@@ -1758,8 +1758,8 @@ void MainWindow::defineActions() {
                          "Alt+L");
   createRightClickMenuAction(MI_ToggleXSheetToolbar,
                              tr("Toggle XSheet Toolbar"), "");
-  createRightClickMenuAction(MI_ToggleXsheetCameraColumn,
-                             tr("Show/Hide Xsheet Camera Column"), "");
+  createMenuXsheetAction(MI_ToggleXsheetCameraColumn,
+                         tr("Show/Hide Xsheet Camera Column"), "");
```
shun-iwasawa: I think this can still be `createRightClickMenuAction` as the command is not included in the `Xsheet` menu.

RodneyBaker:
All looks good here but I'm not sure if I'm missing something.

@manongjohn you marked @shun-iwasawa review suggestion as resolved and we can see that particular code was successfully reverted.

Are you aware of any changes still required to complete this PR?

PR #2792, opened 2019-09-20
Save Specified Frame Numbers of Color Model in Palette

shun-iwasawa:
When loading the color model, you can specify frames to load by typing frame numbers in the `Frames` field in the `Load Color Model` popup.
However, currently such frame numbers are not saved in the palette.
Therefore, when you are using this feature, if you would like to continue your ink&paint work from yesterday you need to load the same color model just for specifying the frames again.

This PR will resolve such trouble.
If user specifies the frames on loading the color model, it will be saved in `fids` attribute of `refImgPath` tag in the palette file (.tpl) so that it will be reproduced on loading.


shun-iwasawa:
Sorry but I'll merge this by myself to verify the feature in the studio.
If there is any problem I'll tackle with it right away.

PR #2799, opened 2019-09-21
Add ability to move curve points using arrow keys

martinvanzijl:
This is a possible fix for #2480 and alternative to #2767. Allows selecting/moving points in the graph editor using the arrow keys:

- Right arrow key selects next point.
- Left arrow key selects previous point.
- Up arrow key moves selected point up 10 units.
- Down arrow key moves selected point down 10 units.

Demo below:

arrow-key-move-demo

I'm happy for the PR to be closed if it does not seem that useful.

### DIFF
--- a/toonz/sources/toonzqt/tonecurvefield.cpp
+++ b/toonz/sources/toonzqt/tonecurvefield.cpp
@@ -460,6 +460,64 @@ void ChennelCurveEditor::removeCurrentControlPoint() {
 
 //-----------------------------------------------------------------------------
 
+void ChennelCurveEditor::selectNextControlPoint()
+{
+  int controlPointCount = (int)m_points.size();
+
+  if (controlPointCount == 0)
+    return;
+
+  int firstVisibleControlPoint = 3;
+  int lastVisibleControlPoint = m_points.size() - 4;
+
+  m_currentControlPointIndex++;
+  if (m_currentControlPointIndex < firstVisibleControlPoint || m_currentControlPointIndex > lastVisibleControlPoint)
+    m_currentControlPointIndex = firstVisibleControlPoint;
+
+  update();
+}
+
+//-----------------------------------------------------------------------------
+
+void ChennelCurveEditor::selectPreviousControlPoint()
+{
+  int controlPointCount = (int)m_points.size();
+
+  if (controlPointCount == 0)
+    return;
+
+  int firstVisibleControlPoint = 3;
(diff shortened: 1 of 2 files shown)


doko-desuka:
Cool idea. Suggestions: 
- If the curve handles can be moved horizontally (I don't remember if they can), to let the right and left keys move them sideways. Then listen to something like `if (e->modifiers() & Qt::ControlModifier)` to switch to the next/previous points.
- To let one-unit adjustments be done. Inkscape uses something like this: key press = 1 unit, key+shift press = 10 units (what you have now).

doko-desuka:
...or using the Qt::Key_Tab / Qt::Key_Tab+Qt::ShiftModifier keys for switching between adjacent points.

RodneyBaker:
Not sure of the status of this group of related PRs.
@martinvanzijl Perhaps we can get a consensus on what the preferred path forward should be.
All the various PRs work well as is but it's not clear 

@gab3d  You've been plussing up the ideas via thumbs up.  Perhaps you can shed some light on your preferences moving forward.

My initial reluctance concerned restricting the overshoot that could be created via the suboptimal implementation that started this set of PRs.  Other than that concern I'm fine with both the restriction within the graph and the addition of the use of arrow keys in this PR.  @doko-desuka 's suggested additions could be added via a subsequent PR.

My current thought is to start with this PR which at least allows for a quick recovery of the handles going out of range off the graph.  Then we can examine the other two PRs to determine if they are still necessary.

What say you all?

cc:  @manongjohn @shun-iwasawa

toonz/sources/toonzqt/tonecurvefield.cpp
```diff
@@ -460,6 +460,64 @@ void ChennelCurveEditor::removeCurrentControlPoint() {
 
 //-----------------------------------------------------------------------------
 
+void ChennelCurveEditor::selectNextControlPoint()
```
RodneyBaker: I believe 'Chennel' should likely be 'Channel' here and in other places in this code.
The spelling doesn't appear to effect any processing but I'll look into that too.

RodneyBaker:
I do see where the spelling of 'Chennel' is used in a few other places in the code also:

RodneyBaker:
As the odd spelling is a carry over from prior code will press forward with this PR and anticipate a spellcheck scrub some time in the future.

This PR:  LGTM

PR #2802, opened 2019-09-21
Fix Crash When Loading MacroFx Containing CurveFx

### LINKED ISSUE #2424: Failure to read pointParam when instancing certain Macro FX

shun-iwasawa:
This PR fixes #2424 

As @manongjohn pointed out (thanks!), the crash caused when loading `<params>` tag under `<super>` tag in the Macro Fx.
I found that the `<params>` tag under `<super>` tag seems to be redundant, since the parameters of macro fx are saved in each fx data under `<nodes>` tag and corrected upon loading.
So, I added a condition to prevent saving / loading `<params>` tag if it is under the `<super>` tag.

I've tested the fix and it seems to be fine as far as I could see.
I would like you all to double-check the fix in order to prevent unexpected troubles.

### DIFF
--- a/toonz/sources/common/tfx/tfx.cpp
+++ b/toonz/sources/common/tfx/tfx.cpp
@@ -703,11 +703,18 @@ void TFx::setNewIdentifier() { m_imp->m_id = ++m_imp->m_nextId; }
 void TFx::loadData(TIStream &is) {
   std::string tagName;
   VersionNumber tnzVersion = is.getVersion();
-
+  // Prevent to load "params" tag under "super" tag on saving macro fx.
+  // For now "params" tag is no longer saved under "super" tag.
+  // This is for keeping compatibility with older versions.
+  bool isInSuperTag = (is.getCurrentTagName() == "super");
   QList<int> groupIds;
   QList<std::wstring> groupNames;
   while (is.openChild(tagName)) {
     if (tagName == "params") {
+      if (isInSuperTag) {  // skip loading "params" tag under "super" tag
+        is.skipCurrentTag();
+        continue;
+      }
       while (!is.eos()) {
         std::string paramName;
         while (is.openChild(paramName)) {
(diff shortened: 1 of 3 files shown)


RodneyBaker:
In trying to test this I am reminded that the stuff directory should be prepopulated with an empty MacroFX directory... if for no other reason than the average user won't know where to place the macros and may very likely typo the creation of the directory leading to the macro not appearing in the menu.

manongjohn:
I carried out a number of loading/saving and even rendering tests, working with old macros and newly created macros that have Curve Fx in it.  All appears to work without any issues.

Beyond the Curve Fx, I am not sure what other Fx would attempt to create `<param>` entries under `<super>` to test.  For completeness, If there are any other Fx that creates param entries, we should probably test them to make sure there is no impact.

RodneyBaker:
It would appear that any additional case that we haven't ran into is independent of this issue although if there is a similar problem elsewhere in the code it should be found and resolved.

As this is a crash and fixing it should not change how those other cases would operate...

Merging.

PR #2806, opened 2019-09-24
Update macOS Info.plist to include missing strings

### LINKED ISSUE #288: Add Version to .app bundle for OS X
We manage our software using systems that rely on the CFBundleShortVersionString key in the Info.plist of an App bundle. OpenToonZ does not have any of the version values entered into the Info.plist, making it difficult for us to manage the application.

Please include the application version (e.g. 1.0) in the CFBundleShortVersionString key of the Info.plist inside the app bundle.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/34130640-add-version-to-app-bundle-for-os-x?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jpturcotte:
Closes #288.
Untested; might have unwanted effects.

### DIFF
--- a/toonz/cmake/BundleInfo.plist.in
+++ b/toonz/cmake/BundleInfo.plist.in
@@ -4,28 +4,26 @@
 <dict>
 	<key>CFBundleDevelopmentRegion</key>
 	<string>English</string>
+	<key>CFBundleDisplayName</key>
+	<string>OpenToonz</string>
 	<key>CFBundleExecutable</key>
 	<string>OpenToonz</string>
-	<key>CFBundleGetInfoString</key>
-	<string></string>
 	<key>CFBundleIconFile</key>
 	<string>OpenToonz.icns</string>
 	<key>CFBundleIdentifier</key>
 	<string>io.github.opentoonz.OpenToonz</string>
 	<key>CFBundleInfoDictionaryVersion</key>
 	<string>6.0</string>
-	<key>CFBundleLongVersionString</key>
-	<string></string>
 	<key>CFBundleName</key>
-	<string></string>
+	<string>OpenToonz</string>
 	<key>CFBundlePackageType</key>
 	<string>APPL</string>
 	<key>CFBundleShortVersionString</key>
-	<string></string>
+	<string>1.3.0</string>
 	<key>CFBundleSignature</key>
 	<string>????</string>
 	<key>CFBundleVersion</key>
-	<string></string>
+	<string>1.3</string>
 	<key>CSResourcesFileMapped</key>
 	<true/>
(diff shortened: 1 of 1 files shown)


manongjohn:
Shouldn't the CFBundleVersion = 1.3.0 and CFBundleShortVersionString = 1.3?

From my understanding CFBundleVersion typically contains build numbers, so it could look something like 1.3.0.{build}.  Then CFBundleShortVersionString would just be 1.3 or 1.3.0.

Based on this, I was actually debating updating the CMake file instructions to dynamically set CFBundleVersion, CFBundleShortVersionString at the end of the build so we wouldn't have to keep updating this file.  Just wasn't sure if it was really worth the effort and just updating this with each release was sufficient.

jpturcotte:
I triple-checked, but it is actually in the correct order: https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html.
Also, for CFBundleVersion: "While developing a new version of your app, you can include a suffix after the number that is being updated; for example 3.1.3a1. The character in the suffix represents the stage of development for the new version. For example, you can represent development, alpha, beta, and final candidate, by d, a, b, and fc. The final number in the suffix is the build version, which cannot be 0 and cannot exceed 255. When you release the new version of your app, remove the suffix."
I think this could be automated, but I do not know how.

manongjohn:
Ok. I somehow never ran across that document when trying to find out how to populate those fields. Other places seemed to explain it similarly but not quite and they provided examples that seemed opposite to what you were doing, hence my question. 

If we don't care to include some sort of build number in CFBundleVersion, then there is no point automating it and just updating this file should be enough.

jpturcotte:
Yes, it is quite confusing. I agree with what you said. The only use I could see is for nightlies. Of course, for everything else including RCs, it is not too burdensome to update these manually. Does anyone beside Iwasawa-san have a working macOS build environment?

artisteacher:
> Does anyone beside Iwasawa-san have a working macOS build environment?

I do. Do you want me to see how this builds with my setup?

jpturcotte:
Please do. Thank you very much!

artisteacher:
This builds just fine. I can see the version number in the "Get Info" window from Finder and also in crash reports.

jpturcotte:
Thanks a lot, @artisteacher! This is ready to merge.

PR #2808, opened 2019-09-26
Fix deleted Camera1 returning due to Camera Column refresh

### LINKED ISSUE #2784: Camera1 cannot be deleted when the camera column is visible
### Steps to Reproduce

1. Show camera column
1. In Schematic, create a new camera (Camera2)
1. Switch the active camera to Camera2
1. Delete Camera1
1. Create a new camera again

### Expected Results

A new camera should be Camera1

### Actual Results

Camera1 and Camera3 are unexpectedly created.

This happens only when the camera column is visible.

### System Information
* **OpenToonz Version**:  latest

manongjohn:
This PR fixes #2784

When deleting a camera, if it was the current camera showing in the xsheet, it forced it to Camera1 (index 0), which would make it come back later when checking camera states. Changed it to switch to the index of the active camera.

This does not handle the existing issue I mentioned regarding delete Camera1, closing and reopening OT and finding it has been recreated.  That will likely be in a separate PR.

### DIFF
--- a/toonz/sources/toonzlib/tstageobjectcmd.cpp
+++ b/toonz/sources/toonzlib/tstageobjectcmd.cpp
@@ -793,7 +793,8 @@ void removeStageObjectNode(const TStageObjectId &id, TXsheetHandle *xshHandle,
     return;
 
   if (id.isCamera() && xsh->getCameraColumnIndex() == id.getIndex())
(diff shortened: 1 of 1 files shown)


turtletooth:
LGTM.  Thanks @manongjohn

Issue #2775, opened 2019-09-17
Vector drawings are sometimes hard to select


artisteacher:
Vector shapes/drawings can be difficult to select. Currently a vector object can only be selected by clicking on the center of the stroke. Clicking on a fill doesn't select anything. Similarly, if the viewer is zoomed in too much, clicking on the edge of a stroke will not select it. Groups have the same problem.

It's pretty standard to be able to select closed shapes by clicking anywhere on the shape - whether it is the fill or the edge of an outline.

age3rcm:
adobe programs have developed neat ways to vector workflow, as someone who has been with these programs for 5 or so years, i feel that opentoonz is missing on a feature like this, to be able to select the vector from what you see as a stroke. maybe selecting the fill itself is too much for opentoonz way, but something as selecting vectors easier will improve vector workflow

artisteacher:
It is hard to say what would be the best way forward with this - especially since fillable vector regions can be created by any number of strokes (some of which may be part of other fillable areas).

I do think that at minimum, clicking anywhere on a stroke should select that stroke (no matter how close to the edge when zoomed in). Also clicking anywhere inside a group (whether it is stroke or fill) should select that group. If a group is entered with a double-click (or etc.), then the individual parts of the group can always be selected normally.

Otherwise, perhaps if a fillable region is selected, any boundaries to that region could also be selected when clicking on the fill. Or perhaps, transparent regions shouldn't be selected? In that case, if a region with a fill (other than transparent) is detected, then any boundaries to that region could also be selected when clicking on the fill. Though if clicking to select a stroke was more reliable, I probably wouldn't care about selecting through clicking on the fill...

artisteacher:
Selecting the stroke (but not the control points) can also be challenging with the control point editor if the viewer is zoomed in too much.

RodneyBaker:
Closing as fixed by #2809.

artisteacher:
It's so much better! Selecting strokes with the control point editor can still be difficult when zoomed in too much.

PR #2809, opened 2019-09-27
Improve vector stroke selection

manongjohn:
This PR addresses #2775 by allowing users to select a vector stroke anywhere the stroke is drawn, accounting for thickness, not just from a fixed distance from the center line.

The fixed distance from center line is still used for 0 thickness lines or anything thinner than the fixed distance.

Caveat: Picking the right stroke where multiple strokes overlap may require moving closer to the center line or a clear section of the stroke you want.

### DIFF
--- a/toonz/sources/tnztools/vectorselectiontool.cpp
+++ b/toonz/sources/tnztools/vectorselectiontool.cpp
@@ -94,13 +94,21 @@ FourPoints getFourPointsFromVectorImage(const TVectorImageP &img,
 //-----------------------------------------------------------------------------
 
 bool getStrokeIndexFromPos(UINT &index, const TVectorImageP &vi,
-                           const TPointD &pos, double pixelSize) {
+                           const TPointD &pos, double pixelSize, TAffine aff) {
   if (!vi) return false;
   double t, dist2 = 0;
-  double maxDist  = 5 * pixelSize;
-  double maxDist2 = maxDist * maxDist;
-  vi->getNearestStroke(pos, t, index, dist2);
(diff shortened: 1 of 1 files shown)


artisteacher:
This will be so helpful! Thank you.

RodneyBaker:
I've been staring at this one for a few days and I'm not seeing a whole lot of difference although I saw some.  Part of this might relate to the fact that I've never felt vector lines were particularly hard to select before.  Perhaps that's because I learned to select closer to the ends?  (unsure)

Is it the subtle difference that is the change here?

I've looked at the code repeatedly and the only thing that comes to mind is that perhaps the selection area hot zone could be increased by a factor of 2.  (again unsure)
I'll do some more side by side comparisons this and previous releases.

Aside:  In the original issue #2775 one aspect of that discussed the shape being selected by clicking on/selecting its fill.  I don't think this PR attempts to implement that portion and if not just want @artisteacher to know.

manongjohn:
Make sure your lines are very thick or zoom in to make the lines appear wider.   To truly see it, zoom in a lot or make the lines extremely thick, then try to grab the line closer to the edges.

artisteacher:
At least on a Mac, the difference is very obvious (assuming a thick or very zoomed in stroke). It's significantly easier to select a stroke this way. I don't particularly need to select by clicking on the fill as long as strokes are easy to select.

RodneyBaker:
Calling this one good.  The obvious areas for me is where the thinnest parts of the line are easily selected via this release.

LGTM

PR #2810, opened 2019-09-27
Fix Assertion Failure on Selecting Camera Column

shun-iwasawa:
This PR will resolve assertion failure occurs with the following steps in debug mode:

1. Select Animate tool.
1. Select camera column.

The assertion failure occurs at [line 212 of tstageobject.cpp](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonzlib/tstageobject.cpp#L212) ;
```
const TStageObjectId TStageObjectId::ColumnId(int index) {
  assert(0 <= index && index <= StageObjectMaxIndex);
  return TStageObjectId(COLUMN << StageObjectTypeShift | index);
}
```
There were cases that the function `TStageObjectId::ColumnId()` is called with the argument `index` being -1 when the camera column is selected.


RodneyBaker:
LGTM

PR #2811, opened 2019-09-27
Narrow Camera Column

shun-iwasawa:
This PR will narrow the camera column in Xsheet.

narrow_camera_column

Now the camera column has 22 pixels width.
There is no changes in the timeline view.


manongjohn:
Nicely done, looks great and quick too! :)

I reviewed the code and overall looks good to me.

The only problem with the Classic layout is that there is no ability to get to the `Config` popup which you normally get by long pressing the Preview area.  Without it, they cannot change to a different camera if there are others. For Classic layout, i would suggest making the exception and allow the Config button to display like you have in the Classic Revised layout.

shun-iwasawa:
Actually I made clicking the camera icon to open the camera configuration popup in the Classic layout.
I agree it's too unfriendly and it would better to display the config button. I'll work on it.

toonz/sources/toonz/xshcellviewer.cpp
```diff
@@ -3439,6 +3456,11 @@ void CellArea::createKeyLineMenu(QMenu &menu, int row, int col) {
   if (rType != TDoubleKeyframe::Constant)
     menu.addAction(cmdManager->getAction(MI_UseConstantInterpolation));
```
manongjohn: I know this is a little out of scope for this PR, but since you are updating this Context menu, can you move `Constant Interpolation` to the top of the list so it matches the list in the Function Editor?

After more consideration it I also recommend removing `Expression Interpolation` and `File Interpolation` since those options require special input.

RodneyBaker:
I concur with the removal of Expression and File Interpolation from that menu.
It is useful but, for those that need that functionality in that same space... and I occasionally am one...  all that need be done is open the Preferences window and have it hovering nearby to make that same on the fly assignment in real time.  Removing the extra option will lesson confusion.

manongjohn:
Right-click the `Lock` button in `Compact` mode doesn't bring up the Lock context menu. It is below it where the `Config` button is.  It's not adjusting for the lock button being shifted up.

manongjohn:
Also in Classic-Revised mode, right-click the Lock button brings up the Preview context menu and right-click the Config button brings up the Camstand context menu.

manongjohn:
Sorry, last one: In Classic mode, right-click the Camera thumbnail brings up the Preview context menu.

shun-iwasawa:
Modified the right click menu. Thank you @manongjohn for detailed testing!

manongjohn:
LGTM

PR #2814, opened 2019-09-29
Stop Width Change on Style Editor on Maximize

manongjohn:
Similar to the change to make the Film Strip stop changing it's width on maximizing or changing OT's main window size, I've set it so the Style Editor also stops changing it's width.

I chose Style Editor because on most drawing rooms, I have to keep shrinking that pane width to increase the drawing area after resizing OT's main window.


RodneyBaker:
Looks good.  Thanks @manongjohn.
That was a simple addition.  :)

PR #2816, opened 2019-10-03
New Fx: Corridor Gradient Fx Iwa

shun-iwasawa:
This PR will introduce a new Fx : `Corridor Gradient Fx Iwa` .

The original purpose of this fx is to easily generate a `Depth` reference image to be used in `Bokeh Ref Fx Iwa` for indoor scenes such as in a room or a corridor like as follows:
 
corridorgradient

Before this fx they needed to combine linear gradients for each side of walls, ceiling and floor to make the depth reference image. Instead, this fx uses the same gadget as the `Free Distort Fx` to specify the shape of far side of the wall and perspective of the corridor. 

This PR also modified the fx gadget for quad shape, which is used in this fx and the `Free Distort Fx`.
I'll write the details in the next comment.

### Parameters
* **Shape** : specifies the shape of the gradient. Either `Quadrangle` or `Circle` . 
* **Type** : specifies the interpolation type of the gradient.
* **Bottom Left, Bottom Right, Top Left, TopRight** : positions of vertices of the inner / outer shape.
* **Color** : colors for each end of the gradient.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1272,6 +1272,20 @@
   <item>"STD_iwa_TextFx.boxColor" "Box Color"</item>
   <item>"STD_iwa_TextFx.showBorder" "Show Border"</item>
 
+  <item>"STD_iwa_CorridorGradientFx" 			"Corridor Gradient Iwa" 	</item>
+  <item>"STD_iwa_CorridorGradientFx.shape"	"Shape"		</item>
+  <item>"STD_iwa_CorridorGradientFx.curveType"	"Type"		</item>
+  <item>"STD_iwa_CorridorGradientFx.bottom_left_in"	"Bottom Left  X"	</item>
+  <item>"STD_iwa_CorridorGradientFx.bottom_left_out"	"Bottom Left  X"		</item>
+  <item>"STD_iwa_CorridorGradientFx.bottom_right_in"	"Bottom Right  X"	</item>
+  <item>"STD_iwa_CorridorGradientFx.bottom_right_out"	"Bottom Right  X"		</item>
+  <item>"STD_iwa_CorridorGradientFx.top_right_in"	"Top Right  X"	</item>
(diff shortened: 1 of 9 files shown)


RodneyBaker:
I tested out various setups and am glad to report that everything worked well on Windows.
I see that Travis/OSX has some conflicts.

A few things here and there that didn't seem to work initially all have workarounds of one form or another.  The most serious of those might be how the source image may need to be either adjusted in advance or scaled up to allow for bleeding of canvas color into the result of the FX.  Short of having an option to extend pixels outside of the camera frame I'm not sure what to suggest there.

shun-iwasawa:
@RodneyBaker
Thank you for testing the fx. I'll fix the problem on macOS build.
Regarding the canvas frame spill problem, you can also exploit Tile Iwa Fx for expanding the source image to the outside of the camera frame. Please make sure to activate the `Mirror Horizontally` and the `Mirror Vertically` options.

tileiwa

shun-iwasawa:
Here are usages of the new quadrangular gadget:

### Corner handles
quad_corner
* **No modifiers** : scale the shape toward the opposite corner.
* **+Shift** : scale the shape keeping the aspect ratio.
* **+Alt** : scale the shape toward the median point of the shape.
* **+Ctrl** : move the point independently.

### Edge handles
quad_edge
* **No modifiers** : move the both ends of the edge.
* **+Shift** : translate the edge along with X or Y axis.
* **+Alt** : symmetrically move the opposite edge.

### Body
 quad_body
* **No modifiers** : translate the whole shape.
* **+Shift** : translate the shape along with X or Y axis.
* **+Ctrl** : rotate the whole shape around the median point.
* **+Ctrl+Shift** : rotate by 90 degrees.

RodneyBaker:
Thanks for the additional information Shun.  Very useful!
I note that TileIwa FX is needed in most cases such as this as the regular Tile FX will not do the job.

shun-iwasawa:
Thank you for testing. I'll merge this by myself.

PR #2821, opened 2019-10-06
Fix lipsync popup layout

manongjohn:
This PR does the following:

1) Fixes the Import Magpie File popup layout

Fields and labels had inconsistent heights causing things not to align properly. Made sure all label and field heights were consistent.

2) Relabeled "Import Magpie File" to "Import Toonz Lip Sync File"

MAGpie is no longer available, as far as I can tell. (If it is still available, please provide link).  I changed the label in order to support other caption software that might be able to export to OT's specifically formatted lip sync file (.TLS).

Here is a manually generated .TLS file I used to test (remove the .txt extension)

[test.tls.txt](https://github.com/opentoonz/opentoonz/files/3726739/test.tls.txt)

image

### DIFF
--- a/toonz/sources/toonz/filebrowserpopup.cpp
+++ b/toonz/sources/toonz/filebrowserpopup.cpp
@@ -2012,7 +2012,7 @@ void ReplaceParentDirectoryPopup::initFolder() {
 // ImportMagpieFilePopup
 
 ImportMagpieFilePopup::ImportMagpieFilePopup()
-    : FileBrowserPopup(tr("Import Magpie File")) {
+    : FileBrowserPopup(tr("Import Toonz Lip Sync File")) {
   setOkText(tr("Load"));
   addFilterType("tls");
 }
--- a/toonz/sources/toonz/magpiefileimportpopup.cpp
+++ b/toonz/sources/toonz/magpiefileimportpopup.cpp
@@ -57,7 +57,7 @@ MagpieFileImportPopup::MagpieFileImportPopup()
 
   beginVLayout();
 
-  setLabelWidth(45);
+  setLabelWidth(60);
 
   addSeparator(tr("Frame Range"));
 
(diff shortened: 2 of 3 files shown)


RodneyBaker:
Initial look and testing.  Looks good.
I'm glad to see this has received some attention as it was almost unusable in its earlier state.
If I find some time I'll try to put together a Preston Blair phoneme set to share here for others to use.

RodneyBaker:
If @turtletooth is available it would be good for him to look at this as he was involved in some of the prior coding for the lipsync feature.

turtletooth:
This is actually a different popup than I worked on.  I was going to use this, but it was so specific to Magpie that I decided to make a different approach.

RodneyBaker:
Thanks for the clarification Jeremy. 
Good to know!

manongjohn:
I think this was Magpie focused only because Magpie might have been the only free software available for making closed caption files? Just a guess.

At any rate, the file loaded in OT isn't even specific to magpie. Someone created a script that would convert magpie data to the Toonz Lip Sync format. In reality, that can be done for any closed caption software output.

The TLS file layout is actually pretty simple.

The nice thing about TLS format is you can define the phonemes you need, which might be different for other languages. I'm not sure how flexible the Apply Lip Sync Data popup is, though I do like that interface better.

At some point it might be better to merge that. Perhaps the one @turtletooth made should be adapted to read TLS files also.

RodneyBaker:
Pardon me for attempting to extend the functionality when I should just be testing what is there.
But I like what I see thus far.  There's lots of room for lip sync and beyond.
(More testing later tonight)

ToonzSyncFile

gab3d:
I think this dialog is more powerful than the old one, as it allows to see all the drawings at the same time.

imagen

manongjohn:
@gab3d I definitely agree with you. I like the new one much more which is why I do recommend that we remove the old one and make the newer one work with TLS files also.  The old one is a bit more dynamic, which I like, allowing from 0 - 9 phonemes in any order where as this one appears to be fixed at 10 specific preston-blair phonemes.   The old one also allows you to only use a section of the file which is nice too.

I think this PR can move forward for now until I, or if someone else, has time to re-configuring the new one.

RodneyBaker:
I'll take that as an okay to merge.  Thanks @manongjohn for bringing this feature back into play.

There is something to be said for both features and I hope we aren't too quick to deprecate either one.  In the interim I hope to use both.

manongjohn:
Aside: MAGpie was able to export to TLS format by putting the stuff\config\export-toonz.lua in a particular spot in the MAGpie installation.

Apparently Papagayo-ng says you can add more export formats by updating 2 python files, 1 for the option and the other to write it out.  I know it's kind of moot because OT can already load in the Papagayo-ng data files, but for S&Gs, I created a TOONZ export format for it. :) 

Replace the 2 files in C:\Program Files (x86)\Papagayo-NG\papagayo-ng with the contents of this zip file: [toonz-papagayo-ng.zip](https://github.com/opentoonz/opentoonz/files/3737579/toonz-papagayo-ng.zip)

This version will put the phrase words in frames of the note levels.  Add the Text Iwa FX and you can see the words as it is being spoken.

PR #2824, opened 2019-10-08
Nonlinear Slider for the Brush Size Options

shun-iwasawa:
This PR will change the behavior of the brush size sliders in the tool options bar, from linear to nonlinear so that it allows users to control precisely in smaller size and also to set larger size.
nonlinear_slider

The transition of the value was based on the brush size slider of PhotoShop. In the graph below the horizontal line shows the position of the handle in slider and the vertical one size of the brush within a range of 1 to 1000. The blue line shows the transition of conventional slider and the green one the transition of nonlinear slider.
nonlinear_slider_graph

Following parameters will be switched to the nonlinear slider:
- `Size` of the Brush tool (except when selecting MyPaint styles).
- `Size` of the Eraser tool.
- `Size` of the Paint Brush tool.

The maximum values of these parameters are expanded from 100 to 1,000.

-------
In association with the above, this PR also changes the size of the Brush tool when 
- drawing on Raster level,
- with `Pressure` option is active,
- w/o selecting MyPaint style, and 
- using mouse.

In the above condition, the "pseudo" pressure value was changed from 0.5 to 1.0, in order to gain consistency with the brush tool for other level types.

-------
**CAUTION!** 
Changing the brush size more than 100 in the newer version and then launching the older version of OT will cause crash due to the lack of error handling for out-of-range of parameter.

### DIFF
--- a/toonz/sources/include/toonzqt/doublefield.h
+++ b/toonz/sources/include/toonzqt/doublefield.h
@@ -102,11 +102,19 @@ onSliderChanged(int value),
 class DVAPI DoubleValueField : public QWidget {
   Q_OBJECT
 
+  bool m_isLinearSlider;
+
 protected:
   RollerField *m_roller;
   DoubleValueLineEdit *m_lineEdit;
   QSlider *m_slider;
   QWidget *m_spaceWidget;
+
+  void setLinearSlider(bool linear) { m_isLinearSlider = linear; }
+
+private:
+  double pos2value(int x) const;
+  int value2pos(double v) const;
 
 public:
   DoubleValueField(QWidget *parent, DoubleValueLineEdit *lineEdit);
@@ -158,7 +166,7 @@ signals:
   void valueChanged(bool);
 
   /*! This signal is emitted only when users edited the value by hand
-*/
+   */
   void valueEditedByHand();
 };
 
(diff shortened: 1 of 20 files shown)


RodneyBaker:
Thank you Shun.  This works great.
I haven't yet tested with a mouse but I don't expect that to be any different.
I did test with shortcut keys and such.  I don't think I'd every tried to draw live while using the shortcut keys before.  

Opentoonz is just a joy to draw in these days.
Little improvements like these go a long way.

beeheemooth:
@shun-iwasawa Does this mean that it will be necessary to make new presets for the brush instead of the old ones? 
Will this feature be implemented for other tools (for example, Iron Tool, Pinch Tool, Magnet Tool) that also need more fine tuning of small values?

shun-iwasawa:
@beeheemooth 

> Does this mean that it will be necessary to make new presets for the brush instead of the old ones?

You can still use the brush presets made in the older versions.

> Will this feature be implemented for other tools

Thank you for the suggestion. I additionally made the following parameters to use nonlinear sliders:
- `Size` of the Finger tool. 
The maximum value is expanded from 100 to 1,000.
- `Size` of the Pinch tool. 
The maximum value is unchanged from 1,000 as it seems to be sufficient.
- `Size` of the Magnet tool.
The maximum value is expanded from 100 to 500. Also modified the source removing the parameter conversion which seems to be unnecessary.

Please note that Iron tool has no parameters to implement.

beeheemooth:
Oops, I wrote "Iron Tool" by mistake, sorry.

shun-iwasawa:
Sorry but I'll merge this for testing the look and feel of the modification in the studio.
If there is any incovenience I'll address it right away.

PR #2829, opened 2019-10-10
Fix reading escaped double-quotes in quoted strings

### LINKED ISSUE #2827: Quotations in level name result in empty scene on open
### Issue Summary

If quotation marks are part of a level name, the entire scene is empty after re-opening OT.

Screen Shot 2019-10-08 at 3 52 39 PM
Screen Shot 2019-10-08 at 3 56 59 PM
Screen Shot 2019-10-08 at 4 02 01 PM
Screen Shot 2019-10-08 at 4 02 09 PM

### Steps to Reproduce

1. Load raster levels into a scene where the name includes " and "
2. Also load levels without any " in the name
3. Save all
4. Close and re-open OT
5. Load the scene
### Expected Results

The scene to load normally
### Actual Results

There is a warning notification that there is a problem with the scene and that some files may be missing. The entire Xsheet/timeline is empty - not even properly named levels or "missing levels" with red text. The levels names do still display in the Scene Cast but without any frame numbers. The levels do still display normally in the File Browser.

I know it's not a best practice to name using symbols & etc. but it's weird that the wrong name format impacted the display of the entire scene including properly named levels.
### System Information
* **OpenToonz Version**: 1.3 and latest nightly build
* **Operating System**: macOS 10.14.6
* **CPU**: 
* **Memory**: 
* **Graphics Card**: 
* **Graphics Tablet**:

manongjohn:
This PR fixes #2827

Normally OT stops you from loading a filename with double-quotes (") in it, but there appears to be ways around that (bug or not), that can allow it to happen.  Short of fixing all possible ways this can happen, this fix will ensure that if it does get into a filename, the scene will still load properly.

The issue: 
When OT reads in strings that are encapsulated in ", it reads character by character.  The 1st " is the start of the string but it assumed the next " is the end of the string and stopped reading the rest. It wasn't accounting for an escaped "  (\\") which is normally considered part of the string.

The fix:
Added logic to track the escape character (\\). If a \ appears immediately before a ", it will consider the " as part of the string and not the string terminator.

NOTE: It is not possible to create filenames with " in it on Windows. For testing purposes, I manually modified a scene file to have a quoted filename containing  \\" in it.  In existing code, OT fails to load the scene properly as explained. With this fix, the scene loads successfully.

### DIFF
--- a/toonz/sources/common/tstream/tstream.cpp
+++ b/toonz/sources/common/tstream/tstream.cpp
@@ -24,7 +24,8 @@ namespace {
 string escape(string v) {
   int i = 0;
   for (;;) {
-// Removing escaping of apostrophe from Windows and OSX as it's not needed and causes problems
+// Removing escaping of apostrophe from Windows and OSX as it's not needed and
+// causes problems
 #ifdef LINUX
     i = v.find_first_of("\\\'\"", i);
 #else
@@ -353,10 +354,11 @@ TOStream &TOStream::operator<<(string v) {
   }
   int i;
   for (i = 0; i < len; i++)
-    if ((!iswalnum(v[i]) && v[i] != '_' && v[i] != '%')
-		|| v[i] < 32   // Less than ASCII for SPACE
-		|| v[i] > 126  // Greater than ASCII for ~
-		) break;
+    if ((!iswalnum(v[i]) && v[i] != '_' && v[i] != '%') ||
+        v[i] < 32      // Less than ASCII for SPACE
+        || v[i] > 126  // Greater than ASCII for ~
+        )
+      break;
   if (i == len)
     os << v << " ";
   else {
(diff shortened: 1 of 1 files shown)


gab3d:
Wouldn't it be safer to just strip the double-quotes of any file name at save time?
I mean, if it is known that Windows don't support it, then allowing for file names with such characters in it would break portability of scene (and potentially other) files in a project... 🤔

manongjohn:
I don't want to change filenames if it isn't necessary and I don't think that is the best way to deal with this issue.

My fix is appropriate. Regardless of where it appears in the scene file, OT should be able to read strings with \\" without issues.   It doesn't matter how it got into the scene file.

Keep in mind that other software can create scene files, like Storyboard.  If it wrote out a filename with \\" the scene would not load into OT successfully without this change.

RodneyBaker:
I'm good with this proposed approach.
While safer to strip the characters out, aligning OT's capability to use some standard filenaming conventions that require the use of quotes (i.e. to allow for spaces in url names) is very appealing.  I would guess that perhaps that is how the quotes got into the field in the first place... via copy/pasting urls.  Unknown.

manongjohn:
I don't know that its safer to strip character out vs just allowing a technically valid filename to be read correctly. I've seen a file rename effectively delete the file it was trying to rename due to some computer glitch while performing the operation. We've seen OT accidentally deleted files for some unknown reason. Trying not to give it another reason to lose a file.

manongjohn:
@shun-iwasawa It appears something has changed on the Travis macOS build environments that is now causing our macOS builds to fail.  

Comparing to a prior successful macOS build log, it appears to be suddenly including LZMA libraries where it wasn't before.  I also see that it's throwing a warning the Xcode 8.3.3 is outdated as well as an error building LLVMSupport.dir which it didn't build before.

Can you take a look to see how to fix this?

artisteacher:
This fix worked on my end. Thanks @manongjohn for tackling this so quickly!

RodneyBaker:
This is more of a Mac related issue so I cannot test this but all appears to be in order in the code.

@manongjohn let us know how you would like to proceed as I don't think @shun-iwasawa or @turtletooth have operational Macs set up for testing purposes.

cc:  @artisteacher Thanks for the validation that it works on your Mac!

PR #2830, opened 2019-10-11
Enable Geometric Tool for Motion Path

shun-iwasawa:
This PR will enable Geometric tool for drawing motion path.
OT already had sources for that, so I reverted it with fixes for avoiding crash.

This PR also fixes the availability of the tools, so that motion path can be edited even if the current column is hidden.

### DIFF
--- a/toonz/sources/tnztools/geometrictool.cpp
+++ b/toonz/sources/tnztools/geometrictool.cpp
@@ -78,9 +78,9 @@ static TPointD rectify(const TPointD &oldPos, const TPointD &pos) {
   const TPointD directions[] = {TPointD(1, 0),  TPointD(h, h),  TPointD(0, 1),
                                 TPointD(-h, h), TPointD(-1, 0), TPointD(-h, -h),
                                 TPointD(0, -1), TPointD(h, -h)};
-  TPointD v        = pos - oldPos;
-  int j            = 0;
-  double bestValue = v * directions[j];
+  TPointD v                  = pos - oldPos;
+  int j                      = 0;
+  double bestValue           = v * directions[j];
   for (int k = 1; k < 8; k++) {
     double value = v * directions[k];
     if (value > bestValue) {
@@ -518,7 +518,7 @@ TPointD Primitive::calculateSnap(TPointD pos) {
         else if (areAlmostEqual(outW, 1.0, 1e-3))
           m_param->m_w1 = 1.0;
         else
-          m_param->m_w1      = outW;
+          m_param->m_w1 = outW;
         TThickPoint point1   = stroke->getPoint(m_param->m_w1);
         snapPoint            = TPointD(point1.x, point1.y);
         m_param->m_foundSnap = true;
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Completed initial testing and this is very useful.
Very nice indeed.

Will continue to test.

RodneyBaker:
For those wishing to try this new Geopathing approach here's a link to the (Windows) download.

[Opentoonz (Geopath) portable link](https://ci.appveyor.com/api/buildjobs/db3hcatw30p0541u/artifacts/toonz%2Fx64%2FOpenToonz.zip)

Recommend:
- While downloading, name this zip file something useful like "Opentoonz (Geopaths)".  Then when extracted you will automatically create two directories inside a "Opentoonz (Geopaths)" folder: Opentoonz and Opentoonz stuff.

- Rename the Opentoonz stuff directory 'portablestuff' and drag/drop it inside the Opentoonz directory.  This will make the release a portable application independent of all other Opentoonz applications on your system and will use the portablestuff directory as its root data folder.

- Inside the Opentoonz directory locate the Opentoonz.exe and run it.
You are now operating a portable Opentoonz with capability to create motion paths with the Geometry tool.

Note:   The link above will no longer work once this PR is merged.  So if you wish to keep it download and save a copy locally.  All appveyor releases expire 6 months from the date of creation unless rebuilt.

RodneyBaker:
@shun-iwasawa 
Could the path making process allow creation of more than one spline?
Say for instance if the Shift key was held down while drawing that might signal that a path was still being drawn?

The reason I ask is that we can make very complex paths already but if we could create more than one it might allow for some very innovative usage.  A simple case would be that of drawing two lines and then connecting them via the Tape Tool.

Perhaps this is already possible and I missed it.

RodneyBaker:
Probably not important but I was just noting that Position Path is not exposed in Current.txt like the other channels. 

https://github.com/opentoonz/opentoonz/blob/6817df33ac50b6be01a2e7b060ea04be5f70f742/stuff/config/current.txt#L816-L827

shun-iwasawa:
Thank you @RodneyBaker for testing this and suggestions.

> Could the path making process allow creation of more than one spline?

I don't think it would be easy, as the motion path is supposed to be referred by using a parameter which defines a proportional position in the single stroke.
As you already noticed, if you would like to create complex curve it would be reasonable workaround to create the curve in the vector level and then copy/paste in on the motion path.

> Probably not important but I was just noting that Position Path is not exposed in the Config file current.txt like the other channels.

You're right. It seems that only the name "posPath" is hard-coded unlike other channel names.
It's not a critical problem as they are currently not translatable. When it comes to making them translatable it should become consistent with the others.

artisteacher:
I haven't had a chance to give this a try. Does poly-line work? You ought to be able to get a fairly complex path with poly-line. Though poly-line would be more useful for making complex paths if it could make cusp control points (usually done while holding alt) so that handles could be adjusted individually on the fly.

RodneyBaker:
@artisteacher Yes, polyline works and works well.

As Shun mentioned before, really complex paths should like be created in vector (using any method to include erasing, Tape Tool, etc) and then copy/pasted into the motion path.
If I'm not mistake I'd say the best workflow there would be to initially create a quick line path and the replace that path via the copy/paste.

Also keep in mind that once a path is created it can be manipulated with other tools such as Smooth, Magnet, etc.  The path is a vector path much like any other spline.

RodneyBaker:
LGTM

PR #2835, opened 2019-10-13
Fix Travis macOS builds

manongjohn:
This is an attempt to fix the failing macOS Travis builds.

1) Fix LZMA compile/linking errors
Prior successful macOS Travis builds did not detect LZMA libraries when building libtiff.a so it built without it. For some reason it is now detecting it and causing the builds to fail later when linking libimage to libtiff built with LZMA support.

For now, disabling LZMA support in libtiff for macOS builds only.

2) Fix Xcode 8.3.3 outdated warnings

Homebrew now recommends a minimum macOS 10.13. We were building on 10.12. Building on 10.12 using Xcode 9.2 even after fixing the LZMA issue was either throwing other linkage errors or compiling homebrew components so slowly that Travis killed the build for taking so long.

Upgraded to Xcode 10.1 which is the last Xcode version Travis has for macOS 10.13. This seemed to fix the problems with taking too long to build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ matrix:
         - os: linux
           compiler: clang
         - os: osx
-          osx_image: xcode8.3
+          osx_image: xcode10.1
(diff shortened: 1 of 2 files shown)


manongjohn:
Travis macOS builds are now completing successfully with these changes. Since this change only affects Travis builds, I'm going to merge this myself.

PR #2837, opened 2019-10-15
Selection pivot fix

### LINKED ISSUE #2836: Pivot is not selectable when overlap on (border and handle box) selection area - Selection Tool
### Issue Summary

Dear @manongjohn , Thank you very much for fix Pivot Selection Tool. There is this issue in 1.3 stable too.

Pivot overlap selection area

### System Information
* **OpenToonz Version**: 1.3 stable, NB `310c4f0`

manongjohn:
This PR does the following:

1) Fixes #2836 by prioritizing the pivot point selection above the selection handles and box.

2) For Vector Selections, this stops the pivot from automatically resetting back to center of selection area every time the selection box is redrawn after moving/deforming selection. Toonz Raster and Raster already do not reset pivot after moving/deforming selection.

### DIFF
--- a/toonz/sources/tnztools/selectiontool.cpp
+++ b/toonz/sources/tnztools/selectiontool.cpp
@@ -972,6 +972,12 @@ void SelectionTool::updateAction(TPointD pos, const TMouseEvent &e) {
     }
     maxDist  = 5 * pixelSize;
     maxDist2 = maxDist * maxDist;
+    if (!isLevelType() && !isSelectedFramesType() &&
+        tdistance2(getCenter(), pos) < maxDist2) {
+      m_what     = MOVE_CENTER;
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Thanks for the quick turnaround on this one @manongjohn.
I note that even the other positive side effects from the previous issue can still be used with this fix.
So this is a win all the way around.

LGTM

beeheemooth:
@manongjohn I find it very useful that the pivot remains in place even when deselecting vectors. In this case, switching between vector levels, you can rotate the lines around a common fixed center, which was very difficult to do before. I ask you to make this option a preference, please.

melieconiek:
Second this one--I was finding it quite handy. If there was an elegant way to keep it fixed while you were working that'd be really useful.

PR #2842, opened 2019-10-19
Fix pivot location on new/modified vector selections

manongjohn:
PR #2837 introduced an issue where the pivot location was not being recalculated to selection center when selecting strokes directly, using freehand/polyline mode or modifying an existing selection by adding or removing a stroke.

Like before, it should now recalculate the pivot whenever a new selection is made or an existing one is modified by adding/removing a stroke.  The pivot should not move after deforming the selection.

### DIFF
--- a/toonz/sources/tnztools/vectorselectiontool.cpp
+++ b/toonz/sources/tnztools/vectorselectiontool.cpp
@@ -587,11 +587,10 @@ void DragSelectionTool::VectorDeformTool::applyTransform(FourPoints bbox) {
   VectorFreeDeformer *freeDeformer =
       static_cast<VectorFreeDeformer *>(tool->getFreeDeformer());
 
-  const bool stayedTheSame =
-      bbox.getP00() == freeDeformer->getPoint(0) &&
-      bbox.getP10() == freeDeformer->getPoint(1) &&
-      bbox.getP11() == freeDeformer->getPoint(2) &&
-      bbox.getP01() == freeDeformer->getPoint(3);
+  const bool stayedTheSame = bbox.getP00() == freeDeformer->getPoint(0) &&
+                             bbox.getP10() == freeDeformer->getPoint(1) &&
+                             bbox.getP11() == freeDeformer->getPoint(2) &&
+                             bbox.getP01() == freeDeformer->getPoint(3);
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Interesting issue.
I've been testing this and I find I like that the pivot stays in place although I also see benefits of it being recalculated.

I'll be glad to hear others weigh in on this.

manongjohn:
Continuing the discussion from #2837 here regarding not recentering pivot when adding or removing a stroke to an already established selection...

I don't want to add yet another preference setting for something so small, however, I am not opposed to making it the default behavior unless there is someone against it.

Anyone against it?

melieconiek:
I believe you should go ahead and fix this. The pivot center resetting on a new selection is the expected behavior in OpenToonz. It's also a behavior consistent with virtually every other software tool out there.

It would be useful if there was a way to temporarily lock the pivot in place, though. I can't really think of an elegant way to do that.

beeheemooth:
@manongjohn If you do not want to add preferences (I understand this is a bunch of extra work), then maybe there is another solution. Is it possible to make double-clicking on the center return it to its default position?

melieconiek:
Perhaps the lock could be toggled by double clicking on it?

manongjohn:
Option on the tool option bar to lock pivot in current location for all selections?

melieconiek:
A checkbox on the tool option bar would work.

beeheemooth:
Agree

melieconiek:
To be clear, looking for it work as it currently does in the nightly--the pivot would still move along with translations, but simply wouldn't reset it's position if the selection changed. Really love the vector selection and handle changes, btw. :D

manongjohn:
I'd like to merge this to fix/restore the pivot location behavior when directly selecting strokes or adding/removing strokes to existing selections.

The pivot will **not** move back to center on an existing selection when the selection is moved or deformed.

Has anyone gotten a chance to test this?

NOTE: I will not be adding the `Lock Pivot` option with this PR. It should be handled separately as a Feature Request.

manongjohn:
I'm going to merge this now. If there are any more issues found with the pivot, I'll take care of it.

PR #2823, opened 2019-10-07
Fix Main Window Full Screen Mode

manongjohn:
This PR fixes #2796 and fixes #2215

There is a known Qt limitation concerning Full Screen mode on Windows system.

The suggested workaround was only applied to the Image viewer to allow it to go to full screen. I applied the same workaround to the main window to fix the full screen issue.

PR #2847, opened 2019-10-21
Fix Crash on Launch without Viewer

shun-iwasawa:
This PR will patch a problem related to the recent fix #2823 as follows:
- OT crashes when launching without any OpenGL-related widgets (e.g. Viewer, Style Editor, Flipbook, etc.) contained in the main window.

### DIFF
--- a/toonz/sources/toonz/main.cpp
+++ b/toonz/sources/toonz/main.cpp
@@ -233,7 +233,7 @@ project->setUseScenePath(TProject::Extras, false);
   // Imposto la rootDir per ImageCache
 
   /*-- TOONZCACHEROOTの設定  --*/
-  TFilePath cacheDir               = ToonzFolder::getCacheRootFolder();
+  TFilePath cacheDir = ToonzFolder::getCacheRootFolder();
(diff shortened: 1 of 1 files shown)


gab3d:
something related to launching OT i've found today:
when you fresh-install the latest nightly, the Basic room shows no Style Editor, but when you open another one and try to dock it somewhere, the lost style editor reappears.

manongjohn:
Hmm. Interesting. 

I can't seem to reproduce the issue on my Win10 machines before this fix is applied.  I tried restarting OT with the initial room being a newly created blank room with no controls in it, but still can't get it to crash.  Is there a certain way to show this problem so I can verify the fix?

manongjohn:
@gab3d  I confirmed that issue with the Style Editor in the Basic room. It started with PR #2814, not this one.  The other way to restore that is to resize the main OT window while still in the Basic room.

A separate Issue should be opened for this problem.

manongjohn:
Ok. I was able to duplicate this crash by removing all rooms except Browser and Farm and confirmed this changed fix it.

Issue #2844, opened 2019-10-19
Pinch & Magnet Tool size sliders are non-linear.


melieconiek:
The size sliders for the Pinch and Magnet tools are non-linear--is this intended?

I think this change makes sense for the brush and eraser tool--non-linear sliders give them more precision at the lower end where they're most often used. Other tools like the Pinch and Magnet tool require more precision at higher values.

Tested in nightly.

gab3d:
perhaps holding a modifier key while adjusting it to switch to linear? 🤔

melieconiek:
I've already switched them back--don't want to sound dramatic, but they were unusable. I had to enter the value in the box instead.

melieconiek:
...and I should have looked at the PR beforehand, the change was intended. What are feelings on reverting this, though?

shun-iwasawa:
Switching the slider to non-linear is supposed to be introduced with decupling (x10) the maximum values. But I decided not to expand the maximum value for the Pinch tool and I only quintupled (x5) for the Magnet tool. It may be a reason of your feeling of user-unfriendliness.
I'll try making  both sliders with proper maximum values and would like to get some feedback.

RodneyBaker:
@melieconiek Did this resolve the issues you were having?
Will reopen this issue if it did not.

PR #2848, opened 2019-10-21
Modify Maximum Values for Pinch and Magnet Tools Size

shun-iwasawa:
Responding to #2844, this PR will expand the maximum values for Pinch and Magnet tools in order to maintain the usability of non-linear sliders compared to the linear ones.
- `Size` of the Pinch tool is expanded from 1,000 to 10,000.
- `Size` of the Magnet tool is expanded from 500 to 1,000.

I begin to think that the `Size` slider of Pinch tool can be reverted to linear, because unlike other sliders it changes one-dimentional parameter.
Anyway, I would like to get feedback from users.

### DIFF
--- a/toonz/sources/tnztools/magnettool.cpp
+++ b/toonz/sources/tnztools/magnettool.cpp
@@ -142,7 +142,7 @@ public:
       : TTool("T_Magnet")
       , m_active(false)
       , m_oldStrokesArray()
-      , m_toolSize("Size:", 10, 500, 20)  // W_ToolOptions_MagnetTool
+      , m_toolSize("Size:", 10, 1000, 20)  // W_ToolOptions_MagnetTool
   {
     bind(TTool::Vectors);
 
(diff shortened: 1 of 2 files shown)


melieconiek:
Thank you for the attention to this.

Tested the appveyor build--the increased range for the magnet tool makes it much easier to set the range at larger scales--useful for tweaking entire drawings. The size here is ~500px:

_temp_o

Like the magnet tool, the appropriate scale for the tweak tool can vary quite a bit depending on what it is you're tweaking. For longer lines this value can be quite large. So it's possible that you may need to be setting precise values in the upper range of the toolbar slider.

---

On a related note, the size handle in the viewer is still capped to a maximum of 1000. What we want that range to be set to is a separate issue, but if is being changed, we may want the ranges to be consistent.
image

RodneyBaker:
Question regarding tthe Magnet Tool:

**Is there a specific reason why the Minimum Value is 10 and not 1?**
Here I must assume values less than 10 are seen to defeat the intended purpose and be better handled by other tools such as Pinch or Control Point Editor.

It just seems odd to me that as I'm working with the Magnet Tool and trying to get the influence refined to influence finer detail that the minium value on the slider would be 10.

PR #2854, opened 2019-10-22
Fix bounding box computation in TXshSimpleLevel

caryoscelus:
I'm not entirely sure what the `TDimensionD(info->m_lx, info->m_ly))` line was supposed to be doing, but it clearly doesn't take into account the endpoint from info and it seems to be making every bounding box the same size. With the fix it seems to work correctly

### DIFF
--- a/toonz/sources/toonzlib/txshsimplelevel.cpp
+++ b/toonz/sources/toonzlib/txshsimplelevel.cpp
@@ -2294,9 +2294,8 @@ TRectD TXshSimpleLevel::getBBox(const TFrameId &fid) const {
     if (!info) return TRectD();
 
     bbox = TRectD(TPointD(info->m_x0, info->m_y0),
-                  TDimensionD(info->m_lx,
-                              info->m_ly)) -  // Using lx, ly is less ambiguous
-           0.5 * TPointD(info->m_lx, info->m_ly);
+		  TPointD(info->m_x1, info->m_y1))
+          - 0.5 * TPointD(info->m_lx, info->m_ly);
 
     if (info->m_dpix > 0.0 && info->m_dpiy > 0.0)
       dpiX = info->m_dpix, dpiY = info->m_dpiy;


manongjohn:
If this is something users can see, is there an image or something you can include in the PR's description so we can see what impact this change is having or the problem it is resolving?  This way we can know what to look for while testing?

I believe that function is used for a number of things, internally.  Would need to see what that is impacting.

toonz/sources/toonzlib/txshsimplelevel.cpp
```diff
@@ -2294,9 +2294,8 @@ TRectD TXshSimpleLevel::getBBox(const TFrameId &fid) const {
     if (!info) return TRectD();
 
     bbox = TRectD(TPointD(info->m_x0, info->m_y0),
-                  TDimensionD(info->m_lx,
-                              info->m_ly)) -  // Using lx, ly is less ambiguous
-           0.5 * TPointD(info->m_lx, info->m_ly);
+		  TPointD(info->m_x1, info->m_y1))
+          - 0.5 * TPointD(info->m_lx, info->m_ly);
```
manongjohn: My concern about this change is when creating a TImageInfo variable,  m_lx and m_ly are passed in, but m_x1 and m_y1 are defaulted to -1 so if those are not properly set at some point, your change may cause problems.  I think that's why it is basically creating the bounding box by specifying the topleft point and then using the ImageInfo width/length to create the box, rather than specified points.

It might be worthwhile to test m_x1 and m_y1 for negative values and use the old logic, if it is.

caryoscelus:
@manongjohn unfortunately i don't know if there's any place user can see it directly. I've stumbled into this while working on another feature, which isn't really ready yet. I can probably make an image of this bounding box in action or share the terrible WIP code if that would help

toonz/sources/toonzlib/txshsimplelevel.cpp
caryoscelus: > when creating a TImageInfo variable, m_lx and m_ly are passed in, but m_x1 and m_y1 are defaulted to -1 so if those are not properly set at some point, your change may cause problems

TImageInfo can be created without m_lx and m_ly too. And m_x0/m_y0 are set to zero by default, so unless they are set later, info is quite meaningless.

> It might be worthwhile to test m_x1 and m_y1 for negative values and use the old logic, if it is.

That would overcomplicate the code, i'm afraid. If i understand correctly, negative values would result in empty rectangle, which is anyway what's returned in case there's no info. Perhaps it would be better to investigate whether TImageInfo is ever actually initialized without proper x1/y1 values

caryoscelus:
bbox

toonz/sources/toonzlib/txshsimplelevel.cpp
manongjohn: One would hope that not passing in lx and ly  would simply default to 0s.  Granted we've had to fix crashes if boxes were 0 length or width. 

x0 and y0 set to 0 by default is just the origin which is rather normal for any type of coordinate plane so I'm not concerned by that not being initially set.  Most image's topleft is (0,0).

Looking at the comment in imagemanager.h for  `isImageCompatible()`, I gather that x1/y1 would be negative if the image hasn't been cached by the ImageManager, yet.  A while back I ran into issues where the logic was accessing an icon image but hadn't been cached yet and was causing a crash.

I'll let other developers who might be more familiar with the usage of that function weigh in. Might be ok as you have it.  I'm just trying to make sure we aren't introducing another possible cause for a crash. :)

caryoscelus: > I'm just trying to make sure we aren't introducing another possible cause for a crash. :)

Yeah, i understand that. It's just that i don't want to introduce another code complication - it took me some time to investigate it as it is.

If there is no other way to avoid the crash, then ok, but even then i'd rather have the function return empty rect or the whole image rect than whole image rect offset by coordinates of drawn area beginning like it does now. Because such "half-working" state is very confusing

shun-iwasawa:
Thank you @caryoscelus , I think the fix is correct.
The modified function `TXshSimpleLevel::getBBox()` seems to be called from only the function `TXsheet::getBBox()` . And `TXsheet::getBBox()` is used only for creating a texture to be used when the sub-xsheet level is deformed using the Plastic tool.

I confirmed that the texture quality when animating the sub-xsheet containing toonz raster level is improved with this PR. :+1:

Jenkins

caryoscelus:
@shun-iwasawa good to hear it fixes something i didn't even know about ;)

RodneyBaker:
LGTM as is. Back to others for merging as I don't know if additional changes are pending for this PR.

PR #2858, opened 2019-10-25
Enable Animate tool to control precisely by pressing the Alt key

shun-iwasawa:
This PR will enable Animate tool to control precisely by pressing the Alt key.
When pressing the Alt key, an extra decoration will be shown in the mouse cursor and one tenth of the mouse movement will effect to the control.


gab3d:
Sorry to ask, but:
Could a **Ctrl + arrow keys** (or some other modifier key) be implemented for moving the current column 1 pixel at a time?
It could also add to the precision, while positioning elements on screen.

shun-iwasawa:
@gab3d 

> Could a Ctrl + arrow keys (or some other modifier key) be implemented for moving the current column 1 pixel at a time?

I once thought so too, but the arrow keys are already occupied for moving the current frame, so IMO it would be confusing to assign the different behavior to Ctrl+Arrow key.

Please note that the Selection tool has an ability to move the selection by using arrow keys. This is achieved by overriding the assignment of arrow keys only when there is something selected. Without the selection, you can still use the arrow keys for moving the current frame even if the Selection tool is active.
The Animate tool has no special state like the Selection tool. Therefore you only can override the arrow keys when the Animate tool is active, which will lose ability to move the current frame.

gab3d:
@shun-iwasawa 

> The Animate tool has no special state like the Selection tool. Therefore you only can override the arrow keys when the Animate tool is active, which will lose ability to move the current frame.

yeah, I realize that, that's why I mentioned a posible method using a modifier key like Ctrl to access to it.

I use the function in the Selection tool quite often but that, of course, makes permanent changes inside of a drawing, and not changes to the position of a column, which could also be of real good use, in my opinion.
Regarding the arrows being used for frame change, I don't know what others could think, but I don't see it to be a problem if the same keys would be used for tweaking the position of a column/layer by using Ctrl or some other modifier key to access this behavior. 🤔

RodneyBaker:
Agreed that the use of Ctrl echoing the behavior found with Selection Tool would be a good thing.

RodneyBaker:
Merging as the additional suggestions fit best with other PR.

PR #2859, opened 2019-10-25
Add "Hide Selected" Command to the Function Editor Spreadsheet

shun-iwasawa:
This PR will add "Hide Selected" command to the context menu in the column header area of the Function Editor spreadsheet.
It enables users to hide all selected channels without toggling in the tree view. 

hide_function_sheet_columns

### DIFF
--- a/toonz/sources/toonzqt/functionsheet.cpp
+++ b/toonz/sources/toonzqt/functionsheet.cpp
@@ -462,6 +462,18 @@ void FunctionSheetColumnHeadViewer::mousePressEvent(QMouseEvent *e) {
 
     getViewer()->selectCells(rect);
   }
+  // Switch selection before opening the context menu
+  // if the clicked column is out of the selection
+  else if (e->button() == Qt::RightButton) {
+    QRect selectedCell = getViewer()->getSelectedCells();
+    if (selectedCell.left() > currentC || selectedCell.right() < currentC) {
+      int lastKeyPos = 0;
+      std::set<double> frames;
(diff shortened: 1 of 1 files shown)


RodneyBaker:
LGTM.  
I almost merged this but thought I would mention this:

Is this hiding/unhiding something that can be added to History/Undo/Redo?
After Hiding a few columns I instinctively hit Undo and noted that History skipped all of the hiding/unhiding and did undo on previous events.  

It doesn't take long to cure ourselves of doing this of course but the thought occurs to me that if we accidentally hide some columns it would be useful to them them back via Undo before moving on.

shun-iwasawa:
@RodneyBaker 

> Is this hiding/unhiding something that can be added to History/Undo/Redo?

Hiding / unhiding the channels is not a change to the scene data, but rather a change to UI. The states of channel visibility are stored in each Function Editor panel. Basically the changes to UI are not stored in the History, just like you cannot undo the change of the brush size.
So, I think we need to take another way to revert the operation.

hayoko:
I would merge as is, looks good.

You don't want to add history to non-destructive changes, exactly in the way Shun Iwasawa has already described, let's not even think about it.

There are already three points of action, selecting, opening a menu and hitting confirm. If the user did the wrong thing that's totally their mistake, so just re-open the columns, no biggie.

RodneyBaker:
Agree.  The Undo History was an itch I had to scratch and I appreciate the feedback and explanations.
LGTM

PR #2867, opened 2019-10-31
Revised vector selection pivot re-centering logic

manongjohn:
Some Discord users were reporting new issues with the vector selection pivot logic and I was seeing additional issues when in Level Edit mode.

In retrospect, the implemented changes were...well...bad, I must admit.

I restored the original logic and revised it again in a different way to prevent the re-centering of the pivot when a selection was being moved, rotated, deformed or thickness changed.

The existing and default behavior is to recenter the pivot with undo/redo operations or whenever the selection objects change like adding/removing new strokes.  The pivot will still move when the bounding box is resized unfortunately.

I do note there are issues with pivots moves with undo/redo operations, since isolated pivot move history is tracked separately.  Undo/Redo operations do not account for original pivot location so it can throw off the position history. These issues will not be fixed with this PR.

### DIFF
--- a/toonz/sources/tnztools/vectorselectiontool.cpp
+++ b/toonz/sources/tnztools/vectorselectiontool.cpp
@@ -737,6 +737,10 @@ void DragSelectionTool::VectorDeformTool::leftButtonUp(const TPointD &pos,
   m_isDragging = false;
 
   tool->notifyImageChanged();
+
+  VectorSelectionTool *selectionTool =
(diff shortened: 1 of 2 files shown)


melieconiek:
Everything seems to be working correctly here. :)

RodneyBaker:
> The pivot will still move when the bounding box is resized unfortunately.

Yay!  I like that functionality.
If desiring to reset the pivot to center all we need to do is reselect.

LGTM

PR #2872, opened 2019-11-11
Fix navigation keyframe shortcuts skipping

### LINKED ISSUE #2871: Next/Prev keyframe shortcuts sometimes skips keys
### Issue Summary
Reported by a Discord user, sometimes when using the Next/Prev keyframe shortcuts, it will skip over keyframes instead of moving to the keyframe nearest to it.

### Steps to Reproduce
1. Open a new scene
2. Create a level and extend it for several frames
3. Add several keys, at least 3 but more is better.
4. Go to first key and use Next keyframe shortcut (Ctrl+.) -> It will go to the 2nd keyframe
5. Go to another room with an xsheet/timeline
6. Switch back to the room you started in
7. Go to the first key and use Next keyframe shortcut (Ctrl+.) -> It will skip to the 3rd keyframe
8. If you go to another room you haven't visted yet with an xsheet/timeline and go back and use the shortcut again, it will skip over 2 keyframes.

### System Information
* **OpenToonz Version**: latest nightly

manongjohn:
This PR fixes #2871 

When switching rooms, the keyframe navigators were not being disconnected. When switching back to the room, the keyframe navigators were getting another connection so when the signal was being emitted to handle keyframe navigation, it would execute it multiple times.

For some reason `disconnect(m_framehandle)` was not enough to remove all connections. I added logic to specifically disconnect the slots responsible for moving keyframes.

### DIFF
--- a/toonz/sources/toonzqt/keyframenavigator.cpp
+++ b/toonz/sources/toonzqt/keyframenavigator.cpp
@@ -148,6 +148,12 @@ void KeyframeNavigator::showEvent(QShowEvent *e) {
 void KeyframeNavigator::hideEvent(QHideEvent *e) {
   if (!m_frameHandle) return;
   disconnect(m_frameHandle);
+
+  disconnect(m_frameHandle, SIGNAL(triggerNextKeyframe(QWidget *)), this,
+             SLOT(onNextKeyframe(QWidget *)));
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Confirm the original issue and that this PR fixes the problem.
LGTM

PR #2873, opened 2019-11-11
New Fx: Spin Gradient Iwa Fx

shun-iwasawa:
This PR will add a new fx; Spin Gradient Iwa Fx for generating fan-shaped gradient between two specified angles.
Pixels outside of the gradient range will be "clamped" - i.e. filled with color of either start or end angles, whichever is nearer.

spingradient

### Parameters

- **Center** : specifies the center position of the gradient.
- **Type** : specifies the interpolation type of the gradient.
- **Start Angle** and **End Angle** : the angle range of gradient. The gradient is generated from Start to End angles, in a counterclockwise direction.
- **Start Color** and **End Color** : colors for each end of the gradient.

Start Angle and End Angle handles in the Fx gadget can be dragged with following modifiers:

- **+Shift** : move the handle every 10 degrees.
- **+Ctrl** : move the both handles at the same time.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1286,6 +1286,14 @@
   <item>"STD_iwa_CorridorGradientFx.inner_color"	"Inner Color"		</item>
   <item>"STD_iwa_CorridorGradientFx.outer_color"	"Outer Color"		</item>
 
+  <item>"STD_iwa_SpinGradientFx" 			"Spin Gradient Iwa" 	</item>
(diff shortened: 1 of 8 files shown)


RodneyBaker:
Nice stuff going on here.
For a moment I was wondering if perhaps we could also have a middle point/breakdown color between the start and end angles but then I realized we can use more than one spin gradient fx to get that and a whole lot more done.

TwoSpinGradients

LGTM

artisteacher:
This works really well, though I wonder if the name should be conical instead of spin.

PR #2874, opened 2019-11-12
Fix Type Tool Cursor Jumping

shun-iwasawa:
This PR fixes a following problem of Type Tool when inputting text with non-Latin alphabets such as Japanese.

- While typing the text, pressing the Shift key unexpectedly moves the cursor to the beginning.

In the type tool [QInputMethodEvent](https://www3.sra.co.jp/qt/relation/doc/qtgui/qinputmethodevent.html) is used only when typing non-latin alphabets. `TypeTool::m_preeditRange` specifies the location where the text to be inserted.
Pressing the Shift key reset the location (at [line 1534](https://github.com/opentoonz/opentoonz/compare/master...shun-iwasawa:g/fix_type_tool_cursor_jumping?expand=1#diff-b408ae6d02f0b8239a28e3ecbec5b9a3L1534)) before going out of the function, which was the cause of the problem.

### DIFF
--- a/toonz/sources/tnztools/typetool.cpp
+++ b/toonz/sources/tnztools/typetool.cpp
@@ -700,7 +700,7 @@ void TypeTool::updateStrokeChar() {
 
 void TypeTool::updateCharPositions(int updateFrom) {
   if (updateFrom < 0) updateFrom = 0;
-  UINT size                      = m_string.size();
+  UINT size = m_string.size();
   TPointD currentOffset;
   TFontManager *instance = TFontManager::instance();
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@shun-iwasawa   All appears to be in order.
I'm not sure how thoroughly I can test non-latin alphabets without adjusting my operating system to fully test however.  I can press into that if necessary.
Otherwise, recommend merging this PR yourself at your own discretion.

turtletooth:
LGTM.

PR #2875, opened 2019-11-12
Enable to create folder in the Startup Popup

shun-iwasawa:
This PR will enable to create folder in the Startup Popup, if the specified path in the "Save In" field does not exist.

This PR also sets the focus proxy from `FileField` widget to the path field, so that canceling the dialog will properly focus the path field.

### DIFF
--- a/toonz/sources/toonz/startuppopup.cpp
+++ b/toonz/sources/toonz/startuppopup.cpp
@@ -274,15 +274,15 @@ StartupPopup::StartupPopup()
   bool ret = true;
   ret      = ret && connect(sceneHandle, SIGNAL(sceneChanged()), this,
                        SLOT(onSceneChanged()));
-  ret = ret && connect(sceneHandle, SIGNAL(sceneSwitched()), this,
+  ret      = ret && connect(sceneHandle, SIGNAL(sceneSwitched()), this,
                        SLOT(onSceneChanged()));
-  ret = ret && connect(newProjectButton, SIGNAL(clicked()), this,
+  ret      = ret && connect(newProjectButton, SIGNAL(clicked()), this,
                        SLOT(onNewProjectButtonPressed()));
-  ret = ret && connect(loadOtherSceneButton, SIGNAL(clicked()), this,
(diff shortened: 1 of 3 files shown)


RodneyBaker:
@shun-iwasawa  Would it be possible to have Opentoonz save the scene file to disk (even though empty) upon clicking the 'Create' button in the Startup Popup?  

Rationale:  There seems to be a number of issues that relate to users not saving their scene and/or levels and the current workaround appears to be to immediately perform a 'Save All' after the Startup Popup closes and the user is looking at the blank canvas.  If they do not save at this point the risk of losing their scene and/or files increases.

I must assume the scene name entered in the Scene Name field gets stored in memory at present awaiting a later time when the scene and/or levels will be saved.  

Establishing the scene in the location specified by the user would seem to be a reasonable thing to expect when a user clicks 'Create'.  As things work now, afterward any changes will flag the scene/levels as dirty prompting the user to save their files.

shun-iwasawa:
@RodneyBaker

> Would it be possible to have Opentoonz save the scene file to disk (even though empty) upon clicking the 'Create' button in the Startup Popup?

@turtletooth had already prepared [the line for that](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/startuppopup.cpp#L496-L497)! 
It seems to have been left open for discussion but IMO it would be reasonable to save the scene on clicking "Create" as you pointed.

RodneyBaker:
> @turtletooth had already prepared the line for that!

I've always like that Jeremy guy.  ;)

RodneyBaker:
@shun-iwasawa  Regarding your original PR itself.
LGTM

turtletooth:
Good work @shun-iwasawa.  LGTM

RodneyBaker:
Post mortem:
The activation of code saving the scene file upon creation wasn't implemented via this PR.
Perhaps it still requires a discussion prior to implementation.

https://github.com/opentoonz/opentoonz/blob/98a68d5c9b720890d06406ba92f5643250cf6052/toonz/sources/toonz/startuppopup.cpp#L507

PR #2876, opened 2019-11-12
Fix ffmpeg loaded levels

### LINKED ISSUE #2870: How OpenToonz kills uploaded files (video attached)
### Issue Summary

OpenToonz deletes uploaded files (bitmaps and gifs) without warning after saving if they have been modified. Also, along with them, auxiliary files are deleted regarding deleted images - xml with tracking data.

ezgif-1-46cce27093dc

### Steps to Reproduce

1. Load one or more images (or gif) into the scene. They automatically fall into the "extras" folder. Watch the file in the folder.
2. Use the raster brush or eraser to make any changes to the image.
3. Save the scene.
4. The file from the "extras" folder will be deleted.

### System Information
* **Operating System**: windows 8.1

manongjohn:
This PR accomplishes the following regarding ffmpeg processing of GIF/MP4/WEBM files:

- Correctly read frame rate from the source file and use internally as needed ( fixes #2870 )

The frame rate was not being read from the file correctly and in many cases defaulting to 0. When used for ffmpeg, a 0 frame rate causes ffmpeg to fail. This utlimately lead to original files being lost when being updated in a save operation.

Beyond reading the frame rate correctly from the files, for GIF processing, if the frame rate = 0 (not an animated gif) it will default to 12, similar to other file processing.

- Make GIF/MP4/WEBM files loaded into OT through ffmpeg as Read Only.

The way files are read into OT through ffmpeg does not necessarily mean it will be written back in the same exact way when saving changes to it through ffmpeg.  Allowing one to save changes to any of these files can result in unintended side effects to the original animation, particularly when adding/removing frames.

For now, we will make these files Read Only, similar to PSD files.  Future enhancements may remove this.

- Speed up loading frame counts in File Browser

Currently, the frame counts for GIF/MP4/WEBM files is performed by running ffmpeg to extract all the frames of the file in the cache directory and count them.  This is problematic if there are a number of these files in a directory to be scanned and/or if a file has a lot of frames, leading to long wait times and potentially locking up OT for a long time.

Modified the logic to not extract the frames from the file until it is actually loaded into OT with a Load Level operation.

Instead, we will use the following information from the file to calculate the expected amount of frames as: r_frame_rate * duration

### DIFF
--- a/toonz/sources/image/ffmpeg/tiio_gif.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_gif.cpp
@@ -194,8 +194,6 @@ TLevelReaderGif::TLevelReaderGif(const TFilePath &path)
   m_lx                    = m_size.lx;
   m_ly                    = m_size.ly;
 
-  ffmpegReader->getFramesFromMovie();
-  m_frameCount = ffmpegReader->getGifFrameCount();
   // set values
   m_info                   = new TImageInfo();
   m_info->m_frameRate      = fps;
@@ -239,6 +237,10 @@ TDimension TLevelReaderGif::getSize() { return m_size; }
 //------------------------------------------------
 
 TImageP TLevelReaderGif::load(int frameIndex) {
+  if (!ffmpegFramesCreated) {
+    ffmpegReader->getFramesFromMovie();
+    ffmpegFramesCreated = true;
+  }
   return ffmpegReader->getImage(frameIndex);
 }
 
--- a/toonz/sources/image/ffmpeg/tiio_gif.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_gif.cpp
@@ -92,7 +92,7 @@ TLevelWriterGif::~TLevelWriterGif() {
   preIArgs << "-v";
   preIArgs << "warning";
   preIArgs << "-r";
-  preIArgs << QString::number(std::max(m_frameRate, 12.0));
+  preIArgs << QString::number((m_frameRate < 1 ? 12.0 : m_frameRate));
   if (m_palette) {
     postIArgs << "-i";
     postIArgs << palette;
--- a/toonz/sources/image/ffmpeg/tiio_gif.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_gif.cpp
@@ -92,7 +92,7 @@ TLevelWriterGif::~TLevelWriterGif() {
   preIArgs << "-v";
   preIArgs << "warning";
   preIArgs << "-r";
-  preIArgs << QString::number(m_frameRate);
+  preIArgs << QString::number(std::max(m_frameRate, 12.0));
   if (m_palette) {
     postIArgs << "-i";
     postIArgs << palette;
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -258,8 +258,8 @@ ffmpegFileInfo Ffmpeg::getInfo() {
   } else {
     QFile infoText(tempPath);
     getSize();
-    getFrameRate();
     getFrameCount();
+    getFrameRate();
     infoText.open(QIODevice::WriteOnly);
     std::string infoToWrite =
         std::to_string(m_lx) + " " + std::to_string(m_ly) + " " +
@@ -313,7 +313,7 @@ double Ffmpeg::getFrameRate() {
     fpsArgs << "-select_streams";
     fpsArgs << "v:0";
     fpsArgs << "-show_entries";
-    fpsArgs << "stream=avg_frame_rate";
+    fpsArgs << "stream=r_frame_rate";
     fpsArgs << "-of";
     fpsArgs << "default=noprint_wrappers=1:nokey=1";
     fpsArgs << m_path.getQString();
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -306,24 +306,22 @@ TRasterImageP Ffmpeg::getImage(int frameIndex) {
 }
 
 double Ffmpeg::getFrameRate() {
-  if (m_frameCount > 0) {
-    QStringList fpsArgs;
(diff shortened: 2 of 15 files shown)


RodneyBaker:
I like the changes here.


Regarding the read only files.
I agree that making the images read only is a good idea.
I would be good to somehow indicate to the user the images' read only status.  I'm not sure the cursor change is enough to keep users from just thinking there is something wrong with the Brush Tool (assuming they try to draw on a frame).

Aside:  This would be another PR but I cannot help but think there is an opportunity to use conversion in lieu Import and Load to get past some of the issues related to initial ingest of files.  Perhaps there could be four options in the Load dialogue:  import, load, convert and cancel?  The conversion would trigger the convert found via Browser Room, etc.

toonz/sources/image/ffmpeg/tiio_gif.cpp
```diff
@@ -92,7 +92,7 @@ TLevelWriterGif::~TLevelWriterGif() {
   preIArgs << "-v";
   preIArgs << "warning";
   preIArgs << "-r";
-  preIArgs << QString::number(m_frameRate);
+  preIArgs << QString::number(std::max(m_frameRate, 12.0));
```
turtletooth: Can a gif have a lower fps than 12?

turtletooth: Would it be better at:
preIArgs << QString::number((m_frameRate < 1) ? 12 : m_frameRate);

manongjohn: I have a gif that I downloaded from some site some time ago that has a r_frame_rate=100/10 (10fps) with a duration = 0.1 which calculates to a 1 frame gif, which it is, so it can be less than 12.

I prefer to preserve whatever frame rate is obtained from the file. If for some strange reason there isn't one, i'm assuming 12

Actually the only reason it was 0 was because the getFrameRate() was called before getFrameCount() and there was a condition to only get frame rate if frame count > 0 so it was never getting rate properly.

turtletooth: Line 95 as it stands will overwrite a fps lower than 12 with 12.

manongjohn: Think i meant min, but realize that will not help if it's 0. Will fix as suggested. Thanks for reviewing.

turtletooth:
LGTM.  Thanks @manongjohn

RodneyBaker:
We may have pulled the trigger too soon on this one.

manongjohn:
Is there another issue?

RodneyBaker:
There are quite a few issues with not being able to work with gif (and presumably MP4 and Webm) in the ways prior to this PR.  Ones that immediately come to mind  vary from simple retiming operations to selecting parts of an image and copy/pasting those selections into a new level.  

There are also issues such as not being able to edit a toonz raster level that has been cleaned up from a gif raster level.  In that case the process will work but only after the scene and level have been saved and the scene reloaded.  It apparently still reads as read only after the cleanup.

The biggest issue appears to be that we are losing some of the benefits/strengths of the gif and mp4 file formats when we can't work with them because they are read only.

Additionally, we lose many of the benefits of standard raster levels to include blending with mypaint brushes or simply drawing onto the images.

Some of this certainly could be overcome by adjusting workflow but we are working against a standard workflow where most users expect to open an image and immediately be able to interact with it.

Possible workarounds?
Hmmm... not sure.
Perhaps give the user an option to load as read only or convert to editable TIF sequence?

I like the underlying idea of read only but it seems to have added several new levels of complexity.

manongjohn:
You can always use OT to convert a gif/mp4/webm to a PNG sequence or other editable format before using if you need to copy parts of it.  If you are simply using it for visual reference or rotosciping, then you can use it directly.

If people are expecting to modify an actual GIF/MP4/WEBM directly, they will be sadly mistaken.  OT doesn't do that. When loading a GIF, OT converts to PNG files with ffmpeg., then the PNG files are loaded into OT. When saving, the PNG files are packed back together with ffmpeg into a temp file, then the temp file overwrites the original GIF file.  Any special properties in the original GIF are lost.  This is especially bad for MP4s because we don't extract the audio. The audio is lost when the file is recreated.

To show how bad this can be: I took a single image gif, added a second frame and saved.  I closed OT and loaded the gif back in.  The gif when loaded now has 20 frames (only 2 original frames). The 1st 10 frames was the original drawing. The next 10 frames somehow merged the 2 frames.

Some tools automatically flag a level to be updated, like the Tracker Tool even if you made no actual change to any of the drawings, so when you save the scene, you've destroyed the original file for no reason.

We can make it editable if there is demand for it, but I have no current intentions of ensuring the integrity of the original attributes is maintained in the saving process.

RodneyBaker:
There are several competing issues here.
My primary concern is the editability of images brought into Opentoonz.
Users will expect image they bring into OT to be editable.

Concerning the defaulting to 0 for gif.
Take the following with a very large grain of salt... because I don't really know but am trying to understand.

Gif format doesn't really have a framerate.
Instead the format uses a delay so a setting of 0 should mean 'no delay'.
If we default that to 12 then we should expect to see 12 copies of the same frame (upon conversion).
Framerate for gifs is largely dictated by the browser/application it is being viewed in and it will play the frames as fast as it can taking into account any injected delay.
This might be why you are seeing frames magically appear?

manongjohn:
I actually know why frames are "magically appearing", but I'm not working out how to fix it.  If someone else wants to do that, then great. I'll gladly test their changes.

As for your questions.  Find out how to use ffprobe to get the properties of any gif file, then see if your statements still hold true and if your questions get answered.

artisteacher:
On the positive side, no one will be accidentally drawing directly on video they didn't intend to draw on. 

For now, an easy way to make a gif/mp4 editable directly in OT, is merging the mp4 or gif onto a new raster level. The merged level is immediately editable. I do wish that the level strip icons & column header icon updated once merging finalized though. There's no real indicator that merging was successful unless you remove or hide the original level(s).

> This is especially bad for MP4s because we don't extract the audio. The audio is lost when the file is recreated.

It would be so nice if audio could be extracted! I suppose there would need to be a load option/preference about loading with or without audio.

PR #2878, opened 2019-11-14
Linear Gradient Fx Revised

shun-iwasawa:
This PR will replace the Linear Gradient Fx by the new one with more flexibility.
The new fx has two point parameters `Start Point` and `End Point` to directly specify the positions of the both ends of gradient, instead of using combination of `Size` parameter and column transformation.

lineargradientrevised

I decided not to add this fx as a new one (like "Linear Gradient Fx Iwa") but replace the conventional one having the same fx name ("Linear Gradient Fx") for sake of simplicity.
The source of the conventional one is kept so that users can still load the fx made with the older versions.

Start Point and End Point handles in the Fx gadget can be dragged with following modifiers:

- **+Shift** : move the handle keeping gradient direction.
- **+Ctrl** : symmetrically move the both handles at the same time.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1294,6 +1294,16 @@
   <item>"STD_iwa_SpinGradientFx.endAngle"	"End Angle"		</item>
   <item>"STD_iwa_SpinGradientFx.endColor"	"End Color"		</item>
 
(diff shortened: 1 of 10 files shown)


RodneyBaker:
Very nice.
The new controllers you are bringing online with these fx updates will be very useful.

LGTM

PR #2880, opened 2019-11-14
Handle missing ffmpeg info

manongjohn:
This is a followup to PR #2876.

A Discord user was testing the nightly and found the gifs he was using were not loading into OT properly.  Oddly, when I used the same gifs from the same sites, I was not having issues.

Working with him to debug using ffmprobe directly, it was determined that for some reason the gifs on his system had N/A for the duration, which caused the calculation of the frame count to be 0. The gifs on my system had duration valued appropriately. Not sure why there is a difference.

At any rate, I've added checks to fall back to use the original ffmpeg attributes  info (ave_frame_rate/nb_read_frames) if the r_ave_frame_rate or duration attributes don't come back with valid data.


manongjohn:
FYI, Discord user tested artifact and it corrected the issue.

RodneyBaker:
Will merge this as it is an immediate fix for known issue.

PR #2882, opened 2019-11-15
Scene/Level Editor changes

### LINKED ISSUE #1061: InknPaint Workflow Request (Background Color)
Currently we use Clip Studio Paint or Retas for InknPaint, what I find is that the painters I work with prefer to paint on a **neutral gray BG** or **off-white BG** instead of a **pure white BG**.

Navigating frames with the XSheet allows us to disable Camera BG Color using ComboViewer, but navigating frames using the FilmStrip causes a permanent white BG to be displayed.

I'd just like to request adding the ability to either turn the white BG off for FilmStrip or allow the user to pick a color in preferences, if possible.

### LINKED ISSUE #2877: [Discussion] Level Edit backgrounds
Currently when you switch to Level Edit mode (click frame in Level Strip), the background for all levels goes gray but the Raster has a white bounding box, and Toonz Raster goes a has filled white box.

Is there a particular reason why they are like this? Do we want it to stay this way?  Should Raster show the filled white box or should Toonz Raster show only the white bounding box? 

As the Vector one just shows gray and Raster also going gray with just the bounding box, I would lean towards Toonz Raster only showing just a white bounding box so it's obvious you've switched to Level Edit mode.

I know this has been somewhat discussed before and we've all just been like "that's how it is and you just need to know the differences", but I think it should be changed to avoid the confusion this causes to new users.

Thoughts?

manongjohn:
This PR resolves #1061 based on discussion from #2877 ( closes #2877 )

The following changes have been made:

### Level Editor
- Added filled white box for Raster Levels to match Toonz Raster Levels.
- Added `Preferences` -> `Colors` -> `Level Editor Box Color` 
   - Default color matches `Viewer BG Color`
- Viewer title bars changed from `Level: A.1 ...` to `[LEVEL]: A.1 ...` to highlight editing mode
- Project information removed form title bar to reduce visual noise

### Scene Editor
- Viewer title bars changed from `Scene: xyz ...` to `[SCENE]: xyz ...` to highlight editing mode
- Project information removed form title bar to reduce visual noise
- Added Zoom % to ComboViewer's title bar

### Xsheet viewer
- Project information removed from title bar to reduce visual noise

### DIFF
--- a/toonz/sources/toonz/sceneviewer.cpp
+++ b/toonz/sources/toonz/sceneviewer.cpp
@@ -1174,7 +1174,8 @@ void SceneViewer::drawCameraStand() {
 
     TImageP image = tool->getImage(false);
 
-    TToonzImageP ti = image;
+    TToonzImageP ti  = image;
+    TRasterImageP ri = image;
     if (ti) {
       TRect imgRect(0, 0, ti->getSize().lx - 1, ti->getSize().ly - 1);
       TRectD bbox = ToonzImageUtils::convertRasterToWorld(imgRect, ti);
@@ -1184,8 +1185,22 @@ void SceneViewer::drawCameraStand() {
       if (ToonzCheck::instance()->getChecks() & ToonzCheck::eBlackBg)
         imgRectColor = TPixel::Black;
       else
-        imgRectColor = TPixel::White;
+        imgRectColor = Preferences::instance()->getLevelEditorBoxColor();
       ToolUtils::fillRect(bbox * ti->getSubsampling(), imgRectColor);
+    } else if (ri) {
+      TRectD bbox = ri->getBBox();
+      bbox.x0 -= ri->getBBox().getLx() * 0.5;
+      bbox.x1 -= ri->getBBox().getLx() * 0.5;
+      bbox.y0 -= ri->getBBox().getLy() * 0.5;
+      bbox.y1 -= ri->getBBox().getLy() * 0.5;
+
+      TPixel32 imgRectColor;
+      // draw black rectangle instead, if the BlackBG check is ON
+      if (ToonzCheck::instance()->getChecks() & ToonzCheck::eBlackBg)
+        imgRectColor = TPixel::Black;
+      else
+        imgRectColor = Preferences::instance()->getLevelEditorBoxColor();
+      ToolUtils::fillRect(bbox * ri->getSubsampling(), imgRectColor);
     }
     glPopMatrix();
   }
@@ -1722,7 +1737,7 @@ double SceneViewer::projectToZ(const TPointD &delta) {
   GLint viewport[4];
   double modelview[16], projection[16];
   glGetIntegerv(GL_VIEWPORT, viewport);
-  for (int i = 0; i < 16; i++)
+  for (int i      = 0; i < 16; i++)
     projection[i] = (double)m_projectionMatrix.constData()[i];
   glGetDoublev(GL_MODELVIEW_MATRIX, modelview);
 
@@ -1884,9 +1899,8 @@ void SceneViewer::zoomQt(bool forward, bool reset) {
     if (reset || ((m_zoomScale3D < 500 || !forward) &&
                   (m_zoomScale3D > 0.01 || forward))) {
       double oldZoomScale = m_zoomScale3D;
-      m_zoomScale3D =
-          reset ? 1
-                : ImageUtils::getQuantizedZoomFactor(m_zoomScale3D, forward);
+      m_zoomScale3D       = reset ? 1 : ImageUtils::getQuantizedZoomFactor(
+                                      m_zoomScale3D, forward);
 
       m_pan3D = -(m_zoomScale3D / oldZoomScale) * -m_pan3D;
     }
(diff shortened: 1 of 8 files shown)


konero:
Very nice! Thank you.

shun-iwasawa:
Thank you @manongjohn , please let me review this.

toonz/sources/toonz/sceneviewer.cpp
```diff
@@ -1172,21 +1174,6 @@ void SceneViewer::drawCameraStand() {
...
-    if (ti) {
-      TRect imgRect(0, 0, ti->getSize().lx - 1, ti->getSize().ly - 1);
-      TRectD bbox = ToonzImageUtils::convertRasterToWorld(imgRect, ti);
-
-      TPixel32 imgRectColor;
-      // draw black rectangle instead, if the BlackBG check is ON
-      if (ToonzCheck::instance()->getChecks() & ToonzCheck::eBlackBg)
-        imgRectColor = TPixel::Black;
-      else
-        imgRectColor = TPixel::White;
```
shun-iwasawa: Sorry but I think I misunderstood your proposal.
I thought that you were planning to just make the box color editable, instead of replacing the box by whole background.

We do need the box itself to recognize the border of the image.
Also the background must become black when the "Black BG" option is activated.

Can you please reintroduce the box for Toonz Raster and Raster levels, replacing `TPixel::White` of this line with `Preferences::instance()->getLevelEditorBgColor()`?

manongjohn: Later in the discussion it was suggested (if I understood correctly) to just have a different background color for the Level Editor.  I felt it made sense and still allowed Studio Ghibli to have their white background.  The bounding box is still there for both Toonz Raster and Raster levels so you know where the image will be and can be drawn.

Sorry, but I don't understand the original intention of just having TR having a filled white box and why this doesn't work for them. I do not need an explanation other than this is just another case of how it has to be.

I will go back to my original proposal and make just the box color changeable.

shun-iwasawa: > Sorry, but I don't understand the original intention of just having TR having a filled white box and why this doesn't work for them.

The reason why I inserted the white box in Ghibli edition was to keep consistency with their conventional Ink&Paint software, Toonz4.6. IMO the white background has its roots in that the drawing on white paper was put under celluloid on Ink&Paint work.
I implemented only for Toonz Raster because Ghibli does not use other level types for animation levels.

It is true that OT shows raster bounding box, but I think for them it is more of visual cue for grasping the placement of each level on compositing scenes. On ink & painting they'd prefer to have more prominent boundary than light-gray and thin border line. The filled box was one of the simple solutions for us.
Everyone has different opinions on UI... I'm sorry to trouble you.

artisteacher: What if the bounding box boundary was made more prominent?

hayoko:
Ah, for me I was talking about an option to the box, so we can have something dull like what was mentioned in #1061.

image

toonz/sources/toonz/sceneviewer.cpp
shun-iwasawa: Hmm, of course I must confirm to my colleagues.., but I think the prominent border line may become noisy on compositing and if we make the line thicker we need to be careful to prevent the line from hiding the edge pixels of the image.

gab3d: @shun-iwasawa 
I find the virtue of consistency in @manongjohn solution, while I understand what you are saying about the needs and habits of Ghibli's staff.

What if one of the following ideas was implemented for Toonz Raster and Raster levels, so that Ghibli's staff could still have a stronger image size cue, while preserving the same background for all level types?:
- The **Level Editor BG Color** could be used as the background *inside* the limits of the level image, and the **Viewer BG Color** used *outside* of those limits.
- Alternatively, a new "image size masking" feature (with a % fade value), could be implemented, that would allow to darken the outside of the image limits in Level Edit mode, so there would be much more contrast between inside and outside areas of the level image.

shun-iwasawa: @gab3d 

> - The Level Editor BG Color could be used as the background inside the limits of the level image, and the Viewer BG Color used outside of those limits.

Please correct me if I am wrong. It seems that the first option is exactly talking about "the box" which is currently implemented (by me) as white-only on Toonz-Raster-only. Right?
If so, only the first option would be sufficient for us.

gab3d: yep, I think both options would achieve the goal of more clearly define the limits of the level image, but are somewhat different one from another.
The first one is very similar to what you had implemented, other than the Level Editor BG Color is now customizable.
The second one attains the contrast in a different way (by fading a mask over BG color) and has the benefit of not depending on the Viewer BG Color at all to produce the contrast, so if a user decides to change the Viwer BG color to the same color as the Level Editor BG Color, the image limits would still be there correctly.

manongjohn:
Changed to match my original proposal. PR description updated.

As for masking the Level Editor Viewer (outside of box) color. Can we agree to just darken the color by some fixed % of the normal Viewer BG Color? This would be another indicator that they've switched to Level Edit mode.

hayoko:
> As for masking the Level Editor Viewer (outside of box) color. **_Can we agree to just darken the color by some fixed % of the normal Viewer BG Color?_** This would be another indicator that they've switched to Level Edit mode.

I'd suggest no, because you need a neutral tone for true color design, otherwise the artist may pick values to compensate for the color shift caused by the surrounding color, this is known as the [Chubb illusion](https://en.wikipedia.org/wiki/Chubb_illusion).

> image
> _Colors can appear different depending on their surrounding colors and shapes. The two small squares have exactly the same color, but the right one looks slightly darker. ([From Wikipedia](https://en.wikipedia.org/wiki/Color))_

I think some other people's suggestions have needlessly over-complicated this (no offence), I wasn't aware that knowing if you're in level edit mode was a problem, I don't think it is, you can just enable the table [_View > Table_] in the viewer which disappears when you enter level edit mode already and there's nothing you could add that would be more obvious than that.

manongjohn:
Good point about the Chubb Illusion. Probably best to leave it as is.

While some changes were made to somewhat address the new user confusion over the edit modes, this PR is also trying to make it a bit more consistent between the raster level types and provide some additional control over the box for both new and existing users.

artisteacher:
> you can just enable the table [View > Table] in the viewer which disappears when you enter level edit mode already and there's nothing you could add that would be more obvious than that.

@hayoko New users often don't realize that there is a level edit mode - hence the confusion. Even when you know what's going on, the table in itself may not make the editing mode clear. If you are using camera view or are zoomed in enough, the table is not visible.

Anything that makes level editing mode more consistent and clear is a positive change. Thanks @manongjohn for working on this!

shun-iwasawa:
For the sake of consistency when editing vector levels there could be the entire background filled with `Level Editor Box Color` or a box filled with `Level Editor Box Color` with the size of current camera. The latter idea may mislead users into thinking that there is a boundary of the image, but it can also be a yardstick to know the scale of drawing when it is put on the scene.
Anyway I think it can be held for later consideration.

LGTM Thank you @manongjohn !

PR #2883, opened 2019-11-15
Fix tool option on Command bar crash

### LINKED ISSUE #2881: Crash when turn out Global Key in Plastic Tool

ezgif-1-f375a09c6753

### Steps to Reproduce

1. Turn on Animate mode (Plastic Tool)
2. Turn on Global Key
3. Turn out Global Key

### System Information
* **OpenToonz Version**: 2019-11-13 23:11:15 UTC

manongjohn:
This PR fixes #2881 

To prevent crash, added test for a valid Tool Handle before using it similar to other options that also accessed the tool handle before using it.

I updated other tool options that had unchecked tool handle usage just in case.


RodneyBaker:
Thank you @manongjohn 
LGTM

PR #2884, opened 2019-11-16
Fix folded columns slipping

shun-iwasawa:
This PR will fix the folded xsheet columns "slipping" when you reorder / insert / remove columns.
Now the folded columns will keep their contents unchanged through column reordering.

### DIFF
--- a/toonz/sources/include/toonz/columnfan.h
+++ b/toonz/sources/include/toonz/columnfan.h
@@ -93,6 +93,9 @@ of column identified by \b col.
 
   void saveData(TOStream &os);
   void loadData(TIStream &is);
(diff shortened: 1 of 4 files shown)


RodneyBaker:
@shun-iwasawa I ran into an issue with folded columns still slipping after performing an Undo.

I'm currently looking deeper but want to get that on your radar.
The basic setup I was testing at the time was copying columns and in the process of resetting the for additional testing (via Undo) I noticed some strangeness that I haven't yet been able to confirm.

RodneyBaker:
Verified problem and that this PR corrects the issue.
LGTM

PR #2885, opened 2019-11-17
Fix playback skipping first frame

### LINKED ISSUE #2879: Playback basically skips the first frame.
When playing back, it basically instantly goes to the second frame in the playback range.  

To test:
Create a scene with a handful of frames.
Set the fps to 1
Playback should stay on frame 1 for a second, but it skips straight to frame 2.

This is especially problematic if rewind after playback is disabled.

manongjohn:
This PR fixes #2879

When playback started, it immediately changed to the next frame then waited. Modified logic not to change frame on initial start so it just waits for the 1st frame.

NOTE:  I only changed line 194. The rest of the changes were from beautification


RodneyBaker:
LGTM

Issue #2581, opened 2019-03-28
Levels disappear 


artisteacher:
### Issue Summary

Occasionally levels are missing when a scene is opened, after starting OpenToonz. When this happens, the level is still associated with the scene and the column displays with red text. Looking in the browser, the actual file is no longer there. The "backup animation levels when saving" preference is enabled and sometimes the backup is missing too. In case it makes a difference, this is using an alternate project location (usually Documents) and using all of the default folders (drawings, extras & etc.).  Most likely this happens after a crash, but it is infrequent and so is hard to track.

### System Information
* **OpenToonz Version**: 1.3
* **Operating System**: macOS 10.13.6
* **CPU**: 
* **Memory**: 
* **Graphics Card**: 
* **Graphics Tablet**:

RodneyBaker:
A likely problem with missing files and where the files are not located inside the stuff directory is that the source files have been moved.  Short of refraining from using files outside of the project directory structure it is not clear how best to proceed to minimize chance for loss of referenced files. 

One option might be to create a 'Secure Mode' in Preferences designed to intentionally restrict the user from performing risk without expressly acknowledging that risk.  This would be a long term developmental track with a primary goal to reduce variation.

With regard to Levels themselves disappearing there are several possibilities:

- Source was referenced in saved Scene but never saved itself
- Source was overwritten by another file an then subsequently updated out of scope (therefore no longer being the same name as the original sourced file
- Source was moved
- Source was corrupted
- Program, Project, Scene was moved relative to source file (source is still in same location but things that once referenced it have changed or moved
- Other things not itemized here

artisteacher:
I agree that sometimes it must be user error due to things being moved or from saving the scene without saving levels. Sometimes though, there does seem to be something else going on. I know that I almost never move things from a project and I always save all (and have auto-save enabled). Perhaps it's the timing of saving coinciding with a crash. Does the scene file save before the levels with save all?

RodneyBaker:
Levels can save from a number of different triggers.
The Save All is just the most efficient one observed thus far.

When considering loss of files I focus on risk.
We gained a great deal of freedom when Opentoonz could save outside of the stuff directory but it came with some added risk in that files outside of the stuff directory can have many things happen to them which will cause problems.

I don't see this as being in the realm of user error.
It is a referencing and processing problem.
It's not unlike all the dead internet links that occur after time.

What we could do is categorize the risks and attack them from that angle.
[...]

PR #2888, opened 2019-11-20
Fix file loss issues

### LINKED ISSUE #2363: Shift + Drag/Drop into OT with Load... Deletes files
### Issue Summary
Using Shift + Drag/Drop to bring files into OpenToonz brings up the Import/Load dialog.
This 'moves' the file into OpenToonz directory structure.
Opting for Import works as expected.
Opting for 'Load' can easily lead to loss of the files.

### Steps to Reproduce
Drag/Drop one or more images (PNG is good option) into OpenToonz and use import or load from the dialog.  This should work as desired.
Use Shift + Drag/Drop to do the same thing also using Import option.
This should work as desired.
Use Shift + Drag/Drop and use the Load option and note that the original is moved to Recycle Bin (deleted) and the image is place in OpenToonz but a Save All will not place the file in the /extras directory.  After reopening the Scene (via Recent Used Scenes listing or other method) the link will be displayed as red (broken link)

### Expected Results
Shift + Drag/Drop should move the file(s) to the target directories (i.e. copy the file(s) and then delete the originals upon successful copy) and then update the Level reference to the new location.

or

Load should be refactored or made to not be an option while 'moving' via Shift + Drag/Drop.

 
### Actual Results
Files are not moved but sent to the Recycle Bin (deleted!).

### System Information
* **OpenToonz Version:  Most current nightly release as of 20181105 
* **Operating System:  Win10 

Thanks!

manongjohn:
This PR does the following to prevent accidental file loss:

1) Some levels use an interim temp file to hold changes while saving. This temp file eventually replaces the original file.  If for some reason the temp file is not created, the logic still deletes the original file and then finally fails to save because the missing temp file can't be renamed.

Added a check to verify the temp file exists before deleting the original. If not found, an exception is thrown and the save fails.

This may address  #2581

NOTE: At present time, failed level saves only notify the user if they did a "Save Level". "Save All" hides the failure notification. This should be addressed separately.

2) On Windows systems, Shift+Drag/Drop of a file does a Move file which basically copies to new destination and then removes the old. When dragging a file from outside OT (explorer) and dropping into OT, OT accepts the drop, but then Windows deletes the original file. If the user does not modify the level and save, the file is lost when quitting

Added logic to convert Move-Drops to Copy-Drops in various places in OT.  The following were impacted:

   - Color Models
   - Tree Item viewers (details of a Tree item), like the right pane in File Browser and Scene Cast
   - Flip Books
   - Scene Viewers (ComboViewer and regular Viewer)
   - Xsheet/Timelines
   - Palette Viewers

This fixes #2363

3) If you attempted a move-drop of a non-studio palette into the Studio Palette viewer, the drop does nothing, but the original file is deleted.

The Studio Palette only allows you to drag/drop Global or Project Palettes already in the Studio Viewer. Any other palette files, like level palettes, do nothing. I added logic to block the ability to drop non-studio palettes into the Studio Palette Viewer.

4) Attempting to convert a file in the File Browser back to the same type (png -> png), results in the file being lost

Added logic to skip converting a file if the source and destination files names are exactly the same.

5) If you drag a file from the right pane of the File Browser back to the same directory in the File Browser tree (left pane), a copy of the original file is created as a numbered file and the original file is deleted. (A.png -> A_1.png)

Added logic to ignore moving file if the source's parent directory is the same as the destination directory.

### DIFF
--- a/toonz/sources/toonz/colormodelviewer.cpp
+++ b/toonz/sources/toonz/colormodelviewer.cpp
@@ -123,7 +123,10 @@ void ColorModelViewer::dragEnterEvent(QDragEnterEvent *event) {
     std::string type = fp.getType();
     if (type == "scr" || type == "tpl") return;
   }
-  event->acceptProposedAction();
+  // Force CopyAction
+  event->setDropAction(Qt::CopyAction);
+  // For files, don't accept original proposed action in case it's a move
+  event->accept();
 }
 
 //-----------------------------------------------------------------------------
@@ -138,7 +141,10 @@ void ColorModelViewer::dropEvent(QDropEvent *event) {
       loadImage(fp);
       setLevel(fp);
     }
-    event->acceptProposedAction();
+	// Force CopyAction
+	event->setDropAction(Qt::CopyAction);
+	// For files, don't accept original proposed action in case it's a move
+	event->accept();
   }
 }
 
--- a/toonz/sources/toonz/convertpopup.cpp
+++ b/toonz/sources/toonz/convertpopup.cpp
@@ -127,7 +127,12 @@ void ConvertPopup::Converter::run() {
 
     m_saveToNopaintOnlyFlag = false;
 
-    if (TSystem::doesExistFileOrLevel(dstFilePath)) {
+    if (dstFilePath == sourceLevelPath) {
+      DVGui::info(tr("Level %1 converting to same file format; skipped.")
+                      .arg(levelName));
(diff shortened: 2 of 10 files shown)


artisteacher:
This will be very helpful - thanks @manongjohn!

gab3d:
appreciated!!

RodneyBaker:
Thanks @manongjohn 
LGTM

PR #2890, opened 2019-11-20
Warn on failing level save during Save All

manongjohn:
When you `Save Level` and there is an error saving, OT will immediately notify the user there was a problem saving.

When using `Save All` or `Save All Levels`, there is no warning if a level failed to save.  This can result in loss of work.

This PR will add logic to display a warning for every level that fails to save during a `Save All` or `Save All Level` operation, provided the failure actually throws an exception.

### DIFF
--- a/toonz/sources/toonzlib/sceneresources.cpp
+++ b/toonz/sources/toonzlib/sceneresources.cpp
@@ -237,9 +237,7 @@ void SceneLevel::save() {
         try {
           m_sl->save(actualFp, oldActualPath);
         } catch (...) {
-          DVGui::warning(
-              QObject::tr("Can't save") +
-              QString::fromStdWString(L": " + actualFp.getLevelNameW()));
+          throw;
         }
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I'm curious what the user might face if there are a lot of level fails due to high number of levels being used in a scene.  After the initial acknowledgement could the user be prompted with an option to acknowledge all remaining notifications?

Ideally as users we do want to know each and every failure.
I'm just concerned with users getting caught in one of those very long loops requiring interaction/acknowledgement.

Perhaps a 'Cancel All Remaining Error Messages' option?

manongjohn:
I'll look at building a list of failed saves and displaying at the end of the save process.

RodneyBaker:
I'm not able to test via triggering by forcing failed levels but all checks out otherwise.
LGTM and thanks for the additional solution of error notification at end of saving @manongjohn

PR #2892, opened 2019-11-22
Fix initial Style Editor palette

### LINKED ISSUE #2385: Style Editor tabs missing on open
### Issue Summary

When first opening OT, only the color and settings tab are visible in the Style Editor, even if there is an active/selected level & styles associated with that level.

screen shot 2018-11-20 at 11 15 05 am
### Steps to Reproduce

Open an existing scene when you open OT.
### Expected Results

For the tabs to be visible if a level is active/selected.
### Actual Results

Only color and settings tabs are visible. To see the tabs, you have to switch to another style.
### System Information
* **OpenToonz Version**: nightly build & 1.2.1
* **Operating System**: MacOS 10.13.5, 10.14
* **CPU**: 
* **Memory**: 
* **Graphics Card**: 
* **Graphics Tablet**:

manongjohn:
This PR fixes #2385

When loading a scene, this change will set the initial palette to the palette of the level in Column 1, if there is one.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1884,6 +1884,10 @@ bool IoCmd::loadScene(const TFilePath &path, bool updateRecentFile,
   app->getCurrentScene()->notifyNameSceneChange();
   app->getCurrentFrame()->setFrame(0);
   app->getCurrentColumn()->setColumnIndex(0);
+  TPalette *palette = 0;
+  if (app->getCurrentLevel() && app->getCurrentLevel()->getSimpleLevel())
+    palette = app->getCurrentLevel()->getSimpleLevel()->getPalette();
+  app->getCurrentPalette()->setPalette(palette);
   app->getCurrentXsheet()->notifyXsheetSoundChanged();
   app->getCurrentObject()->setIsSpline(false);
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
The general operation of this PR works fine and it works well for me personally.
I like how users can leverage this to load default palettes via the palette they use in Column 1.
This also makes me think we can now more trivially create a default 'empty' scene pre-supplied with an initial frame (or set of empty frames) and a predesignated working palette.

I'm not sure studios will want Column 1 to take precedence for their working environment so... over to you @shun-iwasawa.

manongjohn:
Right now, if there is no level in column 1 at scene creation (new or loaded), i force no palette in Style Editor to show. I can make it go back to showing the CLEANUP palette if this is an issue.

Can someone explain how the CLEANUP palette is even brought up in the Style Editor normally? Once you click on a level, i found no other way to bring it back. The only reference to a CLEANUP palette I found was in the Cleanup screens (not room) and it doesn't use the Style Editor unless i missed something.

RodneyBaker:
After more consideration and testing I don't think my previous concerns are valid given the nature of Style Editor default palette and don't want to delay the gain of initially inheriting the Column 1 palette.   

LGTM

artisteacher:
So much better! Thanks!

Issue #2794, opened 2019-09-20
Output Settings window is not scalable and scrollable


2dvision:
### Issue Summary

It is not important, just for the report.

If you expand all section in **Output Settings** window and if you have a small screen, You can't see all this window easily.

### Screenshot or Video Reference

![Output Setting wondow is not scalable and scrolable](https://user-images.githubusercontent.com/16509592/65342856-dee5ab80-dbe8-11e9-8be3-16434ec418d6.jpg)

RodneyBaker:
Thanks for reporting this. 
Hopefully this will be fixed as part of the UI scrub going into v1.4.

This particular issue could likely be resolved by allowing the Output settings window to be scaled and resized.  There is something of a problem with scaling and resizing though that would still leave an issue.

Another approach (in addition to allowing rescale/resizing) might be to place the Other Settings portion of the window to the right of the Camera Settings and File Settings area.  This would result in a dialog box wider than tall which would resolve the deeper issue of not being able to see all the options at the same time.

gab3d:
the scheme of Preferences window couls also be used. with categories to the left and options to the right.

or, alternatively, tabs could be used also...

there are several possibilities, really, for accomodating those options.

RodneyBaker:
@gab3d This might be a good time to discuss some of our collective approach to development.

Your comment is more of a suggestion to redo the Output Settings dialog... to fix it for good! :)
Such should be an ongoing development topic and we need a designated place for those.... but here is a good place too.

Most of what I focus on in my reviews is short term fixes that can be implemented rapidly (or at least more rapidly than major UI overhauls).  This is largely triage but can carry us through until the larger development is locked in and pursued.

The initial report is the triggering that let's us know that 'something isn't right here'.
Far from being a bad thing report of issues are great opportunities.
My interest is in how we can rapidly address the surface issue (resolving the users problem) while triggering the deeper look into underlying issues and opportunities.

So far so good!  :)

gab3d:
@RodneyBaker 
yeah, I get it and agree.
mine, was just a comment.

RodneyBaker:
@gab3d I mostly posted that for others looking in that don't know the process.
I know you know... but this is the type of thing I wish I knew when I was first getting started.

Heck, maybe everyone already knows the process and can clue me in.  :)

hayoko:
Just add a QScrollArea? 🤔

RodneyBaker:
Relates tangentially to #2566

PR #2894, opened 2019-11-22
Overhaul the Preferences

shun-iwasawa:
**PLEASE NOTE: I need this PR to be carefully tested before merging, as I may miss something!**

This refactoring PR will overhaul the preferences.
The source code for the preferences was very messy, having separate similar functions for every variables' I/O.
I tried to clear them up by putting almost all the preference values in one container (Id-QVariant map). Now, I managed to reduce 2,000+ lines of code!

I roughly tested and find that the change has (hopefully) no bad side effect, but I would like this to have further tests from other reviewers before merging.

### DIFF
--- a/toonz/sources/toonzlib/preferences.cpp
+++ b/toonz/sources/toonzlib/preferences.cpp
@@ -40,11 +40,7 @@ typedef std::vector<LevelFormat> LevelFormatVector;
 
 //-----------------------------------------------------------------
 
-const char *s_bool[2] = {"0", "1"};
-
-const char *s_show0ThickLines = "show0ThickLines",
-           *s_regionAntialias = "regionAntialias",
-           *s_levelFormats    = "levelFormats";
+const char *s_levelFormats = "levelFormats";
 
 const char *s_name = "name", *s_regexp = "regexp", *s_priority = "priority";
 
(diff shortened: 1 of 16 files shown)


RodneyBaker:
Travis seems to have failed building on the last two PRs.  
It's not clear if that is due to the PRs themselves or Travis itself.

manongjohn:
There is an issue with the macos builds that started yesterday. It's not using the brew installed GLEW library and the default stuff has changed somehow.

I noticed it on my local build environment that emulates Travis. Looking into it as it will affect OTX builds.

toonz/sources/toonzlib/preferences.cpp
```diff
@@ -563,675 +319,491 @@ Preferences::Preferences()
...
 }
 
 //-----------------------------------------------------------------
 
-void Preferences::enablePreviewAlwaysOpenNewFlip(bool on) {
-  m_previewAlwaysOpenNewFlipEnabled = on;
-  m_settings->setValue("previewAlwaysOpenNewFlip", on ? "1" : "0");
+PreferencesItem &Preferences::getItem(const PreferencesItemId id) {
+  assert(m_items.contains(id));
+  if (!m_items.contains(id)) return PreferencesItem();
```
manongjohn: Linux and macOS builds are failing here with this message

/Users/travis/build/opentoonz/opentoonz/toonz/sources/toonzlib/preferences.cpp:668:37: error: non-const lvalue reference to type 'PreferencesItem' cannot bind to a temporary of type 'PreferencesItem'
  if (!m_items.contains(id)) return PreferencesItem();

shun-iwasawa: Thank you for the notice, I've made the same mistake again.
I removed the problematic line.

RodneyBaker:
I've been testing this PR (from user POV via appveyor portable) as I get the chance and haven't noticed anything odd, out of place or non-functioning.  

Off topic:  There are a number of things that need to be addressed with Preferences that would fit more into a second 'Reorganize the Prefererences' PR.  Most of these fit into general refinement category although there are a few cases of things meant to work that do not; example:  the cursor changes when hovering over the popup window borders and corners to suggest Preferences window can be resized but window cannot be resized/rescaled.

gab3d:
@RodneyBaker 
...but the Preferences window can be resized, at least on Windows. 🤔

manongjohn:
I can resize Preferences. 

I did notice sometimes the cursor on mouse overs doesn't change to show a pane's edge can be moved until after you try to move it. I think it happened with Style Editor and Level Editor, the 2 frames that are designed to keep size when main window resizes. I haven't seen that with other windows though.

RodneyBaker:
Interesting.
Apparently I was on the Animation tab and attempting to make the window smaller and it is/was constrained to be as large as the contents of other tabs.  

Attempt to make the Preferences window as small as possible while the Animation tab is open and see how small you can make the window.  At a guess I'd say there is a minimum size allowable to accommodate the other tabs.

shun-iwasawa:
>Attempt to make the Preferences window as small as possible while the Animation tab is open and see how small you can make the window. At a guess I'd say there is a minimum size allowable to accommodate the other tabs.

Yes, it's an expected behavior of `QStackedWidget` used for the stacked pages in the popup... You need to nest it in `QScrollArea` to make it shrinkable smaller than the minimum size.
It's similar to the issue #2794 regarding the output settings popup.

I further tested this PR in our studio and confirmed that it works OK with 3DLUT settings, which is quite special among others. 
Thank you for testing, I'll merge this.

manongjohn:
@shun-iwasawa Somehow this PR is not keeping History correctly. It only keeps the last action. Can only undo the last action.

shun-iwasawa:
@manongjohn I'll check it right away!

toonz/sources/toonzlib/preferences.cpp
```diff
@@ -219,287 +208,28 @@ bool Preferences::LevelFormat::matches(const TFilePath &fp) const {
...
-  getValue(*m_settings, "viewShrink", m_shrink);
-  getValue(*m_settings, "viewStep", m_step);
-  getValue(*m_settings, "blanksCount", m_blanksCount);
-  getValue(*m_settings, "askForOverrideRender", m_askForOverrideRender);
-  r = 255, g = 255, b = 255;
-  getValue(*m_settings, "blankColor.r", r);
-  getValue(*m_settings, "blankColor.g", g);
-  getValue(*m_settings, "blankColor.b", b);
-  getValue(*m_settings, "undoMemorySize", m_undoMemorySize);
-  setUndoMemorySize(m_undoMemorySize);
```
shun-iwasawa: Ah! I missed this line in ctor!

shun-iwasawa:
@manongjohn Fixed the problem. Thank you so much for the notice!

PR #2897, opened 2019-11-22
Allow note frame creation on Animation Sheet

### LINKED ISSUE #2893: Note level malfunction when "Use Xsheet as Animation Sheet" option on.
When the "Use Xsheet as Animation Sheet" option is on, and you create a Note level. 
The first cell is automatically created, and you can double click on it to input text, but you can't create new cells in that column, while the aforementioned option remains active.

Considering that now "Use Xsheet as Animation Sheet" option is the default mode, I think this should be fixed before 1.4 release.

Thanks!!

manongjohn:
This PR fixes #2893

Added exception to allow empty cell rename on doubleclicks for Note levels when `Use Xsheet as Animation Sheet` is enabled so that new note frames can be created wherever needed in that column

Unless otherwise requested, `Enable auto-stretch frame` does not auto extend the prior note frame to meet the newly created frame like it normally would on drawn levels.

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -2988,8 +2988,12 @@ void CellArea::mouseDoubleClickEvent(QMouseEvent *event) {
   // in modalita' xsheet as animation sheet non deve essere possibile creare
   // livelli con doppio click: se la cella e' vuota non bisogna fare nulla
   if ((Preferences::instance()->isAnimationSheetEnabled() &&
-       m_viewer->getXsheet()->getCell(row, col).isEmpty()))
-    return;
+       m_viewer->getXsheet()->getCell(row, col).isEmpty())) {
+    TXshColumn *column = m_viewer->getXsheet()->getColumn(col);
+    if (!column ||
+        column->getColumnType() != TXshColumn::ColumnType::eSoundTextType)
+      return;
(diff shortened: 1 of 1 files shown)


gab3d:
that was fast! 😄 
thank you!

artisteacher:
This is definitely needed. I do think other level types would benefit from the same behavior, unless it interferes with something important.

manongjohn:
There is a comment (in Italian) that says it shouldn't be allowed, but doesn't give reason why. Not sure I want to challenge that.

As it is, when I do it on a regular level and try to change the frame on an exposed frame, held or not, it doesn't do what you would expect. The frame gets renumbered in the Level Strip. The xsheet gets updated differently depending on if you have sync level strip drawings to xsheet turned off or on. All probably bugs that should be reviewed and fixed before the next release.

RodneyBaker:
LGTM

DarrenTAnims:
Q. When "Use Xsheet as Animation Sheet" isn't enabled, what happens when you try to create a new note? If nothing, wouldn't it make sense to always allow creating a new note on a frame in a note level? I'm not sure what else you would want to do when you double click in a cell.

manongjohn:
In other modes you can enter an empty cell by double clicking it. That is normally disabled in Animation Sheet.  This PR relaxed that restriction just for note levels.

PR #2900, opened 2019-11-23
Fix macOS brew installed glew issue

manongjohn:
This PR fixes the current cause of macOS Travis build failures.

When installing glew libraries using brew, it is creating the symbolic link /usr/local/lib/cmake/glew which points to glew cmake information. This symbolic link is causing cmake to incorrectly detect the actual installed glew libraries and later fails to build OT because of it.

As a temporary workaround, I updated the script to detect and delete the symbolic link after installing glew through brew.


manongjohn:
All builds successful. I'm going to merge this now to prevent future PR builds from having failures.

PR #2903, opened 2019-11-25
Allow Selection Tool copy when level is not editable

### LINKED ISSUE #2901: "Select + Copy" from PSD images or convert to Toonz Raster level?
@shun-iwasawa @manongjohn 
this is not a bug report, but a quick question and possibly an idea about a current limitation of OT.

While helping in migrating a project that some artists had previously started using Photoshop (where they had drawn several level frames as layers) we imported the PSD layers as frames into a single Xsheet column, so far so good, but then they tried to actually use that drawings to edit them and paint them. Naturally PSD is a read-only format, so I suggested copying and pasting or converting the whole level to Toonz Raster, to achieve it, but then found out neither of the two options are currently really possible.

1. As selecting in a read only drawing is disabled, we could not attempt to select and copy the PSD content to later paste it into a new TR level.

2. Converting that whole column into a TR level isn't available either, which could've been massively useful too. I've did manage to convert the drawings into a vector level and then from there to a TR level, but in the process the drawings were altered (as expected), so it wasn't a viable path either...

So the question being: Is there a real issue in converting read only PSD levels into Toonz Raster or standard Raster levels? or is it just an absent feature that could be added in a relatively easy way?

Thank you for listening!!

manongjohn:
This PR resolves #2901 .

The Selection Tool is currently unavailable whenever the column is locked or the level is read-only. This change will make the Selection Tool available under these conditions.

The change include the following:

- Enables the Selection Tool for Vector, Toonz Raster and Raster levels under the following conditions:
   - The level is marked as read-only because of file permissions or is a PSD/GIF/MP4/WEBM file
   - The column in the xsheet/timeline is locked
   - The column is not editable because it is a mesh-deformed column
- Copying a selection is the only thing allowed.  Moving, deforming, cutting and pasting selections into are blocked.
- The Selection Tool toolbar options will become unavailable.
- The selection box will lose the deformation handles and center pivot.

image

### DIFF
--- a/toonz/sources/include/tools/rasterselection.h
+++ b/toonz/sources/include/tools/rasterselection.h
@@ -143,6 +143,8 @@ Can be different from getSelectionBound() after a free deform transformation. */
   void pasteSelection();
 
   bool isTransformed();
+
+  bool isEditable();
 };
 
 #endif  // RASTER_SELECTION_H
--- a/toonz/sources/include/tools/strokeselection.h
+++ b/toonz/sources/include/tools/strokeselection.h
@@ -66,6 +66,8 @@ public:
 
   void setSceneHandle(TSceneHandle *tsh) { m_sceneHandle = tsh; }
 
+  bool isEditable();
+
 public:
   // Commands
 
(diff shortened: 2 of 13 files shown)


gab3d:
Thank you @manongjohn , this is awesome!! 😄

RodneyBaker:
Glad to have the ability to select parts of an image (read only in this case) and look forward to other PRs that resolve some of the other fringe issues with ffmpeg loaded images.

LGTM

PR #2904, opened 2019-11-25
Mark PSD and ffmpeg loaded files as readonly in Level Strip

manongjohn:
Normally, read-only files (based on file permissions) are denoted in the Level Strip with a forbidden icon in the lower left corner of each frame.

This PR adds the forbidden icon for PSD/GIF/MP4/WEBM files also so users know they are read-only files.

image


RodneyBaker:
I hadn't realized that read-only icon was already there.
Learn something new every day.  :)

LGTM

Perhaps in the future we can adjust the color of that icon to make it red.

manongjohn:
It's rare to see that icon in the level strip. I think its mostly used when you are using SVN and files are checked in.

PR #2907, opened 2019-11-28
Add Create Blank Drawing command

manongjohn:
Requested by a few users (and questioned why there is no command for it  by many others), this PR adds a command to allow a user to create a blank drawing.

This command only works on the Xsheet/Timeline and factors into account Autocreation mode and Auto-stretch settings.

If Autocreation mode is `Disabled`, the command will temporarily change the setting to `Enabled` to create the frame according to those settings and then disables that preference again.

Multiple blank drawings can be created by selecting any range of cells on the xsheet/timeline.

Level/frame creation rules still apply meaning blank drawings cannot be created where they wouldn't normally be able to.

### DIFF
--- a/stuff/profiles/layouts/rooms/Default/menubar_template.xml
+++ b/stuff/profiles/layouts/rooms/Default/menubar_template.xml
@@ -212,6 +212,7 @@
         </menu>
         <separator/>
         <command>MI_Autorenumber</command>
+        <command>MI_CreateBlankDrawing</command>
         <command>MI_Duplicate</command>
         <command>MI_MergeFrames</command>
         <command>MI_CloneLevel</command>
--- a/stuff/profiles/layouts/shortcuts/defopentoonz.ini
+++ b/stuff/profiles/layouts/shortcuts/defopentoonz.ini
@@ -110,6 +110,7 @@ MI_ConvertToToonzRaster=
 MI_ConvertToVectors=
 MI_ConvertVectorToVector=
 MI_Copy=Ctrl+C
+MI_CreateBlankDrawing=Alt+D
 MI_CursorOutline=
 MI_Cut=Ctrl+X
 MI_DeactivateAllColumns=
--- a/stuff/profiles/layouts/shortcuts/otadobe.ini
+++ b/stuff/profiles/layouts/shortcuts/otadobe.ini
@@ -110,6 +110,7 @@ MI_ConvertToToonzRaster=
 MI_ConvertToVectors=
 MI_ConvertVectorToVector=
 MI_Copy=Ctrl+C
+MI_CreateBlankDrawing=F7
 MI_CursorOutline=
 MI_Cut=Ctrl+X
 MI_DeactivateAllColumns=
--- a/stuff/profiles/layouts/shortcuts/otharmony.ini
+++ b/stuff/profiles/layouts/shortcuts/otharmony.ini
@@ -110,6 +110,7 @@ MI_ConvertToToonzRaster=
 MI_ConvertToVectors=
 MI_ConvertVectorToVector=
 MI_Copy=Ctrl+C
+MI_CreateBlankDrawing=Alt+Shift+R
 MI_CursorOutline=
 MI_Cut=Ctrl+X
 MI_DeactivateAllColumns=
--- a/stuff/profiles/layouts/shortcuts/otretas.ini
+++ b/stuff/profiles/layouts/shortcuts/otretas.ini
@@ -110,6 +110,7 @@ MI_ConvertToToonzRaster=
 MI_ConvertToVectors=
 MI_ConvertVectorToVector=
 MI_Copy=Ctrl+C
+MI_CreateBlankDrawing=
 MI_CursorOutline=
 MI_Cut=Ctrl+X
 MI_DeactivateAllColumns=
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -17,6 +17,7 @@
 // TnzTools includes
 #include "tools/toolutils.h"
 #include "tools/toolhandle.h"
+#include "tools/toolcommandids.h"
 
 // TnzQt includes
 #include "toonzqt/strokesdata.h"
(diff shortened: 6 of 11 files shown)


melieconiek:
Build faster appveyor. :V

manongjohn:
Hehe.

Just realized there is 1 additional change I want to make: Allow blank frame creation on held frames in `Use Xsheet as Animation Sheet` mode in accordance to how frame creation works there.

manongjohn:
Hmm. Can remove the temp tool switching too, i think.

Is there a default shortcut I should set or leave it blank? 

Anyone know what the shortcuts are for blank frame creation on Harmony, Adobe, or Retas so I can set defaults there?

gab3d:
I think is a good idea to assign a default shortcut for this command.

RodneyBaker:
A little thought on the shortcut for 'new blank frame'...

I'm seeing Shift + N as a 'create new' although I can see where that might be problematic if working any where near a text field.

Alt + N is currently assigned to create new Level (which invokes Level popup given default level type set in Prefernces.  Alt + Shift + N would logically fit into current shortcuts although I would be tempted to personally reverse those so that Alt + N created the immediate blank frame and Alt + Shift + N allows for control of more options.  The downside of such a shortcut key combination is how it's not an easy placement for the fingers such as with Alt + N.   The logic however of adding the shift key is such that it suggests a higher level of detail or optioning.

Because changing Alt + N from the default might cause issues with users conditioned to see the results of the Level popup I'd say that is the ideal target to hit.   Perhaps the new blank frame shortcut can be assigned to Alt + Shift + N and users can swap the two assignments as they see fit.  It does not logically make sense that a Shift key would be added in order to get less options in creating a level so I'm only suggesting this as a interim measure until coordinating a change to the Alt + N default shortcut is green lit.

melieconiek:
Did you add creating a new frame in the middle of an exposure? It doesn't seem to do anything and if undone results in this:
image

Executing the command doesn't auto-stretch the frame when autocreation is set to 'Disabled' or 'Enabled' and 'Enable auto-stretch frame' is grayed out. Works as expected if autocreation is set to 'Use Xsheet As Animation Sheet.'
image
image

Thanks for this. :)

gab3d:
@RodneyBaker 
Ctrl + E/B or Alt + E/B seems to be available right now.
"E" for "empty"
"B" for "blank"

RodneyBaker:
I'd think key combinations using E for things might be best reserved for future features/enhancements related to Erasing and B for things related to Brushes.  This mirroring current shortcut keys E (Erase) and B (Brush).

manongjohn:
Auto-stretch only works for `Use Xsheet as Anination Sheet`. It is grey'd out for other modes, so for Disabled/Edit, its working as designed.

Will look to see why it won't create between levels. It should. I found some other issues and moved this to WIP to fix them. I also want to throw some warning messages when blank frames can't be created.

manongjohn:
I think Adobe uses F5 for blank frames and F6 for keyframes?  Strangely in OT's default Adobe shortcut settings, keyframes is Ctrl+F6.

snapai:
In Flash/Animate:
A "frame" is an xsheet cell that a drawing occupies. a "keyframe" is a drawing, and a "blank keyframe" is a drawing with no content.

F5: Insert Frame - extends the current exposure by 1
⇧F5: Delete Frame - reduces the current exposure by 1
F6: Insert Keyframe - converts the current frame to a duplicate drawing
⇧F6: Clear Keyframe - on the start of a drawing span, deletes it and makes it a continuation of the previous span
F7: Insert Blank Keyframe - creates a new drawing with no content.

What you're doing, completely removing the frame and leaving an open space on the xsheet is not something that's even in Flash's data model.

In Toon Boom Studio, it's "Insert Blank Cell" and it's "Shift-R" (where R is the shortcut for add exposure) - but it adds it *before* the current frame. It has a complementary command, "Delete Blank Cell" that will remove the current frame, pulling future frames closer, but only if there isn't a drawing on it.

manongjohn:
@snapai 

> What you're doing, completely removing the frame and leaving an open space on the xsheet is not something that's even in Flash's data model.

I think you misunderstand what this command is doing? This is putting a blank drawing in an empty cell on the xsheet, basically F7 in Flash as you described.

Agnyy:
Sorry, but I'm afraid I don’t understand correctly. As I understand it, this function creates just empty cells between frames? And does all this happen in automatic mode?
It turns out that such a function will work well together with automatic animation of vector layers?
image
And now you don’t need to create empty frames yourself and can you just animate and not worry?

manongjohn:
@Agnyy No, it's still a manual process.  But now you can select any number of cells on the xsheet/timeline and create blank frames in them with a single command.

RodneyBaker:
Thanks for adding the multiple frame creation @manongjohn 
Very useful.

LGTM
Assuming you don't currently plan to add anything else this PR is ready to merge.

manongjohn:
I guess i'd still to give a little time to find a decent shortcut for it.

At the very least, i can add shortcuts for the preconfigured alternate shortcuts for Adobe and Harmony. There doesn't seem to be any for Retas that I can easily find on-line.

snapai:
Suggesting the command/menu item be "Drawing" not "Frame" to correspond with the "Duplicate drawing" command from #2342 
One of these?:
- Create Blank Drawing
- Create Empty Drawing
- Insert Blank Drawing
- Insert Empty Drawing

manongjohn:
@snapai `Drawing` does sound better. I still would use `Create` as I don't want it to imply it would shift drawings down. Toss up between `Blank` and `Empty` though i am leaning towards `Blank`

melieconiek:
I would like to be able to have autostretch on for the command when autocreation is set to 'disabled' and 'enabled.' It'd have to be in a way that made sense in the options, though.

In my case, I want the features of 'Use Xsheet as Animation Sheet' with 'Enable auto-stretch frame' on--but on command, not when I draw.

manongjohn:
That's a little off topic. Better suited on the Autocreation discussion. However, with shun-iwasawa 's proposal and my plan on adding commands/buttons, it will be closer to what you are requesting i believe.

melieconiek:
I don't think discussing how this command should work is off topic. Perhaps if the command had it's own settings--so the autocreation with the pen could be set to 'Disabled' while the command could use 'Use Xsheet as Animation Sheet.'

manongjohn:
The bare essence of this command is just to create a blank drawing, but in keeping with how frames are created, it's best to follow autocreation logic so it factors in existing default level type and auto-stretch settings.  I will not be creating separate settings for this.

With the changes proposed in Autocreation discussion, those changes would apply to this command automatically once implemented. In that discussion, Auto-stretch will become available for Enabled settings and new frames will be added to existing levels...exactly like it does for Animation Sheet mode.  So if you are in Disabled mode, this command temporarily forces Enabled mode, using auto-stretch, if enabled and adding to the same level.

melieconiek:
I misunderstood the proposed changes in the autocreation discussion then. Sorry about that.

I appreciate your hard work on this. :)

manongjohn:
I'm thinking of making `Alt+D` or `Ctrl+D` the default shortcut for this command.  Any objections or preference?  I kind of like `Alt+D` myself.

RodneyBaker:
LGTM
Thanks @manongjohn!

beeheemooth:
Thank you, @manongjohn !

PR #2911, opened 2019-12-02
Fix tool crashes on Palette and Zerary column

### LINKED ISSUE #2910: Tools used on Palette and ZeraryFx columns cause crash
### Issue Summary
When using Autocreation `Enabled` mode, some tools are available for use in empty cells for ZeraryFx and Palette Columns. If you try to use those tools, it will crash OT.

### Steps to Reproduce
1. Create or load a scene that has a Palette column or a ZeraryFx column (checkerboard fx) in it.
2. Set Autocreation mode to `Enabled`
3. Switch to Brush Tool
4. Find an empty cell after the last exposed frame in the Palette or ZeraryFx column
5. Attempt to draw in the viewer. When done drawing -> CRASH

manongjohn:
This PR fixes #2910

Added logic to disable Level tools like brushes, geometry tools and type tool on blank frames if it is a Palette, ZeraryFx or Mesh column since you cannot add new levels to these columns.

### DIFF
--- a/toonz/sources/tnztools/tool.cpp
+++ b/toonz/sources/tnztools/tool.cpp
@@ -875,6 +875,12 @@ QString TTool::updateEnabled(int rowIndex, int columnIndex) {
         m_application->getCurrentObject()->getObjectId().getIndex());
   }
 
+  bool isZeraryCol =
+      column ? (column->getZeraryFxColumn() ? true : false) : false;
+  bool isPaletteCol =
+      column ? (column->getPaletteColumn() ? true : false) : false;
+  bool isMeshCol = column ? (column->getMeshColumn() ? true : false) : false;
+
   // Check against splines
   if (spline && (toolType & TTool::LevelTool)) {
     return (targetType & Splines)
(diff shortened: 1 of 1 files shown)


RodneyBaker:
In the process of testing this PR I have run into a problem (using latest nightlies) in that if I create a standard raster level Opentoonz will crash shortly after using any tool.  What I'm doing right now as a test for that crash:

Create standard rater level
Draw a rectangle with Geometry tool
Go back to toolbar to select a tool (Fill Tool for instance)
Somewhere near this point Opentoonz will crash (The cursor will not change to that of the appropriate tool).
I also am experiencing the crash if I simply wait after creating the raster level... hovering the cursor over the tool menu/canvas.

If confirmed by someone else I'll open a new report on the crash.
This is not likely to be related to this PR.  I just discovered it while testing it.

manongjohn:
I cannot seem to replicate the crash, though I am seeing that a Raster Level in Col 1 shows a black background.

RodneyBaker:
There's some current strangeness on my system where installing a clean copy of Opentoonz doesn't even add the OT icons as the installer normally would smy current... but very odd... issue is very likely local in nature.
I'll investigate.

RodneyBaker:
@manongjohn My apologies, I thought this PR was already merged.
Are there still any changes or additions pending?

manongjohn:
No changes pending.  Just need to make sure i didn't break something.

RodneyBaker:
I initially had some concern with not being able to place levels beneath the aforementioned level types.  For instance, my greater concern was with Plastic Meshes as they can be treated as frame by frame mesh animations of different levels and I thought perhaps this PR might prevent that functionality.  After much testing, nothing appears broken here.

LGTM

PR #2912, opened 2019-12-03
(WIP) Fix crash when using keyboard shortcuts while moving a line

### LINKED ISSUE #2148: Crash:  Ctrl-z and x while moving a line

martinvanzijl:
Fixes #2148.

Somehow I messed up the original PR (#2686), so I created this one instead. This avoids crashing when pressing Ctrl+X (cut), Delete, Ctrl+Z (undo) or opening a dialog (e.g. with Ctrl+N) while moving a vector line.

Video below:

moving-line-demo

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -517,6 +517,8 @@ transformation.
     return m_selectedFrames;
   }
 
+  virtual void beforeCut() {}
+
 public:
   static std::vector<int> m_cellsData;  //!< \deprecated  brutto brutto. fix
                                         //! quick & dirty del baco #6213 (undo
--- a/toonz/sources/tnztools/strokeselection.cpp
+++ b/toonz/sources/tnztools/strokeselection.cpp
@@ -515,6 +515,8 @@ void StrokeSelection::deleteStrokes() {
     undo = new ToolUtils::UndoPath(
         tool->getXsheet()->getStageObject(tool->getObjectId())->getSpline());
 
+  tool->beforeCut();
+
   StrokesData *data = new StrokesData();
   data->setImage(m_vi, m_indexes);
   std::set<int> oldIndexes = m_indexes;
@@ -628,6 +630,8 @@ void StrokeSelection::cut() {
   if (isSpline)
     undo = new ToolUtils::UndoPath(
         tool->getXsheet()->getStageObject(tool->getObjectId())->getSpline());
+
+  tool->beforeCut();
 
   StrokesData *data = new StrokesData();
   data->setImage(m_vi, m_indexes);
--- a/toonz/sources/tnztools/vectorselectiontool.cpp
+++ b/toonz/sources/tnztools/vectorselectiontool.cpp
@@ -34,6 +34,9 @@ using namespace DragSelectionTool;
 //    Global variables
 //********************************************************************************
 
+static TPointD lastDragPos; // previous mouse-drag position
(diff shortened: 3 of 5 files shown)


martinvanzijl:
As an alternative, it may be easier to simply disable these shortcuts in "SceneViewer::event" if a tool is busy with a mouse-drag operation. Something like below:

```
if (tool->isDragging() && (actionId == "MI_Undo" || actionId == "MI_Redo" ... || ... ))
   // prevent shortcut trigger
```

RodneyBaker:
This PR deals specifically with certain shortcut key combinations but I'm curious if there is a function/class that would allow for all shortcut keystrokes to be disabled during the drag operation.  Rationale:  If problems occur with other keystroke combinations they will need to be explicitly identified in code as well.  The plus side would be to future proof the process while the downside might be added delay in responsiveness (even if only very small).

martinvanzijl:
@RodneyBaker Yes, this is also possible. I've made another PR to demonstrate this approach here: #2914. It's a lot simpler to code, so may be the better solution.

RodneyBaker:
Nice.  Yes, I like that approach.

RodneyBaker:
Sorry.  Didn't mean to close.  Meant to respond.

Yes, I like the approach from #2914 and if somehow in the future some realtime feature that requires shortcut key combinations with  simultaneous movement of vector lines then the code for that feature can opt in and as an exception to the rule.

RodneyBaker:
@martinvanzijl  It looks to me like this is your choice on which solution to move forward... others don't appear to have a preference one way or the other.

gab3d:
Sorry if this is already taken into account by this PR.
But just to remind you that, for example, the Animate tool in its All mode uses the Shift modifier key to constraint certain transformations.

martinvanzijl:
@gab3d Thanks for the tip. This so far only effects Vector Selection tool, not Animate tool, so should be OK. If we extend this to the Animate tool in the future, we'll have to ensure we don't disable Shift.

@RodneyBaker I'm leaning towards #2914, but the code should be reviewed in either case.

@shun-iwasawa or @manongjohn If you have time, can you please review the code for this and #2914? Let me know which you prefer, if any.

RodneyBaker:
Closing this PR as #2914 is the preferred PR moving forward.

PR #2913, opened 2019-12-03
Block reserved file names

### LINKED ISSUE #2908: AUX.pli Windows error.
Hello,

I've reached a point in one of my projects when Opentoonz tried to create a file with a name "AUX.pli" with it's autocreation tool. I work on Windows and I've found out you can't create any file named "AUX". 

Opentoonz creates automaticly new drawing named "AUX", but failes to save it. It's pretty annoying because I have to do some tricks to pass this AUX name everytime I try tu automaticly add a new drawing.

I think you have to prevent opentoonz from creating file with that specific name at least for Windows users.

I run latest nightly build of Opentoonz on latest update of Windows 10. 

Here's article about this issue: https://www.howtogeek.com/fyi/windows-10-still-wont-let-you-use-these-file-names-reserved-in-1974/ 

Here's list of reserved filenames: CON, PRN, AUX, NUL, COM1, COM2, COM3, COM4, COM5, COM6, COM7, COM8, COM9, LPT1, LPT2, LPT3, LPT4, LPT5, LPT6, LPT7, LPT8, and LPT9

Thanks,
Bartosz Ostrowski

manongjohn:
This PR fixes #2908

Added logic for the Windows version to 
1. Skip generation of level names AUX, COM, CON, LPT, NUL, PRN
2. Prevent saving names that are reserved file names AUX, COM1-9, CON, LPT1-9, NUL, PRN

### DIFF
--- a/toonz/sources/include/toonzqt/gutil.h
+++ b/toonz/sources/include/toonzqt/gutil.h
@@ -117,6 +117,8 @@ QString DVAPI toQString(const TFilePath &path);
 bool DVAPI isSpaceString(const QString &str);
 bool DVAPI isValidFileName(const QString &fileName);
 bool DVAPI isValidFileName_message(const QString &fileName);
+bool DVAPI isReservedFileName(const QString &fileName);
+bool DVAPI isReservedFileName_message(const QString &fileName);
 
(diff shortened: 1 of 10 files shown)


RodneyBaker:
LGTM

PR #2914, opened 2019-12-04
Disable keyboard shortcuts while dragging vector line

martinvanzijl:
Possible fix for #2148. This is an alternative to #2912.

This disables all keyboard shortcuts while dragging a vector line. 

Video demo below; this shows shortcuts disabled while dragging, but enabled again when not dragging:

disable-shortcuts-demo

This could be extended for other tools such as the Brush, too.

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -513,6 +513,9 @@ transformation.
   // Tools.
   virtual bool isPencilModeActive() { return false; }
 
+  // return true if the tool is busy with a mouse drag operation
+  virtual bool isDragging() const { return false; };
+
   void setSelectedFrames(const std::set<TFrameId> &selectedFrames);
   static const std::set<TFrameId> &getSelectedFrames() {
     return m_selectedFrames;
--- a/toonz/sources/tnztools/vectorselectiontool.cpp
+++ b/toonz/sources/tnztools/vectorselectiontool.cpp
@@ -1990,6 +1990,12 @@ void VectorSelectionTool::onImageChanged() {
 
 //-----------------------------------------------------------------------------
 
+bool VectorSelectionTool::isDragging() const {
+  return m_enabled && m_leftButtonMousePressed;
+}
+
+//-----------------------------------------------------------------------------
+
 void VectorSelectionTool::doOnDeactivate() {
   m_strokeSelection.selectNone();
   m_levelSelection.selectNone();
--- a/toonz/sources/tnztools/vectorselectiontool.h
+++ b/toonz/sources/tnztools/vectorselectiontool.h
@@ -338,6 +338,8 @@ protected:
     return m_cursorId;
   }
 
+  bool isDragging() const override;
+
 private:
   class AttachedLevelSelection final : public LevelSelection {
     StrokeSelection &m_strokeSelection;  //!< Selection of strokes to be seen at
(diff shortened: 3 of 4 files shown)


RodneyBaker:
@martinvanzijl If no one objects, recommend we merge this PR as the preferred solution and close the other PR.

RodneyBaker:
@manongjohn When you have some time can you comment on which approach you find to work better?

manongjohn:
I do prefer this implementation. It's simple and easily expandable to other tools if needed.  Probably would have fixed 

I can't think of any OT tools where we would be performing some sort of dragging operation that would require or make use of a system keyboard shortcut.

manongjohn:
The only thing i can think of this not covering is OT shortcuts whose default shortcut is a system shortcut. It will be a problem if a user changes that and we still allow that operation to happen.

Example, replacing OT's Cut command shortcut (Ctrl+X) with a letter 'x'.  This fix would only capture Ctrl+X as that is considered a system shortcut, but X will not be and OT will try to execute the cut action.  I'd need to confirm this but not at computer atm.

martinvanzijl:
@manongjohn Thanks for your comments. Let's go with this PR then, instead of #2912.

> Example, replacing OT's Cut command shortcut (Ctrl+X) with a letter 'x'. This fix would only capture Ctrl+X as that is considered a system shortcut, but X will not be and OT will try to execute the cut action. I'd need to confirm this but not at computer atm.

It looks like this scenario also works OK. I tried this and the shortcut was still disabled as expected. See below:

cut-using-x-demo

shun-iwasawa:
@martinvanzijl 
Can you please remove `(WIP)` from the title if it is ready to be reviewed? Thanks!

martinvanzijl:
@shun-iwasawa Done, thanks for the reminder. Ready for your review.

In particular, do you think the name `isDragging()` is OK, or should it be renamed to `shouldDisableShortcuts()`?

shun-iwasawa:
@martinvanzijl 

>In particular, do you think the name isDragging() is OK, or should it be renamed to shouldDisableShortcuts()?

I think `isDragging()` is fine, considering that you also added an accurate description there ( `return true if the tool is busy with a mouse drag operation` ).
I'm not sure but this function can be renamed to more abstract ( like `isBusy()` )  in the future development if some tool turns out to be needed to block shortcuts for reasons other than dragging.

RodneyBaker:
If not for the conflict with toonz/sources/tnztools/vectorselectiontool.h I'd say this PR is ready to be merged.

martinvanzijl:
Thanks, @RodneyBaker.

@shun-iwasawa To resolve the conflict, do I use "git pull --rebase upstream master", update the code, then "git push origin disable-shortcuts-while-tool-is-dragging"?

shun-iwasawa:
@martinvanzijl 
Sorry, but I'm still not familiar with git commands and I'm not sure if `git pull --rebase upstream master` will update your origin/master with the upstream.
Here are my routine for resolving conflicts:

`> git checkout master`  // move to the local master branch
`> git pull upstream master` // sync the local master with the upstream
`> git push origin master` // sync the remote master with the local
`> git checkout disable-shortcuts-while-tool-is-dragging` // move to the PR branch
`> git rebase master` // rebase the commits to the latest version of local master
--- update the code --- 
`> git push -f origin disable-shortcuts-while-tool-is-dragging` // push the changes. `-f` (force update) is needed as the history now conflicts between the local and the remote due to rebasing.

If anyone knows better way, please let me know.

manongjohn:
I do those steps as well, but I tend not to do `git rebase and opt do a `git pull` instead. Resolve any conflicts, then `git push`

martinvanzijl:
@shun-iwasawa Thanks, I used your suggested commands and it looks like it worked fine.

shun-iwasawa:
LGTM Thank you @martinvanzijl for the fix !

PR #2568, opened 2019-03-17
Note Level enhancement and crash fixe

manongjohn:
This PR does the following with Note Levels

1) Adds a Note Level Cell Context Menu 
image

2) Fixes a crash when trying to `Replace Level` on a Note Level.

PR #2915, opened 2019-12-04
Recommit Add Note Level context menu

manongjohn:
This PR reintroduces the Note Level context menu that was originally merged as part of PR #2568 but was somehow lost when it was overwritten with an older version of the file by a later commit.


manongjohn:
Jenkins needs to be called on this again

RodneyBaker:
LGTM
Thanks for the quick turnaround @manongjohn

PR #2916, opened 2019-12-04
Quick fix for build error

shun-iwasawa:
This PR is to quick-fix the build error arises after merging #2894 .

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -246,7 +246,7 @@ public:
     int r0, c0, r1, c1;
     selection->getSelectedCells(r0, c0, r1, c1);
     if (c0 < 0) c0 = 0;  // Ignore camera column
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I'm curious if a run of jenkins prior to the build would have been sufficient to catch this.

I ask for two reasons:  a similar issue occurred recently with another PR (as well as apparently a few in the past) and I'm also curious if having Jenkins run automatically as part of the build process might be a option.   I can see where we might not want it to be automated but I can see benefits as well.

manongjohn:
LGTM

shun-iwasawa:
@RodneyBaker 
I think one of the reasons of this problem is that I made Jen.kins to run automatically as soon the PR is submitted. This way cannot catch errors caused by changes after the submission.
I previously triggered Jen.kins manually just before merging the PR. It had a smaller risk of such error.
Nonetheless, the automatic-running is IMO still advantageous.

PR #2924, opened 2019-12-06
Allow double-click in Load Scene dialog to open selected scene

martinvanzijl:
This PR allows you to double-click on a scene in the "Load Scene" dialog to open it.

Video below:

double-click-demo

### DIFF
--- a/toonz/sources/toonz/dvitemview.cpp
+++ b/toonz/sources/toonz/dvitemview.cpp
@@ -1301,6 +1301,7 @@ void DvItemViewerPanel::mouseReleaseEvent(QMouseEvent *) {}
 void DvItemViewerPanel::mouseDoubleClickEvent(QMouseEvent *event) {
   int index = pos2index(event->pos());
   if (index < 0 || index >= getItemCount()) return;
+  if (m_viewer) m_viewer->notifyDoubleClick(index);
   if (!getModel()->canRenameItem(index)) return;
   QRect captionRect = getCaptionRect(index);
   if (!captionRect.contains(event->pos())) return;
--- a/toonz/sources/toonz/dvitemview.h
+++ b/toonz/sources/toonz/dvitemview.h
@@ -412,6 +412,10 @@ public:
     emit clickedItem(index);
     emit selectedItems(m_panel->getSelectedIndices());
   }
+  void notifyDoubleClick(int index) {
+    emit doubleClickedItem(index);
+    emit selectedItems(m_panel->getSelectedIndices());
+  }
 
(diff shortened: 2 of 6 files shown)


artisteacher:
Sweet! Should this be extended for the load levels dialog?

gab3d:
Let me tell you you've probably made the simplest most wanted PR in a long time! 😸 
Thank you!

manongjohn:
2 things I noticed while briefly playing with it

- If you bring up a File Browser window or go to the Browser room, you  can't double-click a scene to load it.  I think it would be nice if it worked there as well. 

Aside: Perhaps in the future, double-clicking on any file type in the File Browser window would attempt to load it as a scene or level, depending on the file extension.

- When in the Browser room, if you highlight a scene but then use File -> Load Scene, it uses whatever you highlighted.  In current version, it still brings up the Load Scene dialog.  Not a big deal to me. It kind of makes more sense this way but thought I would note a difference in behavior if it wasn't intended.

martinvanzijl:
Thanks all for your feedback.

@artisteacher 
> Sweet! Should this be extended for the load levels dialog?

Yes, this makes sense. Suggest we do that in a future PR. I'd like to get just this change committed first.

@manongjohn 
> If you bring up a File Browser window or go to the Browser room, you  can't double-click a scene to load it.  I think it would be nice if it worked there as well.

This makes sense, too. Suggest we do so in a future PR.

> Aside: Perhaps in the future, double-clicking on any file type in the File Browser window would attempt to load it as a scene or level, depending on the file extension.

Also good for a future PR. Might take a lot of work, though.

> When in the Browser room, if you highlight a scene but then use File -> Load Scene, it uses whatever you highlighted. 

Actually, this happens for me in the current master, too. I don't think I've changed this. Can you please check if it happens for you in the current master?

DarrenTAnims:
Thank you. Thank you. Thank you.
This is standard in all other apps on Windows, so is something I do every time I try and load a scene. And every time I curse that it doesn't work. Great job.

manongjohn:
I checked the latest nightly and confirm what you are seeing. I think I was comparing against the original release. The behavior must have changed along the way and didn't notice until now.

Anyway. LGTM.

RodneyBaker:
LGTM
I agree that it would be ideal to have the same double click capability in other areas of Opentoonz.

Thanks @martinvanzijl !

beeheemooth:
Cool! But I always download the scene through the browser.

martinvanzijl:
@RodneyBaker Thanks, I'll try to make a feature request for the rest.

PR #2928, opened 2019-12-09
Add Shift Keys commands

manongjohn:
This PR adds `Shift Keys Down` and `Shift Keys Up` commands allowing users to move keys up and down the xsheet based on selections with either shortcuts, buttons or context menu actions

Keys will not shift if it runs into another key that is not being shifted.

Ways to shift keys:
1. Cell context menu:
image

Usage: Select 1 or more frames in a column (or columns) on the xsheet and use one of the commands.  Selection does not need to contain keys.

Results: Any keys in that are in the selection or after the selection will be shifted by the number of frames in the column that are selected.  If you select 5 frames in the column, it will shift by 5 frames.

Keys before the selection will not be shifted.

2. Keyframe context menu:
image

Usage: Select specific keys on the xsheet and use one of the commands

Results: Only the selected keys will be shifted by 1 frame.

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -1408,6 +1408,10 @@ void TCellSelection::enableCommands() {
   enableCommand(this, MI_CloneLevel, &TCellSelection::cloneLevel);
   enableCommand(this, MI_SetKeyframes, &TCellSelection::setKeyframes);
 
+  enableCommand(this, MI_ShiftKeyframesDown,
+                &TCellSelection::shiftKeyframesDown);
+  enableCommand(this, MI_ShiftKeyframesUp, &TCellSelection::shiftKeyframesUp);
+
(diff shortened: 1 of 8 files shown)


RodneyBaker:
> Keys will not shift if it runs into another key that is not being shifted.

My current testing has all keys including those not selected shifting.
Double checking as I'm likely just using the feature wrong.

manongjohn:
If you don't specifically select keys (keys stay white) and just select cells, then all keys in and below the selection area in the xsheet will move. (cell shift)

If you select keys specifically (they turn blue), only those keys will shift and stop if it runs into another one. (key shift)

If you do a Ctrl+Select to select cells+keys, it treats it as a cell shift, not a key shift.

RodneyBaker:
Thanks for the info @manongjohn 
That's got it working.

RodneyBaker:
LGTM

Issue #2929, opened 2019-12-10
Auto Inbetweening Ease In vs Ease Out results are backwards


manongjohn:
Putting aside the whole "Ease In/Out" label being reversed, there are a few places where "Ease In" and "Ease Out" occur by themselves.  Per the Animator's Survival Kit, our auto-inbetween Ease In and Ease Out options appear to be backwards.

Here is what our auto-inbetweening from Vector Range option does:

![image](https://user-images.githubusercontent.com/19245851/70496448-9cdc4f80-1add-11ea-8b60-e2fb69d24c66.png)

Ease In and Ease Out auto-inbetweens occur as separate entries in the following places and should be switched

- Level Strip's Inbetween dialog
- Vector Range toolbar option

Testing the Ease In and Ease Outs in the various gradient Fx appear to be correct.

PR #2931, opened 2019-12-10
Add Vector Guided Drawing Auto-Inbetween option

manongjohn:
This PR adds Inbetween options to Vector Guided Drawing. This was inspired by #1423.

image

I've leveraged the Vector Guided feature with the Vector Range inbetweening to get the Guided Auto-Inbetween feature. As such, any issues inherent to Guided Drawing or Range Inbetweening usage could also apply to Guided Inbetweening.

Usage Notes

This applies to all commands/options
- Can be used in Scene Edit or Level Edit modes.
- Only enabled when `Closest Drawing` or `Farthest Drawing` is selected.
- When using guided inbetweening, it is recommended that you use frames that are in numerical order, regardless if in Level Edit or Scene Edit mode
- Guided inbetweening is done at a Level point of view, even when you do this in Scene Edit mode.
   - If you use guided inbetweening on exposed Frames 1-3,5-6 (5 frames, skipping Frame 4) in Scene Edit mode, the inbetweening will actually occur over Frames 1-6 (6 frames).
   - If you use guided inbetweening on exposed Frames order 1,4,3,2,5 in Scene Edit mode, the inbetweening will occur over ordered frames 1,2,3,4,5.
   - Shortcut entries for these options are found under Right-click Menu Commands

### Auto Inbetween
- You can use fixed onionskins, relative onionskins or a mixture of both. The onionskin markers that are closest or furthest will  be used accordingly.
- You can use guided inbetweening using Previous onionskin markers, Next onionskin markers or both
    - When using both, currently the interpolation is done from previous to current and then current to next, independently.  It is not a continuation from previous to next.
- The Vector Range option overrides Vector Guided Drawing Auto-Inbetween

Example of usage:

guided_inbetween3

NOTE: Yes, the ease in and ease out are backwards in this animation but have been corrected in a different PR. 

### Select Previous Frame Guide Stroke/Select Next Frame Guide Stroke/Select Prev & New Frame Guide Strokes
- When pressed, you will enter the Guide Stroke selection mode.  If you click on an onion skinned stroke from a Guide frame, it will set that as the current Guide Stroke.
- The chosen Guide Stroke will remain fixed until you choose another Guide Stroke or use the `Reset Guide Stroke Selections`
- These commands can be used from the following tools: Selection, Brush, Geometric, Type, Control Point Editor, Pinch, Pump, Magnet, Iron, Cutter

### Reset Guide Stroke Selections
- When pressed, this will revert any specifically selected Guide Stroke and resume automatic selection.
- These commands can be used from any tool.

### Tween Select Guide Strokes
- This will create inbetween strokes from the `Previous Frame Guide Stroke` to the `Next Frame Guide Stroke`
- Can be used from any Vector based tool that allows you to select Guide Strokes

### Tween Guide Strokes to Selected
- This will create inbetween strokes from the `Previous Frame Guide Stroke` and/or the `Next Frame Guide Stroke` to a selected stroke in the current frame
- This can be used from the Selection Tool

### Select Guide Strokes & Tween Mode
- This will enter you into a state where you are in a constant process of picking Previous and Next Frame Guide Strokes and automatically tweening them once the pair is made.
- Accidentally clicking an invalid selection (blank spot, a stroke not in the right frame, etc) will not kick you out of this mode. It will wait until you click a valid stroke before moving on to the next step.
- To stop this mode you either need to switch tools or click the `Reset Guide Stroke Selections` button or other Guided inbetweening button


melieconiek:
Since the stokes are being inbetweened one at a time, would it be feasible to set this up so you could select which stroke you'd like to draw using forward/back keybindings for thumbing through the strokes on the source frame?

This wouldn't jive with the inbetween command on the level strip--but vector guided drawing could default to the stack order unless one of the selection commands was used to change the source stroke.

Would be a promising solution--I don't want to pile a feature request on this, though.

manongjohn:
@melieconiek I can see where that can be useful like purposefully skipping a stroke or perhaps retween a prior stroke so it appears to split.  I will look into it.  I have some ideas on how I would need to implement it. For now I'd like to consider that a separate request from this one.

Acaunt:
And how can I check this feature?

manongjohn:
@Acaunt Follow the instructions listed here to pull the artifact from this PR so you can test it locally: https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_test_prs.md

Acaunt:
Thank you so much.

melieconiek:
It might also be useful if there was some way you could use this to tween to an already existing stroke, perhaps by selecting it an executing a command. Another separate request, though.

Thanks for this, it will be useful.

gab3d:
@manongjohn 
nice feature!

is it me or this new feature also needs to be reversed? (ease in - out)

if I understood correctly, from the image you posted in #2929 this was the easing results **before** the switching:
imagen
...but they look the same to what you now show as results of this PR 🤔

manongjohn:
The PR to reverse ease in and ease out was merged this morning.  The artifacts in this PR were created before the merge so it won't have that fix yet.  I'll retrigger the recreation of the artifact so we can verify.

RodneyBaker:
We need to watch #2935 to make sure it doesn't cause issues with this PR.
That PR itself shouldn't be problematic but perhaps the order of merging...

beeheemooth:
This PR is exactly what I wrote here: https://groups.google.com/forum/#!category-topic/opentoonz_en/feature-requests/fJds1ebeLtM
@manongjohn  Thank you for your work!

melieconiek:
I went ahead and tried to use this for a little work. Selecting an reference stroke via key command then drawing works really well, I like it.

For inbetweening existing strokes it's a bit cumbersome--it takes three different commands and two clicks to tween using backward/forward selection. When you're tweening many strokes, this adds up. Since there are three separate commands, I also found myself hitting the wrong keys from time-to-time. A single command followed by selecting the backward then forward stroke, resulting in a tween, might be a little faster (one command, two clicks.)

Tweening using a selection is a little better, but is similarly cumbersome. Again, selecting your stroke in the viewer, a single command, and selection via click would be speedier.

The results are fantastic, though. I love what you're doing here.

manongjohn:
Will look into options to improve tweening existing strokes.

hayoko:
Nice work! In the future it would be neat if you gave this stuff its own window with push buttons and radio buttons. 🙏

manongjohn:
Ok. I've added 2 new options:
- Select Prev & Next Frame Guide Strokes
- Select Guide Strokes & Tween Mode

See PR description for more information.

Barring any major issues, I would like to see this in the upcoming 1.4 release, with or without a window for these settings.

I am looking into setting up some sort of settings window for this, but don't want it to hold up getting included into 1.4 because of it.

openanim:
@manongjohn Thank you for its promising features. Is it possible to select multiple strokes or group of strokes before tweening with these different commands?

manongjohn:
@openanim This is for individual strokes only, I'm afraid.  It doesn't even consider grouped strokes.

shun-iwasawa:
I tested the feature and found this is very nice.
The most notable point I think is that the guided tweening feature frees users from restrictions of drawing order and one-to-one correspondence of the strokes.

Although I came up with some small ideas for refactoring, I think it can be handled afterwards in separate PR and it'd better to merge as-is considering the schedule.

Thank you @manongjohn ! LGTM

openanim:
@manongjohn, how to use the normal Guided drawing mode whithout auto-tween? Currently, I have an immediate interpolation of my empty frames between my keyframes but it is not a behavior that I want every time, to have control over it.Is there a way to do it?

manongjohn:
@openanim Just uncheck Vector Guided Drawing -> Auto Inbetween to return back to normal Guided Drawing. It should have been off by default.

openanim:
Oh perfect , thanks !

If we could also define shortcuts for commands, auto inbetween/ease in /out /ease in out/ it would be cool to speed up the workflow.

openanim:
Can someone explain to me how this part works?

"Tween Guide Strokes to Selected

    This will create inbetween strokes from the Previous Frame Guide Stroke and/or the Next Frame Guide Stroke to a selected stroke in the current frame
    This can be used from the Selection Tool"

I got a result once without understanding the process. All of my other tests give nothing. The selected stroke is no longer selected after using the command Tween Guide Strokes to Selected, but nothing happens.

beeheemooth:
@openanim you must select one vector with the Selection Tool and the other is from the guide

openanim:
@beeheemooth this is what I think I did right. But nothing is happening

this is what I think I did right. But nothing happens, where am I wrong?
GD_pan

openanim:
Ah yes I got it! Thanks, I think my mistake came from the fact that I try to apply it on the current frame ... not very logical :)

PR #2934, opened 2019-12-11
Level Settings for Multiple Selection

shun-iwasawa:
This PR improves the Level Settings popup.
- Enabled with multiple selection and editing them together. When you select different types of levels, only the common properties become editable.
- Enabled undo for level settings changes.
- Fixed a bug that the change in the file path of palette level did not take effect.

levelsettings


gab3d:
amazing!

RodneyBaker:
Outstanding.  This will be a great time saver.

RodneyBaker:
Outstanding.  This will be a great timesaver.
Still exploring but looking very good thus far.

artisteacher:
@shun-iwasawa -  On a similar note it would be really helpful to be be able to apply an interpolation mode to a multiple selection in the Function Editor. Right now switching interpolation modes after animating can be very time consuming.  We can already select multiple columns or cells - but right now the interpolation mode only changes for the first pair of keyframes selected.

gab3d:
@artisteacher 
You can already do so by selecting the range of frames you desire in the Function Editor Spreadsheet and right clicking to select one of the **Change Interpolation** sub-menu options.
You can also do it in a similar way, using the Graph Editor and right clicking, but it will only work for a single channel at a time.

The Function Editor Interpolation drop down menu still can't do it with multiple segments, I'm affraid.

PR #2938, opened 2019-12-12
XDTS File Support

shun-iwasawa:
This PR enables to import and export XDTS files, the general-purpose digital exposure sheet file format.

## What is XDTS ?

XDTS stands for EXchange Digital Time Sheet. According to [the news release by Celsys](https://www.celsys.co.jp/en/topic/2018112901) :

> ■ Standard format for digital exposure sheets
> With the aim of supporting the move to digitalized animation processes, Celsys proposes a general-purpose timeline information file format that will improve compatibility with other software and smoothly integrate with animation process management tools at production companies. This format has been created in collaboration with Toei Animation Co., Ltd., the largest animation production company in Japan, and is available for anyone to use.

The format is basically written in JSON (except the first line indicating the format) . The detailed file specification is [here](http://vd.clipstudio.net/clipcontent/paint/app/ToeiAnimation/XDTSFileFormat_en.pdf) .

XDTS is available in Clip Studio Paint EX (Ver. 1.8.4 and newer) and [Toei Animation Digital Exposure Sheet for Windows](https://www.clipstudio.net/ja/dl/toeianimation#en). Some Japanese developers also made it readable in other software, such as [AfterEffects](http://sunrisemoon.net/program/ae/script/TimeSheet/) , [TVPaint](https://forum-jp.tvpaint.com/viewtopic.php?f=15&t=379&p=1943#p1943) and [UAT](https://www.u-at.net/) .

## Loading XDTS file in OT

1. Just use `Load Scene` command. In addition to the TNZ file, you can also choose XDTS files from the Load Scene popup.
1. Once you choose the XDTS file, another popup is opened (as shown in the above) for specifying the file path of each level, since XDTS format does not contain file location information.
1. The popup try to find and auto-suggest the level file locations in the same folder as the XDTS file by default. You can also leave the file fields empty. OT will create an empty level in such case.

**Attention:**
- If you save the scene loaded from XDTS file, instead of overwriting it, the scene will be saved in TNZ file with the same name as the XDTS file.
- Only the "Cell" field will be loaded. the "Dialog" and the "Camerawork" fields will be ignored.
- Only the first page of the timetables will be loaded even if the file contains multiple pages.
- Non-digit cell data and special symbols (like _Inbetween symbol_ and _Reverse sheet symbol_) will be ignored. 

## Exporting XDTS file from OT

1. Use `Export Exchange Digital Time Sheet (XDTS)` command. The command is placed at `Menu bar>File>Export` by default. Existing users need to refresh the menu bar settings or to introduce the command manually. 
1. The _current_ Xsheet will be exported. If you are in sub-Xsheet, it will be exported. 
1. XDTS file will be exported in the same folder, with the same name as the current scene file. (e.g. the scene saved in `C:/foo/bar/scenename.tnz` will export `C:/foo/bar/scenename.xdts` )

**Attention:**
- Any information other than placement of the cells will be ignored.
- Only the cells containing Toonz Raster / Toonz Vector / Raster / Mesh levels will be exported. 
- If multiple levels are placed in one column, only the first one will be exported and the other levels will be ignored.
- Currently the frame number suffix will be ignored.


RodneyBaker:
Oustanding addition to Opentoonz.

I'm currently getting a crash when I have just imported an xdst file successfully but then try to load another one.  I'm prompted for what to do with the old scene... choose 'Disregard' from the menu and... crash.

Also, a random workflow thought:
It would be nice if after the notification of the successful export Opentoonz would open a window for the directory containing the xdts file.  I'd go so far as to suggest that opening of the window could be done if the export succeeds and the popup message would only appear if the export is not successful.

shun-iwasawa:
@RodneyBaker Thank you for the feedback.
Made the notification popup type from "Warning" to "Information" and enable to open containing folder from it.
I couldn't reproduce the crash yet.

RodneyBaker:
Thanks for the addition.
I'll continue to test and will report specific steps to recreate the crash if when I can lock that down.

RodneyBaker:
This might be too broad a request but the thought came while testing...

Given that there is an export option for xdts files via Menu > File > Export..
Would it be appropriate to have a option to import xdts files via Menu > File > Import?

RodneyBaker:
I haven't been able to reproduce the crash.
If no other additions/changes are planned this side of v1.4 this PR is ready to merge.

shun-iwasawa:
Thank you @RodneyBaker for testing the feature.
I integrated this to `Load Scene` command instead of making some command under `Import` submenu because the term "Import" gives me the impression that the command will bring something into the existing scene, rather than switching scenes.

I'll merge this by myself as I'd like ohter studios to test the feature.

gab3d:
I also would expect something "Imported" to fall into the existing scene...

Issue #2905, opened 2019-11-25
(Discussion) Autocreation modes


shun-iwasawa:
Currently the preferences option `Drawing > Autocreation` has been changed to `Use Xsheet as Animation Sheet` by default.
There are users feeling inconvenient with it, especially that a new drawing is unexpectedly created (with suffixed number) when renumbering the cell:
https://groups.google.com/forum/#!msg/opentoonz_en/M3W27jxGvj4/cbRtLPkfAgAJ

I'm concern about bigger confusion when the next version will be released. 
IMO I'd like to modify the option more ideal, or at least to revert the default to `Enabled` .

Please let me know your thought. Don't you feel any inconvenience with the current option of `Use Xsheet as Animation Sheet` ?

I discussed with an animator in our studio about an ideal numbering method and conceived of the following:

```
When creating a new drawing in some vacant cell,
search the previous cells in the current column and take the maximum drawing number.
The new drawing will be with the number of (maximum + 1).
The following frames will be automatically renumbered with incremented numbers as needed.
(i.e. the drawing will be inserted keeping ascending order in the column)
```

If time permits, I'll try to introduce the above as an alternative behavior of `Autocreation = Enabled` when `Sync Level Strip Drawing Number Changes with the Xsheet = ON` .

beeheemooth:
I remember the time when there was "Enabled" mode. 80% of questions from users were "how to create a new frame on Xsheet?". There was no timeline then, and users could easily get used to the idea that Xsheet is a table. But now there are users who draw in the timeline and do not even suspect the existence of Level Strip (they have this window turned off as unnecessary). And they draw (most often in the raster) duplicating frames and deleting them right on the timeline, not suspecting the intricacies of Autocreation.

However, in order not to argue about who is comfortable with which Autocreation mode, it may be worth looking at the problem from the other side. I have come across many times that different tasks require different Autocreation modes. Most often, I go to Preferences in order to change the "Autocreation" and "Enable auto-stretch frame" modes. Maybe there is still the opportunity to transfer these options to the Xsheet Toolbar, or at least make them available as operations?
 
Here's the topic on the Google forum regarding the above :https://groups.google.com/forum/#!category-topic/opentoonz_en/feature-requests/w3oz6Fxdxbc

manongjohn:
[...]

PR #2943, opened 2019-12-16
Modify Autocreation Feature

shun-iwasawa:
This PR modifies the autocreation feature based on comments in #2905 , also including my proposed options based on the discussion with an animator in our studio.

### New Options Introduced
autocreation
I divided up the behaviors which were available in conventional `Use Xsheet as Animation Sheet` mode and made them optional like as follows:

- **Numbering System** : option is either `Incremental` and `Use Xsheet as Animation Sheet` . It specifies a number to be set for new drawing. ( `Incremental` will be just like the former `Autocreation:Enabled` option. )
- **Enable Auto-stretch Frame** : same as before, except that it now works regardless of other options. 
- **Enable Creation in Hold Cells** : specifies whether a new drawing will be created when start drawing in the "hold" cell - i.e. the cell with the same contents as the previous one. (This behavior was enabled only in the "former" `Use Xsheet as Animation Sheet` mode). Note that when you use the `Create Blank Drawing` command, now this option will become temporarily activated so that you can insert drawings in hold cells regardless of the current state.
- **Enable Autorenumber** : when inserting a new drawing between existing frames, with this options being ON, the following frames will be automatically renumbered so that the frame orders becomes straightforward as possible.

Whole options are grouped and nested in the `Enable Autocreation` box.
The former `Autocreation mode` is deprecated in this PR. OT will automatically reproduce the similar environment for existing users.

### Other changes

- "One level for one column" basis - Expose the same level even when drawing on the cell further down from exiting level on the column in any modes.
- When typing numbers on the cell, if the drawing exists, expose it in any modes, instead of creating the new drawing with suffix.
- Typing suffix is now available in any modes.
- Enabled to create a new drawing in a cell without images (i.e. cells with red numbers). You can now type numbers in the Xsheet before actually draw the images.

### Default values

I made the default values as follows, but are still open for discussion:

- **Enable Autocreation** : Enabled
- **Numbering System**: Incremental
- **Enable Auto-stretch Frame**, **Enable Creation in Hold Cells**, and **Enable Autorenumber** : Enabled

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -518,14 +518,23 @@ transformation.
   }
 
 public:
-  static std::vector<int> m_cellsData;  //!< \deprecated  brutto brutto. fix
-                                        //! quick & dirty del baco #6213 (undo
+  struct CellOps {
+    int r0;
+    int r1;
+    enum Type { ExistingToNew = 0, BlankToExisting, BlankToNew } type;
+  };
+  static std::vector<CellOps>
+      m_cellsData;  //!< \deprecated  brutto brutto. fix
+                    //! quick & dirty del baco #6213 (undo
   //! con animation sheet) spiegazioni in
   //! tool.cpp
   static bool m_isLevelCreated;  //!< \deprecated  Shouldn't expose global
                                  //! static variables.
   static bool m_isFrameCreated;  //!< \deprecated  Shouldn't expose global
                                  //! static variables.
+  static std::vector<TFrameId> m_oldFids;
+  static std::vector<TFrameId> m_newFids;
+  static bool m_isLevelRenumbererd;
 
 protected:
   std::string m_name;  //!< The tool's name.
@@ -578,9 +587,8 @@ public:
   }
   ImagePainter::VisualSettings &visualSettings() { return m_visualSettings; }
 
-  virtual double getPixelSize()
-      const = 0;  //!< Returns the length of a pixel in current OpenGL
-                  //!< coordinates
+  virtual double getPixelSize() const = 0;  //!< Returns the length of a pixel
+                                            //!< in current OpenGL coordinates
 
   virtual void invalidateAll() = 0;    //!< Redraws the entire viewer, passing
                                        //! through Qt's event system
@@ -600,7 +608,7 @@ public:
   //! return the column index of the drawing intersecting point \b p
   //! (window coordinate, pixels, bottom-left origin)
   virtual int posToColumnIndex(const TPointD &p, double distance,
-                               bool includeInvisible = true) const = 0;
+                               bool includeInvisible = true) const    = 0;
   virtual void posToColumnIndexes(const TPointD &p, std::vector<int> &indexes,
                                   double distance,
                                   bool includeInvisible = true) const = 0;
@@ -632,9 +640,9 @@ public:
 
   virtual void rotate(const TPointD &center, double angle) = 0;
   virtual void rotate3D(double dPhi, double dTheta)        = 0;
-  virtual bool is3DView() const      = 0;
-  virtual bool getIsFlippedX() const = 0;
-  virtual bool getIsFlippedY() const = 0;
+  virtual bool is3DView() const                            = 0;
+  virtual bool getIsFlippedX() const                       = 0;
+  virtual bool getIsFlippedY() const                       = 0;
 
   virtual double projectToZ(const TPointD &delta) = 0;
 
--- a/toonz/sources/include/tools/toolutils.h
+++ b/toonz/sources/include/tools/toolutils.h
@@ -152,10 +152,13 @@ protected:
   bool m_isEditingLevel;
   bool m_createdFrame;
   bool m_createdLevel;
-  bool m_animationSheetEnabled;
-  std::vector<int> m_cellsData;  // represent original frame range when
-                                 // m_animationSheetEnabled, m_createdFrame and
-                                 // !m_isEditingLevel; see tool.cpp
+  bool m_renumberedLevel;
+  std::vector<TTool::CellOps>
+      m_cellsData;  // represent original frame range when
+                    // m_animationSheetEnabled, m_createdFrame and
+                    // !m_isEditingLevel; see tool.cpp
+  std::vector<TFrameId> m_oldFids;
+  std::vector<TFrameId> m_newFids;
   TPaletteP m_oldPalette;
   std::string m_imageId;
   static int m_idCount;
@@ -504,6 +507,14 @@ TRasterPT<PIXEL> rotate90(const TRasterPT<PIXEL> &ras, bool toRight) {
   }
   return workRas;
 }
+
+bool DVAPI doUpdateXSheet(TXshSimpleLevel *sl, std::vector<TFrameId> oldFids,
+                          std::vector<TFrameId> newFids, TXsheet *xsh,
+                          std::vector<TXshChildLevel *> &childLevels =
+                              std::vector<TXshChildLevel *>());
+
+bool DVAPI renumberForInsertFId(TXshSimpleLevel *sl, const TFrameId &fid,
+                                const TFrameId &maxFid, TXsheet *xsh);
 
 }  // namespace ToolUtils
 
--- a/toonz/sources/include/tools/toolutils.h
+++ b/toonz/sources/include/tools/toolutils.h
@@ -510,8 +510,7 @@ TRasterPT<PIXEL> rotate90(const TRasterPT<PIXEL> &ras, bool toRight) {
 
 bool DVAPI doUpdateXSheet(TXshSimpleLevel *sl, std::vector<TFrameId> oldFids,
                           std::vector<TFrameId> newFids, TXsheet *xsh,
-                          std::vector<TXshChildLevel *> &childLevels =
-                              std::vector<TXshChildLevel *>());
+                          std::vector<TXshChildLevel *> &childLevels);
 
 bool DVAPI renumberForInsertFId(TXshSimpleLevel *sl, const TFrameId &fid,
                                 const TFrameId &maxFid, TXsheet *xsh);
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -134,6 +134,7 @@ public:
               QVariant defaultValue, QVariant min = 0, QVariant max = -1);
 
   void setCallBack(const PreferencesItemId id, OnEditedFunc func);
+  void resolveCompatibility();
 
   PreferencesItem &getItem(const PreferencesItemId id);
   bool getBoolValue(const PreferencesItemId id) const;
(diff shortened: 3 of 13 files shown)


artisteacher:
Will autorenumber change the frame numbers in the level strip? If so, would it cause problems with old scenes using the same level?

RodneyBaker:
@artisteacher asked: 

> Will autorenumber change the frame numbers in the level strip? If so, would it cause problems with old scenes using the same level?

My current testing suggests the old data stays in place until updated and... does a really good job with that.  I've even been testing moving back and forth from 'Incremental' to 'Use Xsheet as Animation Sheet'... and moving levels into and out of columns... and Opentoonz reacts as I would expect.

Surely there will be some issues found that require some tweaking but so far so good.

gab3d:
Thanks @shun-iwasawa !! it seems a very good improvement!

> "One level for one column" basis - Expose the same level even when drawing on the cell further down from exiting level on the column in any modes.

I couldn't test it myself yet, but I have a question:
**One level for one column** refers to the inability to add drawings to different levels exposed in the same column?
From the top of my head, I'd think that it could be a good idea to take the immediate previous existing level in the column to decide to which level to add drawings to...
(Perhaps I didn't understand right.) 🤔

manongjohn:
@shun-iwasawa Some compiling issues with Linux and macOS builds.

artisteacher:
>  it could be a good idea to take the immediate previous existing level in the column to decide to which level to add drawings to...

Good idea!

manongjohn:
If I am not mistaken, I believe it's supposed to model the behavior of `Use XSheet as Animation Sheet` and use the level just before it to determine what level type to make instead of just creating a brand new level.

gab3d:
> If I am not mistaken, I believe it's supposed to model the behavior of `Use XSheet as Animation Sheet` and use the level just before it to determine what level type to make instead of just creating a brand new level.

If that's the case, great! 😄 
Although, in that scenario, I still can't understand to what **One level for one column** refers to...

RodneyBaker:
@gab3d I was initially concerned that 'one level per column' meant we wouldn't be able to have more than one level in a column.  I'm glad to report that is not the case.   Basically, what @manongjohn said.  :)

shun-iwasawa:
@artisteacher 

>Will autorenumber change the frame numbers in the level strip? If so, would it cause problems with old scenes using the same level?

As @RodneyBaker pointed, yes, and yes. It renumberes the level strip and synchronize Xsheet with new numbering. However, just like the preferences option `Sync Level Strip Drawing Number Changes with the Xsheet` , it only cares the current scene and if the level is used in another scene there will be a problem. (Actually I reuse the function of `Sync Level Strip Drawing Number Changes with the Xsheet` for this option. Thank you @turtletooth ! )
It is difficult to automatically update other scenes containing the renumbered levels. Although IMO this option is still usable in most cases assuming that users usually complete drawings before reusing it in another scene.

@gab3d 

As @manongjohn and @RodneyBaker already pointed, the term `"One level for one column" basis` is nothing more than a statement of the modification in the autocreation behavior. The former `Autocreation:Enabled` option had been creating a new level when users draw in a cell further down from the existing level. Which was usually not desired.

shun-iwasawa:
One thing I'm unsure regarding this PR is a behavior when `Numbering System = "Use Xsheet as Animation Sheet" ` and `Enable Autorenumber = ON`.

Let's say, we have a scene like this:
before

When you draw in the frame 3, OT should no doubt insert the drawing # 3 (as it is under `Use Xsheet as Animation Sheet` numbering system ). However, for renumbering, I think there could be a three options as follows...

1. A "weak" renumbering; just shift numbers of conflicting drawings. In this example, only the existing drawing # 3 conflicts and is shifted to # 4 , regardless of the location of the drawing. (This is a current implementation in this PR.)
renumber1

1. A "medium" renumbering; renumbers all **following** drawings with scene frame numbers where they locate. Drawings in the upper cells (In this example, the drawing # 1 in the frame 2) will be left unchanged.
renumber2

1. A "strong" renumbering;  renumbers all drawings in the level. It will be just like executing `Autorenumber` command in the current column every time when inserting a new drawing.
renumber3

I would appreciate your thoughts and opinions.

gab3d:
@shun-iwasawa 

> As @manongjohn and @RodneyBaker already pointed, the term "One level for one column" basis is nothing more than a statement of the modification in the autocreation behavior. The former Autocreation:Enabled option had been creating a new level when users draw in a cell further down from the existing level. Which was usually not desired.

Thanks for clarification, I misinterpreted it.
It's OK for me.

gab3d:
@shun-iwasawa 
> As @RodneyBaker pointed, yes, and yes. It renumberes the level strip and synchronize Xsheet with new numbering. However, just like the preferences option Sync Level Strip Drawing Number Changes with the Xsheet , it only cares the current scene and if the level is used in another scene there will be a problem. (Actually I reuse the function of Sync Level Strip Drawing Number Changes with the Xsheet for this option. Thank you @turtletooth ! )

> It is difficult to automatically update other scenes containing the renumbered levels. Although IMO this option is still usable in most cases assuming that users usually complete drawings before reusing it in another scene.

Still, as it could potentially ruin scene work done previously, could it be possible to implement some kind of warning before using it for the first time in a session? 
or perhaps having an preference option where a level gets automatically renamed if modified that way? (so it'd not overwrite the original one on disk, when saved). 🤔

hayoko:
Nice, personally I'm leaning towards the **strong renumbering** example since it seems more orderly.

gab3d:
@shun-iwasawa  

Regarding your proposed renumbering options:
I think I'm between **Weak** and **Medium** options, I consider Strong to be too intrusive.
(Usually, what you want is to minimize the chance of problems when working, and changing too much in an automatic way, will no doubt increase the chance that something unexpected happens.
Plus, if you really want to auto-renumber all the exposed drawings of a level, you can use the *Autorenumber* function at anytime.) 

Anyway, perhaps there could be a dropdown in the Preferences to choose the renumbering mode to use? 🤔

manongjohn:
I agree with @gab3d in that the more automatic it is, the more likely something unintended will happen. I believe there have been issues with renumbering in the past, even a crash. For that reason, i would prefer "Weak".

It would be good to have options, so I'd support a preference setting, defaulted to "Weak".

Aside: @beeheemooth suggested autorenumbering for `Rename`, `Step`, `Each` and `Time stretch`.  Perhaps this mode were made into a Preference setting ot could be expanded in the future to use it in these operations and any other operations that involve shifting cells.

konero:
Maybe weak would be my preference, it's really difficult to say!

My issue with naming was OpenToonz convention was a bit confusing, take this for example if we compare this scenario with Clip Studio...

- I create a drawing on frame 2
- Then I move the drawing to frame 3
- I create a new drawing at frame 2
- **What do I have?**

| **Fr #** | CSP | CSP Layers | OT | OT Level Strip |
| --- | --- | --- | --- | --- |
| **# 1** | 1 | 1 | 1 | 1 |
| **# 2** | 1a | 1a | 2a | 2 |
| **# 3** | 2 | 2 | 2 | 2a |

So OpenToonz is not referring to what's already in the x-sheet, it's just adding an appendix to the new drawing because 2 already exists in the level strip. I think it should detect which frame comes before it in the x-sheet and then use that with an appendix, so it's the right way round in the level strip and named properly in the x-sheet.

manongjohn:
Oh, other things to consider when using Medium or Strong options, what if a level is used in multiple columns and the same frames are exposed in different places? What happens to frames that are not exposed at the time it is renumbered? Are they pushed to the end?

hayoko:
Ah, I didn't think of that, in that case I would go with weak, strong would indeed cause some breakage.

But now that @konero mentions appendixes would that not work better if it was fixed? It's a common naming method in animation, no learning curve to it etc. 🤷‍♂

shun-iwasawa:
Thank you guys, then I will keep the renumbering behavior to "weak". 
Regarding making the other behaviors optional, I'll put it off to the future development ~~because of my laziness~~ as I think it would better to heve users' feedback to see if they actually need it.

@konero 
With this PR the drawings will be in order when you set `Enable Autorenumber` option. The drawings will be like as follows:

| Fr # | w/ Autorenumber | 
| --- | ---| 
| # 1  | 1 (unchanged) | 
| # 2 | 2 (new) | 
| # 3 | 3 (renumbered from the drawing 2) | 

Can you please let me know for reference, what does CSP create when you insert a drawing at frame 2 in the following arrangement?

| Fr # | CSP | 
| --- | ---| 
| # 1  | 1 | 
| # 2 | ★ | 
| # 3 | 1a | 

@manongjohn 
>Oh, other things to consider when using Medium or Strong options, what if a level is used in multiple columns and the same frames are exposed in different places? 

If the scene frame is selected, IMO the renumbering would be based on the current column. Other columns containing the same level will be just updated accordingly and may not be in order.

>What happens to frames that are not exposed at the time it is renumbered? Are they pushed to the end?

I think the unexposed-and-conflicted frames would be with suffix, having the number unchanged.

artisteacher:
I think I'd rather have autorenumber off by default, but the weak option does seem the least destructive.

gab3d:
@shun-iwasawa 
> Other columns containing the same level will be just updated accordingly and may not be in order.

Sorry, after renumbering is done, this means either:
A) The exposed drawings in other columns do not change, and their numbers are updated in the Xsheet?
or
B) The exposed frame numbers in other columns do not change, so the exposed drawing changes?

Apparently method A has no disadvantages, but in case it's B it should be good to alert users that there are other columns in the Xsheet with the same level exposed, so they can control them after the operation is done.

shun-iwasawa:
@artisteacher 
Regarding the default value of `Enable Autorenumber` , I agree there are both merits and demerits in each option. I suggested it to be ON by default based on the assumption that new users start using OT by drawing some animation in single column. There still be room for consideration.

@gab3d 
Sorry for not telling enough, I meant the method A, as long as the other columns belong to the current scene.

konero:
> Can you please let me know for reference, what does CSP create when you insert a drawing at frame 2 in the following arrangement?

@shun-iwasawa Here's what CSP does in that case. If a layer named 2 **does not** exist, it will be...

|Fr # | CSP (Timeline) | CSP (Layer Ordering) |
|-- | --| -- |
|# 1 | 1| 1 |
|# 2 | 2| 2 |
|# 3 | 1a| 1a |

If it **does** exist, it will be...

|Fr # | CSP (Timeline) | CSP (Layer Ordering) |
|-- | -- | -- |
|# 1 | 1 | 1 |
|# 2 | Layer 1 | Layer 1 |
|# 3 | 1a | 1a |

I had no idea it would do that, that was interesting.

shun-iwasawa:
@konero Thank you for the quick reply. Actually I was puzzled over a similar problem while working on this PR and yes, it's interesting to know their method.

gab3d:
@shun-iwasawa 
I'm in for defaulting it at ON (in Weak mode).
Reason being: it is meant for making users' life easier by resolving a conflict in a better way than currently, and it'd only really fulfill that means if it requires no conscious user intervention to operate.

shun-iwasawa:
@gab3d 
Yeah, IMO the important thing for setting the default value is to assume usage of new users. If it is demanded, OT can determine whether the user is new or not and separate the default values for them.

Well, I'll merge this by myself as there seems to be no malfunctions reported and I think it'd better to be tried in the nightly build for obtaining further feedback.

PR #2946, opened 2019-12-17
Use std::isfinite instead of _finite

otakuto:


RodneyBaker:
Thanks for the refactoring @otakuto 
LGTM

PR #2949, opened 2019-12-18
Level Settings Box on the Load Level Popup

shun-iwasawa:
levelsettings_in_loadlevelpopup

This PR will add ability to set level settings on loading levels by using the Load Level Popup.
It is useful when, for instance, setting appropriate dpi on loading images which have no dpi information.

- The `Level Settings` box is located next to the `Arrangement in Xsheet` options box which opens by pressing the plus (+)  button at the bottom of the popup.
- The level settings will be set to each loaded level at the same time, if the level type is allowed to set it.
- This feature will take effect only when the `Level Settings` box is visible when clicking the `Load` or the `Apply` button.
- This feature will override the settings in `Preferences > Loading > "Level Settings by File Format"`

### DIFF
--- a/toonz/sources/toonz/filebrowser.cpp
+++ b/toonz/sources/toonz/filebrowser.cpp
@@ -299,7 +299,7 @@ FileBrowser::~FileBrowser() {}
 
 //-----------------------------------------------------------------------------
 /*! when the m_folderName is edited, move the current folder accordingly
-*/
+ */
 void FileBrowser::onFolderEdited() {
   TFilePath inputPath(m_folderName->text().toStdWString());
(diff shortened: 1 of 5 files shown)


artisteacher:
Could we please have the arrangement options too?

RodneyBaker:
Checks passed but conflicts in filebrowser popup.

shun-iwasawa:
Rebased to the current master.

@artisteacher 

> Could we please have the arrangement options too?

Sorry but I'm not sure I could understand your comment properly but...
The arrangement option is already in the load level popup and it will also be kept in this PR.

RodneyBaker:
Thank you @shun-iwasawa

PR #2953, opened 2019-12-22
Fix Duplicate Frame numbering in Scene Edit mode

### LINKED ISSUE #2944: Frame numbering when using Duplicate Drawing
### Issue Summary

I want to pay attention to how two related functions work: Duplicate Drawing and Create Blank Drawing in Autocreation mode: Use Xsheet As animation Sheet (watch the video) In this case, the Duplicate Drawing function does not take into account the current frame number, but creates a duplicate with the following sequence number, that is, after applying the duplication operation, you must always apply Autorenumber so that there is no confusion in the Level Strip.

ezgif-6-35c34744afa1

### Expected Results

In my opinion, the Duplicate Drawing function should follow the frame numbering as well as Create Blank Drawing so that these two operations can be used harmoniously and logically.

### System Information
* **Operating System**: WIndows 8.1

manongjohn:
This PR fixes #2944 for duplicating frames using the xsheet. 

Modified the logic to do the following:
- No longer requires enabling preference `Sync Level Strip Drawing Number Changes with the XSheet`
- Uses Autocreation logic to determine when/how to create duplicate frames and number them according to mode
   - If Autocreation is `Disabled`, the logic will temporarily switch to `Enabled` to perform the duplication
- Allows you to select and duplicate multiple frames across multiple columns as desired.

Duplicating frames from Level Strip is not impacted by this change.

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -1254,6 +1254,45 @@ public:
 
 //-----------------------------------------------------------------------------
 
+class DuplicateDrawingUndo final : public ToolUtils::TToolUndo {
+  TFrameId origFrameId;
+  TFrameId dupFrameId;
+
+public:
+  DuplicateDrawingUndo(TXshSimpleLevel *level, const TFrameId &srcFrameId,
+                       const TFrameId &tgtFrameId)
+      : TToolUndo(level, tgtFrameId, true) {
+    origFrameId = srcFrameId;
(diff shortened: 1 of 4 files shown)


manongjohn:
Seems the Jenkins builds didn't kick off automatically.

RodneyBaker:
Thanks for the fix @manongjohn

beeheemooth:
Thanks! I hope that this will be implemented in the upcoming pre-releases.

PR #2954, opened 2019-12-23
Fixes on the Stop Motion Support

shun-iwasawa:
This PR modifies the stop motion support (#2635) which is just merged;
- Added license files of the `libjpeg-turbo` and the `OpenCV` libraries to comply with their licenses.
- Made OpenCV version flexible. Specifying locations of headers and libraries is now left to CMake.
- Removed `thirdparty/opencv2` folder as they can now be loaded from originally-installed location.
- Modified the `How to build` instruction accordingly.
- Added the translated content to Japanese version of `How to build` instruction.

@turtletooth 
I temporarily set the minimum available version of OpenCV to v4.1.0 (which I used for testing the feature). Please let me know if you have tested and confirmed working with older versions.

### DIFF
--- a/doc/how_to_build_win.md
+++ b/doc/how_to_build_win.md
@@ -72,7 +72,7 @@ Rename the following files:
 
 ## Building with extended stop motion support for webcams and Canon DSLR cameras.
 You will need three additional libraries.
-  - [OpenCV](https://opencv.org/)
+  - [OpenCV](https://opencv.org/) (v4.1.0 and later)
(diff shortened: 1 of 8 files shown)


shun-iwasawa:
@manongjohn 

> Stop Motion is only going to be for Windows?

Yes, as of now.
It seems that the Canon SDK package includes Macintosh version as well, so it has a potential to be available on MacOS in the future.

artisteacher:
> Stop Motion is only going to be for Windows?
Yes, as of now.

That's super disappointing. 😭

PR #2955, opened 2019-12-24
Fix Pasted Macro Fx Node Placement

shun-iwasawa:
This PR fixes a problem as follows:

1. In the Fx Schematic, create any fx (say, Blur fx) and a macro Fx. Connect the blur fx to the Xsheet node.
1. Right-click the macro fx and select `Copy` . 
1. Click the link between the blur fx node and the Xsheet node. Right-click and select `Paste Insert` .
1. The pasted macro fx node appears at the same position as the copied one.

I fixed this so that the pasted macro fx will be placed at next to the upsteam node of the link.


RodneyBaker:
Thank you @shun-iwasawa

Issue #2868, opened 2019-11-02
v1.4 (New Version) General Release


RodneyBaker:
@shun-iwasawa has approved the following general schedule going forward into v1.4:

**Timeframe**
- [x] Mid-December, 2019 - **FEATURE FREEZE**.  New feature PR posted after this will not be included
- [x] Two weeks for bug tracking & translation update
- [x] Beginning of 2020 - Release of v1.4 RC  (done : Jan 17)
- [x] Two+ weeks for urgent bug fix or releasing RC2 (if needed)
- [x] Late January 2020 - Release of v1.4 (done ! : Feb 1)

Comments or suggestions are appreciated.  Thank you!

RodneyBaker:
Checklist (pending)

RodneyBaker:
Change list (WIP)

Completed
- [ ] #2867 Revised vector selection pivot re-centering logic 
- [ ] #2866 Fix customize commands tree structure 
- [ ] #2859 Add "Hide Selected" Command to the Function Editor Spreadsheet 
- [ ] #2858 Enable Animate tool to control precisely by pressing the Alt key 
- [ ] #2855 Fix blank Style Editor panel 
- [ ] #2854 Fix bounding box computation in TXshSimpleLevel 
- [ ] #2848 Modify Maximum Values for Pinch and Magnet Tools Size 
- [ ] #2847 Fix Crash on Launch without Viewer 
- [ ] #2842 Fix pivot location on new/modified vector selections 
- [ ] #2839 Initialize Slider to Linear 
- [ ] #2838 Enable Pencil Mode for Toonz Raster Eraser in Freehand and Polyline types 
- [ ] #2837Selection pivot fix 
- [ ] #2835 Fix Travis macOS builds 
- [ ] #2832 Update Travis to xcode9.2 
- [ ] #2831Fix source comment typos 
- [ ] #2830Enable Geometric Tool for Motion Path 
- [ ] #2829Fix reading escaped double-quotes in quoted strings 
- [ ] #2825 Korean Translations 
- [ ] #2824 Nonlinear Slider for the Brush Size Options 
- [ ] #2823 Fix Main Window Full Screen Mode 
- [ ] #2821Fix lipsync popup layout 
- [ ] #2819 Fix Column Rename Field Position 
- [ ] #2816 New Fx: Corridor Gradient Fx Iwa 
- [ ] #2814 Stop Width Change on Style Editor on Maximize

shun-iwasawa:
[...]

PR #2957, opened 2019-12-25
Vector Guided Drawings Menu (modified by #2992)

shun-iwasawa:
This PR slightly modifies the UI and behavior of the `Vector Guided Drawings` menu which was recently updated ( #2931 by @manongjohn  ). Hoping this will not interfere the original intention of the feature..

- Replaced the exclusive items under `Vector Guided Drawing` by radio buttons, putting them into `QActionGroup` .
- Made the `Vector Guided Drawing` submenu and the `Hide ZeroThickness Lines` command to be displayed only when the current tool is for vector levels.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -836,7 +836,7 @@ QCheckBox:disabled {
   color: #808080;
 }
 .CheckBox::indicator,
-QMenu::indicator,
+QMenu::indicator:non-exclusive,
 QCheckBox::indicator,
 .GroupBox::indicator,
 QGroupBox::indicator {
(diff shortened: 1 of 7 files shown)


manongjohn:
@shun-iwasawa Thanks for tackling the UI refractor you mentioned with this PR. I have not tested it, yet, but did review the code for my own learning purposes, and it all looked good.

To make it more visible and easier to get to, I was actually thinking of moving this to the `View` Menu and make it work like `Shift and Trace`. It would bring up a pseudo tool options bar for the `Vector Guided Drawing` options. Just not sure I can finish it before the feature freeze for the upcoming release.

shun-iwasawa:
@manongjohn 
> To make it more visible and easier to get to, I was actually thinking of moving this to the View Menu and make it work like Shift and Trace. It would bring up a pseudo tool options bar for the Vector Guided Drawing options. 

It sounds very nice.

> Just not sure I can finish it before the feature freeze for the upcoming release.

I'm sorry to say,[ the feature has already been freezed](https://github.com/opentoonz/opentoonz/issues/2868#issuecomment-566981996) and we're in a phase of bug fixing and translation update.
~~So, all you need is to pretend your PR is bug-fix by just adding "Fix" at the head of the title. :-)~~
Joking aside, I think it's barely OK if it is ready before releasing the release candidate.

manongjohn:
@shun-iwasawa Thank you for this change. I did merge this into #2992 and made some modifications so it was better integrated with the new panel.

This can be closed if ok with you.

PR #2959, opened 2019-12-26
Fix Build Error

shun-iwasawa:
This PR will quick-fix the build error after introducing #2943 .
Again, CI checks failed to catch the error because they ran before merging of the changes causing the error ( #2953 ) .


shun-iwasawa:
Jenkins again

shun-iwasawa:
I'll merge this by myself as it is obvious.

Issue #2941, opened 2019-12-15
[Feature Request] Use Double-Click to Open Scenes and Levels


martinvanzijl:
### Issue Summary
We recently added the ability to use double-click to open a scene in the "Load Scene..." dialog (#2924). It would be useful to extend this to also include the following:

- [ ] Double-click on a scene to open it in the File Browser Window and Browser Room.
- [ ] Double-click to open level(s) in the Load Levels dialog.
- [ ] (Maybe) Double-clicking on any file type in the File Browser window to attempt to load it as a scene or level, depending on the file extension.
- [ ] Avoid duplicated code with the single-click function.

martinvanzijl:
Hi @manongjohn, for this one, what would be the test cases?

> Double-click on any file type in the File Browser window to attempt to load it as a scene or level, depending on the file extension.

PR #2961, opened 2019-12-26
Allow double-click to open scenes in File Browser and Browser Room

martinvanzijl:
For #2941. Demo below for File Browser Window:

file-browser-demo

And for the Browser Room:

browser-room-demo


martinvanzijl:
One issue with this is that in the Browser Room, it is difficult to tell that the scene is loading/has loaded. It would help if there was a "Loading Scene..." dialog with a progress bar. Has such a feature ever been suggested or implemented?

RodneyBaker:
The only progress bars I recall seeing is via the rendering progress dialog.
You might look into that.

As for this PR... LGTM.

PR #2966, opened 2019-12-29
Add new Help Menu options

### LINKED ISSUE #2958: Add New Links in Help Menu
I'm thinking about making a **What's New** section in the OT Manual, where all the Release Notes from each successive version were on display.
I also think it'd be great if the user could click on a **What's New** entry in the **Help** menu inside OT, to access directly to the correspondent section in the Manual to that version of OT.

I guess this would improve users' experience when upgrading to new versions, as well as would also help raising visibility of Release Notes, which currently are somewhat difficult to access once launch time has passed.

What do you guys think? 🤔

manongjohn:
This PR resolves #2958

The following new `Help` menu options will be added

`What's New...` -> https://github.com/opentoonz/opentoonz/releases/latest
`Community Forum...` -> https://groups.google.com/forum/#!forum/opentoonz_en
`Report an Issue...` -> https://github.com/opentoonz/opentoonz/issues

image


gab3d:
Please make the URLs translatable, so we can point to forums / new features list, etc. in other languages.

manongjohn:
The only one that can be translated to a different language page would be the `Community Forum` link, which there is currently only a Japanese forum, I think.   The `What's New` is the release notes for the latest release which will be in English and the `Report an Issue` just goes to the Issues page in github.

I do have some concerns about going to specific language forums as the main support is in English and that would likely be the best place to get help.

gab3d:
> The only one that can be translated to a different language page would be the Community Forum link, which there is currently only a Japanese forum, I think.

There is also a [Spanish forum](https://groups.google.com/forum/#!forum/open-toonz--comunidad-en-espaol) too.

Regarding the **What's New** option, I will translate the Release Notes (as I always do) and this time I would like the What's New from Help menu to point to: 
https://opentoonz.readthedocs.io/es/latest/whats_new.html
(Of course the English version of the Notes would also be placed in 
https://opentoonz.readthedocs.io/en/latest/whats_new.html 
...either if the program links to there or not)

> I do have some concerns about going to specific language forums as the main support is in English and that would likely be the best place to get help.

I have mixed feelings on this... but yes, I understand your point. 
All in all I guess all users know that by default there will be English forums somewhere, what they would not know for sure is if there are similar forums in their language. 
Also, please consider that (for example) both OpenToonz's UI and Manual is fully in Spanish, so there could be a good case for people wanting to make use of OT without knowing a single word in English. So that forum, abide having less movement, could well be a valuable source for them. 🤔 

In particular, my personal goal while translating OT and its Manual into Spanish, is that *"a little kid, in the middle of nowhere in Central/South America"* could find in OT the means for developing his/her artistic inclinations for animation, with the sparse resources he/she would most probably count on (both material and educational). 
In that sense, I can't imagine myself asking that kid to go to an English forum to see why the MyPaint brushes don't react to the Size parameter of the Brush tool, the same way as with other brushes.
If you know what I mean... 🦄

RodneyBaker:
In the interest of making sure this PR gets into the Release Candidate I'll press forward with merging this.    Allowing for translation is important but we don't want to hold the feature back for that alone.

As for the URLs, I understand the default for programmers would be to keep them hard coded but from an open source usability standpoint the ideal location for the URLs would be to have them available for editing via Preferences.  Then users can customize the location as necessary to allow for translations, localizations and/or apply some form of language pack to update Preferences.ini.

As for these links (and this PR)... outstanding. 
Thanks @manongjohn for the addition!

shun-iwasawa:
First of all, thank you @manongjohn for adding the links.
Sorry for my late response, but I'm concerning about the following points with this PR:
- Regarding `Community Forum...` link, just like the situation @gab3d pointed, there aren’t a lot of people that can speak English in Japanese animation industry. I think it'd be friendly for them to navigate to the forum of their language.
- Regarding `Report an Issue...` link, I'm a little disinclined to navigate users to github issues page, because it implies that users need to create the github account before posting a bug report and the report should be written in English. IMO this can be replaced by the link to `TroubleShooting` category in the forum of each language, since I think the Google account is more common between users.

manongjohn:
@shun-iwasawa 

I didn't get a chance to make the forum link translatable before it was merged. I will submit another one to correct it.

As for reporting issues...maybe it should be called `Report a Bug` instead so as not to confuse it with general usage issues and questions.

Fyi, other software that i've used that directed me to similar bug reporting sites required me to create a login.

Additionally, in my opinion, all bugs should be reported in English in a common location so any developer can look at it and work it.

Putting issues on a different site is going to be a pain and likely forgotten as development is done in github. I don't generally go to the forum for issues to work and most issues are tracked and resolved better here, in my opinion.

As it is, feature requests hidden or elsewhere tends not to be remembered nor look at. When was the last time a developer looked at Closed Feature Requests on github? Out of sight, out of mind.

If you require we change issue reporting to point to the forum, then I can make the change when I make the forum translatable. Just let me know.

RodneyBaker:
A few thoughts related to @shun-iwasawa's concerns:

Community Forum
Agree, it would be good to find an optimum way to match the Community Forum link with the users language.  Short of sending user first to a landing page (i.e. in the online documentation) I don't have any specific suggestion to achieve that result. 

Regarding Bug Reporting
Having the link go to github seems most appropriate for a number of reasons and also as the bar to actually reporting a bug is higher than that of troubleshooting allows users to consider carefully their pending report.  

Emphasis on Community
Taken as a whole the primary link of interest in the Help menu is that of the Opentoonz Forum and users who frequent the Opentoonz forum will know where to post discussions related to troubleshooting and from there they can be assisted in submitting formal bug reports.

I did originally suggest the link for submitting a bug report via Help menu point to the Troubleshooting forum as I felt the more reports generated there that open forum the better.  I am easily convinced either way however and will adapt accordingly.  An additional issue with changing the target for reporting away from github will be that as with the Community Forum it would need to parse for language preferences.  That doesn't seem realistic and as such the link to github for formal reporting is preferred.

Possible solution for all cases
An option that might suit all cases and interests would be to expose the URLs in Preferences so they can be customized by the user (or modified by a language pack/additional installer).  As can be anticipated, this also would have a potential downside in that users could more easily opt to forego the preferred/official locations.

gab3d:
**Forums**
Well, as previously stated, I'm in favor of making this link translatable, so for the languages for which there currently are running community forums, we should point the link there. In case there isn't such a forum for a particular language, the link should remain "untranslated", pointing to the English forum.

Alternatively, I don't see the option of pointing this link to a landing page in the Manual as a *bad idea*, since in that landing we could offer more than one choice to users: i.e. Google forum, Discord forum, Facebook group, etc. 
What I don't know yet how to handle is what happens with languages which doesn't yet have a translation of the Manual (most of them), but do have local forums/groups? 🤔 

**Bug Reporting**
I'm with @manongjohn in this one... besides acknowledging that there is some sort of obstacle in having to make a GitHub user to be able to report a bug, I think this should raise the probability that reports would come from people that are willing to make a follow-up of their own reports, and not just drop the bomb and disappear. In most other software that I've used, this kind of action always requires some sort of previous registration too.
I also understand that, for developers, it's easier/more productive to have a central location where to look for bugs, while deciding in what to spend their developing time; not to mention that in forums there could more probably come reports in other languages unknown to developers, effectively making them useless.

**Links editable in Preferences**
I've never seen this before in any software.
While it could have its advantages to some organizations with in-house developers, I think it won't help us get as much valuable feedback as possible, since those links could potentially point to anywhere beyond our sight/control. 🤔

hayoko:
Aye, doesn't make sense to make them user editable, you just translate the string.

I don't know if this is necessary so feel free to completely ignore this, but it's nice when an external link is marked by some kind of icon to separate it from other user actions, otherwise "Community..." and "About..." there's no telling which is a link and which is a software action. There's no Unicode for it, because of reasons I guess, but maybe U+1F855 can work?

**Community...** 🡕
**About...**

Since it looks similar to the standard icon for it...

image

Can't really use an icon, because Mac... 😞

gab3d:
it seems good to me.

PR #2975, opened 2020-01-04
Fix "Merge Levels" Command to Update Thumbnails

shun-iwasawa:
This PR fixes the Merge Levels command to update thumbnails properly.
Related to the command, it also fixes & refactors in the following points:

- Made the dialog message ( asking "Group strokes by vector levels?" added in #2762 ) translatable.
- Made the command to be canceled if users close the dialog by pressing a "x" button.
- Simplified the function detecting that the selected column contains vector level.
- Added a proper history string for vector merging operation.


RodneyBaker:
Something that needs to be addressed in the future with Merge Groups is the restriction that prevents frames with one stroke from being merged.  There is no easy way for a user to discover this restriction and therefore the tendency when first exploring might be to think that Merging doesn't work.  Perhaps there is a good reason for the restriction (which I believe requires three strokes)?

PR #2976, opened 2020-01-04
Make Stop Motion Texts Translatable

shun-iwasawa:
This PR will make the following UI texts translatable:
- Messages in the warning popups ( `Please start live view before capturing an image.` and `Cannot capture webcam image unless live view is active.` )
- The `Start Live View` button text.

This PR also fixes the part of checking the tab labels to compare with the translated strings.


RodneyBaker:
Thanks!

PR #2982, opened 2020-01-05
Use std::sort instead of obsoleted qSort

otakuto:
qSort is obsoleted.
We should use std::sort.

https://doc.qt.io/qt-5/qtalgorithms-obsolete.html#qSort


RodneyBaker:
Thanks @otakuto!

PR #2984, opened 2020-01-07
Fix Saving-Purposed File Browsers Not to Clear Name Field on Moving Folders

shun-iwasawa:
This PR will revert the following behavior in File Browser popups, which has been unavailable after introducing #2949 :
- When switching folders by clicking the folder item in the right panel, previously-selected item's name remains in the name field, so that users can reuse it for inputting the file name. It was useful, for instance, to manually save backups with the same file name, in another location. 

Please note that now the above behavior becomes availble only in "saving-purposed" popups ( like the `Save Scene As` popup, the `Save Level As` popup , etc. ). For loading-purposed ones still clear the name field on moving folders because they are basically used by selecting existing items without typing the name field, and some of them (such as the `Load Level` popup) need to update UI along with the item selection.


RodneyBaker:
Thanks Shun!

PR #2987, opened 2020-01-07
Fix Crash on Loading XDTS and Other Improvements

shun-iwasawa:
This PR fixes several problems & improves regarding XDTS input as follows:

- Fixed crash on clicking `...` button in the file field of XDTS popup when loading XDTS file for the second time.
- Fixed OT to load XDTS properly when it contains data in negative frame. Actually containing data in negative frame is not officially allowed in XDTS specification, but Clip Studio Paint may export such data.
- Fixed the frame range of flip console to be properly updated on loading.
- Swap positions of `Load` and `Cancel` buttons to be consistent with the file browser popup.
- Made XDTS popup to suggest the level locations not only in the same folder as the .xdts file, but also sub-folders with the same name as levels (like `$scenefolder/A/A..png` ). Because CSP exports level images into sub-folders.
- Made the column names to be the same as containing level names.
- If the `duration` value in XDTS is smaller than the maximum frame number where the data actually input, such duration is now set to both in preview range and output range.

### DIFF
--- a/toonz/sources/toonz/filebrowserpopup.cpp
+++ b/toonz/sources/toonz/filebrowserpopup.cpp
@@ -2258,13 +2258,15 @@ void BrowserPopupController::openPopup(QStringList filters,
   m_browserPopup->initFolder(TFilePath(lastSelectedPath.toStdWString()));
   m_browserPopup->setFileMode(isDirectoryOnly);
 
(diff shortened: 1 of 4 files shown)


RodneyBaker:
Thanks!

PR #2992, opened 2020-01-08
Add Guided Drawing Controls Pane

manongjohn:
This PR moves the Vector Guided Drawing actions from the Viewer right-click context menu to the following locations

1. New entry in View Menu
image

Command bar/Shortcut entry for this actions can be found under `Menu Commands` -> `View`

2. New entry in Windows Menu
image

Command bar/Shortcut entry for this panel can be found under `Menu Commands` -> `Windows`

3. New Guided Drawing Controls pane
image

Command bar/Shortcut entries for the buttons can be found under `Misc`

Added Commands: `Flip Next Guide Stroke Direction` and `Flip Previous Guide Stroke Direction` 
- Works only with Farthest and Closest options


gab3d:
That's great!

RodneyBaker:
Looks very good thus far.
I won't have time to thoroughly test this until later today but a question...
Does #2957 need to be merged before this PR is implemented?

manongjohn:
No. #2957 will no longer be needed if this is merged.

I have an upcoming commit pending for this PR so don't merge just yet. :)

gab3d:
Thanks for the stroke flipping command!
That's awesome! 😺

turtletooth:
What about people who want to tweak guided drawing without a full pane?

gab3d:
I see this, much like in other program is the Align & Distribute panel, something that you don't have to maintain always opened or docked, but instead use it as something temporal, that you open while working on guided drawing and then dismiss it.

If a button for opening it could be placed somewhere in the UI (command bar, xsheet bar, even in the viewer itself), then it would be just a click away. (a shortcut could also be a good way, obviously)

The right click alternative also works, of course, but honestly with so many options available the workflow gets a little cumbersome...

manongjohn:
@turtletooth I can certainly bring back the right-click menu option back. What options would you want to see there? I don't think all then new options need to be available in the context menu.

RodneyBaker:
@manongjohn 
In part to address some of Jeremy's concerns:
Would it be possible for the 'Vector Guided Drawing' option to automatically be toggled on when the 'Guided Drawing Controls' window is activated?

Rationale in two parts:  
1.  Both options have to be on in order for Guided Drawing to work (I kept forgetting to turn one or the other on and had to remind myself)
2.  This would potentially remove the requirement to have 'Vector Guided Drawing' in the menu (as is in this PR) so it could return to the Right click menu.

turtletooth:
I'm not actually using OT much lately and don't really use the guided drawing, but I can foresee someone wanting to quickly change for example farthest to nearest without opening a full panel.  But since I'm not actually using it much, I should probably be quiet and let those who do use it give their thoughts on an ideal workflow.

manongjohn:
> I can foresee someone wanting to quickly change for example farthest to nearest without opening a full panel.

It's a valid point. For quick changes, I'd probably also keep Auto Inbetween and Interpolation settings, but not add the buttons unless requested.

> But since I'm not actually using it much,
*gasps* Blasphemous! :)

> I should probably be quiet and let those who do use it give their thoughts on an ideal workflow.
It's all good.  You originally implemented Vector Guided Drawing and had ideas on how it would/should work so I say that gives you some authority on the subject.

beeheemooth:
Personally, I really like the proposed panel. Is great that it can be fixed. Thanks @manongjohn !

shun-iwasawa:
Jenkins again

manongjohn:
Restored right-click menu options, though left out buttons for now.

Can now turn guided drawing off from panel by setting Guide Frame to `Off`

shun-iwasawa:
Jenkins

shun-iwasawa:
Currently the flipping stroke operations seem not to make the level dirty and are not undo-able.
I'll merge this anyway as it can be separated to the future issue.

Thank you @manongjohn for the addition! LGTM

PR #2994, opened 2020-01-08
Fix Crash on Reordering Nodes Containing Grouped Camera Node

### LINKED ISSUE #2978: Stage Schematic: Reordering nodes causes a crash if a camera is grouped
Video Example:
OT Grouped Camera Crash

Repro steps:

1. Create a group-able node e.g. a level or a pegbar
2. Select the created node and the camera node and group them (the option to group does not appear on the camera node)
3. Click the Reorder Nodes button.
OT will crash.

The crash will only occur if the group is closed.

Edit: Whoops, messed up the title

shun-iwasawa:
This fixes #2978


RodneyBaker:
Confirm this fixes the issue.  LGTM

PR #2995, opened 2020-01-08
Fix Phantom Fill Check

### LINKED ISSUE #2902: Phantom Fill Check
### Issue Summary

When you collapse multiple columns in sub-xsheet, empty columns with the same numbers still show the “Fill Check” data for the old columns. If you create new columns in place of the old ones, then in addition to the closed areas of the current level, they will also display gray shading of those levels that are in the sub-Xsheet in the columns with the same numbers.

ezgif-1-0cd537627a3d

Moving columns inside sub-Xsheet also affects columns outside sub-Xsheet.

ezgif-1-49504e7ffd04

### System Information

shun-iwasawa:
This Fixes #2902 

I fixed logic for defining the current column. When the columns are collapsed in subxsheet, the index of subxsheet column in the parent xsheet will be valuated instead of each column index in the subxsheet.

### DIFF
--- a/toonz/sources/toonzlib/stage.cpp
+++ b/toonz/sources/toonzlib/stage.cpp
@@ -115,11 +115,11 @@ void updateOnionSkinSize(const PlayerSet &players) {
 bool descending(int i, int j) { return (i > j); }
 
 //----------------------------------------------------------------
-}
+}  // namespace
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thanks Shun!

PR #3001, opened 2020-01-08
Use std::isnan instead of _isnan

otakuto:


RodneyBaker:
This PR likely needs a Jenkins refresh.

RodneyBaker:
Thank you @otakuto

PR #3002, opened 2020-01-08
Remove tgraphicss.cpp

otakuto:


shun-iwasawa:
Let me check Jenkins CI for Windows is alive.

Jenkins again

shun-iwasawa:
Hmm it seems to hang. I'll fix it tonight.

RodneyBaker:
There is some interesting code in tgraphics.cpp.
For instance, the ability to draw specific vector shapes; diamonds, cross, etc. in addition to standard shapes.  While I must assume it was deprecated... it's not clear to me why this code (as well as some other used code) was dropped.

shun-iwasawa:
LGTM Thank you @otakuto !

PR #3003, opened 2020-01-09
Fix Viewer to Unlock After Stopping Playback

### LINKED ISSUE #2817: The brush does not unlock after stopping playback.
### Issue Summary

Stopping the animation with a shortcut ("Loop") blocks the brush. The lock disappears if you remove the cursor from the canvas and return it back.

ezgif-4-380f685ffae6

### Expected Results

After the animation has stopped playing with the shortcut, the brush should be available for use.
### Actual Results

The brush is locked and you cannot draw with it
### System Information
* **OpenToonz Version**:  latest

shun-iwasawa:
This fixes #2817 .


RodneyBaker:
LGTM

Issue #2989, opened 2020-01-07
[DRAFT] Posting Guidelines (1st time posters, please read!) [DRAFT]


manongjohn:
_This is a draft of a notice targeting 1st time posters. This issue is just to discuss suggestions and revisions.  Any feedback is welcome. When ready, I will submit a new Issue that will be Pinned and locked for comments._

### Guidelines for Posting Issues
----
- Please use this site for posting software issues, bugs and crashes only.

- Posts of the following nature should be redirected to the appropriate category in the official [OpenToonz Google Group](https://groups.google.com/forum/#!forum/opentoonz_en)

	- Feature and Enhancement requests 
	- How-to questions
	- General questions about OpenToonz

- Before posting an Issue:

	- Search Issues to see if someone has already submitted the problem or something like it.  If you have additional information, feel free to add to the discussion or just let us know you're having the same issue too.
	- Make sure you are running the latest Official version or the most recent Nightly version of OpenToonz.
		- If you are using an OpenToonz variant (Morevna, OTX, other), download the most recent Nightly and see if it can be duplicated there because the issue you are experiencing may be specific to that variant.
	- Gather as much information as you can regarding the bug/issue.  Examples of things you can provide:
		- Tell us what version of OpenToonz you are on especially it isn't the current official one.
		- Tell us what Operating System you are on and whether you are using a tablet, using a pen or a mouse, etc
		- Reliable steps to reproduce.
		- Any system logs related to a crash. (Windows Event logs, macOS crash logs, etc.)
		- Screenshots.
		- Animated jpg/gifs or links to videos showing the problem.
		- A sample scene, level file or project illustrating the problem, if possible.

- General conversion

	- Try and be constructive in feedback in Issues and Pull Requests.
	- Don't be rude.

Thanks,
OpenToonz Community

RodneyBaker:
Looks good.
One piddly thing...
Perhaps you mean 'conversation' instead of 'conversion' in that last bullet point?

hayoko:
I reworded your post a little to sound more formal and compacted some info to make it a lot quicker to read. Feel free to use, edit or discard. I think you can grab the post layout by quote replying.

_(Although I personally dislike disallowing feature requests here.)_

---

# Posting Guidelines
Only bug or crash reports are accepted here.

The following kinds of posts should be made in the official [OpenToonz Forum](https://groups.google.com/forum/#!forum/opentoonz_en).
- Feature requests
- How-to
- General discussion

## Before You Post
[...]

PR #3006, opened 2020-01-10
Update GitHub templates

hayoko:
This is to accompany what's being discussed in #2989.

Updates the issue templates to use the newer model in GitHub, it will allow contributors to choose a bug report, feature request or be sent to either the English or Japanese user forum when they click new issue.

The links are configurable in the `config.yml`, I believe they are correct.

Please note, I made it so when a feature request is made it will automatically label it.

---
image


shun-iwasawa:
@hayoko 
Thank you for the proposal. I tested the changes in [your branch](https://github.com/hayoko/opentoonz/tree/update-github-templates). I didn't know this feature in github issue template.
Although it provides a option to Japanese user forum which I originally demanded for non-English speaking users (thank you!), this way seems to require to sign in / sign up to GitHub account before opening the options.
For Japanese users, I think for this version I'll just add notifications in the translation text of the `Report a Bug` popup, suggesting that Japanese user forum is also available.

RodneyBaker:
Recommend the Feature Request option (via this PR) be implemented for a test period to determine pros versus cons with a goal to adjust as necessary.  A monthly review identifying what (if anything) related to feature requests is not operating optimally as desired can be conducted with changes to follow.  If after initial testing it is determined use of github for feature requests is not ideal then the Opentoonz community can be notified.

As noted by @shun-iwasawa, the translation capability for links allows for some customization for directing language specific assistance.  This helps alleviate one important obstacle in that not all Opentoonz user can communicate their requests in the English language.

To take a slightly contrarian view for a moment...
If implemented, all involved should be prepared for the higher volume of github issues being opened.

hayoko:
Oh, yeah that's fair enough. I swapped the title for the label for feature requests, I've not used this before so I hope it works, if a user makes a feature request it gets auto labelled, this will make feature requests filterable and since they will have the label anyone can just ignore them if they want.

For example, in the search field type `is:open is:issue -label:"feature request"` and you won't see feature requests anymore in the feed, so being swamped by requests is no big deal. I do this for some projects and bookmark the URL.

But you will need to rename the '**Feature / Request**' label to '**feature request**' in lower-case.

Actually, I would urge renaming all labels to lower-case and never use special characters, I made a list of labels that would be good in [my branch](https://github.com/hayoko/opentoonz/labels) you can view, also please use muted pastel colors for labels, don't use black or primaries, it's harsh and too masculine.

BTW having a list of feature requests attracts people entering things like GSoC.

Just my advice, with love. ❤️

RodneyBaker:
I'm easily convinced...
- [x] Rename the 'Feature / Request' label to 'feature request' in lower-case.
- [ ] Rename all labels to lower-case
- [ ] Never use special characters
- [ ] Don't use black or primaries, it's harsh and too masculine.

Most labels have changed back and forth a few times depending on how often someone has put forth a preference.  

Most labels are now all lowercase with a few outliers that we'll either change to lower case, leave as is or re-label to something else.

The blacks and primaries were intentional... even impersonal... but largely based on best practices.  If necessary we can discuss the rationale behind each.

All labels are subject to change... and wind direction... to best meet the needs of the Opentoonz community so we can update them as convenient or necessary.

manongjohn:
I believe I remember reading a long while ago that Bountysource looks for `Won't Do` label on closed issues to determine whether or not the funds should be automatically reverted back to their owners. Not sure if it is case sensitive.

hayoko:
I had a look at their faq, they write if it's marked "won't fix", Git's default label is written "wontfix" so it'd be an oversight if they didn't account for that, but I put nothing past them, their site crashes more often than me drunk on an ice rink.

> Never use special characters

I worded that wrong, you should when appropriate, but not like there is now, like build / install are not the same thing and feature / request makes no sense.

> As for you personal preferences in color scheme for labels... can we respectfully pass?

If you like, but the best practice btw is the tried and true which is built into Git if you tap that refresh button. 😋 

image image

artisteacher:
I think it would be helpful for the bug reports to have a section for crash reports - possibly combined with the screenshots section.

RodneyBaker:
@artisteacher said:

> I think it would be helpful for the bug reports to have a section for crash reports - possibly combined with the screenshots section.

It might be good to have a checklist where the reporter could check if the issue relates to:

- [ ] Bug
- [ ] Crash
- [ ] Other with reproducible steps as detailed below
- [ ] Difficult to reproduce

Note that I'm not suggesting these would be the actual checklist items but something along those lines.

I would have suggested this sooner but I don't know of any way to get the checklist to run horizontally which would be ideal as the checkboxes could be the first thing seen at the top of every report.  Then the reporter and others could really hone in on the current status and nature of the report.  

An example of this checklisting can be seen in reports tagged as Known Issues where an extra advantage of checklists in the initial posting is that the counting of checked vs unchecked options is included automatically near the title of listed reports.  Example:  2 of 5, 5 of 5.  This makes it a useful means of measuring at a glance the nature and status of any given report.

See this know issues listing for example of githubs automated checklist counting at work:
https://github.com/opentoonz/opentoonz/issues?q=is%3Aopen+is%3Aissue+label%3A%22known+issue%22

The reporter of course is able to remove any boilerplate text from the templates.

manongjohn:
These checklists are annoying and frankly an eyesore. I want to see the issue first, not checklists. I shouldn't have to scroll down to read a short description of an issue. In some cases the checklist takes up more space than the issue description.

Move them to bottom if you insist on using them for administrative purposes.

hayoko:
@RodneyBaker Checklists would over-complicate things, helping to report a problem should be super-duper simple for anyone with no work involved.

You don't need to mix capitilization of labels, not even for acronyms, just choose one, I can change the label here if you're set on capitalization. Also consider pastel post-it colors, with all the black this repo just doesn't look as much fun as other repo's.

By the way feel free everyone to start an issue on my repo if you want to test the feature request auto label thing before activating it here to see how it works, I don't mind at all and can just delete them later.

https://github.com/hayoko/opentoonz/issues/new/choose

manongjohn:
@RodneyBaker Even though i see you moved them to the bottom as I requested, can you please get rid of these checklists?  I'm sorry, I know you mean well, but I just find this is making a lot of unnecessary noise in the body of the issue when trying to look at them.  

In some cases you are already assigning labels of the same name to them (known issue, workaround) making those checklist entries redundant.  For some things there are other means of identifying this information without even opening the issue like Developer Assigned.  Developers can assign issues to themselves when they are working on it which can be see from the issues page.  If you know a developer is working on an issue and it is not assigned, assign it to them, especially if they put in a PR for it already. Bountysource tags and updates short descriptions if there is a bounty on it.  If not, and there is a bounty by some other means, add the label. 

In all honesty, as a developer, none of the entries in the checklists helps me at all when reading the issue.  I don't need to see any of this when I'm trying to understand if I will work on the issue or not.  

If any other developer working issues feels otherwise and wants to keep it, then I'll begrudgingly let it be, but my preference is to have these removed and leave the issues as they are.

And, please, don't go label happy as a response.  I don't need to be blinded by a myriad of different colors when looking for issues to work on. :)

@hayoko I love the New Issue chooser you are proposing. Really nice and clean. I would actually suggest changing OT's `Report a Bug` link to that if it gets implemented.

RodneyBaker:
> none of the entries in the checklists helps me

@manongjohn Please try to understand.  These checklists are not for you.
I know you have a lot of work to do and do not like distractions.  
No one does.

Please note that you have full contributor permissions but many assisting here and working through issues do not.  A few months ago there were 500 open issues that almost no one was reviewing.  Presently there are less than 300 and while checklists weren't a major part of that the same basic methodology was used.  

As we work to reduce the number of open issues there will be little or no need for checklists and less need for labeling also.

One of the next things I plan to do is ping everyone who has submitted a report to encourage them to review their report in light of the release of v1.4.  

Having said all this... I aim to please and will plan to minimize checklists further and eliminate them if possible.  Currently the only checklists I have added (other than those others have added) are those in the Known Issues category.  The reason for the added checklist is that many of those are stagnant... more than other reports.  When the issues hit a wall and can't move forward the checklists serve a purpose... to let everyone know why the issue isn't moving forward.

If you somehow thought checklists were appearing in every category and not just Known Issues... Please disregard all of the above.  :)

RodneyBaker:
What needs to be done to move this PR forward?

konero:
This seems fine to me, better than the current template, it seems complete so not sure why it's labelled hiatus? Is there something you're waiting for?

RodneyBaker:
> Is there something you're waiting for?

Indications of life and/or interest?

konero:
Not sure I follow, but I posted a modified version of this here: #3244 if OP isn't updating it. But as for interest, it'd be good to update since it's the newer template model anyway.

RodneyBaker:
Closing in light of ##3244 moving forward.

PR #3013, opened 2020-01-11
Fix Crash on Applying "Adjust Current Level to This Palette" to Palette Level

### LINKED ISSUE #3007: 'Adjust Current Level to This Palette' command crashes with Palette levels
Here's a somewhat obscure reproducible crash

Video example:
Palette Crash

Repro steps:
1. Load a palette into a column (click and drag it onto the XSheet/Timeline).
2. With the newly created palette level selected, right-click on a palette in the Studio Palette window and choose the 'Adjust Current Level to This Palette' command.
OT will then crash.

OT Ver: Jan 10th Nightly
OS: Windows 10

shun-iwasawa:
This fixes #3007


RodneyBaker:
Thanks!

PR #3017, opened 2020-01-12
Revised English Fx Help files by gab3d

### LINKED ISSUE #3015: Revision of English Fx Help files.
@manongjohn @shun-iwasawa 
As suggested by @RodneyBaker in #2761 here I upload a revision I did of the previously translated English help files for FXs.

The revision was made for improving the meaning/wording/layout of the previous English translation, that was many times confusing.

Here is a sample of the kind of revisions made to the help files:
in **PerspectiveDistortIwa** help, *Overview* section:
**Previous:**

> This effect is for generating the movement of the ground at the time of Follow (Front Moves Fast, Back Moves Slow = Motion Parallax).
Since the deformed trapezoid is stretching the picture to the left & right, there is no vertical distortion.
In other words, the contraction in the vertical direction by the depth must be incorporated in the pre-input image.
Also, if you want to apply this effect to the background to parallel move, the movement of the background must be folded into the SubXsheet.

**Revised:**

> Allows to generate a motion parallax movement effect on the Source image, where things in
front move faster than things in the back.
The image width is stretched to the sides, to transform it into a trapezoid, but it will not
be stretched vertically.
In other words, the contraction in the vertical direction caused by depth, should be
incorporated beforehand in the image.
Also, if you want to apply this effect to a moving background, the movement of the background
must be collapsed into a SubXsheet.

Or the **CrossDissolveIno** effect help, *Clipping Mask* option description (also present in a lot of other effects):
**Previous:**

> When ON,
Material will not exist in the location (Back) (Alpha Value Zero), and remains clear.
The default setting is OFF.

**Revised:**
> When ON, it will take the Back port alpha information to define the alpha of the
resulting image, leaving areas where the Alpha value is zero as transparent.
The default value is OFF.

In the attached zip files you can find PDF, HTML and TEX files for the English help files.
I've introduced a couple of minor fixes in some of them before uploading them here, so please use these ones.

Thanks.

[english docs (tex) v2.zip](https://github.com/opentoonz/opentoonz/files/4050213/english.docs.tex.v2.zip)
[english docs v2.zip](https://github.com/opentoonz/opentoonz/files/4050214/english.docs.v2.zip)

manongjohn:
This PR resolves #3015

Thanks @gab3d for the updates.


RodneyBaker:
Thanks!

PR #3019, opened 2020-01-14
More Translation Update for V1.4

shun-iwasawa:
This PR updates translation (.ts) files for last-minute changes.
It also updates Japanese translation file.


manongjohn:
@shun-iwasawa Is it possible to make the Default Room names translatable?

shun-iwasawa:
@manongjohn 

> Is it possible to make the Default Room names translatable?

I haven't tested yet, but preparing some `QMap` table associating the default room names and "tr"-ed names and when creating the room tabs replacing the name by translated ones would probably work?

PR #3023, opened 2020-01-15
Fix Studio Palette Hang Up on Clicking Item in the Tree View

shun-iwasawa:
This PR fixes the following problem in the Studio Palette:

- When clicking the item in the tree view, if you move any pixels while pressing the mouse button, the studio palette hangs a while.

The problem was due to starting a drag and drop operation.
I modified it so that the drag and drop starts only when user moves the cursor a certain distance ( `QApplication::startDragDistance()` is set to 10 pixels by default )  with a button held down.


RodneyBaker:
LGTM

PR #3024, opened 2020-01-15
Fix UI Font Colors

### LINKED ISSUE #3020: Font colors in the default UI theme.

shun-iwasawa:
This fixes #3020 

- FIxed the Plastic Skelton items to be the same color as others.
- Fixed the Export Scene popup tree view to share the same style sheet with the one in the File Browser. (The object name in the style sheet is renamed from `DvDirTreeView` to `StyledTreeView` and shared by both widgets.)
- Fixed the Studio Palette tree item not to be black.


RodneyBaker:
Thanks!

PR #3029, opened 2020-01-16
Fix Panel Rearrangement

### LINKED ISSUE #3022: Style Editor rearrangement bug
GIF example:
OT Style Editor Bug

Repro steps:

1. Create a new room
2. Create any window, (the example gif uses a Combo Viewer) and make it fill the whole space
3. Create a Style Editor window and try to place it above or below the Combo Viewer

The Style Editor window will not be placed correctly.

Edit: Upon further testing, it seems that step 2 is not necessary, all that is needed is to create a Style Editor window and attempt to make it fill the whole screen.
The only difference is that if you only have the Style Editor window, it is possible to drag it again and return it to a working state. However, from that point onward, you are unable to snap any other windows.

OT Ver: Jan 15th Nightly
OS: Windows 10

shun-iwasawa:
This fixes #3022 .

Before temporalily fixing the panels width, I added a check if all the horizontal alignment of the panels in the main window are to be fixed.
And OT actually set them to be fixed width only if there is at least one normal (shrinkable) panel in the alignment.

### DIFF
--- a/toonz/sources/toonzqt/docklayout.cpp
+++ b/toonz/sources/toonzqt/docklayout.cpp
@@ -201,9 +201,9 @@ QSize DockLayout::maximumSize() const {
 
 QSize DockLayout::sizeHint() const {
   QSize s(0, 0);
-  int n        = m_items.size();
+  int n = m_items.size();
   if (n > 0) s = QSize(100, 70);  // start with a nice default size
-  int i        = 0;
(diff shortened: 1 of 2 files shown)


shun-iwasawa:
I'll merge this by myself. If there are any problems I'll tackle with it during the release candidate period.

Issue #2292, opened 2018-09-14
Mac build issues 


artisteacher:
### Issue Summary

Following the build instructions for MacOS either only partially works or results in a crash on open. I’m wondering if the instructions need to be different for High Sierra? Any assistance would be appreciated!
### Screenshot or Video Reference

<img width="300" alt="1a44c0b1-65ca-4f22-a2d2-1b9c195f0167" src="https://user-images.githubusercontent.com/21159116/45524428-77f31280-b79c-11e8-9140-83814ae5399d.png">

### Steps to Reproduce

Following the step between 2 & 3 to add the stuff folder causes problems when trying to open the resulting build. On open, there is a warning notification behind the startup screen that says the menubar settings template is missing (the warning is barely visible & has to be dragged into view). At that point OT needs to force quit. Using the stuff folder from a normal OT install allows the build to open. However the mypaint brush icons are missing (except for the no brush icon) and changing to the vector tab in the style editor always causes an immediate crash. 
### Expected Results

For OT to open if the stuff directory is created, for the icons to all be visible, and for no crashing with the vector tab. 
### Actual Results

If there is no OpenToonz folder in the Applications directory, copying the stuff folder fails. If there is an existing stuff folder in Applications, then the terminal command places the stuff folder in the 1.2 stuff folder. Either way, due to the path issues, OT can’t open completely. 

 If there is an existing stuff folder and terminal command is changed for the source to be stuff/ then the contents of the stuff directory appear to copy correctly. However in that case, there is a silent crash on open.

When using an existing stuff folder, OT opens but continues to have path issues with mypaint brush icons & crashes when changing to the vector tab. The vector tab crash does occur periodically anyways, but this is every time. This crash seems to stem from CustomStyleManager::StyleLoaderTask::run()
### System Information
* **OpenToonz Version**:
* **Operating System**: 10.13.6
* **CPU**: Intel Core i7
* **Memory**: 16GB
* **Graphics Card**: Radeon Pro 460 & Intel HD Graphics 530
* **Graphics Tablet**:

artisteacher:
Now that I've updated the OS to 10.14, I'm still struggling to get a completely successful build on MacOS following the current build instructions. I updated everything that could be updated (including QT) and there are definitely still path issues. If I try to create the stuff directory when building, I get a crash before the loading window finishes:
[...]

PR #3031, opened 2020-01-16
Add support for GitHub Actions

otakuto:
GitHub Actions is easier than Travis and Appveyor for new developers.

### DIFF
--- a/ci-scripts/osx/travis-install.sh
+++ b/ci-scripts/osx/travis-install.sh
@@ -6,11 +6,3 @@ brew install clang-format
 # from Homebrew 1.6.0 the old formula for obtaining Qt5.9.2 becomes invalid.
 # so we start to use the latest version of Qt. (#1910)
 brew install qt
-# temp workaround to brew installed glew cmake info overriding glew lib detection
-# which causes compiling issues later
-if [ -L /usr/local/lib/cmake/glew ]
-then
-   echo "Symbolic link '/usr/local/lib/cmake/glew' detected. Removing to avoid glew lib detection issue!"
-   ls -l /usr/local/lib/cmake/glew
-   rm /usr/local/lib/cmake/glew
-fi
--- a/.github/workflows/CMakeLists.txt
+++ b/.github/workflows/CMakeLists.txt
@@ -1,0 +1,38 @@
+cmake_minimum_required(VERSION 2.8.12)
+
+project(mypaint LANGUAGES C)
+
+find_package(json-c CONFIG REQUIRED)
+
+include_directories(${JSON-C_INCLUDE_DIR})
+
+set(sources
+  brushmodes.c
+  fifo.c
+  helpers.c
+  mypaint-brush-settings.c
+  mypaint-brush.c
+  mypaint-fixed-tiled-surface.c
+  mypaint-mapping.c
+  mypaint-matrix.c
+  mypaint-rectangle.c
+  mypaint-surface.c
+  mypaint-symmetry.c
+  mypaint-tiled-surface.c
+  mypaint.c
+  operationqueue.c
+  rng-double.c
+  tilemap.c
+  utils.c
+)
+set(headers
+  mypaint-config.h
+  mypaint-glib-compat.h
+  mypaint-mapping.h
+  mypaint-matrix.h
+  mypaint-symmetry.h
+  config.h
+  mypaint.h
+)
+
+add_library(libmypaint STATIC ${sources} ${headers})
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -52,8 +52,8 @@ jobs:
         cp appdir/usr/share/applications/*.desktop appdir
         cp appdir/usr/share/icons/hicolor/*/apps/*.png appdir
 
-        mkdir -p artifact/OpenToonzPortable
-        mv appdir/usr/share/opentoonz/stuff artifact/OpenToonzPortable/portablestuff
+        mkdir artifact
+        mv appdir/usr/share/opentoonz/stuff artifact/portablestuff
         rmdir appdir/usr/share/opentoonz
 
         wget -q -c 'https://github.com/linuxdeploy/linuxdeploy/releases/download/continuous/linuxdeploy-x86_64.AppImage'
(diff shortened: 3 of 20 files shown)


otakuto:
You can see logs and artifacts here.
https://github.com/otakuto/opentoonz/commit/c4f6911808f492b28e3236d68e9ae04a421cbd20/checks

shun-iwasawa:
Looks very nice. I'll see in detail after releasing the next version.

manongjohn:
Nice. 

Would have loved to use this for OTX, but it uses the latest versions of macOS (Catalina) and Ubuntu (18.04) so wouldn't be able to make portables compatible for older systems like I can with Travis currently.

Could still configure macOS builds to create a portable as a dmg file for testing purposes (nightlies too?), similar to OTX, though it would only work on Catalina.  You can see how I build macOS portables here: https://github.com/manongjohn/OTX/blob/master/ci-scripts/osx/travis-buildpkg.sh

I build Linux portable AppImage with a similar script using an opensource tool: https://github.com/manongjohn/OTX/blob/master/ci-scripts/linux/travis-buildpkg.sh

otakuto:
Thank you for your reviews.
OK, I will try to create portable artifacts.
As far as I can see, GitHub Actions supports only the latest macOS for now.
https://help.github.com/en/actions/automating-your-workflow-with-github-actions/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources
Would it be better to add Ubuntu-16.04?

manongjohn:
i would suggest 16.04.  I believe the opensource tool "linuxdeployqt" that I use to build the AppImage actually gives a warning about using it on the most recent version as the AppImage could have compatibility issues with older versions.  This is why i still build on 14.04.

otakuto:
Could you check to these artifacts work well.
I don't have a Mac computer.
https://github.com/otakuto/opentoonz/actions/runs/33007185

artisteacher:
I can test it when I get home today. I don't currently have Catalina at school.

artisteacher:
Like the OTX release, trying to open the first time fails. Afterwards, the app opens without issue. I did run into a couple problems that I've had when building myself #2292. The icons for the mypaint brushes are missing, though they are functional. That has to be some sort of path issue. Also, "Saving All" sometimes causes a crash.

Despite those issues, it's great to be able to test without needing to build manually. Such a time saver!

otakuto:
@artisteacher
Thanks.
Using cp -rp for now.

@manongjohn
linuxdeploy worked well for build on Ubuntu 18.04.
Cloud you check for artifacts.
https://github.com/otakuto/opentoonz/actions/runs/34461108

manongjohn:
@otakuto I was able to run the linux artifact (gcc) on my 18.04 VM.  Can you make sure, before it zips up the file that the AppImage on either Linux builds that it has Execute permissions. I recommend "chmod 755" on it after it is done creating it.  I had to manually change it before I could run it.

otakuto:
@manongjohn 
Comments fixed.
https://github.com/otakuto/opentoonz/actions/runs/40496962

MrBlockCat:
This sounds extremely helpful, thanks for your work on this! It seems like we may also need updated documentation on how to build locally with this once this PR is merged?

artisteacher:
I don't think this impacts building locally. Once this is merged we'll be able to test PRs without needing to build ourselves. That said, building locally on mac isn't perfect with the current directions.

MrBlockCat:
So we'll keep the options and as such documentation to build locally, and have this as an alternative? Interesting! And yeah, the directions aren't perfect as of now, though they're closer than a lot of build directions I've seen elsewhere...

MrBlockCat:
I just thought about this, but would nightly/production releases be built using this?

manongjohn:
No. This would just be for testing PRs.

Nightlies and official releases are created on build servers which only Shun-iwasawa has access to.

MrBlockCat:
Ah, thank you for the clarification!

otakuto:
@shun-iwasawa @manongjohn 
Is there anything else I should do?

ci-scripts/osx/travis-install.sh
```diff
@@ -6,11 +6,3 @@ brew install clang-format
 # from Homebrew 1.6.0 the old formula for obtaining Qt5.9.2 becomes invalid.
 # so we start to use the latest version of Qt. (#1910)
 brew install qt
-# temp workaround to brew installed glew cmake info overriding glew lib detection
-# which causes compiling issues later
-if [ -L /usr/local/lib/cmake/glew ]
```
manongjohn, line 11: Is there a reason you are removing this? To my knowledge, this is still needed for OSX Travis builds to build successfully.

manongjohn:
I removed the comment I made about the glew change. I finally saw the fix you did.  

So merging this will automatically enable this to work or is there something additional that needs to be done from a repo configuration that needs to be done?

otakuto:
Pobably, Repo configuration don't needs to be done.

artisteacher:
I've figured out the mypaint icon/saving issue when building on macOS was due to incorrectly building with an old version of libpng (from mono frameworks) but running on a newer version of libpng. Maybe something similar is happening here?

MrBlockCat:
If it helps, here's the PR @artisteacher made to fix the issue they described, but for existing build methods: #3215, line 229

manongjohn:
Just something to consider before implementing this, ...

Travis and AppVeyor cause GIT LFS Storage/Bandwidth usage.  I don't know how OT manages this, but I know for personal accounts there is a 1 GB limit on how much LFS Storage or Bandwidth usage is allowed per month before you have to purchase more.  Storage doesn't appear to be an issue, I think, but Bandwidth will be with nightly pulls, PR builds and people updating forks (which count against the org).

As it is, every OTX release uses 50% of LFS Bandwidith when generating 1 windows, 1 macOS and 2 Linux builds.  I can't imagine how much OT uses up or how much it's costing them per month if Dwango is blanket purchasing additional units per month.

Github Actions use Build Storage/Minutes instead.  The minutes burned for the same build depends on OS it seems (linux=1x, macOS=10x, windows=2x).  Travis and Appveyor builds take around 30-35 minutes.  That's 490 minutes burned for just 1 PR.  There is also storage costs to consider.

I would imagine from a Dwango point of view, they'd want to do some cost analysis before using or even switching to this.  

At any rate, this should not be merged without Dwango's support.

RodneyBaker:
I'm for tabling this (i.e. placing on hiatus) until more is known and a proper decision can be made.
Perhaps individual forks/branches could turn Github Actions on to resolve this for the needs of  the individual developer.  Then each developer could tailor the Github Actions to their specific requirements.

RodneyBaker:
Information about migrating to Github Actions (from say... Travis):

https://docs.github.com/en/actions/migrating-to-github-actions/migrating-from-jenkins-to-github-actions

Note:  I am not suggesting we do this just pointing to the information for those interested.

RodneyBaker:
Closing pending further discussion and interest especially in light of the bandwidth issue associated with Github Actions.

manongjohn:
Not sure if I overlooked this or if it somehow changed later on, but reviewing Github Actions recently for OTX, I found that build minutes quota apparently only applies to Private repos.  As OT is a Public repo, I think it can use it without any additional cost or limitations.

manongjohn:
FYI, It seems Travis-ci.org is shutting down in Dec, requiring everyone to migrate to Travis-ci.com. Not sure if it will be automatic.  However, there is a bigger issue in that Travis will be pretty much removing free building of open source projects.  You now need to start purchasing build credits.  This will certainly impact PR build tests.  

Existing users are converted to a Free plan with 10K credits to start, but then must somehow purchase more or switch to another plan.  Cost of builds varies, with macOS builds costing a lot more credits than Linux builds.

In short, you might want to revive this PR if you want to continue testing Linux/macOS builds automatically.  There is a storage cost for keeping artifacts generated by Github Actions..  May or may not want to have those saved. I am not 100% sure, but I think it still uses some Git LFS Bandwidth when building, but that is probably ok in this case since you'd be replacing Travis usage with this one.

PR #3046, opened 2020-01-21
Fix Window Activating Behavior

### LINKED ISSUE #3045: The Camera Capture window moves to the background and does not return when the cursor moves to some panels.

ezgif-3-694924393b32

### System Information

shun-iwasawa:
This will fix #3045 .
Currently the Palette and the Studio Palette panels will be activated (i.e. move on top and set keboard focus) when the mouse cursor enters in order to enable quick operations.
This PR adds a condition for such behavior so that the panels will not be activated when the current top-level window is a dialog like the Camera Capture popup.


RodneyBaker:
Thanks!

PR #3052, opened 2020-01-22
Keeping Separate Colors Parameters

shun-iwasawa:
**Please do not merge before releasing v1.4 as this is feature-PR.**

This will improve `Separate Colors` feature as follows so that users can save processing parameters and reproduce the same result afterwards.

- After executing the processing, parameters will be saved in a settings file ( `.sep` file ) with the same name and the location as the level. ( i.e. after processing the level `C:\foo\bar\A..tif` , the settings will be saved at `C:\foo\bar\A.sep` )
- When opening the `Separate Colors` popup, if the settings file for the target level exists, it will be loaded.
- The setting file can also be imported from / exported to any location by using newly-added `Save Settings` and `Load Settings` buttons.

This PR also improves the `Save In` field. When user manually typing non-existent path, a new folder will be created after confirmation.

### DIFF
--- a/toonz/sources/toonz/separatecolorspopup.cpp
+++ b/toonz/sources/toonz/separatecolorspopup.cpp
@@ -8,6 +8,7 @@
 #include "tiio.h"
 #include "toutputproperties.h"
 #include "filebrowser.h"
+#include "filebrowserpopup.h"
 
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Very nice addition.
LGTM

Issue #2729, opened 2019-08-24
Cleanup - Line Processing - None. Bring it back?


gab3d:
@shun-iwasawa 
@manongjohn 

Could it be possible to bring back the **None** option for the Line Processing mode in the Cleanup Settings panel?
It's useful for just stabilizing drawings without touching anything else in the image.

Of course you can stabilize them while you clean them up, but that does not cover all use cases, for example the following one:

- I've got some scanned drawings of a character, and I want to be able to overlay, to a certain extent, the original drawings over the cleaned up ones just to give the final image a more natural and rough look.
The problem is that without the **None** option, I can only stabilize the cleaned up level, but the original scanned one is not stabilized, so to be able to do it I have to resource to other software like After Effects, for example.

If OT could allow again to stabilize without further processing of the image, this could also be resolved inside OT, which of course would be better.

Thanks!!

RodneyBaker:
It's actually painful to close some of these feature requests as most would be so very useful...

Based on the discussion in #1207 feature requests are going to be closed here on GitHub unless there is a developer actively working on the feature. Feature requests can be discussed at the OpenToonz Google Group: https://groups.google.com/forum/#!forum/opentoonz_en

PR #3053, opened 2020-01-23
Revert Delete Unused Styles right-click command

### LINKED ISSUE #3035: Delete unused styles right-click item missing?
### Issue Summary
I couldn't find the Delete unused styles command in the right-click menu, I had to set it to a shortcut.

### Steps to Reproduce
Right-click in the palette, there's no option (for me at least).

### Expected Results
For it to be a right-click menu item in the palette.

### Additional Context
I could almost swear it used to be there, so it could be user error.

shun-iwasawa:
This PR resolves #3035 .

Note that this command is relatively time consuming especailly when you have many styles in the palette or many frames in the level, as it checks all frames the usage of each style.

The command was originally in RMB menu in the main pane (where styles are placed) of the palette. 
I decided to move the command to RMB menu of the upper area ( i.e. tabs & tool bar area ) considering that;  1) the command modifies whole palette rather than each styles and  2) it may be annoying for users if they accidentally click the command by mistake.
(Now I remember that the latter reason is why I asked to remove the command from the RMB menu when I was developping the Ghibli edition.)

Also I modified / fixed the following:
- Make this command to open a confirmation popup before actually deleting styles.
- Fixed the logic to check the style usage for vector levels.
- Prevent the command to appear on `Raster Drawing Palette` , locked palettes, or inactive palettes.

In association with the palette UI, I also modified the view options menu (which pops from the humberger button) to have radio buttons in order to let users to know the options are exclusive and to see which option is selected.

### DIFF
--- a/toonz/sources/include/toonz/fullcolorpalette.h
+++ b/toonz/sources/include/toonz/fullcolorpalette.h
@@ -35,6 +35,7 @@ public:
   void clear();
   TPalette *getPalette(ToonzScene *scene);
   void savePalette(ToonzScene *scene);
+  bool isFullColorPalette(TPalette *palette) { return m_palette == palette; }
 };
 
 #endif  // FULLCOLOR_PALETTE
--- a/toonz/sources/include/toonzqt/paletteviewer.h
+++ b/toonz/sources/include/toonzqt/paletteviewer.h
@@ -82,6 +82,7 @@ protected:
   TPaletteHandle *m_paletteHandle;
   TFrameHandle *m_frameHandle;
   TXsheetHandle *m_xsheetHandle;
+  TXshLevelHandle *m_levelHandle;
 
   QScrollArea *m_pageViewerScrollArea;
   PaletteViewerGUI::PageViewer *m_pageViewer;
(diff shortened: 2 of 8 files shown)


gab3d:
@shun-iwasawa 
Thanks for reverting this one. It will be very handy to clean-up palettes.! 

...in the heels of bringing back some simple features 😉 😉 
if you could check #2729 it'd be just great! I know I'd be using that one a lot (cos now I have to go to After Effects just for stabilizing drawings before cleanup)

shun-iwasawa:
@gab3d OK I'll look into it.

gab3d:
Thank you! :)

hayoko:
Awesome stuff, thanks!

RodneyBaker:
Thanks!

RodneyBaker:
I slight postmortem thought...
It seems a little strange that we can delete the black #1 style using this.
This actually might be ideal in some ways but it almost seems that the black #1 style should be similar to the transparent #0 style in how it persists.

This isn't a suggestion to change it... just an observation.

PR #3054, opened 2020-01-23
Fix Load Folder crashing while loading tga files

### LINKED ISSUE #3051: Unable to load/import folder with .tga files inside
### Issue Summary
OpenToonz is unable to load/import a folder with _.tga_ files

tga_bug

### Steps to Reproduce
I use sample data provided from OpenToonz official website
If you load a folder by drag & drop operation or Load Folder command in _File_ menu with any_ _.tga_ files in it, OpenToonz will refuses to load/crashed

### Expected Results
OpenToonz should load all assets into the xsheets columns/timeline rows

### Actual Results
Drag & Drop operation: 
OpenToonz refuses to load assets when .tga files are inside the folder

Load Folder command:
OpenToonz crashed 

### System Information
* **OpenToonz Version**: OpenToonz 1.4rc
* **Operating System**: Windows 10 Build 19035
* **CPU**: AMD Ryzen™ 3 2200G
* **Memory**: 8 GB
* **Graphics Card**: (APU) Radeon™ Vega 8 Graphics

manongjohn:
This PR fixes #3051

The current logic in `Load Folder` appears to want to internally convert TGA files named as `Axxxx.tga` to OT recognized sequenced file naming convention `A.xxxx.tga`.  

When `Load Folder` encounters the naming convention `A_xxxx.tga` (an accepted OT sequenced file name), it still does not think it's a properly named sequenced file and wants to put a `.` in the right place to changed it.  However, when extracting the name from the full path, it returned the name as `A` since it saw it as a sequenced file name. Due to the shortened name, OT crashed when it attempted to insert the `.` which is now outside the string's bound.

Modified logic to determine if the file name is already an accepted OT sequenced file name and skip renaming to prevent the crash and allow the level to load correctly.

NOTE: TGA files named as `Axxxx.tga` will not normally load as a sequenced level through Load Level. Instead these files are loaded as individual levels.

Due to the renaming logic in `Load Folder`, `Axxxx.tga` files will fail to load completely since it converts it to `A..tga` but cannot find files with that naming convention. This issue should be handled separately once it is decided if `Axxxx.tga` files should be loaded as sequenced files or not.

### DIFF
--- a/toonz/sources/toonz/loadfoldercommand.cpp
+++ b/toonz/sources/toonz/loadfoldercommand.cpp
@@ -182,9 +182,11 @@ TFilePath multiframeResourcePath(const TFilePath &fp) {
 
 TFilePath retasComponentPath(const TFilePath &fp) {
   std::wstring name = fp.getWideName();
-  assert(name.size() > 4);
 
-  name.insert(name.size() - 4, 1, L'.');
+  // Assumes the name is Axxxx.tga and needs to be changed to A.xxxx.tga
+  if (name.size() > 4 && fp.getDots() != "..")
+    name.insert(name.size() - 4, 1, L'.');
+
   return fp.withName(name);
 }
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thanks!

PR #3058, opened 2020-01-24
Revert "None" line pocessing option to the cleanup settings

### LINKED ISSUE #2729: Cleanup - Line Processing - None. Bring it back?

shun-iwasawa:
This resolves #2729 
Also fixed the cleanup settings to properly load the default settings on launching the software.

### DIFF
--- a/toonz/sources/toonz/cleanupsettingspane.cpp
+++ b/toonz/sources/toonz/cleanupsettingspane.cpp
@@ -108,13 +108,14 @@ CleanupSettingsPane::CleanupSettingsPane(QWidget *parent)
   QPushButton *loadBtn  = new QPushButton(tr("Load"));
   QPushButton *resetBtn = new QPushButton(tr("Reset"));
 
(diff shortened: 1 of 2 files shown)


gab3d:
Thank you so much for this one @shun-iwasawa , you're the best! 😸

PR #3064, opened 2020-01-27
Modify Creating Scene from Startup Popup

### LINKED ISSUE #3034: Creating a scene from the Startup popup doesn't save the scene
### Issue Summary
I don't understand why this basic functionality hasn't been implemented. From the popup dialog, hitting create project creates all of the folders necessary for a new project, including saving the project file. Why does hitting create scene not save the scene file? Clearly there's something I'm missing, but it confuses many users. You have to create a drawing level and hit save all to save the scene (and update the application header bar). Consistency would give users more comfort in using the application.

In the attached clip you can see that after creating the project, the project name is shown in the application header, but on creating the scene, the scene name shows as Untitled41.

There was some work already completed, but left commented out (I'm guessing because it needs more testing) in this PR: https://github.com/opentoonz/opentoonz/pull/2875

ScreenCapture

Here's the location for that in code: https://github.com/opentoonz/opentoonz/blob/98a68d5c9b720890d06406ba92f5643250cf6052/toonz/sources/toonz/startuppopup.cpp#L507

### Expected Results
On clicking Create Scene, I'd expect the scene name to show in the application header.
And if we choose to do so, for the scene file to be created on disk.
However, considering the conversation in #3036, we might not want the scene file saved, but the header should definitely update as feedback to the user.

### System Information
* **OpenToonz Version**: V1.3 Built Oct 5 2019 10:17:51

shun-iwasawa:
This PR resolves #3034 .

- Made the scene to be saved on the disk right away when the `Create Scene` button is clicked. Unlike the levels, users still can edit the untitled scene without being saved just by closing the popup.
- Fix the title bar to be updated when creating the scene.

### DIFF
--- a/toonz/sources/toonz/startuppopup.cpp
+++ b/toonz/sources/toonz/startuppopup.cpp
@@ -504,10 +504,11 @@ void StartupPopup::onCreateButton() {
       size);
   TApp::instance()->getCurrentScene()->getScene()->getCurrentCamera()->setRes(
       res);
-  // this one is debatable - should the scene be saved right away?
-  // IoCmd::saveScene();
+  // save the scene right away
+  IoCmd::saveScene();
   // this makes sure the scene viewers update to the right fps
   TApp::instance()->getCurrentScene()->notifySceneSwitched();
+  TApp::instance()->getCurrentScene()->notifyNameSceneChange();
 
   hide();
 }


RodneyBaker:
Thanks.  I've been looking forward to seeing this one.  :)

DarrenTAnims:
Me too. That's an excellent fix that'll help us all. Great job

PR #3065, opened 2020-01-27
Fix Crash on Exporting XDTS When There are No Columns to be Exported

shun-iwasawa:
This PR fixes the `Export Exchange Digital Time Sheet (XDTS)` command as follows:

- Fixed crash when there are no columns to be exported (e.g. empty scene).
- Enabled to select export destination.


RodneyBaker:
Nice catch.

PR #3067, opened 2020-01-27
Reset Undo when Saving Levels Preference Option

shun-iwasawa:
**It's devatable if it should be merged into v1.4 - as there is a known issue (see below).**

Currently, the undo history is cleared when user uses `Save Level` , `Save Level As` or `Overwrite Palette` command. 
This behavior was originally demanded by Ghibli ink&paint staffs.
Since undo can change non-working frame without realizing it, they would like to use the `Save Level` command when they finish each frame and "pack up" their work by clearing undo.

This PR will make this behavior optional. The option will be in `Preferences > Saving` category and set to ON ( = do clear undo when saving levels ) by default.

**Known Issue :** I confirmed that some level-editing operations do not set dirty flag to the level on undo.  I think it is difficult to fix all of them to set dirty flag properly before releasing v1.4 . 
I'm leaning towards keeping this PR open (unmerged) until after releasing v1.4, since there seems to be no incovinience reported with the current behavior so far.

### DIFF
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -65,6 +65,7 @@ enum PreferencesItemId {
   //----------
   // Saving
   rasterBackgroundColor,
+  resetUndoOnSavingLevel,
 
   //----------
   // Import / Export
--- a/toonz/sources/toonz/filebrowserpopup.cpp
+++ b/toonz/sources/toonz/filebrowserpopup.cpp
@@ -1651,7 +1651,8 @@ bool SaveLevelAsPopup::execute() {
     DvDirModel::instance()->refreshFolder(fp.getParentDir());
 
     // reset undo memory!!
-    TUndoManager::manager()->reset();
+    if (Preferences::instance()->getBoolValue(resetUndoOnSavingLevel))
+      TUndoManager::manager()->reset();
   }
 
   if (ret)
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -2788,8 +2788,8 @@ public:
     }
 
     // reset the undo before save level
-    // TODO: この仕様、Preferencesでオプション化する
-    TUndoManager::manager()->reset();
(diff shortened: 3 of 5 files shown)


gab3d:
I'm fine with giving this the proper time to be able do it right. 👍

RodneyBaker:
I'm all for having the option in Preferences.
I'd prefer the option to reset Undo to be off by default rather than on as it's generally better to maintain historical data rather than release it although... read on...  ;)

As for timing of merger I'm good either way but it will be good to consider the broader case of workflow as desired by Studio Ghibli (and presumably others) where clearing/resetting Undo history is preferred.  In such a case it would be ideal if as with clearing the cache (via 'clear cache folder') reset of Undo History could be triggered via command/shortcut key.

Regarding Undo History itself this is a broader area of interest where it would be optimal if history could be saved off line (via file) and accessed/ as needed.  Resetting of Undo history then becomes less of a backup plan to recover from error and more of a workflow enhancement feature.

In light of the bigger picture I'd say this reset of Undo history at Level Saving is more of a fix/revert than a new feature and thus can be added to v1.4 with a goal to consider related features and enhancements related to Undo history moving forward toward v1.5.

RodneyBaker:
I can see the benefit in clearing undo history.
And I especially like that we can turn the option on/off.
Thanks much for giving us this option.

I've tested this PR and LGTM but I will wait to allow others time to test also.
As I like a as full a history as possible via Undo/Redo I have a feeling I will usually turn this option off.

manongjohn:
Can you please make this option Off by default? Not everyone's workflow will be like Studio Ghibli and may Save Level while working a Scene and suddenly find they cannot Undo stuff.

manongjohn:
I would like to see this turned Off by default also.

gab3d:
I guess I'd prefer it to be off by default too.

RodneyBaker:
Merging this as it gives us the same usage as current releases but adds the ability to save history that we could not before.  

Adjusting changes to Preferences and defaults can be picked up in the (pending) Preferences scrub.

PR #3071, opened 2020-01-28
Fix CurveFx Crash

### LINKED ISSUE #3069: Curves FX crash
gif example:
Curves Crash

Repro steps:
1. Create a Curves node
2. Drag the top right point to the bottom left corner or the bottom left point to the top right corner

OT will crash

OT Ver: 1.4 [Jan 24 Nightly]

shun-iwasawa:
This PR fixes #3069 

Now I modified the curve panel so that the first and the last control points will not be at the same position. If they are at the same horizontal positions (i.e. forming step-like curve to get a binarized image), the points cannot be closer than 16pixels vertically. 

Related to the Curve Fx UI I also modified the following in this PR:
- Fixed a problem that the curve handles are sometimes not visible.
- Made text color of the channel bar to be style-sheet-aware.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -1630,6 +1630,9 @@ FxSettings QToolBar QToolBar {
 #FxSettingsHelpButton:hover {
   background-color: #a8bee7;
 }
+ChannelBar {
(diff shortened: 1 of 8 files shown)


RodneyBaker:
Thank you @shun-iwasawa

PR #3074, opened 2020-01-29
Fix Vector Tape Tool on Joining Points at the Same Position

### LINKED ISSUE #3072: Join vector using tape tool may create buggy stroke
### Issue Summary
When using tape tool to join two strokes together, if those points are at the exact same position, it might create a buggy stroke which contains one too few points at one end. When trying to manipulate the missing point, opentoonz will crash.

https://www.youtube.com/watch?v=snDS_qMRy-Q

### Steps to Reproduce
1. create a toonz vector level
2. draw a stroke using geometric tool with Arc shape and snap option checked
3. draw another stroke but with one end snapped to one end point of the first stroke
4. use tape tool with normal type, endpoint to endpoint mode, join option selected, and smooth NOT selected to connect the connecting end points (the two you snapped together in step 3)
5. use control point edit tool to manipulate the missing point

### Expected Results
creates a not buggy stroke. If possible, the resulting stroke should have only 1 point in the place where it connects.

### Actual Results
the result stroke will contain one too few points and when you try to manipulate it, opentoonz will crash.

### other info
If you try to move the connecting point first, the stroke will be fixed. (video: https://youtu.be/hcRo29Iyg4Y)
If the smooth option is checked in tape tool, this won't happen
If you join two lines, this won't happen.
If you draw line first then an arc, this will happen. If you draw arc first then a line, this won't happen.

log when the crash happens: opentoonz/toonz/sources/tnztools/controlpointselection.cpp:528: bool ControlPointEditorStroke::isSpeedInLinear(int) const: Assertion `index < (int)m_controlPoints.size()' failed

### System Information
* **OpenToonz Version**: latest nightly (1af1ae59bc6275dc28650f25d6a98bc66fe3b848)
* **Operating System**: Arch linux 5.4.11-arch1-1 with i3 as desktop environment
* **Graphics Tablet**: Mouse

shun-iwasawa:
This fixes #3072 
Now the joining behavior of the vector tape tool w/o Smooth option changed as follows:
If the connecting points are at the exact same position, the tool will unite them into one control point instead of inserting zero-length segment between them.

Please note that if the connecting points have different thickness, the united point after joining will be with an average thickness of them.

### DIFF
--- a/toonz/sources/common/tvectorimage/tvectorimage.cpp
+++ b/toonz/sources/common/tvectorimage/tvectorimage.cpp
@@ -574,13 +574,13 @@ TRectD TVectorImage::getBBox() const {
   for (UINT i = 0; i < strokeCount; ++i) {
     TRectD r           = m_imp->m_strokes[i]->m_s->getBBox();
     TColorStyle *style = 0;
-    if (plt) style     = plt->getStyle(m_imp->m_strokes[i]->m_s->getStyle());
+    if (plt) style = plt->getStyle(m_imp->m_strokes[i]->m_s->getStyle());
     if (dynamic_cast<TRasterImagePatternStrokeStyle *>(style) ||
         dynamic_cast<TVectorImagePatternStrokeStyle *>(
             style))  // con i pattern style, il render a volte taglia sulla bbox
                      // dello stroke....
       // aumento la bbox della meta' delle sue dimensioni:pezzaccia.
-      r  = r.enlarge(std::max(r.getLx() * 0.25, r.getLy() * 0.25));
+      r = r.enlarge(std::max(r.getLx() * 0.25, r.getLy() * 0.25));
     bbox = ((i == 0) ? r : bbox + r);
   }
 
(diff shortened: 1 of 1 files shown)


gab3d:
awesome, thanks!!

RodneyBaker:
I think I'm seeing a problem when testing the Appveyor release for this PR.

The problem:  joining two closely spaced control points leads to control point moving to canvas center.

Steps to reproduce:
(Setup) in order to make sure the start and end control points of a line are as close in proximity as possible we will use the Geometry Tool to create the shape/line and the Cut Tool to sever the line to make two control points to join.

- Draw a circle with the Geometry Tool
- Use Cut Tool to sever line  (I cut it at 12 o'clock position for this test)
- Using Tape Tool with 'Join' option checked connect/join the previously severed location
Note the results

- Continue cutting and rejoining parts of the circle as desired.

Tape Tool Results

The new point does not always go to canvas center but almost always moves toward that location.

shun-iwasawa:
Thanks, I'll look into it.

shun-iwasawa:
@RodneyBaker I think I fixed the problem. 
Please note that the Cutter tool seems to have another problem ( #3075 ) so it'd better to test this PR with strokes drawn with Brush tool, snapping one end to the other. Thanks!

RodneyBaker:
Thanks @shun-iwasawa  I've been following your efforts to resolve this with great interest.

> it'd better to test this PR with strokes drawn with Brush tool, snapping one end to the other. 

Will do.  Thanks!

RodneyBaker:
The primary issue I was seeing does appear to be the one related to the Cut Tool.
Using the Tape Tool (via this PR) with brush strokes is working quite well.

It looks like a Jenkins call may be needed for this PR.

shun-iwasawa:
Jenkins again

RodneyBaker:
That worked.  Thanks!

PR #3076, opened 2020-01-30
Fix Duration Items in Clapperboard to Show Output Range

shun-iwasawa:
This PR fixes the duration items in clapperboard.
If the output range is specified in the output settings, such frame duration will be displayed instead of the xsheet length.


RodneyBaker:
LGTM

PR #3079, opened 2020-01-31
Add snap option to control point editor tool

pojienie:
This PR adds snap option for Control Point Editor, just like Geometric tool. It allows you to snap a control point onto a point of an existing line.

https://www.youtube.com/watch?v=f_0-59Wzk1A

A few known problems with this PR:

1. The functionality is very similar to that of tape tool. So this feature may not be useful for most people which means it's a dead weight to the codebase.
2. New code is mostly copied from geometric tool so there might be a code duplication issue.
3. Whether you check snap option or not, the snap sensitivity menu will be shown unlike geometric tool where the menu will be hidden if snap option is not checked.
4. In drawSnap, there is m_tool in geometric tool which provides getPixelSize function but control point editor tool doesn't have it so I'm not sure what negative effect this might bring.
5. behavior when dragging control point is different. Now, if you drag a control point at its corner, your mouse will stay at its control point's corner. With this PR, once your mouse drags, your mouse will be dragging at its center. Again, not sure what negative effect this might bring.
6. it can not snap to itself.

### DIFF
--- a/toonz/sources/tnztools/controlpointeditortool.cpp
+++ b/toonz/sources/tnztools/controlpointeditortool.cpp
@@ -770,13 +770,16 @@ void ControlPointEditorTool::leftButtonDrag(const TPointD &pos,
                                                 // deselezionata l'ultima
                                                 // selezione nel movimento
 
-    TThickPoint cp =
-        m_controlPointEditorStroke.getControlPoint(m_lastPointSelected);
-    TPointD controlPoint = TPointD(cp.x, cp.y);
+    if (m_lastPointSelected >= 0) {
+      TThickPoint cp;
+      TPointD controlPoint;
+      TPointD newPos;
 
-    TPointD newPos;
-    newPos = calculateSnap(pos);
-    delta  = newPos - m_pos + (m_pos - controlPoint);
+      cp = m_controlPointEditorStroke.getControlPoint(m_lastPointSelected);
+      controlPoint = TPointD(cp.x, cp.y);
+      newPos       = calculateSnap(pos);
+      delta        = newPos - m_pos + (m_pos - controlPoint);
+    }
 
     m_pos = pos;
 
--- a/toonz/sources/tnztools/controlpointeditortool.cpp
+++ b/toonz/sources/tnztools/controlpointeditortool.cpp
@@ -231,7 +231,7 @@ TPointD ControlPointEditorTool::calculateSnap(TPointD pos) {
   TVectorImageP vi(TTool::getImage(false));
   TPointD snapPoint = pos;
   if (vi && m_snap.getValue()) {
-    double minDistance     = m_snapMinDistance;
+    double minDistance = m_snapMinDistance;
 
     int i, strokeNumber = vi->getStrokeCount();
 
@@ -241,7 +241,7 @@ TPointD ControlPointEditorTool::calculateSnap(TPointD pos) {
 
     for (i = 0; i < strokeNumber; i++) {
       stroke = vi->getStroke(i);
-      if(stroke != selfStroke){
+      if (stroke != selfStroke) {
         if (stroke->getNearestW(pos, outW, distance) &&
             distance < minDistance) {
           minDistance = distance;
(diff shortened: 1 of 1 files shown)


gab3d:
Thanks!! looks really useful 😄

beeheemooth:
Thanks, man! This feature is expected by many users.

RodneyBaker:
@pojienie  Welcome to Opentoonz development!
You are off to a great start.

As v1.4 is about to be released I believe merging of this PR (as a new feature) needs to wait until after the official release.
@shun-iwasawa Please correct me if I am reading that wrong.

As for this PR...

Looking VERY good.
I did initially make the mistake of trying to snap a line to itself... even when I had read your list of things the feature can't do.   Whoops.    I figured it out though.

This will be a very useful enhancement of the control point editor tool.

RodneyBaker:
I have to ask...

I don't suppose you could add a 'Join' option where if the snapping of control points is to the end of a spline/line then the ends would automatically be joined.  @shun-iwasawa just now added some code relating a little to that with the Tape Tool.  That PR will not be merged until after v1.4 goes final release also.

shun-iwasawa:
Thank you @pojienie for the contribution! As @RodneyBaker mentioned, this PR will be reviewed after v1.4, which will be released very soon.

pojienie:
Thank you guys for your kind words!

Yes, I understand this pull request will only be reviewed after the release of v1.4.

As for the join option, I'll first wait for this PR to be reviewed then create another PR for it if I'm capable of implementing it. Thanks @RodneyBaker for referencing the #3074 and suggesting the idea and @shun-iwasawa for making the #3074 PR, it'll be very helpful!

beeheemooth:
@pojienie The problems you described in practice are not at all significant. The Tape Tool has some limitations, for example, it cannot connect the ends of vectors at an acute angle, smoothing only the vector. With the help of your feature, this can be done. I played around with it and found one crash. Holding down the Ctrl button you can add new points, but when snap is on, this operation fails. (You need to hold Ctrl, click on the vector and pull). Otherwise, I did not notice other problems, everything works well. 
ezgif-3-2a2283081e50

pojienie:
@beeheemooth thank you for testing! I know where the problem is: In order to get the dragged control point, I need to use m_lastPointSelected, however, when a point is created, the value is set to -1 so the assert in getControlPoint failed.

There are two possible way to fix:
1. guard it so that snap won't happen when m_lastPointSelected is -1
2. change the behavior when creating point so that m_action is not set to CP_MOVEMENT but NONE.

To change the least amount of behavior, I've made a commit that implemented the first one. However, I do think the second one is more reasonable but also more risky.

RodneyBaker:
The more I test the more I like about this PR.
For example, when selecting multiple control points we can drag the entire selection along the path of the target spline.  And this based on whichever control point we use to move the selection.
Nice.

jrhyme:
The developer fixed the crash that was caused by clicking on the control button. You can think of adding to the release. Failures are nowhere to be seen. Shun. Check)

beeheemooth:
@pojienie Thanks for fixing the crash! Now everything works well. I hope that your feature will be implemented soon. Vector issues receive so little attention. Thank you for your work!

RodneyBaker:
LGTM

pojienie:
@RodneyBaker Thank you for your kind words and for merging!

AlmostMikhail:
Hi, can you create a snap that snaps halfway of another drawing as shown in Adobe illustrate in this video : https://youtu.be/kdQaw0O6j2A

It a bit more similar here because you don't have to select each control point to snap a drawing, it automatically snaps halfway or at the edges when you move the drawing. 

Thanks

williebluebussell:
hi guys where can i download a working brush like this... please hope you can answer back... and if its out... where is it...

RodneyBaker:
@williebluebussell Can you clarify more?  This topic is about the control point editor so I'm not sure what 'brush' you are referring to.

Whatever you are looking for you can probably find it somewhere in the latest nightly release.
Link:  https://github.com/opentoonz/opentoonz_nightlies/releases

Alternatively there are forks that have some additional brush tool related features not yet in Opentoonz such as Tahoma2D.
Link:  https://tahoma2d.org/download/

PR #3089, opened 2020-02-03
Use standard library instead of boost

otakuto:


RodneyBaker:
Thanks for the refactor.
LGTM

PR #3093, opened 2020-02-05
Camera Capture Using OpenCV

shun-iwasawa:
This PR will change the camera capture popup to use OpenCV instead of QCamera, referring to the stop motion feature introduced by @turtletooth .

I confirmed that the change notably improves the frame rate for some USB cameras, which seems to be thanks to `MJPG` codec which can be selected with OpenCV.

Since OpenCV is included only when `WITH_STOPMOTION` flag is ON, currently this change only appears when building the software with that flag. Without the `WITH_STOPMOTION` flag the camera capture feature remains using Qt's QCamera class.

In future development, if there is no problem I'm thinking of making OpenCV usage not to be limited to the build with  `WITH_STOPMOTION` flag..
That is to say, 1) all developers will be needed to install OpenCV, 2) `WITH_STOPMOTION` flag will be renamed to, say, `WITH_CANON_DSLR_SUPPORT` indicating usage of CANON sdk, and 3) the stop motion feature will become available in any build at least with web cameras.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -1530,6 +1530,9 @@ ProjectPopup QLabel {
 #GearButton {
   qproperty-icon: url('../Default/imgs/white/gear.svg');
 }
+#GearButton::menu-indicator {
+  image: "";
+}
 #SubfolderButton {
   qproperty-icon: url('../Default/imgs/white/subfolder.svg');
   padding-left: 6px;
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -1530,6 +1530,9 @@ ProjectPopup QLabel {
 #GearButton {
   qproperty-icon: url('../Default/imgs/white/gear.svg');
 }
+#GearButton::menu-indicator {
+  image: "";
+}
 #SubfolderButton {
   qproperty-icon: url('../Default/imgs/white/subfolder.svg');
   padding-left: 6px;
--- a/stuff/config/qss/Default/Default.qss
+++ b/stuff/config/qss/Default/Default.qss
@@ -1530,6 +1530,9 @@ ProjectPopup QLabel {
 #GearButton {
   qproperty-icon: url('imgs/white/gear.svg');
 }
+#GearButton::menu-indicator {
+  image: "";
+}
 #SubfolderButton {
   qproperty-icon: url('imgs/white/subfolder.svg');
   padding-left: 6px;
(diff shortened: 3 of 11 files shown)


shun-iwasawa:
FYI In my test environment using HOZAN L-835 USB Camera with 1920x1080 pixels size, the capturing time improved from 238.8msec (4fps) to 67.5msec (15fps).

manongjohn:
> In future development, if there is no problem I'm thinking of making OpenCV usage not to be limited to the build with `WITH_STOPMOTION` flag..
> That is to say, 1) all developers will be needed to install OpenCV, 2) `WITH_STOPMOTION` flag will be renamed to, say, `WITH_CANON_DSLR_SUPPORT` indicating usage of CANON sdk, and 3) the stop motion feature will become available in any build at least with web cameras.

Funny, I was thinking about this the other day.. I was going to ask if there was a way to compile to allow stop motion with just web cameras. Great minds think alike.😁. I think this would be a great idea.

Side note: if OpenCV is going to be a part of the build now, it would be great to update the Dwango Plugins that only work with OpenCV 3.1.0 to use the latest. It would certainly simplify installing them since the DLL will already be there.

artisteacher:
> Since OpenCV is included only when WITH_STOPMOTION flag is ON, currently this change only appears when building the software with that flag.

So this is currently only for Windows? Will it be a lot of work to get Stop Motion running on macOS too?

shun-iwasawa:
Jenkins again

shun-iwasawa:
Jenkins again

shun-iwasawa:
@artisteacher 
>So this is currently only for Windows? 

I'm sorry to say, yes.

>Will it be a lot of work to get Stop Motion running on macOS too?

I'm not sure but I'll work on it. Thank you for your patience!

RodneyBaker:
Camera capture is working great here.

This PR says it has a conflict with penciltestpopup.cpp but from what I can tell it is working fine.
Other than the conflict I'm not sure what is preventing this from moving forward.

turtletooth:
@manongjohn What are your thoughts on enabling OpenCV capture on all builds of OT?  I'm all for it, but want to hear your thoughts.

RodneyBaker:
Merging this so that other related fixes can also be merged (REF:  #3221)

quirinux-so:
Hi, I am trying to use my webcam in Debian GNU / Linux and I can in version 1.3 but not 1.4. Do you think there is something I can do when compiling? If so, please indicate the precise commands, since I am not an expert. I leave you more information here. 

https://github.com/opentoonz/opentoonz/issues/3333

Thank you!

PR #3094, opened 2020-02-05
Enable to Reload Read-only Level from Level Strip

shun-iwasawa:
This PR will enable the `Reload` command in RMB menu of the Level Strip, regardless of whether the level is read-only / subsequence or not.
Even if the level is read-only this command is still needed to update the level, since it can be modified by another software.

This PR will also enable to reload PSD file, just like the same manner as #2415 .


RodneyBaker:
LGTM

PR #3095, opened 2020-02-05
Fix XDTS Export Feature to Use Output Range for Duration

shun-iwasawa:
This PR modifies XDTS export feature as follows:

- If the current xsheet is the top xsheet in the scene and the output frame range is specified, set the "to" frame value as "duration", instead of the xsheet frame length.
- If the last cell in the frame range is occupied, put the Null Cell Symbol ("X") at the following cell.


RodneyBaker:
I don't quite follow everything that is going on with regard to Output Range but everything checks out here.  Perhaps somewhere down the line we'll see some walkthroughs and tutorials that conver use of XDTS in Opentoonz that will assist users in fully understanding.

Thanks Shun!

PR #3099, opened 2020-02-13
Fix Convert to TLV

shun-iwasawa:
This PR will fix `Convert...` command used with the `File Format` option being set to `tlv` .
- Fixed crash on converting if the mode is set to `Unpainted TLV` and the source level contains empty (= no drawing) frame.
- Fixed a bug that the `Start` and the `End` fields do not work. (It converted all frames in the source level regardless of the range specified by them.)
- Fixed a bug that the convertion do not work properly when it is used to more than one selected level.
- Made the progress bar to show the progress properly when converting with the `Unpainted TLV` mode.

### DIFF
--- a/toonz/sources/toonz/convertpopup.cpp
+++ b/toonz/sources/toonz/convertpopup.cpp
@@ -256,6 +256,8 @@ void ConvertPopup::Converter::convertLevelWithConvert2Tlv(
     tlvConverter->abort();
   } else {
     int count = tlvConverter->getFramesToConvertCount();
+    popup->m_progressDialog->setMinimum(0);
+    popup->m_progressDialog->setMaximum(count);
     bool stop = false;
     for (int j = 0; j < count && !stop; j++) {
       if (!tlvConverter->convertNext(errorMessage)) {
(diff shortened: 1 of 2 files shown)


RodneyBaker:
@shun-iwasawa 
I still get crashes with the appveyor build of this PR.
I've tried a few times and I get similar results.

I haven't yet gotten to tests of start and end frames etc. because I haven't been able to keep the conversion to TLV from crashing.  This with Unpainted option on and a variety of different image formats.

Note that I am using both the File menu's Convert option as well as the Right Click > Convert via the Browser window.

RodneyBaker:
In looking at this again, the crash seems to relate to the target image not having the correct naming convention,  For example, the same image with imagename(dot)0001(dot).extension converts fine (without crashing) while one named imagename(dot)extension crashes (or in some cases creates a broken TLV file.

RodneyBaker:
This is a very useful update given that it is ideal in many/most cases to convert from standard image fomrat to OT's native TLV.  As such, this PR could use some additional testing to move it forward.

Here the Windows appveyor build if anyone cares to test that:

https://ci.appveyor.com/api/buildjobs/n0o95kdnjkwtwhxm/artifacts/toonz%2Fx64%2FOpenToonz.zip

RodneyBaker:
Merging this as a fix to current and significant issues being experienced with TLV conversion.

PR #3105, opened 2020-02-19
Streamline Fx Settings

shun-iwasawa:
This PR will enhance refresh response of the Fx Settings on clicking a fx node in the Fx Schematic.

When switching the current fx, the function `buildSceneFx()` is called.
It can be time-consuming especially when working in large scene, since the function traces all upstream nodes. (In an extreme case, in a scene containing about 300 fx nodes it takes several seconds after clicking nodes to refresh the contents.)

After investigation, I found that calling `buildSceneFx()` seems to be needed to render the swatch preview ( = a small preview area in the bottom of the Fx Settings ).
So, I modified it to call `buildSceneFx()` only when the swatch preview is enabled.


RodneyBaker:
Very nice.
LGTM

PR #3106, opened 2020-02-19
Prevent Preview from Recomputation-checking when Switching Camstand Toggle

shun-iwasawa:
This PR will enhance response of clicking the camstand toggle in the Xsheet while preview flipbook is opened.

The signal `xsheetChanged()` will invoke "recomputation-checking" in `PreviewFxManager`, which traces all upstream fx nodes starting from the previewed node and check if any of fx/object paramaters has been changed or not. It can be time-consuming especially when working in large scene.

For now, clicking any part of the column header will emit `xsheetChanged()` signal, however toggling camstand view does not effect rendered result unless the `Enable Column Color Filter and Transparency for Rendering` option (in the Scene Settings) is enabled.
So, I made the `xsheetChanged()` signal to be emitted only after a manipulation which will affect the render result.


RodneyBaker:
LGTM
Thanks Shun!

PR #3107, opened 2020-02-19
Fix Slowness with Frozen Preview

shun-iwasawa:
This PR will enhance response when editing scene while the "frozen" preview flipbook is open.
(i.e. when the "pause" button in the top-right corner of the flipbook is enabled.)

For now, even if the preview is frozen, each chage to the scene will cause some computation in background such as updating render settings, updating camera and updating fx aliases, etc.
They can be time-consuming, especially when working on a large scene.

After investigation, I found that these computations seem to be unnecessary. So I made them to be done only once when unfreeze the preview flipbook.


RodneyBaker:
You are making me realize I don't use the Flipbook/Preview enough.
Thanks Shun.  LGTM

PR #3109, opened 2020-02-19
Fix Cache Fx Not Working

### LINKED ISSUE #3108: Cache Fx Not Working
### Issue Summary

Cache Fx feature in the Fx Schematic does not work under some condition. 

### Steps to Reproduce

1. Build the fx schematic like as follows:
    cache_schematic
    - Connect any time-consuming fx to the level. In the example image above, I used Fog Fx Ino with large radius value. Then set the fx to be cached.
    - Connect CheckerBoardFx to Xsheet Node.
1. Preview the scene. It will take some time before opening the preview result due to the time-consuming fx.
1. (To see the correct behavior) without closing the flipbook, double-click the CheckerBoardFx node and edit parameter of the fx via Fx Settings. You can promptly get the result.
1. Keeping the Fx setting showing  CheckerBoardFx parameters, click an Xsheet cell which contains level A.
1. Again, edit the parameter of CheckerBoardFx in the Fx Settings and see how long does it take to finish rendering.

### Expected Results

It should be rendered promptly.

### Actual Results

The time-consuming fx is computed as if the cache feature does not work.

### System Information
* **Operating System**: Win10

shun-iwasawa:
This PR fixes #3108 .

When there is a changes in fx, the cached fxs are checked if it should be recomputed or not.
The check is actually to see if the "current Fx" is contained in upstream nodes of the cached fx.

However, there is a case that the "current fx" is not the one displayed in the Fx Settings.
For now, when clicking in the xsheet cell, the current fx will be switched to the clicked column ( inside OT, columns are also handled as Fxs, but with no parameters ). In such case, the current fx will be switched but the fx settings popup remains to show the previously-selected fx paramters because it is more useful than updating (= showing nothing).
It causes misunderstanding by OT that the edited Fx is contained in the cached Fx tree.

To resolve the issue, I made the current fx not to be changed by clicking the xsheet cell.
Exceptionally, if the cells containing zerary Fx is clicked, current fx will be switched to such fx and the Fx Settings will be updated accordingly.

### DIFF
--- a/toonz/sources/toonz/xshcolumnviewer.cpp
+++ b/toonz/sources/toonz/xshcolumnviewer.cpp
@@ -2039,8 +2039,6 @@ void ColumnArea::mousePressEvent(QMouseEvent *event) {
           return;
         setDragTool(XsheetGUI::DragTool::makeColumnSelectionTool(m_viewer));
       }
-      // synchronize the current column and the current fx
-      if (column) TApp::instance()->getCurrentFx()->setFx(column->getFx());
     }
     // clicking on the normal columns
     else if (!isEmpty) {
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Thank you @shun-iwasawa !
LGTM

PR #3110, opened 2020-02-20
reverse snap option when holding control key

pojienie:
This pull request allows you to temporarily reverse snap option by holding control key in both geometric tool and control point editor tool. This is especially useful when drawing a line or any shapes that don't allow you to release left mouse button using geometric tool as you don't get a chance to reverse the snap option mid drawing.

Brush tool already has this functionality by holding alt key so I didn't touch it.

video demonstrations:
https://youtu.be/AY0vCMq0WvY
https://youtu.be/mWYV5zO5kAU

There is one problem, though... inconsistency. In brush tool, you can hold alt key to temporarily reverse snap. However, in this pull request, I used control key to do it because alt key have another usage in both geometric tool and control point editor tool... Any suggestion?

### DIFF
--- a/toonz/sources/tnztools/controlpointeditortool.cpp
+++ b/toonz/sources/tnztools/controlpointeditortool.cpp
@@ -217,7 +217,7 @@ public:
   // returns true if the pressed key is recognized and processed.
   bool isEventAcceptable(QEvent *e) override;
 
-  TPointD calculateSnap(TPointD pos);
+  TPointD calculateSnap(TPointD pos, bool ctrlPressed);
   void drawSnap();
   TPointD getSnap(TPointD pos);
   void resetSnap();
@@ -226,11 +226,11 @@ public:
 
 //-----------------------------------------------------------------------------
 
-TPointD ControlPointEditorTool::calculateSnap(TPointD pos) {
+TPointD ControlPointEditorTool::calculateSnap(TPointD pos, bool ctrlPressed) {
   m_foundSnap = false;
   TVectorImageP vi(TTool::getImage(false));
   TPointD snapPoint = pos;
-  if (vi && m_snap.getValue()) {
+  if (vi && (m_snap.getValue() != ctrlPressed)) {
     double minDistance = m_snapMinDistance;
 
     int i, strokeNumber = vi->getStrokeCount();
@@ -777,7 +777,7 @@ void ControlPointEditorTool::leftButtonDrag(const TPointD &pos,
 
       cp = m_controlPointEditorStroke.getControlPoint(m_lastPointSelected);
       controlPoint = TPointD(cp.x, cp.y);
-      newPos       = calculateSnap(pos);
+      newPos       = calculateSnap(pos, e.isCtrlPressed());
       delta        = newPos - m_pos + (m_pos - controlPoint);
     }
 
--- a/toonz/sources/tnztools/geometrictool.cpp
+++ b/toonz/sources/tnztools/geometrictool.cpp
@@ -476,11 +476,11 @@ public:
   virtual void onActivate(){};
   virtual void onDeactivate(){};
   virtual void onImageChanged(){};
-  TPointD calculateSnap(TPointD pos);
+  TPointD calculateSnap(TPointD pos, bool ctrlPressed);
   void drawSnap();
   TPointD getSnap(TPointD pos);
   void resetSnap();
-  TPointD checkGuideSnapping(TPointD pos);
+  TPointD checkGuideSnapping(TPointD pos, bool ctrlPressed);
 
   virtual TStroke *makeStroke() const = 0;
 };
@@ -495,12 +495,12 @@ void Primitive::resetSnap() {
 
 //-----------------------------------------------------------------------------
 
-TPointD Primitive::calculateSnap(TPointD pos) {
+TPointD Primitive::calculateSnap(TPointD pos, bool ctrlPressed) {
   m_param->m_foundSnap = false;
   if (Preferences::instance()->getVectorSnappingTarget() == 1) return pos;
   TVectorImageP vi(TTool::getImage(false));
   TPointD snapPoint = pos;
-  if (vi && m_param->m_snap.getValue()) {
+  if (vi && (m_param->m_snap.getValue() != ctrlPressed)) {
     double minDistance2     = m_param->m_minDistance2;
     m_param->m_strokeIndex1 = -1;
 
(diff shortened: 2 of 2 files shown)


beeheemooth:
Without a doubt, this is a very useful feature and I am glad about it.
 However, I want to express my opinion about the buttons. I think that the ALT key would be more appropriate in this case, since the snaping in the Brush Tool is also temporarily switched by the ALT key (that it does not start the first time is another question, khmm). Yes, in the Geometric Tool, the Alt button is also used to create shapes (Elipse, Rectangle) from the center of the click, which in my opinion is more of a user’s preference in Tool Option Bar (to which the user can assign his hotkeys if necessary) than the mode, on which needs to be temporarily switched.

I still can’t come to terms with the fact that in Toonz Raster the extension of the Frame Range chain in Eraser / Fill / Tape Tools is carried out by the SHIFT button, and Vector Frame Range uses CTRL for the same purpose (since the straight line mode is bind on SHIFT). Especially when you do a long chain of operations, and out of habit, instead of CTRL, hold SHIFT. It is necessary to cancel several operations. This is terrible, incredibly inconvenient! A general approach is needed to assign operations to the ALT, SHIFT, CTRL, Scroll buttons, especially since the user does not have the ability to reassign them. Otherwise, there will be complete chaos. Will the user always need to carefully monitor which tool he is currently using? at what level? which button should he hold in this case, ALT or CTRL? It will be easier for the user NOT to use such delusional hard-coded hotkeys.

I can give many more paradoxical situations as an example. For example: you draw using the Vector Frame Range Brush while holding ALT for snapping, switch to another frame and to extend the chain you press CTRL, and also hold ALT, because in this frame you also need it - bam ... surprise - you are resizing brushes, mua-ha-ha ...

Definitely I'm for the **ALT** button.

sorry for this indignant monologue

artisteacher:
I'd prefer not to use the alt key since I'm still hoping the modifier can be added to the polyline mode of the geometric tool for making a cusp point. It's a standard behavior for pen tools in vector programs and gives significantly more control when making custom vector shapes. We are so close to the polyline mode being fully functional as a pen tool! Ctrl is also used for the polyline mode to switch between curved/straight segments, so that could be a problem.

Maybe snapping should be a two modifier shortcut?

beeheemooth:
@artisteacher you are talking about what you “still hope”, I am talking about what is already there. There is a vector brush snapped to ALT. Why use a different button for other tools?

artisteacher:
@beeheemooth I agree they should all be the same - I just think snapping should use a different modifier. Alt currently has other uses - both in existing features and in an industry standard behavior that is currently missing.

RodneyBaker:
Shift is often used to "reverse the direction" of a navigation command.

pojienie:
@beeheemooth Don't worry, I totally agree with you. As long as the functionality is the same (reverse snap option in this case), the shortcut should be the same. Otherwise it would cause situations like you described.

@RodneyBaker Yeah, when I was trying out brush tool's reverse snap, I noticed the "File" in the menu kept being selected. Although it doesn't hinder drawing at all.

Unfortunately shift key is also used in geometric tool already. For example, to draw a square you can hold shift key and choose rectangle in geometric tool. It doesn't matter in control point tool as shift key is only used when moving segments and snap won't work when moving segments.

So we basically have 3 options: 1. don't accept this pull request(I'll keep the branch myself, btw) 2. accept the inconsistency or, 3. offend the existing users (by changing existing hot key of brush tool or geometric tool). 

Personally I think 2 and 3 are big no-no in any good quality software. We might end up choosing 1.

artisteacher:
While changing existing hotkeys is not ideal, this would be most helpful if the shortcut were consistent across all the tools that utilize snapping. Is there another option, outside of shift, ctrl, and alt?

RodneyBaker:
The shortcut key conflicts element is a different issue largely outside of the scope of this PR.
That that will need to be investigated and resolved.
I'd say the full fix there is the make sure that any shortcut key assignment can be modified by the user so that no user need fight against the default hotkey assignments... they can modify them.  

I would vote for option 2)  Accept the inconsistency (for the current timeframe)

@pojienie What is the likelyhood that we could get the 'reverse snap' option to be customizable via Configure Shortcuts menu?  

cc:  @manongjohn @martinvanzijl (as you might have important insights into resolving the problem.

artisteacher:
> The shortcut key conflicts element is a different issue largely outside of the scope of this PR.
That that will need to be investigated and resolved.

To a certain extent yes. However shift, ctrl, and alt are all used already for the geometric tool. Reverse-snap should really use something else as a modifier.

pojienie:
@RodneyBaker If all shortcut key could be customized with sensible default, it would be awesome. But, yeah, all this issue is outside of the scope of this PR.

To your question: Kinda unlikely but I'll study to see if I can do it.

@artisteacher Thank you for mentioning ctrl key is also used in geometric tool, I didn't realize this is the case when I made this PR. Yeah, in this case a new, hopefully customizable, shortcut key should be used.

Just in case... To anyone who don't know when control key is used in geometric tool: when drawing polyline, you can hold down mouse button to set "speed in" and "speed out". When holding ctrl key, it will remove "speed out" when you set next point.

jrhyme:
Cool thing, still amazed at the convenience, but. Due to known problems with vectors that cannot be solved, when connecting points, the software may crash at any time. Today 4 times lost small projects)

openanim:
Thanks a lot for theses news snap on CP editor tool, really helpfull.

beeheemooth:
@pojienie I wonder if it is possible to implement a snapping for the Cutter Tool cursor? Many users request a feature to remove vector overflow, but in OT, according to the developers, this is difficult to do. But in essence, we only need to cut the vector at the place of its intersection with another vector, and the binding would allow this to be done accurately. Removing a pieces of the vectors is possible manually. Just a thought.

pojienie:
@beeheemooth Not sure we should be discussing the feature here but I've tried implementing the feature you said. However, due to #3115, I'm not able to complete it. And, since the feature is not possible until #3115 is solved, I'm not sure if I should submit a pull request.

And yes, those developers are right, intersection detection in vector image is hard...

openanim:
@pojienie By the way, It is not related to the problem of beeheemoth nevertheless, the cutter tool also has a disadvantage when several strokes cross, it becomes very difficult to choose which to cut.

Would it be possible to restrict the choice of a stroke using a shortcut or a modifier? So that, the cut tool cursor remains along the selected stroke to facilitate cutting.

pojienie:
@RodneyBaker I've found out that the functionality has already been implemented. Although a bit differently. You can set a shortcut in File -> Configure Shortcuts -> Tool Modifiers -> Snap (or just search snap)

Once you set it, whenever you press the shortcut key, the snap option will be reversed. It works in Brush Tool, Geometric Tool, and, surprisingly, Control Point Editor tool.

I think this PR can be closed now.

@openanim Seems easy enough. I'll see what I can do.

RodneyBaker:
@pojienie Very interesting find.  
I'm a little sad to see this PR go but... as you say, the functionality is already there.
Closing.

PR #3116, opened 2020-02-27
Enable Actual Pixel View in Cleanup Preview Mode

shun-iwasawa:
This PR makes `Enable Actual Pixel View on Scene Editing Mode` preference option to work when the Cleanup Preview mode or the Camera Test mode is active.
In such case, now the cleanup camera's dpi will be used to compute proper scaling ratio in order to display pixels in actual size.


RodneyBaker:
LGTM
Thanks Shun!

PR #3120, opened 2020-02-29
Fix Palette Loading on Re-Cleanup

### LINKED ISSUE #3119: Fails to Revert Palette After Re-Cleanup
### Issue Summary

When re-cleanup the level, OT fails to set proper palette.

### Steps to Reproduce

1. Cleanup some raster level (say, cleanup A..tif to A.tlv). Note that the palette of A.tlv contains cleanup styles + default palette styles.
1. New scene.
1. Load the raster levele A..tif again.
1. Cleanup the level again.
1. In the popup, select the option `Cleanup all selected drawings overwriting those previously cleaned up` .
1. See the palette of the level A.

### Expected Results

The palette should be the same as the first cleanup result .

### Actual Results

The palette of the level A is unchanged from `Raster Drawing Palette` even though it is converted to toonz raster level.
Note that this problem does not affect the palette file (A.tpl).

### System Information
* **Operating System**: Win 8.1

shun-iwasawa:
This PR fixes #3119 .

When keeping the original palette, OT determines which palette to keep by the level type.
If the level is raster level type, the palette is loaded from cleanup destination.

However, the level type is changed to toonz raster at the line 938 ( `sl->makeTlv(outputPath);` ).
It made OT to misunderstand that the target level is always toonz raster.

So, I moved the part of keeping the original palette before changing the leve type.

### DIFF
--- a/toonz/sources/toonz/cleanuppopup.cpp
+++ b/toonz/sources/toonz/cleanuppopup.cpp
@@ -932,6 +932,33 @@ QString CleanupPopup::setupLevel() {
        notLineProcessed = (sl->getType() != TZP_XSHLEVEL);
 
   if (lineProcessing) {
+    /*--- Keep original palette which will be reverted after cleanup ---*/
+    if (m_keepOriginalPalette) {
+      if ((sl->getType() == TZP_XSHLEVEL || sl->getType() == TZI_XSHLEVEL) &&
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Nice catch.
Thanks Shun.

PR #3121, opened 2020-03-01
(WIP) Make sure opacity slider doesn't go off screen

### LINKED ISSUE #1752: Column transparency menu extends off screen
I have a couple of workflow issues with the transparency button on the levels panel.

Clicking and holding to get the transparency popup is not hugely intuitive, but a bigger problem for me is the fact that the popup doesn't take screen edge into account. When the level is up against the edge of my screen, the popup goes off the right edge of the screen (and I can't adjust the transparency, of course).

A second, minor issue is that when the level display is disabled, the display button itself disappears. This could be confusing for newer users (like me) who have to click in an empty space to get the level to reappear.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/54118682-gui-issues-with-levels-transparency-button?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

martinvanzijl:
Fixes #1752. This makes sure the opacity slider popup doesn't go off-screen to the right. Demo below - note that for column 5, the pop-up is moved left so that it fits on the screen.

popup-demo

This works for me with a single-monitor setup. I still have to test a multiple-monitor setup; will aim to do so tomorrow.

### DIFF
--- a/toonz/sources/toonz/xshcolumnviewer.cpp
+++ b/toonz/sources/toonz/xshcolumnviewer.cpp
@@ -62,6 +62,7 @@
 #include <QLabel>
 #include <QComboBox>
 #include <QPushButton>
+#include <QDesktopWidget>
 
 #include <QBitmap>
 //=============================================================================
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@shun-iwasawa  Looks like a jenkins call is required for this.

martinvanzijl:
@RodneyBaker Thanks for waiting. I found and fixed an issue when using a multiple-monitor setup: when the app window was split over two monitors, the popup would always be put on the leftmost monitor, even if the xsheet column was on the rightmost one. Have fixed this now.

Below is an example (rather wide, entire two monitors in one shot). In this case, the popup should be (and is) left in its original position.

multiple-monitor-demo

@shun-iwasawa or @manongjohn, could you please review my code before we merge? I am using the `QDesktopWidget` class, which I think is deprecated - is this OK?

RodneyBaker:
Adding this note concerning current usage of the QDesktopWidget class.
I see that there are 19 instances currently being used.
So if deprecated those might have to be reconsidered also.

https://github.com/opentoonz/opentoonz/search?p=2&q=QDesktopWidget&unscoped_q=QDesktopWidget

martinvanzijl:
@RodneyBaker Thanks for investigating. Since it is already used elsewhere within the code, I guess using QDesktopWidget here is OK for now.

RodneyBaker:
Concur.  We'll press forward with it for now then.
Thanks Martin!

PR #3138, opened 2020-03-05
Add snap at intersection in cutter tool

pojienie:
This PR allows you to cut at intersection point when using cutter tool.

demonstration video: https://youtu.be/dHxIGU7zEmo

Note that I purposefully set the "snap at intersection" as "snap" in code so that it can be switched on and off by pressing shortcut key(You can set the shortcut in File -> Configure Shortcuts -> Tool Modifiers -> Snap (or just search snap)).

### DIFF
--- a/toonz/sources/tnztools/cuttertool.cpp
+++ b/toonz/sources/tnztools/cuttertool.cpp
@@ -203,6 +203,103 @@ public:
     // glPopMatrix();
   }
 
+  double getNearestSnapAtIntersection(TStroke *selfStroke, double w) {
+    TVectorImageP vi = TImageP(getImage(false));
+    if (!vi) {
+      return w;
+    }
+
+    std::vector<DoublePair> intersections;
+    int i, strokeNumber = vi->getStrokeCount();
+    double diff;
+    double nearestW = 1000;
+    double minDiff  = 1000;
+
+    // check self intersection first
+    intersect(selfStroke, selfStroke, intersections, false);
+    for (auto &intersection : intersections) {
+      if (areAlmostEqual(intersection.first, 0, 1e-6)) {
+        continue;
+      }
+      if (areAlmostEqual(intersection.second, 1, 1e-6)) {
+        continue;
+      }
+
+      diff = abs(intersection.first - w);
+      if (diff < minDiff) {
+        minDiff  = diff;
+        nearestW = intersection.first;
+      }
+
+      diff = abs(intersection.second - w);
+      if (diff < minDiff) {
+        minDiff  = diff;
+        nearestW = intersection.second;
+      }
+
+      if (selfStroke->isSelfLoop()) {
+        diff = abs(1 - intersection.first) + w;
+        if (diff < minDiff) {
+          minDiff  = diff;
+          nearestW = intersection.first;
+        }
+
+        diff = intersection.first + abs(1 - w);
+        if (diff < minDiff) {
+          minDiff  = diff;
+          nearestW = intersection.first;
+        }
+
+        diff = abs(1 - intersection.second) + w;
+        if (diff < minDiff) {
+          minDiff  = diff;
+          nearestW = intersection.second;
+        }
+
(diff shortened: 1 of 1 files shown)


beeheemooth:
Wonderful! This option seemed to be always here, it significantly improved the Cutter Tool, and made the cutting process controlled at the intersection of lines. The only problem that I noticed is that when using the brush and using the Cutter Tool most often closed areas for filling are not defined, however this is an old problem with vector areas for filling. Cut / Paste vectors in this case solves the problem. So far I have not received a single crash. 
@pojienie Thank you very much for your work!

RodneyBaker:
Looks very good.
A very welcome addition to the Cut Tool.

With apologies for the feature creep feedback here...
A highly desired feature for Opentoonz is the ability to remove individual strokes at the end of an intersection.  This PR comes very close to that and so I cannot resist suggesting that perhaps in the future an additional option here would be to automatically delete (or alternatively to 'tag') the short  side of the line just cut, based on some user set and adjustable length.  

The deletion is the desired route but for the purpose of explanation 'tagging' would, instead of deleting, assign the eligible segment(s) designated at the time of cut to a new color swatch which in turn could either be recolored (i.e. disappeared via color change) or deleted via deletion of the sacrificial color swatch.  We can do this now manually but automating the process would be ideal.

The 'tagging' process would thereby plus up the feature in a way to improve upon Toonboom's equivalent feature where the line segments can only be deleted.

General note on usage:  This PR improves upon my current workflow that uses 'Simply Vectors' in the following ways to achieve similar results:
- Individual segments can be targeted for segmentation versus the all or nothing approach with 'Simplify Vectors'**
- The line caps don't from the newly cut segments don't need to be changed as they usually do during use of 'Simplify Vectors'.

**I note that use of 'Simplify Vectors' *after* cutting with this PR's feature is perhaps even more useful now in that it connects/joins the modified segments after the excess/unnecessary segments are removed.  

Thank you!

RodneyBaker:
Thanks @pojienie!

pojienie:
@beeheemooth Thank you for testing! The filling problem has been a problem in vector drawing but it is a bug I wouldn't be able to solve.

@RodneyBaker Thank you for merging!

I might be able to implement the "deletion" part of the feature. Like, if the option is checked and user cuts at an intersection, the shorter segment or segment that is closer to the mouse can be automatically deleted. Tagging part is a lot harder but it would be nice if it is implemented!

Also, may I ask what is "Simplify Vectors"?

RodneyBaker:
> what is "Simplify Vectors"?

Please see #1410 for information on what is currently (officially?) called 'Replace Vectors with Simplified Vectors'.  I've used the terms so often I've grown weary of the longer name so refer to it as 'Simplify Vectors'.  ;)

It can be found in the menu via Level > Optimize > Replace Vectors with Simplified Vectors.

RVwSV

pojienie:
Oh, I see! Thank you for letting me know such a powerful feature!

melieconiek:
thanks, very useful :D

can this be made to be enabled/disabled by holding down a modifier key? CTRL perhaps.

pojienie:
@melieconiek I've thought about that but I decided to go against it. You can't assign a single control key in configure shortcut menu so it must be hard-coded. We had a bit of discussion in #3110 and I don't think it's a good idea to use hard-coded shortcut key here.

While a bit different, you can use the shortcut menu to configure the "snap" key as I've mentioned in the PR.

openanim:
About the shortcut, very slight UI issue, when using it with the brush, it does not activate the drop-down menu of snap levels (low / med / hight /) if it has not been activated manually beforehand. Something that can be fixed later.

melieconiek:
@oijienie That's significantly slower and susceptible to mixups, but I understand the issue. We need a solution for hardcoded keybindings across the board, including the modifier keys--it's a problem.

Would it be possible to put in a falloff--the snap wouldn't kick in until it's a certain distance from the closest intersection? Do you think that would work?

I've been using this today for work, very handy, thank you again. :)

PR #3139, opened 2020-03-06
Fix Particles Fx Control Image Obtaining

shun-iwasawa:
This PR will fix obtaining area of control images of Particles Fx, especially for the images with inifinite bounding box.
Input images with infinite bounding box can be input from most of zerary effects like Linear Gradient Fx. Currently in such case Particles Fx obtains the control image with the same size as the output tile size.
However, there is a case that the particle is generated outside of the camera when the source rectangle or the source control image is larger than the camera size. Particles born outside of the control image uses transparent (RGBA=0,0,0,0) pixel as reference.

To resolve such problem, this PR will enlarge obtaining size of control image to cover the source region.


RodneyBaker:
Thank you Shun!

PR #3142, opened 2020-03-08
Improve multi-layer style picker

### LINKED ISSUE #2843: Improve switching vector levels with Style Picker
### Issue Summary
When `Multi Layer Style Picker: Switch Levels by Picking` is enabled and you try to single click a stroke on a Vector level in another level, OT will not switch to that level, not even if you just single click over the center of the stroke.

To get it to work you have to click-drag over the center of the stroke.

This can probably use a similar enhancement made for Improve vector stroke selection (#2809)

### Steps to Reproduce
1) Enable `Multi Layer Style Picker: Switch Levels by Picking`  in Preferences
2) Create a Vector Level (A) and draw something in it
3) Create a 2nd Vector Level (B) and draw something in it
4) Switch to the Style Picker Tool
5) While on Level B click over a stroke in Level A -> Nothing happens
6) While on Level B cick then drag over the center of stroke in Level A -> Finally switches as it crosses the center of the stroke

### Expected Results
Regardless if single clicking or click-dragging, the level is switched no matter what part of the stroke you select in the other level

### System Information
* **OpenToonz Version**: 1.3

martinvanzijl:
Fixes #2843. Allows clicking anywhere on a stroke in another layer to pick it -- not just the center.

Demo below:

style-picker-demo

**Still to do**:

* Get the actual pixel size from the viewer, instead of hard-coding it 1.0. Here: https://github.com/opentoonz/opentoonz/blob/01ee1736d95607fcfe1815c1948a18d121fe6667/toonz/sources/toonzlib/stagevisitor.cpp#L256
* Maybe use a shared function instead of copying and pasting the code from `TTool::Viewer::doPickGuideStroke`.

### DIFF
--- a/toonz/sources/toonzlib/stagevisitor.cpp
+++ b/toonz/sources/toonzlib/stagevisitor.cpp
@@ -250,9 +250,22 @@ void Picker::onImage(const Stage::Player &player) {
     if (r) styleId   = r->getStyle();
     if (styleId != 0)
       picked = true;
-    else if (vi->getNearestStroke(point, w, strokeIndex, dist2) &&
-             dist2 < m_minDist2)
-      picked = true;
+    else if (vi->getNearestStroke(point, w, strokeIndex, dist2)) {
+      // based on TTool::Viewer::doPickGuideStroke
+
+      double pixelSize = 1.0; // TODO: Get actual pixel size from viewer.
+      double maxDist   = 5 * pixelSize;
+      double maxDist2  = maxDist * maxDist;
+      double checkDist = maxDist2 * 4;
+
+      TStroke *stroke = vi->getStroke(strokeIndex);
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@martinvanzijl I know this is a WIP but am curious if there is anything that can be done to move this PR forward.  Perhaps what you have now is sufficient enough to implement while pushing the 'still to do' elements toward other PRs?

Your call, I just hate to see a good PR stagnate.  :)

martinvanzijl:
@RodneyBaker Thanks for the reminder; I was meaning to get back to this.

For getting actual pixel size from the viewer, it seems like `m_minDist2` is the pixel size to the power 4, so I take the square root of the square root to get the pixel size.

For using a shared function, we can probably refactor this later. The amount of duplicated code is rather small, anyway.

Ready for review/merging.

PR #3144, opened 2020-03-10
Fix MacOS Camera Capture Resolution Problem

### LINKED ISSUE #3114: MacOS Camera Capture Fails to Capture with Proper Resolution
### Issue Summary

The problem was [reported in Japanese forum](https://groups.google.com/forum/?hl=ja#!category-topic/opentoonz/%E4%B8%8D%E5%85%B7%E5%90%88%E5%A0%B1%E5%91%8A/59-2-38kzxA) .
In MacOS version, Camera Capture feature always capture images with specific resolution regardless of the value specified in the `Resolution` combo box.
For instance, Hozan L835 USB camera always captures with 1280x960 pixels and MacBook built-in camera (FaceTime camera) always captures with 1280x720 .

Please see the original post for screenshots.
After it takes the first frame of level A, OT shows "Warning" since the resolution set in the combo box (2592x1944) is not match with the resolution of the first feame which has been wrongly taken with 1280x960.

### Steps to Reproduce

- In MacOS version, use the Camera Capture.
- Take images with some resolution values and check if the captured levels are with correct resolutions.

### Expected Results

Levels should be with the resolutions set in the `Resolution` combo box.

### Actual Results

Levels always have specific resolutions, regardless of the `Resolution` combo box. 

### System Information
* **OpenToonz Version**: 1.3 and 1.4
* **Operating System**: MacOS

shun-iwasawa:
This fixes #3114 .
There seems to be a difference in behavior of `QCamera` between Win and MacOS.
In MacOS stopping and restarting the camera is not needed, and, what is worse, it seems to cause unwanted reset of resolution. 
Also I removed `m_dummyViewFinder` as it seems to be unnecessary now.

### DIFF
--- a/toonz/sources/toonz/penciltestpopup.cpp
+++ b/toonz/sources/toonz/penciltestpopup.cpp
@@ -1494,10 +1494,6 @@ PencilTestPopup::PencilTestPopup()
   m_subHeightFld        = new IntLineEdit(this);
   QWidget* subCamWidget = new QWidget(this);
 
-#ifdef MACOSX
-  m_dummyViewFinder = new QCameraViewfinder(this);
-  m_dummyViewFinder->hide();
-#endif
   //----
 
   m_resolutionCombo->setMaximumWidth(fontMetrics().width("0000 x 0000") + 25);
@@ -1903,10 +1899,6 @@ void PencilTestPopup::onCameraListComboActivated(int comboIndex) {
 
   m_currentCamera = new QCamera(cameras.at(index), this);
   m_deviceName    = cameras.at(index).deviceName();
-#ifdef MACOSX
-  // this line is needed only in macosx
-  m_currentCamera->setViewfinder(m_dummyViewFinder);
-#endif
 
   // loading new camera
   m_currentCamera->load();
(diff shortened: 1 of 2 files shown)


RodneyBaker:
I cannot fully test this as it is Mac related but foresee no issues going forward.

cc:  @artistteacher (keeping you in the loop)

manongjohn:
If you cannot test a mac specific change, then why are you merging it? 

Surely you could have asked and waited for a mac user to build and confirm it and check for side effects that are not obvious from looking at the code.

RodneyBaker:
@manongjohn Thanks for the feedback.  
Here is a list of PRs that would also benefit from your attention:
https://github.com/opentoonz/opentoonz/pulls

shun-iwasawa:
I'll ask the original reporter in Japanese forum to test the fix. Thanks!

MrBlockCat:
Hi! I'm on a MacBook, and while I didn't test the old version as a control group to make sure it does have the issue, if I followed the steps in the original issue correctly it appears this is now fixed on the nightly build!

PR #3145, opened 2020-03-10
Fix Background Color of Export Level Popup Fails to be Alpha-enabled in Non-English UI

shun-iwasawa:

- In non-English UI, the background color field in the Export Level popup does not become alpha-enabled even if the exported format allows to have transparent background.

When checking whether the format has alpha channel, the property name should be obtained by using `TProperty::getName()` instead of `TProperty::getQStringName()` , since the latter function returns translated string for UI.


RodneyBaker:
LGTM

PR #3147, opened 2020-03-11
Fix Crash on Copying Column Connected to Plugin Fx

shun-iwasawa:
This PR fixes a crash problem which can be reproduced by following steps:

-  Load some level to the column 1 and build the fx schematic like as follows;
plugin_crash
  In the above example, I used `DWANGO_LightBloom` plugin fx.
- Select the column 1 and try to copy it ( `Ctrl+C` ). OT crashes immediately.

The cause of the crash was that the plugin fx has been wrongly derived from `TZeraryFx` instead of `TRasterFx` .


RodneyBaker:
I'm not set up right now for using of the plugins but will move in that way shortly if no one tests and validates this first.   The code itself... LGTM.  :)

RodneyBaker:
Thank you Shun!
LGTM

Issue #514, opened 2016-06-23
feature request: Improve Gap Check lines


blurymind:
Currently gap check lines have a number of problems:

1. They are too thick, so they actually get in the way of visibility:
![gapcheck](https://cloud.githubusercontent.com/assets/6495061/16311333/11776df8-3968-11e6-8b01-84d06485c1e7.PNG)
2. They become black when you enable fill check - considering that you will most probably have gap and fill check enabled together, and that your lineart is likely going to be black - the black color on these lines is not helpful here. 
3. Gap check lines are dependent on the camera zoom - that makes them hard to see if you zoom out a lot - gap check lines thickness should not be dependent on zoom level - it should always be one value.

-We need a way to control gap check line thickness, or have that thickness set to 2 pixels - and have it independent from zoom level
-We need gap check lines to keep their color (magenta) when fill check is enabled.

blurymind:
Very often gap check draws a gap line where there is no gap:
![gapcheckwrong](https://cloud.githubusercontent.com/assets/6495061/16316903/9917e644-3980-11e6-9a07-21a9a0db0ed4.PNG)

Why does it do that?

gab3d:
Currently (as of last nightly build) Gap Check lines are hardly seen.
Would it be possible to change the color of them to a more contrasting one, or just to increase its width a little bit?
Thanks!

![gapcheck](https://user-images.githubusercontent.com/18104091/30508265-09f0c36e-9a69-11e7-89cc-aa28240bedbc.png)

RodneyBaker:
It would be good to review this issue and update status.

RodneyBaker:
There are a number of issues relate to Gap Check that fall more into the new features category than bug but current workflow can be troublesome.  I've retitled this issue as "Gap Check Improvements" to be a general weigh station and collection point for topics related to Gap Check as most of these are known issues awaiting a fuller scrub of the gap check feature set.

A few issues with the Gap Check that have been raised include:
- [ ] Line Thickness
- [ ] Setting the Distance for the Gap Check to use (via Tape Tool options)
What might not be obvious to users is that the Distance setting of the Tape Tool is what controls Gap Check line display.  So for anyone that can't get Gap Check to work adjust that setting.  Note that the setting is grayed out (i.e. unavailable) until specific other options of the Tape Tool are turned on/off.
- [ ] Color
- [ ] Accuracy / False positives

RodneyBaker:
Slipping to 2019 development cycle.

[...]

PR #3148, opened 2020-03-11
Preserve gap indicator color when fill check is enabled

martinvanzijl:
For #514, addresses this issue about the gap check indicators:

> 2. They become black when you enable fill check - considering that you will most probably have gap and fill check enabled together, and that your lineart is likely going to be black - the black color on these lines is not helpful here.
 
This fixes that, making sure they are still magenta even when fill check is enabled. Demo below:

gap-check-demo


RodneyBaker:
Very nice.
LGTM

PR #3160, opened 2020-03-18
Update .gitignore to re-ignore .DS_Store files

MrBlockCat:
This (re)adds GitHub ignoring .DS_Store files that Macs create

Sorry for the mess! It turns out since I'm inexperienced at GitHub, I accidentally undid ignoring .DS_Store files in the PR to make the Plastic Tool icon in the tool bar vector. I think I've got things sorted now, I'm using separate branches which is certainly progress!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,6 @@ ipch/
 *.creator
 *.files
 *.includes
+
+#macOS
+.DS_Store


RodneyBaker:
Somewhat my fault.  
I saw that removal in the code and should have asked if that's really what you wanted to do.

PR #3161, opened 2020-03-18
Makes all Command Bar/Xsheet Toolbar icons SVG

### LINKED ISSUE #3128: Some Command Bar Icons are Pixelated
### Issue Summary
On the Command Bar, some icons appear pixelated on screen, with the exception of text-based ones like 1's, 2's, and 3's. There are some other exceptions I've noticed too, namely the three New Level buttons. I've attached an example.

Screen Shot 2020-03-03 at 12 08 13 PM

### Steps to Reproduce
Use a high resolution display (this becomes pretty noticeable to me because I'm using a high-resolution 2560 x 1600 display.)

### Expected Results
All icons show up at native resolution

### Actual Results
Some Command Bar icons show up pixelated

### System Information
* **OpenToonz Version**: 1.4.0
* **Operating System**: macOS Catalina
* **CPU**: Quad Core i5
* **Memory**: 8 GB LPDDR3
* **Graphics Card**: 
* **Graphics Tablet**:

MrBlockCat:
Fixes #3128 

toolbar_arrow.png was replaced with toolbar_arrow.svg, but I don't know where it's used in the code so this shouldn't be pulled into OpenToonz until that's sorted out!

Also, sorry about the mess last time I tried making this! I'm using separate branches now, which should hopefully make things much, much more manageable.


MrBlockCat:
Sounds like toolbar_arrow probably'll work with no more commits needed! All this needs to wait on now is my ability to make minor tweaks to output.svg, and in the meantime I'll try and set up a build environment to test stuff!

MrBlockCat:
I'm having an issue when I build OpenToonz myself where the toolbar arrow (the one that expands it if it would otherwise overflow) shows up as a grey (barely visible with the Dark theme) vector arrow, rather than the normal black pixely one. I've attached photos. I don't know why this happens or where the grey arrow is even coming from (it's not the new one I made), but it makes it so I can't test whether or not toolbar_arrow.svg shows up! It also happened when I built the main opentoonz branch, so I don't think this issue is caused by this commit. I'm on macOS Catalina, building with Xcode...

Normal icon:
Screen Shot 2020-03-18 at 9 07 13 PM

Mysterious grey icon (it's there I promise):
Screen Shot 2020-03-18 at 9 07 40 PM

That said, the other icons do show up normally!

RodneyBaker:
In Preferences > Tools at the bottom is an option that should adjust the results of those arrow icons.
That option turns on/off the context sensitive response of the Tool icons.

The arrows are generally there to express that not all Tools are appearing although it becomes the grayed out icon in some cases when all available tools are displayed but the toolbar space isn't quite adequate to terminate the whole toolbar properly.

MrBlockCat:
The Toolbar Display Behaviour setting seems to only make the arrow appear/disappear (depending on if all the icons can fit), not make it turn back into the normal black one. Clicking the arrow does still pull up more icons, meaning they don't all fit, which makes me think what you described may not be quite what's happening...

manongjohn:
That double-arrow button you are showing is the QToolbar widget's expansion button which is automatically added when all the buttons on the toolbar cannot be displayed. To my knowledge we don't have much control over that.

MrBlockCat:
Ah, thanks! I thought there was something weird about it. Is the black, pixelated one just what it looks like in older versions of QT, which presumably is what OpenToonz is publicly built with, then?

RodneyBaker:
The black arrows can be clicked on to show the hidden tools/icons.
The grey arrows mean there is nothing else to see (therefore deactivated and cannot be clicked),

manongjohn:
Actually.  It appears the QToolbar extension button can be somewhat styled in the QSS files using `QToolButton#qt_toolbar_ext_button {}`  Not sure what full set of options are available for styling it.

manongjohn:
The arrows for the extension button is the stock image, likely a png image. The image can be replaced with styling, though I'm not sure how to change the image to show a down arrow for the vertical toolbars vs right arrow for horizontal toolbars like it does now. Might need to be code-styled to use a specific image.

FYI, I don't believe the toolbar_arrow.png is being used anywhere.

MrBlockCat:
@RodneyBaker I see what you're describing now, but that's a different grey arrow! It seems to be black when not all tools are visible, grey when all tools are visible but the "Collapse toolbar" arrow can't fit, and it disappears completely when the "Collapse toolbar" arrow can fit.

Here's the Collapse toolbar arrow, since there are so many arrows, to avoid confusion:
Screen Shot 2020-03-18 at 11 09 15 PM

Interestingly, now that I know how to find that other grey arrow, I tried doing that on my nightly build with its updated (I'm assuming that's what it is) arrow, and it does still slightly change color (screenshots below) when all tools but not the Collapse toolbar arrow are visible.

Nightly with all tools and Collapse toolbar arrow visible
Screen Shot 2020-03-18 at 11 12 18 PM
(It's not there)

Manual with all tools and Collapse toolbar arrow visible
Screen Shot 2020-03-18 at 11 13 05 PM
(It's not there)

Nightly with all tools but not Collapse toolbar arrow visible
Screen Shot 2020-03-18 at 11 13 45 PM
(It's grey)

Manual with all tools but not Collapse toolbar arrow visible
Screen Shot 2020-03-18 at 11 15 11 PM
(It's even lighter grey)

Nightly with not all tools visible and Collapse toolbar arrow not visible
Screen Shot 2020-03-18 at 11 15 47 PM
(It's black)

Manual with not all tools visible and Collapse toolbar arrow not visible
Screen Shot 2020-03-18 at 11 16 28 PM
(It's grey)

@manongjohn From googling it looks like the background color of the button might change its color? I'm not sure, and that all goes well above my head, so I'd have to leave that to someone else! If it's easiest for everyone, I can leave in the toolbar_arrow.svg in this commit to replace toolbar_arrow.png in case that comes in handy for replacing the existing arrow with that one and make a separate issue to fix that arrow (but I wouldn't be able to make that issue's PR)

RodneyBaker:
The grey arrow will not appear after the big black arrow (pointing up) because that black arrow designates there is nothing else below.  Any blank space thereafter is filled by a spacer as the toolbar area gets larger.  Perhaps it's that spacer your refer to as the empty gray icon.

MrBlockCat:
If I'm interpreting what you said correctly, the spacer you refer to is a different thing than the new QT gray arrow that's usually hard to see. Let me try and explain: What the nightly build's QT arrow does works fine in all scenarios color wise, the only problem is it's pixelated! But when I manually build with a newer version of QT, QT instead uses a vector arrow with lighter colors, which causes a problem in most of OpenToonz's themes, because the new QT arrow uses grey where the old arrow used black. Since most themes use a shade of grey for the background, it makes the new arrow hard to see, especially in OT's Dark theme and least so in the Light theme, which is what I was trying to show in the set of those first three screenshots.
What I'm hoping is possible is either to change the arrow's color from the default to black, or maybe even better change it to use the toolbar_arrow.svg icon so QT can't mess with it more in the future.

Hopefully that makes sense, this is very hard for me to explain because there are so many different arrows going on!

MrBlockCat:
I noticed this used over at [QDarkStyleSheet](https://github.com/ColinDuquesnoy/QDarkStyleSheet/blob/master/qdarkstyle/style.qss), which might reveal how to use the button:

```
QToolButton#qt_toolbar_ext_button {
  background: #32414B;
  border: 0px;
  color: #F0F0F0;
  image: url(":/qss_icons/rc/arrow_right.png");
}
```

Maybe if we used `color: #000000;` and/or `image: url("PathGoesHere");` we could make the arrow uniform across QT versions? Unfortunately I don't know where that code would go, but I was able to find the toolbar's code [here](https://github.com/opentoonz/opentoonz/blob/9e8f4af94befc2a38ce67cd3707fc712c0701498/toonz/sources/toonz/toolbar.cpp) if that helps!

MrBlockCat:
For now, it's probably best if this is merged as is, and I can make a separate issue about the double arrow always being gray. Please let me know if you guys disagree or find an issue with this PR!

MrBlockCat:
I said I was done, but I want to get the arrows just right... Sorry! The latest look much nicer in preview, I'll test them in OpenToonz again soon

MrBlockCat:
I think they look pretty good now! Not quite the same as the originals but about as close as I could get them.
Here's some of them in practice:
Screen Shot 2020-03-25 at 12 58 55 AM
Screen Shot 2020-03-25 at 12 59 47 AM

Versus the originals:
Screen Shot 2020-03-25 at 1 00 09 AM
Screen Shot 2020-03-25 at 1 00 27 AM

RodneyBaker:
Running through the build nothing pops out as out of place... but I am not on a high rez monitor either.

We'll give others a little time to check things out but from here things look good.

PR #3165, opened 2020-03-18
Curves Fx Enhancement

shun-iwasawa:
This PR will improve the Curves Fx . 

- Enabled to stretch handles to any lengths. It can be expanded even to outside of the graph area.
- Handle positions are now displayed and controlled in decimal number.
- Added input fields for displaying and editing the selected point.
- Enabled to enlarge the graph (x2 scaled) for detailed operation.
- Added an option to change the value range from `0-255` to `0.0-1.0` .
- Narrowed the minimum interval between control points from 16 to 4 in order to gain consistency with the one of Adobe Photoshop's tone curve feature.
- Keyboard controls are now changed as follows:
  - **Arrow keys** : move the selected point to each direction by 1 unit.
  - **Shift+Arrow keys** : move the selected point to each direction by 10 units
  - **Ctrl + Right or Left arrow key** : switch the selected point.
- Changed the interpolation method between key frames. Instead of just interpolating position of each point separately, left and right handles are now interpolated in angles and lengths from connected control points.

**NOTICE**
- the `Range` and the `Enlarge` states are not saved in the scene, as it is just for UI settings.
- This PR will replace the existing Curve Fx. The fx created in the previous versions can be loaded in this version and vice versa. Please note that the curve shape may become different especially in animation (i.e. between the key frames), since the interpolation method has been changed.

### DIFF
--- a/toonz/sources/common/tparam/ttonecurveparam.cpp
+++ b/toonz/sources/common/tparam/ttonecurveparam.cpp
@@ -148,11 +148,81 @@ void TToneCurveParam::setCurrentChannel(ToneChannel channel) {
 //---------------------------------------------------------
 
 QList<TPointD> TToneCurveParam::getValue(double frame) const {
+  // compute the handle angle and length
+  // in case the handle length is 0 on one side, take the oppositte handle to
+  // calculate the angle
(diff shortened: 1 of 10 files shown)


RodneyBaker:
Very nice!
LGTM

Issue #2864, opened 2019-10-30
Tangled Commands


beeheemooth:
### Issue Summary

Here I use the Brush Tool with Range mode. I need to draw a line in the first frame, so that it snaps to the line that is already drawn. To do this, I hold down the ALT key to enable temporary snapping. In the sixth frame, I need to draw another line, but this is not the last line in this range, so I hold down the CTRL key to be able to continue drawing lines in the "Range" mode and also hold down the ALT key for snapping. Oops ... CTRL + ALT resize the brush.
### Screenshot or Video Reference

![ezgif-1-152a5d6e1b4b](https://user-images.githubusercontent.com/25724133/67857212-d5702d00-fb1e-11e9-93c8-21919badf968.gif)

### System Information
* **OpenToonz Version**: latest
* **Operating System**: Windows 8.1

RodneyBaker:
Can you reassign your shortcut keys to  allow for the desired workflow?

beeheemooth:
@RodneyBaker This is the problem: shortcuts cannot be reassigned. If I could do this, I would completely remove the ability to resize the brush (and eraser) using shortcuts - this is a very inaccurate and difficult to control function. It's my opinion. I would also change the ability to draw straight vector lines from the SHIFT key to the CTRL key, since the extension of the frame range in Toonz Raster is done by holding down the SHIFT key, while in Vector Levels this is done by holding down the CTRL button - this is very inconvenient and you can always mix up the buttons when you forget what type of level you draw.

RodneyBaker:
Agree a goal would be to have consistency of shortcut keys across all level types to the greatest extent possible.

artisteacher:
This is also a problem with touch gestures and the double-click behavior #2338.

andeon:
@RodneyBaker Apparently, this has been fixed due to the PR linked to this issue, so the issue can be closed.

PR #3170, opened 2020-03-22
Add option to disable use of ctrl+alt for changing brush size

martinvanzijl:
For #2864. This adds a checkbox to the _Tools_ section in the _Preferences_ window: "Use Ctrl+Alt to Resize Brush". This is checked (on) by default. If you un-check it, the Ctrl+Alt shortcut will not resize the brush when you move the mouse.

Video demo below. Here I am moving the mouse with Ctrl+Alt pressed, before and after un-checking the preference:

resize-brush-demo

This hopefully fixes the problem in the issue summary, but there are other shortcut issues mentioned in the comments (https://github.com/opentoonz/opentoonz/issues/2864#issuecomment-547905964), which I haven't addressed.

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -333,6 +333,9 @@ public:
   int getLevelBasedToolsDisplay() const {
     return getIntValue(levelBasedToolsDisplay);
   }
+  bool useCtrlAltToResizeBrushEnabled() const {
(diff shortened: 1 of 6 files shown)


RodneyBaker:
Thank you Martin!
LGTM

martinvanzijl:
@RodneyBaker Thanks for merging. Should I update the documentation to mention the new check-box? If so, where do I do this?

PR #3176, opened 2020-03-24
fix tape tool not connecting two strokes when both strokes are not grouped

### LINKED ISSUE #2937: After grouping, the Tape Tool does not connect ungrouped vectors.

ezgif-6-8dd207006146

### Steps to Reproduce

1. Draw three vector strokes in the same order as in the video.
2. Select and group the middle stroke (RMB> Group)
3. Using the Tape Tool, try connecting the first stroke to the third. For some reason this is not possible, although neither the first stroke nor the third is grouped.
### Expected Results

Vectors must connect as they are not grouped.
### Actual Results

Vectors do not connect
### System Information

pojienie:
video: https://youtu.be/GfrtMT2YZL8

this fixes #2937. The reason why it happens is because the strokes are not in the same ghost group. To the user, they simply appear to be not grouped. But in the code, they are assigned a ghost group.

whenever you add a stroke to a vector image, if you're not in group, you're adding the stroke to an existing or a new ghost group.

Tape tool checks if two strokes that are being taped are in the same group. If they do, you can tape them. Otherwise, you can't. However, it also checks if the two strokes are in the same ghost group even when they seem to be just two not-grouped strokes to the users.

This PR makes tape tool check one additional condition before skipping a stroke: one of them are in any not-ghost group. That is, if they are both in ghost group, check them instead of skipping them.

### DIFF
--- a/toonz/sources/tnztools/vectortapetool.cpp
+++ b/toonz/sources/tnztools/vectortapetool.cpp
@@ -240,14 +240,14 @@ public:
   //-----------------------------------------------------------------------------
 
   bool onPropertyChanged(std::string propertyName) override {
-    TapeMode                 = ::to_string(m_mode.getValue());
-    TapeSmooth               = (int)(m_smooth.getValue());
-    std::wstring s           = m_type.getValue();
+    TapeMode       = ::to_string(m_mode.getValue());
+    TapeSmooth     = (int)(m_smooth.getValue());
+    std::wstring s = m_type.getValue();
     if (!s.empty()) TapeType = ::to_string(s);
-    TapeJoinStrokes          = (int)(m_joinStrokes.getValue());
-    AutocloseFactor          = (double)(m_autocloseFactor.getValue());
-    m_selectionRect          = TRectD();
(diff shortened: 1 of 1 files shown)


RodneyBaker:
The inability to Tape lines in different groups is still something of an issue and one we need to stare at along with bigger picture processes such as color fill of things not assigned to same group.  That obviously is for another time/development.

You do have me curious if color fill might not checking for something similar in ghost groups as well.

As for this PR...
LGTM

Thanks @pojienie

pojienie:
@RodneyBaker Thank you for merging!

Yeah, a lot decision and, code modification to be made if we want to allow tape tool to tape strokes that are different groups.

Also, just tested, color fill also checks ghost group: https://youtu.be/f2Sa0k-2a9I
Do I need to create an issue for this?

RodneyBaker:
Yes, I would create a new issue for that.
Nice find!

melieconiek:
@pojienie Thank you for this, and for looking into the related fill issue. :)

PR #3178, opened 2020-03-24
Add auto delete option in cutter tool

pojienie:
video https://youtu.be/RvMJdPBsA-c

The reference picture is from: https://youtu.be/iT_oMHF8Bs0?t=336.

this PR allows cutter tool to automatically remove stroke that is closer to the mouse when cutting at intersection. 

This almost always works. However, as shown in the video, if the intersection looks like this:
a
It may not detect it as an intersection. A workaround for this is to disable auto remove and remove them yourself.

"remove vector overflow" does it by checking edge list. However, from what I see, edge list only works in closed area and thus, I wasn't able to adapt it to the cutter tool.

I wish someone could take this and improve upon it so that it always works.

### DIFF
--- a/toonz/sources/tnztools/cuttertool.cpp
+++ b/toonz/sources/tnztools/cuttertool.cpp
@@ -324,6 +324,7 @@ public:
     if (vi->getNearestStroke(pos, pW, strokeIndex, dist) && pW >= 0 &&
         pW <= 1) {
       double w;
+      double snapW;
 
       strokeRef = vi->getStroke(strokeIndex);
 
@@ -350,21 +351,29 @@ public:
         w = strokeRef->getParameterAtLength(len);
       }
 
+      snapW = w;
       if (m_snapAtIntersection.getValue()) {
-        w = getNearestSnapAtIntersection(strokeRef, w);
+        snapW = getNearestSnapAtIntersection(strokeRef, w);
       }
 
       std::vector<DoublePair> *sortedWRanges = new std::vector<DoublePair>;
 
       if (strokeRef->isSelfLoop()) {
-        sortedWRanges->push_back(std::make_pair(0, w));
-        sortedWRanges->push_back(std::make_pair(w, 1));
+        sortedWRanges->push_back(std::make_pair(0, snapW));
+        sortedWRanges->push_back(std::make_pair(snapW, 1));
       } else {
-        if (w == 0 || w == 1)
+        if (snapW == 0 || snapW == 1)
           sortedWRanges->push_back(std::make_pair(0, 1));
         else {
-          sortedWRanges->push_back(std::make_pair(0, w));
-          sortedWRanges->push_back(std::make_pair(w, 1));
+          if (snapW != w && m_autoDelete.getValue()) {
+            if (snapW < w)
+              sortedWRanges->push_back(std::make_pair(0, snapW));
+            else
+              sortedWRanges->push_back(std::make_pair(snapW, 1));
+          } else {
+            sortedWRanges->push_back(std::make_pair(0, snapW));
+            sortedWRanges->push_back(std::make_pair(snapW, 1));
+          }
         }
       }
 
--- a/toonz/sources/tnztools/cuttertool.cpp
+++ b/toonz/sources/tnztools/cuttertool.cpp
@@ -25,6 +25,7 @@
 using namespace ToolUtils;
 
 TEnv::IntVar SnapAtIntersection("CutterToolSnapAtIntersection", 0);
+TEnv::IntVar AutoDelete("CutterToolAutoDelete", 0);
 
 //=============================================================================
 namespace {
@@ -166,15 +167,20 @@ public:
 
   TPropertyGroup m_prop;
   TBoolProperty m_snapAtIntersection;
+  TBoolProperty m_autoDelete;
 
   CutterTool()
       : TTool("T_Cutter")
       , m_mouseDown(false)
       , m_cursorId(ToolCursor::CutterCursor)
-      , m_snapAtIntersection("Snap At Intersection", false) {
+      , m_snapAtIntersection("Snap At Intersection", false)
+      , m_autoDelete("Auto delete", false) {
     bind(TTool::VectorImage);
     m_prop.bind(m_snapAtIntersection);
+    m_prop.bind(m_autoDelete);
+
     m_snapAtIntersection.setId("Snap");
+    m_autoDelete.setId("AutoDelete");
   }
 
   ToolType getToolType() const override { return TTool::LevelWriteTool; }
@@ -428,6 +434,7 @@ public:
 
   void onActivate() override {
     m_snapAtIntersection.setValue(SnapAtIntersection ? 1 : 0);
+    m_autoDelete.setValue(AutoDelete ? 1 : 0);
   }
   void onEnter() override {
     if ((TVectorImageP)getImage(false))
@@ -444,12 +451,19 @@ public:
 
   void updateTranslation() override {
     m_snapAtIntersection.setQStringName(QObject::tr("Snap At Intersection"));
+    m_autoDelete.setQStringName(QObject::tr("Auto Delete"));
   }
 
   TPropertyGroup *getProperties(int targetType) override { return &m_prop; }
 
   bool onPropertyChanged(std::string propertyName) override {
+    if (propertyName == m_snapAtIntersection.getName()) {
+      if (!m_snapAtIntersection.getValue()) m_autoDelete.setValue(false);
+    } else if (propertyName == m_autoDelete.getName()) {
+      if (m_autoDelete.getValue()) m_snapAtIntersection.setValue(true);
+    }
     SnapAtIntersection = (int)(m_snapAtIntersection.getValue());
+    AutoDelete         = (int)(m_autoDelete.getValue());
     return true;
   }
 


RodneyBaker:
Thanks so much for this @pojienie 
This is going to make a lot of Opentoonz users very happy.

My initial testing suggests that the deleting is a little too aggressive in that it will remove a segment but then also remove another segment next to it.  Perhaps if a delay is built in so that a pen movement or touching of the tablet doesn't get detected too soon after the first trigger that might prevent that unintentional removal?

Back to testing! :)

pojienie:

I didn't have a drawing tablet so I never have the problem you described... I am going to assume that, whenever you use the pen to touch the drawing tablet, it's as if mouse's left button being clicked. But in your case, it clicks twice.

I can add a built-in delay by getting current time every time you left click/pen touches tablet and check if the times between clicks are within certain threshold. If they are, don't do anything. Otherwise, do things as usual. 

I have two questions here: 

1. Is it the correct way to solve this problem? Maybe it should be applied more generally instead of just cutter tool so that every tool can benefit from this (This will require a separate PR). Maybe the threshold should be in preference so that user can adjust it. Can you give me some guidance here as to how I should proceed?

2. How long should the threshold be? Too long and it will feel not-responsive. Too short and the problem will persist, just less likely. I've added a patch here. Could you apply the following patch and test the threshold by modifying 0.5 for me?

```
--- a/toonz/sources/tnztools/cuttertool.cpp
+++ b/toonz/sources/tnztools/cuttertool.cpp
@@ -156,6 +156,7 @@ public:
 
 class CutterTool final : public TTool {
 public:
+  std::chrono::time_point<std::chrono::system_clock> m_lastTimeMouseDown;
   bool m_mouseDown;
 
   TPointD m_vTan;
@@ -171,6 +172,7 @@ public:
 
   CutterTool()
       : TTool("T_Cutter")
+      , m_lastTimeMouseDown(std::chrono::system_clock::now())
       , m_mouseDown(false)
       , m_cursorId(ToolCursor::CutterCursor)
       , m_snapAtIntersection("Snap At Intersection", false)
@@ -307,6 +309,16 @@ public:
   }
 
   void leftButtonDown(const TPointD &pos, const TMouseEvent &) override {
+    // if time between two consecutive clicks are too short, ignore it.
+    std::chrono::time_point<std::chrono::system_clock> now;
+
+    now                                = std::chrono::system_clock::now();
+    std::chrono::duration<double> diff = now - m_lastTimeMouseDown;
+    m_lastTimeMouseDown                = now;
+    if (diff.count() < 0.5) {  // 0.5 seconds
+      return;
+    }
+
     if (getViewer() && getViewer()->getGuidedStrokePickerMode()) {
       getViewer()->doPickGuideStroke(pos);
       return;
```

melieconiek:
I am not experiencing the multiple cuts like that. I'd be against adding any kind of delay to the tool since it might causes issues when making quick, successive cuts.

Everything is working fine here on my end. This is going to be a huge time saver. Thank you! :D

It we be awesome to see this adapted to a cutter/line segment eraser tool similar to what we have in Blender, CSP, Cacani, and Harmony. They all use an intersecting line selection method (much faster--requires less precision,) can handle multiple lines at once, and can cut out line segments between two intersections.

RodneyBaker:
@pojienie  I can't test the patch at the moment as I am away from my desktop and full build capability.  Hope to settle back into it soon.  

Thanks for the validation/feedback @melieconiek 
Interesting that you aren't experiencing the same issue as it's very noticeable on my end.
I'll explore my pen settings and see if perhaps the sensitivity is a primary factor.
I agree that if others aren't experiencing similar problems we shouldn't introduce any intentional delay.  Adjusting wacom sensitivity might be enough to introduce any required delay for my case.

My goal for today with regard to this PR would be to record a session of drawing with this PR and let that speak for what I'm experiencing.  I spent a few hours last night... just random doodles and cutting/deleting.

From my experience testing the appveyor build there are at least three issues that I feel need to be addressed with this PR but @melieconiek has thrown a new spin on this if the main issue I'm experiencing isn't being seen elsewhere.    It would be good to get more people testing to ratify this PR.  The more this feature gets out for testing and usage the better.

jrhyme:
No need for a delay here. I will be brief.

beeheemooth:
I don’t have the problem that Rodney described, so I don’t see the point in delay, as this will slow down the workflow (btw, delay when switching tools just kills me).

@pojienie, as you noted, after applying this feature, that is, after removing a piece of the vector, the intersection point is no longer determined. This does not happen if the "Auto Delete" option is disabled. I wonder why this happens, and what an alternative solution can be besides disabling the option, as this will significantly complicate the use of this feature.

I noticed such a thing: determining the intersection probably depends somehow on the angle that the intersecting T-shaped vectors form. If the angle is sharp, the intersection is determined and snap works. If you increase the angle, then at some point snap stops working. Perhaps there are some limitations for snap, is my assumption true?

ezgif-4-d7a1ac290df3

RodneyBaker:
If the consensus is to merge and then troubleshoot any outliers later in follow up PRs I'm all for it.
My personal preference is to resolve conflicts prior to merger so as not to introduce a feature with problems but as no one except me is experiencing any problem I lean toward merging the PR with follow up testing and adjustment as appropriate.

In the meantime... still testing... and enjoying the feature despite the primary issue I've logged.  :)
Hopefully @manongjohn @martinvanzijl @shun-iwasawa and others can test also and provide their feedback.

pojienie:
@melieconiek Thank you for testing! Not entirely sure what you meant by "adapted to a cutter/line segment eraser tool", as this is cutter tool... could you elaborate? I don't use those tools you mentioned.

@jrhyme Thank you for testing!

@beeheemooth Thank you for testing! About your discovery, nice find! I can't tell you if the angle plays a role or not, I'll need to study more closely into the algorithm used to find intersection. And yeah, the problem may be a deal breaker to use this feature as, whenever it happens, you need to revert->disable auto delete->cut those line->delete them yourself->enable it back. All the time saved by using this feature would be wasted again.

@RodneyBaker Thank you for all the effort you've put into this! About the "double cut" problem, judging from everyone else's feedback, it might be because of your particular setup. But yeah, hopefully more people could chime in and we could see if it really is the case.

Also, you mentioned you have 3 issues that need to be addressed in this PR. What are those?

RodneyBaker:
You know the first two issues:
1) The issue of cutting a line but the line and intersection isn't recognized so nothing can process.
2) The issue of multiple lines getting deleted (i.e. difficulty in only deleting the desired line segment)

Additional issues:
3) Some care must be taken to ensure that a longer or inappropriate part of the targeted segment doesn't delete the full line which can be considerably long.

This one probably needs to be demonstrated and I alluded to a workaround above of temporarily switching off the auto-delete cutting the line in an appropriate place and then returning to delete the desired segment.  An example of this might be to draw a shape (perhaps a circle) and draw overlapping lines that intersect that shape.  Somewhere in the process of deleting/removing sections 'accidentally' target the circle/shape and note the extent of the deletion.  

The idea here is that in more complex drawings we might more easily target a section unknowingly which causes more to be deleted than we desire.  In a perfect world the removal would only proceed to the next intersection and then stop even if it's not the line section we want.

This is very likely an extension of the current feature of this PR but the underlying problem is that is can be a little too easy to select an inappropriate part of the line.  To avoid this the user can keep a hand on the keyboard read to trigger Undo to get back to the previous state and the retarget the appropriate location.

4) Occasionally the wrong side of a segment that intersects a line will be removed. 
What I think could address the issue would be some method that double checked to ensure only the removal of the shorter end of any given segment that intersects a line.  I do realize there will be cases where the longer end is the one that needs to be removed so this would need to be optional (i.e. via checkbox)

5) Similar to 4 above.  It appears that the direction of release of pen/mouse influences which side of a intersected line gets removed.  I've taken to a form of 'flicking' the pen in the direction of the part I want to remove because that intentional movement seems to increase accuracy of results considerably.  This may relate more to the precision of where the cutting gnomon is at the precise moment triggered.

6)

melieconiek:
@pojienie Yeah, I was referring to the tools as they're named in other software. Blender, CSP, Cacani, and Harmony all have a tool that erases vector strokes up to nearest two intersections. They use an intersecting selection method, can handle multiple strokes at once, and can erase a segment in the middle of a line when there are intersections in both directions.

This is what it looks like in Clips Studio Paint:
_demo

Outside the scope of this PR, but something many of us would love to see in OpenToonz. Saves an enormous amount of time during cleanup.

I don't mean to diminish your work here though--this PR alone will be a huge time saver. Thank you. :)

RodneyBaker:
In exploring what I'll call Case 1, which is when deleting a segment which leaves a dangler that isn't recognized as having an intersection, I note an interesting workflow that might hint as to how to approach a remedy programmatically.

In this image I show something of a Line Checkers approach with the progression from initial drawing to a final state where I would delete a red color swatch that is used to mark up any locations that fit Case 1.  In the second step the green circle simply identifies the section that fits Case 1.  All other line segments cut/deleted properly.  It is interesting to note that in some instances of Case 1 working the rest of the image first and then going back to the troublesome segment can allow the intersection to register and therefore the segment to be deleted properly.  This suggests to me that the bias/magnitude at the intersection might play a part in the proper registration of an intersection.

Note that in this experiment the purpose of drawing the red line at the troublesome location is to allow the cut tool to delete the Case 1 segment which deletes as expected.

Aside:  It seems to me that the additional capabilities that @melieconiek demonstrates would be good candidates for an update of the Eraser Tool where an option allows for regular erasing but additionally extends the erasure of each segment to the closest point of intersection.  This could then _theoretically_ even use the Frame Range feature to cut/delete through a sequence of images.

pojienie:
@melieconiek Gosh that is an amazing feature! Thank you for letting me know! I'll admit that, after seeing your demonstration, I want to abandon this PR altogether and try to implement that one instead... As that feature single-handedly solves all issues @RodneyBaker mentioned: (note that the order of the following list corresponds to that of @RodneyBaker 's)

1. I can make it so it identifies strokes to be deleted before actually deleting, thus making intersection algorithm work. (not completely as the root cause is still not solved)
2. no more "double cut" problem: if your pen/mouse don't go through there, it won't get deleted. And thus, no delay is necessary.
3. it only deletes the strokes until next intersection so you almost always know exactly what you're going to delete.
4. same as 2, it will not delete the wrong side of the segment if your pen doesn't go through there.
5. same as 4, no more flicking needs to be done
6. since you need to draw through the segment you want to delete, no more left/right-handness.

It will also make the cutter tool "pure", since cutter tool is used to, well, cut. It never occurred to me that auto delete feature is more of an eraser thing than a cutter thing. Why can cutter tool delete when there is eraser tool that is specifically designed to delete?

On top of all that... I don't need to look into and try to fix the boring and difficult part of the code. (I've actually looked into it... it's well beyond my ability to fix.)

@RodneyBaker Thank you for listing all the issues you have and doing so much testing! As I've said above, I feel the feature that @melieconiek shows can solve all the issues you listed at once.

Of course, I may be a bit too optimistic as there will be some problems that I fail to foresee.

melieconiek:
@pojienie No pressure of course! I meant that some of the work you've done here might be applied to a line segment eraser tool like this. It's a commonly requested feature--I believe there's a bounty on it as well, so you might want to look that up. :)

If you don't own a copy of Clip Studio Paint, the demo of Cacani is free and unlimited. It has an excellent implementation of this that you could play around with: https://cacani.sg/ (As well as some other fantastic cleanup tools.)

Blender's grease pencil also has a cutter tool, but it's a bit tempermental in my experience.

---

I'd love to see this merged in the mean time. It can be a bit finicky in some situations, but no more so than the cutter tool was already.

I can't say this enough, thank you for this and some of the other work you've been pushing recently, it's a huge help. :)

RodneyBaker:
I was excited before but now you've got me ramped up even more.  ;)
I'm all for merging this PR 'as is' pending further implementations in other PRs.
The feature right now is highly useful and a considerable plus up of the Cut Tool.

pojienie:
I'm extremely happy to say the feature that @melieconiek mentioned is done!

video: https://youtu.be/Nspvo_WSOs8

one question though: what should this functionality be named? Currently I hijacked "freehand" option to test this feature as you can see in the video.

also, sorry, super tired right now. But I'm also so excited about this that I want to share first. Will edit this comment later.

manongjohn:
Very nice! Maybe something along the lines of "Trimmer"?

RodneyBaker:
Outstanding work @pojienie!
Looking very good.

Not entirely sure about name but 'Cut Line' seems to be an appropriate term.

melieconiek:
Awesome stuff! :D

I would prefer this be added a separate tool so that you could quickly call up both the eraser and segment eraser separately. Having to use a dropdown to access it would be slow enough that I'd spend most of my time drawing with it disabled where I would otherwise be using it.

If this is to be added to the eraser tool instead of added as a new tool, I'd recommend a direct shortcut for it akin to the 'Fill Tool - Areas' and 'Fill Tool - Lines' bindings for the fill tool.

beeheemooth:
When I proposed the idea of ​​adding a snap to the Cutter Tool, I thought, “It’s unlikely that there will ever be an eraser like in CSP in OpenToonz. We need to look for an alternative using the functions that we already have.” But, @pojienie , you did it! Congratulations! Of course, this should work with the Eraser Tool, not the Cutter Tool. I do not like how this is implemented in TBH with a knife - it only deletes one segment at a time.
By the way, about the name: in OT there is already a similar function in the Fill Tool (Line mode, TR level), the option is called "Segment". If your feature will be in the form of an option for the Eraser Tool, then it can be called similarly.

 Will this thing work with Frame Range?

jrhyme:
You are a great fellow! Everyone has been waiting for such a thing for a very long time) but I would like to inquire) can it be implemented also in the toonz raster?

RodneyBaker:
Something of an aside related to this PR.
I do hope it can still be implemented as it is still quite useful and plusses up the Cut Tool.
I will defer to you @pojienie on whether or not you want it to be merged.

KaseyFarron:
Wow, it's so good!

In Clip the similar mode for this is called "[Erase up to intersection](https://www.clipstudio.net/en/promotion/trial_tour/004/)" just FYI if it helps any for names.

artisteacher:
What about "Segment Erase"?

pojienie:
(I guess I am not editing that comment)

Thank you all for kind words and name suggestion! I've created a new PR: #3194

for its name, I'll go with "Segment" because @beeheemooth mentioned that it's already been used in Fill Tool.

@melieconiek There is already an eraser tool so I don't think adding a new tool of similar functionality is a good idea. I do agree adding a shortcut that allows you to switch to the modes instantly instead of selecting it in drop-down menu. Maybe future PR, though.

@beeheemooth currently it doesn't work with frame range (not even invert) yet.

@jrhyme it doesn't work in raster level.

@RodneyBaker If I am to decide, I'd suggest not to merge this PR. My reason is that cutter tool should not be able to remove. But I'm not the target users of opentoonz so my reason here may be invalid in a "efficiency" based point of view.

RodneyBaker:
Closing this so it doesn't accidently get merged.
(which I wouldn't mind happening)  ;)

RodneyBaker:
@pojienie There has been additional interest in the Cut Tool version of the Segment Erase feature also being implemented.  I tend to agree on several levels to include:

1) The Cut Tool version of the Feature has a slightly different feature set that is complimentary to the Erase Tool version

2) The Cut Tool gains important functionality through the feature.

Perhaps something to consider...

pojienie:
@RodneyBaker Just to be sure, by "the feature" you meant this PR, right? 

Personally I don't agree with the philosophy that a knife can be used to erase something... But yeah, currently, cutter tool really is quite feature-less.

I've restored the branch. I think it is possible to reopen this PR then merge this. If not, I'll submit another PR with the same branch. You're the maintainer of Opentoonz so the decision and responsibility is on you :)

RodneyBaker:
Yes, this feature is the one under consideration.
As for the philosophy... I follow your logic and agree with the general premise.

Think of it like cutting a rope where the untethered part of the rope falls away and disappears after being cut... or a vine.  Yes, technically the part is not disappearing but the purpose is to cut away and discards a part of the object.  Think someone that is sewing some clothing and cuts away with scissors or knife that part they don't need and throws it away in the trash can.

A specific part of the feature as you've implemented it in the Cut Tool is the ability to sever the line at any given point and have the remainder of that portion fall away.

Shun Iwasawa is the maintainer of Opentoonz... don't get me in trouble here.  ;)
And hey, if you want to give me credit for your excellent work... I'm all for it!  (kidding!)

pojienie:
Oops, sorry! You're the one who most actively monitoring all the issues and PRs here so I always thought you're the maintainer :)

I understand what you mean now. Never thought about those metaphors so thank you for making sense of this PR for me!

jrhyme:
All this is of course excellent) but it is a pity that these PRs will not work in the Toonz raster. He is more popular in this program, since the Ghibli studio uses it.

manongjohn:
My 2 cents:

Cutter Tool was designed to split strokes, not remove portions of it. Adding this functionality confuses the function of the tool. So I do not agree this should be added to the Cutter Tool.

The logic was added to the Erase tool and it makes sense there because it is a tool for removing unwanted things, in this case a segment.

For the purposes of the bounty, I think the bounty has been fulfilled and awarded. It may not be on the Cutter Tool, but it is in the Eraser Tool where it makes sense.

RodneyBaker:
@manongjohn This is a good opportunity to discuss what should be done in cases like this where some think a feature should be added and other don't.  There is no doubt to me that the feature would be useful so I of course lean toward inclusion.  I also of course want to proceed in a way that is best for Opentoonz.

I base my interest in seeing the feature included on a number of things, a few of which I have mentioned above as well as the request of others who I have asked to post here so their interest can be known.  My own interest is motivated primarily by usage... I've used the feature via Cut Tool and found it to be very useful.

@pojienie has expressed a similar concern about the Cut Tool's purpose and the paradigm of it's current use.  Perhaps it would be good to better understand what form that paradigm takes regarding the functionality of the Cut Tool as it moves forward in development in the future.  We don't have to imagine much for the proposed paradigm of included the option... as it can be observed right now.  Perhaps for the Cut Tool implementation... as requested elsewhere... this particular implementation would do something other than remove the target segment; assign it to a specific color style where it could be further modified... even removed if desired... by the user.

I ask all this because 2 cents coming from developers  carry a lot of weight... more like 50 bucks... and so must be carefully considered.

beeheemooth:
Let's weigh the pros and cons:
1. Cutter Tool deletes only one vector, and the Eraser Tool - two (and more). Thus, the Cutter Tool cannot solve the problem of the T-intersection, and the Eraser Tool can.
2. Removing a small owerflow with the Eraser Tool is sometimes quite problematic, the Cutter Tool (as it seemed to me) does this better.
3. It is unlikely that the Cutter Tool will have a Frame Range mode.

turtletooth:
I'm with those who don't think it fits here.  I have played with the eraser tool and it works fantastic!  Thanks @pojienie 

I don't mind trying it out first though before I give it a hard no on my side.

@jrhyme It's probably never going to happen for Toonz Raster since vector levels are a totally different beast than raster.  Sorry.

beeheemooth:
In the TR, the Fill Tool can paint over segments, why not delete them in the same way?
ezgif-2-3cbd6ad9e577

turtletooth:
@beeheemooth Good point.

jrhyme:
I’ve talked about this, therefore nothing is impossible.

RodneyBaker:
Closing as this PR is not likely to move forward.

PR #3181, opened 2020-03-26
Update macOS build steps

artisteacher:
This adds a step to remove an incompatible directory, updates the name to macOS and lists the last tested boost version as 1_72_0.

### DIFF
--- a/doc/how_to_build_macosx.md
+++ b/doc/how_to_build_macosx.md
@@ -1,5 +1,5 @@
 
-# Setting Up the Development Environment on MacOSX
+# Setting Up the Development Environment on macOS
 
 ## Necessary Software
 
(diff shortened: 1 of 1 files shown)


MrBlockCat:
Awesome, thank you! Do you think you could change the Homebrew execution statement to the new one on their website, `/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"`? I'm still new to all this, but my impression is that's outdated too

RodneyBaker:
Anything else needed for the MacOS build steps prior to merging this?

MrBlockCat:
Nothing I'm aware of! In steps with the QT version the version's outdated, but as far as I'm aware it's the new QT version that's causing the grey double arrow issue so it's probably best to keep it outdated for now

PR #3182, opened 2020-03-26
Modify "Save Images" Command

shun-iwasawa:
This PR modifies the "Save Images" command, which is in right mouse button menu of the flipbook panel.

- Fixed an initial folder path of the browser popup to be decoded ( i.e. convert alias like "+outputs" to regular file path ) so that the browser opens properly.
- Moved the command from under the "Load / Append Images" command to the bottommost of the menu in order to avoid being clicked by mistake when users would like to load something.


RodneyBaker:
Thank you Shun!
LGTM

PR #3183, opened 2020-03-26
Update more macOS Build Instructions

MrBlockCat:
Updates the rest of boost from 1_69_0 to 1_72_0 and tweaks some other more minor things.

Sorry! I thought I'd caught everything from the recent PR but I noticed there were still a couple outdated commands, so this should fix that, and some minor typos as a bonus!

### DIFF
--- a/doc/how_to_build_macosx.md
+++ b/doc/how_to_build_macosx.md
@@ -63,8 +63,8 @@ $ git clone https://github.com/opentoonz/opentoonz
 $ cd opentoonz
 $ git lfs pull
 $ cd thirdparty/boost
-$ mv ~/Downloads/boost_1_69_0.tar.bz2 .  #or whatever the boost filename you downloaded is#
-$ tar xvjf boost_1_69_0.tar.bz2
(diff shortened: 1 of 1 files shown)


manongjohn:
The build instructions should show the minimal requirements to build. Boost 1_69 is still valid so i would not update that.

MrBlockCat:
I agree it should show the minimum requirements to build (and I kept the minimum required version for boost the same) but as boost was recently changed to 1_72_0 elsewhere in this doc and that’s the version you have by following the directions it seems to make sense to update these commands to follow what does happen when you download boost, especially as following the build instructions involves redownloading boost every time you build, meaning you’d be working with their most updated version

manongjohn:
Then we'd be updating these instructions every time they released a new version.

I think the only reason I even had it there was because that is what I built with at the time I drafted these instructions as an example.

Not a big deal if its updated.

artisteacher:
The last tested version (which I updated) doesn't necessarily need to match the version number in the build steps themselves. However it'll temporarily save us a step, so I don't mind the update.

Is the official release still built with QT 5.9.2? It would be nice to update the cmake file at some point.

PR #3184, opened 2020-03-27
Correct library path in Linux installation doc

martinvanzijl:
I wasn't seeing any MyPaint brushes in OpenToonz. When I changed the TOONZLIBRARY path from "$HOME/.config/OpenToonz/stuff/projects/library" to "$HOME/.config/OpenToonz/stuff/library" in my config file ($HOME/.config/OpenToonz/SystemVar.ini), I could see the MyPaint brushes, and a lot more vector brushes, etc.

It looks like the path was wrong in the installation doc, so this PR is to correct it.


RodneyBaker:
This will make a lot of Linux users happy as most think Mypaint brushes cannot be used.

martinvanzijl:
@RodneyBaker Thanks, although this fix only applies to installing from source. There may still be issues with the pre-packaged Linux versions (.deb files and so on). Do you happen to know where those are managed?

RodneyBaker:
I would be guessing.
Perhaps @blackwarthog can assist in this as he surely knows.

blackwarthog:
fix me if i'm not right
as i know we does not provide .deb or .rmp packages
and have no scrips or instruction to make it

RodneyBaker:
Thanks for the confirmation Ivan.
It was worth a shot.  :)

PR #3187, opened 2020-03-29
Fix typos in the Preferences menu

### LINKED ISSUE #3136: (Proposal) Rename Preferences > Interface > Icon Size as "Level Strip Icon Size"
Icon Size, from an outside perspective, sounds like all the icons in OpenToonz should change with it. However, what it changes is the icons in the Level Strip. To make this clearer to beginners, the "Icon Size" setting should probably be renamed to "Level Strip Icon Size"

### LINKED ISSUE #3163: Viewer Shrink Setting Typo (one extra space)
The setting, in Preferences > Interface, has two spaces between Viewer and Shrink rather than one.

MrBlockCat:
Fixes #3163, fixes #3136

The Korean translation issue no longer occurs for me after reinstalling everything on my Mac, meaning there was some kind of underlying issue.


MrBlockCat:
After again testing every language in the Preferences menu compared to nightly, the only thing that doesn't translate but did is still with the Korean translation in [Preferences > General's Undo Memory Size (MB):](https://github.com/MrBlockCat/opentoonz/blob/845bc731f0b31109cc307a1174f4b63c9f067d22/toonz/sources/translations/korean/toonz.ts#L7065-L7068)

MrBlockCat:
I'd like to change Sub-xsheet or sub-xsheet (those are the most common capitalizations at least) to Sub-Xsheet or sub-Xsheet, to match with Xsheet's capitalization. Are there any objections to that?

RodneyBaker:
I kind of prefer Sub-xsheet but for no specific reason.  (maybe because it conserves space  ;)? )
The documentation does appear to standardize on Sub-Xsheet with the additional capitalization.
I would think as long as it's standardized that would be the ideal.

MrBlockCat:
I can go back in a later PR and make this more consistent across the UI, but that should take care of it in the Preferences!

MrBlockCat:
It may be worth noting that where it was translated, Level Strip Icon Size still translates to Icon Size, or at least whatever it translated to when it was that. I'm trying to minimally mess with translations, so that seemed like the best idea to do that. Please let me know if I should approach it another way! Otherwise, the only issue remaining is that single Korean translation.

RodneyBaker:
Most translations updates can be done separately and by those who know the language well.
It wouldn't be good to make a change in another language that actually made the reading worse.

MrBlockCat:
Great! That makes the only remaining issue that Korean translation, I believe. I may do some testing to try and track down what causes it, though I'm tempted to wait for the GitHub Actions to be put in before testing lots of tweaks for that as it would save me a ton of time and it seems very close to being put in!

RodneyBaker:
@MrBlockCat Did you say you wanted to press forward with this PR and leave the Korean issue as was already?

MrBlockCat:
The Korean translation problem confuses me, since it works outside this PR but doesn’t in this PR. I suspect an underlying issue, but as my MacBook is dead for now I don’t have a good setup to test/change this. It’s probably best to wait a few days until I can, and I’ll see if I can somehow figure out the cause of the issue. That said, if I can’t fix the issue this still fixes way more translations than the one it breaks so it makes sense to me to merge this regardless of the Korean issue; If that happens it’s then up to you!

RodneyBaker:
We shouldn't merge this PR if we know it introduces a new bug.
We'll hold off for now.

MrBlockCat:
My MacBook is back to a functioning state, but the hard drive needed to be replaced as part of fixing it. Because of that, I've had to reinstall everything. For some reason, now I no longer am able to replicate the Korean translation issue, it works as it should now! I don't understand why only that one translation was impacted, but I think I was right that it was some kind of underlying issue. I think this PR is ready to be merged!

RodneyBaker:
Thanks @MrBlockCat 
Will proceed with this fix.

PR #3188, opened 2020-03-29
Enable Ink #1 Check for vector levels in the viewer

martinvanzijl:
When I enabled "View --> Ink#1 Check" on a vector level, it would not highlight the ink for Style number 1 in the viewer. (It would do so in the thumbnail, though, if the thumbnail was refreshed by e.g. drawing another line.) 

This PR makes the Ink#1 Check work in the viewer for vector levels. Demo below:

ink1-check-demo


RodneyBaker:
Interesting.  Apparently no one has reported that issue before.
Nice find.

I confirm the problem and that your PR fixes that.
Thank you Martin.

PR #3192, opened 2020-03-29
Force all thumbnail icons to refresh when toggling Paint Check

### LINKED ISSUE #1122: Filling a shape with paint check on results in wrong color in thumbnail.
paint_check

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/43379551-filling-a-shape-with-paint-check-on-results-in-wrong-color-in-thumbnail?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

martinvanzijl:
Fixes #1122. All thumbnail icons in the film strip are now refreshed when toggling Paint Check. They are also refreshed when toggling the other checks in the view menu (Ink Check, Black Bkg Check, etc.)

Demo below before the fix:

thumbnail-demo-before

After the fix:

thumbnail-demo-after

I recommend testing with a large scene with a lot of frames to see that this doesn't cause any slowdown. It seems to go quickly for me.

### DIFF
--- a/toonz/sources/toonz/sceneviewer.cpp
+++ b/toonz/sources/toonz/sceneviewer.cpp
@@ -297,6 +297,16 @@ void invalidateIcons() {
   s.m_paintIndex = mask & ToonzCheck::ePaint ? tc->getColorIndex() : -1;
   IconGenerator::instance()->setSettings(s);
 
+   // Force icons to refresh
+  TXshLevel *sl = TApp::instance()->getCurrentLevel()->getLevel();
+  if (sl) {
+    std::vector<TFrameId> fids;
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thanks Martin
Tested with large projects including an extended scene with mostly auto-inbetweened vector shapes.  All with no noticeable slow down.
LGTM

Issue #2861, opened 2019-10-27
Vectors:  Stroke/brush origin extra control points


RodneyBaker:
### Summary
All vector brushes appear to be especially vulnerable to distortion/end looping when Smoothing is set to greater than 33

![LeadingEdgeBrushError](https://user-images.githubusercontent.com/2576432/67628815-9e0d3200-f839-11e9-9a11-fffdb2f91428.png)
Note:  Strokes are being drawn from right to left.
The bump near the right is part of the vector brush.

### Steps to Reproduce
Create a vector level 
Select a vector brush from Style Editor (some will more easily display the leading edge of the stroke than others)
Draw vector lines
Note that the leading edge of the stroke will be distorted/looped or hooked over.
(Use the Iron Tool) to smooth it out)
Adjust Smoothness above and below the value of 33 and note the difference.

### Expected Results
Ends of vector line should run smoothing but directly into the control point.

### Actual Results
Leading edge of vector line hooks/loops over deforming stroke

--- 
Status
---
- [x] Known Issue
- [x] Documented Workaround(s)
- [ ] Documented Code Location(s)
- [ ] Priority Established
- [ ] Bounty Attached
- [ ] Developer Assigned/Attached

zarco1794:
If a stroke is done quickly, smoothing has no effect. I don't know if this has to do with the issue raised here
![Animated GIF-source](https://user-images.githubusercontent.com/39812147/67642712-5a740000-f8ed-11e9-8445-3ca73ee4f0aa.gif)

RodneyBaker:
@zarco1794  In your example it looks to me like you have Smoothing set to zero for both slow and fast drawing.  It is normal additional smoothing calculations are not triggered when Smoothness is set to zero.  Drawing a faster stroke does give us some generic smoothness (as seen in your example though so that is a good way to draw!)  Those experiencing brush lag are encouraged to turn Smoothness down to zero (if they can live with the results of course).  Shun Iwasawa described in another issue/topic more about the specifics of the calculations and their impact.

RodneyBaker:
I was noticing this again today with regular vector brush while using touchpad on my laptop.
The result of the stroke being a concentrated and overlapped group of approx. 5 control points at the beginning of the stroke all almost hidden because they are overlaid over each other.

I note that running 'Replace Vectors with Simplified Vectors' resolves the problem (leaving only two control points in most cases).

[...]

Issue #2526, opened 2019-02-18
Crash with Remove Vector Overflow (gif).


beeheemooth:
### Screenshot or Video Reference

![ezgif-1-d528686fa912](https://user-images.githubusercontent.com/25724133/52955226-d009f280-3394-11e9-8098-dc7b580dbd5f.gif)

### Steps to Reproduce

"Do it like Simon says..."

1. Create new Vector level.
2. Make sure the Brush Smooth option is similar to mine (it seems that it plays some role in reproducing crash, I think).
3. Draw letter "H" in order like i do.
4. Use Eraser Tool Type: Normal to erase the middle parts. (Turn on Fill Check and Gap Check)
5. Select all lines and choose Ramove Vector Overflow
6. Undo this (think: "damn, I forgot to erase the top and bottom!")
7. Redo this (think: "nevermind.. I'll erase it manually")
8. Try to erase or edit with Control Point Editor.
9. Crash.

### System Information
* **OpenToonz Version**:  v1.3
* **Operating System**:  Windows 8.1

beeheemooth:
It is possible that one of the reasons for the crash could be this:
![ezgif-1-c9a784fb4a6e](https://user-images.githubusercontent.com/25724133/52965614-e8d2d200-33ad-11e9-97aa-783596dec557.gif)

manongjohn:
Following your exact steps from the description and even the video, I am unable to reproduce this crash or that strange extra loop when drawing.

manongjohn:
After raising smooth to over 30 (I used 45+) at @beeheemooth 's suggestion and drawing lines quickly, I was able to reproduce the extra loop as well as the crash.

Seems related to bad intersection data.  Looking further into it.

RodneyBaker:
Somewhat related:
I note that the extra control points (extra loop on the end of a spline/path) tends to get created when using the Magnet Tool in some situations.  Perhaps some form of 'IsFlat' query can be used to reduce the number of control points for lines that approach being less than a specific increase.  For instance, a straight line should only require two control points because a recursive 'IsFlat' query would result in 'true' for all segments, thereby iterating through a process of control point reduction until arriving at two control points (for the straight line case).

RodneyBaker:
This should hopefully be resolved.
@beeheemooth can you confirm?
Thanks

beeheemooth:
Confirmed.

PR #3194, opened 2020-03-30
add segment mode in eraser tool

pojienie:
This PR allows you to remove segments when using eraser tool

video: https://youtu.be/Nspvo_WSOs8
related PR: #3178

name choice: since @beeheemooth mentioned that "segment" is already used in fill tool

Algorithm:

1. find all strokes intersect with eraser and where they intersect
2. find closest intersection for each intersection with eraser to form a range
3. merge all ranges of the same stroke by using interval merging algorithm
4. make complement range
5. actually do the erasing and construct undo data

It doesn't support invert and frame range yet. It doesn't support raster level. Since freehand mode is already in place, I used its code to do everything except actually erasure.

note that the problem where it wouldn't be able to detect intersection like this still exist. Because the root cause wasn't fixed yet.
 a

### DIFF
--- a/toonz/sources/tnztools/vectorerasertool.cpp
+++ b/toonz/sources/tnztools/vectorerasertool.cpp
@@ -54,6 +54,7 @@ namespace {
 #define RECT_ERASE L"Rectangular"
 #define FREEHAND_ERASE L"Freehand"
 #define POLYLINE_ERASE L"Polyline"
+#define SEGMENT_ERASE L"Segment"
 
 //-----------------------------------------------------------------------------
 
@@ -341,6 +342,8 @@ private:
 
   void eraseRegion(const TVectorImageP vi, TStroke *stroke);
 
+  bool eraseSegments(const TVectorImageP vi, TStroke *eraseStroke);
+
   void multiEraseRect(TFrameId firstFrameId, TFrameId lastFrameId,
                       TRectD firstRect, TRectD lastRect, bool invert);
   void doMultiErase(TFrameId &firstFrameId, TFrameId &lastFrameId,
@@ -379,6 +382,7 @@ EraserTool::EraserTool()
   m_eraseType.addValue(RECT_ERASE);
   m_eraseType.addValue(FREEHAND_ERASE);
   m_eraseType.addValue(POLYLINE_ERASE);
+  m_eraseType.addValue(SEGMENT_ERASE);
   m_prop.bind(m_selective);
   m_prop.bind(m_invertOption);
   m_prop.bind(m_multi);
@@ -409,6 +413,7 @@ void EraserTool::updateTranslation() {
   m_eraseType.setItemUIName(RECT_ERASE, tr("Rectangular"));
   m_eraseType.setItemUIName(FREEHAND_ERASE, tr("Freehand"));
   m_eraseType.setItemUIName(POLYLINE_ERASE, tr("Polyline"));
+  m_eraseType.setItemUIName(SEGMENT_ERASE, tr("Segment"));
 }
 
 //-----------------------------------------------------------------------------
@@ -439,7 +444,8 @@ void EraserTool::draw() {
       tglDrawCircle(m_brushPos, m_pointSize);
     }
     if ((m_eraseType.getValue() == FREEHAND_ERASE ||
-         m_eraseType.getValue() == POLYLINE_ERASE) &&
+         m_eraseType.getValue() == POLYLINE_ERASE ||
+         m_eraseType.getValue() == SEGMENT_ERASE) &&
         m_multi.getValue() && m_firstStroke) {
       TPixel color = blackBg ? TPixel32::White : TPixel32::Red;
       tglColor(color);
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Looking good thus far.

segment erase

I confirm our old enemy the old Case 1 non-interesection registering segment.
That certainly is a pain.
For a workaround I'm currently using the Cut Tool to go in and quickly dissect the line at that intersection so that segment erasing will remove the troublesome segment.

I'm running into an issue where a small section of the segment that is erased remains.
My current theory is that it is the little hook that tends to appear at the end of a vector line in many cases. Reference #2526 and #2861 for more information on that.

Here's the PLI file if you want to use that.
[horse.zip](https://github.com/opentoonz/opentoonz/files/4401344/horse.zip)

RodneyBaker:
Experiencing a lot less problems as I get use to the workflow...

horse

pojienie:

About the "case 1" problem: As I said it in the PR, I didn't solve the root cause (actually, I couldn't). But this PR should mitigate such problem as you can now remove two strokes at once, like so:

test 0001

The red line indicates where I use the eraser. In the top row, if you remove strokes one by one, you'll encounter the problem. However, if you remove two strokes at once as in the bottom row, it will correctly detect the intersection and remove the strokes you want to remove.

About the other problem: Your theory is probably right since those "hooks" also appeared in your horse.pil file. I'm sure this issue should be solved when the line is drawn instead of in this PR.

RodneyBaker:
Posting my intent to merge this as is.
Open call for testing!  :)

Here's a link to the appveyor build (for Windows) for those so inclined that do not build locally:

https://ci.appveyor.com/api/buildjobs/jtva65vp3k2n1eps/artifacts/toonz%2Fx64%2FOpenToonz.zip

You'll want to rename the Opentoonz stuff directory as 'portablestuff' and drag/drop it into the Opentoonz directory.  That will ensure you are getting everything built from this PR.

jrhyme:
Maybe Shun will come and combine it into a night build?) Or I misunderstood ..

RodneyBaker:
It can be put into a nightly build right now but... testing... testing first.  ;)

jrhyme:
Nightly builds will be updated and this function hangs in the list like the rest) as you wish) In any case, we will test everything, otherwise there will be no version 1.5.

RodneyBaker:
Time to get this into the hands of more users.

beeheemooth:
@pojienie Great job! Thanks!
Users ask if the Eraser Tool can be switched to Segment mode while holding down the Ctrl or Alt button. If not, will there be a hotkey for the "Segment" type?

jrhyme:
I think everything will be with time)

pojienie:
@RodneyBaker Thank you for testing and merging!

@beeheemooth Thank you! I've thought about adding the shortcut but it will take me some time to study the shortcut system.

gab3d:
@pojienie 
I know I'm late, but just didn't want to miss the opportunity to say a big thank you for this feature.
It's amazingly good, and much certainly one of the top requested features by users since long ago.
So, kudos to you! 🥇 
It's so refreshing to have you onboard!

Regarding @beeheemooth suggested "Ctrl" feature, it'd certainly be a great way of using it on the fly.

pojienie:
@gab3d Thanks!

I have added all shortcuts in #3213 so hard-coded control key is not needed now.

gab3d:
@pojienie 
thanks, I've just seen that PR and surely it adds more workflow possibilities than just the Ctrl variant suggested, and also for a lot of other tools.
you've done a great work! :)

one question I'd like to make is (although perhaps I didn't fully understand the workflow of using the new shortcuts available), as far as I understood if, for example, you use the assigned "Eraser - Segment" shortcut to erase some stroke segments; then you use the Brush tool shortcut (B) to draw some more strokes; and finally you use the (E) shortcut to be able to manually erase some parts of a stroke. 
What would be the state of the Segment option when you activate the Eraser tool through the E shortcut in the last step?

if the Segment option is to remain active, this could slower the workflow (because the user would need to deselect the option, before being able to use the standard erasing mode). 🤔

pojienie:
@gab3d Yes, after 3rd step (pressing E to switch to eraser tool) the eraser will be in segment mode. But you can always use other shortcuts to change that (I added shortcuts to all modes so you can always, say, use "Eraser - Normal" shortcut to switch back to normal mode)

I am against the temporary switch (holding control and it is segment mode, otherwise it's the previous mode) because of two reasons:

1. I don't want to hard-code it
2. why does only segment mode have this special treatment? And what if other users think it's more intuitive to have rectangular mode when holding control? How about users who wants "freehand" mode when holding control? Whose opinion is more important?

The above, of course, is just my opinion. But I'm the developer of this PR so I get to call the shots :stuck_out_tongue: That said, if other developers want to do it, I won't intervene :wink:

gab3d:
@pojienie 
Thank you for the explanation.
Yes, undoubtedly from a functionality standpoint the possibilities are all there, and are almost endless for a lot of tools, and not just Eraser, and that's certainly great. :)

Having said that, and viewing it now from a pure workflow (and also from a new user's) perspective though, I think the solution offered right now is a little cumbersome.
Basically, two main things to consider:
- Standardization. The new "Eraser - Segment" function has no default shortcut assigned.
- Simplicity. I predict the standard shortcut for Eraser will be "Eraser - Segment", so 'E' shortcut would be effectively superfluous, as it offers less functionality (it activates the tool, but it doesn't change the Segment option).

So I think that either we should make the 'E' shortcut control de "Eraser - Segment" function or preserve it like it is right now, but enable a modifier key (such as Ctrl) to temporarily activate the Segment option to have a strong performance while using this awesome new feature.

I would like to know what others think about it as well...

RodneyBaker:
Having 'E' be the default for Erase - Segment wouldn't work for the default level type; toonz raster.

RodneyBaker:
Just a follow up note to announce that segment erase for raster levels (smart raster) can now be tested via Tahoma appveyor artifact:

https://ci.appveyor.com/project/turtletooth/tahoma2d/builds/36395193/artifacts

PR #3195, opened 2020-03-30
Improve Browser up button

martinvanzijl:
When using the Load Scene dialog (or other dialogs), the "Up" button sometimes would not go to the parent directory. It seems to only go up if the parent folder exists in the tree-view on the left.

This PR makes it always go to the parent directory if possible.

Demo before the fix below. Note how it gets "stuck" at the "sandbox" folder and does not go up any further:

browser-up-button-before

Demo after:

browser-up-demo


RodneyBaker:
I confirm the issue and that this PR fixes it.
Thanks Martin!

Issue #1554, opened 2017-11-04
(WIP) Hide line tool Request


openanim:
<details>Hi folks,

Currently to hide specifics parts of drawing on OT we have to cut the parts, then use 0 thickness or 0 alpha. But this process is long and break the tweening process. 

I would like in the future to see a feature to hide the strokes in opentoonz, something intuitive and above all, very fast without damaging the existing tweening process.

I think we should think about the best integration for Opentoonz with its current tools, all ideas will be welcome, there is already a discussion link on the forum https://groups.google.com/forum/#!category-topic/opentoonz_en/feature-requests/dsME50E4t8g  and https://groups.google.com/forum/#!category-topic/opentoonz_en/feature-requests/G_AgxoI37QY  and here  https://github.com/opentoonz/opentoonz/issues/746 on Github.

I would like to submit a bountysource because it is a very useful feature in many situations without having to use masks, cutter or erasers : Bodyturns, headturns, parts of body, cycles, cut-out etc ...

Personally, I like 2 softwares approaches. Moho and Cacani. I much prefer the cacani approach because it allows maximum flexibility and customization that also includes the moho approach, but the simple moho approach will also save significant time anyway compared to the workflow currently available in OT. 

Here is my proposal on this topic:

Creation / integration of a "hide tool" with the properties of some tools currently available in OT, essentially the selection tool.

- On the one hand :
we could choose the way we want to hide our stroke as Type : endpoint to endpoint (segment) / Freehand / polyline/ Freely , the last type should allows maximum flexibility , in this way we can choose that we want to hide easly. Cf. See video and picture below.

![hide tool request](https://user-images.githubusercontent.com/29790512/32406322-b67b2960-c176-11e7-8e1d-e4b2997099a6.gif)

![hide tool](https://user-images.githubusercontent.com/29790512/32406335-dc099874-c176-11e7-85a2-4865bf87f2fa.png)

- And on the other hand :
the mode of application found in the selection tool to be able to quickly affect specific strokes as shown above.

Video above it's how Cacani works on this :
2 segments tools  one to use by segment and the other freely on the parts of the drawing to hide. 
That said, there are others approachs about this feature request, the main thing would be to provide Opentoonz with an intuitive tool in the future to easily hide lines without breaking them.

@ developers : Currently Bountysource don't works with github account, I will submit that once it will be repaired.

</details>

Updated this request. 

[...]

Issue #746, opened 2016-08-20
Suggesting a hide line tool for toon vector level


arh2009:

RodneyBaker:
Can you provide a little more information regarding how this tool would work and what it would be designed to do?

Currently, vVector lines can be hidden via turning the transparency from 255 to 0 on the M channel (RGBM) in the same fashion as turning use of an Alpha Channel.  This applies to any Palette Style used with either lines or areas.

arh2009:
sure, just like anime studio tool...
i think a tool for that will be save a lot of time 
![hide edge animestudio1](https://cloud.githubusercontent.com/assets/20434396/17837246/babbf40e-67d7-11e6-8bf5-8fb6b3a6df33.gif)

arh2009:
it can be hide a part or some of edges or line segments, without cutter the line segment at the first

RodneyBaker:
Here's something of a workaround... but I think you already know that method.


![colorlines](https://cloud.githubusercontent.com/assets/2576432/17837854/1dfe8558-6784-11e6-8bc2-45f2635101e8.gif)

arh2009:
@RodneyBaker : is it a connected line or just separate lines in a group ? , as i know we can not do that if all the lines was connected.
I wish opentoonz has hide line tool for a single line connected with many segments...

arh2009:
![sample2](https://cloud.githubusercontent.com/assets/20434396/17840428/5beae606-6831-11e6-8a76-a7a261e26e9b.png)

animking:
yes, it's needed.

blurymind:
I dont understand what's wrong with the approach @RodneyBaker  suggested.
It seems to be exactly the feature that you are requesting.
It already is a part of OT - you just dont know how to do it and are used to how anime studio does it.

Imo opentoonz has a better design - as it lets you set any color to the line. If you want to make the line disappear, simply set it to a transparent style

animking:
I guess you didn't use anime studio.

blurymind:
I used it when it was called moho

blurymind:
I guess you want to be able to hide a part of a line, without having to cut it

RodneyBaker:
@blurymind   I see the issue they are running into.  It would certainly be better if we could isolate a part of a line.

If you'll pardon the long post, I believe the solution to this is going to be 'Named Groups'.
[...]

PR #3196, opened 2020-03-31
implement frame range for segment mode of eraser tool

pojienie:
This PR implements the "frame range" option of segment mode of eraser tool.

video: https://youtu.be/ig9KVzgpch8

Code wise, it's surprisingly simple, this shares 99% of the code with freehand mode so this feature should behave exactly the same as freehand mode in terms of tweening.

But since segment mode erases all segments touched by the eraser, slight inaccuracy might cause segments that you want to keep be removed or vice versa. In simple cases, this will work well. However, problems might arise when dealing with complex drawing.

### DIFF
--- a/toonz/sources/tnztools/vectorerasertool.cpp
+++ b/toonz/sources/tnztools/vectorerasertool.cpp
@@ -288,6 +288,9 @@ public:
   void onDeactivate() override;
 
 private:
+  typedef void (EraserTool::*EraseFunction)(const TVectorImageP vi,
+                                            TStroke *stroke);
+
   TPropertyGroup m_prop;
 
   TEnumProperty m_eraseType;
@@ -342,15 +345,18 @@ private:
 
   void eraseRegion(const TVectorImageP vi, TStroke *stroke);
 
-  bool eraseSegments(const TVectorImageP vi, TStroke *eraseStroke);
+  void eraseSegments(const TVectorImageP vi, TStroke *eraseStroke);
 
   void multiEraseRect(TFrameId firstFrameId, TFrameId lastFrameId,
                       TRectD firstRect, TRectD lastRect, bool invert);
   void doMultiErase(TFrameId &firstFrameId, TFrameId &lastFrameId,
-                    const TStroke *firstStroke, const TStroke *lastStroke);
+                    const TStroke *firstStroke, const TStroke *lastStroke,
+                    EraseFunction eraseFunction);
   void doErase(double t, const TXshSimpleLevelP &sl, const TFrameId &fid,
-               const TVectorImageP &firstImage, const TVectorImageP &lastImage);
-  void multiEreserRegion(TStroke *stroke, const TMouseEvent &e);
+               const TVectorImageP &firstImage, const TVectorImageP &lastImage,
+               EraseFunction eraseFunction);
+  void multiErase(TStroke *stroke, const TMouseEvent &e,
+                  EraseFunction eraseFunction);
 
 } eraserTool;
 
@@ -991,7 +997,7 @@ void EraserTool::leftButtonUp(const TPointD &pos, const TMouseEvent &e) {
   } else if (m_eraseType.getValue() == FREEHAND_ERASE) {
     closeFreehand(pos);
     if (m_multi.getValue()) {
-      multiEreserRegion(m_stroke, e);
+      multiErase(m_stroke, e, &EraserTool::eraseRegion);
       invalidate();
     } else {
       eraseRegion(vi, m_stroke);
@@ -1003,14 +1009,15 @@ void EraserTool::leftButtonUp(const TPointD &pos, const TMouseEvent &e) {
     double error = (30.0 / 11) * sqrt(getPixelSize() * getPixelSize());
     m_stroke     = m_track.makeStroke(error);
     m_stroke->setStyle(1);
-    bool erasedSomething = eraseSegments(vi, m_stroke);
-    if (erasedSomething) {
+    if (m_multi.getValue()) {
+      multiErase(m_stroke, e, &EraserTool::eraseSegments);
+      invalidate();
+    } else {
+      eraseSegments(vi, m_stroke);
       invalidate();
       notifyImageChanged();
     }
     m_track.clear();
-    delete m_stroke;
-    m_stroke = NULL;
   }
 }
 
@@ -1038,7 +1045,7 @@ void EraserTool::leftButtonDoubleClick(const TPointD &pos,
     TStroke *stroke = new TStroke(strokePoints);
     assert(stroke->getPoint(0) == stroke->getPoint(1));
     if (m_multi.getValue())
-      multiEreserRegion(stroke, e);
+      multiErase(stroke, e, &EraserTool::eraseRegion);
     else {
       eraseRegion(vi, stroke);
       m_active = false;
@@ -1218,8 +1225,8 @@ static bool doublePairCompare(DoublePair p1, DoublePair p2) {
   return p1.first < p2.first;
 }
 
-bool EraserTool::eraseSegments(const TVectorImageP vi, TStroke *eraseStroke) {
-  if (!vi || !eraseStroke) return false;
+void EraserTool::eraseSegments(const TVectorImageP vi, TStroke *eraseStroke) {
+  if (!vi || !eraseStroke) return;
 
   int strokeNumber = vi->getStrokeCount();
   std::vector<int> touchedStrokeIndex;
@@ -1248,7 +1255,7 @@ bool EraserTool::eraseSegments(const TVectorImageP vi, TStroke *eraseStroke) {
 
   // if the eraser did not touch any strokes, return
   if (touchedStrokeIndex.size() == 0) {
-    return false;
+    return;
   }
 
   // find closest intersections of each end of the touched place of each stroke
@@ -1406,7 +1413,7 @@ bool EraserTool::eraseSegments(const TVectorImageP vi, TStroke *eraseStroke) {
   }
 
   TUndoManager::manager()->add(undo);
-  return true;
+  return;
 }
 
 //-----------------------------------------------------------------------------
@@ -1488,7 +1495,8 @@ void EraserTool::eraseRegion(
  * si deve effettuare una cancellazione.*/
 void EraserTool::doMultiErase(TFrameId &firstFrameId, TFrameId &lastFrameId,
                               const TStroke *firstStroke,
-                              const TStroke *lastStroke) {
+                              const TStroke *lastStroke,
+                              EraserTool::EraseFunction eraseFunction) {
   TXshSimpleLevel *sl =
       TTool::getApplication()->getCurrentLevel()->getLevel()->getSimpleLevel();
   TStroke *first           = new TStroke();
@@ -1542,7 +1550,8 @@ void EraserTool::doMultiErase(TFrameId &firstFrameId, TFrameId &lastFrameId,
       else
         app->getCurrentFrame()->setFid(fid);
     }
-    doErase(backward ? 1 - t : t, sl, fid, firstImage, lastImage);
+    doErase(backward ? 1 - t : t, sl, fid, firstImage, lastImage,
+            eraseFunction);
     notifyImageChanged();
   }
   TUndoManager::manager()->endBlock();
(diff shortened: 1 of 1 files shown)


jrhyme:
Well what can you say, you're just a magician)

beeheemooth:
There is one nuance that I have long wanted to write about, but it is forbidden to write posts under the category "feature request" here. The fact is that vectors created using auto-inbetweens are located in frames of the corresponding interpolation, therefore, to erase them, the "segment" must also have these interpolation modes. In Toonz Raster levels, by default, the eraser and fill have only linear interpolation mode (this is noticeable when you export auto-inbetweened drawing from the vector level to TR level) The lack of interpolation modes for these tools complicates the workflow in places where the non-linear type of interpolation is used .
Is it possible to have interpolation modes for the Eraser Tool?

RodneyBaker:
Nice work @pojienie 
I'm mostly just testing what I'll call the simple case for now.
That is where all frames are either the same or very similar.
As the primary use for the Frame Range option I'd say that is working spectacularly.

Thanks!
Back to testing...

beeheemooth:
I tested the function work this way. I was interested in the behavior of the erasure line shape. By the way, this is a very slow process.
ezgif-2-af89cb2fd647
ezgif-2-4b3818ee670b

pojienie:
@jrhyme Thank you!

@RodneyBaker Thank you for testing and merging!

Yeah, for complex case Segment mode is simply too fragile as it erases all segments the erasure line merely touches, unlike other mode where you need to cover the strokes fully to remove it.

Not sure about the crash you encountered. I occasionally encounter crashes too when I was trying to trace the teeth of a cartoon character but re-opening opentoonz always fixes such problem and thus making it ir-reproducible. I developed a habit where I would close and re-open opentoonz every now and then just so that it won't crash on me. Maybe these two crashes are related?

@beeheemooth The erasure line uses exactly the same linear interpolation as auto-inbetweening. So you can just create two strokes in two different frames of the same level and pretend it's the erasure line.

And yes, detecting intersection is a very expensive process. My algorithm taking O(m + n * m) * O(intersection algorithm) doesn't help either. (where n is how many points the erasure line touches strokes and m is how many strokes there are)

(ps. I might try adding interpolation mode to eraser tool but no promise)

beeheemooth:
The Eraser Tool is also a drawing tool - any artist will tell you that.

openanim:
Thanks for all news vectors features theses are very helpful. I tested segment to clean projet with krita assistants, it became super easy ! And frame range helps to correct mistakes in the auto inbetweening process.

Thoughts : Would it be possible to have a version of your segment that does not erase the lines but just make them invisible (like 0 thickness or zero alpha) ? This would be usefull in particular to  keep the auto inbetweenning process iunbroken and help in many situations related to this kind of workflow.

A bit like the segment function in toonz raster which does not erase the lines but make them invisible. Unfortunately the conversion to vectors breaks the process of inbetweening.

For more details

https://github.com/opentoonz/opentoonz/issues/746

I also created a bounty for this request here. If you are also interested about that https://github.com/opentoonz/opentoonz/issues/1554.

Anyway, thanks for thes latests enchancements very useful !

pojienie:
@beeheemooth But I'm no artist so eraser is used to erase.

just kidding! I understand what you mean!

@openanim $.$

So, I tried implementing it: https://youtu.be/Krehy6dJo0s

I've encountered a problem while testing... The fill doesn't always work. In fact, it rarely works. I tried several times just to got the video.

openanim:
@pojienie Thanks for trying tackling with this request. 

Is the filling problem present when you don't create a group too ?

The second problem will break the auto inbetween process.

Perhaps a thickness-oriented approach available with the selection tool or pump tool could be a solution ? Although I think that currently your segment method is the fastest approach in its use.

zero thickness


The current problem with the pump tool is that it does not allow the thickness value to reach 0. Unlike its gadget available with the selection tool.

RodneyBaker:
@pojienie @openanim 
A workaround that demonstrates a possible way forward for what you are trying to do:

- Draw a vector Tic, Tak, Toe board
- Run Level > Optimize > Replace Vectors with Simplified Vectors on it
- Using Brush Tool and a new (sacrificial color swatch... i.e. the color red) and using line mode... recolor the target segments red
- Using Select tool in Same Style mode... select one of the target segments
- Apply the desired change to that selection (i.e. line thickness change, recolor, or delete the color segment). 

If this basic process was automated it would be able to fulfill the desire.

I call this approach the Sacrificial Color Swatch because if the color swatch is copied just prior to deletion then it can be recovered easily and used again for subsequent targeting of more segments.

Sacrificial Color Selection

If automated there are several thing that could use improvement.  For instance, for the Replace. Vectors with Simplified Vectors phase we don't want to target every intersection just those that are necessary to targeting the selection of vectors we want to modify.

For the Sacrificial color swatch phase we don't want to have to create color swatches and then maintain them if some other method works better.  Although I do like the idea of color swatches maintaining a form of customizable group that can be labeled and organized.  This is the way that groups of vectors... the ones we currently cannot see... should be operating.

Replace Vectors with Simplified Vectors does not always create ideal intersections or cap the vector line's ends optimally.

etc. etc.

Hope my description of the process isn't too confusing.
I should supply a video walkthrough of the process as well and will be glad to if necessary.

jrhyme:
> @pojienie @openanim
> A workaround that demonstrates a possible way forward for what you are trying to do:
> 
> * Draw a vector Tic, Tak, Toe board
> * Run Level > Optimize > Replace Vectors with Simplified Vectors on it
> * Using Brush Tool and a new (sacrificial color swatch... i.e. the color red) and using line mode... recolor the target segments red
> * Using Select tool in Same Style mode... select one of the target segments
> * Apply the desired change to that selection (i.e. line thickness change, recolor, or delete the color segment).
> 
> If this basic process was automated it would be able to fulfill the desire.
> 
> I call this approach the Sacrificial Color Swatch because if the color swatch is copied just prior to deletion then it can be recovered easily and used again for subsequent targeting of more segments.
> 
> Sacrificial Color Selection
> 
> If automated there are several thing that could use improvement. For instance, for the Replace. Vectors with Simplified Vectors phase we don't want to target every intersection just those that are necessary to targeting the selection of vectors we want to modify.
> 
> For the Sacrificial color swatch phase we don't want to have to create color swatches and then maintain them if some other method works better. Although I do like the idea of color swatches maintaining a form of customizable group that can be labeled and organized. This is the way that groups of vectors... the ones we currently cannot see... should be operating.
> 
> Replace Vectors with Simplified Vectors does not always create ideal intersections or cap the vector line's ends optimally.
> 
> etc. etc.
> 
> Hope my description of the process isn't too confusing.
> I should supply a video walkthrough of the process as well and will be glad to if necessary.

It would be desirable.

jrhyme:
> > @pojienie @openanim
> > A workaround that demonstrates a possible way forward for what you are trying to do:
> > 
> > * Draw a vector Tic, Tak, Toe board
> > * Run Level > Optimize > Replace Vectors with Simplified Vectors on it
> > * Using Brush Tool and a new (sacrificial color swatch... i.e. the color red) and using line mode... recolor the target segments red
> > * Using Select tool in Same Style mode... select one of the target segments
> > * Apply the desired change to that selection (i.e. line thickness change, recolor, or delete the color segment).
> > 
> > If this basic process was automated it would be able to fulfill the desire.
> > I call this approach the Sacrificial Color Swatch because if the color swatch is copied just prior to deletion then it can be recovered easily and used again for subsequent targeting of more segments.
> > Sacrificial Color Selection
> > If automated there are several thing that could use improvement. For instance, for the Replace. Vectors with Simplified Vectors phase we don't want to target every intersection just those that are necessary to targeting the selection of vectors we want to modify.
> > For the Sacrificial color swatch phase we don't want to have to create color swatches and then maintain them if some other method works better. Although I do like the idea of color swatches maintaining a form of customizable group that can be labeled and organized. This is the way that groups of vectors... the ones we currently cannot see... should be operating.
> > Replace Vectors with Simplified Vectors does not always create ideal intersections or cap the vector line's ends optimally.
> > etc. etc.
> > Hope my description of the process isn't too confusing.
> > I should supply a video walkthrough of the process as well and will be glad to if necessary.
> 
> It would be desirable. (I'm talking about the video)

beeheemooth:
@openanim I understand what you mean. But keep in mind that it is unlikely that vectors created using autoinbetweens will keep a stable border of thicknesses, given the mess that happens with vector points in this case.

openanim:
@RodneyBaker The problem with using simplify vectors this will break the pose-to-pose tweening process. Like all destructive tools as eraser cutter etc. What you take off should match between poses.

The other problem with this function is also that it will simplify the vectors and therefore modify their shape and length and number or stroke, even minimal between poses.

@beeheemooth

In fact, the change in thickness is not too much of a problem for me as long as the structure is identical and is not broken. What you do before you begin an auto inbetween command.

--
--

An example in cacani may be obvious, I hope.

I draw my two poses first. I apply at the level of the bridge of the nose a hidden segment. This will not break the auto inbetween. But by applying an eraser it broke it.

Among other things, by cleaning my drawing entirely in hidden segment. If tomorrow I need to change the timing or reuse it in another way. I can do it much more easily without having to everything again.

RodneyBaker:
@artisteacher said: 
> The problem with using simplify vectors this will break the pose-to-pose tweening process. Like all destructive tools as eraser cutter etc. What you take off should match between poses.

Yes, my breakdown is just demonstrating the closest process we have that meets the current need.
I am not suggesting that process actually be used but rather that since @pojienie has demonstrated a mastery of using current code for other uses that might be useful code to consider in moving forward as it contains a similar process. 

Currently simply vectors breaks the intersections but perhaps it could create the control point there without breaking the continuity.  There are some potential problems there which are well documented in 3D topology which leads to such things as Pixar's SubDiv approach etc.  The main obstacle from my view being that when (in code) we encounter an intersection we don't necessarily know which way to move forward to establish continuity... should the line go straight forward?  Should it make a right turn?  Left turn?  Terminate?  The code can't read the users mind... or can it?  This is where weighting comes into play where a user can establish rules for favoring one path over another.

beeheemooth:
@openanim Since manongjohn created the Vector Guided Drawing Control panel, the problem of vector recalculation has ceased to be relevant. In tandem with the new Eraser Tool features, the workflow becomes multifaceted. By the way, I saw a video tutorial on Cacani, where the user did an animation of the character’s walk. It took quite a while to remove or hide the intersecting line of legs in each frame.

pojienie:
@openanim Pump tool can make a line's thickness to 0. But a line can only disappear if the whole line's thickness is zero.

@RodneyBaker I don't quite understand what you meant by 

> Using Brush Tool and a new (sacrificial color swatch... i.e. the color red) and using line mode... recolor the target segments red

so.. yeah, a video would be much appreciated.

openanim:
@beeheemooth 

Of course both are big steps and open multiple workflows more free. 
Others points can be covered  if we had a tool to hide the lines, thaughts.  And the walks are good examples also. In cacani, you can quickly switch from the hidden or normal stroke state, it is convenient to achieve the desired effect.

I recently saw a walking contest on cacani with a tutorial in Russian but I did enough not pay attention, are you talking about that one or another older ? But yeah about the time consuming, it can actually take time like a cleanup depends the complexity. Cacani uses a system called boundaries strokes to remedy this. It’s a sort of improved frame range. But we risk to be off topic .

@pojienie Yes indeed, I forgot that it is the wanted portions which cannot make it. 
In any case, I appreciate the attempt but if this function proves more difficult to set up and will take you too long, do not hesitate to postpone it, no pressure. The new eraser is already a spectacular step for the workflow ;).

beeheemooth:
@pojienie I managed to fix one crash with frame range. I enclose the level at which this crash occurs.

[Test_crash.zip](https://github.com/opentoonz/opentoonz/files/4446849/Test_crash.zip)


ezgif-2-617d64e50259

beeheemooth:
I seem to have found the cause of the crash

ezgif-2-509b6f97b59d

pojienie:
@beeheemooth thank you for reporting!

A few things I discovered:

- The crash also happens in freehand mode because it shares the same code with segment mode.
- if you use polyline mode to erase, the Undo data is wrong.
- if you use Rectangular mode, it operates fine. However, the code is still incorrect, explained below.

In the explanation below, fid means a level's index(the number that appears at right side of a cell in xsheet). Also the frame numbers are based on your test_crash project.

For freehand and segment mode crash, The problem is in [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/tnztools/vectorerasertool.cpp#L1548)

If you start erasing at frame 139, `startRowInXSheet` would be 138, `fid.getNumber()` would be 139. 138 + 139 - 1 = 276. Does frame 276 has level? No, so when it tried to get fid by using [getCurrentFid()](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/tnztools/vectorerasertool.cpp#L1388), it returns `<noframe>` and thus, it crashes.

c

I modified it to be `startRowInXSheet + i` and it seems to fix the problem. But there is an edge case where some other level is in between, like so:

b

I'm still thinking what's the expected behavior in such cases.

For rectangular mode, it uses `fid.getNumber() - 1` (in [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/tnztools/vectorerasertool.cpp#L905)) And thus, if you move your level M to the first frame so that it looks like this: 

a

Then erase it using rectangular mode and frame range will also result in a crash. Because it tried to set current frame to `fid - 1` which is 139 - 1 = 138. Frame 138 doesn't have anything so it crashes.

beeheemooth:
as far as I understand fids.size () refers to frames in Xsheet and not Level Strip, then if there is another level in the column, as you showed in the screenshot, perhaps the program should give a warning about the impossibility of this operation.

pojienie:
Oh! `fids.size()` is just size of the fids vector. It actually has nothing to do with xsheet frames nor level strip frame. In this context, it means how many frames the user wants to erase.

Hmm, giving a warning is a good idea. Using this means it'll only need to deal with only one level.

Another edge case I can think about is:

a

Where "start" means the frame the user starts erasing using frame range and "end" means the frame the user ends erasing using frame range. Note that the frame number of level strip is basically random.

Yet another edge case I can see is when there is a space in between: 

a

I kinda understand why the frame range option was originally only able to be done on level strip...... There are some strange corner cases if you do it on xsheet.

beeheemooth:
@pojienie Unfortunately, I'm really bad in C ++, sorry. )
It’s just that in previous versions of OT using the Vector Frame Range Brush, I observed that the operation of the tool depends on the frame order in Xsheet. When reordering or Autorenumber, the program often crashed. But one glitch that I managed to detect turned out to be a very useful feature. 

I created a vector level of 12 frames, shifted the entire row of frames in the Xsheet one frame down, applied Autorenumber, returned the row of frames to their original location (thus the column started from the second frame) and painted with a vector brush (Frame Range) while holding CTRL. At the same time, after each stroke (even if I skipped a few frames for auto inbetweens), the program made the transition to the next frame, which seemed very convenient to me, since this is how the Fill Tool in TR Levels behaves (only there you need to hold Shift)

But recently, I have not observed this “useful” glitch (oh no, I do not miss it). Apparently some changes have occurred.

By the way, the location of frames can be in the most unexpected way, thereby paving the way for the non-standard use of tools in the workflow.

ezgif-2-e39b650eb58d

PR #3200, opened 2020-04-01
Allow loading levels/images with numeric names

### LINKED ISSUE #2945: Levels with two character alpha numeric filename fail
### Issue Summary
When loading a level from a file, the program cannot parse its name if it begins with a number.  If you select such a file in the Load Level dialog or drag and drop it onto the program, the filename will simply show up blank (except for its extension). If you manually enter the filename, it will go missing when the program asks if you want to import it or load it.

opentoonz_error

### System Information
* **OpenToonz Version**:  1.3
* **Operating System**:  Windows 10

martinvanzijl:
Fixes #2945. Video demo below - here I select images with name "1a.png" and "1234.png", and the filename is loaded correctly in the Load Level dialog:

numeric-filename-demo

I tested a sequence, too ("series.0001", "series.0002") - also works fine, still.

### DIFF
--- a/toonz/sources/common/tsystem/tfilepath.cpp
+++ b/toonz/sources/common/tsystem/tfilepath.cpp
@@ -606,8 +606,7 @@ std::string TFilePath::getLevelName() const {
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Works great.
Thanks @martinvanzijl !

Issue #1733, opened 2018-01-14
Cut Tool Feature Request [$120]


UncleTan45:
This could go a long way in terms of quality of life for the users of Toonz.
https://www.youtube.com/watch?v=iT_oMHF8Bs0

At 5:31 I want the cut tool to work like this if you mess with the tool options on the cut tool. This should be easy with the 'remove vector overflow' coding that's already in the program. It'd just need a slight adjustment.

artisteacher:
That would be really handy!

beeheemooth:
In OpenToonz it is possible to do this in Vector and Toonz Raster levels using other approaches to solving this problem.
[https://www.youtube.com/watch?v=N-Gds2B2pLk&t](url)
[https://www.youtube.com/watch?v=_Jr88wsNJfs](url)
However, in this case, you need to do a little more action, but you can embrace a larger range of frames and as a result, the speed of work should increase.

Orphanlast:
If you're wanting to cover a large number of frames, then yeah that's an effective meathod. @beeheemooth but using the frame range brush and auto inbetweener isn't always what's needed. I still say this is something that's nice to have on the fly. 

I do believe that using this feature for the cut tool will probably destroy vector guided drawing capability, but, some of us have no intention of using vector guided drawing.

RodneyBaker:
Feature Requests must go the route of discovery via the OpenToonz forum.
As such launching a discussion there will get the ball rolling and let other know of your interest.

[Link to Feature Request section of the OpenToonz Forum](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!categories/opentoonz_en/feature-requests)

Orphanlast:
@RodneyBaker so my $70 is just poof gone, is what you're saying?

manongjohn:
If there is a bounty on this, post link in here so we know and wont close it.

Orphanlast:
The OP has no idea how to do that. And if I can do that, I have no clue how. 

What is the problem? And it still sounds like my $70 went down the shitter. 

I am seriously not happy right now.

Orphanlast:
and what's worse is it sounds like you guys have the power, at a whim to make any of my donations go down the toilet, using this donation platform.

I've been generous, and I'm getting hosed, and you guys can hose me any time, any bounty, at any time you want if I continue using this platform.

This is a volatile donation platform. Figure it out! FIX IT!

Orphanlast:
[...]

PR #3202, opened 2020-04-01
Add shortcut to eraser modes

pojienie:
this PR allows you to assign shortcuts to eraser modes.

a

In the code, I had to make it switch to normal then to the respective mode. Without doing so, it will switch back to normal mode if it's already in that mode. e.g. user press freehand shortcut -> switch to eraser freehand mode -> user presses it again -> switch to eraser normal mode.

Tried to find the root cause but to no avail. Well, this workaround works.

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -2452,6 +2452,7 @@ void MainWindow::toggleEraserNormal() {
 
 void MainWindow::toggleEraserRectangular() {
   CommandManager::instance()->getAction(T_Eraser)->trigger();
+  CommandManager::instance()->getAction("A_ToolOption_Type:Normal")->trigger();
   CommandManager::instance()
       ->getAction("A_ToolOption_Type:Rectangular")
       ->trigger();
@@ -2459,6 +2460,7 @@ void MainWindow::toggleEraserRectangular() {
 
 void MainWindow::toggleEraserFreehand() {
   CommandManager::instance()->getAction(T_Eraser)->trigger();
+  CommandManager::instance()->getAction("A_ToolOption_Type:Normal")->trigger();
   CommandManager::instance()
       ->getAction("A_ToolOption_Type:Freehand")
       ->trigger();
@@ -2466,6 +2468,7 @@ void MainWindow::toggleEraserFreehand() {
 
 void MainWindow::toggleEraserPolyline() {
   CommandManager::instance()->getAction(T_Eraser)->trigger();
+  CommandManager::instance()->getAction("A_ToolOption_Type:Normal")->trigger();
   CommandManager::instance()
       ->getAction("A_ToolOption_Type:Polyline")
       ->trigger();
@@ -2473,6 +2476,7 @@ void MainWindow::toggleEraserPolyline() {
 
 void MainWindow::toggleEraserSegment() {
   CommandManager::instance()->getAction(T_Eraser)->trigger();
+  CommandManager::instance()->getAction("A_ToolOption_Type:Normal")->trigger();
   CommandManager::instance()->getAction("A_ToolOption_Type:Segment")->trigger();
 }
 
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -450,6 +450,14 @@ centralWidget->setLayout(centralWidgetLayout);*/
   setCommandHandler(MI_PickStyleAreas, this, &MainWindow::togglePickStyleAreas);
   setCommandHandler(MI_PickStyleLines, this, &MainWindow::togglePickStyleLines);
 
+  /*-- Eraser Tool modes --*/
+  setCommandHandler(MI_EraserNormal, this, &MainWindow::toggleEraserNormal);
+  setCommandHandler(MI_EraserRectangular, this,
+                    &MainWindow::toggleEraserRectangular);
+  setCommandHandler(MI_EraserFreehand, this, &MainWindow::toggleEraserFreehand);
+  setCommandHandler(MI_EraserPolyline, this, &MainWindow::toggleEraserPolyline);
+  setCommandHandler(MI_EraserSegment, this, &MainWindow::toggleEraserSegment);
+
   setCommandHandler(MI_About, this, &MainWindow::onAbout);
   setCommandHandler(MI_OpenOnlineManual, this, &MainWindow::onOpenOnlineManual);
   setCommandHandler(MI_OpenWhatsNew, this, &MainWindow::onOpenWhatsNew);
@@ -2314,6 +2322,8 @@ void MainWindow::defineActions() {
                           "");
   createToolOptionsAction("A_ToolOption_Type:Polyline", tr("Type - Polyline"),
                           "");
+  createToolOptionsAction("A_ToolOption_Type:Segment", tr("Type - Segment"),
+                          "");
   createToolOptionsAction("A_ToolOption_TypeFont", tr("TypeTool Font"), "");
   createToolOptionsAction("A_ToolOption_TypeSize", tr("TypeTool Size"), "");
   createToolOptionsAction("A_ToolOption_TypeStyle", tr("TypeTool Style"), "");
@@ -2360,6 +2370,18 @@ void MainWindow::defineActions() {
   createAction(MI_PickStyleAreas, tr("Style Picker Tool - Areas"), "",
                ToolCommandType);
   createAction(MI_PickStyleLines, tr("Style Picker Tool - Lines"), "",
+               ToolCommandType);
+
+  /*-- Eraser Tool modes --*/
+  createAction(MI_EraserNormal, tr("Eraser Tool - Normal"), "",
+               ToolCommandType);
+  createAction(MI_EraserRectangular, tr("Eraser Tool - Rectangular"), "",
+               ToolCommandType);
+  createAction(MI_EraserFreehand, tr("Eraser Tool - Freehand"), "",
+               ToolCommandType);
+  createAction(MI_EraserPolyline, tr("Eraser Tool - Polyline"), "",
+               ToolCommandType);
+  createAction(MI_EraserSegment, tr("Eraser Tool - Segment"), "",
                ToolCommandType);
 
   createMiscAction("A_FxSchematicToggle", tr("Toggle FX/Stage schematic"), "");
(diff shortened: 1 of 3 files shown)


beeheemooth:
@pojienie Tools that are of type "Normal" (+ Rectangular, Freehand, Polyline) switch back to type "Normal" when you press the hotkey again. This is the Eraser Tool, Fill Tool, Tape Tool. The Selection Tool is not of type "Normal" and the types do not have this behavior. (This is just my observation).
Honestly, I thought that you would make Segment a separate option, as in the Fill Tool, so that it was available, say, only when the Freehand type was selected, and you could switch from Freehand to Segment. But let it be as it is.
Thank you by the way, that you assigned the continuation of the chain the erase operation to the Shift key, and not to Ctrl (as it is implemented in the Vector Brush Tool, because of this terrible confusion).

 Also accept thanks from a large number of our users for your efforts.

artisteacher:
@pojienie - Thanks for all your hard work! In addition to this, would it be possible to cycle through the types with the shortcut for the eraser? Ie. press E to enter the eraser, pressing E when the eraser is active lets you toggle through the different types. This would be so helpful for all of the tools with types as well as the animate & geometric tools. Less to memorize!

melieconiek:
Everything seems to be working here! Thank you!

If you haven't already, you might want to see if you can get the funds released for the bounty on the cutter tool. I'm not sure how that works.

RodneyBaker:
This works great @pojienie.  
I like @beeheemooth and artisteacher 's thoughts of cycling through the modes and the ease by which repeatedly pressing the same key (default key of E) would move from erase mode to erase mode. 

Off topic workflow thought:  As of now, I rarely use rectangle and polyline mode personally so in the current scheme of things I mostly just need to set three keys.  At present I'm playing with the shortcuts being: E (Erase Tool), Shift E (Segment Mode) and Control Shift (Normal Mode).  Hard to say how long I'll leave it that way but that key combination is working nicely at present for me. 

I assume further additions related to this PR would be in a separate PR but will delay merging for others to have time to test and provide feedback.

Thank you for allowing users the use of shortcut keys for your outstanding feature.

pojienie:
@beeheemooth Thank you for sharing your observation!

Yeah, tool types that has no "normal" wouldn't be switched back and it would otherwise. Funny thing is, if I change the eraser tool's "normal" mode to "normal_1" and it wouldn't switch back when you press the shortcut key the second time. There must be some special code for the word "normal" but I just couldn't find it.

At the time I did think about adding it as an option instead of another mode. But I feel if only freehand can use it, adding a mode made more sense to me. But then I realized polyline could also use it...... ugh, too late.

The chain hotkey is not programmed by me! It was already there, I just reused their code. So thank them instead!

Also, thank you for letting me know that there are more users of this tool than I thought! 

@artisteacher Thanks for suggestion! Cycling through them would be awesome. I'll try, but no promise. The shortcut system is nothing short of magical as I described above for @beeheemooth.

@melieconiek You meant #1733, right? Thank you for reminding me! Not entirely sure if the segment mode would qualify as their cutter tool is even more amazing.

@RodneyBaker Yeah, I like the cycling idea, too!

By "further additions related to this PR", did you mean other shortcuts that directly switch to some particular mode of a particular tool? If that is the case, I should be able to make it happen. But I urge you not to merge this PR if this is the case. I'll create another one that has all tool modes shortcuts.

RodneyBaker:
Yes, if the moving from one mode to another via a shortcut key could be added that would be very good reason to delay merging this PR and allow another to replicate what is currently accomplished herein.

pojienie:
During my study, I discovered that the functionality actually already exists... Although slightly less convenient:
a
To use this, you need to press eraser tool then corresponding mode. And it will switch back to normal if you use the shortcut when it's already in that mode.
There is also one for animate tool, just search "Active axis". Again, you need to switch to animate tool then use the shortcut.
Hmmm, not entirely sure how I should proceed now. Should I still add a "switch to X tool and also Y mode" to all tools and modes? What do you guys think?

beeheemooth:
I like the approach that Shun Iwasava implemented in Shift and Trace: when this tool is turned on, the keys F1, F2, F3 are used to scroll through the drawings. If the Shift and Trace function is disabled, then F1, F2, F3 perform other functions assigned to them. The idea of this is that with a certain active tool, some reserved keys (for example, F1 ... Fn) can switch modes or options of the current tool.

For example, Animate Tool is on:

F1 - Position (modes in order)
F2 - Rotation
F3 - Scale
F4 - Shear
F5 - Center
F6 - All

Selection Tool:

F1 - Rectangular
F2 - Freehand
F3 - Polyline
F4 - Standart
F5 - Selected Frames
F6 - Whole Level
F7 - Same Style
F8 - Same Style on Selected Frames
F9 - Same Style on Whole Level
F10 - Boundary Strokes
F11 - Boundaries on Selected Frames
F12 - Boundaries on Whole Level

etc...

With this approach, we can save a huge number of shortcuts, and introduce a logical order for the tool options.


However, this idea also has its drawbacks:
1. The buttons are hardcoded.
2. In different types of levels, different instrument settings
3. For some instuments (for example Selection Tool) the number of options may exceed 12 reserved buttons.
4. Some tools when switching modes add additional options.

Hmm ... Well, in general, I just fantasized on this subject.

artisteacher:
The cycling would be more convenient, if it could share the shortcut for the tool itself, instead of needing a 2nd shortcut for the type. There are already so many shortcuts in OpenToonz - it's tricky finding available key combinations & it's a lot to remember.

RodneyBaker:
@artisteacher I agree with your assessment.
While this is outside the scope of this PR the whole of shortcutting should be reviewed with an eye to simplify the hotkey process to maximum advantage of the user.

The would include considering the distance and time (however small) that it takes the user to change from one keyboard location to another especially to identify strange combinations that force users to use awkward key combinations or look at the keyboard.

This is one of the reasons I like the idea of being able to hit the E key once to activate the Erase Tool and hit it again to move to another ideal setting within the context of the Erase Tool.  This is perhaps the area where things get tricky because what is important to one user won't be to another.  
I'm not sure how that would be addressed short of updating the shortcut key process to allow for a finer resolution on how hotkeys work.

Regarding this PR...

> The cycling feature that @artisteacher @RodneyBaker mentioned is in there, too. Just set a shortcut key to the one without mode (e.g. "Type" in the above picture) and it will automatically cycle.
> 
> Hmmm, not entirely sure how I should proceed now. Should I still add a "switch to X tool and also Y mode" to all tools and modes? What do you guys think?

You remind me of that old lesson where the first step is to consider what we already know... to inventory current resources.  I must admit there appears to be a lot about the current shortcut process I do not know.  For instance:

> press eraser tool then corresponding mode. And it will switch back to normal if you use the shortcut when it's already in that mode. 

For some reason this doesn't quite compute.
I must assume this means select the Erase Tool icon with mouse/pen and then enter the shortcut combination for the desired tool.  If so, this is logical; i.e. activate the tool and then enter the mode.

At any rate, how to proceed... I'm not entirely sure.

MrBlockCat:
Something else that cycles through tool options with one key is the Geometric Tool's "Geometric Shape" shortcut seen below, though it only does so if the Geometric Tool is selected, and doesn't select the Geometric Tool for you (in a related note, I'd love to see a keyboard option for each geometric shape like you've done here for the erasers!)
Screen Shot 2020-04-02 at 3 07 33 PM

I have no clue how the code for keyboard shortcuts works, but hopefully this helps you out if you do decide to go for @artisteacher 's idea!

I definitely agree keyboard shortcuts in general should be re-reviewed at some point, there are many different aspects it would be great to simplify, add to, etc.

pojienie:
@beeheemooth I would probably need to revamp the whole shortcut system to implement that in... Judging by my current understanding of the shortcut system, it's out of my league currently. And, as you said, it would require hardcoding which is something I want to avoid always.

@artisteacher Yeah, I'd prefer same shortcut for cycling and for selecting as well. (first time you press shortcut, it switches to that tool. Second time you do so, it cycles the mode.) I saw there is a function that allows me to get current tool selected but I haven't verified if it would work. If it would, such functionality will be possible.

@RodneyBaker Yeah, a reorganize of shortcuts is desired. The person who reorganize would need to know what's the commonly used tool and what's not. They would also need to get a lot of feedback to achieve it as everyone's workflow is different. I can **probably** implement it if anyone comes up with a design since all the code are already in place and I only need to rearrange them. Still, no promise though.

Great lesson by the way. There are two features now that I implemented for nothing since the functionality already exists, lol.

Also, sorry for my bad English. By that I meant, for example, if the current tool is eraser and mode is normal and you press freehand mode shortcut, it will switch to freehand mode. However, pressing it again and it will switch back to normal mode. The same applies for all other modes except normal.

@MrBlockCat Thank you for sharing your observation! I'll study that part as well!

openanim:
I am not too fan of the idea of ​​imposed shortcuts. For all the bad sides you listed @beeheemooth .

Personally with the functionality of the current cycles, what I would like is to have a shortcut maybe a modifier that alternates the direction of the cycle in order to make it shorter when necessary.

pojienie:
It seems to me, there are 3 types of shortcut (as far as switching tools is concerned):

1. general: shortcut that works on multiple tools. 
Advantage: one shortcut key for several tools, less shortcut keys to remember.
Disadvantage: takes slightly longer to switch if you also want to switch tool
For example: "Type - Freehand" can be used in selection tool, fill tool, and eraser tool.
note: this category also provides mode-cycling.

2. tool switching: shortcut that switches tool and switches only tool.
Advantage: after pressing the shortcut, you can be sure of two things: 1. you're using the tool and 2. the tool's mode is the mode you last used.
Disadvantage: this only does one thing, need to be combined with shortcuts in general category.

3. tool specific: shortcut that works only on one tool. 
Advantage: one shortcut does two things: switching tool and switching mode, and is thus faster
disadvantage: a lot of shortcuts to remember.
For example: "Fill Tool - Areas" switches to fill tool and its mode to areas
note: tool-specific mode-cycling will be in here too, instead of tool switching, explained below. Not sure if changing direction that @openanim is possible, but if it is, it's also in here.

I personally consider all 3 of these categories are important and no two can be merged. However, this is because I do not know how most people use opentoonz and thus I must treat all 3 categories equally.

The reason why I said that the tool-specific mode-cycling belongs to tool-specific category: When I use opentoonz, I have a habit that I'll press keys in tool-switching category several times. For example: I use geometric tool a lot but each time I draw a stroke using it, I'll instinctively press the shortcut key every time I draw a stroke even though I am already using geometric tool. I certainly don't want the mode to be changed.

What are you guys' thought?

melieconiek:
Anything that requires a combination of successive key presses is an issue for me (select tool, select mode.) **I prefer direct shortcuts to whatever it is I intend to use. I should know what tool I have selected and in what mode without having to look at the tool option bar.**

I like what you've done here with this I would love to see it brought to other tools, like the tape tool's mode options.

---

As an aside--Clip Studio Paint has combined tools with many options just like OpenToonz. CSP handles this by allowing you to define presets for all tools and assign shortcuts to them. There is only one 'Brush' tool--the vector eraser modes are actually part of this, so if you want a binding for the segment or line erasers you have to define a preset and binding for these. It's a very flexible system.

This is something that could be made to work for OpenToonz.

artisteacher:
I hadn't realized how many shortcuts fall under #1. I would like to see more of these "general" shortcuts have a default shortcut set - at least for type, mode, shape, and snapping. Maybe general shortcuts would be a good choice for the function keys?

As far as #3 goes, there are tool specific shortcuts, they just don't change the tool at the same time as the mode or etc. I can see the benefit of number 3, though I'm unlikely to learn that many individual shortcuts myself.

MrBlockCat:
A thought based on @pojienie saying you press the key every time you do something without intent of changing something: Maybe there could be an option in Preferences to enable or disable rotating between things within a tool with the same key as to go to that tool?

pojienie:
@melieconiek It is possible to achieve what you want currently: Just press the shortcut for tool then the shortcut for mode.

Also, as you said, that Clip Studio Paint feature/revamp is probably too big for me to handle...

@artisteacher Adding default shortcut in code is easy but deciding which shortcut to use to use as default is completely outside of my ability here......

About category 3 stuff, at least for me, I'll probably just set a few that I used most often. But, even if they don't exist, I can always just combine category 1 and category 2 to achieve it.

@MrBlockCat Hmm, a good idea. People who like to bind the same shortcut key to both tool switching and mode cycling can just enable it. People who don't can disable it can combine both category 1 and 2 shortcut to achieve tool switching and mode cycling whenever they want.

Although, I need to study how to add an option in the preference in this case.......

Thank you guys for sharing your opinion! From what you guys shared here, I am starting to see category 3 being quite useless. And thus, I guess this PR can be closed? Since everything I added in this PR all falls in category 3. And category 1 and 2 are already there. (I'll submit another PR to add "Type - Segment" as I forgot to add it when I implemented it)

melieconiek:
@pojienie I agree that you should add the 'Type - Segment' shortcut, but this functionality does NOT currently exist and I would very much like to see this PR merged.

MrBlockCat:
For me personally, since I’d love to see the Geometric Tool’s shapes each have their own shortcut options, category 3 in some cases is very appreciated by me! On that basis, I agree that this PR has value to me despite being in category 3.

melieconiek:
I've added this PR to my personal build so thank you for putting it together @pojienie. Maybe this can be revisited later and expanded to include some direct shortcuts for other tools. :)

pojienie:
Maybe I reached conclusion a bit too quick. I thought shortcuts that switches tool and also its mode can be easily achieved by switching the tool then switch the mode instead. And thus I see those shortcuts redundant. If there are needs of such shortcuts, I'll add all of them in in my next PR. It's bound to clutter the "configure shortcut" window but we can discuss how to organize them in the next PR instead of not putting any of them in.

Issue #3137, opened 2020-03-05
Geometric Tool Shapes should have individually mappable keyboard shortcuts


MrBlockCat:
As I was working on the new Adobe Animate keyboard shortcut preset, I noticed that interestingly OpenToonz has a mappable keyboard shortcut to switch to the Geometric Tool (Tools > Geometric Tool in the keyboard shortcuts window), and one to rotate between the shape the tool'll create (Tool Modifiers > Geometric Shape), but you can't as far as I can tell have individual mappings for individual shapes, ie. Rectangle or Polyline. It would be great to have those shortcuts available to save some time when animating and make the presets more accurate. I'm thinking in the Tool Modifiers section, instead of Geometric Shape, maybe we could expand that into:

Geometric Tool Shape (what Geometric Shape does now, no dash to indicate it's different than the others)
Geometric Tool - Rectangle
Geometric Tool - Circle
Geometric Tool - Ellipse
Geometric Tool - Line
Geometric Tool - Polyline
Geometric Tool - Arc
Geometric Tool - Polygon

artisteacher:
In case you are not aware, you can use the shortcut for the geometric shape to cycle through the different shapes - as long as in Preferences -> Tools, you select "Cycle through the available options" in the dropdown shortcuts option.

Personally, rather than having individual shortcuts for each shape, I'd love to be able to use the Geometric tool's shortcut to cycle through the shape options. Pressing G the first time would make the geometric tool active. Pressing G when the geometric tool is active would cycle through the shapes. Just one shortcut to remember and no additional shortcuts would need to be set. It could be handy for other tools as well.

MrBlockCat:
Yeah, I’d noticed the Geometric Shape shortcut and that it cycles shapes. I like the idea of merging that shortcut with the one to switch to the Geometric Tool how you said! I think as an option, though, it would be nice to also be able to switch shapes using their own keyboard shortcuts, to save time instead of having to cycle if you prefer that (as I personally would). To combine the best of both ideas, maybe we could have:

In Tool Modifiers:
Geometric Tool - Rectangle
Geometric Tool - Circle
Geometric Tool - Ellipse
Geometric Tool - Line
Geometric Tool - Polyline
Geometric Tool - Arc
Geometric Tool - Polygon

And in Tools:
Geometric Tool (Switches to Geometric Tool when pressed once, if pressed and you’re already in Geometric Tool it starts cycling shapes)

[...]

PR #3213, opened 2020-04-05
Expand tool-switching shortcuts

pojienie:
improvement of #3202 

a

99% of the code I added is boilerplate. I tried using lambda to reduce it but failed. So if anyone knows how to reduce this boilerplate, please share!

I only check Toonz Vector level for all modes/types shortcuts. I believe anyone who uses other levels and notices that some shortcuts are missing should be able to add those shortcuts themselves by reading this PR. (please also submit a PR if you do so)

Now, for all tool switching shortcuts, there will be 5 types of shortcuts:

- tool switching - shortcuts that only switch tools and nothing else (e.g. "Eraser")
- mode cycling - shortcuts that cycles modes/types of current tool (e.g. "Mode", "Type")
- mode switching - shortcuts that switches modes/types of current tool (e.g. "Type - Freehand")
- tool switching + mode cycling - shortcuts that switches tool if you're not using that tool, cycles modes/types if you are using the tool (e.g. "Eraser Tool - Next Type")
- tool switching + mode switching shortcuts that switches tool AND modes/types (e.g. "Eraser Tool - Freehand")

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -452,6 +452,16 @@ centralWidget->setLayout(centralWidgetLayout);*/
   setCommandHandler(MI_EditCenter, this, &MainWindow::toggleEditNextCenter);
   setCommandHandler(MI_EditAll, this, &MainWindow::toggleEditNextAll);
 
+  /*-- Selection tool + mode switching shortcuts --*/
+  setCommandHandler(MI_SelectionNextMode, this,
+                    &MainWindow::toggleSelectionNextMode);
+  setCommandHandler(MI_SelectionRectangular, this,
+                    &MainWindow::toggleSelectionRectangular);
+  setCommandHandler(MI_SelectionFreehand, this,
+                    &MainWindow::toggleSelectionFreehand);
+  setCommandHandler(MI_SelectionPolyline, this,
+                    &MainWindow::toggleSelectionPolyline);
+
   /*-- FillAreas,FillLinesに直接切り替えるコマンド --*/
   setCommandHandler(MI_FillAreas, this, &MainWindow::toggleFillAreas);
   setCommandHandler(MI_FillLines, this, &MainWindow::toggleFillLines);
@@ -2372,6 +2382,16 @@ void MainWindow::defineActions() {
   createAction(MI_EditCenter, tr("Animate Tool - Center"), "", ToolCommandType);
   createAction(MI_EditAll, tr("Animate Tool - All"), "", ToolCommandType);
 
+  /*-- Selection tool + mode switching shortcuts --*/
+  createAction(MI_SelectionNextMode, tr("Selection Tool - Next Mode"), "",
+               ToolCommandType);
+  createAction(MI_SelectionRectangular, tr("Selection Tool - Rectangular"), "",
+               ToolCommandType);
+  createAction(MI_SelectionFreehand, tr("Selection Tool - Freehand"), "",
+               ToolCommandType);
+  createAction(MI_SelectionPolyline, tr("Selection Tool - Polyline"), "",
+               ToolCommandType);
+
   /*-- FillAreas, FillLinesにキー1つで切り替えるためのコマンド --*/
   createAction(MI_FillAreas, tr("Fill Tool - Areas"), "", ToolCommandType);
   createAction(MI_FillLines, tr("Fill Tool - Lines"), "", ToolCommandType);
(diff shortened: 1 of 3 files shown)


jrhyme:
Hmm ..... why I don’t see the like button ...
Well done, as always!

artisteacher:
I think you've found a way to give literally everyone what they want!

artisteacher:
We might need a different organizational structure in the configure shortcuts window. Should we have 2 categories for Tool Modifiers? ie. Tool Specific Modifiers & Universal Tool Modifiers

Alternatively, the tool category could have sub-categories for the tool specific modifiers. Ie. An Eraser subcategory would hold Then the existing Tool Modifiers category could hold the general/universal tool modifiers.
shortcut_window

openanim:
Great job, thanks !

MrBlockCat:
First of all, amazing work! This would already make using OpenToonz so much easier for me. My thought for the organization is to maybe get rid of the Tool Modifiers category, have what @artisteacher ‘s picture shows for the tool specific modifiers, and then have another subcategory within the Tools category for Universal Tool Modifiers. Would reorganizing the tool modifiers impact the ability to load existing shortcut presets?

pojienie:
I agree we need to reorganize the "configure shortcut" window but I don't want to call the shot here because I'm not the target user of Opentoonz. Also, the coding aspect seems to be more difficult than it should be and thus, in order to reorganize, we might need to revamp the corresponding part of the code...

@artisteacher I implemented one solution for the the image you provided:
a

BUT, it heavily relies on translation being consistent with English since it checks the translated string in order to categorize. I'm still looking for a way to organize them without relying on translated string. But the code feels like it inherently doesn't support such things......

@MrBlockCat loading shortcuts doesn't seem to rely on the way "configure shortcut" window is displayed. And thus reorganizing shouldn't cause any problem with loading presets.

RodneyBaker:
Looking good thus far.
It seems to me there is yet another type of shortcut that perhaps falls a bit into mode switching but the problem with that categorization is that I'm not sure it is limited to the switching of modes.  In thinking about it it makes me think of a possible second shortcut type... a generic passthrough switch as well.  I suppose that one would logically be type 0 which would include the shortcut option being off entirely.

The type of shortcut I'm speaking of is often demonstrated by doing something with the pen or mouse but doing it while holding down the Shift key.  Depending on where the user is at that moment the addition of the Shift tends to do one of two things; drops the current process in favor of a lower process or conversely raises the process to a higher level functioning.

What success might look like in the type of shortcut with using the Eraser Tool in the Segment mode for instance would be:

The user is erasing segments of a drawing.  While erasing the segments are removed as expected when using the segment mode.  The user adds the use of Shift key to modify usage and... while holding down the Shift key... the result is that the Erase Tool acts as it normally would as an eraser in the default/normal mode as if no options were selected.

So in a way, this type of shortcut is like a passthrough except it negates the presence of any other options other than the default.

That's the first branch of that type of shortcut... the user using the Shift modifier to drop back to the default state.

The second is what happens when the Shift modifier is held while in the default state.
This can be more complex because each user will have a different preference based on workflow and need.  

Therefore, an elevated shiftstate is a good candidate for customization to fit localized and temporal needs whereas the bypass/passthrough state is a good candidate for hard coding to achieve maximum standardization.

In my estimation it would be ideal if the user could hold the shift key down and know with confidence what the results would be.  In the first case... that the tool would be using the default behavior if using customized options and in the second case... their personal customizations.

pojienie:
@RodneyBaker I see where you're going, it's like a secondary tool that you can quickly access by holding shift, except it is also context aware. Having such option would be awesome, for sure, but, eh, it's completely out of my league. Maybe you can create a feature request to log it and, hopefully, someone would be able to implement it in the future.

About this PR, I'm sad to say that I wasn't able to find a way to categorize the shortcuts without relying on translation.

RodneyBaker:
> it's like a secondary tool that you can quickly access by holding shift, except it is also context aware.

You are right about this being another PR of course.  I'll do some additional R&D and see where I can make a good case for the shift operation as well as additional context awareness and customization.  Initially, my focus was just on the shift/negation... not much context awareness needed there... but your feedback has me thinking about ways the approach could be made even better.

There is one downside I can think of which would make this key check need to be a Preference and that is the time it takes to verify a signal from the keyboard has been intercepted.  This isn't a lot of time initially but could add up especially if the code was not optimized for it.  BUT... I believe Opentoonz already performs this check as it is continuously waiting to receive keyboard input.

pojienie:
Yeah if it is going to exist, it is definitely another PR.

CPU resources wise, rendering, smoothing for brush tool, and tweening take far more resources than checking keyboard keys so you probably don't need to worry about this unless it really becomes a usability problem.

RodneyBaker:
I've minimized most posts in this PR in order determine and what is still needed.
Improves #3202.  Fixes #3137.
Looking good!

melieconiek:
Excellent! Thank you for this!

RodneyBaker:
I'd say 4 days is more than enough time for the usual suspects to provide feedback.
LGTM

openanim:
@pojienie Unless I haven't seen it, it seems that the entries in the selection tool are not affected by the mode shortcut ?

pojienie:
@openanim Yeah, I missed it... Also, in the code, they are not the same "mode" for some reason.

MrBlockCat:
I just noticed something: Is there a difference between Tool Modifiers > Geometric Shape, Geometric Shape Rectangle, etc. (Lines 2411 to 2428 in the code) and Tools > Geometric Tool, Geometric Tool - Rectangle, etc. (Lines 2523 to 2538)?

pojienie:
The former shortcuts change the mode of current tool only. If current tool doesn't support geometric shape, the former shortcuts do nothing. (Only geometric tool supports geometric shape currently)

The later shortcuts change current tool to geometric tool and its mode.

PR #3215, opened 2020-04-07
mac frameworks cmake fix

artisteacher:
This resolves almost all of #2292. Currently, if libpng exists in frameworks (like mono.framework), that is used when building instead of the intended libpng installation. Somehow the incompatibility between libpng versions prevents mypaint brush icons from displaying. It also causes crashes on making a new scene and sometimes on saving. Setting frameworks to be used last, allows the correct libpng library to be used. This could potentially prevent other libraries from unintentionally being used from frameworks when building.

This also changes "installation" to be spelled correctly.


RodneyBaker:
The code suggests all is well but would prefer to have a Mac tester validate this cmake change.
Otherwise this look ready to be merged.

MrBlockCat:
Unfortunately, my MacBook fairly spontaneously died so I personally won't be able to test things for about a week while Apple fixes it. Best of luck though! I love to see improved Mac support

MrBlockCat:
I can now try and test this! Is all I need to do to try building with this branch, or do I need to install another version of cmake elsewhere or something more complex?

RodneyBaker:
This PR seems to have been forgotten.

artisteacher:
> This PR seems to have been forgotten.

There's nothing else I can do on my end. I don't think this would have a negative impact. It doesn't prevent frameworks libraries from being used - just makes them the last resort if libraries aren't in the expected places. It is really necessary for building on macOS if certain software (ie. Unity) is installed.

RodneyBaker:
As far as I can tell the only thing this PR is waiting for is the confirmation from a Mac user that it works as described with no ill effects.  For my part, what I see in the code is very straightforward.

We do have one confirmation through @MrBlockCat 

I'll call intent to merge this PR and we'll see if other interested parties can fit a quick review into their schedule.  

cc:  @turtletooth  (as I believe you were the one that requested the PR from @artisteacher)

PR #3216, opened 2020-04-07
Add interpolation in vector eraser tool

pojienie:
This PR allows you to set interpolation in vector eraser tool.

video: https://youtu.be/34zmqKEn_c4

I moved the inbetween calculation from filmstripcommand(used when you select two frames in a level and click on "INBETWEEN") to TInbetween so that eraser tool can share the same calculation.


jrhyme:
Taking off my hat) Well done, as always)

RodneyBaker:
This PR... LGTM.
Will delay to allow others to test and provide feedback.

beeheemooth:
@pojienie what you did is just great! Thank you!

PR #3217, opened 2020-04-07
Keep focus on textboxes when moving mouse to another panel

### LINKED ISSUE #1262: UI:  Text fields lose focus  while selected
hoverbug
the textfield in the edit window looses its selection when hovering away from it.
you can only change the number when keeping the cursor exactly over the textfield.
other texfields work properly.
i use it often to flip my characters heads horizontaly by setting horizontal scale from 100% to -100%

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46050145-loose-selection-in-edit-text-field?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

martinvanzijl:
Fixes #1262. Adds an option to keep focus on text boxes when moving the cursor to other panels. This is in the Preferences window under "Interface --> Keep Textbox Focus When Moving Mouse to Another Panel".

Video demo below:

textbox-focus-demo

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -231,9 +231,6 @@ public:
   }
   void setColorCalibrationLutPath(QString monitorName, QString path);
   QString getColorCalibrationLutPath(QString &monitorName) const;
-  bool isKeepTextboxFocus() const {
-    return getBoolValue(keepTextboxFocus);
-  }
 
   // Visualization  tab
   bool getShow0ThickLines() const { return getBoolValue(show0ThickLines); }
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -231,6 +231,9 @@ public:
   }
   void setColorCalibrationLutPath(QString monitorName, QString path);
   QString getColorCalibrationLutPath(QString &monitorName) const;
+  bool isKeepTextboxFocus() const {
+    return getBoolValue(keepTextboxFocus);
+  }
 
   // Visualization  tab
   bool getShow0ThickLines() const { return getBoolValue(show0ThickLines); }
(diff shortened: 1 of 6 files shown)


martinvanzijl:
To be honest, I'm not sure we really need a preference for this. Maybe we should just make it the standard behaviour. Let me know your thoughts.

MrBlockCat:
My opinion as a relatively inexperienced user is that I'd want this option on! (aka I'd like to keep focus despite mouse movement) I can see how keeping it as it is could be helpful, though, so I'm curious what others think

bobbybutterfly:
Thanks @martinvanzijl awesome! But i think "keep focus on textboxes" could be the standard behavior like in all the other software i know

RodneyBaker:
My initial thought was to be in favor of adding the Preference but having it on by default.
You do have me wondering about cases where focus on textboxes would not be the preferred behavior and... I can't think of any.

pojienie:
Thank you so much! Like everyone else has said, I would recommend make it on by default. Or even remove the setting altogether and just make it default.

Note that this PR also works on palette names. (tested under 18.04 ubuntu)

manongjohn:
I would forget the preference and just make it a default behavior. It should only lose focus once you click on something in the next pane.

papio-pl:
Thanks. I think you can turn it on by default. Right now it's very annoying;)

martinvanzijl:
Thanks for your comments. I'll remove the preference and just make it the standard behaviour, since everyone seems to prefer having it on. Will commit shortly.

martinvanzijl:
Done. Ready for review/merging.

RodneyBaker:
A very useful improvement.
LGTM
Thanks Martin

PR #3219, opened 2020-04-08
add freehand mode to control point editor tool

pojienie:
This PR allows your to use freehand mode to select control point

video: https://youtu.be/nokEyuM-rn8

Code wise, I copied a lot from selection tool and eraser tool. Freehand, rectangular, and polyline are too common for selection across tools (selection tool, eraser tool, and now control point editor tool all use almost exactly the same code for rectangular and freehand mode) so maybe they should be put to somewhere else so that they can share.

Also fixed two warnings that I've been wanting to fix for a long time.


jrhyme:
It’s scary to imagine what will happen next, the animation of the points?))) I need to drink for you)

beeheemooth:
@pojienie this is certainly wonderful, but does it make sense if we can’t undo the action on highlighting points?

RodneyBaker:
@beeheemooth Do you mean 'undo the action' as in 'invert the selection' or more strictly some form of undoing?  Not sure I'm following.  Undoing the selection of any group of control points can be accomplished simply by clicking anywhere other than another control point on the canvas.

@pojienie Thank you for adding this standardization across tools.  Very nice.
This helps not for usage... but for documentation and training too.
Testing now...

beeheemooth:
@RodneyBaker look at the video: you select points, spend time and effort on it, you have only one chance to select the necessary points at once, since you cannot use the addition or exclusion of additional points from the selection (using Shift or Ctrl). And so, when you finally selected all the points you need, you want to drag them ... oops, you clicked beside and the selection was canceled. And you need to re-select the points, since there is no operation to undo the highlighting of those points that you selected a second ago.

RodneyBaker:
Ah, the thing you speak of to me is what I use in other software as 'named
groups' where every time a selection of control points is made a temporary
group is created.  This automatically generated and temporary group will
disappear if not given a name.  Then at any time that group of control
points can be reselected, renamed, joined with other groups... because when
two named groups are selected... yet another temporary group containing the
contents of all other groups selected is created.

This approach is much better than any undoing but... also allows for that
because renaming of control point groups is an event which is entered into
event history... thereby allowing undo/redo of the event.

In my estimation the feature you allude to is one of many steps on the way
toward animation of control points.

RodneyBaker:
Ah, the thing you speak of to me is what I use in other software as 'named
groups' where every time a selection of control points is made a temporary
group is created.  This automatically generated and temporary group will
disappear if not given a name.  Then at any time that group of control
points can be reselected, renamed, joined with other groups... because when
two named groups are selected... yet another temporary group containing the
contents of all other groups selected is created.

This approach is much better than any undoing but... also allows for that
because renaming of control point groups is an event which is entered into
event history... thereby allowing undo/redo of the event.

In my estimation the feature you allude to is one of many steps on the way
toward animation of control points.

Added:  Another thing we currently seem to lack in Opentoonz is the ability to select more than one  group of control points (i.e. more than one line/stroke) and see all of the control points at the same time.  Currently we can only see/access the control points of a single line.  There is a slight exception to this in that the Snap feature is aware of the line, there just is no presentation to the user of the control points of the target spline.  Adding this capability seems a prerequisite going forward to such things as 'gluing' control points together and other merger/connectivity operations.


namedgroups

pojienie:
@jrhyme Heh, most features I have done are just "this exists in tool X, why doesn't tool Y have it?" kind of features so what I did is just glorified copy-paste.

@beeheemooth I agree that, at least, it should have a way to allow us to "select previously selected control points". Better if it allows us to name the grouped control points as @RodneyBaker mentioned.

Also, you can select individual control points in addition to what you already selected by either holding control key or right clicking on it. Although this can only be done one by one.

RodneyBaker:
Posting intent to merge.
Any additional feedback will be appreciated.

jrhyme:
@pojienie But still, no one dared to make such functions that everyone had been waiting for so long) the program is capable of much)
@RodneyBaker I think something new can be done in a separate PR.

PR #3221, opened 2020-04-09
Stop Motion Improvements

turtletooth:
This makes a number of improvements to the stop motion toolset:

- Camera settings are saved and loaded for each scene
- Added ability to export an image sequence based on an xsheet column
- Display battery level for Canon DSLR cameras
- Stop making .bak files on saving a stop motion level
- Focus check location on DSLR cameras can be controlled with arrows and numpad
- Focus check location now highlighted with a box in the viewer
- Fix some shortcuts
- Fix white frame flash after playback
- Option for the camera to be drawn below or above other levels
- Webcam stop motion for all builds
 (This will require all developers to use OpenCV from now on)
- Time lapse option for stop motion
- Frame number is sent over a serial port for motion control to arduino/rasp pi.


RodneyBaker:
Will #3093 need to be rebased if this gets merged first (or vice versa)?

turtletooth:
Most likely.

turtletooth:
Actually, I just looked, and it looks like we didn't touch the same files, so we should be good.

turtletooth:
@shun-iwasawa I saw the crash reports using EDSDK.  I am unable to reproduce the crash over here.  Any ideas?

turtletooth:
@shun-iwasawa I just made it so that the canon sdk isn't initialized unless the stop motion panel is actually open.  I'm curious if that helps.

shun-iwasawa:
> @shun-iwasawa I saw the crash reports using EDSDK. I am unable to reproduce the crash over here. Any ideas?

Neither am I. I haven't heard any other crash report related to EDSDK. I think we need more cases reported before investigating the condition. 

Anyway thank you for the improvements! I'll ask my colleague to lend me his canon camera for testing the improvement when I will go to studio next week. ( We are basically working remotely from home for social distancing now. )

Jenkins again

turtletooth:
@shun-iwasawa Thanks for testing.  Would you mind if I also use this PR to implement OpenCV support by default and make the CMake option for DSLR only?

turtletooth:
One more question. . . 
I would like to implement showing the stop motion camera underneath layers that are above it in the xsheet/timeline.  

I don't really understand the renderer very well to implement this.  Any ideas on this?

shun-iwasawa:
@turtletooth

>@shun-iwasawa Thanks for testing. Would you mind if I also use this PR to implement OpenCV support by default and make the CMake option for DSLR only?

Thank you, that really helps. Please do! 

>I would like to implement showing the stop motion camera underneath layers that are above it in the xsheet/timeline.
>I don't really understand the renderer very well to implement this. Any ideas on this?

Currently the stopmotion image seems to be [added](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/sceneviewer.cpp#L1987) after [visiting all columns](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/sceneviewer.cpp#L1966) in the scene. I think inserting the stop motion layer node at the proper position instead of [appending it to the end of nodes](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonzlib/stagevisitor.cpp#L1038-L1041) will probably work.

turtletooth:
> Currently the stopmotion image seems to be added after visiting all columns in the scene. I think inserting the stop motion layer node at the proper position instead of appending it to the end of nodes will probably work.

That's the part that I don't quite get yet.  There seems to be a lot of magic that happens on Stage::visit that I'm not sure where I would try and put my code.

I don't want to bother you with it, but if you had any quick pointers, that would help.

turtletooth:
@shun-iwasawa I made OpenCV and Stop Motion available at all times.  There is a new flag WITH_CANON for DSLR support.

I tested the camera capture window and it works very nicely with OpenCV.  

What do you want me to do with the _Qt versions of penciltestpopup and cameracapture . . .?

I will need some extra help making the mac version work since I don't have access to a mac anymore.  The pieces are all there, however.

artisteacher:
Installing opencv breaks building OpenToonz altogether on macOS.
Screen Shot 2020-04-09 at 10 39 19 PM
The same errors were previously reported in #2496, #1566, and #40. @yehjaemin had a fix which didn't end up being merged #2499. I was able to build the master successfully with that change. I've been attempting to build from the mac stop motion PR. For some reason, EDSDK.h is currently showing errors that prevent a successful build. I'll make another attempt with this PR tomorrow.

turtletooth:
@artisteacher I would try and build this branch without the WITH_CANON option first, to simplify getting the other 2 pieces working first.  This branch can do webcam stop motion without any options checked now.

Once you can get a successful build without Canon, then we can try and get Canon working.

manongjohn:
In #2835 I disabled LZMA support for Travis builds due to those errors. Is LZMA needed for Cannon SDK? If not, need to disable that from XCode builds as well.

manongjohn:
LZMA is used when building our custom libtiff.a. 

Seems we should update the macOS build instructions to disable LZMA support as we do in Travis builds.

It should be changed to this: 

```
$ cd ../tiff-4.0.3
$ ./configure --disable-lzma && make
```

@artisteacher Try rebuilding your libtiff.a as above and then rebuilding with Xcode.

artisteacher:
Disabling LZMA support seems to work. Build made it 91% but fails at penciltestpopup.cpp. 
[pencil_test_errors.txt](https://github.com/opentoonz/opentoonz/files/4462354/pencil_test_errors.txt)
Should the opencv version matter? Homebrew installs 4.3.0.

turtletooth:
@artisteacher 
That code should only be for Windows machines.

Can you put an #if WIN32 after line 1927
`int PencilTestPopup::translateIndex(int camIndex) {
#if WIN32`

and an #endif and return after 2002
`
CLEAN_ATTRIBUTES()
#endif
return camIndex;
`

also around 2016, we need to add some if:
`
#if WIN32
    if (!m_useDirectShow) {
      // the webcam order obtained from Qt isn't always the same order as
      // the one obtained from OpenCV without DirectShow
      m_cvWebcam.open(translateIndex(camIndex));
    } else {
      m_cvWebcam.open(camIndex, cv::CAP_DSHOW);
    }
#else
    m_cvWebcam.open(translateIndex(camIndex));
#endif
`

That should take care of the Windows only code.

artisteacher:
Thanks @turtletooth, that helps. Next question, libjpeg-turbo can be installed through homebrew, but then it ends up in usr/local/Cellar/jpeg-turbo. Should we update the cmake paths to look for the system install location over thirdparty?

For now I've followed the Windows instructions and copied the lib and include folders to the thirdparty/libjepeg-turbo64 location. The build is failing at the last minute since it can't find `thirdparty/libjpeg-turbo64/lib/turbojpeg.lib`
Screen Shot 2020-04-10 at 3 34 30 PM

turtletooth:
That would probably be fine @artisteacher .  I'm also not sure if lib files are still called lib on Mac.  Feel free to try new things in CMake.

manongjohn:
I know this may be asking quite a bit, but is it possible to separate the piece that makes OpenCV and Stop Motion available all the time so people with webcams can use it? Any Canon specific changes/improvements could be in another PR.

If you can I might be able to help work out any necessary macOS/Linux Travis build scripts changes, which also helps with fixing the macOS/Linux manual build instructions.

turtletooth:
@manongjohn It is separated as a build option in this PR.  There is a new WITH_CANON build flag in CMake.  This can be built without the Canon SDK.

manongjohn:
@turtletooth I have changes to macOS build script, CMakeLists.txt files and some code fixes for Travis builds and manual builds for stopmotion without Canon support.  The changes were to use the brew installed jpeg-turbo library.  I had to make a few code fixes due to some build issues and Stop Motion Controls crash I encountered. I will be submitting a PR to your branch with my changes at some point, but I'm having a little issue with the controller.

I can open the Stop Motion Controller, but when I select my webcam as my camera, it doesn't turn it on.  Start Live View doesn't turn it on either.  As a gauge, the Camera Capture popup does turn on my camera as soon as I select my webcam.  Am I doing something wrong?  Not sure where to look to see why it's not starting my webcam.  Any suggestions on where I should start looking?

Once I verify the macOS builds are working, I'll switch to Linux builds to make sure it I can get the controller to open but I won't be able to test the webcam  since I can't get my webcam to work properly through my Linux VM.

turtletooth:
@manongjohn Are you building on a Mac or a PC?

manongjohn:
Mac VM

turtletooth:
That's weird.  I just looked and @shun-iwasawa is using more or less the same code on penciltestpopup as I'm using in stopmotion.cpp.

Are you able to set breakpoints and see if getWebcamImage is being hit?

manongjohn:
I can debug through Xcode, though I find it slow for me, probably because it's a VM.

I will attempt to see what's wrong since you have given me something to look for.

turtletooth:
@manongjohn I figured it out.  I fixed the code on penciltestpopup for this but not stopmotion.

the inside of initWebcam on stopMotion.cpp should be:

```
#if WIN32
  if (!m_useDirectShow) {
    // the webcam order obtained from Qt isn't always the same order as
    // the one obtained from OpenCV without DirectShow
    translateIndex(index);
    m_cvWebcam.open(m_webcamIndex);
  } else {
    m_webcamIndex = index;
    m_cvWebcam.open(m_webcamIndex, cv::CAP_DSHOW);
  }
  if (m_cvWebcam.isOpened() == false) {
    return false;
  }

#else
    m_cvWebcam.open(index);
#endif
  return true;
```

manongjohn:
That worked. Thanks.  I'll include it with my changes and will be submitting a PR to you shortly

turtletooth:
Thanks a ton for helping to troubleshoot and get this working on mac.  I'm sure @artisteacher will be thrilled.

artisteacher:
Very true! I can't say how much I appreciate you both!

manongjohn:
@artisteacher  No prob. Since I don't have the Canon camera or the SDK, you'll still need to build and test with Canon SDK but at least this part is 1 less thing you need to work out.

Plus I get to make sure it builds properly for OTX builds :)

@turtletooth PR submitted against your branch for macOS builds.  I'll work on the Linux (Ubuntu) build changes and submit a PR for that as well.

Are you able to work the Appveyor build failure? I'm assuming you are going to add OpenCV stuff + license to the thirdparty directory and modify CMakeLists.txt accordingly

turtletooth:
@manongjohn Thanks.  I merged the PR.

I can take a look at Appveyor.

artisteacher:
So building without the SDK works just fine. Thanks @manongjohn!

With the SDK, it is failing on EDSDK.H on line 27. For some reason Xcode is ignoring `#ifdef --MACOS--` and trying to load windows.h.

artisteacher:
And there is a crash when using the camera capture feature with the built in camera.
Screen Shot 2020-04-11 at 5 48 57 PM

manongjohn:
For the `#ifdef --MACOS--`, you will need to probably update the toonz\sources\CMakeLists.txt and look for where it `add_definitions()` under MacOS builds (2 places) and  add `-D--MACOS--` to the list.  You'll need to rebuild the build folder.  If that works, we can add it to the build scripts as part of this PR.

Not sure about that other issue.

manongjohn:
Built a version of OTX with this PR included and tested on a different macOS VM.  It crashed immediately when it tried to open.

I discovered that the `macdeployqt` program pulled in the opencv and jpeg-turbo *.dylib files into the `OpenToonz.app/Contents/Frameworks` folder as expected, however some of those still contained references to other brew'd libraries in `/usr/local/Cellar` when it should have been changed to look within the Frameworks folder.  Not sure how the nightly/official builds are created and work with the installers, but might run into the same issue.  Not sure how to fix this yet.

turtletooth:
I found the crash on capture - at least on Windows.  jpeg-turbo includes its own copy of jpeglib.h.  I had to rename that file to make OT use the correct version for saving levels.  

I think that eventually we could get rid of lib-jpeg that we use now and move everything over to jpeg-turbo.

artisteacher:
It includes jpeglib.h on macOS too. The crash on my end is happening as soon as I try to select the built in camera. Probably the new security settings they added to macOS.

Updating cmakelists.txt allowed the build to go further. Now I'm seeing a couple of repeated errors on stopmotion.cpp.
1. Expected ';' at end of declaration - lines 233, 3474
2. Use of undeclared identifier mySize - lines 236, 243, 268, 3477, 3486, 3543, 3564
Screen Shot 2020-04-11 at 6 49 35 PM

artisteacher:
I was able to edit info.plist in Xcode so that I can access the built in camera. When selecting the built in camera for the first time, a notification pops up so permission can be granted to use the camera. The camera turns on (the light is on), but nothing shows up in the capture window - it stays black & says that the camera in unavailable. Pressing capture does nothing. Closing out of the camera capture window & opening it up again does work for some reason.

turtletooth:
Can you try putting this before the "expected ;" lines:

```
#ifdef MACOSX
    typedef long long __int64;
#endif
```

turtletooth:
Does capturing still cause a crash?

artisteacher:
So as long as I build with:
`<key>NSCameraUsageDescription</key>`
`<string>OpenToonz needs access to the camera in order to use Camera Capture</string>`
in toonz/cmake/BundleInfo.plist.in - there's no crash on my end. The camera works from the start, once I grant it permission. The only issue I've found is that using the refresh button stops the built-in camera from working. It only works again after closing/re-opening the camera capture window.

artisteacher:
```
#ifdef MACOSX
    typedef long long __int64;
#endif
```
I tried adding this in, but I'm still seeing the same errors.

turtletooth:
How about this:

```
#ifdef MACOSX
    UInt64 mySize = 0;
#else
  unsigned __int64 mySize = 0;
#endif
```

artisteacher:
That works! Next problem, it can't find thirdparty/canon/Library/EDSDK.lib. There wasn't a library folder in the SDK - there are Framework & Header folders.

turtletooth:
Almost there. . .
I don't think the file it wants is called edsdk.lib on Mac
What are the files in the Framework folder?

artisteacher:
Inside of Framework is a directory called EDSDK.framework. Inside of EDSDK.framework is a symlink to EDSDK, a versions directory and a symlink to a resources directory.

turtletooth:
Can we try this?

can you rename your canon folder in thirdparty to EDSDK

on line 284 of CMakeLists.txt change canon to EDSDK

and then  change line 411 to:
```
    pkg_check_modules(EDSDK REQUIRED edsdk)
    find_library(EDSDK_LIB edsdk ${EDSDK_LIBRARY_DIRS})
    message("**************** edsdk lib:" ${EDSDK_LIB})
```

This is a just a guess.

artisteacher:
Doesn't work. "A required package was not found".

turtletooth:
Another option is to copy the contents of the Framework directory to the Library directory in thirdparty/Canon

and change line 411 to 
```
set(CANON_LIB ${SDKROOT}/canon/library/EDSDK.framework)
```

turtletooth:
If that doesn't work, is there a section of xcode called Groups&Files?

The instructions on the Canon manual say:
"Add EDSDK.framework to Groups&Files"

artisteacher:
That builds all the way but crashes on trying to open.
Screen Shot 2020-04-11 at 10 21 57 PM

turtletooth:
It builds.  Now we just need the right file to be found at runtime.

turtletooth:
The manual says:

> Place EDSDK.framework in an application directory such as Contents/frameworks/. It is also possible to load “EDSDK.framework” as a source file. 

Can you place EDSDK.framework in the folder that OpenToonz runs in?

artisteacher:
Putting EDSDK.framework in the same folder does help it run. The macOS security settings are a hassle though. When I first try to open OpenToonz, I get:
Screen Shot 2020-04-11 at 10 47 55 PM
Going to Preferences -> Security & Privacy -> General, this can be overridden by clicking "Open Anyway":
Screen Shot 2020-04-11 at 10 52 03 PM
Afterwards, trying to open again, the first warning about "can't be checked for malicious software" comes up again but there's a 3rd option to open.

The stop motion controls work great so far! It recognized my camera without issue.

turtletooth:
Yes!!!
That is awesome!  So glad.  Thanks for sticking with it.  I really appreciate it.

I wonder if there is a way around the warnings on Mac.

Would you be willing to make a PR against my branch to get the changes you made included?

artisteacher:
It is really fantastic! 

I hope so. It's excessive with the SDK added.

Should we try to get the path issue resolved so that EDSDK.framework doesn't need to be added to the app's parent folder?

turtletooth:
For the mac build instructions, I don't think there is a way around that. The file/folder is needed for the program to run.  On Windows, there is a file that is required to be moved to the OpenToonz directory for it to work too.

But for automatic builds,though, I think that can be worked into the build script.

artisteacher:
Overall this is working great, though there's something funny going on with frame 1. When working in a new scene, live view is only visible on the first frame if "Show Live View on All Frames" is checked. When you press capture for the first time, it sounds like a photo is taken, but nothing get placed in the Xsheet. Pressing capture again, the image for frame 2 gets exposed on the sheet for frame 1. Looking in the extras folder, the first image is there. If you press the right arrow button for the next frame (next to the current frame - ie. 1) before taking the first picture, live view starts working as expected and the first photo is exposed properly.

turtletooth:
I just want to make sure we are seeing the same issue.  

When I first open OT, the camera behaves as expected.  But if I create a new scene after opening, I see what you are seeing.  Hitting the "Current Frame" button fixes it and gets the stop motion camera placed on frame 1 of the xsheet with the green indicator.

Is this the case for you?

artisteacher:
Yes, I think so. It seems to happen for any scene (new or not) that has not previously captured images with the stop motion feature. Hitting either the current frame or next frame button fixes it.

turtletooth:
K- I think I fixed it.

manongjohn:
Likely going to have issues with automating Linux Ubuntu builds for this because we'd need to build openCV 4 from source and that requires a newer version of cmake to build. I'd also need to build cmake from source to do that.  The packages for these in the Ubuntu 14.04 and 16.04 repos are older versions.  It will certainly increase the build time.

Not sure how other Linux flavors will fair.

turtletooth:
We can remove linux from the supported stop motion platforms if we can't figure it out.
@pojienie Are you using linux for your work?  Do you have any insights?

manongjohn:
Currently working out the steps for downloading and compiling cmake/opencv for the travis-install and build scripts.  It's just not very pretty.  I can only provide steps for Ubuntu since that is what Travis builds with.  Other Linux flavors will likely need to do similar steps.

turtletooth:
@artisteacher I submitted a fix for the issue you pointed out, and changed the controls for DSLR cameras to sliders so they are easier to work with.

Please let me know if you see any other issues so I can deal with them quickly.

Thanks for testing this and taking the time to get it working on mac.

artisteacher:
@turtletooth - I'm happy to help! I've put my mac build changes in a pull request. I updated the mac build directions while I was at it.

The frame 1 issue is fixed. The sliders work well. 

Running OT through Xcode, I see this failed assertion when capturing the first photo in a new level:
`2020-04-13 21:03:39.106898-0400 OpenToonz[84510:10075229] [Framework] FSEventStreamFlushSync(): failed assertion '(SInt64)last_id > 0LL'
2020-04-13 21:03:39.107162-0400 OpenToonz[84510:10075229] [Framework] FSEventStreamFlushSync(): failed assertion '(SInt64)last_id > 0LL'`
It hasn't seemed to impact functionality so far.

Running through Xcode, if the camera falls asleep, OT hangs. Without Xcode, if the camera falls asleep it often crashes.
[sleep_crash.txt](https://github.com/opentoonz/opentoonz/files/4472713/sleep_crash.txt)

If the camera is asleep when OT starts, it doesn't crash or hang, but the controller doesn't recognize the camera is there.

What exactly should the refresh button do? All it seems to do on my end is deselect the active camera.

Focus check is really out of sync with the "pick focus" selection box - it seems to be selecting a spot somewhere below.
Screen Shot 2020-04-13 at 9 54 31 PMScreen Shot 2020-04-13 at 9 54 08 PM

General nitpicky feedback 🙃:
1. The stop motion controller window might benefit from a minimum width or horizontal scrolling
2. It would be nice to have a checkbox to enable/disable the images/live view being flipped. Working with an animation stand, the current functionality would be perfect but not when using a tripod for claymation or etc.
3. In my opinion, the behavior of the Xsheet frame field and buttons would be better if they allowed for directly navigating the Xsheet. Ie. typing frame 14 and pressing enter would switch to frame 14, making that the active frame. Likewise the left and right arrows would change the active frame forwards or backwards. Similarly, changing the frame in the Xsheet would change the Xsheet frame number in the controller. Then the current frame button wouldn't be necessary.
4. The focus check button should have an off and an on state so that there is a visual indicator showing if it is on or not
5. The pick focus behavior would benefit from refinement. I'm not sure of the best behavior but it could use a visual indicator that the focal area has changed. Perhaps "pick focus" should have an off/off state like the check focus button. Then after selecting the the new area of focus, the rectangle would move to that point.

turtletooth:
Hey @artisteacher Thanks for the feedback!  A couple clarifying questions:

Is the image flipped by default in your viewer?  On Windows, I had to flip it over, I think.

It looks almost like the pick frame rectangle spot is flipped vertically from the actual zoom location.  Can you let me know if that seems to be the case?

If so, those seem like easy fixes.

For number 3, I agree with the first part.  As for setting the number by navigating the xsheet, how would that work with reviewing frames?

I agree with 4 and 5.

Quick note on focus check picking, you can select a new location by either clicking or using the arrow keys to move the box.  You can get out of focus check mode currently by clicking a location, hitting enter or hitting escape.  I can make clicking not turn it off however.

turtletooth:
The refresh button should update the camera list if you plug in a camera while OT is already open.
I can make it not disconnect a camera if it is already activated.

Canon cameras should automatically show up in the list even if they are plugged in/woken up with OT already going.

artisteacher:
> Is the image flipped by default in your viewer?

It is flipped by default. I thought it was deliberate!


> It looks almost like the pick frame rectangle spot is flipped vertically from the actual zoom location. 

This seems to be the case. It's probably showing what would be picked if the view was flipped.

> As for setting the number by navigating the xsheet, how would that work with reviewing frames?

I see your point.

>  using the arrow keys to move the box

Good to know. I tried to click and drag the box to a new position at first. Obviously that didn't go so well!

turtletooth:
Just to clarify - is it flipped only vertically, or flipped in both directions.

artisteacher:
Just vertically.

artisteacher:
Also, shutting down the camera while it is active in OT causes a crash.

turtletooth:
Ok, I think I fixed the flip and the crash.

turtletooth:
Good news.  Appveyor builds now and the Mac side of Travis builds too.  I'm not sure what to do with the linux side of Travis.  It looks like Travis doesn't have an easy way to pull in OpenCV, so I'm wondering if it would time out by having to build it from source, and then build OT.  

@pojienie Do you have any insights about building and getting this working on Linux?  The Canon DSLR support won't work, but stop motion with webcams could.

manongjohn:
Still working on Linux scripts.

Building OpenCV from source does take a lot of time unfortunately. Not sure what the timeout limit is and if it would be any faster than my VM.

Can stop motion without Canon support work with an earlier version of OpenCV? I can pull openCV 2.4 without compiling which would make building much simpler.

Also running into issues with cmake detecting libjpeg-turbo8. Says its installed but I don't see it. Took me a bit to get it right for macOS builds, so not entirely surprised.

turtletooth:
@artisteacher I think I took care of all(?) of the issues you mentioned.

There is a visual indicator now for both focus check and pick focus check
There is a min width to the panel
You can drag the pick focus check around
I think I took care of the crash
I actually made it so the camera doesn't go to sleep, but simply turns off live view if the camera hasn't been used in a while.
Changing the xsheet number in the panel changes the active frame in the xsheet.

Please let me know if you see any other issues.

turtletooth:
I've got basic messages sent to an Arduino working for motion control.

manongjohn:
@turtletooth FYI...there are a number of places where you are doing a `#if WITH_CANON`  or `#if WIN32` and it should really be `#ifdef WITH_CANON` or `#ifdef WIN32`

turtletooth:
Thanks @manongjohn I'm changing them as I come across them.

turtletooth:
This is ready for testing.

manongjohn:
@turtletooth FYI, I submitted a PR against your branch to fix the Linux Travis builds.

turtletooth:
Awesome. Thanks so much for that!

turtletooth:
@manongjohn That did it!  You're amazing.  Thanks!

manongjohn:
Thanks.  The stuff you did here is far more impressive.  Nice work!

I just need to iron out an issue with the macOS packaging for OTX builds then I can try playing around with it more in OTX.

turtletooth:
Just a heads up- to enable motion control for motors through arduino, I added QSerialPort as a requirement.  I need to update the Windows build instructions for this.

manongjohn:
@shun-iwasawa Please look at the Jenkins builds on this PR as well.  May or may not be related to the other PR where they failed.

artisteacher:
So I built with the latest changes this morning & ran into a few issues.
1. My DSLR Canon SL2 is no longer recognized (a webcam & the native camera both worked)
2. If you press capture too quickly after the previous capture, this warning pops up:
Screen Shot 2020-04-22 at 10 46 36 AM
3. If live view is left on when switching cameras, it needs to be turned off then on again for the new camera

turtletooth:
Thanks for the info.

1.  Oh no!  Does your camera show up in the drop down still?
2. Live view pauses while processing an image.  But the popup would be annoying.  I can fix that.
3. I can fix that.

#1 is the biggest issue for me.  I don't want to have a regression.  I'll investigate right away.

turtletooth:
@artisteacher Silly question, but did you make sure the with_canon flag was checked in the build?

artisteacher:
I probably didn't!

artisteacher:
Ok, I obviously wasn't awake enough earlier. It finds the camera just fine now. Last couple issues:
1. The live preview shows the correct orientation of the camera. The actual photos are upside down.
2. After navigating directly in the Xsheet with live view on, it needs to be turned off and on again when returning to the current stop motion frame for live view to work.

turtletooth:
Phew!  Glad the camera is recognized.  I'll look into the issues today.

turtletooth:
@artisteacher I think I fixed them all.  I accidentally introduced a couple bugs working on pausing live view, but I think they are worked out.

Please let me know if you find anything else.

manongjohn:
I noticed with the Stop Motion Controller that the Opacity control is fixed to the bottom of the  box and is independent of the Controls or Options tab.  Is that intentional or is it supposed to be on one or the other?

The Options tab panel doesn't allow you to scroll down to get to the controls towards the bottom when the window is vertically compressed.  This would be bad for smaller screens especially when docked in a small space.  As it is, with the default height of the popup, I didn't know there were additional controls below Time Lapse because of where the Opacity controls covered the other stuff just right.

Are the DSLR Options supposed to be enabled even for Webcam users?

turtletooth:
@manongjohn 
The opacity slider is intentionally always visible, as adjusting it is pretty common in stop motion.

I'll look at enabling scroll bars on the tab pages if needed.

The DSLR options ideally are hidden to start with, but when I tried to hide them at the start, the wrong tab page would open clicking the options tab.  I'll take a look again.

turtletooth:
@manongjohn Thanks for catching those.  FIxed.

manongjohn:
@shun-iwasawa Can you please look into why Jenkins is failing on this PR?

artisteacher:
I built today and live view has stopped working. It just shows a white canvas or frames from other levels. This is both for my canon sl2 and for the built-in camera. When you take a photo, it flashes in the viewer for a moment then returns to the white canvas or the active frame from other visible levels.

turtletooth:
I add a commit last night to get the camera working in the proper order in the xsheet.  But there was an option that I forgot to account for.  Would you mind trying it now?

artisteacher:
That works! There is an issue with the light setting. If you try to take a capture with "Screen 1" checked, the monitor displays the selected color, while the picture is taken. Once the image is captured, the screen changes to black. I can't exit without toggling the main window's full screen mode.

On a related note, I was hoping OT would recognize a second display that's connected to my laptop through apple airplay. No such luck.

turtletooth:
This looks like a mac specific issue.

In stopmotionlight.cpp line 121, can you change it to:
```
m_fullScreen1->showMaximized();
```

turtletooth:
Another thing to try would be to insert 
```
m_fullScreen1->showNormal();
```
before the hide on line 149

artisteacher:
The first change has no impact.

The second option helps. The color window shrinks down & then can be closed:
Screen Shot 2020-05-04 at 3 59 57 PM
Auto-closing (or hiding?) would be nicer though.

So, the second display (using airplay) seems to work, as long as it is already connected before OT is opened. The second display is not recognized if it's connected after OT is open.

If the second display is turned on for the color feature, the second display switches to black after the display of the color with the camera capture. Except when the light feature is active, the black screen stays until OT closes. Additionally, disconnecting & reconnecting the second display while OT is open seems to cause some graphics issues when using the light feature.

turtletooth:
Can you try changing the hide at line 149 or 150 to close?

Right now I'm only having you test the first monitor to see if we can get it working.  The code for the closing the other 2 monitors is right below that.

artisteacher:
That behaves the same.

turtletooth:
Hmm. . . how about accept() ?

turtletooth:
Btw. . . is live view picking showing in the right place now?  And are the inverted images fixed?

artisteacher:
Nope.

Live view picking is correct. Inverted images are fixed.

turtletooth:
That's tricky.  I can close it with hide, close or accept.  
So show normal makes it a normal window, but nothing closes it?

turtletooth:
Can you add 
```
m_fullScreen1->windowHandle()->close();
```
after the line that tells it to hide, close or accept (whichever one you have in there right now)

Thanks for helping figure this out.

turtletooth:
That will also need #include <QWindow> at the top - line 9

artisteacher:
That works! I used `#include <QWindow>` for line 9. I copied the same code for multiple monitors. A second display works perfectly. I assume a third would as well.
Screen Shot 2020-05-04 at 9 12 52 PM
Though I just tried commenting out `m_fullScreen1->close();` & it worked fine without that line. So lines 150, 155 & 160 seem to be unnecessary on macOS.

The light options all work perfectly now. I like the addition of the red dot to show the active camera frame. Time lapse works really well.

I do still sometimes get a crash when the camera turns off or is unplugged.
[canon_crash.txt](https://github.com/opentoonz/opentoonz/files/4578561/canon_crash.txt)

artisteacher:
Another thing, hovering over the the test button, triggers the button. I'm just using a touchpad - I don't have a mouse.

turtletooth:
Ok, I added the code for the lights, and I think I fixed the test button on hover.

Regarding the crashes, I think that is also mac specific.  I tried a bunch of unplugging and turning off mid live view set.  I do have OpenToonz configured to keep the camera awake, so it shouldn't be an issue if people leave the camera alone.  

If you can run xcode in debug mode and see if it shows where it crashes or send me a crash report, I can try and figure out where it is crashing.

Thanks for all your help on this.  I think in the end OpenToonz is going to be a pretty top notch stop motion software.

turtletooth:
I tried something that might fix the crash on mac.  When you have a chance, could you try and crash it again?

manongjohn:
I'm looking to add this to OTX's May update for webcam users.  Is it stable enough or should I wait?

I'm also getting this when I pull your branch into mine: 
```
From https://github.com/turtletooth/opentoonz
 * branch                stop_motion_improvements -> FETCH_HEAD
   17bac0a49..f14288ce7  stop_motion_improvements -> turtletooth/stop_motion_improvements
Downloading thirdparty/libjpeg-turbo64/dist/turbojpeg.dll (724 KB)
Error downloading object: thirdparty/libjpeg-turbo64/dist/turbojpeg.dll (d9f87d7): Smudge error: Error downloading thirdparty/libjpeg-turbo64/dist/turbojpeg.dll (d9f87d770b205312635a9d87ce862d0fa6a00b9f74fd9fb09449c38bd37cb2dd): [d9f87d770b205312635a9d87ce862d0fa6a00b9f74fd9fb09449c38bd37cb2dd] Object does not exist on the server: [404] Object does not exist on the server
```

Any idea why?

turtletooth:
I just did a lfs push.  Can you try now?

I do think it's stable enough right now.  There will probably be a couple of unforeseen bugs, but I can knock them out quickly.

artisteacher:
If I unplug the camera cord then plug it back in quickly, that seems to always trigger a crash. This is what I get in Xcode:
Screen Shot 2020-05-05 at 12 19 39 PM

turtletooth:
Hmm. . . that crash didn't really help.  Crashing on main.cpp is hard to debug.
Does it still crash on turning the camera off or unplugging it?

artisteacher:
Also, the test button works perfectly now - no weird hover issue. 

Is line 165 on stopmotionlight.cpp a typo? I thought it should be `m_fullScreen3->showNormal();`. It might not matter, but lines 153, 160, and 167 (the first close occurrences) don't seem to be necessary on macOS.

artisteacher:
It does still crash with both. The crash reports look the same:
[canon_crash2.txt](https://github.com/opentoonz/opentoonz/files/4582956/canon_crash2.txt)

turtletooth:
Thanks for checking.  I'm not sure about what could be causing that.  Does everything else seem to be working ok?

artisteacher:
Everything else seems to work just fine so far.

manongjohn:
See if you can get Jenkins to start on this one, unless it starts on my comment this time

turtletooth:
You did it.

manongjohn:
lol.  it's definitely hit or miss when I do it. :)   Need @shun-iwasawa to see why the Je.kins builds are failing. Likely because it needs some stuff installed on the build machines?

I did manage to get this built into the OTX May update, though the Travis macOS build was crashing on startup when I tested it in Catalina.  Had to replace the macOS version with a manual build that would work.

manongjohn:
FYI, a reddit user posted that they crash on startup.  Windows 10 error log says the faulting module was EDSDK.dll.  The odd thing about it is the user was able to open OT a week ago, briefly looked at the canon stuff, but hadn't used a camera then closed OT (at some point).  When the user tried to open OT today, it crashed in EDSDK.dll.

turtletooth:
That's not the first one I've heard of that.  I wish I could duplicate it.

manongjohn:
Update from the reddit user regarding the dll crash

```
Turned out EDSDK.dll was also a part of Canon EOS Utility and other associated Canon apps, and that they were also running in the background. I uninstalled them since I don't really have any use for them anyway and I figured this was the easiest way to solve it. This solved the EDSDK problem. I wonder if this could be of any use to the other user? Thanks for the help
```

turtletooth:
I think I'm going to be branching off and making a stop motion focused version of OpenToonz.  Do you want me to leave the Canon stuff in here or take it out if I'm not going to be around to support it.  All the webcam stuff will still work.

gab3d:
I think it would be better to keep it integrated.
What reasons lead you to that conclusion?
It'd be sad to see OT forked...

shun-iwasawa:
@turtletooth 

The recent failures of Jen.kins were due to missing of libjpeg-turbo files ( `jerror.h turbojpeg.h and turbojpeg.lib` ). I found that when once Jen.kins run on this PR, it removes libjpeg-turbo files when pulling another PR later on. It seems to be because this PR firstly starts to record these files in git.

I try modifying Jen.kins script, adding lines to copy the libjpeg-turbo files if they are missing before starting build. So let’s wait and see :-)

manongjohn:
Closed? This PR was supposed to add the ability for webcam users to use the Stop Motion feature. Any way you can separate out just that piece and strip out any Canon specific changes?

turtletooth:
The webcam stuff works already.  As long as the Canon flag isn't used in CMake, it will build as is.

turtletooth:
I don't mind leaving this open, but I would like to get it merged soon.

shun-iwasawa:
Jenkins (for testing Windows build)

turtletooth:
J*nkins is still failing.

shun-iwasawa:
@turtletooth 
Jen.kins Windows build seems to fail because it cannot find 32bit version of OpenCV library when building 32bit module for QuickTime I/O (`t32bitsrv`) .
Is it possible to make it to use OpenCV only when built in 64bit?

turtletooth:
I'm sorry I can't do that right now.

turtletooth:
Closing until someone is able to fix the 32 bit issue.

PR #3222, opened 2020-04-09
Fix Bokeh Fxs to Render Iris Pattern Properly

shun-iwasawa:
This PR fixes several problems in `Bokeh Iwa` and `BokehRef Iwa` fxs as follows:
- The rendered iris pattern sometimes becomes chipped at the right and the top edges.
- The pattern sometimes offset by 0.5 pixel.
- The source image sometimes offset by 0.5 pixel, which can cause noticeable change in brightness especially when blurring small highlights.

Here is an example, applying the `Bokeh Iwa` fx to a single white dot at center with animating the bokeh amount:

Before the fix:

After the fix:


RodneyBaker:
Thank you Shun!
LGTM

PR #3226, opened 2020-04-12
Fix multiple warnings

rozhuk-im:
Fix simple C warnings  during build on FreeBSD with clang and gcc.

### DIFF
--- a/toonz/sources/common/expressions/tgrammar.cpp
+++ b/toonz/sources/common/expressions/tgrammar.cpp
@@ -538,7 +538,7 @@ public:
   bool matchToken(const std::vector<Token> &previousTokens,
                   const Token &token) const override {
     int i = (int)previousTokens.size();
-    return i == 1 && token.getText() == "?" || i == 3 && token.getText() == ":";
+    return ((i == 1 && token.getText() == "?") || (i == 3 && token.getText() == ":"));
   }
   bool isFinished(const std::vector<Token> &previousTokens,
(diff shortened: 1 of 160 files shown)


RodneyBaker:
@manongjohn Could you add this to the next OTX build?
I know OTX isn't building for FreeBSD but there are so many changes in this PR it would be good to have additional validation on other OSs that there are no adverse effects.  
Everything looks good here I just don't want to see anything get missed.

rozhuk-im:
This is not FreeBSD specific changes (except malloc.h replace) - I can build #3231 without this, after change ```#include <malloc.h>``` in few files.

There is many changes in ```if()``` conditions, that clang show as suspect to wrong.
Also this includes: https://github.com/opentoonz/opentoonz/pull/925 - it was from gcc warnings/suggestions.

Please review this PR before merge.

RodneyBaker:
In testing thus far I've yet to see something that doesn't work and have seen some improvements.
Some of which I can't quite explain.
For instance, with this PR inserting a Color Emboss FX pipes the flow of the schematic correctly into the xsheet node whereas the regular release of Opentoonz does not... leaving the user to connect the node manually.  

Other minor improvements noted as well.

RodneyBaker:
After running through the 160 files effected by this PR multiple times...
Posting intent to merge this.

#925 should no longer be needed although specific instances of abs/fabs may need to be reviewed on case by case basis.  The majority of cases are handled by std::abs.

Issue #1717, opened 2018-01-05
RGB Picker Tool side effect of zoom/skew


beeheemooth:
At first I thought that this thing happens only in OTX, but it turned out that in the Morevna Edition the same bug.
Try to play this bug in OT 1.2 (I have not installed it):
1.Use some non-standard vector fill.
2. Switch to the RGB Piker Tool.
3. For some reason now, this tool has the ability to zoom, which can deform even the camera.
![ezgif-3-a9859e5d2e](https://user-images.githubusercontent.com/25724133/34609912-51f8a206-f227-11e7-8331-22803ade93d8.gif)

artisteacher:
This also happens in 1.2. For this to occur, both of the gradient colors have to be changed from the default. After this occurs, using reset view and fit to window do not fix the extreme zoom or angle distortion for the viewer/camera. Resetting the center of the camera in the schematic does fix the zoom - but not the angle distortion.

After changing to other vector styles & back to the gradient, I can’t seem to trigger the effect. This may not be related but at the same time that the zoom effect stopped, the gradient fill started to display properly in the viewer (in OS X several of the vector styles overflow #891) though they continue to display with the overflow in preview/render and in the drawing icons.

RodneyBaker:
I accidentally ran into this and thought the ability to skew the camera view was very cool.
Later, I wasn't able to get the behavior to repeat but that is very likely because I was no longer using the Color Pick Tool.

beeheemooth:
After "Delete Lines, Styles and Areas" (in this case with a problem style) the behavior of the tools becomes inadequate even if a new scene is created. Only restarting OpenToonz solves this problem.
![ezgif-2-b5297fcfa0](https://user-images.githubusercontent.com/25724133/37691124-446b688c-2cb8-11e8-877b-b820e8d7fd33.gif)

DarrenTAnims:
I've just stumbled across this too in V1.4. The above all describe the issue well. The only thing I've got to add is that it only messes up the viewer in the current room. Change to another room and until you change the colour, the viewer works fine.

ladlon:
Hi.  I was directed to this thread when I posted my own bug report today about this same issue.

I had a scene where I had a character (multi-level, parented), and a photographic background image.
[...]

PR #3230, opened 2020-04-13
fix hook problem with vector brush tool

pojienie:
related issue: #2526 #2861 

The "hook" problem is that, when you use vector brush tool to draw quickly with high smooth value, there will be a "hook" at the start of the stroke.

a

The problem no longer exists on my end when I applied this fix. However, I don't have a drawing tablet to confirm that it won't happen if you are using it.

Explanation: If I didn't misunderstand [this article](http://www.mvps.org/directx/articles/catmull/), CatmullRomInterpolate(p0, p1, p2, p3, samples) function generates `samples` points between p1 and p2 and thus all points generated should be inserted between p1 and p2. However in

https://github.com/opentoonz/opentoonz/blob/f49503f713d2942e4afa418c2e1f3b56085ddd78/toonz/sources/tnztools/toonzrasterbrushtool.cpp#L702-L705

where

* p0 = points[0]
* p1 = points[0]
* p2 = points[1]
* p3 = points[2]

the code inserts generated points BEFORE p1 and I assume that's the reason that caused the hook problem.

### DIFF
--- a/toonz/sources/tnztools/toonzrasterbrushtool.cpp
+++ b/toonz/sources/tnztools/toonzrasterbrushtool.cpp
@@ -701,7 +701,7 @@ static void Smooth(std::vector<TThickPoint> &points, int radius) {
   if (points.size() >= 3) {
     std::vector<TThickPoint> pts;
     CatmullRomInterpolate(points[0], points[0], points[1], points[2], 10, pts);
-    std::vector<TThickPoint>::iterator it = points.begin();
+    std::vector<TThickPoint>::iterator it = points.begin() + 1;
     points.insert(it, pts.begin(), pts.end());
 
     pts.clear();


RodneyBaker:
You've grabbed my attention with this one for sure @pojienie 
Thanks for addressing this issue.
Will test!

RodneyBaker:
@pojienie In attempting to put together a few good test cases for this I ran into an instance where I still got some psuedo-hooks.   I think it can be reproduced at times when both Accuracy and Smoothness is turned all the way down to zero and short strokes are drawing very quickly.

If I'm reading this right it would be another issue... for another PR... that deals with the termination of splines at the terminal control point where the magnitude of the spline isn't sufficiently... hmmm... perpendicular.    A similar thing might relate to where when we are drawing say, a circle and attempting to have it end at the same point of the line's origin it doesn't like to meet smoothly. 

I'm not stating this quite right so I'll go back to testing...

RodneyBaker:
I am WAY out of my comfort zone here but something is not quite adding up to me.

This is perhaps more for spline patch surfaces but here is a 'makeflat' breakdown:

MakeFlat:
p0 = p0
p1 = p0 + (p1-p0).Norm() * (p3-p0).Normalize()
p2 = p3 + (p2-p3).Norm() * (p3-p0).Normalize()
p3 = p3

This seems to force values to be inside of p1 and p2.

REF:  Martin Hash's "A COMPLETE ALGORITHM FOR REAL-TIME SPLINE-BASED CHARACTER ANIMATION"

pojienie:
@RodneyBaker First of all, thank you for testing!

Not sure about the pesudo hook problem you mentioned, but if you change smooth value to 0, the `Smooth()` function won't even be called. so, yeah, it's probably another issue.
https://github.com/opentoonz/opentoonz/blob/f49503f713d2942e4afa418c2e1f3b56085ddd78/toonz/sources/tnztools/toonzrasterbrushtool.cpp#L773-L782

Also, I am not able to follow you here... what's "makeflat" doing here?

RodneyBaker:
The makeflat test (or perhaps it is more of an isflat test) is used to determine when splines and patches are 'flat enough' to render smoothly so subsequent processing is no longer required.  The paper I reference by Martin Hash goes into more detail but that is for 3D modeling and animation.

Aside:  I do wish that Smoothness setting allowed for a little more granularity as I tend to use values over 0 but don't always know where the sweet spot will be and have to reset it all the time.  I generally set it either all the way on (very slow) or to about 50.   I'm not complaining though as I'm pretty happy with the way vector lines draw now.

I was testing this PR earlier and wanted to validate via use of vector brushes but ran into a separate problem with crashing when selecting the vector brush tab.  Not sure what that relates to... corrupt brushes perhaps.  The reason for using the vector brushes is that the hooks are considerably more noticeable if/when they are created.

pojienie:
I see... So you're suggesting that "isflat" test can be added to the smooth process, right? Could be used to improve the performance of smooth process.

I couldn't reproduce the vector tab crash, perhaps it's because I wasn't using QT 5.14.1. It does remind me issue #1717 though.

beeheemooth:
@pojienie, good work! Thank you!

ezgif-2-2e5ee13e7468

RodneyBaker:
Thanks @beeheemooth for the additional confirmation.
Thanks @pojienie for the fix!

Issue #3127, opened 2020-03-03
Option to Hide Frame Bar in Viewer


MrBlockCat:
I'm not sure what the technical name for the bar is, but I'm referring to the one on the bottom of the viewer that you can move what frame you're on with. I've attached an image of it below.

<img width="1108" alt="Screen Shot 2020-03-03 at 11 10 37 AM" src="https://user-images.githubusercontent.com/37186692/75794977-9dccf800-5d3f-11ea-9c8c-1297fe65a8ef.png">

This option would fit in nicely with the other show/hide options right above the frame bar, which let you show or hide everything in the bar right above the frame bar.

<img width="163" alt="Screen Shot 2020-03-03 at 11 13 06 AM" src="https://user-images.githubusercontent.com/37186692/75795193-f7352700-5d3f-11ea-8bc2-25d1ac034c5d.png">

This option would be great to have because the bar can be fairly redundant when you have the Timeline view on screen, and more screen space is always nice!

### System Information
* **OpenToonz Version**: 1.4.0
* **Operating System**: macOS Catalina
* **CPU**: Quad Core i5
* **Memory**: 8 GB LPDDR3
* **Graphics Card**: 
* **Graphics Tablet**:

MrBlockCat:
Added with #3233!

PR #3233, opened 2020-04-14
Allow to Hide Console in Viewer

turtletooth:
This allows the Console (playback controls and playback slider) to be hidden in the viewer.  If you know the shortcuts, this can save screen space.

NoConsole

This adds the show/hide menu options that are available in the combo viewer.

menu

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -75,7 +75,7 @@ namespace {
 
 void initToonzEvent(TMouseEvent &toonzEvent, QMouseEvent *event,
                     int widgetHeight, double pressure, int devPixRatio) {
-  toonzEvent.m_pos      = TPointD(event->pos().x() * devPixRatio,
+  toonzEvent.m_pos = TPointD(event->pos().x() * devPixRatio,
                              widgetHeight - 1 - event->pos().y() * devPixRatio);
   toonzEvent.m_mousePos = event->pos();
   toonzEvent.m_pressure = 1.0;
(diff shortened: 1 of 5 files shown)


MrBlockCat:
Awesome, thank you! I know I’ll be using this option. To me this PR is already great as is, but might it be possible to split the setting into two, to be able to keep the playback controls visible but not the playback slider or vice versa? If it’s difficult to code that to work that’s okay. In either case, this fixes #3127!

RodneyBaker:
Very nice Jeremy.
LGTM

beeheemooth:
I think this thing should also be in the "View" tab. 
Thank you, @turtletooth

RodneyBaker:
Concerning @beeheemooth's feedback...
One of the useful things about this PR is how hiding of the console only applies to the current viewer.  Perhaps a version of this PR could be added to the View menu that hIdes the console for all viewers.  Adding to the View menu is a little more complex than the Right Click context in that:

If the same hiding process is added to the View menu the code would have to allow for:
- Determining whether to hide some or all of the consoles
- Determine whether or not to update currently open viewers
- Determine whether to hide the consoles of viewers that will be opened in the future

It seems to me that a similar feature located in the View menu might need to have a more complex logic and perhaps even a set of suboptions in Preferences.

gab3d:
very nice addition, thank you @turtletooth !

one interesting thing present in Maya, that helps to use the UI with less visible panels is a shortcut that allows the animator to drag the mouse left/right to scrub through the frames. it's very handy, really...

it works while the user keeps pressing the 'K' key + draging the mouse in the said direction.
I know this PR is not for that, but I take the opportunity to mention it for the sake of having these kind of options on the radar.
thanks again!

turtletooth:
@gab3d You can actually do that in OT with Ctrl + Shift and drag.

turtletooth:
Hey all- guess what's next?

what

gab3d:
@turtletooth
ohh... thanks! 
I didn't know/remember it. 😊

gab3d:
I love a horizontal Level Editor!

RodneyBaker:
A post mortem note...

The hiding of the playback controls and playback slider doesn't appear to get set anywhere so doesn't persist from one Opentoonz session to the next (after restart).

PR #3234, opened 2020-04-14
Allow undo/redo converting level to vectors

martinvanzijl:
Adds ability to Undo and Redo the _Level --> Convert --> Convert to Vectors..._ command. Video demo below:

undo-convert-demo

Haven't tested using a column with linked Fx's but assume they will be OK.


RodneyBaker:
Thank you Martin.

Aside:  One of my thoughts on workflow was rather we'd be able to run the conversion... copy some or all of the vectors... undo... and still have access to what we copied.  I'm happy to report that is in fact the case.

Of course, there are several other benefits to be gained too.

LGTM

PR #3236, opened 2020-04-14
Frame slider and Play bar can be hidden separately

turtletooth:
This allows the user to hide the playback bar and the frame slider separately.
It also allows the sound and locator icons to be hidden from the play bar using the menu
Finally it reorders the menu of the play bar to show up in the order they are in the bar.


RodneyBaker:
Thanks for the fast turnaround on this one Jeremy!
LGTM

PR #3237, opened 2020-04-15
Enable sending files to Trash in Browser on Linux

martinvanzijl:
Allows sending files to the Trash (recycle bin) from Browser windows on Linux. Demo below, shows me deleting a file and then restoring from the Trash in Ubuntu:

linux-delete-file-demo-after

This works for me on Ubuntu 18.04, but I'm not sure about other Linux distributions. Is anyone else able to test?

### DIFF
--- a/toonz/sources/common/tsystem/tsystempd.cpp
+++ b/toonz/sources/common/tsystem/tsystempd.cpp
@@ -59,7 +59,10 @@
 #include <dlfcn.h>
 #include <utime.h>
 #include <sys/time.h>
-
+#include <QDir>
+#include <QFileInfo>
+#include <QTextStream>
+#include <QUrl>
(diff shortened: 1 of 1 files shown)


martinvanzijl:
Also, this currently only works for deleting files from the home (local) drive. I haven't implemented deleting from external drives yet, and this will just print a warning to the console: "Deleting files on external drives in Linux is not implemented yet."

RodneyBaker:
Could use a Linux user to test and render feedback on this.
Looks good from here thus far.

cc:  @pojienie

pojienie:
I can confirm that it works on Artix Linux.

RodneyBaker:
Thanks @pojienie 
And thank you Martin!

PR #3238, opened 2020-04-17
Horizontal Level Strip

turtletooth:
This will allow the level strip to be switched to a horizontal orientation.  The toggle is in the context menu.

Any requests or ideas for where to put the level select drop down?

what

### DIFF
--- a/toonz/sources/toonz/filmstrip.cpp
+++ b/toonz/sources/toonz/filmstrip.cpp
@@ -723,7 +723,8 @@ void FilmstripFrames::paintEvent(QPaintEvent *evt) {
     }
 
     // navigator rect
-    if (naviRect.isValid() && fid >= 0 && fid == getCurrentFrameId()) {
+    if (m_showNavigator && naviRect.isValid() && fid >= 0 &&
+        fid == getCurrentFrameId()) {
       p.setPen(QPen(Qt::red, 1));
       if (m_isVertical) {
         p.drawRect(naviRect.translated(0, oneFrameHeight * i));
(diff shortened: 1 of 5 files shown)


DarrenTAnims:
Wow! That's really useful.

Does it change orientation automatically or is there a key to press or menu item? If it's not automatic, could it have a button (possibly using the same icon) as the xsheet/timeline, placed at the top left? Then it'll be easy to discover.

Regarding the drop down. I think it's fine where it is, but could you make it the width of the largest item instead of the full width? Then this makes it part of a toolbar, ready for placing other UI elements on it, as new features are added.

jrhyme:
Excellent))) it will be very convenient)

RodneyBaker:
I don't suppose a menu could be added directly in front of the level strip; just above when vertically oriented and just to the left when horizontal?

jrhyme:
@RodneyBaker Why? It would be convenient just above the timeline.

RodneyBaker:
My assumption is that Level Strip can be used... without the timeline or xsheet.
So, if it's not in/attached to the Level Strip... it cannot be accessed.

beeheemooth:
@turtletooth Wow! True, the ability to set Onion Skin (and Shift and Trace) markers has not yet been realized. I also noticed that in the vector level, frames can be dragged only for the first selected frame, since starting from the second click anywhere invokes the "inbetweens" command. But still - it looks impressive!

I once proposed a similar feature request. I thought that a drop-down list could be placed on the same row as the scroll slider. But now it seems to me that this was a bad idea, since it still takes up a lot of space vertically, and I also did not take into account the fact that the panel should have a title for which it can be moved.
There is still an idea to remove the drop-down list altogether and make it possible to select levels in the Level Strip context menu.
Please make it possible to pan the frames row in Level Strip using the pressed scroll (as in the Xsheet or on the canvas) I think the logic should be the same here.
ls

gab3d:
@turtletooth 
> Any requests or ideas for where to put the level select drop down?

perhaps a imagen menu somewhere?
...like in the Style Editor.

konero:
Just for reference, reminds me of the Retas cel bank window, it has navigation buttons and other stuff, the layout is snapped to width with dynamic tiling instead of scrolling horizontally.

image

**Row1:** Arrows navigate back and forth between levels, combobox is what you already know, reload button, open, delete, sub-menu, file name, file size, file resolution.
**Row2:** File location, cel count.

I wouldn't hide the combobox though.

turtletooth:
This is ready for testing.

RodneyBaker:
Looking VERY good Jeremy.
Love that the menus/dropdown can be hidden.

manongjohn:
@shun-iwasawa Can you check to see why the Windows & Default Jenkins builds failed?

manongjohn:
Heh...once again, I use the "J" word in a sentence then suddenly they kick off on their own, but I bet if I purposefully try to restart those builds just by saying the "J" word nothing happens.

beeheemooth:
23969681_556372181379327_5162408436180189184_n

shun-iwasawa:
Jenkins again

shun-iwasawa:
Expecto Jenkins Patronum

shun-iwasawa:
Removed all intermediate files. Jenkins again

RodneyBaker:
Everything passed here @turtletooth 
Do you want to hold for now or merge this?

turtletooth:
I'm fine merging now, unless there are any major bugs that have been noticed during testing.  I can fix bugs as they show up after merging.

beeheemooth:
There's an old problem again. :(

ezgif-2-146de173a2af

turtletooth:
Yeah. I had to revert that to make horizontal possible.  I’ll see if
there’s a way to have both.

turtletooth:
K - fixed it in #3252

papio-pl:
Great addition. Thank you:)

artisteacher:
This is great! I agree with this:

> Could we have a button, possibly using the same icon as on the xsheet/timeline, to toggle between horizontal/vertical placed at the top left? Then it'll be easy to discover. Or can the window not decide itself, based on whether the panel it taller or wider?

It would be amazing if the window automatically chose the orientation based on the space it was being placed in. It could be based on the red placement guideline - if the line is horizontal, the level strip would be horizontal. If the guideline was vertical, the level strip would be vertical. Otherwise, if there was a button, I would just use the the same one as the style editor orientation button.

PR #1976, opened 2018-05-15
Fix Note Button Text Clipping

konero:
Note buttons have a liberal amount of padding inherited from stylesheets, this will remove most of it by using the predefined `#PushButton_NoPadding` block in the stylesheet.

**Before:**
note_before
**After:**
note_after

PR #3243, opened 2020-04-19
Add search bar in fx browser

pojienie:
This PR allows you to search in Fx browser.

video: https://youtu.be/THvFztcCJX0

I created this because I constantly forget where "palette filter" FX is.....

code wise, I copied some from toonz/shortcutpopup.cpp for the QT widgets. Was about to copy the search algorithm as well but I failed to do so so I had to write my own version. Logic is basically the same, though.

During testing, I encountered a bug where if nothing is selected and you click on "insert" or "add", opentoonz will crash. I also fixed this bug in this PR.

### DIFF
--- a/toonz/sources/toonz/insertfxpopup.cpp
+++ b/toonz/sources/toonz/insertfxpopup.cpp
@@ -494,8 +494,9 @@ TFx *InsertFxPopup::createFx() {
   TXsheet *xsh      = scene->getXsheet();
 
   QTreeWidgetItem *item = m_fxTree->currentItem();
-  QString text          = item->data(0, Qt::UserRole).toString();
+  if (item == NULL) return 0;
 
+  QString text = item->data(0, Qt::UserRole).toString();
   if (text.isEmpty()) return 0;
 
   TFx *fx;
--- a/toonz/sources/toonz/insertfxpopup.cpp
+++ b/toonz/sources/toonz/insertfxpopup.cpp
@@ -143,16 +143,34 @@ TFx *createMacroFxByPath(TFilePath path) {
 // FxTree
 //=============================================================================
 
+void FxTree::displayAll(QTreeWidgetItem *item) {
+  int childCount = item->childCount();
+  for (int i = 0; i < childCount; ++i) {
+    displayAll(item->child(i));
+  }
+  item->setHidden(false);
+  item->setExpanded(false);
+}
+
+//-------------------------------------------------------------------
+
+void FxTree::hideAll(QTreeWidgetItem *item) {
+  int childCount = item->childCount();
+  for (int i = 0; i < childCount; ++i) {
+    hideAll(item->child(i));
+  }
+  item->setHidden(true);
+  item->setExpanded(false);
+}
+
+//-------------------------------------------------------------------
+
 void FxTree::searchItems(const QString &searchWord) {
   // if search word is empty, show all items
   if (searchWord.isEmpty()) {
     int itemCount = topLevelItemCount();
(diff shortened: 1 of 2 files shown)


DarrenTAnims:
This look superb! Great job @pojienie. You're an animal in here.

KaseyFarron:
Awesome!

gab3d:
wow!!  thank you!!
that's just awesome... 

perhaps you could check to see why this happens in the FX Browser in other languages?
imagen

as you can see in the image, there are two text strings cropped in the buttons (I've put the original strings over them, for you to check)

jrhyme:
@gab3d In Russian, the same thing, in all functions of the program.

konero:
Oh very nice, I requested this once before!

@gab3d I fixed a similar issue in the past: https://github.com/opentoonz/opentoonz/pull/1976

In the stylesheet there's an override to disable padding. You can enable it by using the following on the button:

```C++
buttonThatHasTheProblem->setObjectName("PushButton_NoPadding");

Those buttons could have better margins though, that could fix it.

**Comment by gab3d** [comment]
@konero 

Thanks for pointing it out!
I'm no programmer though, I can only cry about it 😸

**Comment by jrhyme** [comment]
I still do not understand where and what to click)

**Comment by RodneyBaker** [comment]
> I still do not understand where and what to click

From what I've tested it appears the search is only accessible via xsheet/timeline creation using the New FX method.

**Comment by jrhyme** [comment]
@RodneyBaker I said about the incorrect display of languages.

**Comment by pojienie** [comment]
@DarrenTAnims @KaseyFarron Thanks!

@gab3d Following @konero 's direction, it could be done by adding 3 lines in the following code snippet. I didn't test any other languages, though.
```
insertBtn->setObjectName("PushButton_NoPadding");
addBtn->setObjectName("PushButton_NoPadding");
replaceBtn->setObjectName("PushButton_NoPadding");
```
https://github.com/opentoonz/opentoonz/blob/26b3167ad04fcda4772eaea6e18f9960d340c926/toonz/sources/toonz/insertfxpopup.cpp#L213-L227
a

@konero Thanks a lot for the direction for @gab3d 's issue! Could you submit a PR? I'm not familiar with QT so I don't really know what `insertBtn->setObjectName("PushButton_NoPadding");` does exactly without further studying it. And without understanding it, I fear that I might create regression. 
The code location is the same as the above code snippet.

turtletooth:
@pojienie It's safe to include that change in your PR.

pojienie:
@turtletooth alrighty. I've committed the changes in. Thanks for confirmation!

Agnyy:
Very cool, reminds a panel in the After effect.
image 
 It would be cool if these ponels could be fixed. In the likeness of AE.

It would be convenient if there was another panel effect controls.
image

DarrenTAnims:
There is @Agnyy. You can dock the effects editor. It's really useful. I use it all the time now. Unfortunately, if you double click on a node, you still get the effects editor popup window too.

WolfInABowl:
@Agnyy I agree that it would be useful to be able to dock the FX Browser, currently it is only a popup and not treated as a standard UI panel.

gab3d:
I agree it'd be good to be able to dock the FX Browser as a panel.

gab3d:
Regarding the FX Settings panel (which is already dockable), I'd love it could evolve to be a general Settings panel, for whichever object is the currently selected one in the interface, be that an effect, a column, a level, an output node, a camera, etc. 
This would simplify tremendously the logic of the UI, and any user would know where to find the settings for anything anytime.

RodneyBaker:
Thank you @pojienie!

PR #3249, opened 2020-04-23
Re-add Spanish translation Preferences spacing

MrBlockCat:
@gab3d This was the only spacing I could find in any language (besides English) that changed between words in my Preferences typo-fixing PR, but please let me know if there was anything I missed!


gab3d:
thank you!

Issue #3254, opened 2020-04-24
Changing the default Units needs a restart


DarrenTAnims:
**Description**
After a clean install of OpenToonz, if you try to create a raster level, it shows the size in mm. Change it to pixels in the preferences, then try to create another raster level and it still shows in mm. You have to restart to see mm.

**To Reproduce**
1. Clean install of OT
2. Click the New Raster Level button on the xsheet toolbar
3. Notice that the size is in mm, with DPI showing too.
4. Cancel that dialog & open the Preferences (File > Preferences)
5. On the interface page (2nd down), change the Unit from mm to pixels.
6. Close the prefs dialog
7. Click the New Raster Level button on the xsheet toolbar
8. The size still shows as mm

**Expected behavior**
1. After step 7, you should see the size in pixels
2. Also, it still shows DPI as an option. This makes no sense.

**Solution**
When seeing the size in pixels on the level select dialog, don't show DPI.

Also, either:
1. (preferred) Allow changing the units in the preferences dialog, then immediately creating a new level using those units
2. (Not ideal, but consistent) Put an asterisk next to the units option, so the user knows to restart.

DarrenTAnims:
Also, after changing my default units to mm, on the Drawing tab, the new level's size is still shown in mms, which is really unhelpful. I've not idea how large 406.4mm x 228.6mm at 120 dpi relates to in pixels.

turtletooth:
@DarrenTAnims Where on the drawing tab do you see the size?

DarrenTAnims:
> @DarrenTAnims Where on the drawing tab do you see the size?

... here:

![image](https://user-images.githubusercontent.com/28575907/81800674-18b64a00-950b-11ea-9eb4-0d329adee7da.png)

RodneyBaker:
I'm currently not seeing this problem although I need to do more testing to fully confirm.
@DarrenTAnims Can you confirm this is still an issue?

PR #3261, opened 2020-04-29
Use correct default units when re-showing Create Level popup

martinvanzijl:
Fixes part of #3254. The _New Level_ pop-up should now use correct measurement units for width and height, even if they are changed in the Preferences. Demo below:

change-unit-demo

This should fix the problem from the issue description. The first comment (https://github.com/opentoonz/opentoonz/issues/3254#issuecomment-619154377) mentions about the Drawing tab still showing old units. I haven't looked at that.

### DIFF
--- a/toonz/sources/toonz/levelcreatepopup.cpp
+++ b/toonz/sources/toonz/levelcreatepopup.cpp
@@ -345,11 +345,15 @@ void LevelCreatePopup::showEvent(QShowEvent *) {
     m_dpiLabel->hide();
     m_widthFld->setDecimals(0);
     m_heightFld->setDecimals(0);
+    m_widthFld->setMeasure("camera.lx");
+    m_heightFld->setMeasure("camera.ly");
   } else {
     m_dpiFld->show();
     m_dpiLabel->show();
     m_widthFld->setDecimals(4);
     m_heightFld->setDecimals(4);
+    m_widthFld->setMeasure("level.lx");
+    m_heightFld->setMeasure("level.ly");
   }
 }
 


turtletooth:
LGTM.  Thanks.

beeheemooth:
@martinvanzijl I installed today's nightly-builds with your PR and noticed that the new Toonz Raster levels are created without frames (in Level Strip the icons are indicated as "no icons"). I'm not sure if this glitch is related to your PR, but previous nightly-builds do not have such a problem.

beeheemooth:
nor can you draw
ezgif-2-95cb5ec6acb1

DarrenTAnims:
I'm guessing this might have something to do with the dialog showing the size as 0 x 0. Or is this an unrelated thing?

beeheemooth:
Woow, @DarrenTAnims I didn’t notice this! Indeed, when creating a level, the width and height always have zero dimensions. It can be fixed?

DarrenTAnims:
In 1.4, they don't have dimensions of 0, they show the size in pixels, inches or whatever, so this seems to have crept into this fix.

beeheemooth:
I can imagine how many questions there will be from users ...

beeheemooth:
wait, what is it?...

I just click on the input boxes, and they seem to “remember” the numbers

ezgif-2-db26d7199fea

turtletooth:
I pulled the trigger too fast on this.  Seemed like a simple fix.  Looking into it now.

martinvanzijl:
Sorry, looks like this is indeed a problem introduced by my changed. I should have looked more closely. The call I made to `setMeasure` resets the values to zero, here:

https://github.com/opentoonz/opentoonz/blob/c7ce1288f47f546f4bab8c876a85764ff9de662c/toonz/sources/common/tunit/tunit.cpp#L448

@turtletooth If you don't find an easy fix, I recommend reverting this PR entirely.

turtletooth:
@martinvanzijl I've got it fixed.  Thanks.

PR #3262, opened 2020-04-29
Add locking stroke functionality in cutter and pump tool

pojienie:
This PR allows you to "lock" a stroke when using cutter and pump tool by holding control key.

video: https://www.youtube.com/watch?v=frl7fsuCZok

I created this because using pump tool on eyelash stroke is a pain if there is another stroke(e.g. hair) at the end of the eyelash.

### DIFF
--- a/toonz/sources/tnztools/cuttertool.cpp
+++ b/toonz/sources/tnztools/cuttertool.cpp
@@ -164,6 +164,8 @@ public:
   int m_cursorId;
   double m_pW;
 
+  int m_lockedStrokeIndex;
+
   TPropertyGroup m_prop;
   TBoolProperty m_snapAtIntersection;
 
@@ -300,7 +302,7 @@ public:
     return w;
   }
 
-  void leftButtonDown(const TPointD &pos, const TMouseEvent &) override {
+  void leftButtonDown(const TPointD &pos, const TMouseEvent &e) override {
     if (getViewer() && getViewer()->getGuidedStrokePickerMode()) {
       getViewer()->doPickGuideStroke(pos);
       return;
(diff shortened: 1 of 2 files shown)


beeheemooth:
Very useful! Thanks!

openanim:
Thanks, works fine for me !

jrhyme:
Let's merge, I want to use on Linux ahah)

beeheemooth:
@pojienie What about the Pinch Tool? He also likes to jump from one vector to another.
It is absolutely impossible to select the edge of the desired vector in a place where several lines converge.

turtletooth:
I'll look at this soon.

gab3d:
@pojienie 
nice job! it will increase productivity of the tool 👍 

perhaps, now that you've discovered the power of the hardcoded Ctrl key variant 😉 is there any chance of adding it to #3194 too? it'd be just the same, really. pretty please! 💟 

one way or te other, thanks for all of your contributions to OT, you're doing a very good job lately!

pojienie:
@beeheemooth I don't use pinch tool at all so I missed it...

@gab3d The reason why I hardcoded this was because it makes no sense to have a GUI element for this functionality and, without a GUI element, I have no idea how to add it to shortcut menu. So, no.

gab3d:
@pojienie 
I get the point, and certainly you're free to develop as you understand is the best way.

In spite of this, at least to me, workflow's priority is higher than UI, or having many options; so if I insisted with this was just to raise the case for a focus on workflow, over having many options 'per se' (even though having options is always good, of course).
In brief, what I say is that it might not be enough for us to add a lot of options in OT to make it a good solution, if the usage of each one of them is not well though out and optimized in the best way possible.

I'm certainly grateful for you're efforts, please don't take it as a bad criticism, but I'd certainly like you to consider other poins of view as well.
Thanks

RodneyBaker:
Thanks @pojienie  This will make a nice addition to the Cutter and Pump Tools.

This is just waiting on @turtletooth's review, otherwise it looks good to move forward.

Aside:  This is one of those features that may be hard to discover.
Short of word of mouth and documentation I'm not sure how to promote it more in users minds.

RodneyBaker:
Pressing forward with this as Jeremy has a lot going on right now.

PR #3266, opened 2020-04-30
Fix level create popup units

turtletooth:
Fixes the issue with #3261


RodneyBaker:
Looks like a Jenkins refresh is required.

turtletooth:
Oh Jenkins

RodneyBaker:
cc:  @shun-iwasawa @manongjohn

manongjohn:
Let's see if this works...probably won't...but let's tey anyway...  Jenkins

manongjohn:
As usual. Jenkins build won't start when I want it to but does when I don't expect it to.

turtletooth:
I'm going to merge this since the current level creation is broken.

PR #3270, opened 2020-05-05
New Fx: Glare Fx Iwa

shun-iwasawa:
This PR will introduce a new fx "Glare Iwa", which reproduces glare in photography, a star-shaped streak pattern due to Fraunhofer diffraction of light caused by diaphragm or [star filters](https://hoyafilterusa.com/collections/star-effects/products/hoya-cross-screen). 

An unique characteristic of this effect is that users input the diaphragm or the filter shape for computation, instead of inputting the streak pattern itself.

## Parameters
- **Render Mode** : either "Filter Preview" or "Render". The "Filter Preview" mode is for editing the single streak pattern by displaying it at the center.
- **Intensity** : changes the brightness of the light.
- **Filter Size** : changes the size of the pattern.

The following parameters are for deforming the pattern shape by applying noise to the length of the streak in radial direction. 
- **Noise Factor** : specifies contributing rate of the noise.
- **Noise Size** : specifies size of noise pattern.
- **Noise Octave** : specifies the granularity of the noise.
- **Noise Evolution** : specifies evolution of noise pattern.
- **Noise Offset** : offsets position of the noise to be applied to each streak.

## Tutorial
https://youtu.be/uZLj7TyIxds

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1305,6 +1305,17 @@
   <item>"STD_iwa_LinearGradientFx.wave_frequency"	"Frequency"	</item>
   <item>"STD_iwa_LinearGradientFx.wave_phase"		"Phase"		</item>
 
+  <item>"STD_iwa_GlareFx" 			"Glare Iwa" 	</item>
+  <item>"STD_iwa_GlareFx.renderMode"	"Render Mode"		</item>
+  <item>"STD_iwa_GlareFx.intensity"	"Intensity"		</item>
+  <item>"STD_iwa_GlareFx.size"	"Filter Size"		</item>
+  <item>"STD_iwa_GlareFx.rotation"	"Filter Rotation"		</item>
+  <item>"STD_iwa_GlareFx.noise_factor"	"Noise Factor"		</item>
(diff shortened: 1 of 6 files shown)


gab3d:
@shun-iwasawa
great effect!! thank you...

could it be possible that the effect already had a Composite option, where to specify between (say) Normal, Screen, Add?
seeing the youtube video I see you add a composite node just to iverlay it onthe the actual background, but if the effect node could contain itself a basic compositing capability, then 90% of times the schematic would be simpler.

Anyways, I'm glad we have an effect of such quality and flexibility now. :)

RodneyBaker:
@shun-iwasawa I'm not sure whey the builds are failing but this PR as well as a few others seem to have that same problem.

shun-iwasawa:
I'll look into it

shun-iwasawa:
jenkins

shun-iwasawa:
jenkins

shun-iwasawa:
jenkins

RodneyBaker:
That seems to have done the trick.
Thanks Shun for the new FX!

PR #3272, opened 2020-05-08
Remove references to linetest and flash

turtletooth:
There was a lot of old code left over from when Toonz could output flash files and from the linetest features.  This clears out those pieces in order to have a cleaner codebase. 

Along the way I found some buried features:
The ability to use a vector layer as a mask without fx.  This doesn't work for renders, but for quick tests it could be handy.  I placed the toggle in the column header menu for vector levels only:
mask2

The ability to set interpolation steps from the xsheet or timeline:
interpolation

And an export panel that will export what the viewport sees before fx.  This doesn't work with fx or with mesh levels.  (It tried to work with mesh levels, but it would crash, so I blocked it for now).  This could be useful for teachers or people who prefer simple animations without fx.  This will export the masks mentioned above.  It may be a more stable render option if you don't need fx.  It can also export multiple scenes into the same output file/image sequence.  Just drag multiple scenes in and it should work.

export

This PR hits a lot of files and removes a lot of code, so I'd love for @shun-iwasawa or @manongjohn to take a look at this.


gab3d:
thank you for tidying up the code, also those features are interesting.

regarding the masking feature, is this some sort of Column/Layer property or a Level property?

turtletooth:
@gab3d The masking is tied to the column.  It will mask out the column immediately above it ( or immediately to the right if using the xsheet).  The masking column has to be a vector, but the mask is applied to any type of column.

gab3d:
@turtletooth
thanks!

turtletooth:
Let's try this. . . Jenkins

manongjohn:
Looks like it started for you...this time. 

Appveyor failing for the following:

```
[00:16:46] C:\projects\opentoonz\toonz\sources\tcomposer\tcomposer.cpp(427): error C2027: use of undefined type 'TPropertyGroup' [C:\projects\opentoonz\toonz\x64\tcomposer\tcomposer.vcxproj]
[00:16:46]   c:\projects\opentoonz\toonz\sources\include\tiio.h(23): note: see declaration of 'TPropertyGroup'
[00:16:46] C:\projects\opentoonz\toonz\sources\tcomposer\tcomposer.cpp(427): error C2227: left of '->getProperty' must point to class/struct/union/generic type [C:\projects\opentoonz\toonz\x64\tcomposer\tcomposer.vcxproj]
[00:16:46] C:\projects\opentoonz\toonz\sources\tcomposer\tcomposer.cpp(427): error C2227: left of '->getValueAsString' must point to class/struct/union/generic type [C:\projects\opentoonz\toonz\x64\tcomposer\tcomposer.vcxproj]
```

turtletooth:
@manongjohn Thanks for seeing the TComposer error.

adamearle:
That masking system could be really good. I am wondering if there are any nuggets in the natron code that could give us masking and controlled feathering,

https://youtu.be/47opK56DUow?t=130

RodneyBaker:
@adamearle This is a different type of masking in that it's code that was already present for setting columns to be masks.

morevnaproject:
I wonder, how hard it would be to add option to invert the mask? ^__^

RodneyBaker:
@morevnaproject 
I like the way you think.
I would think that given values of 255 (white) or 0 (black) it would be fairly trivial to flip those values and allow the mask to invert.

There are other places in Opentoonz where similar benefit from this inversion of alpha channel/masking.  For example:  When painting with MyPaint brushes on raster levels with the 'Lock Alpha' option on the user can paint on any area except the area (presumably in black) in the alpha channel.  It would be useful to be able draw/paint on areas outside of the shape specified by the alpha channel via an invert option that flipped the values stored in the alpha channel (i.e. change black values to white and white values to black).  With this approach the user can draw/paint initially to specify where they want to apply color then use the mask (i.e. lock alpha with or without invert) to specify whether when they want to draw/paint inside a shape versus outside of the shape.

RodneyBaker:
As much as I hate to say it the found features introduce too many bugs and potential for confusion to add 'as is'.  As such I wouldn't recommend adding them until such a time as they can be developed more.   Merging the removal of unused references with linetesting and flash seems appropriate though.

Regarding the column masking which would be the most popular and useful of the found features if we could isolate the use cases and deal with crashes it would be a good candidate.  Use of column masking however is certainly complemented by the use of the found feature Export dialogue which is quite useful in it's present state.  I'd go so far as suggesting it be added but it has been suggested that having multiple export dialogues is sure to be confusing.  This is a shame because this hidden Export does have some very useful and unique features to it.

I'm going to rename this PR with (WIP) because I'm not sure how anyone intends for it to move forward.  Feel free to adjust title as necesary.

turtletooth:
I rehid the export panel and the masking command.  All that's left is the ability to change the step of an interpolation from the xsheet and removing the old code.

RodneyBaker:
This is looking good.
Open call for feedback prior to merging.
(I'm still reviewing and testing)

RodneyBaker:
Thank you Jeremy for cleaning out the code.
LGTM

PR #3275, opened 2020-05-11
Remove Changing Pins on IK

turtletooth:
IK is currently a mess simply due to the fact that we can change which bone is the pinned bone.

The calculations for this are horribly broken and lead to errors in many different scenarios.  

This removes the ability to change which bone is pinned.  This still allows temporary pins, but the root of the skeleton will always be the main pin with this change.

Clicking on the joint will no longer change the main pin.  Shift clicking sets a temporary pin.

Clicking on the root joint now allows the user to move the skeleton around.  

I would love for changing pins to work, as it allows some neat options for workflow, but it is currently broken far beyond my ability to repair.  And I would like to see IK have some reliable functionality even if we lose changing pins.

As it stands, the unpredictable nature of the current IK system is just one more thing that would frustrate users and drive them away.

### DIFF
--- a/toonz/sources/include/tgeometry.h
+++ b/toonz/sources/include/tgeometry.h
@@ -65,7 +65,7 @@ public:
 \sa rotate270
 */
 template <class T>
-inline TPointT<T> rotate90(const TPointT<T> &p)  // counterclockwise
+inline TPointT<T> rotate90(const TPointT<T> &p)  // 90 counterclockwise
 {
   return TPointT<T>(-p.y, p.x);
 }
@@ -76,7 +76,7 @@ inline TPointT<T> rotate90(const TPointT<T> &p)  // counterclockwise
 \sa rotate90
 */
 template <class T>
-inline TPointT<T> rotate270(const TPointT<T> &p)  // clockwise
+inline TPointT<T> rotate270(const TPointT<T> &p)  // 90 clockwise
 {
   return TPointT<T>(p.y, -p.x);
 }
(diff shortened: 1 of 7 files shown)


beeheemooth:
ezgif-2-27caa80f2618

turtletooth:
@beeheemooth I'm sorry, but I'm not sure what to be looking for in the image above.

beeheemooth:
@turtletooth I tested your build and made sure that it works. Of course, the slip problem is still present as before, and the final details will need to be corrected frame by frame. It turns out hybrid inverse kinematics: the center is animated automatically, and the ends frame by frame. 
Thank you for the work done!
 I would like to hear the opinion of other users on this subject.

turtletooth:
I have an idea on how to fix the slip.  

As far as this PR, something needs to be done about IK, because what we have is just going to lead to user frustration.  I don't know of any other software, including Toom Boom that allow the user to  change the root of an IK skeleton.  

If there is a developer willing to take on this task, that would be awesome, but I played around with it a couple days, and couldn't find a strong solution.  There are too many scenarios and areas that can go wrong.  What might work for one setup, doesn't work for a different.

Eventually the ability to do what Toon Boom calls nails would be nice, where you can move the root and certain joints try to stay in place.  But that is further down the road.

turtletooth:
@beeheemooth I think I misunderstood the slip you were talking about.  I think it is something will have to be done by frame after the fact.

My main reason for wanting to make this change is that I want the parts of OpenToonz that we release to work reliably.  IK right now is a disaster.  Unless we have a developer who has a better understanding of how to make it work, I don't think it should be exposed to the public user base.

RodneyBaker:
LGTM
I realize there is much more to be done with IK but we'll leave that for other PRs as it is important to get this added so that IK can at least be effectively  utilized.

openanim:
Sorry to back on it as I missed this change, but please consider also the possibility to bring back it as before.

IMO, there are many cases where the current behavior is insufficient. Even though the old behavior was not straightforward and optimal, I personally find it much more effective. Currently, it seems like to be a regression to me.

PR #3277, opened 2020-05-12
Make the tool options scrollable

### LINKED ISSUE #1277: Clipping Text when in Selection Mode
## **Selection Mode menu can’t be read because it gets cut off sometimes - icons will solve this issue and would be better for other languages.**

clippingtextin selectionmode

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/46163047-clipping-text-when-in-selection-mode?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #1977: [Proposal] Taller Tool Options Bar?
That's just a question that keeps coming to my mind:
Could it be possible to make the Tool Options Bar dynamically adjust to a two row layout, when there is an excess of required space to show all of the selected tool's options?

For example, currently the Animate tool (when in All mode) has so many options that even in a 1920 wide monitor, there are still some options that end off screen.

Also, thinking of people using OT in some notebooks which resolution is even less than that, I imagine that other tools' (like Brush or Select) options would, too, be facing the same problem.

I'm thinking of an option in Preferences to allow this behaviour or not.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/58411860-a-taller-tool-options-bar?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F33713530&utm_medium=issues&utm_source=github).
</bountysource-plugin>

turtletooth:
Fixes #1277
Fixes #1977 (Sort of)

This makes the options bar scrollable if the screen space is too small.

### DIFF
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -109,7 +109,7 @@ protected:
   QHBoxLayout *m_layout;
 
 public:
-  ToolOptionsBox(QWidget *parent, bool isScrollable = false);
+  ToolOptionsBox(QWidget *parent, bool isScrollable = true);
   ~ToolOptionsBox();
 
   virtual void
--- a/toonz/sources/tnztools/tooloptions.cpp
+++ b/toonz/sources/tnztools/tooloptions.cpp
@@ -535,7 +535,7 @@ ArrowToolOptionsBox::ArrowToolOptionsBox(
   m_nsPosField =
       new PegbarChannelField(m_tool, TStageObject::T_Y, "field", frameHandle,
                              objHandle, xshHandle, this);
-  m_zField        = new PegbarChannelField(m_tool, TStageObject::T_Z, "field",
+  m_zField = new PegbarChannelField(m_tool, TStageObject::T_Z, "field",
                                     frameHandle, objHandle, xshHandle, this);
   m_noScaleZField = new NoScaleField(m_tool, "field");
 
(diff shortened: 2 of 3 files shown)


turtletooth:
Jenkins

RodneyBaker:
Very nice.
LGTM

RodneyBaker:
Post mortem feedback:
Although I can see some cases where it might be desired to have the scrollbar remember it's general location, it seems to me that when switching tools the toolbar should revert back to the original alignment where the left-most options are presented.  This being most desirable because it is assumed that the options to the left on the toolbar are the most used and therefore of most interest to the user.

Currently with the implementation of the scrollable menu... which works great... we lose the focus of the most important options because the scrollbar does not reinitialize and realign to the left.

Perhaps at the initiation/activation of a toolbar the scroll could default to left?
This might create some additional interesting options if/when the user opens more than one toolbar as they could have each toolbar display a specific option.

The main point is that when switching from tool to tool the scrollbar should likely align to left.
I'm not sure how to accomplish this without resetting all toolbars the user might have opened independently except if it is designed to revert to left-aligned upon initial creation.

RodneyBaker:
@turtletooth 
I believe I'm seeing in the Plastic Tool's toolbar where the scrollable area is not properly defined.

In the attached image the scrollable area is over to the right leaving the part to the left truncated.

ScrollError

openanim:
Just a thought about the many options for the animate tool. In harlequin there is a kind of burger to select or deselect the desired parameters. Maybe we can bring it back in OT ?

burger menu ui

By the way i confirm also the display issue on plastic tool option bar.

PR #3284, opened 2020-05-14
Fix Scrollable Plastic Options

turtletooth:
Fixes an issue with #3277 
Good catch Rodney.
@konero I had to change some qss for this since DvScrollWidget was overriding the QPushButton.  What I did works, but there might be a more elegant solution.


RodneyBaker:
Thank you Jeremy for the quick fix!

PR #3285, opened 2020-05-14
Fit to viewer on load scene

turtletooth:
This stops the big zoom in when you load a scene.  I've often had students not realize they were only drawing on a portion of the canvas because of the zoom in.


RodneyBaker:
A VERY welcome addition.
Thanks Jeremy!
LGTM

DarrenTAnims:
Aha !!! At last. This has always bugged me. Zooming to 100% on load when the drawing are larger than the viewer always bothers me. The 1st thing I do after loading a scene, EVERY TIME, it so zoom out a little so I can see everything. Thanks Turtleman

PR #3288, opened 2020-05-17
Fix lip sync extend to end

### LINKED ISSUE #3218: Importing a Lip Sync file doesn't start or end with the correct 'rest' drawing
This is a small issue and one that I can work around. However, it should be an easily fixed issue.

Importing a lip sync file using the menu item: Xsheet > Apply lip sync data to column, correctly uses the 10 drawings I've selected in the dialog, but if my lip sync info doesn't start on frame 1, it adds drawing #9 in those frames. It should add drawing #10, the rest frame. Likewise, at the end of the animation, it adds a single frame of drawing #9 or if you tick the box at the bottom of the dialog (Extend rest drawing to end marker), it extends drawing #9 to the end, when it should use drawing #10.

Here's the dialog showing that drawing 10 is the rest drawing:

LipSyncModal

### System Information
I'm on Windows using version 1.4

turtletooth:
Fixes #3218

Fixed the icon size in the lip sync popup window 
and fixed extend rest drawing to end.

### DIFF
--- a/toonz/sources/include/toonzqt/icongenerator.h
+++ b/toonz/sources/include/toonzqt/icongenerator.h
@@ -105,6 +105,8 @@ public:
   // icons from toonz levels
   QPixmap getIcon(TXshLevel *sl, const TFrameId &fid, bool filmStrip = true,
                   bool onDemand = false);
+  QPixmap getSizedIcon(TXshLevel *sl, const TFrameId &fid, std::string newId,
+                       TDimension dim = TDimension(0, 0));
   void invalidate(TXshLevel *sl, const TFrameId &fid,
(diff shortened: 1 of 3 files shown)


turtletooth:
Note, this only works for Morevna or version 2.0.  Version 1.0 doesn't set a rest frame at the beginning.

gab3d:
@turtletooth
sorry, 2.0 of what software? 🙄

turtletooth:
Jenkins

DarrenTAnims:
Thank you Mr Turtle. That's amazing

shun-iwasawa:
Jenkins

PR #3298, opened 2020-05-20
Fix Multiple Flipbooks Playback At the Same Time

### LINKED ISSUE #3296: Unable To Play More Than One Flipbook At the Same Time
**Describe the bug**
After introducing #3278 it becomes unavailable to play more than one flipbooks at the same time unless the `Link Flipbooks` command is used. 

**To Reproduce**
1. Open two flipbooks and load some files respectively.
2. Press the Loop Play button of one of them.
3. Press the Loop Play button of the other flipbook. 

The second flipbook doesn't start to play. The first one stops playing instead.

**Expected behavior**
The second flipbook starts to play without stopping the first one.

**Device Information:**
 - OS: Win 10
 - OpenToonz Version: Latest

shun-iwasawa:
This PR fixes #3296 

On synchronizing playback between viewers, I added a line for distinguishing the console owner between (viewer / combo viewer) and (flipbook / color model) so that the sync works only on the viewers / combo viewers.

### DIFF
--- a/toonz/sources/toonzqt/flipconsole.cpp
+++ b/toonz/sources/toonzqt/flipconsole.cpp
@@ -172,11 +172,11 @@ void PlaybackExecutor::run() {
   TUINT32 loadedInstant, nextSampleInstant = timeResolution;
   TUINT32 sampleTotalLoadingTime = 0;
 
-  TUINT32 lastFrameCounts[4] = {0, 0, 0,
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I confirm the original problem and that this PR fixes it.
Thank you @shun-iwasawa!

PR #3300, opened 2020-05-21
Use "Current Column Color" for the current column in the Function Editor

shun-iwasawa:
This PR will make the Function Editor to use `Current Column Color` value in the preference instead of the style sheet in order to gain consistency.
The current column color in the spread sheet and the current item color in the tree view have been changed.

currentcolumncolorforfunctioneditor


RodneyBaker:
LGTM
Thank you @shun-iwasawa

Issue #2467, opened 2019-01-08
Undefined libconv when build on Linux


christopherwoo:
### Issue Summary

When build on Ubuntu 18.04, it failed for "undefined reference to libiconv".

### Screenshot or Video Reference

```
[ 81%] Linking CXX executable ../bin/OpenToonz
../lib/opentoonz/libtnzstdfx.so: undefined reference to `libiconv_open'
../lib/opentoonz/libtnzstdfx.so: undefined reference to `libiconv_close'
../lib/opentoonz/libtnzstdfx.so: undefined reference to `libiconv'
```
### Steps to Reproduce

cd toonz/build
cmake ../sources
make

### Expected Results

something but not "undefined reference"

### Actual Results

```
[ 81%] Linking CXX executable ../bin/OpenToonz
../lib/opentoonz/libtnzstdfx.so: undefined reference to `libiconv_open'
../lib/opentoonz/libtnzstdfx.so: undefined reference to `libiconv_close'
../lib/opentoonz/libtnzstdfx.so: undefined reference to `libiconv'
collect2: error: ld returned 1 exit status
toonz/CMakeFiles/OpenToonz.dir/build.make:9972: recipe for target 'bin/OpenToonz' failed
make[2]: *** [bin/OpenToonz] Error 1
CMakeFiles/Makefile2:1018: recipe for target 'toonz/CMakeFiles/OpenToonz.dir/all' failed
make[1]: *** [toonz/CMakeFiles/OpenToonz.dir/all] Error 2
Makefile:129: recipe for target 'all' failed
make: *** [all] Error 2
```
### System Information
* **OpenToonz Version**: git commit 01e9518533aff5d76defc8b0b487c03e87bf1417
* **Operating System**: Ubuntu 18.04
* **CPU**: x86_64/Intel
* **Memory**: ４GB
* **Graphics Card**: Intel Corporation 2nd Generation Core Processor Family Integrated Graphics Controller 
* **Graphics Tablet**: none

rozhuk-im:
This commit remove unused iconv dep from code and should fix this issue: https://github.com/opentoonz/opentoonz/pull/3231/commits/bdac84a72d513a16394dbc3163c05b711e689a4c

rozhuk-im:
PR: https://github.com/opentoonz/opentoonz/pull/3304

PR #3304, opened 2020-05-22
Remove ICONV dep

rozhuk-im:


RodneyBaker:
This is looking good so far but I am in early stage of testing as much of this is unfamiliar territory.
Just wanted you to know someone else is looking into the PR @rozhuk-im.

toonz/sources/CMakeLists.txt
```diff
@@ -458,6 +458,8 @@ elseif(BUILD_ENV_UNIXLIKE)
     find_package(LZO REQUIRED)
     message("LZO:" ${LZO_INCLUDE_DIR})
 
+    find_library(PTHREAD_LIBRARY pthread)
```
martinvanzijl, line 461: Why is pthread needed?

What is the risk the user does not have pthread installed?

martinvanzijl:
Overall this looks OK. To clarify the changes:

- Iconv dependency removed.
- Pthread dependency added.
- Some unused functions removed, .e.g `locale_to_jp`.
- Some function declarations removed from header files, since they are not used externally, e.g. `mbs_to_wcs`.
- In igs_resource_msg_from_err_win.cpp, in `std::string mbs_from_ts`, the code-page parameter is now hard coded to `CP_THREAD_ACP` (this was the default value). Before the change it could also be either `CP_UTF8` (from `utf8_from_cp932_mb`) or `932` (from `cp932_from_utf8_mb`) but these functions have both been removed.

The affected code seems to be called from the "Fog Ino" effect, if there is an error in one of its threads. I'm not sure how to trigger this, though. 

Compiles and runs OK for me on Ubuntu 18.04. This should also be compiled/tested on Windows and Mac. I recommend testing an animation with the "Fog Ino" effect and seeing it still works OK.

martinvanzijl:
@shun-iwasawa Do you happen to know who wrote the "Fog Ino" effect? It would be nice to check with them too.

toonz/sources/CMakeLists.txt
rozhuk-im: This is required it app uses threads, ```#include <pthread.h>``` + any ```pthread_*()``` requires to add this.
On linux and probably mac pthread is somewhere in standard lib, IMHO, that is why this work before. But on BSD pthread requires additional lib for linker.

shun-iwasawa:
@martinvanzijl 
> Do you happen to know who wrote the "Fog Ino" effect?

The related codes were written by @masafumi-inoue . 

@rozhuk-im 
Thank you for your modifications for FreeBSD support. 
I sought a comment from @masafumi-inoue regarding this PR. Unfortunately he doesn't have enough time for reviewing for now. He is not sure if there would be no problem by removing the iconv-related codes as he wrote it a long time ago. 
He's just wondering if, isn't it an option that installing libiconv to FreeBSD instead of removing dependency.

rozhuk-im:
@shun-iwasawa This is not FreeBSD related code. Linux users also affected: https://github.com/opentoonz/opentoonz/issues/2467

FreeBSD have iconv lib, and there was 2 options to build OpenToonz:
1. Add -liconv to CMake file
2. Check is it really used and remove if not.

I found that this code does not used, you may recheck this to be sure that there is no mistake.
IMHO this is artifact from pre utf-8 times. :)

shun-iwasawa:
Jenkins

shun-iwasawa:
Jenkins

shun-iwasawa:
@masafumi-inoue reviewed and verified the changes.
I also confirmed that it works with no problem both on Windows and MacOS.

LGTM Thank you @rozhuk-im

rozhuk-im:
https://github.com/opentoonz/opentoonz/issues/3597

PR #3317, opened 2020-05-29
Fix Pegbar Restoring

### LINKED ISSUE #3315: OT crashes on pasting pegbar
**Describe the bug**
OT crashes on pasting pegbar in the Stage Schematci under some conditions.

**To Reproduce**
1. Create 3 pegbars.
1. Connect the Peg1 to the Peg3 as a child.
1. Copy the Peg1
1. Delete the Peg3
1. Paste Insert. OT crashes.

**Expected behavior**
OT should not crash and the peg should be pasted under the Table node

**Screenshots**
pegbar_crash

**Device Information:**
 - OS: Win10

### LINKED ISSUE #3316: Collapse in the Stage Schematic malfunction
**Describe the bug**
Using Collapse command in the Stage Schematic enables users to specify pegbars to be included in the sub xsheet. However, it sometimes unexpectedly creates unnecessary pegbars and fails to connect nodes properly in the sub xsheet.

**To Reproduce**
1. Create a level and 2 pegbars. Arrange them like as follows:
image

1. Select the `Peg2` and transform position by using the Animate tool.
1. Select the `Col1` and the `Peg2` nodes.
1. Right click the `Col1` and select `Collapse`.
1. Open SubXsheet and see how the pegbars are arranged.
image
1. There are 3 pegbar nodes. The transformations is set to `Peg3` but it is not connected to `Col1`.

**Expected behavior**
There should be only one pegbar node correspondent to the `Peg2` in the sub xsheet and should be a parent of the `Col1` node.

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #3315 , and fixes #3316 as well.

The major changes are as follows:
- Pasting pegbars will now try to restore the copied pegbar ID. If the copied pegbar ID is already used, then use the first available ID. Collapsing pegbars also assigns the same pegbar IDs as the ones in the parent xsheet.
- Connection between nodes will be restored only if the both parent and child nodes are copied & pasted. If the parent node is not copied together, it will be reconnected to the Table node on pasting.  The same applies to collapsing too.

### DIFF
--- a/toonz/sources/toonz/subscenecommand.cpp
+++ b/toonz/sources/toonz/subscenecommand.cpp
@@ -211,7 +211,7 @@ bool mustRemoveColumn(int &from, int &to, TXshChildLevel *childLevel,
     if (app != childLevel) {
       removeColumn = false;
       if (from != -1 && to != -1) {
-        rangeFound            = from <= row && row <= to;
+        rangeFound = from <= row && row <= to;
         if (!rangeFound) from = to = -1;
       }
       continue;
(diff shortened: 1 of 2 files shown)


RodneyBaker:
LGTM

PR #3322, opened 2020-06-03
Set "Fit to window" zoom when loading/creating scenes

morevnaproject:
By default OpenToonz sets zoom level 1:1 when loading new scenes.

This PR makes it set "Fit to window" zoom everytime when user loads (or creates) a scene.

This is useful when you work with many scenes and switching them often.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -20,6 +20,7 @@
 #include "versioncontrol.h"
 #include "cachefxcommand.h"
 #include "xdtsio.h"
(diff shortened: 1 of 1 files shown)


morevnaproject:
This is how working process looks after this PR applied - https://youtu.be/m9RoWmrrACs?t=853

turtletooth:
This has already been fixed with #3285 
The solution shown in this PR doesn't take into account all viewers.

morevnaproject:
Woops, I was working with old version... Thanks for that fix, @turtletooth ! ^__^

PR #3327, opened 2020-06-05
Linux Build Instructions: Fix command for launching OpenToonz

morevnaproject:
Shell will not consider LD_LIBRARY_PATH variable if it is placed on separate line.

### DIFF
--- a/doc/how_to_build_linux.md
+++ b/doc/how_to_build_linux.md
@@ -177,8 +177,7 @@ If you need to debug the application, you should be able to use `cmake -DCMAKE_B
 You can now run the application:
 
 ```
-$ LD_LIBRARY_PATH=./lib/opentoonz:$LD_LIBRARY_PATH
-$ ./bin/OpenToonz
+$ LD_LIBRARY_PATH=./lib/opentoonz:$LD_LIBRARY_PATH ./bin/OpenToonz
 ```
 
 ### Performing a System Installation


manongjohn:
It should work. I've not had issues doing that. What Linux version are you using?

Does this work instead?

```
export LD_LIBRARY_PATH=./lib/opentoonz:$LD_LIBRARY_PATH
```

morevnaproject:
Yes, with `export` that will work too.

I am on Linux Mint 19, bash shell.

The `export` directive makes variable available for subsequent commands. Without it next commands executed in the same terminal will not "see" it.
https://superuser.com/questions/18988/difference-between-a-b-and-export-a-b-in-bash

I suspect it works for you because you run those commands from script?
If you run from script, then all commands are executed in the same shell and the sequence will work.
But if you enter those commands one-by-one in terminal, then it will not work, because terminal executes each command in separate shell.

manongjohn:
Not running from script. I use bash shell on Ubuntu for quick build testing every so often. I actually forget to run that statement most of the time.

morevnaproject:
If I am not mistaken, ubuntu uses `dash` shell instead of `bash`, so maybe it works differently...

morevnaproject:
> I actually forget to run that statement most of the time.

This is actually the reason why I put this statement in a single command.
This way you can get this single command from shell history (arrow key up) and run. 
If you have 2 statements, then you have to remember to run `export LD_LIBRARY_PATH=...` every time when you open new terminal window.

manongjohn:
Ubuntu uses bash. My account uses bash.

I would probably put that export in my profile so I wouldn't have to type that all the time. 😁

morevnaproject:
> Ubuntu uses bash. My account uses bash.

Ah, you right. ^__^

> I would probably put that export in my profile so I wouldn't have to type that all the time.

Great idea! ;)

morevnaproject:
@manongjohn You right, it works in terminal without export. 

```
$ A=123
$ echo $A
123
```

Probably I made a mistake somewhere... ^__^"

PR #3328, opened 2020-06-05
Fix expression reference in sub xsheet

### LINKED ISSUE #3326: Expression referencing fails in sub xsheet
**Describe the bug**
Expression fails to reference parameter of another object in newly collapsed sub xsheet.

**To Reproduce**
1. Load or create some level in the column 1.
2. Select column1 and `Collapse`. (a sub xsheet will be created.)
3. Select the sub xsheet and `Open Sub-xsheet` . (Sub xsheet opens, containing the column you just collapsed.)
4. [In sub xsheet] In Function Editor, set the expression keyframe to `Col1.ew` referencing some another parameter like `table.ew`.
1. [In sub xsheet] Try changing the referenced value (`Table.ew`) and see if `Col1.ew` follows with it.

In my environment, `Col1.ew` seems to fail to refer the parameter. 
Instead, it actually refers the same object in the parent xsheet. (i.e. in this case, `Col1.ew` in the sub-xsheet changes according to `Table.ew` in the parent xsheet.)
Note that the reference recovers once you save and reopen the scene.

_**EDIT**_
The same problem occurs as well, when copying column in some xsheet and pasting it in another sub-xsheet. The pasted column cannot refer the values in the pasted sub-xsheet.

**Expected behavior**
Expression should work properly in just-collapsed sub-xsheets.

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #3326 .

Each parameters is needed to set "grammar" associated with the current xsheet.
If you bring some object from the parent xsheet to another sub-xsheet, you need to reset the grammar one or it still refer to the parameters in the parent xsheet.

### DIFF
--- a/toonz/sources/toonz/subscenecommand.cpp
+++ b/toonz/sources/toonz/subscenecommand.cpp
@@ -1123,9 +1123,16 @@ void bringPegbarsInsideChildXsheet(TXsheet *xsh, TXsheet *childXsh,
   for (pegbarIt = pegbarIds.begin(); pegbarIt != pegbarIds.end(); ++pegbarIt) {
     TStageObjectId id        = *pegbarIt;
     TStageObjectParams *data = xsh->getStageObject(id)->getParams();
-    childXsh->getStageObject(id)->assignParams(data);
+    TStageObject *obj        = childXsh->getStageObject(id);
+    obj->assignParams(data);
     delete data;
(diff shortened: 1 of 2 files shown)


RodneyBaker:
LGTM
Thank you @shun-iwasawa!

PR #3336, opened 2020-06-09
Fix pegbar lost on exploding sub xsheet

### LINKED ISSUE #3335: Loss of pegbar info on exploding sub xsheet
**Describe the bug**
Exploding a sub xsheet with `Bring relevant pegbars in the main xsheet.` option can unexpectedly overwrite existing pegbar in the parent xsheet by the one from the sub xsheet. Thus the overwritten pegbar information disappears after exploding. Undoing the Explode operation cannot recover it.

**To Reproduce**
1. Load / create 2 levels. Put a level `A` to the column 1 and a level `B` to the column 2.
1. Collapse the column 1.
1. Create 4 pegbars in the parent xsheet. Remove the `Peg3`.
1. Create 1 pegbar in the child xsheet.
1. Connect the stage objects like this:
explode_problem
1. Move the N/S and E/W positions of the `Peg4` in the parent xsheet and the `Peg1` in the child xsheet, to points distant from each other. (This is for the sake of clarity.)
1. Explode the sub xsheet choosing `Bring relevant pegbars in the main xsheet` option.

In my environment the `Peg1` in the sub xsheet is brought out to the parent xsheet being renamed to `Peg4`. The original `Peg4` is lost. Therefore the column2 unexpectedly moves to the same location as the column1.

**Expected behavior**
Pegbars brought out from the exploded sub xsheet should not overwrite the existing pegbars in the parent xsheet.

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #3335 .

When bringing out the pegbars from the sub xsheet, I added a lines to find the first available pegbar index for them instead of applying sequential indices from the "former-Table" pegbar.

### DIFF
--- a/toonz/sources/toonz/subscenecommand.cpp
+++ b/toonz/sources/toonz/subscenecommand.cpp
@@ -538,6 +538,11 @@ void bringObjectOut(TStageObject *obj, TXsheet *xsh,
     assert(id.isPegbar());
     pegbarIndex++;
     TStageObjectId outerId = TStageObjectId::PegbarId(pegbarIndex);
+    // find the first available pegbar id
+    while (xsh->getStageObjectTree()->getStageObject(outerId, false)) {
+      pegbarIndex++;
+      outerId = TStageObjectId::PegbarId(pegbarIndex);
+    }
     TStageObject *outerObj =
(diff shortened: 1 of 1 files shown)


RodneyBaker:
LGTM
Thanks @shun-iwasawa

PR #3349, opened 2020-06-12
Fix fxs missing on exploding sub xsheet

### LINKED ISSUE #3347: Loss of fxs when exploding a sub xsheet
**Describe the bug**
When exploding sub xsheet, fxs connected downstream of the xsheet node disappear.

**To Reproduce**
1. Build the fx schematic like as follows:
image
1. Explode the sub xsheet. The pegbar options does not matter either way.
1. Open the group and see if all the fxs are brought out properly.

In my environment the fxs connected downstream of the xsheet node ( the Directional Blur Fx in this example ) disappear.
image

**Expected behavior**
All fxs should be brought out to the parent schematic in order to keep the rendering result unchanged.

**Device Information:**
 - OS: Win10

### LINKED ISSUE #3348: On exploding sub xsheet, the Xsheet node is replaced by multiple Over fxs instead of one
**Describe the bug**
(It is actually not a bug, but IMO is unnecessary behavior.)
When exploding sub xsheet, if the sub xsheet node is connected to some fx node other than the Xsheet node, the Xsheet node in the sub xsheet schematic is replaced by the Over Fxs reproducing the stacking order of the nodes.
Currently the Over fx node is stacked for each connecting nodes. however, the Over fx has dynamic input port (i.e. a new port is added when the last port is connected) so the composition can be reproduced by only one Over fx.

**To Reproduce**
1. Build fx schematic like as follows:

2. Explode the sub xsheet.
3. Open the group and see how the xsheet node is replaced by multiple Over fx nodes:
image

**Expected behavior**
The xsheet node can be replaced by one Over fx node like as follows, for simplicity:

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #3347 and resolves #3348 .
As far as I tested, it seems to work as I expected - exploding the sub xsheet brings all fxs out so that it can reproduce the render result.

### DIFF
--- a/toonz/sources/toonz/subscenecommand.cpp
+++ b/toonz/sources/toonz/subscenecommand.cpp
@@ -478,7 +478,7 @@ TFx *explodeFxSubTree(TFx *innerFx, QMap<TFx *, QPair<TFx *, int>> &fxs,
   if (!xsheetFx) {
     if (innerDag->getCurrentOutputFx() == innerFx)
       innerFx = innerFx->getInputPort(0)->getFx();
-    if (!innerFx) return 0;
+    if (!innerFx) return nullptr;
     bringFxOut(innerFx, fxs, outerDag, fxGroupData);
     TOutputFx *outFx = dynamic_cast<TOutputFx *>(innerFx);
     if (outFx)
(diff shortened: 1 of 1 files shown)


RodneyBaker:
LGTM
Thank you @shun-iwasawa

PR #3353, opened 2020-06-15
Add more shapes to Adobe Animate keyboard shortcuts

MrBlockCat:
Thanks to @pojienie 's awesome work with adding lots of new keyboard shortcuts, I've expanded the Adobe Animate keyboard shortcut preset to have more accurate shape shortcuts! I tested it by manually replacing the stuff folder with the one in my local repository, and it seems to work as it should.


RodneyBaker:
LGTM
Thank you @MrBlockCat

PR #3355, opened 2020-06-15
fix a bug where level in xsheet gets replaced when cancelling saving

### LINKED ISSUE #3329: 'Save Level As' - Level Replaced When Action is Canceled
**Describe the bug**
Bug appears after executing the 'Save Level As...' command while 'Replace Toonz Level after SaveLevelAs command' in the options is enabled. If the user tries to save over an existing file and hits 'Cancel' when asked if they want to overwrite the file, the file will still be replaced on the Xsheet. The action is not able to be undone.

**To Reproduce**
1. Enable 'Replace Toonz Level after SaveLevelAs command' under the General tab in the settings dialog.
2. Create two new levels on the xsheet and save both (Level A, Level B.)
3. Place the xhseet cursor on the first level (Level A) and execute the 'Save Level As...' command.
4. Attempt to save over the second level (Level B.) When presented with a dialog box asking if you'd like to save over the existing file, hit 'Cancel.'

_The level will be replaced on the xsheet even though the action was canceled._

**Expected behavior**
Nothing. The action should be canceled.

**Device Information:**
 - OS: Windows
 - OpenToonz Version: Latest Nightly

pojienie:
This fixes #3329 

The reason why this bug exists is because saveLevel(const TFilePath&) function doesn't return `false` if user presses cancel. Gladly, the function it calls does so it's an easy fix.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1531,7 +1531,11 @@ bool IoCmd::saveLevel(const TFilePath &path) {
   if (realPath.getType() == "")
     realPath = TFilePath(realPath.getWideString() + ::to_wstring(dotts + ext));
 
-  saveLevel(realPath, sl, false);
+  bool ret = saveLevel(realPath, sl, false);
+  if(!ret){ //save level failed
+    return false;
+  }
+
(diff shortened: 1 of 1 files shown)


RodneyBaker:
LGTM
Thank you @pojienie

melieconiek:
Thank you @pojienie for the fix!

Issue #3282, opened 2020-05-13
Auto smooth arc tool 


openanim:
I wonder if the arc tool could have certain attributes of the tape tool in order to simulate a continuous curve tool.

In particular endpoint to endpoint, join strokes and smooth.

When joining 2 strokes with the tape tool with its attributes, a smooth curve is created.

If the arc tool could do this continuously with each addition of a new stroke to obtain an "auto smooth line" effect, we could maybe simulate a polyline curve tool, currently absent in OT.

Currently with the polyline tool we needs to create the curves manually. So similarly, perhaps a definable smooth value between two segments for the polyline tool could allow that.

Which would help with the lineart, the filling etc.

![auto smooth arc tool](https://user-images.githubusercontent.com/29790512/81850195-f26acb80-9557-11ea-8655-2c88a2396e5e.gif)

pojienie:
Is this what you want? https://youtu.be/ekCwA1bbvP4

basically I add 1 shape in geometric tool: MultiArc. It allows you to immediately start another arc shape after one is completed. The start point of the new arc is the end point of the previous arc.

Currently there is a huge bug when you only enable "join vector" and not "smooth". But I want to know if this is how you want it to work first.

openanim:
@pojienie Thank you so much, to tackle the implementation, that's exactly that !

Ideally, I wanted to do something like this. 

![Curve tool cacani](https://user-images.githubusercontent.com/29790512/85221826-584c3d80-b3b7-11ea-8eba-4ef9814544e4.gif)

Its more as a "polyline with curve" But I think your version fits perfectly with the tools present in OT and I am very grateful to you. Possible to be able to test it ?

pojienie:
wow, how does the software you're using know how "curvy" the lines should be?

I have submitted PR #3360 and it should have apveyor image after awhile.

(btw the artwork is from Regular Show and not by me, I just used it as an example that this feature is useful for drawing her kind of hair)

openanim:
Yes cacani has some good tools ... :) Thanks for this, exciting !

beeheemooth:
This is a great feature, but it doesn't seem to work properly with Vector Guided Drawing in the case meant that the line created with the help of MultiArc be whole (Join Vector).
![ezgif-4-5119d2389ebb](https://user-images.githubusercontent.com/25724133/85232765-3f20ac80-b40a-11ea-9c17-83cdf7337402.gif)
[...]

PR #3360, opened 2020-06-21
Add multi arc shape in geometric tool

pojienie:
related feature request: #3282 

video: https://youtu.be/ekCwA1bbvP4

This pull request adds a shape in Geometric Tool called Multi Arc. The way Multi Arc works is almost the same as Arc but, as soon as you finished the first Arc, it immediately starts another with the starting point of the new Arc being the same as the ending point of the previous Arc.

With Join option, all Arc drawn by Multi Arc will be joined as if you used tape tool on them. Same with Smooth option. Luckily, Auto Group and Auto Fill work automatically with the newly added code so I did not need to tackle with that.

However, this PR still has some bugs that I haven't been able to fix:

1. If join option is enabled and smooth option is disabled, sometimes the stroke will have missing control point (see #3072 and #3074) (fixed)
2. If join option is enabled, smooth option is disabled, and you draw a closed shape. The stroke WILL have problem (try dragging the control point of the connecting control point). (fixed)
3. If join option is enabled and you draw a closed shape, the Multi Arc tool won't automatically stop like how Multi Line works. (fixed)

Also, I did not test it in anything but vector level so I expect there are bugs when you use raster levels as well.

Edit: As @beeheemooth and @openanim reported:

1. use undo commend while drawing multiarc causes either crash or joining wrong strokes. (fixed)
2. Vector guided drawing doesn't work as expected. (fixed)
3. thickness doesn't work (fixed)
4. join vector shortcut doesn't work (fixed)

Edit: removed join option so that it is consistent with polyline shape (it will always join now)
double click ends the multiArc shape so that it is consistent with polyline shape

another issue as reported by @openanim:

1. no entry of multiArc shape in shortcut list (fixed)

### DIFF
--- a/toonz/sources/common/tvectorimage/tvectorimage.cpp
+++ b/toonz/sources/common/tvectorimage/tvectorimage.cpp
@@ -1034,8 +1034,8 @@ bool TVectorImage::Imp::areWholeGroups(const std::vector<int> &indexes) const {
     if (!m_strokes[indexes[i]]->m_groupId.isGrouped()) return false;
     for (j = 0; j < m_strokes.size(); j++) {
       int ret = areDifferentGroup(indexes[i], false, j, false);
-      if (ret == -1 ||
-          (ret >= 1 && find(indexes.begin(), indexes.end(), j) == indexes.end()))
+      if (ret == -1 || (ret >= 1 && find(indexes.begin(), indexes.end(), j) ==
+                                        indexes.end()))
         return false;
     }
   }
@@ -1282,7 +1282,7 @@ void TVectorImage::mergeImage(const TVectorImageP &img, const TAffine &affine,
         } else {
           img->m_imp->m_strokes[i]->m_groupId =
               TGroupId(groupId, img->m_imp->m_strokes[i]->m_groupId);
-	}
+        }
       }
     }
   }
@@ -2154,7 +2154,8 @@ VIStroke *TVectorImage::Imp::joinStroke(int index1, int index2, int cpIndex1,
 
   // check if the both ends are at the same postion
   bool isSamePos = isAlmostZero(tdistance2(stroke1->getControlPoint(cpIndex1),
-                                           stroke2->getControlPoint(cpIndex2)));
+                                           stroke2->getControlPoint(cpIndex2)),
+                                1e-4);
   // connecting the ends in the same shape at the same postion
   // means just making the shape self-looped
   if (isSamePos && index1 == index2) {
@@ -2908,7 +2909,7 @@ void TVectorImage::Imp::regroupGhosts(std::vector<int> &changedStrokes) {
              ((currGroupId.isGrouped(false) != 0 &&
                m_strokes[i]->m_groupId == currGroupId) ||
               (currGroupId.isGrouped(true) != 0 &&
-                  m_strokes[i]->m_groupId.isGrouped(true) != 0))) {
+               m_strokes[i]->m_groupId.isGrouped(true) != 0))) {
         if (m_strokes[i]->m_groupId != currGroupId) {
           m_strokes[i]->m_groupId = currGroupId;
           changedStrokes.push_back(i);
--- a/toonz/sources/common/tvectorimage/tvectorimage.cpp
+++ b/toonz/sources/common/tvectorimage/tvectorimage.cpp
@@ -1034,8 +1034,8 @@ bool TVectorImage::Imp::areWholeGroups(const std::vector<int> &indexes) const {
     if (!m_strokes[indexes[i]]->m_groupId.isGrouped()) return false;
     for (j = 0; j < m_strokes.size(); j++) {
       int ret = areDifferentGroup(indexes[i], false, j, false);
-      if (ret == -1 || (ret >= 1 && find(indexes.begin(), indexes.end(), j) ==
-                                        indexes.end()))
+      if (ret == -1 ||
+          (ret >= 1 && find(indexes.begin(), indexes.end(), j) == indexes.end()))
         return false;
     }
   }
@@ -1282,7 +1282,7 @@ void TVectorImage::mergeImage(const TVectorImageP &img, const TAffine &affine,
         } else {
           img->m_imp->m_strokes[i]->m_groupId =
               TGroupId(groupId, img->m_imp->m_strokes[i]->m_groupId);
-        }
+	}
       }
     }
   }
@@ -2154,8 +2154,7 @@ VIStroke *TVectorImage::Imp::joinStroke(int index1, int index2, int cpIndex1,
 
   // check if the both ends are at the same postion
   bool isSamePos = isAlmostZero(tdistance2(stroke1->getControlPoint(cpIndex1),
-                                           stroke2->getControlPoint(cpIndex2)),
-                                1e-4);
+                                           stroke2->getControlPoint(cpIndex2)));
   // connecting the ends in the same shape at the same postion
   // means just making the shape self-looped
   if (isSamePos && index1 == index2) {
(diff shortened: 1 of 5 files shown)


openanim:
@pojienie The features are really cool and works as expected. Except for the problems you listed and the crashes that prevent a more thorough test.

Quick feedback and some suggestions:

1. It seems to me that we cannot change the thickness of the stroke at this time.

2. Until the final stroke is completed with the active options, Could "undo" command work by deleting the last arc in progress instead of the entire stroke

(as it works with the polyline tool).

3. When both options are active

with the Guided drawing feature, it would be good if the following guide only started after the last stroke of a multiarc series and not the first.

To avoid visual offset of the guides. I do not know if it is something easy to set up?


Thank you for this nice addition.

pojienie:
Thank you for testing! I'll add things you've shown in the known bugs.

I'll delay any kind of bug fixes to guided drawing as there are other more serious issues.

The join shortcut should be an easy fix. I probably just gave it wrong ID for shortcut to work.

RodneyBaker:
Did some initial testing.  Thanks @pojienie for developing this.

When deciding to terminate drawing of the Multi-Arc line it seems the only way to do that is to select another Tool.
With the normal Arc tool double clicking signals the user wants to stop drawing.
With Multi-Arc double clicking seems to just tell OT to keep drawing where the user left off.
Is there a specific or preferred way to stop drawing with Multi-Arc?

Aside:  In other software I use a Right Click (or Escape) is used to terminate.

openanim:
@pojienie The thickness and shortcut fixes work well for me.

@RodneyBaker Same feeling about doble click that should works to end multiarc like it's works with the polyline tool. "Enter" key works to finish line but while you use shortuct for enable/disable smooth for exemple, it's more easy to doble click than press enter in the workflow.

pojienie:
@RodneyBaker Now there are 4 ways of ending the multiArc shape:

1. switch to another tool
2. double click
3. press enter
4. draw a closed shape

Polyline also uses these 4 ways of ending a stroke so I mimicked it.

@openanim Yeah I forgot to add the shortcut entry. I'll add it when I have time.

All changes and issues fixed are added/edited in the PR.

openanim:
I wonder what others think about that but as far as it would not complicate your task in resolving the pending issues, in my opinion, join vectors options should remain available as it brings certain benefits.

One of the reasons that comes to my mind, during the auto inbetween process, this would help a lot because sometimes we need to cut  a stroke to obtain several individual strokes to match with precision a shape. 

It  work, exactly like a "integrated cutter tool" by cutting the stroke at the chosen location with precision while retaining the advantage of having strokes still well connected with each other despite that join vectors option is deactivated. And consolidates the filling process. 

Nevertheless it is a task which can be done by others ways also even if you stay as it is now. And i ll understand the simplicty of this. what do you think ? 

Otherwise i noticed that the double click works well with the mouse but I had a slight problem by double clicking with my stylus, the multi arc stops but the stylus still snaps the canvas without being able to choose the desired direction after that. Need to be confirmed.

snap still active multi arc


Thanks for the lasts commits !

pojienie:
The reason why I removed join option was because Polyline didn't use it. I like consistency within a program because it would make a program easier to learn. For example, a newbie might be confused why lines are joined together when using polyline shape when join option is not enabled and the same isn't happening in MultiArc.

I do see the benefit of having the join option though. I might make another PR to add it back, however, I'm going to change polyline shape to also add not-joined lines if join option isn't enabled if that's the case. Anyway, I'm going to focus on this PR first :)

Ah, I don't have drawing tablet so I only tested the double clicking funcitonality with mouse. So... we might need another developer to fix this bug...... Current workaround might be using shortcut or pressing enter to cancel multiArc.

openanim:
@pojienie Personally, I am in favor of the possible solution that you are proposing since it will not destroy the previous workflow of the polyline tool and give more flexibility. But if you decide to stay as it is now, that's fine with me too. For the stylus side, I will use the workarounds available until a possible fix.

pojienie:
So, this is finally done. I guess.

The git history is very messy because I made a bunch of mistakes while implementing this. I only learnt the correct way of doing this feature while I was implementing #3369.....

I'm not proud of the implementation of undo but I couldn't come up with anything better.

MultiArc should now behave similar to Polyline.
1. When undoing/redoing while drawing, it would only undo/redo the last mouse click.
2. It would only add stroke to image when you actually finished it. (for example, if you switch tool mid drawing, it will NOT add any stroke)
3. You can double click, press enter, or draw a closed shape to stop drawing.

All issues should all be fixed. But this still requires testing and, if possible, code review.

openanim:
I played with it, a little while on vector and raster level, and everything worked fine for me! No crash and double click with the stylus also works on my tablet :). I find multiarc more useful than the functionality in cacani that I originally requested.

I just don't know what to think about the color that highlights the strokes.

maultiArc color strokes

Thanks for great work !

pojienie:
@openanim First, thank you for testing! Personally I couldn't find any serious bugs in this feature but my usage is very limited. Also, happy to hear that the double click works for you without me doing anything about it!

And that's an interesting bug. However, I couldn't reproduce it for some reason. Can you tell me what you did when you move your mouse out of the screen? I tried moving my mouse to various location that might cause the color change but to no avail.

openanim:
@pojienie Sorry for the invisible part which makes you think probably that this is a bug. I'm not sure.

It seems that depending on the style that you choose the color of highlighted will be different. Although It seem also that using the numeric keyboard as a shortcut to change styles does not update this one until we have the cursor on the xsheet. But actually, theses are very minors things.

multiarc strokes highlights


when using different styles like in a process for lines separations, it might seem weird to the eye that it changes continuously imo.

multiarc lines seps

pojienie:
Oh I see! Yeah the color changing according to style is already a thing in Arc shape. MultiArc is built on top of Arc so they behave the same in this case.

But I do agree that the color should stay consistent for already-drawn Arc. I'll make a commit when I have time.

openanim:
First thank you for your patience and sorry for the extra work it gives you @pojienie  :(, and to insit about that but by trying the latest appveyor. I can't see the difference from the previous one.

I also just noticed that the arc tool in this PR also has this behavior, however, it is not the usual behavior that I usually observe. Usually, the strokes have a similar color before and after, with different thikcness. (e.g gifs). The arc tool in the official version has a different behavior compared to this PR.

arc tool last NB


It also seems to me that in your very first versions you did not introduce this highlight behavior, so it would not be better to keep them as before for consitency ?Since arc tool or polyline tool dont use it at least visualy, we cannot see that on official versions.

Arc tool PR multiArc tool

pojienie:
Oops! I accidentally copied the code for color from MulitLine and assumed it was originally in Arc as well. The color should now be consistent with the original Arc tool.

manongjohn:
Build failed with this error:

```
C:\projects\opentoonz\toonz\sources\tnztools\geometrictool.cpp(2324): error C2666: 'TThickPoint::operator ==': 3 overloads have similar conversions 
```

pojienie:
b-b-but it builds on my machine...

jokes aside, it does produce warnings on my end. I thought it was okay so my bad. It should now be fixed as the warnings don't appear on my end. Hopefully.

openanim:
Yeah, it works perfectly well, I didnt noted something very annoying. The only detail point that I can note is that the circle at the start of the stroke arrives after the first stroke instead of starting at the same time but it is not annoying. ^^
I think it's done,
Thank you for the great and hard work provided @pojienie !

RodneyBaker:
LGTM
Thank you @pojienie for the multiarc feature!

beeheemooth:
@RodneyBaker  exe ?

RodneyBaker:
Link:  https://github.com/opentoonz/opentoonz/releases/tag/nightly

beeheemooth:
@RodneyBaker There is no MultiArc in Geometric Tools

beeheemooth:
@RodneyBaker  "released this on 16 Nov 2016 " are you joking? :)

RodneyBaker:
Yes, that has been an ongoing joke since the nightlies were set up in 2016.
And apparently it remains just as fresh as ever after all these years. ;)

openanim:
It seems like the smooth option is present in all geometry tools. Maybe make it present only on MultiArc tool to avoid confusion in the future.

The smooth option is however active on toonz raster and standard raster levels. But we have no way to deactivate it or activate it as we can do it with the vector level, for now. It would be good for the future.

It seems also to see or not the smooth effect on the rasters levels, it is necessary to activate or deactivate it on the vector level, and a restart seems necessary. Need confirmation about that.

Issue #3365, opened 2020-06-24
Feature Request: Intersecting Freehand Type for Selection Tool


melieconiek:
Looking for freehand selection type that selects the objects inside the lasso as well as the objects it intersects with. In most cases this is a faster and more precise selection mode. :)

An example below: selecting a set of objects and then deselecting another that was inside the selection--and a mock-up of a simple intersecting freehand selection. In nearly all cases, the stroke you have to draw for the selection is smaller and allows you to more easily avoid objects you don't wish to select.

_(In this simplified example you'd probably just select the lines individually, but for more complex drawings this isn't as easy.)_

![a](https://user-images.githubusercontent.com/42298081/85635646-636bcb80-b64c-11ea-97ec-d5e0f9520039.gif) ![b](https://user-images.githubusercontent.com/42298081/85635654-65358f00-b64c-11ea-9433-9870eacc4b03.gif)

![image](https://user-images.githubusercontent.com/42298081/85636083-85198280-b64d-11ea-864d-ec34d22860b3.png)

openanim:
I remember having a similar request, it would be a good marriage to be able to switch with the current behavior which also has a lot of advantages.

PR #3369, opened 2020-06-27
add include intersection option in selection tool

pojienie:
related feature request: #3365 

video: https://youtu.be/YLiPx2pxNRw

this PR adds a "include intersection" option in selection tool. When this option is unchecked, Polyline and Freehand will only select strokes that is completely inside the region drawn. With it checked, it would also select any stroke that touches the drawn line.

I used an option(or a checkbox) instead of adding a "intersect" type so that polyline mode would also benefit from this.

### DIFF
--- a/toonz/sources/tnztools/vectorselectiontool.cpp
+++ b/toonz/sources/tnztools/vectorselectiontool.cpp
@@ -39,6 +39,8 @@ namespace {
 VectorSelectionTool l_vectorSelectionTool(TTool::Vectors);
 TEnv::IntVar l_strokeSelectConstantThickness("SelectionToolConstantThickness",
                                              0);
+TEnv::IntVar l_strokeSelectIncludeIntersection(
+    "SelectionToolIncludeIntersection", 0);
 
 const int l_dragThreshold = 10;  //!< Distance, in pixels, the user has to
                                  //!  move from a button press to trigger a
@@ -63,7 +65,7 @@ FourPoints getFourPointsFromVectorImage(const TVectorImageP &img,
       TStroke *s = img->getStroke(i);
 
       for (int j = 0; j < s->getControlPointCount(); j++) {
-        double thick                           = s->getControlPoint(j).thick;
+        double thick = s->getControlPoint(j).thick;
         if (maxThickness < thick) maxThickness = thick;
       }
     }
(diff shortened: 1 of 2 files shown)


melieconiek:
This is fantastic dude. Thank you!

openanim:
Awesome ! Thank you for giving a lot of love at the vector level that needed it @pojienie  !

DoctorRyan:
Yeah, this just made a great advancement to the vector drawing.  Well done.

RodneyBaker:
LGTM

Observation:  This is directly related to this PR but noting it for future work.
The options in the Selection Tool's toolbar menu force a few options off screen to where the scroll/arrow needs to be used to get them back into view.  This isn't a major issue but some specific options might be used more than others; for instance, line thickness.  As such having those options scroll out of view is not ideal.  Recommend a future UI scrub consider how best to address the space on the toolbars and how best to:  1) Keep all options in view 2) Alternatively, keep those most likely to be used in view.

gab3d:
@pojienie  very nice, thanks!

Could it be made that if you build the lasso counterclockwise/clockwise the selection would automatically work either in one or the other mode?
3ds Max does it that way, and it's great once you get accustomed to.
Plus, we could get rid of the visible option in the toolbar (and perhaps replace it for a Preference, to be able to get the legacy behavior?)

It could also work for Rectangle selection, this time keeping track of the position of the first drawn point (if it's on the left/right of the final selection rectangle).
🤔

melieconiek:
@gab3d The Rectangle selection type already works this way, so it would totally make sense if Freehand did too. I'll plus one on this, it'd allow you to waffle between selection modes on the fly. :)

beeheemooth:
This feature is so useful! I could not even imagine how much this would facilitate the work with a selection of small parts! In my work I use only freehand mode to select vector details, and I had to zoom back to cover all the necessary vectors. Now there is no need to do this, and this is a huge time saving. Thank you so much!

pojienie:
@gab3d Yeah, Rectangle type already do it: If the starting point of the rectangle is left top/bottom point, you won't select intersected strokes. Otherwise you do. I only saw that part of code after I implemented this...

I've made a PR #3401 that implements the clockwise/counterclockwise feature.

PR #3375, opened 2020-07-02
Expression Reference Manager

shun-iwasawa:
This PR will add a new experimental feature to manage references used in the expressions.

### How To Enable This Feature

In Preferences popup, check the `Automatically Modify Expression On Moving Referenced Objects` option in the `Animation` category.

### What This Feature Does

Expression Reference Manager monitors all the fxs / stage objects references used in the expression. It has 2 major features:

#### When the referenced object is moved, the manager automatically adjusts the expression text so that the reference won't break.
- For instance, let's say you have an expression `col1.ew` in some parameter. If you drag&move the Column1 to Column2, the manager automatically change the expression text to `col2.ew`.
- The changes will be reported in the Message Center panel.

#### When you try to do some operation, if it will remove referenced parameter, the manager opens warning popup in advance. 
- For instance, let's say you have an expression `col1.ew` in some parameter (again). When you try to delete the column1 the manager will open the warning.
- If you proceed the operation in spite of the warning, the referencing parameter will become "ignored" state, i.e. the parameter will no longer be monitored or modified with subsequent operations.
  - There is an exception that if the expression will cause circular reference after the operation. In such case the manager will add "?" to both ends of the expression text to avoid it. 
- The ignored parameter is displayed in red color and with warning sign in the function editor so that users can detect it.
- For instance, let's say you have an expression `col1.ew` in a keyframe of Column2 E/W. When you try to delete the column1 the manager will open the warning. If you proceed deleting the column1, the followings will happen:
  - Column2 will move by one to the left and become Column1.
  - Column1 E/W (former Column2 E/W) becomes "ignored" state. i.e. the expression will no longer be modified by the manager.
  - However, leaving the expression text `col1.ew` unchanged will cause circular reference. So the manager will add "?" to the both ends like `?col1.ew?` . It is not proper syntax but can avoid the circular reference.
- When users manually modify any keyframe in the ignored parameter, the parameter will be monitored again and back to normal state.
- When users try to save scene with any parameter being left in the ignored state, the manager will open the warning popup.

### Known Issues
I marked this feature "Experimental" as there are known issues as follows;
- This feature does not cover the `vertex` expression syntax for referencing the plastic skeleton (like `vertex(2,"Vertex __2").angle`) yet, since I'm not so familiar with it.
- Undoing the operation with ignored parameters will not bring back data properly.

I think I made the feature not to affect the exiting usage as long as the preference option is not checked.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -2372,9 +2372,13 @@ SpreadsheetViewer {
   qproperty-KeyFrameColor: #995d1d;
   qproperty-KeyFrameBorderColor: #db9041;
   qproperty-SelectedKeyFrameColor: #a2835b;
+  qproperty-IgnoredKeyFrameColor: #ac2a39;
+  qproperty-SelectedIgnoredKeyFrameColor: #b25872;
   qproperty-InBetweenColor: #666250;
   qproperty-InBetweenBorderColor: #b0aa91;
   qproperty-SelectedInBetweenColor: #717970;
+  qproperty-IgnoredInBetweenColor: #8a695e;
+  qproperty-SelectedIgnoredInBetweenColor: #93807d;
   qproperty-SelectedEmptyColor: rgba(96, 109, 118, 0.5);
   qproperty-SelectedSceneRangeEmptyColor: rgba(96, 109, 118, 0.5);
   qproperty-TextColor: #d6d8dd;
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -2372,9 +2372,13 @@ SpreadsheetViewer {
   qproperty-KeyFrameColor: #995d1d;
   qproperty-KeyFrameBorderColor: #db9041;
   qproperty-SelectedKeyFrameColor: #a2835b;
+  qproperty-IgnoredKeyFrameColor: #ac2a39;
+  qproperty-SelectedIgnoredKeyFrameColor: #b25872;
   qproperty-InBetweenColor: #666250;
   qproperty-InBetweenBorderColor: #b0aa91;
   qproperty-SelectedInBetweenColor: #717970;
+  qproperty-IgnoredInBetweenColor: #8a695e;
+  qproperty-SelectedIgnoredInBetweenColor: #93807d;
   qproperty-SelectedEmptyColor: rgba(90, 100, 106, 0.5);
   qproperty-SelectedSceneRangeEmptyColor: rgba(90, 100, 106, 0.5);
   qproperty-TextColor: #e6e6e6;
--- a/stuff/config/qss/Default/Default.qss
+++ b/stuff/config/qss/Default/Default.qss
@@ -2372,9 +2372,13 @@ SpreadsheetViewer {
   qproperty-KeyFrameColor: #995d1d;
   qproperty-KeyFrameBorderColor: #db9041;
   qproperty-SelectedKeyFrameColor: #a2835b;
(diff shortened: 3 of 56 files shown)


gab3d:
@shun-iwasawa 
thanks for trying and improve the expressions system! 🥰

RodneyBaker:
I've started to test this and already see how this will be a very welcome addition.
Having the changes appear in the Message Center is much appreciated.

RodneyBaker:
Posting initial intent to merge although I'm still exploring and testing more broadly.
As part of that exploration I am thinking of screen capture recording the highlights in order for those seeking more information to be better informed as well as to encourage additional feedback.

shun-iwasawa:
Thank you @RodneyBaker for testing so far & further exploration.
Since this is quite large change, I would like it to be additionally reviewed by some developper in advance of merging in order to make sure at least it will not affect the exiting usage as long as the preference option is OFF. 
So can you please keep it unmerged even if you find it looks good to you? Thanks.

RodneyBaker:
Agreed.  Is @masafumi-inoue available?

RodneyBaker:
This PR is still awaiting review by interested developer.

RodneyBaker:
@shun-iwasawa Recommend this experimental feature's PR be merged to get it into circulation with nightly releases.
This to aide in seeing it fully explored and vetted moving into the v1.5 release.

RodneyBaker:
While awaiting additional review this might be a good time to mention this...

It would be very useful if the field that contains the address of the currently selected object could be selectable for the purpose of copying.

Here is the location of interest:
image

Short addresses aren't a problem but often when working with expressions with channels deep inside FX or Plastic Tool vertexes the addresses can become quite lengthy.   Example:

image

There may be other useful ways of copying channel addresses as well.
Having this particular field's text accessible for copy/pasting however would be very welcome.

Thanks!

gab3d:
@RodneyBaker 
very good suggestion!

shun-iwasawa:
@RodneyBaker Thank you for the good suggestion.

You might have heard already but you can also middle-button drag the item in the tree view or the column head in the spread sheet to get the expression name: 
drag_expression

RodneyBaker:
@shun-iwasawa I was hoping you would clarify that as I had indeed heard you state before that there was a drag and drop option for expressions and so knew I was missing something very important.  You've closed the loop on that now.  Many thanks!

The problem on my end seems to be that my pen isn't set up for a middle button/mouse movement and I don't have a mouse connected to my laptop so could not confirm the drag/drop.   
I need to set up my pen with the middle mouse  button drag today (that or just add my mouse back and be done with it)!

artisteacher:
> The problem on my end seems to be that my pen isn't set up for a middle button/mouse movement and I don't have a mouse connected to my laptop so could not confirm the drag/drop.

I have the same problem - both apple's touchpad & magic mouse don't have a way to middle click. It would be better to have an alternative modifier that everyone can use.

ladlon:
This all sounds great!

beeheemooth:
@RodneyBaker just assign the middle mouse button to one stylus button :)

RodneyBaker:
Already done @beeheemooth.  See commentary above.  :)

gab3d:
It´s great to be able to drag the parameters.
thank you @shun-iwasawa for this tip!

artisteacher:
>  just assign the middle mouse button to one stylus button

That doesn't help if you don't have a stylus - or if you don't have the correct permissions to change stylus settings. 

If a shortcut can't be manually set, it really needs to be accessible on any device.

shun-iwasawa:
I'll merge this considering that the change has been tested with/without activating the feature for several months in our studio and any critical issue have not been reported so far.
If any matters come up, I'll tackle with it right away.

PR #3378, opened 2020-07-03
Fix Crash on Attempting to Save Read-only Studio Palette

shun-iwasawa:

- When saving a studio palette, if the palette file is set to read only, OpenToonz crashes.

The cause of this problem was that the exception thrown from `StudioPalette::save()` was not caught by any higher functions.
In this PR I added missing `try/catch` statements to not only the usage when saving the palette but also all other cases.


RodneyBaker:
LGTM

PR #3381, opened 2020-07-06
Fix Geometric Tool - MultiArc Shape - Smooth Option

shun-iwasawa:
This PR quick-patches the issue reported by @openanim in https://github.com/opentoonz/opentoonz/pull/3360#issuecomment-653891320

- The Smooth option now appears in all level types.
- The option will be disabled unless the "MultiArc" shape mode is selected.

Actually I feel it would be better to hide the unnecessary options (instead of disabling it) in order to save space of the tool options bar.
But for now I made it consistent with `Polygon Sides` field which is also only enabled in "Polygon" shape mode.


RodneyBaker:

openanim:
@shun-iwasawa Thank you very much for the quick fix !

pojienie:
Nice, thank you so much!

PR #3384, opened 2020-07-07
Fix Saving Toonz Raster Level Failure

### LINKED ISSUE #3383: Saving Toonz Raster Level Fails Just After Adjusting Style Color
**Describe the bug**
This issue has been a problem for our studio over the months.
When saving the level, it sometimes fails with the warning popup saying `Can't open file. Access may be denied or someone else may be saving the same file. Please wait and try again.` Although no one else open the same file.
After investigation, it is found that the problem tends to occur with a level with large numer of  frames, and when user attemps to save level just after modifying the style color of its pallete.

**To Reproduce**
1. Create toonz raster level with large number of frames, (let's say, 100 frames)
1.  Draw on some frame.
1. Adjust some style color. And immediately save the level.

OT sometimes fails to save level with the warning popup.

**Expected behavior**
The level should be saved properly.

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR will fix #3383 

The probable cause of the saving failure was that the file was opened by another thread for updating the icons in the Level Strip in response to the style change.

Before the fix, OT always invalidated and re-generated all icons in the level when a style of its palette was changed (in the function [notifyPaletteChanged()](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/tapp.cpp#L637) ).
After further investigation, I came up with the fact that re-generating level strip icons is not always needed:

- For **Raster Levels**, the icons do not need to be changed. ( The images does not change with the style color change in the `Raster Drawing Palette` )
- For **Toonz Raster Levels**, the icons do not need to be re-generated, because the icons are cached as color-mapped image and the current palette is applied just before using it. (in the function [getIcon()](https://github.com/opentoonz/opentoonz/blob/0dc08facf261210e097186e980ea2ed65dc805c8/toonz/sources/toonzqt/icongenerator.cpp#L98-L99) in icongenerator.cpp ) It only needs to trigger update of relevant panels.
- For **Toonz Vector Levels**, the icons still need to be re-generated since they are cached as raster image, including colors and shapes defined by styles.

So I modified `notifyPaletteChanged()` in tapp.cpp to decide if the icons should be re-generated according to the level types.

This change will reduce unnecessary file access. It will enhance the responce and will minimize the risk of failure on saving levels.

### DIFF
--- a/toonz/sources/image/tzl/tiio_tzl.cpp
+++ b/toonz/sources/image/tzl/tiio_tzl.cpp
@@ -40,7 +40,7 @@ char *reverse(char *buffer, int size) {
   }
   return buffer;
 }
-}
+}  // namespace
 
 static int tfwrite(const char *data, const unsigned int count, FILE *f) {
   return fwrite(data, sizeof(char), count, f);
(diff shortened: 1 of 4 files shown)


RodneyBaker:
LGTM

PR #3386, opened 2020-07-07
fix a bug where moving a control point while holding alt has unintended result

### LINKED ISSUE #3332: Control Point editor bug alt_click
**Describe the bug**

The CPE tool is very weird lately. When you want to create a corner point, it advances the point by several millimeters. A second click continues to move it forward instead allow it to find its handles. It occasionally follows a crash.

**To Reproduce**
Itseems random but happens  fairly regularly

**Screenshots**

CPE_issue
**Device Information:**
 - OS: W10
 - OpenToonz Version:  Nightly jun 05 2020

pojienie:
this fixes #3332

To reproduce this bug:
1. select control point editor tool
2. select a stroke
3. move a control point
4. move another control point while holding alt

The bug happens because `m_action` isn't initialized when you press a control point while holding alt. So, if you drag a control point, `m_action` is initialized as `CP_MOVEMENT`. Then, when you hold alt and drag a control point again, it would then misbehave.

### DIFF
--- a/toonz/sources/tnztools/controlpointeditortool.cpp
+++ b/toonz/sources/tnztools/controlpointeditortool.cpp
@@ -652,6 +652,7 @@ void ControlPointEditorTool::leftButtonDown(const TPointD &pos,
     m_selection.makeCurrent();
   } else if (pointType == ControlPointEditorStroke::CONTROL_POINT) {
     if (e.isAltPressed()) {
+      m_action = NONE;
       m_selection.selectNone();
       m_selection.select(pointIndex);
       initUndo();


RodneyBaker:
@pojienie I confirm the problem and that this PR fixes it (for windows).

@shun-iwasawa / @manongjohn I'm not sure why Travis is failing but it seems unrelated to the PR itself  (REF:  https://travis-ci.org/github/opentoonz/opentoonz/jobs/705715939)

manongjohn:
The `ci-scripts/osx/travis-build.sh` is detecting multiple versions of QT on the macos image for some reason and the logic I added a while back doesn't handle that properly. For cases like this, the script needs to be fixed to select the highest version it finds since that is what Homebrew installs.

openanim:
Thank you very much !

RodneyBaker:
Travis is back on the job and this PR... LGTM
Thank you @pojienie for the bug fix!

PR #3387, opened 2020-07-07
Fix Overwriting Raster Drawing Palette

shun-iwasawa:
- When pressing "Save Palette" button in the palette viewer's tool bar, the raster level palette (`+palettes/Raster_Drawing_Palette.tpl`) is not overwritten properly. OT suggests to save palette to the wrong location, in the same folder as the level files (like `+drawings/A..tpl` for `+drawings/A..tif`) instead.


RodneyBaker:
LGTM
Posting intent to merge.

PR #3389, opened 2020-07-08
Fix travis-install

shun-iwasawa:
This PR is an attempt to fix recent failures in Travis CI - MacOS build.

According to the command log it seems that the older version of Qt (5.13.2) remains together with the latest version (5.15.0) for unknown reasons and it causes failure to set `QT_PATH` properly.


shun-iwasawa:
Confirmed the Travis CI back to normal. I'll merge it by myself as it is quite obvious.

PR #3390, opened 2020-07-08
Fix Viewer Controls Assert Crash (Debug)

konero:
Using viewer controls causes a crash in debug.


RodneyBaker:
LGTM
Thank you for the fix @konero!

PR #3391, opened 2020-07-08
Fill Audio Waveform

konero:
Fill the waveform instead of outlining it.

Soundwave


RodneyBaker:
Posting intent to merge...

This PR is simple and direct but it will be good to allow time for feedback should anyone wish to comment on the change.

beeheemooth:
I hope that the waveform is not vector, otherwise the fill will disappear from time to time ;)

PR #3395, opened 2020-07-09
fix undo data when drawing arc and mutliarc

pojienie:
Oops.

The bug is that, if you draw with Arc or a closed stroke with MultiArc, there would be one too many undo data. So, when you undo, it would set the tool's data instead of removing the stroke you just drew.


RodneyBaker:
LGTM
Thank you for the fix @pojienie 

Will merge this momentarily after the PR currently being merged completes its nightly build.

shun-iwasawa:
Jenkins

PR #3397, opened 2020-07-10
Mode Sensitive Tag for Fx Settings Layout

shun-iwasawa:
This PR enables Fx Settings to hide the unnecessary controls according to its mode specified by another control of the same Fx.

mode_sensitive_box

Some Fx has an ability to render several types of images according to its "mode". And some Fx parameters are only used in a specific mode.
For instance, the `Text` field in the Text Iwa Fx is used only when the `Source` combo box is set to `Input Text` mode.
In another instance, the `Box Color` field in the TimeCode Iwa Fx is needed only when the `Show Box` checkbox is checked.

In such case, this feature will hide unrelated controls and can keep the UI simpler.

### Usage

This feature can be added by adding special tag like as follows to the layout XML file in `$TOONZPROFILE/layouts/fxs` .

```
<vbox modeSensitive="mode_controller_name" mode="mode_indices">
  <control>control_to_be_mode_sensitive</control>
  <control>another_control_to_be_mode_sensitive</control>
   . . .
</vbox>
```

where;
- **mode_controller_name** : Specify the name (parameter id) of control to change the mode. It must be either a combo box (for `TIntEnumParam`) or a checkbox (for `TBoolParam`) .
- **mode_indices** : Specify the mode index in which the controls will be displayed. 
  - If the controller is a combo box, it will be the series of values of `TIntEnumParam` separated by comma "," (N.B. it can be different from the combobox' item indices). 
  - If the controller is a the checkbox, it will be either 1 (=on) or 0 (=off).

For trial, I modified the fx settings layout of Text Iwa Fx and TimeCode Iwa Fx.

### DIFF
--- a/stuff/profiles/layouts/fxs/STD_iwa_TextFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_iwa_TextFx.xml
@@ -2,8 +2,12 @@
 	<page name="Text Iwa">
  	<vbox>
 		<control>targetType</control>
-		<control>columnIndex</control>
-		<control>text</control>
+		<vbox modeSensitive="targetType" mode="1">
+			<control>columnIndex</control>
+		</vbox>
+		<vbox modeSensitive="targetType" mode="2">
+			<control>text</control>
+		</vbox>
 		<control>hAlign</control>
 		<control>center</control>
 		<control>width</control>
--- a/stuff/profiles/layouts/fxs/STD_iwa_TimeCodeFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_iwa_TimeCodeFx.xml
@@ -2,13 +2,17 @@
 	<page name="TimeCode Iwa">
  	<vbox>
 		<control>displayType</control>
-		<control>frameRate</control>
+		<vbox modeSensitive="displayType" mode="0,2">
+			<control>frameRate</control>
+		</vbox>
 		<control>startFrame</control>
 		<control>position</control>
 		<control>size</control>
 		<control>textColor</control>
 		<control>showBox</control>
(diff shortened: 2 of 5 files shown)


RodneyBaker:
Nice addition! That will streamline things considerably.

While you are in the FX dialogue code can you look to see if there is an equivalent to a blank line or CRLF (carriage return line feed) for text or to assist with spacing of controls?  There may be something like that already but I didn't see it when I was surveying the options.

An example usage for such a thing (amongst other things) might be a tab that mostly contains text for informational purposes such as a Credits tab or a description of various options that work well with the FX (in addition to that of the help file, of course).

shun-iwasawa:
@RodneyBaker 

> While you are in the FX dialogue code can you look to see if there is an equivalent to a blank line or CRLF (carriage return line feed) for text or to assist with spacing of controls? There may be something like that already but I didn't see it when I was surveying the options.

For now the fx settings cannot display multiple lines. Instead, the `<label>` tag can be exploited to make blank line by setting empty string in it ( like `<label>""</label>` ).

> An example usage for such a thing (amongst other things) might be a tab that mostly contains text for informational purposes such as a Credits tab or a description of various options that work well with the FX (in addition to that of the help file, of course).

The plugin fxs are [taken aside](https://github.com/opentoonz/opentoonz/blob/5d6d967c5fdfb975b182e440fb8c4a91548f5919/toonz/sources/toonzqt/pluginhost.cpp#L627-L658) and having such information tab (the `Version` tab) since they can be under different copyright / conditions.

RodneyBaker:
Thank you @shun-iwasawa for the addition!

PR #3401, opened 2020-07-12
[Do Not Merge] include intersect when drawing counterclockwise

pojienie:
based on this comment: https://github.com/opentoonz/opentoonz/pull/3369#issuecomment-657152672

video: https://youtu.be/YNpGgJa3Khs

I followed the answer on this link: https://stackoverflow.com/questions/1165647/how-to-determine-if-a-list-of-polygon-points-are-in-clockwise-order to implement the algorithm for detecting if a stroke is clockwise or counterclockwise.

this PR removes "include intersection" option in selection tool. Now, instead of checking an option to include intersection, you draw counterclockwise if you want to include intersection. If you don't want to include them, draw clockwise.

I chose not to add a preference option because 1.4 doesn't have this feature so I don't need to keep it backward compatible. (also because I'm lazy :p )

### DIFF
--- a/toonz/sources/tnztools/vectorselectiontool.cpp
+++ b/toonz/sources/tnztools/vectorselectiontool.cpp
@@ -128,6 +128,25 @@ inline void notifySelectionChanged() {
   TTool::getApplication()->getCurrentSelection()->notifySelectionChanged();
 }
 
+static bool isClockwise(TStroke *stroke) {
+  assert(stroke);
+
+  std::vector<TThickPoint> points;
+  stroke->getControlPoints(points);
+
+  if (points.size() < 3) {
+    return false;
+  }
+
+  double area = 0;
+  for (int i = 0; i < points.size() - 1; ++i) {
+    area += (points[i + 1].x - points[i].x) * (points[i + 1].y + points[i].y);
+  }
+  area += (points[0].x - points.back().x) * (points[0].y + points.back().y);
+
+  return area > 0;
+}
+
 }  // namespace
 
 //********************************************************************************
@@ -1206,7 +1225,6 @@ public:
 VectorSelectionTool::VectorSelectionTool(int targetType)
     : SelectionTool(targetType)
     , m_selectionTarget("Mode:")
-    , m_includeIntersection("Include Intersection", false)
     , m_constantThickness("Preserve Thickness", false)
     , m_levelSelection(m_strokeSelection)
     , m_capStyle("Cap")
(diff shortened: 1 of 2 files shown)


melieconiek:
The selection tool doesn't take into account the viewer being flipped, neither does the rectangle type for that matter.

Excellent stuff, though! Thank you for this. :)

beeheemooth:
Oh no, please don't do this.
 I am very happy with the current behavior of the tool you made. I always select vectors in a clockwise direction (it’s more convenient for me) and would not want to lose this wonderful feature. Is disabling the option not enough for the user? Please think. Or make this option a preference. Thank.

beeheemooth:
In practice, there are situations where the selection of vectors depends on the direction, for example, when you bypass some vectors. I'm afraid that in this case, users will be confused by the result.

beeheemooth:
ezgif-4-e68ce07d61de

This is a simple case of selecting vectors. In practice, situations can be much more complicated. (remember that if you make a 
mis-click, you will need to select everything in a new way) This feature with the direction of stretching the frame is typical 
for highlighting ONLY in Rectangle Mode, since there was no Select All function before and it helped to easily select all vectors.

The selection of numerous similar vector elements (horizontally and vertically mirrored) in different places in this way will require the user to constantly concentrate on the orientation.

Do not forget that the column can be scaled in the negative direction (H or V: -100), the selection of vectors will need to be carried out in the reverse order, and this will be very confusing for users.

pojienie:
@melieconiek Yeah, there is no flipped viewer detection anywhere in the selection tool as far as I know.

@beeheemooth 

The web demonstration is excellent!

I do agree it can be confusing, however, the original rectangle type already behaves somewhat like this (from left to right, no intersected strokes. From right to left, contains intersected strokes). This PR makes it so that they are confusing in the same way. I agree this sounds silly, though.

A preference option seems to be the best way to do this. It could take care of users who wants to select intersected strokes only when a checkbox is checked and users who want to be able to switch between them just by changing the way they select.

beeheemooth:
@pojienie Thank you for your work and understanding!

melieconiek:
@pojienie I meant that flipping the viewer causes the direction you have to draw with the selection tool to flip as well, which is confusing. It should stay the same even when the viewer is flipped. Thank you for you work on this. :)

beeheemooth:
I really hope that after this PR it would never occur to anyone to do the same with Freehand / Segment Eraser.
 Anything but that. :)

yokobell:
Thanks for your work! I really enjoy the features you make.

gab3d:
great addition @pojienie ! 
thanks!! 😊

gab3d:
I also think that is weird than when the viewer is inverted the selection directions get inverted too. 🤷🏻‍♂️

PR #3402, opened 2020-07-12
Color_0 chip redesign

DoctorRyan:
This changes  the default color_0 chip to display a checkered pattern all the way to  the top.  This is for users who keep confusing the button for white.  Studio employee use probably knows what Color_0 stands for already. The change only affects the default chip.  All chips created afterwards with Color_0 will maintain the old style.

Co-Authored-By: turtletooth <jcbullock@gmail.com>

### DIFF
--- a/toonz/sources/common/tvrender/tsimplecolorstyles.cpp
+++ b/toonz/sources/common/tvrender/tsimplecolorstyles.cpp
@@ -590,7 +590,7 @@ void TSolidColorStyle::makeIcon(const TDimension &size) {
     else {
       TRaster32P fg(size);
       fg->fill(premultiply(col));
-      TRop::checkBoard(m_icon, TPixel32(200, 200, 200), TPixel32::White,
+      TRop::checkBoard(m_icon, TPixel32(180, 180, 180), TPixel32(230, 230, 230),
                        TDimensionD(6, 6), TPointD());
       TRop::over(m_icon, fg);
     }
--- a/toonz/sources/common/tvrender/tsimplecolorstyles.cpp
+++ b/toonz/sources/common/tvrender/tsimplecolorstyles.cpp
@@ -591,7 +591,7 @@ void TSolidColorStyle::makeIcon(const TDimension &size) {
       TRaster32P fg(size);
       fg->fill(premultiply(col));
       TRop::checkBoard(m_icon, TPixel32(130, 130, 130), TPixel32(200, 200, 200),
-                       TDimensionD(9, 9), TPointD());
+                       TDimensionD(6, 6), TPointD());
       TRop::over(m_icon, fg);
     }
   }
--- a/toonz/sources/common/tvrender/tsimplecolorstyles.cpp
+++ b/toonz/sources/common/tvrender/tsimplecolorstyles.cpp
@@ -590,8 +590,8 @@ void TSolidColorStyle::makeIcon(const TDimension &size) {
     else {
       TRaster32P fg(size);
       fg->fill(premultiply(col));
-      TRop::checkBoard(m_icon, TPixel32(130, 130, 130), TPixel32(200, 200, 200),
-                       TDimensionD(9, 9), TPointD());
+      TRop::checkBoard(m_icon, TPixel32(200, 200, 200), TPixel32::White,
+                       TDimensionD(6, 6), TPointD());
       TRop::over(m_icon, fg);
     }
   }
--- a/toonz/sources/common/tvrender/tsimplecolorstyles.cpp
+++ b/toonz/sources/common/tvrender/tsimplecolorstyles.cpp
@@ -590,8 +590,8 @@ void TSolidColorStyle::makeIcon(const TDimension &size) {
     else {
       TRaster32P fg(size);
       fg->fill(premultiply(col));
-      TRop::checkBoard(m_icon, TPixel32::Black, TPixel32::White,
-                       TDimensionD(6, 6), TPointD());
+      TRop::checkBoard(m_icon, TPixel32(130, 130, 130), TPixel32(200, 200, 200),
+                       TDimensionD(9, 9), TPointD());
       TRop::over(m_icon, fg);
     }
   }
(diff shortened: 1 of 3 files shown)


konero:
As my colleague pointed to me, you can paint white using color_0 chip on raster levels, his reasoning was that they would fix imported frames using color_0 (white) and color_1 (black), that color_0 is guaranteed to be white being a good thing.

I didn't actually know that myself. I'm just posting the point that was made in any case, since it is a good one.

RodneyBaker:
As this is a graphical UI element it will be good to post an image of what this PR has the Color_0 chip look like:

color_0

I've added a Color_2 chip that displays the current look of the Color_0 chip for comparison.

I personally don't have a strong preference one way or the other.
Let's see if we can't get good enough feedback for this to make **a decision by 15 July** to press forward with the change or set it aside for future consideration.

yokobell:
Hi! I have the same opinion about painting white, I thought color_0 means white on raster? And when you run a cleanup task to convert to TLV since all white is removed the swatch becomes transparent keeping the same workflow. Have I got that wrong and this is a bug? So if this change was accepted won't that actually confuse more people? Since at the moment the swatch shows both properties (it is white or it is transparent depending on your level type).

The alignment of the squares seems different and broken looking: image image

It's more difficult to see which swatches have lower alpha than before, this is with alpha 245 (current nightly) vs this change with alpha of 160 so you can see there is a big margin of difference:
image image

The alpha thing is not a big problem it was just nice that you could see which swatch had lower alpha even by a little.

DoctorRyan:
Thanks for all the feedback.  It was a request that was made because of the confusion issues I was talking about.  So going forward is there a better way to make sure users don't get confused.  Ideas are appreciated.

artisteacher:
If it's only raster levels that use Color_0 chip as white, perhaps on the raster palettes, it should be white. On the toonz levels it should be transparent. Since the levels use different palettes, and Color_0 behaves differently depending on the level, I don't see why it needs to display the same way for each palette type.

DoctorRyan:
Annotation 2020-07-13 091017

So I reset the squares back to default.

jrhyme:
No letters are visible on the transparent frame.

shun-iwasawa:
@konero 

>As my colleague pointed to me, you can paint white using color_0 chip on raster levels, his reasoning was that they would fix imported frames using color_0 (white) and color_1 (black), that color_0 is guaranteed to be white being a good thing.

Wow. I didn't know that either. I checked the source and understand that the alpha value of styles are always ignored in the raster brush, no matter whether the style is `color_0` or not. Instead, the `Opacity` value in the tool option is used to produce pressure-sensitive opacity in the strokes. I don't think it was intentional that the `color_0` consequently becomes the white brush, but it seems to be working nicely.. :-)

RattusaurusRex:
Comparison of the large chip alpha checkerboards (all alphas the same value for each color) in OpenToonz, this PR, and Tahoma:

colortest

DoctorRyan:
@shun-iwasawa My final revision.  Thanks to everybody for their contributions.
Annotation 2020-07-14 114402

shun-iwasawa:
Hmm I'm not sure that your final revision could resolve the original intention of this PR.
> This is for users who keep confusing the button for white.

I don't come up with an ideal solution, though..

DoctorRyan:
@shun-iwasawa I am open to suggestions.  The only thing that was changed was just the color of the checkerboard.  Nobody is really on board with a full checkerboard pattern either.

shun-iwasawa:
@DoctorRyan Thank you for tackling with the UI improvement.
Here I tried to figure out my suggestion:

- Keep the checkerboard colors of styles other than #0 unchanged from black & white, in order to make it easier to recognize the semi-transparency.
- Fill the style #0 chip with the checkerboard pattern to clarify that it is transparent color for the empty area. 
- For the style #0 chip, use `Chessboard Color 1` and `Chessboard Color 2` in the `Preferences > Colors` for the same reason. 
- For the style #0 chip, fill white color behind the style name so that the user can read it and know the RGB value of it.

The drawback of my suggestion would be that it will lose consistency of checkerboard colors between style#0 and other styles.

Here is a mock-up: (based on the comparison image of @RattusaurusRex 's post. Thanks!)

style0_draft

Any feedback would be appreciated.

artisteacher:
Is there a darker grey that would perform better for translucent styles but still read as transparent for color_0? 

I still think color_0 should display as white on raster palettes only - if it's going to behave as white.

DoctorRyan:
The idea of leaving the other colors black in white is an excellent idea as one of the problems with the PR was being able to see the alpha being reduced and increased better.

gab3d:
Thanks for improving this!

I also think the color_0 chip should be full white on the "Raster Drawing Palette".

yokobell:
Although it makes the text readable, it is a shame that the overall look will be inconsistent, it feels a bit like it has been selected in the corner of my eye and I should pay attention to it in some way.

How about excluding large thumbnail from these changes and applying it to just small and medium? That amount of checkerboard pattern on the screen is a bit uncomfortable to look at.

In Harlequin the whole swatch is checkerboard but the max size is the mid size in OpenToonz, even then part of the swatch is taken up by the name field, so it doesn't seem as busy as the large thumbnail view in OpenToonz.

image

DoctorRyan:
@yokobell That is kinda the issue with this PR.  Trying to find a way to communicate the idea in Toonz Raster (Hey this is not white!) while not overloading your senses with harsh images (full checkerboard).  @artisteacher I am looking into creating separate entries for Raster and Toonz Raster for Color 0 but I am still new to a lot of this.  The code in Opentoonz is like the rabbit hole in Alice in Wonderland.  It just keeps going and going and going.............................

DoctorRyan:
So I am stuck on this PR @shun-iwasawa.  I am still learning the code in OT so I can't figure out how to separate the colors of the checkerboard and build the white box.  I did make the color changes you requested by making the chip the same as the preferences colors though.  So this is a shoutout to any devs on here.  I need some help with this PR.  Any support would be appreciated.  I would like to see this PR finished as I think it does have the potential to help a lot of people.

RodneyBaker:
Tagging with 'hiatus' pending renewed interest.

PR #3406, opened 2020-07-15
Make the New Style Button Optional

shun-iwasawa:
This PR modifies the New Style Button feature (which was recently added in #3394)  as follows:
- Made the button optional since users may not would like to create a new style in some workflow. Now it is hide-able from the hamburger menu in the Palette panel toolbar. The default value is ON (visible).
- Slightly shrank the button size to be the same size as style chip.
- Made the plus ("+") mark to be with the same color as the text color of the current theme. Note that I made the color semi-transparent since I thought it was too prominent with full opacity.

image

### DIFF
--- a/toonz/sources/include/toonzqt/paletteviewer.h
+++ b/toonz/sources/include/toonzqt/paletteviewer.h
@@ -166,6 +166,7 @@ protected slots:
   void setIsLocked(bool lock);
 
   void onSwitchToPage(int pageIndex);
+  void onShowNewStyleButtonToggled();
 };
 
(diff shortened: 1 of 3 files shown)


turtletooth:
This looks good.  Thanks for tweaking it.  I added your paint changes to Tahoma:
https://github.com/turtletooth/tahoma/pull/89

RodneyBaker:
LGTM
Thank you  @shun-iwasawa!

DoctorRyan:
Thank you for making the changes on the button Shun.  It looks very nice.

PR #3410, opened 2020-07-20
Fix Loading Pegbars' Expression To Refer Parameters Properly 

### LINKED ISSUE #3405: Expression Fails to Refer Parameter on Loading Scene
**Describe the bug**
Expression fails to refer parameter just after loading the scene under some conditions as follows:
1. When the stage objects' (like columns, cameras, and pegbars) parameter refers an fx parameter.
1. There is another parameter with keyframe and such keyframe is not at the same frame as the first one.

With this problem the scene fails to render the scene properly when you use batch rendering with `tcomposer`.

**To Reproduce**
1. Create Blur Fx.
1. In the Function Editor, set **Camera1 E/W** keyframes from **frame 1** to **frame 3** with the expression interpolation. Set the expression to refer the `Value` parameter of the Blur Fx ( `fx.blur01.value` ). _[This is making the condition 1]_
1. Set **Camera1 N/S** keyframe at **frame 2** with arbitrary interpotation type.  _[This is making the condition 2]_
1. Make sure the Function Editor looks like this:  
  expression_before
1. Save the scene with any name.
1. Load the same scene. Check if **Camera1 E/W** could refer the parameter properly.

In my test, it fails to refer the parameter and the function editor shows 0:
expression_after

**Expected behavior**
The expression should refer the parameter properly.

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR will fix #3405 

I have commented-out the line calling `updateKeyframes()` in `TStageObject::loadData()` .
Calling `updateKeyframes()` before completing loading the scene may cause failure to load expressions properly especially if it refers to a curve which is to be loaded lately in the loading scene process.
So I removed it assuming that `updateKeyframes()` will be called when it is actually needed as it is managed under the lazy data model.

### DIFF
--- a/toonz/sources/toonzlib/tstageobject.cpp
+++ b/toonz/sources/toonzlib/tstageobject.cpp
@@ -258,8 +258,8 @@ bool setKeyframe(const TDoubleParamP &param, const TDoubleKeyframe &kf,
 
   TDoubleKeyframe kfCopy = kf;
 
-  kfCopy.m_frame                        = frame;
-  if (easeIn >= 0.0) kfCopy.m_speedIn   = TPointD(-easeIn, kfCopy.m_speedIn.y);
+  kfCopy.m_frame = frame;
+  if (easeIn >= 0.0) kfCopy.m_speedIn = TPointD(-easeIn, kfCopy.m_speedIn.y);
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #3411, opened 2020-07-21
New Expression Function: Periodic Random

shun-iwasawa:
This PR will add a new function for expression; Periodic Random.
It is based on conventional `random` function with enabling to specify "period" in which the random value is set. The value will be linear-interpolated between the periodic intervals.

### Syntax
```
random_p(period,min,max)
```
where `period` is the intervals of the random value (in frames), `min` and `max` specifies the range of the value.
You can also specify the random seed by the following function;
```
random_ps(period,seed,min,max)
```
where `seed` is the random seed.

The arguments `min` and `max` are optional and the term `random` can be shortened as `rnd`.
Here are all variations of this function:
```
random_p(period) = random_p(period,0,1)
random_p(period,max) = random_p(period,0,max)
rnd_p(period,min,max) = random_p(period,min,max)
rnd_p(period) = rnd_p(period,0,1)
rnd_p(period,max) = rnd_p(period,0,max)
random_ps(period,seed) = random_ps(period,seed,0,1)
random_ps(period,seed,max) = random_ps(period,seed, 0,max)
rnd_ps(period,seed,min,max) = random_ps(period,seed,min,max)
rnd_ps(period,seed) = rnd_ps(period,seed, 0,1)
rnd_ps(period,seed,max) = rnd_ps(period,seed,0,max)
```

### Usage
- It can generate random pattern with long period.
- Unlike the conventional `random` function, `random_p` interpolates values between random numbers in **frames with decimals** like frame 3.5. 

  The decimal-numbered frame is rarely used most of the times, except in the **Motion Blur Iwa Fx**.
  If you would like to add shaky motion to the camera (due to earthquake or explosion, etc.) and to apply motion blur to the scene along with the random movement of the camera, use `random_p` instead of `random` in order to get proper trajectory.

motion_blur

### DIFF
--- a/toonz/sources/common/expressions/tgrammar.cpp
+++ b/toonz/sources/common/expressions/tgrammar.cpp
@@ -267,7 +267,8 @@ public:
 
 //-------------------------------------------------------------------
 
-class RandomNode final : public CalculatorNode {
+class RandomNode : public CalculatorNode {
+protected:
   std::unique_ptr<CalculatorNode> m_seed, m_min, m_max, m_arg;
 
 public:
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thank you @shun-iwasawa for the new expression/function!

F-Burning:
Wow! amazing features : D

It's possible make a transition with the random expression with keyframes to 0?

For example a strong shake, and then a slow transition to 0 values.

RodneyBaker:
HI @F-Burning 
Yes, you should be able to diminish the value by a number of different means.
As my math skills are not what they should be I would probably first experiment with something simple such as subtracting the frame number (or the frame number times another factor) so that as the transition continues the shake values lower and lower until they reach zero.

As a simple example, I like to use values like *.5 to halve values.
If we wanted to divide by the number of the current frame we might use \frame or ... halve the value of the current frame allowing for softer/slower transition...  \(frame*.5).

There are other and better approaches I'm sure but that is to get you moving in the right direction.

F-Burning:
Thank you @RodneyBaker!

PR #3421, opened 2020-07-31
Linear color space option for Add Fx Ino

shun-iwasawa:
This PR will enable Add Fx Ino to compute in linear color space.
Which will make the result less-saturated at bright part and more realistic in some cases.

### New parameters

- **Linear Color Space** : When ON, Images will be converted to linear color space before being added. The default setting is OFF. 
- **Gamma** : Specifies gamma for encoding / decoding color space. The default value is 2.2 . 
- **Source is Premultiplied** : When ON, image’s color data will be divided by the alpha before converting the color space (i.e. image will be unpremultiplied), and then multiply by the alpha afterwards. This will correct the color of the semi-transparent pixels in most cases. The default setting is ON. (since Fx input is usually premultiplied.) 

This commit is based on source for the [ComposeAdd plugin fx](https://github.com/opentoonz/dwango_opentoonz_plugins/blob/master/doc/sample_plugins_manual.md#composeadd) by DWANGO Co., Ltd. (Thus necessary license notice are included in this PR as well.) The major differences from the original ComposeAdd fx are as follows;
- Added the `Source is premultiplied` option.
- Modified the transfer functions to use standard gamma correction.


RodneyBaker:
Thanks for this addition @shun-iwasawa.  This will help to avoid that overexposure we often see when using the node!

Informational question(s) here...
The linear color space processing in the Add Fx Ino node only relates to the specific process of compositing but I'm curious to know what the process uses when the linear color space option is not turned on.  Is that on an image to image basis?

The actual question I was initially starting to formulate was whether this node might now allow us to output images in that state of still being in linear color space but I confess I don't even know how Opentoonz processes images regularly.  Perhaps it is already processed in linear color space and this update just extends the Add Fx Ino processing in that way?

As an example, if we set gamma to 1.0 that will only effect the processing of the Add operation and not leave the complete output in linear color space, correct?

By default does opentoonz process images in linear color space?

shun-iwasawa:
> By default does opentoonz process images in linear color space?

Basically OpenToonz processes images in non-linear color space, which means that RGB channel values are composited as-is.
Some Fxs (like Bokeh Fx Iwa, Motion Blur Fx Iwa, Adjust Exposure Iwa, etc. and the new option introduced in this PR) are using linear color space in order to obtain physically-sound rendering result. Each Fx converts color channels to linear color space only for internal computation of the fx and convert them back to non-linear before passing to the subsequent fx.

AFAIK Photoshop and After Effects work in sRGB non-linear color space by default. Nuke works in 32-bit linear color space.

I'm planning to enhance OT to work in linear color space throughout the entire fx schematic like Nuke... in some distant future.

The reason why I set the default value of the `Linear Color Space` option to OFF is to keep backward compatibility.

RodneyBaker:
Thank you @shun-iwasawa 
That was just the answer I was looking for and you've provided great insight into the process both current and planned.
Much appreciated.

Regarding this PR... LGTM!

PR #3426, opened 2020-08-06
Skip Unnecessary Icon Invalidation

shun-iwasawa:

**To reproduce:**
- Open or create Toonz Raster level with relatively large amount of frames (say, 100 frames).
- Edit the level.
- Toggle some checking option like `Paint Check` or `Ink Check`.
- Immediately try to save the level.

Saving sometimes fails with the warning saying `Can't open file. Access may be denied or someone else may be saving the same file.`

After introducing  #3192 OT regenerates all icons when toggling the check. Regeneration is done by sub threads. The threads grab the level files through their tasks and prevent the main thread to access and save the same level properly.

For Toonz Raster Levels, icons are cached as color-mapped raster. Therefore, they don't need to be regenerated for reflecting the palette changes (like toggling Paint Check which only alters the some style color to red.) And Raster level icons does not need to be updated as the checks have no effect to them. 
So I skip unncessary icon invalidation at line 302 in sceneviewer.cpp, the same modificaiton applied in  `TApp::onLevelColorStyleSwitched()` as well.

### DIFF
--- a/toonz/sources/toonz/sceneviewer.cpp
+++ b/toonz/sources/toonz/sceneviewer.cpp
@@ -297,9 +297,9 @@ void invalidateIcons() {
   s.m_paintIndex = mask & ToonzCheck::ePaint ? tc->getColorIndex() : -1;
   IconGenerator::instance()->setSettings(s);
 
-  // Force icons to refresh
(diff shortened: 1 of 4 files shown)


RodneyBaker:
LGTM

PR #3428, opened 2020-08-06
Fix Frame Range Fill With Tablet

### LINKED ISSUE #3247: Anti-aliasing as an FX node
This is a request for the addition of an anti-aliasing FX node.

OpenToonz is already capable of applying anti-aliasing to raster images.
I am assuming (perhaps wrongfully so) that it would not be hugely difficult to create an FX node that utilizes at least some of this same code.

OpenToonz applies anti-aliasing at a fixed point in the rendering process, this means that the user is unable to work with the aliased source image.
Plugins to smooth images have been created for After Effects, these plugins are able to be placed anywhere in a layer's effects stack. This allows the user to work with the aliased source images for as long as they need to.

Here is an example of something that would be far harder or outright impossible to achieve in OT compared to After Effects:
AE Blending

You can see that the white square is being blended with Dwanko's clothing, these are two separate images that have been layered on top of each other.
As the blending effect works based on pixel colors and is designed to work with aliased images, such as result is possible.
Finally, anti-aliasing is applied, producing a smoothed result.
Without the ability to apply effects to the aliased source image, getting artifact-free results would be difficult due to the increased amount of colors produced by anti-aliasing.

In conclusion, having the ability to apply anti-aliasing when needed, enables a workflow that is designed around the ability to work with aliased images.
Such a workflow can be seen in the Japanese Animation Industry where After Effects is commonly used for compositing.
Having more direct access to the pixel data of images allows for more complex compositing and VFX tasks to be achieved.
There are tradeoffs between OT's approach and the approach used in the above example:
OT, though more limited, does a good job of dealing with anti-aliased images as long as they are ToonzRaster or ToonzVector.
After Effects, though lacking the ability to work as cleanly with anti-aliased images, is more flexible as it is designed around manipulating regular raster images.

### LINKED ISSUE #3263: Red Cross Bug
In Toonz Raster Levels, when I use the Fill Tool with Frame Range mode, the red cross does not turn off (and thus the operation does not end) if I fill the color in more than 11-15 frames using the stylus. There is no such problem if I fill in the color using the mouse.

ezgif-2-40817416e30b

**Device Information:**
 - OS:  Windows 8.1
 - OpenToonz Version: OT 1.4 Apr 24 2020
 - **Tablet; Wacom Intuos 4**

### LINKED ISSUE #3427: Unwanted Point Selection After Frame Range Fill With Tablet
**Describe the bug**
After finishing the frame-range filling by using tablet, some point becomes unexpectedly selected as if the user clicks the first point for another frame-range filling.
Note that this problem should occur only in Windows.

**To Reproduce**
1.  Create or load a Toonz Raster level with relatively large amount of frames (say, 100 frames).
1.  Select Fill Tool, activate `Frame Range` option, `Normal` type and `Areas` mode.
1.  Try frame-range filling from 1 to 100 frame **by using tablet.**

**Expected behavior**
No points selected after finishing the fill.

**Device Information:**
 - OS: Win10
 - Tablet: Wacom Intuos PTK-450

shun-iwasawa:
This PR will fix #3427.
Also fixes #3263.

On Windows, using Tablet triggers both `tabletEvent` and `mousePressEvent` on pen touching.
(I'm not sure if this behavior is only in Qt 5.9 or not.)
And on pen releasing, both `tabletEvent` and `mouseReleaseEvent` is triggered.

So, the normal calling order of the events when "tapping" with the tablet is like as follows:
1. `tabletEvent()` with `QEvent::TabletPress` type
1. `mousePressEvent()`
1. `tabletEvent()` with `QEvent::TabletRelease` type
1. `mouseReleaseEvent()`

However, if the tapping causes time-consuming processing like frame-range fill, `mousePressEvent()` seems to be called AFTER `tabletEvent()` with `QEvent::TabletRelease` type for unknown reasons. Which causes unwanted another "click" just after the filling due to `mousePressEvent()`.

So I called `qApp->processEvents()` before the frame-range fill to make sure to "consume" the mouse press event in advance.

### DIFF
--- a/toonz/sources/tnztools/filltool.cpp
+++ b/toonz/sources/tnztools/filltool.cpp
@@ -237,14 +237,12 @@ public:
   void onAdd() override {}
 
   int getSize() const override {
-    int size1 = m_regionFillInformation
-                    ? m_regionFillInformation->capacity() *
-                          sizeof(m_regionFillInformation)
-                    : 0;
-    int size2 = m_strokeFillInformation
-                    ? m_strokeFillInformation->capacity() *
-                          sizeof(m_strokeFillInformation)
-                    : 0;
+    int size1 = m_regionFillInformation ? m_regionFillInformation->capacity() *
+                                              sizeof(m_regionFillInformation)
+                                        : 0;
+    int size2 = m_strokeFillInformation ? m_strokeFillInformation->capacity() *
+                                              sizeof(m_strokeFillInformation)
+                                        : 0;
     return sizeof(*this) + size1 + size2 + 500;
   }
 
@@ -640,11 +638,10 @@ public:
   }
 
   int getSize() const override {
-    int size =
-        m_selectingStroke
-            ? m_selectingStroke->getControlPointCount() * sizeof(TThickPoint) +
-                  100
-            : 0;
+    int size = m_selectingStroke ? m_selectingStroke->getControlPointCount() *
+                                           sizeof(TThickPoint) +
+                                       100
+                                 : 0;
     return sizeof(*this) +
            m_regionFillInformation->capacity() *
                sizeof(m_regionFillInformation) +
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@shun-iwasawa There was an apparent typo in your reference to the issue this PR will fix.
I've believe I've changed 3247 to 3427 in your post to point it to the correct reference.

beeheemooth:
It seems that here you are talking about the same thing as here #3263

RodneyBaker:
@beeheemooth Can you confirm this fixes your previous problem?

beeheemooth:
I confirm

RodneyBaker:
LGTM

beeheemooth:
@shun-iwasawa thank you!

DoctorRyan:
You the man @shun-iwasawa

Issue #3429, opened 2020-08-06
re-timing keys and drawing at the same


adamearle:
Workflow issue with spacing drawings and keys at the same time in the Timeline and function editor. 

At the moment any animation either done tradigital or cutout animation in retiming is very tedious and time-consuming. Through ongoing use over the years working with all kinds of artists the below list is are much-needed refinements of OpenToonz

Retiming animation should not be as painstaking as it is in OpenToonsz. We believe in the following request that workflow will speed up dramatically. This means that there will be a tighter faster cleaner way to navigate the timeline/ Xsheet & Function (Curve), editor. No more late nights on retiming animations because of OpenToonz outdated use of current modern code and computers. Animators' life quality will become significantly better from a few refinements.

We don't believe the Timeline/ Xsheet and Function (Curve) editor needs to be rewritten simply refined.

Function Editor: (Curve Editor)
> Retime everything across the timeline and Moving and the curve editor
> Select multiple keyframes at once by clicking dragging
> Remove select curve capability
> Remove all Key Spacing options and have just one this can be broken up in to 
    1. Drawing 2. Keys Icons that can be switched on separately or both at the same time
> Spacing out keyframes and drawings "Add space" & "Remove Space" across multiple columns
> Shift Keys backwards and forwards - Need to be changed to "shift keys Forward"/ "backward"

> An option to Reveal in the function editor from the

> Remove double click from the Timeline/ Xsheet to open the function editor and make it Ctrl + double click
> Copy-pasting keys + Hot Keys 
> Solo Keys + Hot Keys 
> Mute keys + Hot Keys 
> Isolating the selection + Hot Keys 
> Invert selection on columns + Hot Keys 

> Preference: Animation / (the option to have broken handles or not in the curve editor) 
> Make the keys at the top larger in the Curve Editor
> The repeat icon could also be seen in the curve editor
> Quick Switching icon so windows can be quickly switched between timelines/ Xsheet Function Editor

> Curve editor needs a normalized view of the curves
> Zoom in and out around on canvas during playback

> Curve editor appearance needs an update so that 

Timeline:
> Just changing and adding these 3 columns means that  retiming could all be done in the timeline
>        1.Add Mask column  
>        2. Add Peg column  
>        3. Add FX column

> Play every "x"frames
> Draw and animate only on odd/ evens

[...]

PR #3431, opened 2020-08-06
Stop Function Editor to Open by Double Clicking Keyframe Icon

shun-iwasawa:
This PR just stops to open Function Editor by double clicking the key icons in the Xsheet / Timeline.
This change is requested in the part of #3429 . This one makes sense and is easy to change.

I think it is enough to be able to open the Function Editor via context menu of the key icons. So I decided not to make the command with modifier keys (like Alt + double click) but just to remove the feature.

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -330,46 +330,36 @@ bool isGlobalKeyFrameWithSameTypeDiffFromLinear(TStageObject *stageObject,
   TDoubleKeyframe::Type type =
       stageObject->getParam(TStageObject::T_Angle)->getKeyframeAt(frame).m_type;
   if (type == TDoubleKeyframe::Linear) return false;
-  if (type !=
-          stageObject->getParam(TStageObject::T_X)
-              ->getKeyframeAt(frame)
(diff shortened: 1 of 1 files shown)


gab3d:
@shun-iwasawa thanks for helping improve the user experience! 

I should most probably make a separate issue for this one, but since is related to retiming keys, I'll try a mention here:
One thing that would be great is to be able to scale the selected keys (either in xsheet and/or function editor) by just dragging the selection from one of its ends (pivoting the scaling operation from the opposite end).
The feature (present in competing software like After Effects) is so useful when correcting timing in motion graphics projects, for example, that I miss it a lot in OT.

To use it (in After Effects): the user should select some keyframes in the timeline and then press Alt + drag from the first or last keys of the selection range.
The selected key range will be scaled, keeping the opposite end fixed, and the dragged end to where the user dragged it.
...if needed I could make a short video of how the feature works. 🤔 

Thanks for listening :)

DarrenTAnims:
Thank you so so much for this. This has been a long-standing annoyance for me (and others). This will make working with keys so much smoother.

RodneyBaker:
Thank you @shun-iwasawa 
LGTM

PR #3457, opened 2020-08-24
All New Icons (And GUI Cleanup)

konero:
The scope of this PR was to cleanup the GUI, but I ended up doing a bit more than that. I've redesigned almost every icon found in the core UI and added icons for **_every command_** found in the main menu bar, as well as many found in the command bar.

The icons are now theme-able (white or black) and are toggled by a user preference in the same place you choose the theme, for now the user must restart for the change to take effect.

**Icon Examples**
---
### Tools

tools_icons

### Command Bar (InknPaint Setup)

90994148-b829b500-e5af-11ea-8d5f-f5953eecde9a

### XSheet Toolbar (Common Actions)

90994448-a85ea080-e5b0-11ea-879a-6355c62cc78a

### Console
console_icons_a

### Mimetypes
image

### Folders
image

### Tasks
Annotation 2020-08-24 032201

### Schematic
schematicIconExample

### New Keyframe Design
I redesigned the look of keyframes, instead of using an image of a key I opted to use the diamond that's now found in most other software, I also updated them in the FX Editor window, so keyframes now share the same design language. The same colors are used in the Function Editor (including the folders).

keyframeExamplekeyframeExample2

A fully filled diamond is a set global keyframe, a half filled diamond is a partial keyframe, a diamond with a dot in the middle means that it's been modified.

_* The one exception is the keyframe icon on the XSheet/Timeline cells, I couldn't figure out how to draw the diamond there._

### XSheet & Timeline
I re-balanced the colors in the XSheet/Timeline to make them more pleasing to the eye. The spreadsheet lines have been given less contrast, the lines that are drawn between cells now pickup the hue from the track color, I made the track cells the focal point instead of the drag handles and the Timeline column header panel had a redesign.

image

### Tweaked the Palette List View UI
I changed the look of the Palette window list view, before it wasn't possible to tell if a batch selection was made and it was troublesome to check from time to time which color was selected. Now it resembles a typical list view, and I made the numpad border more prominent (and gave autopaint label an extra space).

paletteExample

## Note on New Functions
I added three new functions to assist in supporting theme-able icons.

1. `recolorPixmap()`: changes **_pure black pixels_** to any color, so if another color is present in the SVG it will be ignored and carry over. Note: if you must have black present in the image, use an off-color, something like `rgb(3, 3, 3)`.
2. `pixmapOpacity()`: changes opacity of pixmaps.
3. `getIconThemePath()`: builds the path to icon file by prefixing the theme directory with the option set in preferences, eg. `":icons/dark/"` or `":icons/light/"`. Note: This function should only be used when a theme-able pixmap is required and QIcon is not an option, you might also wish to call `recolorPixmap()`, `recolorPixmap(getIconThemePath("folder/icon.svg"))`.

It's not as elegant as I'd like, I'm not too experienced with Qt, this could definitely be updated by someone with more experience.

I changed `createQIcon()`, it will grab an icon using `fromTheme()`, conditionally check for alternate state image files just like it already does, and build the icon using the above functions. It will now add pixmaps for off and on disabled states. By default icons are set to 80% opacity for their off normal state, the boolean now lets you bypass that for 100%.

### How the Icons Work
I opted to use a custom `QIcon::fromTheme()`, you can refer to the [GNOME Icon Theme Specification](https://developer.gnome.org/icon-theme-spec/) table on how to use it, it's very easy. We have two themes, **Light** and **Dark** specified with their own `index.theme` file. Light **_inherits_** everything from Dark, so if Light is missing an icon file it will grab the icon from the Dark theme, for now I didn't need to duplicate icons across themes at all, the only exception is the startup popup logo, Light will grab icons from Dark and they will be recolored by OpenToonz appropriately, but this seemed like a very useful fallback for cases like that, if you absolutely must have two different icon files for each theme, you can drop it in both themes, just be sure to add both files to `toonz.qrc`. From testing, it's not required to add these icon resources to `toonzqt.qrc`.

### Note on Icon Design
They are "pixel perfect" symbolic, which means they are a pure color snapped to a pixel grid. To add alternate colors, use varying shades of alpha, so OpenToonz can recolor them properly, for example black 100%, black 60%, black 30%. If you're unsure, open one of the icons with a different shade such as the Geometry tool icon. When designing a new icon, make sure you enable the pixel grid in your software and enable snap.

Command icons are **_20x20_**, however the `index.theme` file contains specs that allow you to set fixed icon sizes to prevent the SVG from scaling beyond a certain point, so even though Command Bar is set to use `20x20`, if you provide it a `16x16` SVG and the specs in `index.theme` is set to `Size=16` (which means MaxSize) the SVG will remain `16x16` centered in a `20x20` button. You should therefore place the icon according to its width size in the specified folder in `sources/toonz/icons/dark` etc.

I chose `20x20` over `16x16` because it allows to add more detail for complex commands, this is the size found in Adobe and Clip Studio software. Unfortunately this means I had to disable the icons in the main menu as they would shrink to `16x16` and look blurry, I don't personally think they're of any use there though, it's rare to see it these days, but I added them back in at least for the tool icons, the bigger size makes it more comfortable to use on a 16-inch drawing tablet, for example, than if the icon had less detail.

❗ Tip: You can (on Windows) install a little extension called [SVG-See](https://github.com/tibold/svg-explorer-extension), which will allow Windows to display SVG thumbnails in File Explorer.

### Note on Stylesheets
Once again, I've refactored them quite a bit, mainly the toolbars and XSheet classes, it was unavoidable due to the icon size change so I apologize for the many changes to these, I split them out into a separate commit. One other thing to note, is I moved some of the external stylesheet icons into binary so they could take advantage of the new `createQIcon()` recolor abilities.

I did alter some design aspects of the stylesheets, but it's very minor stuff like making the checkboxes stand out in the Neutral (old Light theme), most of the work was for the XSheet/Timeline.

image

### New Light Theme
I added a new "Light" theme, which is actually a **_true_** light theme, the current theme has been renamed to Neutral to better represent its intention. The new Light theme inherits from Neutral, having little need to maintain it.

image

### Misc Changes
- Makes the `TPanelTitleBar` icon state colors changeable in the stylesheets.
- Removes margin around FX Editor window.
- Removes grey line above ConsolePlayBar.
- Removes black lines added to ScrollBar in New Project window (now they don't look broken).
- Removes white line under color sliders in Style Editor.
- Removes margin around ComboBox in Level Strip.
- Makes the text always viewable in the FPS field.
- Increases size of Link/Unlink Handles buttons in the Function Editor.
- Increases size of the motion path switches in Schematic nodes.
- Remove "Category" label from Preferences and "Layer" label from Timeline.
- Onion and Fixed Onion skin dots will draw a darker outline if the lightness of the OnionSkinArea is above 50%.
- Shrink size of ToolBar window by a tiny amount.
- Make Dv list view sorted headers (File Browser) changeable in the stylesheets.
- Make Schematic line colors, Palette selected and numpad colors, as well as other window properties changeable in the stylesheets, this was required to support the new Light theme properly.
</details>


jrhyme:
Well done. Finally, the program will acquire an excellent, modern and convenient look, and it will become more pleasant to work in it.

beeheemooth:
Oh my god ...  O_O Thanks for your titanic work!

FerrLui:
Really amazing, good job.
 Thank you.

пн, 24 авг. 2020 г., 17:17 Kite <notifications@github.com>:

DoctorRyan:
It was worth the wait.  Good job also on the PR explanation so other developers can know what is exactly going on with this.  The best part about this is now if anything needs updated UI wise down the road, it is gonna make it so much easier to implement changes.  Excellent job Konero.

manongjohn:
Nice job!

> * The one exception is the keyframe icon on the XSheet/Timeline cells, I couldn't figure out how to draw the diamond there.

Which icon are you refering to?

The normal "key" icons are `xshcellviewer.cpp`

>  static QPixmap selectedKey = svgToPixmap(":Resources/selected_key.svg");
>  static QPixmap key         = svgToPixmap(":Resources/key.svg");

The one the timeline uses is a drawn directly on the timeine in `drawFrameMarker()`

DoctorRyan:
This is a callout to any Mac/Linux devs out there.  Seems we have errors on the Travis CI builds.  I built my own appveyor build earlier so I know that is working correctly.  @manongjohn @artisteacher @turtletooth

manongjohn:
Doesn't necessarily require a Mac/Linux developer if you know what the error is and how to fix it...

Per Travis logs, Linux/MacOS builds are failing for the following errors:

/home/travis/build/opentoonz/opentoonz/toonz/sources/toonzqt/gutil.cpp:280:33: error: 
      no matching function for call to 'recolorPixmap'
  QPixmap normalPm            = recolorPixmap(normalIcon.pixmap(iconSize));
                                ^~~~~~~~~~~~~
/home/travis/build/opentoonz/opentoonz/toonz/sources/toonzqt/gutil.cpp:249:9: note: 
      candidate function not viable: expects an l-value for 1st argument
QPixmap recolorPixmap(QPixmap &pixmap, QColor color) {
        ^
/home/travis/build/opentoonz/opentoonz/toonz/sources/toonzqt/gutil.cpp:281:20: error: 
      no matching function for call to 'recolorPixmap'
  QPixmap overPm = recolorPixmap(QIcon::fromTheme(overStr).pixmap(iconSize));
                   ^~~~~~~~~~~~~
/home/travis/build/opentoonz/opentoonz/toonz/sources/toonzqt/gutil.cpp:249:9: note: 
      candidate function not viable: expects an l-value for 1st argument
QPixmap recolorPixmap(QPixmap &pixmap, QColor color) {
        ^
/home/travis/build/opentoonz/opentoonz/toonz/sources/toonzqt/gutil.cpp:282:20: error: 
      no matching function for call to 'recolorPixmap'
  QPixmap onPm   = recolorPixmap(QIcon::fromTheme(onStr).pixmap(iconSize));
                   ^~~~~~~~~~~~~
/home/travis/build/opentoonz/opentoonz/toonz/sources/toonzqt/gutil.cpp:249:9: note: 
      candidate function not viable: expects an l-value for 1st argument
QPixmap recolorPixmap(QPixmap &pixmap, QColor color) {
        ^
3 errors generated.

RodneyBaker:
Looking very good!

jpturcotte:
This is monumental. Well done, and thank you!

toonz/sources/toonz/penciltestpopup.cpp
```diff
@@ -1342,19 +1343,22 @@ PencilTestPopup::PencilTestPopup()
   m_captureButton->setIconSize(QSize(30, 30));
 
   m_captureFilterSettingsBtn->setObjectName("GearButton");
-  m_captureFilterSettingsBtn->setFixedSize(23, 23);
-  m_captureFilterSettingsBtn->setIconSize(QSize(15, 15));
+  m_captureFilterSettingsBtn->setFixedSize(24, 24);
+  m_captureFilterSettingsBtn->setIconSize(QSize(16, 16));
+  m_captureFilterSettingsBtn->setIcon(createQIcon("gear");
```
shun-iwasawa, line 1348: In this line the closing bracket seems to be missing.

shun-iwasawa:
Thank you @konero for brushing up the UI and icons. It looks so nice :-)

I tested with the "Dark" theme. Here are some concerns I've noticed so far:

- Xsheet columns becomes slightly brighter. The change seems to degenerate visibility of the white texts on the xsheet. Can you please make the column background colors to have contrast with the text color, as good as before?
  image

- The render task button icons are changed from the clapperboard to the teapot. Pardon me if I'm wrong, but the teapot symbol is recognized only in the context of 3DCG (having its roots in [Utah Teapot](https://en.wikipedia.org/wiki/Utah_teapot)). I'm a bit concerned that some users may not associate it with rendering.

I'll ask staffs in our studio to try this and gather further feedback.

toonz/sources/toonz/penciltestpopup.cpp
konero: Thanks for catching that!

konero:
@shun-iwasawa Thanks for the feedback! All very good points.

I made all the changes as requested, the level colors have been reverted to what they were originally, I can't remember why I brightened them so it can't have been important (I was probably spending too long in the Timeline).

Ah I never knew about the teapot! I always thought it meant "put the kettle on, this is going to take awhile". 😆 

I changed all "render/output" visualization to a clapperboard, to differentiate it from "new scene" I added some stars. If there's a better representation though, I'm happy to change it.

image

image

gab3d:
@konero kudos to you!
excellent work indeed!! 🚀

jrhyme:
@konero Column names are not visible in the light theme ...

konero:
@jrhyme If you're referring to the yellow highlight color when a column is selected, that was added to `Preferences > XSheet` awhile ago, the user will need to change the color there manually to something more suitable.

jrhyme:
Brother Shun, where are you) we missed)

konero:
Someone had a nice idea of using a tint/shade for cell selection, instead of just increasing the brightness, so the cells have a very small amount of blue tint from the main selected color like from some spreadsheet programs. I think this is much better than how I did it originally, feels a little more typical.

cellSelection

beeheemooth:
@konero I really like the columns with the offset line, it feels like they are voluminous, that's cool! But what will the keys look like and where will the key strip be? Will the user confuse the column strip with the key strip if the keys are out of sight?

konero:
@beeheemooth That's because I have **_Preferences > Interface > Display Level Name on Each Marker_** enabled. 😛

beeheemooth:
@konero Arghhh! I never paid attention to it. I was always annoyed by the names of the levels in each section. The side stripe looks very nice and it accentuates the keys. It is a pity that there is no such display mode, but without names.

konero:
Quick tutorial for adding emblems to symbolic icons, I included a template file found in **_icons/templates_** containing some common ones. They each have a main shape and a mask shape, which is 1px larger than the main shape, the mask is used to subtract from the main icon to cut out a hole.

image

manongjohn:
Appveyor had a logger failure and Travis had temporary issue installing QT.  I've restarted all failed builds. They should all complete successfully.

manongjohn:
The OpenToonz banner (startup.png) on the Startup dialog box doesn't switch based on theme.  It always shows the light "OpenToonz" text version.  On the non-light themes, it's fine, however the text is essentially invisible in the `Light` theme.  

Having said that, do you have the theme specific startup.png backwards?  The black text OpenToonz is in the dark/misc folder and the white text version is in the light/misc folder.  I would think you want the opposite.

konero:
@manongjohn Ah, the problem you describe is what I experienced a few days ago, I think I fixed it with https://github.com/opentoonz/opentoonz/pull/3457/commits/6d7f93e293ecaab1c1e2e97bc97110c922e901ee, if you have a build before that change you might still run into that problem. I tested the latest AppVeyor release and it seems to switch fine for me, could you double-check?

Basically, **dark** is for black icons, in this case we want black text to appear on a light interface, and **light** is for white icons, similar to how the directories are named for the QSS images. They are icon theme names independent from the stylesheet names.

manongjohn:
I was prepping OTX last night and wanted to include this so I pulled your branch in.  I also tested the Appveyor artifact from this PR which has a build date of 8/27, and it still shows the problem.

konero:
@manongjohn Thanks for checking! Although I can't reproduce it. 🤔

My build and the latest AppVeyor change it fine, I checked with qDebug and it grabs the correct image OK when switching the toggle. I checked on another machine which was also OK, a colleague tried it and it was OK and some users on Discord kindly tested it out and reported that it changes OK.

You're probably aware of this but just to be sure, the banner is part of the icon theme and not the main stylesheet theme, so it changes using the icon theme toggle which you need to restart OpenToonz to take effect. Could this be it?

manongjohn:
I just downloaded the Appveyor - Release artifact again to be sure and it still does it for me on Win10

Here's what i am doing:

1. Download the Appveyor - Release artifact from this PR
2. Go to my Downloads folder and unzip OpenToonz.zip into `OpenToonz` folder
3. cd to OpenToonz
4. Rename `OpenToonz stuff` to `portablestuff` and move it into `OpenToonz` folder
5. cd to OpenToonz
6. Double-click OpenToonz.exe
7. OT opens to splash page in `Default` theme and the startup dialog uses the white text version of the banner.
8. Open Preferences and change to `Light` theme  -> OT immediately changes to Light theme...nice touch btw!
9. Open Startup Popup again...The OpenToonz text is missing since it's the same color as the theme.  Expecting to see the black text version,  Ok, so I need to restart it...
10. Close and Restart OT -> Loads in `Light` theme as expected and the Startup dialog still shows the banner with the missing OpenToonz text.

I just realized my icons are also still white and not black when I in Light theme, even after restart.  It's also white when I switch to Neutral.  The white still works with the Neutral theme so I thought that was intentional, but it appears it should also be black based on your sample images?

konero:
@manongjohn Oh, are you making sure to check the option in preferences? This is the option where you need to restart.

image

You didn't list in your steps. 🤞 

Unfortunately I wasn't able to make the icons change color based on the stylesheet (it's definitely possible, I have limited knowledge of Qt/C++ to do it though, hopefully someone can improve on it).

manongjohn:
*facepalms* Sorry. My bad. Apparently I didn't read everything I should have so I didn't know I had to check a separate box for that. In my defense, it's not something I would have expected to do when switching themes. :)

Works fine otherwise.

shun-iwasawa:
Sorry for my long silence. I asked Ink&Paint staffs in our studio to try this and will sum up their feedback tomorrow. Thanks!

shun-iwasawa:
@konero 
Here is a list of feedback by Ink&Paint staffs in our studio.

**Somewhat Important**
- The cam-stand view button icon in the Xsheet column header with semi-transparent state is now having small checkerboard square. I think the combination of two motifs (table and checkerboard) may make it harder to recognize at a glance. Can you please make it as clear as before so that users can easily distinguish from the full-opac state?
gui1

**A Matter of Preference / A Matter of Minor Importance**
- The tile bar buttons in Viewer / Combo Viewer are now arranged a bit more tightly than before. It would be nice if they placed with the same interval as before so that users can click the right button for sure.
gui2
- I think it is really a matter of preference (and I think they'll get used to it soon), but they prefer the previous folder icon (without the horizontal line on top) because it is simple.
gui3

**A Matter of No Problem**
- Toolbar buttons are now arranged more tightly (i.e. with smaller padding) than before. I think the change is very helpful especially for users using monitors with small resolution and it can be adjusted by ourselves with style sheet.

They generally like this new & neat GUI changes. Again, thank you @konero !

beeheemooth:
Screen_0035

konero:
@shun-iwasawa Thanks for the feedback! I'm glad they generally liked it. I made all the requested changes as follows.

The folder icons have been reverted back to their original style.

image

I increased the width of the title bar buttons back to what they were originally, actually this was a very good point as they feel much easier to hit using a stylus now! Except preview/sub-preview is missing 1px than before so the shapes can be centrally aligned.

image

I changed the XSheet transparent table icon to the same style as before, it was a little tricky since the icon is now a flat shape but I hope this is a better representation.

image

For toolbar buttons I increased the size back to what it was originally, it wouldn't be possible with just a stylesheet change since I had to adjust the margins in `tpanels.cpp` to shrink the width of the window so I think it's reasonable if I go with the request of your studio to reduce the maintenance for you, since technically I didn't really receive any complaints about small screens, I think if small screens becomes a problem we can instead implement a new layout to the toolbar that can let the icons flow horizontally when the window is resized like in Clip Studio upto their separators. There's also a workaround that users can add tools to the command bar in the meantime.

One point, the buttons are still 1px shorter, this is because Qt for some reason adds an extra bit of padding to the right and bottom of QToolButtons which I reset, which makes icons look like they're out of alignment, it wasn't noticeable before with the original icons and I'm probably just being a bit of a perfectionist (😆) but the alternative is to make them bigger on the top and left to compensate, but this would increase the length of the toolbar more than before, if you could try out the new change hopefully this is a good size!

Original Size: **27x27**
PR Size: **24x24** (the size tested by studio)
Revert Size: **26x26** (reverting back to as closely as possible)

shun-iwasawa:
Jenkins

shun-iwasawa:
Thank you @konero for the quick response and modifications!
Thank you for the excellent work. I'll merge this.

shun-iwasawa:
@konero 
Sorry for being late to notice, but [you have hidden all icons in the menu bar](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/mainwindow.cpp#L1452). 
Did you think it would become too messy if all icons are visible in the menu bar? I think the icons would be helpful for some users to find and remember locations of commands. How about making them visible (or making it optional to be visible)?

image

konero:
@shun-iwasawa Sorry for late reply, I hid them originally since most are 20x20 size which can look out of focus since the QMenu supports only 16x16 size. But looking at it now I think I agree with you and actually software like Maya and Gimp deals with it regardless so we should be fine.

Could you be so kind as to submit the change? I'm not at my build machine to test.

The lines if I remember correctly are these:

Disables all icons:
https://github.com/opentoonz/opentoonz/pull/3457/commits/fbe5ffb1a05cc72097e5a3316cfd177b83b9bb05#diff-f8f6107d9c9fb90a80198f8c948cc64a4f3a65a797f4cdeb3d2c8f5ad5a129c4R1452

Re-enabled for tools:
https://github.com/opentoonz/opentoonz/pull/3457/commits/fbe5ffb1a05cc72097e5a3316cfd177b83b9bb05#diff-f8f6107d9c9fb90a80198f8c948cc64a4f3a65a797f4cdeb3d2c8f5ad5a129c4R1641

Removing those should revert the behavior.

PR #3480, opened 2020-09-16
Replace the palette's freeze button icon

shun-iwasawa:
This PR will replace the "freeze" button icon by the "pause VCR button" icon - the same icon as the ones in the Viewer and the Flipbook in order to gain consistency.

palette_freeze_button

Please note that now the state of the button has opposite meaning: The button had been with the "I/O power" icon which means that the panel synchronizes with the current palette when it is activated. Now the panel won't synchronize when the freeze button is activated.

@konero sorry for the change but I would be happy if you accepted my proposal.

### DIFF
--- a/toonz/sources/toonz/pane.cpp
+++ b/toonz/sources/toonz/pane.cpp
@@ -46,7 +46,6 @@ TPanel::TPanel(QWidget *parent, Qt::WindowFlags flags,
     , m_panelType("")
     , m_isMaximizable(true)
     , m_isMaximized(false)
-    , m_isActive(true)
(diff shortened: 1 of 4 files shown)


DarrenTAnims:
I like this change. It makes much more sense. The one/off button confused everyone I spoke with. Nice fix.

konero:
LGTM! This is much better, nice change.

RodneyBaker:
Travis failed for no apparently good reason but otherwise this PR looks good.

RodneyBaker:
Thank you @shun-iwasawa 
LGTM

PR #3481, opened 2020-09-16
Fix extra selection in palette

DoctorRyan:
When using the new style button, sometimes the old selected style would stay highlighted. This should fix that.
(Note: This bug fix was implemented in Tahoma.)

Co-Authored-By: Jeremy Bullock <turtletooth@users.noreply.github.com>


RodneyBaker:
Thanks for the fix Jeremy and the port/implementation @DoctorRyan 
LGTM

PR #3484, opened 2020-09-17
Render bug

DoctorRyan:
During render, certain cache files are deleted and sometimes the deletion fails. This keeps it from flipping out if the delete fails.

Co-Authored-By: Jeremy Bullock <4576381+jeremybullock@users.noreply.github.com>

### DIFF
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -390,8 +390,12 @@ void TSystem::renameFile(const TFilePath &dst, const TFilePath &src,
 
 // gestire gli errori con GetLastError?
 void TSystem::deleteFile(const TFilePath &fp) {
-  if (!QFile::remove(toQString(fp)))
-    throw TSystemException(fp, "can't delete file!");
+  if (TSystem::doesExistFileOrLevel(fp)) {
+    if (fp.getQString().contains("cache")) {
+      QFile::remove(toQString(fp));
+    } else if (!QFile::remove(toQString(fp)))
+      throw TSystemException(fp, "can't delete file!");
+  }
 }
 
 //------------------------------------------------------------


jeremybullock:
@shun-iwasawa In Tahoma I let it fail on deleting the cache file without throwing an error.  Is there a reason you can think of why that is a bad solution?

shun-iwasawa:
@DoctorRyan @jeremybullock 

> During render, certain cache files are deleted and sometimes the deletion fails. 

Please let me clarify what are the "cache files". Are they intermediate PNG files to be deleted [here](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp#L491) in the source after converting them into FFMpeg format (such as MP4)? Is there any other problematic case?

> In Tahoma I let it fail on deleting the cache file without throwing an error. Is there a reason you can think of why that is a bad solution?

Well, IMO the cache files are rather safe to be left undeleted. But it may gradually squeeze the storage if the problem occurs often and there is no user notification on failure.
Do you have any idea why the deletion failure occurs?

jeremybullock:
@shun-iwasawa It's not the temp png files created for the mp4.  It's the little cache files in the numbered folders in the appdata/local/opentoonz folder.  I was rendering out Dwanko with mp4, and after the first render completed, I'd render again and the second render would throw an error trying to delete one of those cache files.

shun-iwasawa:
@DoctorRyan @jeremybullock
I'm so sorry but I decided to temporarily close this PR for the following reasons:
- I've tried rendering the Dwanko scene but still couldn't reproduce the problem. Considering that there seems to be no similar issues reported so far, I think we need to gather more information of this problem before patching.
- Identifying the file from the cache folder by finding the term "cache" from the path is IMO at risk of misjudging:
    - Users may include the term "cache" for the files in other locations.
    - The [default value](https://doc.qt.io/qt-5/qstandardpaths.html#StandardLocation-enum) of `QStandardPaths::CacheLocation` is `~/Library/Caches/<APPNAME>` in macOS so at least you may need to add `Qt::CaseInsensitive` as the second argument of `QString::contains()` .
- I think it would be better to handle the thrown error than just to skip it. 

Thank you for tackling with the issues and kind understanding.

Issue #3168, opened 2020-03-19
keyframe data not writing out correctly to scene file


manongjohn:
### Issue Summary
Was helping a user who suddenly could not open their scene.  When opening the scene, it would give the warning that not all levels could be loaded and eventually would crashing.

With some debugging, I found it crashed when trying to load/display keyframe data. After some review of the scene file, I found inavlid keyframe data being written to the file

Example bad data:
```
          <keyframes>
            <Ep>
              4 0 3.33333 
            </Ep>
            <Ep>
              <prev>
                0 -3.33333 
              </prev>
              14 0 0 
            </Ep>
            <n>                          <--- This is not a valid interpolation
              <prev>
                0 -3.33333 
              </prev>
              </n>
```

It is unclear what may have caused this to be generated.

manongjohn:
Closing this for now as I'm not sure this is the real issue. Not sure why, but the corrected file stopped working so something else is going on...

manongjohn:
Found the problem this time...I think.  Bad interpolation data was written to the file.  I've updated the post to reflect my new findings

manongjohn:
FYI, found another issue with the keyframe data is generated in a user's file that was suddenly causing crashes. The keyframe data was written to the TNZ file with a missing tagname:

 ```
            <>
              <prev>
                1.24367 
              </prev>
              </>
```

I'm seeing a lot more reports of issues with people working successfully with a scene then closing it and reopening it and it crashing.  I'm starting to suspect whatever is causing this bad data is the issue.

KAPthenoob:
does that make some keyframes not appear in the xsheet. if true, then this is the same problem with me. i use OT morevna 1.4 on windows 7 32bit.

RodneyBaker:
I'm curious if PR #968 might have some impact on what is happening with this issue.

KAPthenoob:
it doesn't work. fyi, I have a different scene using the same level as the broken scene, but there was no problem loading the new scene. I have an assumption that an error occurred when saving keyframe data to the TNZ file, which caused the file and the levels in it to be loaded incorrectly. I think this problem is felt by many new users like me. I have tried a number of open source animate applications, and I found that Opentoonz is better than others for me. I hope this problem can be fixed so that more people will use Opentoonz.
GBU

manongjohn:
If you want to privately send me the .tnz file, I can take a quick look at it.

Easiest to do this on the Discord if you want to communicate directly with me there.

[...]

PR #3485, opened 2020-09-17
Fix incorrectly using None interpolation type

DoctorRyan:
Note: Original fix ported from Tahoma

This PR fixes an problem in Tahoma that also exists in the Official OpenToonz version noted in issue opentoonz/opentoonz#3168

When a new keyframe is created, the keyframe may inadvertently be created as a None interpolation type. There may be multiple ways to do this, but 1 known way is to create a new keyframe before the 1st keyframe in the column and then copy/paste the newly created keyframe after it. This was due to not updating m_prevType in the next keyframe properly.

A None interpolation keyframe is written to the scene file as <n>...<\n> which can lead to a crash while loading the scene or create strange behavior with the keyframes in the column.

This fix should

1. Prevent the creation of a None interpolation type keyframe

2. Throw out None interpolation keyframes when loading the scene in case it encounters one.

### DIFF
--- a/toonz/sources/common/tparam/tdoubleparam.cpp
+++ b/toonz/sources/common/tparam/tdoubleparam.cpp
@@ -757,6 +757,9 @@ void TDoubleParam::setKeyframe(int index, const TDoubleKeyframe &k) {
     dst.m_prevType = TDoubleKeyframe::None;
   else
     dst.m_prevType = keyframes[index - 1].m_type;
+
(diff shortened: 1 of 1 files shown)


RodneyBaker:
There is a secondary fix in the works for other cases similar to this one.
This is a short term fix to remedy the majority of cases.

Thanks to @manongjohn and @DoctorRyan for the fix.

PR #3487, opened 2020-09-18
Additional Style Sheet

shun-iwasawa:
This PR will enable users to customize style sheet from the Preferences.

additional_style_sheet

The style sheet provides an easy way to customize GUI without touching the source code. 
For instance, as shown in the image above, in our studio we enlarge the buttons in the Style Editor at Ink&Paint staffs' request.
However, currently you need to re-introduce the modifications every time you install the newer version of OT in which the style sheet has been updated.

In this PR the modifications will be separately stored in the user preferences so that it will save the trouble of re-introduction.
Furthermore, it allows personalized customization for each user sharing the same Stuff folder in the studio.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -953,6 +953,7 @@ QComboBox QAbstractItemView {
 .LineEdit,
 QPlainTextEdit,
 QLineEdit,
+QTextEdit,
 #TaskSheetItem,
 #tasksRemoveBox,
 #tasksAddBox {
@@ -967,6 +968,7 @@ QLineEdit,
 .LineEdit:focus,
 QPlainTextEdit:focus,
 QLineEdit:focus,
+QTextEdit:focus,
 #TaskSheetItem:focus,
 #tasksRemoveBox:focus,
 #tasksAddBox:focus {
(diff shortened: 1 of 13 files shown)


gab3d:
@shun-iwasawa I suppose you enlarge the buttons in the studio, but what you're really after is to have a a more square shaped style thumbnail, isn't it?

I've seen this rectangular thumbnails as a bad thing for a long time now, but I can't imagine how could they be made square again (as they were in Toonz) without sacrificing so much space in the panel. 🤷🏻‍♂️

If anyone happens to have a good solution for this, it'd be good for making those thumnails more useful again.

konero:
Nice! Works extremely well.

image

RodneyBaker:
Looks like Travis build exceeded the time limit.

RodneyBaker:
Thank you @shun-iwasawa for the feature.
LGTM

beeheemooth:
Sounds very cool! But where can an ordinary user find a list of items that can be changed using this feature?

RodneyBaker:
> where can an ordinary user find a list of items that can be changed using this feature?

Ordinary users are prohibited from use of the PASS (Preferential Additional Style Sheet).
Only super users are allowed access to this level of power.

Kidding!  :)

It will take some time but at least now CSS 'recipes' can start to be published.
Currently the best place to find the examples would be in the QSS/Less files.

RodneyBaker:
Here is a good place to get started:
(See attached)

[Default.qss.txt](https://github.com/opentoonz/opentoonz/files/5272099/Default.qss.txt)

RodneyBaker:
As a post mortem for this feature going into v1.6...
It would be good to include a few sample style sheets for use with the PASS system.

An additional option for the feature that would be useful:
- Be able to save out and load the custom style sheets
-

PR #3498, opened 2020-09-23
Fix Fractal Noise Fx Iwa

### LINKED ISSUE #3497: Fractal Noise Fx Iwa fails to generate consistent noise pattern.
**Describe the bug**
Fractal Noise Fx Iwa fails to generate consistent noise pattern under some conditions, especially when the noise is generated in some bounding box and the fx column is rotated.

**To Reproduce**
1. Load some level to Col1. Create Fractal Noise Fx Iwa in Col2.
1. Create the scene like as follows:
   XSheet and Function Editor
   fnbug1
   Fx Schematic
   fnbug2
   Note that Col2 is rotated by only 1 degree at the frame 2.
1. Preview and compare the results of the two output nodes.

The result of the output from the XSheet node becomes unexpectedly inconsistent, as shown in the GIF below.

**Expected behavior**

Noise patterns must be the same, only except that one result is full screen sized and the other is in the shape of the matte level.

**Screenshots**

The output from Fractal Noise Fx Iwa (correct)
fractalnoise_bug1

The output from XSheet  (wrong)
fractalnoise_bug2

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #3497 , removing unnecessary transformation.

Now the result seems to be OK.
fnbug_ok

### DIFF
--- a/toonz/sources/stdfx/iwa_fractalnoisefx.cpp
+++ b/toonz/sources/stdfx/iwa_fractalnoisefx.cpp
@@ -155,9 +155,6 @@ void Iwa_FractalNoiseFx::doCompute(TTile &tile, double frame,
   TAffine subAff = TTranslation(param.subOffset) * TScale(param.subScaling) *
                    TRotation(-param.subRotation);
 
-  TAffine offsetAff =
-      TTranslation((double)outDim.lx * 0.5, (double)outDim.ly * 0.5);
-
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Travis timed out again (I think)

RodneyBaker:
Travis checks complete.

PR #3499, opened 2020-09-23
fix language definitions

Savraska2:
Weblate can't correctly define some languages.
Снимок экрана от 2020-09-23 22-17-32
Instead of names in English, there should be Russians (in my case, it depending on the your chosen Weblate interface language).
https://weblate-ru.readthedocs.io/ru/weblate-2.7/admin/projects.html?#component-configuration


RodneyBaker:
@Savraska2 I haven't investigated yet...
Should the other PR for the weblate be merged before this one or does it particularly matter?

Savraska2:
Yes, you are right, it does't matter.

RodneyBaker:
Thank you @Savraska2

PR #3500, opened 2020-09-25
Fix standards compliance for some files.

ericonr:
Hi! These are some small changes that simplify the preprocessor magic, and, in my case, make the project compile for musl libc.

Since I don't know the language of some of the comments (I assume it's Japanese, given the links), I didn't know whether I should update their contents.

### DIFF
--- a/toonz/sources/stdfx/igs_resource_msg_from_err_unix.cpp
+++ b/toonz/sources/stdfx/igs_resource_msg_from_err_unix.cpp
@@ -1,3 +1,5 @@
+#define _POSIX_C_SOURCE 200809L /* int strerror_r() */
+#undef _GNU_SOURCE
 #include <cerrno>
 #include <cstring> /* memset */
 #include <vector>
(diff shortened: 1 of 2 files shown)


RodneyBaker:
LGTM
Thank you @ericonr 
We'll leave updating of the comments to in Japanese to @shun-iwasawa (which looks unlikely).

toonz/sources/stdfx/igs_resource_msg_from_err_unix.cpp
```diff
@@ -1,3 +1,5 @@
+#define _POSIX_C_SOURCE 200809L /* int strerror_r() */
```
rozhuk-im, line 1: IMHO setting this is not good idea.
At least on FreeBSD this brokes compilation, because compiler did not get here:
```
#else				/* Default environment: show everything. */
#define	__POSIX_VISIBLE		200809
#define	__XSI_VISIBLE		700
#define	__BSD_VISIBLE		1
#define	__ISO_C_VISIBLE		2011
#define	__EXT1_VISIBLE		1
#endif
```
and ```__XSI_VISIBLE``` required for ```isascii```

https://bugs.llvm.org/show_bug.cgi?id=34330
https://sigrok.org/bugzilla/show_bug.cgi?id=649
https://github.com/Olga-Yakovleva/RHVoice/issues/107
https://groups.google.com/forum/#!topic/native-client-discuss/m2vOMflHJrQ

Not sure that this is FreeBSD specific, may be this llvm specific.

I disable these two lines with ifdef for FreeBSD.

PR #3501, opened 2020-09-26
Fix copy/paste keyframe causing bad data

DoctorRyan:
Note:  This fix from Tahoma was created by Manongjohn.  Info following in his own words about the change.

This PR fix should stop the generation of <n> or <> keyframe data when saving.

The logic that was there assumed and worked correctly for full keyframes, but not partial keyframes. When copy-pasting partial keyframes, it would set m_prevType on channels that were not keyframed. In addition to that, the newly pasted key's interpolation was being incorrectly updated which would lead to the <n>/<> when pasting before a partial keyframe.

Added logic to only update m_prevTypes if the channel was keyframed on the newly pasted key and the next/previous key also had that channel keyframed.

### DIFF
--- a/toonz/sources/toonz/keyframedata.cpp
+++ b/toonz/sources/toonz/keyframedata.cpp
@@ -137,9 +137,12 @@ bool TKeyframeData::getKeyframes(std::set<Position> &positions,
         // interpolation between them using preference setting
         TStageObject::Keyframe prevKey = pegbar->getKeyframe(kL);
         for (int i = 0; i < TStageObject::T_ChannelCount; i++) {
-          prevKey.m_channels[i].m_type =
-              TDoubleKeyframe::Type(Preferences::instance()->getKeyframeType());
-          newKey.m_channels[i].m_prevType = prevKey.m_channels[i].m_type;
+          if (newKey.m_channels[i].m_isKeyframe &&
+              prevKey.m_channels[i].m_isKeyframe) {
+            prevKey.m_channels[i].m_type = TDoubleKeyframe::Type(
+                Preferences::instance()->getKeyframeType());
+            newKey.m_channels[i].m_prevType = prevKey.m_channels[i].m_type;
+          }
         }
         pegbar->setKeyframeWithoutUndo(kL, prevKey);
         pegbar->setKeyframeWithoutUndo(row, newKey);
@@ -149,7 +152,9 @@ bool TKeyframeData::getKeyframes(std::set<Position> &positions,
         if (!pegbar->getKeyframeSpan(row - 1, kP, e0, kN, e1)) kP = row - 1;
         TStageObject::Keyframe prevKey = pegbar->getKeyframe(kP);
         for (int i = 0; i < TStageObject::T_ChannelCount; i++) {
-          newKey.m_channels[i].m_prevType = prevKey.m_channels[i].m_type;
+          if (newKey.m_channels[i].m_isKeyframe &&
+              prevKey.m_channels[i].m_isKeyframe)
+            newKey.m_channels[i].m_prevType = prevKey.m_channels[i].m_type; 
         }
         pegbar->setKeyframeWithoutUndo(row, newKey);
       }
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@shun-iwasawa Appveyor build failed here.
This should correct an oft reported corruption of scene files.

RodneyBaker:
All checks passed.  
LGTM

Thanks @manongjohn @jeremybullock and @DoctorRyan for the fix.

hippiecool:
I've been experiencing similar issues with Load as Sub-Xsheet with scenes that contain many numerous keyframes that indeed were copy/pasted from previous frame rows.   I'm using Nov 2020 AppImage on Ubuntu 20.04 and OT 1.4 latest on Win10.

Steps I do:
1. Create a new scene with lots of keyframes that were indeed copy/pasted throughout the scene. The way I do this is (Hold CTRL key, then use mouse to click & drag to select the desired cells and their keyframes - see pic below.  Then select the target frame cell below starting in its matching column, and hit (Ctrl+V).  It usually doesn't work right away.  Sometimes I must click a different frame cell then re-click the target frame cell and (Ctrl+V) again to get the frames and their keyframes to actually paste.  But it does on 2nd try.  (I do it this way, because the Repeat function doesn't bring its keyframes with it).
2. Save as scene1.tnz
3. Create new scene, called master.tnz then in Column1, Cell1, do Load as Sub-Xsheet scene1.tnz - it instantly crashes.

In general, it usually will work as expected with scenes who's keyframes were created manually, not copy/pasted.  I only copy/paste keyframes and cells when creating a scene were the character is repeating a loop of similar actions (running, skating, etc) so I just have to adjust the (X,Y) position of the character.
 
This is how I COPY Source Keyframes and Cells (HOLD Ctrl key, grab desired cells).  Also notice Col7 name is not unique to just one Column.  I find this in many of my scenes.  Not sure if this is related, I have never Copy/Paste Columns, I always create new Levels in each Column to keep the structure very clean. 

copy paste keyframes


After reading many threads on this issue, I do not see any <n> .... <> when I inspect scene1.tnz.txt in Notepad in Win10.  

Win10 throws an Application Error 1000 pointing to AcLayers.DLL - see pic below.  Does this help you? 


win10 error


Not sure where to go from here? 
Is there something else I should be looking for to get the scene to Load as Sub-Xsheet? 
I'm happy to share my scene1.tnz.txt file for inspection. 
Is there an alternate way to copy/paste keyframes and cells correctly which will not corrupt the file? 

Thanks

manongjohn:
I would suggest attaching your scene file so it can be looked at in more detail to find where it is crashing.

hippiecool:
See attached.  Thank you. 
[scene1.tnz.txt](https://github.com/opentoonz/opentoonz/files/5546943/scene1.tnz.txt)

manongjohn:
I am able to open the scene using Nov 2020 OTX, so it appears unrelated to this particular keyframe issue.

I'm wondering if something with one of your levels is wrong, especially since you are using only vectors.

I would suggest loading each vector level into a new scene and look at all frames to see if one of them might be causing a crash.

hippiecool:
To be clear, I can Open the scene without issue. It crashes when trying to get it to Load as Sub-Xsheet into a Master Scene with other Sub-XSheets. I have experienced this specifically with 3 similar scenes so I will do as you suggest and try loading each vector again in a new scene to look for anomalies in the Levels and report back.  But I don't think that is the case since I have Loaded this character into many master scenes as a Sub-Xsheet.  I am just having troubling with certain scenes across the board (which  only difference is numerous keyframes) which is why I was starting in this issue.

manongjohn:
I was able to successfully import it as a sub-xsheet as well.

I would suggest opening a new issue, if it doesn't appear to be a vector layer issue as this was related specifically to the `</n>` and `</>` bug which is not present in your scene.

You many need to provide the entire project or a dummy project that exhibits the same issue.

hippiecool:
I will create a new issue to fully explain my methodology for using vector layers. 

Thanks

PR #3504, opened 2020-09-27
Fix big endian code

ericonr:
Hi! We use these patches on Void Linux, which has packages for big endian PCC and PPC64 machines. The big endian parts were previously correct, but refactoring around the code base didn't reach the big endian part, so it needed these small fixes.


RodneyBaker:
> We use these patches on Void Linux, which has **packages for big endian PCC and PPC64 machines**.

Hi @ericonr 
Very interesting.
Is there a place where we can find out more about your efforts there?

As for this PR...
LGTM

Thanks!

ericonr:
@RodneyBaker 

Certainly! This is the main link: https://voidlinux-ppc.org/

Glad to help :)

PR #3508, opened 2020-09-28
GUI modifications

shun-iwasawa:
This PR slightly modifies GUI as follows:

1. Function Editor - spread sheet view
    - Changed the column head colors in "Dark" theme so that the ruled lines become more visible.
    - Added 1px vertical line at the end of the column heads.
    - Slightly adjusted positions & added borders so that the panes perfectly align with each other and with Xsheet.

    gui_mod1

1. PSD settings popup (opens when loading .psd files)
    - Removed decorations from read-only QTextEdit in order to clarify it is just a non-editable, multi-lined label.

    gui_mod2


RodneyBaker:
Very appropriate modifications.
Thank you @shun-iwasawa

PR #3510, opened 2020-09-28
Fix Inputting Japanese Characters

### LINKED ISSUE #3509: Fails to Input Japanese Characters on Renaming Styles
**Describe the bug**
OT sometimes fails to input Japanese characters properly when renaming styles.
This problem has been annoyance for Ink&Paint staffs in our studio.

**To Reproduce**
1. Double click a style to start renaming.
1. Before typing, randomly move the mouse cursor.
1. Try typing some Japanese text.
1. With any (un)luck, it fails to input properly as if the consonant and the vowel are typed separately.

**Expected behavior**
Japanese character should be input properly regardless of the mouse behavior.

**Screenshots**
type_bug

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #3509 .

When moving the cursor while renaming styles, `TPanel::leaveEvent()` and `TPanel::enterEvent()` are randomly called even if the cursor does not leave the panel for unknown reasons. This behavior temporary loses the keyboard focus and disrupts input before combining the consonant and the vowel into Japanese character.
This PR prevents losing focus on `TPanel::leaveEvent()`  if the current focused widget is any inheritance of `QLineEdit`.

### DIFF
--- a/toonz/sources/toonz/pane.cpp
+++ b/toonz/sources/toonz/pane.cpp
@@ -119,19 +119,13 @@ void TPanel::enterEvent(QEvent *event) {
   QWidget *w = qApp->activeWindow();
   if (w) {
     // grab the focus, unless a line-edit is focused currently
-    bool shouldSetFocus = true;
-
     QWidget *focusWidget = qApp->focusWidget();
-    if (focusWidget) {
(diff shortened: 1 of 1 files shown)


RodneyBaker:

Issue #3258, opened 2020-04-27
Invalid value for Color Calibration path preference causes crash on startup


martinvanzijl:
**Describe the bug**
I entered an invalid value (added some spaces, I think) to the "Color Calibration 3DLUT file path" in the Preferences window, then closed OpenToonz. The next time I tried to start OpenToonz, it crashed with a segfault message during the splash screen.

I found this quite by accident while testing something else related to preferences.

**To Reproduce**

**WARNING: Please do not try this unless you know how to edit the preferences file manually!**

1. Choose _File --> Preferences_.
2. Go to the _Interface_ page.
3. Check the box for "Color Calibration using 3D Look-up Table".
4. Type an invalid value into the textbox (e.g. "Hello").
5. Close the window.
6. Close OpenToonz.
7. Try to restart OpenToonz. This crashes during the splash screen due to a segfault error.

The only way to fix this is to manually edit the settings file and remove the entry for "colorCalibrationLutPaths". Then restart OpenToonz and it runs correctly.

**Expected behavior**
Invalid value for the path should not cause a crash on startup.

**Screenshots**
1. Setting the preference:

![preference-bug-01](https://user-images.githubusercontent.com/24422213/80421892-4a3aed00-8931-11ea-8d9d-342e9d034128.png)

2. Crash upon closing and restarting:

![preference-bug-02](https://user-images.githubusercontent.com/24422213/80421918-56bf4580-8931-11ea-8808-b682c5435c98.png)

3. Have to delete line from preferences file:

![preference-bug-03](https://user-images.githubusercontent.com/24422213/80421932-5e7eea00-8931-11ea-90b6-17d989c205af.png)

Full video below (zipped OGV):

[preference-bug-demo.zip](https://github.com/opentoonz/opentoonz/files/4542165/preference-bug-demo.zip)

**Device Information:**
 - OS: Ubuntu 18.04 (64-bit)
 - OpenToonz Version: Git 90814c256edff7e11ced473770672520619c1014 (24 April 2020)
 - GPU: Intel® Sandybridge Mobile
 - GPU Driver Version: 3.0 Mesa 19.2.8 (?)

martinvanzijl:
**Useful debugging info**

I debugged it in GDB, and the problem occurs here:

https://github.com/opentoonz/opentoonz/blob/90814c256edff7e11ced473770672520619c1014/toonz/sources/toonzlib/preferences.cpp#L661

[...]

PR #3513, opened 2020-09-30
Apply Color Calibration 3DLUT On-the-fly

shun-iwasawa:
This PR enables to apply and switch 3D Look-up Table files without rebooting OT.
The following preference options will now effect as soon as they are changed:
- Interface category > `Color Calibration using 3D Look-up Table`
- Interface category > `3DLUT File for [Monitor Name]:`

### DIFF
--- a/toonz/sources/include/toonzqt/lutcalibrator.h
+++ b/toonz/sources/include/toonzqt/lutcalibrator.h
@@ -9,6 +9,7 @@
 #include <QOpenGLBuffer>
 #include <QMatrix4x4>
 #include <QOpenGLFunctions>
+#include <QSet>
 
 #undef DVAPI
 #undef DVVAR
@@ -53,7 +54,8 @@ class DVAPI LutCalibrator : public QOpenGLFunctions {
 
 public:
   // static LutCalibrator* instance();
-  LutCalibrator() {}
+  LutCalibrator();
+  ~LutCalibrator();
 
   // to be computed once through the software
   void initialize();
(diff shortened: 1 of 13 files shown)


RodneyBaker:
Very nice @shun-iwasawa 
Having to restart Opentoonz made the previous 3DLut feature a bit difficult to use.
Having the ability to change on-the-fly...   big thumbs up!

3dluts 0001

RodneyBaker:
@shun-iwasawa 
I'm running into a problem where (because I couldn't remember how to create a valid LUT file for use with Opentoonz) I point Opentoonz to the malformed 3Dlut and from there on I have major problems to include that I cannot even use 3dlut files that wree working perfectly before (as those that produced the images in my previous post).

I thought perhaps this related to ffmpeg but when I remove ffmpeg... same results.

I am struggling with remember the specific requirements of the header for the lut file but there seems to be more going on here.

This may relate to https://github.com/opentoonz/opentoonz/issues/3258

RodneyBaker:
Apparently the malformed path problem is corrupting the preferences.ini entry.

`colorCalibrationLutPaths=@Variant(\0\0\0\b\0\0\0\x1\0\0\0\x16\0\x41\0n\0y\0 \0M\0o\0n\0i\0t\0o\0r\0\0\0\n\0\0\0Z\0\x43\0:\0\\\0U\0s\0\x65\0r\0s\0\\\0r\0o\0\x64\0n\0\x65\0\\\0O\0n\0\x65\0\x44\0r\0i\0v\0\x65\0\\\0\x44\0\x65\0s\0k\0t\0o\0p\0\\\0\x43\0i\0t\0y\0_\0\x44\0u\0s\0t\0.\0\x33\0\x64\0l)`

shun-iwasawa:
Thank you @RodneyBaker for the feedback.
I haven't tested enough with using invalid or malformed 3dlut files. I'll test it.

Regarding the preferences.ini entry, it seems OK. The LUT file path is not a single string but the paired list of the monitor name and the LUT path. Qt saves such data in human-unreadable format.

RodneyBaker:
With additional testing it doesn't appear this PR adds any negative results that weren't already there with regard to malformed 3dlut files.  It may just be brining them to light more easily because of the ability to update on-the-fly.

This PR is very welcome.

P.S. It is very good to know about the paired data as path in the ini file.
Very interesting.

RodneyBaker:
Will proceed with merge here.

Aside:  It will be nice to add a few example 3dluts in the Library for users to test with but I'm not certain how best to proceed with presenting optimal LUTs.

PR #3518, opened 2020-10-09
Fix Pasting Freehand-selected Toonz Raster Image

### LINKED ISSUE #3517: Pasting toonz raster selection offset by 1 pixel
**Describe the bug**
Toonz raster image selected with the freehand selection tool sometimes unexpectedly offset by 1 pixel from the original position on pasting.
Note that the selected area should across the bottom or left image boundary.
The problem occurs not always but quite often, like one third of operations.

**To Reproduce**
1. Load any Toonz Raster Level with more than one frames.
1. Select the Selection tool with Freehand type.
1. Select the portion of the image at the frame 1 across the lower image boundary.
1. Copy the selection.
1. Select the second cell in the Xsheet.
1. Paste.
1. Check if the image is pasted at exactly the same position as the copied one. Sometimes it offset by 1 pixel downward.

**Expected behavior**
The image should be pasted at the same position as the copied one.

**Screenshots**
paste_toonz_raster_bug

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #3517 .

When computing intersection between the freehand stroke and the image boundary, some points may slightly run over the boundary due to calcuration error.
It causes unwanted expansion of the pasted bounding box when converting it to integer by truncating.

I added lines to make the intersection points to be exactly match the boundary in order to avoid it.

### DIFF
--- a/toonz/sources/tnztools/rasterselection.cpp
+++ b/toonz/sources/tnztools/rasterselection.cpp
@@ -796,6 +796,19 @@ TStroke getIntersectedStroke(TStroke &stroke, TRectD bbox) {
       TThickPoint p  = getIntersectionPoint(bbox, stroke.getChunk(chunkIndex),
                                            currentSegmentIndex,
                                            chunkIndex == precChunkIndex);
+
+      // exactly match the position with the edge of bbox
+      // or the pasted raster may offset by 1pixel due to truncation in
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I confirm the problem and that this PR appears to resolve the issue.

Issue #3514, opened 2020-10-02
Selective on eraser tool with segment type


openanim:
**Is your feature request related to a problem? Please describe**

On vector level the segment eraser tool is not affected by the selective option.

**Describe the solution you'd like**

It will be great if as others erasers type it was available, for cleanup workflow and consistency.

openanim:
I close it. it seems that it does not close automatically anymore ?

PR #3523, opened 2020-10-11
implement selective option for segment eraser

pojienie:
this implements #3514 

To be honest I never knew what the "selective" option was for before reading the code......


openanim:
Great ! Thanks for this @pojienie.

RodneyBaker:
Very nice.

PR #3526, opened 2020-10-13
XDTS I/O enhancement: Enabled to handle frame number with suffix

shun-iwasawa:
This PR will enable to export / import XDTS (eXchange Digital Time Sheet) format when including frame numbers with suffixes (like `1a`, `2B`, etc. ).

### DIFF
--- a/toonz/sources/toonz/xdtsio.cpp
+++ b/toonz/sources/toonz/xdtsio.cpp
@@ -58,6 +58,24 @@ void XdtsHeader::write(QJsonObject &json) const {
 
 //-----------------------------------------------------------------------------
 
+TFrameId XdtsFrameDataItem::str2Fid(const QString &str) const {
+  bool ok;
+  int frame = str.toInt(&ok);
+  if (ok) return TFrameId(frame);
+  // separate the last word as suffix
+  frame = str.left(str.size() - 1).toInt(&ok);
+  if (!ok) return TFrameId(-1);                              // EMPTY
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Thanks to @wolfinabowl for posted this to assist with testing:

[OTTest.zip](https://github.com/opentoonz/opentoonz/files/5380331/OTTest.zip)

This zip file consists of the XDTS data file and associated images.
In testing he noted:

> I can confirm that OT is able to load the suffixed frames with this PR. 
> The previous behavior simply omitted the frames, leaving a gap.
> OT does not correctly map the drawings to these suffixed frames however as it appears that it automatically renumbers them.

image

shun-iwasawa:
Thank you @WolfInABowl and @RodneyBaker for the feedback! I'll check it.

shun-iwasawa:
I've checked with the sample files in uploaded zip. 

It seemed that the image files are named without suffixes;
image
After I made the file names with correctly suffixed like as follows;
image
.. I could load it correctly from XDTS:


I thought that this feature unwantedly renumbers suffixed frames to non-suffixed. Please let me know if I missed something.

WolfInABowl:
That appears to have been my mistake, I used the `File name` setting `Serial number` when exporting the cels.
Changing that setting to `Cel name` fixes the issue and everything imports into OT correctly.

RodneyBaker:
Thank you for the confirmation @WolfInABowl 
This PR looks good to move forward.
Thanks for testing!

PR #3528, opened 2020-10-13
Fix Text Fx bug

shun-iwasawa:
This PR fixes a problem regarding `Font` field of the Text Fx settings as follows:

**To reproduce:**

1. Create Two Text Fxs. The both fxs are with `Font family = "Segoe UI" size = 30` by default. 
1. Open the first Text Fx's settings.
1. Change the font family to anything else (say,  `Arial` ) and change the size to `10` .
1. Select the second Text Fx and see if the Fx settings updates properly.
1. In our environment, the font size field fails to be updated. i.e. the font size of the second Text Fx becomes `10` .

### DIFF
--- a/toonz/sources/toonzqt/paramfield.cpp
+++ b/toonz/sources/toonzqt/paramfield.cpp
@@ -1579,7 +1579,7 @@ void StringParamField::onChange() {
   if (m_multiTextFld)
     value = m_multiTextFld->toPlainText().toStdWString();
   else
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
Hmm Jen.kins for Windows seems to hang up. I'll check later.

RodneyBaker:
I confirm the issue and that this PR fixes the problem.

PR #3529, opened 2020-10-13
Fix Crash on Undoing Replace-paste of Zerary Fx

### LINKED ISSUE #3084: FX Schematic copypaste + undo crash with zerary nodes
gif example:
Undo Crash

Repro steps:

1. Create a zerary node in the FX Schematic
2. Copy the node and paste it on itself, you will notice the connecting wire is now absent
3. Undo the paste action

OT will crash

Additional:
The connecting wire is no longer present after pasting, unlike with non-zerary nodes.

You can trigger this behaviour by pasting any zerary node on any other zerary node, they do not need to be the same.
No existing connection needs to be present for the crash to occur when undoing.

Edit:
Just found another way to crash that appears related.
Ctrl+V calls the Paste Replace command, however using the Paste Add will result in an immediate crash.

Repro steps:

1. Create a zerary node in the FX Schematic
2. Copy the node and call the Paste Add command

OT will crash

Paste Add Crash

I assume this is because OT is attempting to connect the node to a non-existant input port.

shun-iwasawa:
This PR fixes a crash problem as follows:

**To reproduce:**

1. On Fx Schematic, create some zerary fx.
1. Select and copy the Fx ( `Ctrl+C` ).
1. While selecting the Fx node, paste the Fx ( `Ctrl+V` ). The same Fx will be pasted replacing the existing one.
1. Undo the operation. OT gets crash.

Also fixes #3084

### DIFF
--- a/toonz/sources/toonzlib/fxcommand.cpp
+++ b/toonz/sources/toonzlib/fxcommand.cpp
@@ -2793,9 +2793,9 @@ void UndoAddPasteFxs::initialize(TFx *inFx) {
   m_linkIn = TFxCommand::Link(inFx, ifx, 0);
 
   // Furthermore, clone the group stack from inFx into each inserted fx
-  auto const clone_fun = static_cast<void (*)(TFx *, TFx *)>(FxCommandUndo::cloneGroupStack);
-  for_each_fx(
-      [inFx, clone_fun](TFx *toFx) { clone_fun(inFx, toFx); });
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I confirm the crash and also that this PR fixes the problem.
Thanks also @WolfInABowl for the confirmation of this addressing #3084.
(please let us know if anything remains problematic after merging)

PR #3530, opened 2020-10-14
Palette Filter Fx Enhancement

shun-iwasawa:
This PR will enhance the Palette Filter Fx with `Lines & Areas (No Gap)` option.

A major purpose of the `Lines & Areas (No Gap)` option is to retrieve the portion of image assuming that it will be composited back with the inverse-filtered part of the image after applying some fxs.
For instance, consider the character which is drawn together with its shadow. This option can be used to retrieve the shadow part and composite again after making it semi-transparent by Transparency Fx.

However, currently the option still leave transparent part after deleting lines and causes unwanted gap after re-composition.
This PR enhances the option so that the transparent gap is filled with neighbor pixel's style ( i.e. making the image with No Gap, in the true sense of the word ).

Here is a simple example scene showcasing the enhancement:
In the Fx Schematic the two Palette Filter Fx, one is for deleting lines (with style#1) and the other is for keeping only lines, are composited again. 

Here are the results (click to enlarge):

Please note that the rendering result by existing one has small semi-transparent gaps (where the green background can be can be seen through) especially around the eye of the character. This PR resolves it.

### DIFF
--- a/toonz/sources/common/trop/tropcm.cpp
+++ b/toonz/sources/common/trop/tropcm.cpp
@@ -182,7 +182,7 @@ paints[i]  = TPixelFloat(paints2[i]);
       }
       paints[0] = TPixel32::Transparent;
     } else
-      for (int i  = 0; i < palette->getStyleCount(); i++)
+      for (int i = 0; i < palette->getStyleCount(); i++)
         paints[i] = inks[i] =
             ::premultiply(palette->getStyle(i)->getAverageColor());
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Very nice!
I'm still testing out various combination but everything looks good thus far.

papio-pl:
Thank you! It's very useful. I use this feature a lot:)

PR #3534, opened 2020-10-16
Ease Regulations for the Use of Animate Tool

shun-iwasawa:
This PR will ease regulations for the use of Animation Tool as follows:

**The current conditions**
The tool is disabled if the current object is column and such column is hidden in the viewer.
Neither transformation (like positions, rotation, etc.) nor fx gadgets (like Inner or Outer Size of the Radial Gradient Fx) are controllable.

**The new conditions proposed in this PR**
- Even if the current column is hidden, the transformation is still enabled in case at least one of the columns connected to such column as child objects in the stage schematic is visible in the viewer. This is because the result of transformation can be seen in the Viewer as the change of child columns.
  

- Even if the current column is hidden, fx gadgets are still enabled since the result of the fxs are not visible in the camstand view in the first place.

Additionally, I made the tool to be disabled when the current object is the empty column.

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -446,8 +446,8 @@ return true if the method execution can have changed the current tool
   int getFrame();        //!< Returns the actual frame in use.
   int getColumnIndex();  //!< Returns the actual column index.
 
-  TStageObjectId
-  getObjectId();  //!< Returns a pointer to the actual stage object.
+  TStageObjectId getObjectId()
+      const;  //!< Returns a pointer to the actual stage object.
 
   void notifyImageChanged();  //!< Notifies changes on the actual image; used to
                               //! update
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Everything appears to work as advertised but I'm not sure I'm fully understanding this last part:

"Additionally, I made the tool to be disabled when the current object is the empty column."

It may be that I'm not fully understanding what is meant by 'empty column' here.

Edit:  No, I get is now.  I was thinking that I needed to have an empty column be a parent or child which is not possible.
I see now that what is meant is that the Animate Tool is simply disabled when the column is empty and so therefore will now allow keyframes to be created in that empty column.

All LGTM
Will delay merging for a few more tests and to allow time for others to test.
Posting intent to merge.

RodneyBaker:
What is a bit odd to me is that without this PR change it seems we can move an empty column and effect another column.
Very strange.  This PR actually resolves that.

RodneyBaker:
Merging but want to keep an eye open for a few new workflows/hidden features introduced with this PR.

PR #3537, opened 2020-10-17
Selectable Parameter Name Label

shun-iwasawa:
This PR enables to select parameter name label in the segment viewer panel of the Function Editor in order to copy & paste it into the expression field.
This idea was suggested in [this comment](https://github.com/opentoonz/opentoonz/pull/3375#issuecomment-708597360) by @RodneyBaker . Thanks!


RodneyBaker:
A pleasant surprise!
Thank you @shun-iwasawa for implementing this (as well as the clarification before about drag/dropping expression objects).
LGTM :)

PR #3540, opened 2020-10-19
Fix Particles Fx with Sub-xsheet Control

### LINKED ISSUE #3539: Particles Fx fails to render properly when sub-xsheet column is connected as the control image
**Describe the bug**
This problem have been a problem for us over the years as we could not find the connections with sub-xsheet.
Particles Fx fails to render properly when sub-xsheet column is connected as control with a condition as follows:
- The scene is rendered in multi-thread or rendered from the frame more than 1. (The problem won't occur when rendering in single thread and from the frame 1.)

**To Reproduce**

1. Create a level with two frames for the fx's control image. Draw a red circle in the frame 1 and a green circle in the frame 2.
    particles_bug_1
1. Create a level for the fx's texture image. In this example I drew a small triangle.
    particles_bug_2
1. Collapse the column 1.
1. Create the Particles Fx. Arrange the Fx schematic and the Xsheet like as follows:
   particles_bug_3
1. Open the fx settings of the Particles Fx. Set the `Control Image` of `Source` and `Birth Color` to `1`. Set the `Birth Color -> Birth Intensity` to `100`. Leave other parameters be default.
1. Set the `Preview Settings -> Dedicated CPUs` to `Single`.
1. (Correct result)  Select the frame 1 in the xsheet and Preview the scene. The render result of the frame 2 is like as follows:
   particles_bug_ok
   Note that the 10 red triangles which were born in the frame 1 appears on left, and the other 10 green triangles which were born in the frame 2 are on right.
1. (Wrong result)  Select the frame 2 in the xsheet and Preview the scene. The render result of the frame 2 becomes like as follows:
    particles_bug_ng
    Note that all triangles becomes green and appears on right. 

**Expected behavior**
Rendering results must be consistent regardless of the frame where the computation started from.

**Device Information:**
 - OS: Win10
 - OpenToonz Version: latest nightly

shun-iwasawa:
This PR will fix #3539 .
The Particles Fx is quite special in a sense that it refers the rendering result of the control images in previous frames, when tracing the whole "history" of particles from the start frame to the current rendering frame. 
Sub-xsheet had not assumed such case and it only held one integer value ( `m_frame` ) for each rendering frame which is the frame number in the sub-xsheet to be redirected to.
This means that the sub-xsheet column cannot output the result of previous frames.

I made the sub-xsheet fx (`TimeShuffleFx`) to hold a pointer to the column and to obtain proper frame number to redirect.

### DIFF
--- a/toonz/sources/toonzlib/scenefx.cpp
+++ b/toonz/sources/toonzlib/scenefx.cpp
@@ -68,9 +68,11 @@ private:
   int m_frame;                 //!< Frame this fx redirects to
   TFxTimeRegion m_timeRegion;  //!< Input (outer) valid column frame range
   TRasterFxPort m_port;        //!< Input port
+  TXshCellColumn *m_cellColumn;
 
 public:
-  TimeShuffleFx() : TRasterFx(), m_frame(0), m_timeRegion() {
+  TimeShuffleFx()
+      : TRasterFx(), m_frame(0), m_timeRegion(), m_cellColumn(nullptr) {
     addInputPort("source", m_port);
   }
   ~TimeShuffleFx() {}
@@ -81,6 +83,7 @@ public:
 
     fx->setFrame(m_frame);
     fx->setTimeRegion(getTimeRegion());
+    fx->setCellColumn(m_cellColumn);
 
     return fx;
   }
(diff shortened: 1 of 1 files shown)


RodneyBaker:
> I made the sub-xsheet fx (TimeShuffleFx) to hold a pointer to the column and to obtain proper frame number to redirect.

I'm not familiar with that fx.  Is that something that has been released to the public yet?

shun-iwasawa:
@RodneyBaker 
> I'm not familiar with that fx. Is that something that has been released to the public yet?
On rendering, the sub-xsheet node is converted to a special-purpose "fx" ( named `TimeShuffleFx` ) which will redirect the frame according to the cell number in the sub-xsheet column. It is just for internal use in the software and nothing more than an implementation of the sub-xsheet. I apologize for any confusion!

RodneyBaker:
Thanks for the clarification @shun-iwasawa 
Very good to know.

RodneyBaker:
LGTM

PR #3541, opened 2020-10-19
Fix Crash When Drawing on Absent Level

### LINKED ISSUE #3531: crash on a scene with level no saved
**Describe the bug**
on the vector level, if I save a scene forgetting to save a level, when I load the scene again and use a tool (the brush in example) OT will crash.

1.create a level
2.save the scene
3.load the scene
4.use the brush / geometry / type tool on the level
5.OT will crash

shun-iwasawa:
This PR fixes #3531 .
This PR also takes care of the similar problem occurs with Raster and Toonz Raster levels.

The crash occurred due to the initialization (like setting the palette, resolution and dpi) is not completed which is normally done when loading the resources on opening the scene.
This PR will initialize the level on creating the new frame if the level is empty to avoid crash. The initialization will be done with the values set in the preferences.

### DIFF
--- a/toonz/sources/include/toonz/txshsimplelevel.h
+++ b/toonz/sources/include/toonz/txshsimplelevel.h
@@ -243,6 +243,10 @@ Editable range is contained in \b m_editableRange.
   //!  and assigns the specified level path.
   TImageP createEmptyFrame();
 
+  void initializePalette();
+  void initializeResolutionAndDpi(const TDimension &dim = TDimension(),
+                                  double dpi            = 0);
+
   TDimension getResolution();
 
   TPointD getImageDpi(const TFrameId &fid = TFrameId::NO_FRAME,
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Thank you @shun-iwasawa for the fix and for extending it to other level types as well.
LGTM

RodneyBaker:
@shun-iwasawa I'm watching to see if the nightly build gets properly listed.
While new nightlies gets listed on the main Opentoonz release the last few were not added to the Nightlies release listiing.

shun-iwasawa:
According to the log, it seems that some software is needed to update. I'll do that by tomorrow.
Thank you @RodneyBaker for letting me know!

shun-iwasawa:
I updated Ruby in the Jen.kins machine and now it seems to be back to normal. 👍

PR #3545, opened 2020-10-20
Export Sub Xsheet Column to XDTS

shun-iwasawa:
This PR expands the XDTS exporting feature to export sub xsheet column as well as normal level columns (= columns containing Toonz Vector, Toonz Raster or Raster level). This feature was demanded by some user in Japanese animation industry.

This PR modifies only exporting, but I think it would be nice if we could somehow arrange sub xsheet on importing XDTS as well in the future.


RodneyBaker:
Thank you @shun-iwasawa for getting nightly releases back online and... for this XDTS enhancement!

PR #3546, opened 2020-10-20
Fix Repeat Fx Commands to Work in Non-English Environment

shun-iwasawa:
In Fx Schematic, by pressing `Ctrl + Right Click` you can repeat the last Fx-creating operation (like `Insert Blur`, `Add Color Card`, etc.) for quick building of Fx Schematic.
However, it hadn't worked in non-English environment. This PR fixes such problem.

(I'm thinking of re-examining the best way of implementation of this feature... This fixes the existing bug anyway.)


RodneyBaker:
LGTM
Thank you @shun-iwasawa

PR #3551, opened 2020-10-21
Fix Crash on Reverting Scene

shun-iwasawa:
This PR fixes crash when using `Revert Scene` command while displaying the Level Strip.

To reproduce:
- Load some scene containing a level.
- Open Level Strip if you don't have one.
- `Revert Scene` . OT crashes.


RodneyBaker:
(Lots of autoreformatting in this one)

I confirm the crash with the steps you outline.
LGTM

PR #3559, opened 2020-10-22
Enhance Studio Palette Searching

shun-iwasawa:
This PR enhances studio palette searching procedure when using `Toggle Link to Studio Palette` and `Get Color from Studio Palette` commands.

These commands had tried to find the linked studio palette by checking all files in the `Global Palettes` and `Project Palettes` tree. It takes time especially if there are a lot of files under the `Project Palettes` folder tree. (In our studio, we put not only the studio palettes but also color model files for all scenes under the same folder tree as they are related to the color designing workflow.)

This PR will make an index file listing the "global palette IDs" and the "paths to relevant palettes" (`palette_paths.ini`) under the both root folders of `Global Palettes` and `Project Palettes` so that OT can find the palette efficiently.
An index of some studio palette will be automatically updated when searching it for the first time after its creation / relocation.

### DIFF
--- a/toonz/sources/include/toonz/studiopalette.h
+++ b/toonz/sources/include/toonz/studiopalette.h
@@ -107,6 +107,9 @@ public:
 
   void save(const TFilePath &path, TPalette *palette);
 
+  void removeEntry(const std::wstring paletteId);
+  void addEntry(const std::wstring paletteId, const TFilePath &path);
+
 private:
   StudioPalette();
   TFilePath m_root;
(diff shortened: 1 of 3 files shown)


RodneyBaker:
@shun-iwasawa I am not sure I know enough about the process here to thoroughly test this PR.
As such it may be appropriate for you to merge after you have validated on your end.

I can verify that in testing the appveyor for this PR that I experienced no adverse effects.
Your description seems straightforward enough but it seems likely that I may miss something in the intended workflow using Studio palettes.  (Note:  This suggests I need to learn more about the use of Studio palette which I look forward to!)

RodneyBaker:
I believe I've been able to test enough to say LGTM here... and learned much in the process.
Thank you @shun-iwasawa for the enhancement.

PR #3570, opened 2020-10-27
Remove unused directory and some unused files

SeanKuehl:
Removing lzodriver from sources. Removing a .cpp and a .h file from sources/toonzlib. These files have not been touched in years and should be removed as they don't seem to play any part in the project in it's current state.


shun-iwasawa:
Jenkins

shun-iwasawa:
LGTM

PR #3582, opened 2020-11-09
Fix Raster Rectangular Fill Tool

shun-iwasawa:
This PR fixes the following problem regarding Fill Tool with tablet:

**To reproduce**

- Open some toonz raster level.
- Select the Fill tool
- Select the `Rectangular` type and the `Areas` mode
- Using tablet, double-click(tap) any point on the image.

In our environment the rectangular fill unexpectedly executed with the region from the center of the image and the double-clicked point. The problem was due to the function `FillTool::leftButtonDoubleClick()` is called between `FillTool::leftButtonDown()` and `FillTool::leftButtonUp()` when the second tap, and in the function the clicked pos is reset to the center of the image. 
I added a line to cancel the left button click when the double click is detected so that the filling operation will be skipped.

In this PR I also modified that the rectangular fill operation will not register undo if the specified rectagle is too small or contains no line pixels (i.e. the case that the operation obviously does not change the image) .

### DIFF
--- a/toonz/sources/include/toonz/fill.h
+++ b/toonz/sources/include/toonz/fill.h
@@ -109,7 +109,7 @@ Fill \b rect in raster with \b color.
 else if \b fillInks is false fill only paint delimited by ink;
 else fill ink and paint in rect.
 */
-  void rectFill(const TRect &rect, int color, bool onlyUnfilled,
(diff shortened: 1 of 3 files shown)


RodneyBaker:
I confirm the issue and that this PR addresses the problem.

PR #3583, opened 2020-11-10
Fix Crash on Switching Rooms

shun-iwasawa:
This PR fixes crash on switching rooms, with conditions as follows:

**To reproduce:**
- Display the Style Editor in floating window.
- Enable 3DLUT color calibration.
- Select an empty cell or a cell containing 16bpc raster level so that the hexagonal color wheel is hidden in the Style Editor.
- Switch the rooms. OT crashes.

The crash occurs when calling `initializeOpenGLFunctions()` in the function `LutCalibrator::initialize()` .
When the color wheel is hidden, this function crashes due to absence of OpenGL context.

I added a line to postpone the update of calibrator until the color wheel is shown.

### DIFF
--- a/toonz/sources/include/toonzqt/styleeditor.h
+++ b/toonz/sources/include/toonzqt/styleeditor.h
@@ -152,7 +152,8 @@ class DVAPI HexagonalColorWheel final : public GLWidgetForHighDpi {
   QOpenGLFramebufferObject *m_fbo = NULL;
   LutCalibrator *m_lutCalibrator  = NULL;
 
-  bool m_firstInitialized = true;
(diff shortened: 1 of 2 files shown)


RodneyBaker:
I confirm the crash upon switching rooms with the Style Editor open and 3DLut calibration enabled.
Also confirm that this PR resolves the crash.

Additionally, this seems to address a crash with similar case even without switching rooms.
If the Style Editor is open I can get a crash just enabling 3Dluts.
Using this PR I no longer get that crash either.
In order to get past that to test your case I had to delay opening the Style Editor uhtil AFTER 3DLut calibration was enabled and working.

In other words... LGTM!

PR #3584, opened 2020-11-11
New Fx : Bloom Iwa Fx

shun-iwasawa:
This PR will introduce a new fx, Bloom Iwa Fx, for generating a [bloom effect](https://en.wikipedia.org/wiki/Bloom_(shader_effect)). 

This Fx is based on the [LightBloom plugin](https://github.com/opentoonz/dwango_opentoonz_plugins/blob/master/LightBloom/src/main.cpp) developed by Dwango, with modifications as follows:

- Parameters are reduced & simplified.
- Reduced the block noise appeared especially when the gamma is with higher value.
- Enabled to render alpha gradation.
- Enabled to adjust size with the viewer gadget.

### Parameters

- **Gamma**  specifies how the light decays along with distance.
- **Gain**  specifies intensity of light.
- **Size**  specifies size of bloom. This value is adjustable via a viewer gadget.
- **Alpha Rendering** when activated, the value (or brightness, i.e. the maximum value of RGB channels) is set to the alpha channel as "pseudo" opacity. 

### Caution
Since this Fx uses OpenCV, currently this Fx is only available in Windows version built with `WITH_STOPMOTION` option. I'm thinking of including OpenCV regardless of options or platforms by the release of next version.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1341,6 +1341,12 @@
   <item>"STD_iwa_FractalNoiseFx.dynamicIntensity"	"Dynamic Intensity"	</item>
   <item>"STD_iwa_FractalNoiseFx.alphaRendering"	"Alpha Rendering"	</item>
   
+  <item>"STD_iwa_BloomFx" 			"Bloom Iwa" 	</item>
+  <item>"STD_iwa_BloomFx.gamma"	"Gamma"		</item>
+  <item>"STD_iwa_BloomFx.gain"	"Gain"		</item>
+  <item>"STD_iwa_BloomFx.size"	"Size"		</item>
+  <item>"STD_iwa_BloomFx.alpha_rendering"	"Alpha Rendering"		</item>
+
  <!------------------------------ Tiled Particles Iwa ------------------------------------------->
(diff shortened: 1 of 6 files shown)


RodneyBaker:
@shun-iwasawa  Since I'm getting an error that looks like the Bloom Iwa FX didn't build correctly I will guess the appveyor builds don't build with stopmotion on.

Just for general reference here's an image of what I see in the FX Schematic:
image

The code is a very interesting read through and I had never heard of the term 'power space' but that makes very good sense as a term.  Also... I very much like the idea of including OpenCV as well as all the various plugins in the next release if that can be done.

I have been planning to build locally for some time now but likely won't get that going until this weekend.
As such someone else will need to build and test with stopmotion on to give feedback.
Otherwise everything looks good here thus far.

jeremybullock:
Testing this now.

jeremybullock:
This works well.  Nice one @shun-iwasawa

RodneyBaker:
Merging this based on @jeremybullock's feedback.
Thank you @shun-iwasawa for the new FX!

PR #3596, opened 2020-11-20
Convert To Vector : Align Boundary Strokes Direction Option

shun-iwasawa:
This PR will add a new option to `Convert To Vector` popup; Align Boundary Strokes Direction.
When the option is activated, boundary strokes will be aligned to be clockwise direction, i.e. left to right direction as viewed from inside of the shape. This option will be effective especially when converting a shape with a hole like donuts.

Here is a example:

**Source Image**

**Convert to Vector result**

(The result looks the same between with and without the option)

**Style changed, showing the stroke direction**

Left: Without the `Align Boundary Strokes Direction` option
Right: With the `Align Boundary Strokes Direction` option

Note that the inner boundary stroke is flipped so that the stroke direction becomes the same as viewed from inside of the shape.

This option is made since I'm planning to add a custom vector style in the future which has different appearance between inside and outside of the shape.
This option can be a separate command so that it is applied to any existing vector images. Though, it will be a future development.

### DIFF
--- a/toonz/sources/include/toonz/tcenterlinevectorizer.h
+++ b/toonz/sources/include/toonz/tcenterlinevectorizer.h
@@ -95,8 +95,8 @@ Returns image converted.*/
                        const VectorizerConfiguration &c);
   void applyFillColors(TRegion *r, const TRasterP &ras, TPalette *palette,
                        const CenterlineConfiguration &c, int regionCount);
-  void applyFillColors(TRegion *r, const TRasterP &ras, TPalette *palette,
-                       const OutlineConfiguration &c, int regionCount);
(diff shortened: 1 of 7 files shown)


RodneyBaker:
@shun-iwasawa LGTM via appveyor but the build stalled/timed out on Travis.

shun-iwasawa:
@RodneyBaker 
It seems that the old failed checks (`Build / Windows (pull_request)` and `Travis CI`) are not cleared even if the PR is closed & reopened. Please don't care of them. Since the latest checks are all passed, it's ready to be merged. 👍

RodneyBaker:
Thank you @shun-iwasawa for the new option!

Issue #3352, opened 2020-06-13
Opentoonz generates invalid XML for scene / project files


mbasaglia:
For scene and project files, Opentoonz does not seem to be properly escaping characters, here's an example:

```xml
<property name="Compression Type" type="enum" value="Lempel-Ziv & Welch encoding">
```

As you can see in the attribute "value" it has a raw ampersand "&" instead of encoding it as "&amp;amp;" which is what XML requires.

This makes it difficult to work with opentoonz files, as various tools might fail parsing them.
I'm working on importing opentoonz scenes into Synfig, another open source animation program, but this issue is going to make the process more difficult than it needs to.

Agnyy:
@shun-iwasawa @RodneyBaker  @manongjohn  @ turtletooth  

I apologize for distracting you. Help me figure out the system for saving projects in OT.
I'd like to understand how OT stores data. I would like to be able to tear off this data in other programs.

MrBlockCat:
I wonder if this could be somehow related to #3239 ? They seem suspiciously similar to me... I could be wrong though!

mbasaglia:
> I wonder if this could be somehow related to #3239 ? They seem suspiciously similar to me... I could be wrong though!

No I think that's because "&" is often used to mark the shortcut letter in menu items eg "&File" will display File with an underlined F

ave4:
I confirm the bug, it prevents me from loading the tnz scene into renderchan to analyze the dependencies of the tnz scenes from other scenes and programs.

ave4:
I wrote a patch that renames `"Lempel-Ziv & Welch encoding" to "Lempel-Ziv and Welch encoding"`
Old scenes work as before
```
diff --git a/toonz/sources/image/tif/tiio_tif.cpp b/toonz/sources/image/tif/tiio_tif.cpp
index f8cec502b..1b1d56bc4 100644
--- a/toonz/sources/image/tif/tiio_tif.cpp
+++ b/toonz/sources/image/tif/tiio_tif.cpp
@@ -902,7 +902,7 @@ void TifWriter::open(FILE *file, const TImageInfo &info) {
     std::wstring compressionType =
         ((TEnumProperty *)(m_properties->getProperty("Compression Type")))
             ->getValue();
-    if (compressionType == TNZ_INFO_COMPRESS_LZW)
+    if (compressionType == TNZ_INFO_COMPRESS_LZW || compressionType == TNZ_INFO_COMPRESS_LZW_LEG)
[...]

PR #3599, opened 2020-11-23
Fix incorrect xml in tnz file

ave4:
This patch fixes issue https://github.com/opentoonz/opentoonz/issues/3352 by changing **Lempel-Ziv & Welch encoding** to **Lempel-Ziv and Welch encoding** in new projects

For backward compatibility, old projects and scenes will work as before and will not be overwritten.


RodneyBaker:
LGTM Thanks @ave4 
Will wait to see what @shun-iwasawa can do about Travis before proceeding.

RodneyBaker:
Thank you @ave4 for the fix as well as maintaining backward compatibility.

PR #3601, opened 2020-11-24
[WIP] Try Updating Travis macOS Version

shun-iwasawa:
This PR is an attempt to fix recent errors in Travis macOS builds.
According to the log, it seems to be frozen when brew-installing some dependent libraries which has not been included in the previous builds.
I still not sure the cause of the problem, but I just would like to try with the latest version.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ matrix:
         - os: linux
           compiler: clang
         - os: osx
-          osx_image: xcode10.1
+          osx_image: xcode12.2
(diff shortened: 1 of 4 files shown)


shun-iwasawa:
Unfortunately, despite the trials done in recent days I couldn't managed to make Travis to finish OSX build within the time limit of 50 minutes. Eventually it reached 92% of the procedure but I don't come up with another measure anymore.

It seems to be better to use another CI tool like GitHub Actions :-)

PR #3614, opened 2020-12-02
Glare Fx Iwa Revised

shun-iwasawa:
This PR will revise the Glare Fx Iwa, both in modifying the internal calculation and adding new features as follows.

### Modification

I fixed the chromatic calculation: when summing the intensity of each wavelength of light, weight is modified from constant to proportional to 1/(_wavelength_^2) which is more physically-accurate.
Now the modified fx generates more "white" glare pattern:

In order to keep backward compatibility (as we already had used this Fx for production), I added a new variable `fxVersion` to the `FxAttribute` .
The Glare fx made with the older versions is with `fxVersion=1` (default) and will compute in the previous way. 
The fx created after this version will be saved as `fxVersion=2` and compute in the modified way.

### New Features

Introduced the "Iris Preset" feature, which enables to generate typical glare patterns without an input image from the `Iris` port. 
Here are the types of glare shapes:

And here are the parameters:
- **Iris Scale** specifies the size of the generating iris image at a ratio of `Filter Size`
- **Symmetry** specifies how the streaks with the same length / at regular intervals.
   
- **Appearance** specifies line width of the iris shape which will change the interval of the rainbow coloring. Selecting the option `Filled` will minimize the rainbow coloring in the pattern. 
   
- **Edges** and **Random Seed** are enabled only when `Iris Shape = Multiple Streaks` is selected. `Edges` specifies the number of edges of gear-shaped iris (therefore, specifies number of streaks accordingly). `Random Seed` specifies the random seed to generate asymmetric shape when the value `Symmetry` is less than 1.0.

Also added a value `Chromatic Aberration` which specifies the order and dispersion of the rainbow colors.

Left: `Chromatic Aberration = 1.0` (default value)
Right: `Chromatic Aberration = -1.0`

Furthermore, a new option `Iris` is added to the `Render Mode`, for checking the generated iris pattern.

#### ATTENTION

Since the new parameter `fxVersion` will be saved in the scene, the scene file containing Glare Fx Iwa cannot be opened with the older versions of OT.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1310,9 +1310,16 @@
 
   <item>"STD_iwa_GlareFx" 			"Glare Iwa" 	</item>
   <item>"STD_iwa_GlareFx.renderMode"	"Render Mode"		</item>
+  <item>"STD_iwa_GlareFx.irisMode"	"Iris Shape"		</item>
+  <item>"STD_iwa_GlareFx.irisScale"	"Iris Scale"		</item>
+  <item>"STD_iwa_GlareFx.irisGearEdgeCount"	"Edges"		</item>
+  <item>"STD_iwa_GlareFx.irisRandomSeed"	"Random Seed"		</item>
+  <item>"STD_iwa_GlareFx.irisSymmetry"	"Symmetry"		</item>
+  <item>"STD_iwa_GlareFx.irisAppearance"	"Appearance"		</item>
   <item>"STD_iwa_GlareFx.intensity"	"Intensity"		</item>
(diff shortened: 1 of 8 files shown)


RodneyBaker:
LGTM

Still not sure how to proceed given Travis check failing.

shun-iwasawa:
I'll try closing and reopening this PR to trigger the GitHub Action.

RodneyBaker:
I believe that if restarted all checks should pass for this PR.

RodneyBaker:
Thank you @shun-iwasawa for the revision!

PR #3627, opened 2020-12-10
Fix crash on "no current level" selected in level strip

### LINKED ISSUE #3471: OpenToonz turn off itself when I use "no current level" to drawing.
**Describe the bug**
crash

**To Reproduce**
It crashes this way: >>>>
1. New sence
2. Brush sth on the paper
3. select "No current level"
4. brush sth
5. Opentoonz turned off automatically

**Expected behavior**

**Screenshots**

**Device Information:**
 - OS: Windows 10 (Up to date)
 - OpenToonz Version: OpenToonz 1.4
 2rd Feb 2020
 - GPU: NVIDIA Geforce MX250 / Intel(R) UHD Graphics 620
 - GPU Driver Version: 451.67 / 26.20.100.7926

shun-iwasawa:
This fixes #3471 
I decided to take the solution different from tahoma2d/tahoma2d#149 .

IMO the combo box item `No Current Level` should not allow users to make the current level empty. It causes unexpected problem like this issue and AFAIK there seems to be no point in doing so.
Therefore, I made the item `No Current Level` to be displayed only for notification and remove the item from the list when users click the combo box while some level is current.

### DIFF
--- a/toonz/sources/toonz/filmstrip.cpp
+++ b/toonz/sources/toonz/filmstrip.cpp
@@ -1643,7 +1643,8 @@ void Filmstrip::updateChooseLevelComboItems() {
     }
   }
 
-  m_chooseLevelCombo->addItem(tr("- No Current Level -"));
+  if (m_chooseLevelCombo->count() == 0)
+    m_chooseLevelCombo->addItem(tr("- No Current Level -"));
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #3629, opened 2020-12-11
Fix Github Actions to work: update versions

shun-iwasawa:
This PR modifies Github Actions settings to work on Windows and MacOS environments.

I confirmed it to work on my remote repository.
Windows build seems to take so long time (about 3hrs) mainly for installing and building the dependency library like Qt.


shun-iwasawa:
Update: it seems to use the cache and the Windows build becomes pretty fast.

I'll merge it as it succeed in all builds. 👍

PR #3631, opened 2020-12-11
Actions: Separate workflow yml for re-running

shun-iwasawa:
Currently Github Actions cannot re-run individual/incomplete jobs. 
As a workaround, I separated workflow by platforms so that it can re-run jobs individually by the "re-run all jobs" command.


RodneyBaker:
LGTM

Thank you @shun-iwasawa for pressing into Github Actions!

PR #3641, opened 2020-12-17
Modify Output Settings Layout

### LINKED ISSUE #2794: Output Settings window is not scalable and scrollable

shun-iwasawa:
This resolves #2794 
Output Settings popup is now resizable & scrollable, with category buttons for quick access.
I also revised the sizes of the combo boxes not to be full width of the popup.

Before

After (left : when expanded right : when shrunk)

### DIFF
--- a/toonz/sources/toonz/outputsettingspopup.cpp
+++ b/toonz/sources/toonz/outputsettingspopup.cpp
@@ -201,7 +201,7 @@ OutputSettingsPopup::OutputSettingsPopup(bool isPreview)
 
   QListWidget *categoryList = new QListWidget(this);
   QStringList categories;
-  categories << tr("Camera") << tr("File") << tr("Others");
+  categories << tr("Camera") << tr("File") << tr("More");
   categoryList->addItems(categories);
   categoryList->setFixedWidth(100);
   categoryList->setCurrentRow(0);
@@ -268,9 +268,9 @@ void OutputSettingsPopup::onCategoryActivated(QListWidgetItem *item) {
   } else if (item->text() == tr("File")) {
     label = m_fileLabel;
     frame = m_fileBox;
-  } else if (item->text() == tr("Others")) {
-    label = m_othersLabel;
-    frame = m_othersBox;
+  } else if (item->text() == tr("More")) {
+    label = m_moreLabel;
+    frame = m_moreBox;
   } else
     return;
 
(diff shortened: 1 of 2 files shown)


artisteacher:
What about using "More" instead of "Others"? If you do want to stick with "Others", "Other" would probably be more correct in that context.

RodneyBaker:
Outstanding!  Thank you for this @shun-iwasawa 
This has been a pesky problem for quite some time.

I will delay merging in case any discussion ks desired concerning @artisteacher's thought on the text of the dialogue box.


Additional:
I do have some thoughts related to the Output Settings window/panel but most of that belongs in 2021/v1.6 development and beyond.  There are however two that might be worth of consideration now versus later:

1) Capability of window/panel to dock inside of rooms (specifically the Farm room)
Most users don't venture into the Farm room because they believe it relates only to network rendering or simply think it contains things they have no need of using.  Being able to dock Output settings would allow for useful customization and even repurposing of the Farm room as a Render room or Output room

2) Capability to full screen the Output Settings window/panel
Being able to have render and output options all (or almost all) available and in view without resorting to dropdown menus etc.  will help the user know what options they have available to use.  

Again, these and other rendering/output options are best left to 2021 development cycle and beyond but I want to add it here for future consideration.  Both of these (and other output concerns) best belong in 2021 development cycle and beyond but I thought I'd mention these two as they allow much of the customization necessary to set the user up for success as other changes come.

RodneyBaker:
Thank you @shun-iwasawa for the modification!

gab3d:
@shun-iwasawa 
Thank you for the modifications, very welcome indeed!

I wonder why when selecting each category it still shows all the settings, instead of only showing those of the current one.
Perhaps it's a technical limitation of the window used or you assesed it has some advantages?
Just curious...

shun-iwasawa:
Thank you @gab3d for the comment.
I intentionally made it to scroll in a single panel instead of to switch between category pages.
By doing so you can still display and edit all setting items (or items only in `Camera` and `File` categories) at once by expanding the popup, which was available in the conventional UI and IMO has advantage in the sense of reducing necessary clicks.

PR #3642, opened 2020-12-18
Fix crash on quit OT with selecting stage schematic node

shun-iwasawa:
This PR fixes crash as follows:

To reproduce:
- Launch OT with Viewer and Schematic.
- Select either `Table` or `Camera1` node in the Stage Schematic. Animation tool is selected accordingly.
- Quit OT. It crashes instead of quitting.


RodneyBaker:

PR #3643, opened 2020-12-18
Fix Camera Settings response

shun-iwasawa:
This PR modifies response of the camera settings (used in Camera Settings popup and Output Settings).
Since the field is using `editingFinished()` signal, it emits the signal even when it just lose focus with the value unchanged. This fix will check if the values are actually changed or not and prevent unnecessary emitting of signals like `sceneChanged()` and `xsheetChanged()` which will cause updates in other panels.


RodneyBaker:
LGTM

PR #3644, opened 2020-12-18
Actions: use setup-python to specify Python version

shun-iwasawa:
This PR specifies Python version by using `actions/setup-python` so that it will be installed if it does not exist in the cache.

This PR specifies Qt version as will.


shun-iwasawa:
I'll merge it as the actions successfully finished.

PR #3646, opened 2020-12-18
Tahoma2D porting (Dec 18, 2020) by manongjohn and jeremybullock

shun-iwasawa:
This PR will port several bug fixes done in Tahoma2D as follows:  
- tahoma2d/tahoma2d#401 Fix current brush preset tracking
  by @manongjohn
- tahoma2d/tahoma2d#402 Fix audio memory leak and fix windows distortion
  by @jeremybullock 
- tahoma2d/tahoma2d#506 Fix guidedstroke in sceneviewer 
  by @manongjohn
- tahoma2d/tahoma2d#451 Fix loading brush presets or last brush used (partially)
  by @manongjohn
  This is partially ported as OT still does not have auto close/fill/group options in the vector brush. 
- tahoma2d/tahoma2d#504 Fix crash moving cell with shaderfx 
  by @manongjohn 
- tahoma2d/tahoma2d#511 Fix vector skeleton picking 
  by @jeremybullock
  I think it would better to fix [here](https://github.com/opentoonz/opentoonz/blob/86e8a1aec9025e923b54d09386ff6d85ef87a5ed/toonz/sources/toonzlib/stagevisitor.cpp#L256-L259) estimating that the distance is the 4th power of the pixel size, though...

Thank you @jeremybullock and @manongjohn !


RodneyBaker:
There was a glitch in building with appveyor for this PR but the Github Actions build all passed.
This is something worth watching in the future to see what differences there might be between the two builds.

Thanks @shun-iwasawa  for the port (and @manongjohn and @jeremybullock for the original fixes.

PR #3650, opened 2020-12-23
Making generation of translation projects optional in cmake

shun-iwasawa:
This PR enables to run `cmake` without generating translation projects.
This is needed when building with recent Xcode. Otherwise `cmake` fails because the same source are added to multiple targets which is not allowed by the Xcode "new build system".

With this PR, adding an command line option `-DWITH_TRANSLATION=OFF` enables to avoid the error.


RodneyBaker:
Thank you @shun-iwasawa!

PR #3652, opened 2020-12-23
Fix the onion skin fill and picking range for vector levels

### LINKED ISSUE #3651: Onion Skin Fill does not work properly when the Xsheet frame is selected
**Describe the bug**
When the xsheet frame is selected (i.e. in "scene editing mode"), the `Onion Skin` option of the Fill tool may not work properly, especially when the drawings are not placed in 1's or when the current column is animated.

**To Reproduce**
1. Create a new level (either Toonz Vector or Toonz Raster level) with 3 frames
1. Draw some closed shape in each frame. Fill the shape in the frame 1 with Style # 2 (red color) and the shape in the frame 2 with style # 3 (green color) respectively.
    
1. Make the cells arrangement `Step 2`
1. Select frame 5 (containing the drawing # 3) and extend the  onion skin handle to frame 3 (containing the drawing # 2). Note that the green fill of the drawing # 2 is shown in the viewer.
    
1. Select the Fill tool with `Onion Skin` option activated.
1. Click inside of the shape.

Although only the green color is appeared in the onion skin, the shape unexpectedly filled with red color. 

Furthermore the `Onion Skin` fill option could not pick the style correctly if the column is animated and the shape is offset between the frames.

**Expected behavior**
The `Onion Skin` fill option should pick & use the style of what you click.

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #3651 . Now the `Onion Skin` option of the Fill tool will consider the cell arrangement and the column transformation when the xsheet frame is selected.

This PR also fixes the range computation logic in `Stage::Picker` .
The change affects the behaviors of the tools with features of switching columns by picking the drawing in the viewer; such as Control Point Editor tool, Style Picker tool, Skeleton tool, Animate tool, and Shift&Trace feature. 
These tools are now consider the difference of scales between columns to be picked. Regardless of how the column is scaled, thin stroke can be picked from the same pixel-distance in the viewer.

### DIFF
--- a/toonz/sources/include/tools/stylepicker.h
+++ b/toonz/sources/include/tools/stylepicker.h
@@ -57,14 +57,15 @@ public:
 
   /*-- (StylePickerTool内で)LineとAreaを切り替えてPickできる。mode: 0=Area,
    * 1=Line, 2=Line&Areas(default)  --*/
-  int pickStyleId(const TPointD &point, double radius2 = 1, int mode = 2) const;
+  int pickStyleId(const TPointD &point, double radius, double scale2,
+                  int mode = 2) const;
 
   /*--- Toonz Raster LevelのToneを拾う。 ---*/
   int pickTone(const TPointD &pos) const;
 
   // per pli come sopra, ma ritorna il maincolor
   // per tzp e fullcolor ritorna il colore effettivo del pixel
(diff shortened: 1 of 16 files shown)


RodneyBaker:
I confirm the problem with the outlined steps and that this PR addresses the issue.

Issue #3085, opened 2020-02-02
ninja build broken


eszlari:
The ninja build worked in 1.3.0.
```
ninja: error: build.ninja:1488: multiple rules generate tnzcore/CMakeFiles/tnzcore.ts_lst_file [-w dupbuild=err]
```
https://flathub.org/builds/#/builders/48/builds/1691

(tested with cmake 3.13.4 & ninja 1.9.0)

RodneyBaker:
This could use an assist from someone else that uses ninja to build Opentoonz.
@eszlari The official build has never used ninja but was considering the merits of adding it so if you have additional information related to ninja related issues it will be appreciated.

eszlari:
Actually, the build with ninja fails because they changed a warning into an error between ninja v1.8 and v1.9: https://github.com/ninja-build/ninja/issues/931

Maybe you can take inspiration from this Krita commit to fix it:
https://github.com/KDE/krita/commit/e9b99ee5b75102a454dd728c81581dba92e19c4e

shun-iwasawa:
You can now avoid errors by adding an option `-DWITH_TRANSLATION=OFF` to cmake command.

eszlari:
But doesn't that mean, there are no translation when built with ninja?

shun-iwasawa:
The translation projects which will be generated with `-DWITH_TRANSLATION=ON` are utility projects for updating `.ts` files based on the current source code.
You can still use the [update script](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/translations/update.sh) or manually call `lupdate` to do the same thing.

And if you do not update the translation you don't need them and OT works as-is.

PR #3653, opened 2020-12-24
Actions: use the latest ninja

shun-iwasawa:
Just like Xcode, the latest version of Ninja does not allow to add the same source to multiple targets.  So I needed to install older version of ninja by rolling back Homebrew in Github Actions MacOS build.
Since #3650 was merged, now we can use the latest version of Ninja.

I also modified the MacOS build instruction.


RodneyBaker:
I believe this should address #3085

PR #3660, opened 2020-12-26
Fix crash on deleting stage schematic node

### LINKED ISSUE #3659: Crash when deleting a node from Stage Schematic.
**To Reproduce**
1. Create a new level.
2. In Stage Schematic, select the column node and delete it.
3. Crash
**Screenshots**
ezgif-7-f35303f6d87c

**Device Information:**
 - OS: Windows 8.1

shun-iwasawa:
This fixes #3659 .
Sorry, I broke it when I introduced the Expression Reference Manager. Thank you @beeheemooth for reporting the bug!


RodneyBaker:
Thank you @shun-iwasawa for the fast turn around on the fix.

PR #3662, opened 2020-12-28
Fix crash on saving palette level

shun-iwasawa:
This PR fixes the crash when overwriting palette levels by pressing the `Overwrite Palette` button in the Palette tool bar.
So sorry that I included the bug in #3387 .


RodneyBaker:
@shun-iwasawa I cannot seem to reproduce the crash on my end but must assume I am simply not using the same process.
Will merge this based on the simplicity of the fix along with your validation.

PR #3678, opened 2021-01-12
Cut Level Strip Selection Behavior Fix

### LINKED ISSUE #3612: (Discussion) Jumping to the first frame after cutting a frame in the Level Strip
**Describe the bug**
When you cut a frame from the Level Strip, it jumps to the first frame. If you have few frames it is not noticeable. But if you have a lot of them, then this is very inconvenient, since you constantly need to scroll to the very last frame.

**Expected behavior**
After cutting the frame (in Level Strip), it would be more logical for the jump to occur to the previous frame (if any).

**Screenshots**
ezgif-7-8dc9363db2b7

(I don't know which shortcut is more appropriate: "bug" or "feature request" for this issue. In my opinion, this is a bug.)

**Device Information:**
 - OS: [e.g. Windows] 8.1

shun-iwasawa:
This fixes #3612 .
Now the current frame will be jump to a previous frame of the cut frames, instead of the first frame of the level.

### DIFF
--- a/toonz/sources/toonz/filmstripselection.cpp
+++ b/toonz/sources/toonz/filmstripselection.cpp
@@ -238,10 +238,15 @@ void TFilmstripSelection::copyFrames() {
 void TFilmstripSelection::cutFrames() {
   TXshSimpleLevel *sl = TApp::instance()->getCurrentLevel()->getSimpleLevel();
   if (sl) {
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Merging this as it should pick up on the changes to Github Actions in subsequent builds.
Thank you @shun-iwasawa for the quick turnaround on this enhancement.

PR #3679, opened 2021-01-12
Motion Blur Fx Iwa : default value change

shun-iwasawa:
This PR modifies the default value of the `Index` parameter of Motion Blur Fx Iwa.
It is changed from `0` to `1` since this value is meaningless when less than 1.

Also the `Index` field is now sensitive to the `Reference Object` value.
When the `Reference Object` is set to `Own Motion` or `Table` , the `Index` field becomes hidden.


RodneyBaker:
Will delay merger of this PR long enough to check to see what (if any) changes the adjustment to Github Actions for CI builds on MacOS might make.

RodneyBaker:
Will press forward with merging this although the Mac builds have not yet caught up with the updated code for building...

Thank you @shun-iwasawa for the update to Motion Blur FX Iwa!

PR #3686, opened 2021-01-15
Fix redundant preview invocation

shun-iwasawa:
This PR fixes redundant preview invocation when deleting multiple fx nodes in the Fx Schematic as shown in a GIF below.

preview_problem

This behavior was very cumbersome especially when working on the time consuming fxs to render as it multiplies the rendering time.
This problem was due to the signal `xsheetChanged()` emitted for each deleting of node. Deleting multiple nodes caused multiple preview update requests. I fixed this to emit the signal once at the end of the operation.


RodneyBaker:
Checking this out now.
It looks like the Mac builds are back online. (Thanks for that @shun-iwasawa!)

PR #3696, opened 2021-01-24
Fix Color Tab on Style Editor

konero:
Removes the negative margin set for tab area.

It was previously a dirty fix, but since I modified it in the source awhile back this line is no longer needed in the style sheets, and actually causes the tab to go off screen a bit. Removing it makes the color tab align properly again.

cap1


RodneyBaker:
Thank you @konero 
Will merge once all checks have successfully completed.

PR #3697, opened 2021-01-24
Update Default Room Set

konero:
I know a lot of people discuss this over the many months, maybe years, so I'm attempting to update the default set of rooms, this should hopefully now offer a good balance between absolute beginner up to advanced users, by following a production style flow to the layout.

The new Drawing room which is now the first room that opens should be more than enough for anyone to become acquainted with the program and start animating with a minimal learning curve, while at the same time being useful for people of any skill level. The Studio Palette is an advanced feature, so it doesn't make sense having this in the first room. I've used the Timeline view over the Xsheet due to its familiarity to other software such as Toon Boom, Animate CC, Krita, so if somebody already has experience with these, then there's a good likelihood they'll have little problem adapting.

I've added an InknPaint room intended for inbetween or finished line and paint. It offers a similar experience to the old Drawing room layout by maximising space for the canvas. If the color model panel is unnecessary for one user, they can swap it for something else as I want to avoid duplication as much as possible to keep it simple. I've added this for communicative reasons as well, it's obvious by the name what this room is for, and it's much easier to explain over a Zoom call to somebody. Blender has well-named layouts which makes communication a lot easier.

I added a new room called Schematic which is for a bigger view of the schematic and function editor panels, mainly for FX compositing and useful for overlaying floating windows when necessary. I reconfigured the Palette room, as even on a QHD display the current layout is really limiting, I'm pretty sure that's the old layout from Harlequin, which wasn't great in the first place in my opinion. The other rooms I've kept the same.

I've also added extra commands to the toolbars, since most commands have icons now.

**Command Bar:**
image

**Xsheet/Timeline Bar:**
image

Placing auto input cell number here is similar to Clip Studio's batch specify button.

****Preview of New/Changed Rooms** (click to view full-size):**


RodneyBaker:
If anyone wishes to try this new room setup out, here's a link to the automated build:

(Windows)
https://github.com/opentoonz/opentoonz/suites/1889569118/artifacts/36952235

(Mac)
https://github.com/opentoonz/opentoonz/suites/1889569115/artifacts/36954961

(Linux - Ubuntu)
(clang build)  https://github.com/opentoonz/opentoonz/suites/1889569117/artifacts/36951646
(gcc build) https://github.com/opentoonz/opentoonz/suites/1889569117/artifacts/36951647

RodneyBaker:
image


I'd love to add the timeline to what is seen above but there is a ton of header space in the timeline...  hmmm.... another aspect of Timeline that I don't care for is how much effort it can take to add a layer/column above or below the current one.  With xsheet this is trivial.

konero:
I'm closing this for now, no time to work on this, I'd want to test it on someone who hasn't used OT before and stuff. Others can have a discussion about it or submit a new PR. 👍 

The point Rodney makes about timeline is something to look out for, it can be complicated managing columns at the moment.

DarrenTAnims:
To be fair, adding a blank column is fairly trivial in the timeline; just press the insert key. Or use the context menu to choose to insert it above or below. But I agree that always having a blank row in the xsheet is nice. But you still have to drag it around, so I still find myself hitting the insert key before I paste a drawing. So once you're passed using 3 or 4 columns, the timeline and the xsheet are pretty similar to me.
But the issue is that most new users would expect to see a timeline, so the first room they see should have one.

gab3d:
At first sight all I can say is:
- I'm not especially in favor of the Timeline vs the Xsheet, but nevertheless I like the proposed Drawing room. 
- I don't like the Animate room, without it having a way to select the desired column/layer. For that I think it should include the Xsheet (not the Timeline, because the Function Editor is not prepared yet for a horizontal mode, and it'd be inconsistent).
- As I haven't really used the Palette or Ink&Paint rooms, I leave that to others more involved in those workflows.

On a more general note, I think there should really be a conceptualization of the proposed use cases for a room to exist and, derived from that, to conclude why a certain proposed layout is the most appropriate one for the task.

artisteacher:
As far as having the first room optimized for new users goes, it might be best to keep the level strip out of it. Newbies always get confused by the level vs. scene display in the viewer.

You get used to the Timeline and it's really helpful for displaying more frames in a small space. I do wish that you could drag and drop upwards to create an empty layer though. There's no insert key on Mac keyboards, so some of us are limited to using the context menu to insert. If you create a new level with a cell in frame 1 active, the new level gets placed above the active level - that's easy enough. However if you need a level that starts later in the animation, it's a little annoying to setup from the Timeline.

I don't particularly feel that the Function Editor has to be the same orientation as the Xsheet/Timeline. I often use a small screen at work, so it's hard to fit a usable Function Editor and Xsheet in the same room.

gab3d:
yes, small screens are a problem with OT. I feel some thinking should be put on how to improve the experience for les than 1080 screens (720 is already uncomfortable to work with), as many so notebooks have less than 1080 screens.

I also agree that drag & drop to create a new layer would be ideal.

With respect to the Level Strip I feel it´s useful that it's there to raise the "problem" of what it represent. We cannot aspire for new users to be able to really use the program if they don't understand some of the most basic concepts under which it works, and Levels are one of them. 
To put it in reverse: how can we expect new users to successfully use the program if we consciously try to hide the basic concepts from them? :thinking:

gab3d:
All in all, I would like to stress that having a new general purpose default room (whatever it is called) should be achieved before we release 1.5, as the current one leaves much to be desired in terms of screen usage efficiency, and is a simple task to complete from a technical point of view. So we should be able to come to an understanding on a new, improved, room I guess...

What do you think?

konero:
For small screens I think the user will want to full screen the panel they're currently working in like the viewer. On the official site 1024 is listed as the minimum requirement for height, and it would be extremely difficult to fit a viewer and timeline in the same room at 720 without the viewer being too small to be useful. That titlebar hiding option found in Tahoma 2D would probably help, but only by about 20-40 pixels.

I set my monitor to 720 and I could not work with it, what I did was make the drawing room just a viewer with toolbar and option bar, then made a new room to the right called timeline, and just flip back and forth to each room. In Clip Studio Paint I actually work this way on a 4K monitor, I map the timeline to a shortcut key which opens/closes the window.

I set my monitor to the minimum requirement listed on the site, this is how the drawing room would fit.

image

If a new room set is decided (I strongly agree they should be), I can reopen and update the PR.

konero:
I updated the top post with some new previews. How about these? Anything to recommend?

RodneyBaker:
> ... a new general purpose default room (whatever it is called) should be achieved before we release 1.5,,,

I am often wrong but I don't see this default room set change going into v1.5 initial release.
It's a bit too late in the game for major change to interface and such changes should gain as much feedback as possible prior to such change.
By all means the changes should press into v1.5 immediately after the release as then users would be able to test and discuss the additions via use of the nightly releases.  This would in turn bring about more testing and more use of the nightlies for their intended purpose.  Then when those involved feel satisfied with default room updating they can simply stop submitting PRs and the nightlies will reflect the desired look for v1.5 and beyond... until the next iteration of default room changes.

Also, any changes to WIP PRs can be downloaded and used as if they were nightlies via the automated builds for PC, Mac and Linux.  This further take the impetus off rushing into change because those that desire a given change need only download the artifact they have a copy of Opentoonz running with that change.

konero:
> I am often wrong but I don't see this default room set change going into v1.5 initial release. It's a bit too late in the game for major change to interface

I would strongly urge updating them before a new release, I'd rather anyone new to the program not have to use the poor room layout we currently have and wait another year before they receive the update notification, it's too late by then.

This is a trivial change regardless, nothing that can introduce bugs or require any sort of code review to hold up a rollout.

gab3d:
I am firmly with @konero in this one.
It'd be nonsense to leave the current Basics layout (that's really bad) for another full year.
Let's take the risk!

pikhosh:
I like the new layout of Basics room; But when I opened the software, the size of the Level Strip and Palette panels was bad:

Screenshot from 2021-01-29 05-02-21

But I hope the OpenToonz interface looks more like Tahoma2D. Because Tahoma2D has a clean and comfortable interface and OpenToonz also has active development and a large community! And Tahoma2D has to constantly update itself with OpenToonz, and that's bad! 😕

I wish we had the Tahoma2D interface in OpenToonz so that we no longer need a separate software and waiting for new features to be ported to Tahoma2D. 

Of course, there are definitely reasons for many of the decisions, and I'm really grateful for everyone's efforts for OpenToonz and Tahoma2D. 👏🌹

(Also, I'm sorry for my bad English 😅)

konero:
> I like the new layout of Basics room, But when I opened the software, the size of the Level Strip and Palette panels was bad

@pikhosh I don't experience this using Windows. I'm curious if the same thing happens to you with the original Drawing room? Is there a way you could test this?

If anyone wants to test: [RoomTest1.zip](https://github.com/opentoonz/opentoonz/files/5893600/RoomTest1.zip)

You place them here: `C:\OpenToonz stuff\profiles\layouts\personal\Default.UserName`

pikhosh:
> I'm curious if the same thing happens to you with the original Drawing room? Is there a way you could test this?

With Nightly Build and replacing files; Everything was fine 👍

Screenshot from 2021-01-29 16-54-03

konero:
@pikhosh Thanks for testing again.

It's a strange thing to test, since I'm not experiencing it, not sure what the factor is.

I've updated the rooms again, if anyone could test Drawing and InknPaint room again: [TestRooms2.zip](https://github.com/opentoonz/opentoonz/files/5895283/TestRooms2.zip)

gab3d:
@konero 
In the **Drawing** room, I'd make the Toolbar full wide, and the Level Strip start just below it. 
It's a few height pixels less for Level Strip, but some much needed lot for the (usually crowded) Toolbar.

pikhosh:
> I've updated the rooms again, if anyone could test Drawing and InknPaint room again

This time the Level Strip panel was too small! 😅

Screenshot from 2021-01-29 22-25-15

konero:
@pikhosh Thanks for testing again!

I installed Ubuntu and decided to reconfigure them from there, if you could kindly try this set hopefully it fixes it. 🤞 

[TestRoomsFromLinux.zip](https://github.com/opentoonz/opentoonz/files/5897195/TestRoomsFromLinux.zip)

Also made the change to options bar according to @gab3d feedback.

pikhosh:
> @pikhosh Thanks for testing again!

You're welcome 😅

> I installed Ubuntu and decided to reconfigure them from there, if you could kindly try this set hopefully it fixes it. 

Yes, pretty much everything is fine, Thanks for the changes! 👍🌹

Screenshot from 2021-01-30 14-33-37

But the Tool Option Bar in the InknPaint room has not changed yet:

Screenshot from 2021-01-30 14-34-14

konero:
Thanks again! I noticed the Level Strip can be buggy when it comes to respecting its geometry and I have since managed to recreate what you were experiencing on Windows and Linux, but the bug is far more prevalent on Linux. I tested the current Default room set as well as the Studio Ghibli set and it happens there also, at least I now know it's not something I'm doing!

I looked at the issues and there's been a few about the Level Strip. The fix for which is not in the scope of this PR. IMO it should be made the size of the icons, not sure why that was changed.

I ran into a strange issue a couple times where on each launch of the program, the Level Strip would shrink in width by 2px each time.

I've pushed an update, and I tested on both a clean install for both Windows and Ubuntu, both showed the correct geometry on launch and kept their geometry on being set to full screen afterwards and rebooted.

It's still a possibility the geometry will break again depending on the end user screen resolution, or maybe other factors. In that case the user will just need to resize it again.

Update Files: [TestRooms3.zip](https://github.com/opentoonz/opentoonz/files/5898024/TestRooms3.zip)

gab3d:
@konero 
Agreed on Level Strip. 
Depending on vertical/horizontal orientation, there's no point in its width/height being respectively wider/taller than what it's strictly necessary to show the thumbnails completely (...plus the scroll bar, of course).
But, naturally that issue is for another PR.

Flairieve:
In the current version v1.4, I like how the drawing room has no timeline/xsheet and focuses on the canvas. I assume by all the screenshots that the drawing room is becoming the default room in OpenToonz? I do like how the new drawing room layout is but I feel like it should be just called the basics room if it includes stuff like the timeline and that the current drawing room in v1.4 should be kept.

shun-iwasawa:
Thank you @konero and folks for the discussion over updating the default room set.
I think I'll introduce this change so that we'll have further feedback from users during RC period.

PR #3705, opened 2021-02-02
Fix missing orientation dtor crash

shun-iwasawa:
This PR fixes crash when exiting OT on MacOS, which seems to be related to the following warning at the dtor of `Orientations`:
```
Delete called on 'Orientation' that is abstract but has non-virtual destructor
```

So I added the virtual dtor to `Orientation` and the crash seems to be fixed.


RodneyBaker:
I cannot verify results on Mac OS but the code itself appears self explanatory.
Will merge and await confirmation from our Mac users.

cc:  @artisteacher

PR #3712, opened 2021-02-08
Fix Deleting Fx Link

shun-iwasawa:
This PR fixes the following problem appeared after introducing #3686 :
- On Fx Schematic, selecting and deleting fx link lines ( without selecting any node ) does not seem to work.

In fact, the operation does not update the schematic as the necessary signal (`xsheetChanged()`)  is not emitted when deleting links.
I fixed it so that the signal is emitted once in the deleting operation.


beeheemooth:
Thanks for solving this problem! Yesterday I ran into this glitch and thought to write a report.
ezgif-7-1514d4695af9

RodneyBaker:
Thank you @shun-iwasawa for the fast turnaround on the fix.
LGTM

PR #3714, opened 2021-02-16
Icons for all commands

shun-iwasawa:
image

This PR does following fixes and modifications regarding icons:

- Made all icons visible no only in the command bar, but also in the main menu bar based on [the discussion with @konero](https://github.com/opentoonz/opentoonz/pull/3457#issuecomment-778889841) .
- To resolve [QMenu alignment issue](https://github.com/opentoonz/opentoonz/issues/3597#issuecomment-772480170), I made temporary icons to be generated for all commands to which the icons are not specified.
  - The temporary icon is transparent in 16x16 size (i.e. it is hidden in the menu bar) .
  - In 20x20 size (in the command bar), the generated icon is displayed and contains the first 2 capital letters in the command id.
     image
    For example, the last 4 buttons in the above are `Convert to Vectors` , `Tool Options Bar` , `Each 3` , `Quit` respectively. 
  - Modified menu padding in the style sheet.
  - Please note that the temporary icon is generated only in Windows version since the menu alignment is different and the issue seems not to occur in MacOS.
- Refactored `MainWindow::defineActions()` .
  - Icon names are specified in the argument of `createAction()` to reduce lines.
  - Commands are organized by categories.
- Slightly modified the icon for `Touch Gesture Control` and put into the same storage as other commands.


toonz/sources/toonzqt/gutil.cpp
```diff
@@ -339,6 +339,76 @@ QIcon createQIconOnOffPNG(const char *iconPNGName, bool withOver) {
...
+    QFont font = painter.font();
+    font.setPixelSize(pxSize);
+    painter.setFont(font);
+
+    painter.drawText(rect, Qt::AlignCenter, iconStr);
+
+    painter.end();
+
+    icon.addPixmap(transparentPm);
+    icon.addPixmap(pixmap);
```
konero, line 405:
```suggestion
    icon.addPixmap(pixmap);
    icon.addPixmap(setOpacity(pixmap, 0.15), QIcon::Disabled);
```

I think for now the default disabled icon generated by QIcon doesn't show when the icon is already white, since I think Qt makes a faint white version, it shows a little bit if the icon is dark.

I tested by adding _Right-click menu commands > Blend Colors_ to the Command Bar.

How about adding something like this? 0.15 is the opacity value for disabled icons from createQIcon.

artisteacher:
> the first 2 capital letters in the command id.

The problem with this is that the letters may not correspond to the correct words when translated

toonz/sources/toonzqt/gutil.cpp
shun-iwasawa: I added the semi-transparent icon for the disabled state. Thank you!

shun-iwasawa:
@artisteacher 
> The problem with this is that the letters may not correspond to the correct words when translated

Thank you for pointing it out. I managed to modify the fix so that the temporary icon now uses the command name (in English) instead of the command id.
Now the temporary icon for `Level Strip` command becomes `LS`. It was `FS` before the fix since the command id of the `Level Strip` is `MI_OpenFilmStrip` for historical reasons.

konero:
@shun-iwasawa This looks good! It's a satisfying refactor.

I've resized all the menu icons to 16x16, some needed a design change but only very few of them, while I was gong over them I took liberty to make better icons for Sub-Xsheets. I tried REALLY hard to completely fill the menus as well, but some like "Expose in Level Strip" and "Convert to Vectors" and very few others I just couldn't visualize for the life of me.

I'm not that experienced with git so forgive me, I cloned your branch and worked from that, you can view it here: https://github.com/konero/opentoonz/tree/icons16. The history is a bit messy, even though it built locally I was having issues pushing at one point. I don't know if you'd rather pull changes directly from it and submit here or in a new PR after testing it, or if I should submit the branch as a PR.

To review, simply go through the top menu, make a new level and right-click one of its cells in the Xsheet and go through some sub-menus, and browse to an image file in the Filebowser and right-click it.

image image

_(I'm going to bed now 💤 just in-case you respond too fast!)_

shun-iwasawa:
Thank you @konero ! It looks great. I cherry-picked and shrunk your commits.
This PR will be merged after introducing #3674 and resolving conflicts with it.

le-dragon-dev:
Outch, there is no space on linux.
image

shun-iwasawa:
@le-dragon-dev 
Thank you for the notification. I set up a virtual machine with Ubuntu 20.10 and confirmed the problem with the Actions' artifact.
OTOH I tried building with Qt 5.14.2 and the space seems to be with no problem.
image

The Actions' artifact seems to be based on Qt 5.9.5. I think the difference of Qt versions seem to be the cause of the problem.

So, I will try to modify the problem by following:
- Make OT to check running Qt version and if it is the same or less than `5.9.x` then add the following style sheet to `Preferences > Additional Style Sheet` as the default value:
  ```
  QMenu::Item {
    padding: 3 28 3 28;
  }
  ```
- Update the Actions' runner environment to `Ubuntu 20.04` .

gab3d:
in Level Strip, Toggle Orientation option seems to be out of place:

imagen


also, some menu options are juxtaposed with shortcut info, like these ones:

imagen

konero:
@gab3d The first issue should be fixed by this PR, if you can test by downloading an artifact build: https://ci.appveyor.com/project/opentoonz/opentoonz/builds/37850668/job/f4v12tp4504v4yb0/artifacts

le-dragon-dev:
Works well on Linux Mint 20.1! 🎉

shun-iwasawa:
Thank you @le-dragon-dev for verifying the fix!
I think I'll merge this and let it be tested further in the nightly release.

RattusaurusRex:
Icons on xsheet toolbar, command bar, window bars, etc are pixelated when highlighted on 4k display.  Toolbar tool icons and icons in menus seem fine.

icons
icons2

konero:
@RattusaurusRex I think I got a fix for this in #3740.

image

PR #3719, opened 2021-02-19
Fix New Level Undo

shun-iwasawa:
This PR fixes undo of `New Level` command to remove the created level from Scene Cast properly.
In `CreateLevelUndo` constructor, it registers level amount in Scene Cast before the operation for reversion.
Thus, the creation of the undo must be in advance of creation of the level.


RodneyBaker:
LGTM

Issue #3604, opened 2020-11-26
Opentoonz crashes when typing space in text tool


cardcraft:
**Describe the bug**
When using the text tool and type space in the text EX. "Hello World" the program crashes instantly 

**To Reproduce**
List steps to reproduce the behavior.
Create a new project
Use the text tool
Type something with a space

**Expected behavior**
i expected text with a space in it that can be animated with all the normal tools

**Screenshots**

**Device Information:**
 - OS: Fedora 32
 - OpenToonz Version: 1.4.0 Flatpak install
 - GPU: NVIDIA RTX 750 TI
 - GPU Driver Version: latest

RodneyBaker:
I cannot reproduce this on Windows.
Perhaps it relates only to Linux?

artisteacher:
Not an issue on macOS either.

le-dragon-dev:
Hello! I tested on Linux (Yep, it's the Type Tool) and I have an exception: TException "bad QImage format" (tfont_qt.cpp l.173)

orrin-ahola:
I don't know if this redundant but I am experiencing this behavior as well.

5.7.4-arch1-1
opentoonz 1.4.0-2
NVIDIA GeForce GTX 1050 Ti

le-dragon-dev:
> I don't know if this redundant but I am experiencing this behavior as well.
> 
> 5.7.4-arch1-1
> opentoonz 1.4.0-2
> NVIDIA GeForce GTX 1050 Ti
> 
> (Also happens on my Manjaro laptop)

Hello!

A fix has been made for this, can you try the following appimage?
https://github.com/opentoonz/opentoonz/suites/2190932894/artifacts/45181417

It will be release with the next version of OpenToonz ;)

orrin-ahola:
> > I don't know if this redundant but I am experiencing this behavior as well.
> > 5.7.4-arch1-1
> > opentoonz 1.4.0-2
> > NVIDIA GeForce GTX 1050 Ti
> > (Also happens on my Manjaro laptop)
> 
> Hello!
> 
> A fix has been made for this, can you try the following appimage?
> https://github.com/opentoonz/opentoonz/suites/2190932894/artifacts/45181417
> 
> It will be release with the next version of OpenToonz ;)

Yup, that worked!  I was able to push the space bar without ill effect.

RodneyBaker:
Closing this as fixed but tagging as 'known issue' in order to assist with those that might not be able to find the information.

Issue #2011, opened 2018-05-25
Pressing space when using text tool crashes program on Linux 


steakiestsauce:
### --Issue Summary--
When I try to enter a space while using the text tool, Opentoonz crashes. I tried with the morevna appimage as well as the AUR opentoonz build.

### --Steps to reproduce--
Open opentoonz
Click on text tool
Try to place text and then type ' '/space

### --Actual Results--
Program crash. Says invalid bitmap and segmentation fault

### --Expected Results--
A space is showed in the text and the program doesn't crash

### --System Information--
+ OpenToonz Version:
1.2.0 and (1.2.2 morevna version)

+ Operating System:
Arch Linux

shun-iwasawa:
Thank you @steakiestsauce for reporting.
Unfortunately I couldn't reproduce the crash with neither Windows nor OSX environment.
I may miss out something or this may be Linux-specific.

jpturcotte:
I am able to reproduce this bug on Lubuntu 18.04 with a build compiled on May 19 with a **Toonz Raster** level.

romeojulietthotel:
Last thing Opentoonz says is
 `terminate called after throwing an instance of 'TException'`

Not extremely helpful, I know.

And it only happens on a Toonz Raster level here too.

Maybe some kind of wide character problem?


From a gdb backtrace:

```
> #6  0x00007ffff3f2f237 in TFont::drawChar(QImage&, TPointT<int>&, wchar_t, wchar_t) const ()
>    from /usr/local/lib/opentoonz/libtnzcore.so
> #7  0x00007ffff3f2f476 in TFont::drawChar(TRasterPT<TPixelCM32>&, TPointT<int>&, int, wchar_t, wchar_t) const
>     () from /usr/local/lib/opentoonz/libtnzcore.so
> #8  0x00007ffff6a2e450 in TypeTool::replaceText(std::__cxx11::basic_string<wchar_t, std::char_traits<wchar_t>, std::allocator<wchar_t> >, int, int) () from /usr/local/lib/opentoonz/libtnztools.so
> #9  0x00007ffff6a31fef in TypeTool::keyDown(QKeyEvent*) () from /usr/local/lib/opentoonz/libtnztools.so
> #10 0x000055555586b2aa in SceneViewer::keyPressEvent(QKeyEvent*) ()
[...]

PR #3721, opened 2021-02-20
Fix crash in Type tool on linux (#2011, #3604)

le-dragon-dev:
Looks like for some reasons, the method alphaMapForGlyph (from QRawFront) returns an invalid QImage, even if the QRawFront is valid, only on Linux/Unix.

I also tested with the Qt version 5.15.2 and I had the same issue.

I did a workaround (not really pretty, but it works), really inspired from https://forum.qt.io/topic/61457/qimage-from-alphamapforglyph/9

Tested on my PC (Linux mint 20.1 Cinnamon, Linux 5.4.0-65, 16Go, NVidia GTX 1070)
image


RodneyBaker:
Thank you @le-dragon-dev for the workaround.
Hopefully other linux users can verify this works well on their builds as well.

Issue #3700, opened 2021-01-29
OpenToonz Places Full Screen Mode Drawing Area on Wrong Monitor


BloodyMess:
**Describe the bug**
I have OpenToonz maximized on my preferred display (Gaomon PD2200 Pen Display). When I choose "Full Screen Mode" from the contextual menu, it places the the drawing content area maximized on my other monitor, which is in portrait orientation. This is useless because then I can't use the pen display to draw on it.

**To Reproduce**
List steps to reproduce the behavior.
1. Launch OpenToonz
2. Maximize window on main display.
3. Choose "Full Screen Mode" from the contextual menu.

**Expected behavior**
I expect the full screen drawing area to simply stay on the same monitor that the window is already on.

**Screenshots**
Here is how my monitors are laid out:
![Screenshot from 2021-01-29 07-08-07](https://user-images.githubusercontent.com/12903370/106291897-0d25c300-6201-11eb-8ae0-d19a9bbf9fc5.png)

Here is how OpenToonz is putting the full screen drawing area on the wrong monitor:
![Screenshot from 2021-01-29 07-12-37](https://user-images.githubusercontent.com/12903370/106292306-858c8400-6201-11eb-95ba-bc5be2dfa450.png)

**Device Information:**
 - OS: elementary OS 5.1.7 Hera (based on Ubuntu 18.04 LTS)
 - Kernel: Linux 5.4.0-65-generic
 - GTK Version: GTK 3.22.30
 - OpenToonz Version: Built from current source checked out from master branch.
 - GPU: Sapphire Pulse Radeon RX590 8GB
 - GPU Driver Version: "xserver-xorg-video-amdgpu-hwe-18.04/bionic-updates,now 19.1.0-1~18.04.1 amd64 [installed,automatic]"

RodneyBaker:
This seems likely related to an issue where multiple monitors must be of the same resolution.

You might test that theory by making your screens mirror each other and see if that resolves the issue.

Some users have reported that tweaking their graphics settings and drivers has resolved this problem but in general the issue seems to persist with likely issues upstream related to the use of Qt and graphics cards.

BloodyMess:
Nope. No luck for me with mirroring or with just changing orientation.

[...]

PR #3728, opened 2021-02-23
Fix for Full Screen Mode on X11 with Multiple Monitors

BloodyMess:
I took [my own suggestion](https://github.com/opentoonz/opentoonz/issues/3700#issuecomment-781927440) and implemented it.

I was right about Qt always putting the window at (0,0), but I was initially looking at the wrong class and method. I implemented the fix in FullScreenWidget::toggleFullScreen() in imageutils.cpp in my fork.

It seems to be working fine, but I only have a Linux system to test it on. So I'm hesitant to create this pull request. Also, my coding style is a bit different from what's in OpenToonz, you might prefer to just look at my fix to see what I did (don't worry, it's super easy to read and very well commented.).


BloodyMess:
BTW: In case the code looks funky displayed in the web browser: I always edit using tabs for indentation, with a tab width of 3.

le-dragon-dev:
Hi! Thank you, I'll try this fix later today!

For the tabs/width 3: isn't it better to keep the same code aspect for all the project? (spaces, width 2)
I don't know if the script ./beautification.sh in the sources folder still works.

BloodyMess:
> For the tabs/width 3: isn't it better to keep the same code aspect for all the project? (spaces, width 2)
> I don't know if the script ./beautification.sh in the sources folder still works.

Yes, probably. I wasn't trying to suggest that the project change it's style, just giving a heads up as to why my code might look funky in your browser. That's one of the reasons that I mentioned you guys might want to just read it to see what I did to fix it.

I hadn't heard other users complaining about the full screen mode, which made me think there wasn't much demand for a fix for it, so I originally only intended the fix to be for myself. Once I found that I only needed to change a single function, I figured it couldn't hurt to make it available, and you guys could just do whatever you want with it in your project.

shun-iwasawa:
Thank you @BloodyMess for the fix, I'll check if your modification would work on Windows as well.

Sorry for troubling you, but it would be helpful if you could resolve the conflict. Thanks!

BloodyMess:
> Thank you @BloodyMess for the fix, I'll check if your modification would work on Windows as well.
> 
> Sorry for troubling you, but it would be helpful if you could resolve the conflict. Thanks!

Sorry, I didn't realize that that was an edit field on the Resolve Conflicts page. lol. I know C++ like the back of my hand. But I'm fairly new to using git and github for anything more than building projects from source.

shun-iwasawa:
I don't have multi monitor environment for MacOS but at least confirmed that it works with no problem with single monitor.

Regarding Windows I found another way to fix the problem and will post separate PR later.

LGTM thank you @BloodyMess for the fix!

PR #3730, opened 2021-02-25
Fix Viewer Window Disappearing on Undock

shun-iwasawa:
This PR fixes the problem reported in [the comment](https://github.com/opentoonz/opentoonz/issues/3597#issuecomment-783788804) , that the VIewer window disappears when undocking.

It was the problem with `FullScreenWidget` class which enables the panel to be displayed in full screen. (Thus, the same problem occured in Flipbook.)
It seems that, in recent Qt, calling `QWidget::winId()` at improper timing causes the problem related to OpenGL context handling.


konero:
This fixes it, thanks!

RodneyBaker:

PR #3731, opened 2021-02-26
Stylesheet Fixes for v1.5

konero:
Very minor fixes, no changes otherwise.

- Fixes scene cast panel when floating showing wrong border color
- Color correction for Neutral theme scrollbar handles
- Removes minor code duplication I found in qmenu indicator
- Makes menu icons and indicators a little bit more symmetrically spaced out
- Fixes sporadic bug seen in Linux with tree list branch icons, setting a transparent image instead of empty URL seems to solve it
- Color correction to better differentiate child and mesh columns in darker themes (they were both almost same color)
- Restore the darker column color for FX as numerous people have asked me about it (revert the change from #3457)
- Correct the file browser bg color for Dark theme so that each half of the panes are easier to differentiate due to some environmental lighting conditions


konero:
You can ignore the Win Build error, looks like just a server error, it built fine, plus there were no changes made here that could break anything. Or, hit re-run job for anyone with the permissions.

RodneyBaker:
Thank you @konero

PR #3742, opened 2021-03-02
Fix Icon Rendering for Hdpi

### LINKED ISSUE #3740: [Nightly] Problem Rendering Icons Above 100% Scale Factor
**Describe the bug**
There's a problem rendering icons when display scale factor is above 100% and the toolbars are using a pixmap smaller than the toolbar is set for (iconSize), icons become glitchy and enlarged. Seems to have happened after resizing the icons from 20x20 to 16x16. Playbar and Command Bars are iconSize = 20, I'm only guessing, but the pixmaps may for some reason be upscaling to fit 20x20? It looks OK at 100% scale though, so this is really confusing. Icons that are 20x20 look fine, it's only happening to icon files smaller.

Possible fixes...
- switch to PNG
- make all icons that appear in their toolbars a set size, so resize all 20x20 to 16x16 (that won't be easy design wise)

I'll keep investigating, but if anyone has any idea, I'd be grateful.

Icons pass through various functions in gutil.cpp:
- recolorPixmap
- setOpacity

**To Reproduce**
1. Set environment variable in Windows, QT_SCALE_FACTOR = 2 to set 200% for OT
2. Launch OT, look at Playbar, the icons have enlarged

konero:
Closes #3740, and as reported by @RattusaurusRex, menu icons (16x16) were difficult to repurpose for toolbars because they would resize up to fit 20x20, the theme config file did not help unfortunately, it seems no matter what SVG is going to scale up to whatever `iconSize` is in the toolbar, despite what's said in the docs.

There's also some issue with `setOpacity` function in `gutil`, it works absolutely fine for icons that don't get resized, so I've left it alone, but for icons that do it seems to make them look frayed around the edges, I don't know why. I built-in the `setOpacity` by using `QPainter` this time. We can look at that function later as it's now only used by the viewer pane buttons.

So to solve this, I reconfigured `createQIcon` and added a new function to allow compositing two pixmaps together, and loading this additionally into the generated icon. We composite all icons onto a canvas rect, this seems much more solid than doing it onto pixmaps. I also repurposed some recent code by @shun-iwasawa with the for loop on how the generated acronym icons were generated for Hdpi scaling (very nice).

I hope this is OK, this was really difficult to work out, not a fan of Qt's Hdpi support.

So in summary, two icons of different sizes will be generated into the icon, but ONLY if the icon to be generated is 16x16 in size, this action doesn't happen for any other size, which size is used depends where it's displayed, such as the menu or toolbars.

Before: 
image

After:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ Thumbs.db
 
 # bundled thirdparty libraries
 /thirdparty/boost
+/thirdparty/libjpeg-turbo64
 /thirdparty/lzo
 /thirdparty/tiff-4.0.3
 /thirdparty/LibJPEG/jpeg-9
--- a/toonz/sources/include/toonzqt/gutil.h
+++ b/toonz/sources/include/toonzqt/gutil.h
@@ -113,6 +113,10 @@ QPixmap DVAPI recolorPixmap(
     QPixmap pixmap, QColor color = Preferences::instance()->getIconTheme()
                                        ? Qt::black
                                        : Qt::white);
+QPixmap DVAPI compositePixmap(QPixmap pixmap, qreal opacity,
+                              int canvasWidth = 20, int canvasHeight = 20,
+                              int iconWidth = 16, int iconHeight = 16,
+                              int offset = 0);
 QIcon DVAPI createQIcon(const char *iconSVGName, bool useFullOpacity = false);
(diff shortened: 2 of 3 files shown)


RodneyBaker:
Let's give everyone a chance to test it thoroughly.
Thanks @konero for the fast turnaround on this.

PR #3743, opened 2021-03-02
Fix Temp Icon Over State in Menu

konero:
I overlooked this, need to add the transparent icon for the different modes, otherwise hovering causes the temp icon to display.


shun-iwasawa:
LGTM Thank you @konero for the fix!
Please don't care about the failure on GitHub Actions MacOS build which seem to be due to some cached data.

PR #3745, opened 2021-03-04
Optional Frame Separator Line in Level Strip

konero:
This makes it so it's easier to select a blank frame in the Level Strip and draw directly to the canvas. It's something that was useful in Harlequin it seems, since you scroll in the window to add 'ghost' frames that are selectable, but don't exist unless you draw.

It's optional (right-click under Panel Settings) and is hidden by default.

image

### DIFF
--- a/toonz/sources/toonz/filmstrip.cpp
+++ b/toonz/sources/toonz/filmstrip.cpp
@@ -1284,15 +1284,9 @@ void FilmstripFrames::contextMenuEvent(QContextMenuEvent *event) {
   QMenu *panelMenu           = menu->addMenu(tr("Panel Settings"));
   QAction *toggleOrientation = panelMenu->addAction(tr("Toggle Orientation"));
   panelMenu->addSeparator();
-  QAction *hideComboBox  = panelMenu->addAction(tr("Show Level Select Menu"));
-  QAction *hideNavigator = panelMenu->addAction(tr("Show Navigator"));
-  QAction *hideSeparatorLine = panelMenu->addAction(tr("Hide Separator Line"));
-  hideComboBox->setCheckable(true);
-  hideComboBox->setChecked(m_showComboBox);
-  hideNavigator->setCheckable(true);
-  hideNavigator->setChecked(m_showNavigator);
-  hideSeparatorLine->setCheckable(true);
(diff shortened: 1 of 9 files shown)


konero:
On first commit (I'm likely missing something) but OT doesn't load the setting if set as false on already created rooms when it adds the settings to the ini file, but if you create a new room it does. So I inverted the action to make sure it works in all scenarios. I originally changed the text strings but I reverted it and made my addition written in the same style to be coherent.

toonz/sources/toonz/filmstrip.cpp
```diff
@@ -1958,6 +1989,10 @@ void Filmstrip::load(QSettings &settings) {
   m_showNavigator = navigator == 1;
   m_frames->setNavigator(m_showNavigator);
 
+  UINT separatorLine  = settings.value("separatorLine", 1).toUInt();
```
shun-iwasawa, line 1992: > OT doesn't load the setting if set as false on already created rooms when it adds the settings to the ini file

This is because you set the second argument of `settings.value()` to `1`.
The second argument of this function is `default value` when the the setting doesn't exist.
I think it should be like `(m_showSeparatorLine?1:0)` .

konero: That was it, thank you!

shun-iwasawa:
This feature will be reviewed after releasing V1.5.

RodneyBaker:
This PR title edited to place it into the review queue.

RodneyBaker:
In revisiting this PR I forgot the part about the option being off by default... couldn't figure out why I wasn't getting the separation lines.  All good now though.  ;)

@konero Since this is your PR we will gladly defer to delay the merger of this.
I don't see any conflicts however that would prevent adding it.

From where I'm sitting... LGTM

konero:
@RodneyBaker Thanks for checking.

Indeed, no rush for this one! It was a dirty quick application from a colleagues request awhile back (I cleaned it up for the PR), I think the user-base might be so small however now that I think about it I'm not sure it's worth merging, and OT should get less options like this one, not more, IMO.

I'll close this for now, if a good case comes up later for introducing it it can always be reopened. 👍

PR #3746, opened 2021-03-04
Fix for Full Screen Mode on Windows with Multiple Monitors

### LINKED ISSUE #3700: OpenToonz Places Full Screen Mode Drawing Area on Wrong Monitor

shun-iwasawa:
This PR fixes Full Screen command on Windows, so that the widget stays on the same monitor as before using the command.
Together with #3728 , it resolves #3700 .

Please note that basically I only added a line calling `setScreen()` between `setWindowFlags()` and `showFullScreen()`, since the widget seems to lose the current screen after adding `Qt::Window` flag.
Most of other changes are by clang-format.

This resolution may work on other platforms as well but unfortunately I have no way to test it.

### DIFF
--- a/toonz/sources/toonzqt/imageutils.cpp
+++ b/toonz/sources/toonzqt/imageutils.cpp
@@ -904,165 +904,154 @@ void FullScreenWidget::setWidget(QWidget *widget) {
   if ((m_widget = widget)) layout->addWidget(m_widget);
 }
 
-
 //=============================================================
-	bool							FullScreenWidget::toggleFullScreen(
-//=============================================================
+bool FullScreenWidget::toggleFullScreen(
+    //=============================================================
 
-		const bool					kfApplicationQuitInProgress )		// Indicates whether the application is quiting.
+    const bool kfApplicationQuitInProgress)  // Indicates whether the
+                                             // application is quiting.
 
 /*
- *	DESCRIPTION:
+ *  DESCRIPTION:
  *
- *		Sets the size and location of the widget to either full
- *		screen or normal.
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thank you @shun-iwasawa for the additional tweak.

PR #3747, opened 2021-03-04
Actions Fix: Update Path to Qt5 in MacOS Job

shun-iwasawa:
This PR will fix recent failures in MacOS build of GitHub Actions.
Homebrew recently updated to install Qt v6 by default, so we need to explicitly specify `qt@5` .


RodneyBaker:

PR #3751, opened 2021-03-05
Fix Scene Cast to Update on Pasting

### LINKED ISSUE #3750: Scene Cast does not update on pasting
**Describe the bug**
The Scene Cast panel is for managing assets of the scene. All the levels used in the scene should be registered, or it will cause unexpected problems.

Some pasting commands (such as paste cells, columns, fx nodes and stage nodes) may re-introduce levels which has been removed from the scene cast. However, currently OT is not aware of it.

**To Reproduce**
1. Open Xsheet and Scene Cast panels.
1. Load or Create a level `A` in the column 1. (Level A is registered in the Scene Cast)
1. Select all cells containing the level A and `Cut` .
1. Use `Remove All Unused Levels` command in the Scene Cast context menu. (Level A is removed from the Scene Cast)
1. In Xsheet, `Paste` the cells. 

The Level A is pasted on the Xsheet, but the Scene Cast fails to register it.

You can wrongly create another level with name `A` since the duplication of the level name is checked based on the Scene Cast.
The same problem does occur when you paste columns, fx nodes or stage nodes.

**Expected behavior**
OT should make sure that the all pasted levels are registered in the Scene Cast.

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #3750 .

When pasting cells, columns, fx nodes or stage nodes OT will check the pasted levels and make sure they are registered in the Scene Cast. This is undo-able along with the pasting operation.

### DIFF
--- a/toonz/sources/include/toonzqt/fxselection.h
+++ b/toonz/sources/include/toonzqt/fxselection.h
@@ -156,6 +156,7 @@ signals:
   void doCollapse(const QList<TFxP> &);
   void doExplodeChild(const QList<TFxP> &);
   void doDelete();
+  void columnPasted(const QList<TXshColumnP> &);
 };
 
 #endif
--- a/toonz/sources/include/toonzqt/schematicviewer.h
+++ b/toonz/sources/include/toonzqt/schematicviewer.h
@@ -5,6 +5,7 @@
 
 // TnzLib includes
 #include "toonz/tstageobjectid.h"
+#include "toonz/txshcolumn.h"
 
 // TnzBase includes
 #include "tfx.h"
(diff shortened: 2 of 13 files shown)


RodneyBaker:

gab3d:
I'd like Scene Cast to have a more prominent role in scene management (in a much similar way as After Effects's Project panel). 
From there, there is a clear view of all assets that conform a scene, and it's from there also, that we could eventualy manage the names of sub-xsheets (if that feature is introduced) to be able to reference them in expressions, etc.

...just leaving the comment here for future reference. 😉

PR #3755, opened 2021-03-08
Auto Cleaning Assets Feature

shun-iwasawa:

This PR will add a new preference option which enables OT to remove unused levels from the Scene Cast automatically.

### What is this option for?

I think most users do not open the Scene Cast window or even know the existence of it.
After the level is removed from the Xsheet it remains in the Scene Cast until using the `Remove All Unused Levels` command.
Scene Cast does not allow multiple levels to have the same name.
Therefore, when viewed from users who are not familiar with Scene Cast, when they try to load or create some level it is sometimes "unexpectedly" renamed ( `_1` suffix is added ).

This option will let OT to remove unused asset automatically so that users can work without concerning about Scene Cast.
 
### Usage

Enable the option `Preferences > Loading > Automatically Remove Unused Levels From Scene Cast` .

The default value of this option is OFF (as it is still in trial phase) .

When this option is ON, OT will automatically do the following:

- When Saving Scene, automatically remove all unused level from the scene before saving.
  - The auto-removing is executed when using `Save Scene` , `Save Scene As` or `Save All` commands. The removing is NOT executed when silently overwriting by auto-save feature or when using `Save Sub-Xsheet As` command.
- When users create / rename / paste a level and if there is another level in the Scene Cast with the same level name, OT will check if the existing level is unused in the Xsheet and remove it from the Scene Cast in advance.
- When the level is removed by the above operation, an information message will be displayed in the Message Center window.

The all auto-removing operations are undo-able.

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -270,6 +270,9 @@ public:
   int matchLevelFormat(const TFilePath &fp)
       const;  //!< Returns the \a nonnegative index of the first level format
               //!  matching the specified file path, <I>or \p -1 if none</I>.
+  bool isAutoRemoveUnusedLevelsEnabled() const {
+    return isAutoExposeEnabled() && getBoolValue(autoRemoveUnusedLevels);
+  }
 
   // Saving tab
   TPixel getRasterBackgroundColor() const {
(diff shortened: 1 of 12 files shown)


RodneyBaker:

RodneyBaker:

Also re-noting that this option is off by default intentionally to allow users to trial it out.
Those interested in having it remain off by default or that desire it to be on are encouraged to provide their feedback.

PR #3757, opened 2021-03-11
Add New Option for Level Palette Toolbar Position

konero:
_**Note:** Please don't merge this right away, allow Shun Iwasawa time to review it when he has time, there's no rush._

This adds a new user option to change the position of the toolbar in Level Palette and Studio Palette panels. By default it will be shown at the bottom, you can change it back to the top using the option in the sub-menu. It will save independently of each panel.

**Note on Studio Ghibli Room**
------
I know there's a comment that the reason the toolbar is in the position it is is by staff request at Studio Ghibli, so I decided to try and gracefully handle this situation. I made it so if the current room choice is the Studio Ghibli room set then to keep the current behaviour, so the toolbar will default to being set on top for both currently docked and newly created floating panels after updating, no user intervention is required to set it back, it should be automatic.

toolbotm1

---

_There's a key value shared across all room sets for floating panels stored in `popups.ini` by current design, if you create a floating panel within the Default set, then switch to the Studio Ghibli set, then the behaviour might not be what you expected, since the key value is now set for bottom for floating panels. Setting your preferred option will overwrite it though, so it's not a problem, but keep that in mind when testing this._


RodneyBaker:
I like how it looks/operates on the bottom like that.  Nice idea!

gab3d:
This should be the default location of Palette buttons, it makes so much sense. 👏👏

RodneyBaker:
Will merge this to allow for sufficient time testing this with the release candidate.

konero:
Thanks, but I should have been more clear requesting not to merge without checking, I'm only semi-confident in the code, I don't want it to have corrupted any user profile settings or anything. 😬 

Although I tested it a great deal on some workstations and am confident using it personally, Shun might need to do the same at his end, as I made some bold assumptions in the default settings.

I know you're real busy @shun-iwasawa and you must have been pinged more times than a Google server lately, but before the final release could you double-check this one? I don't mind if it needs to be reverted in the meantime. 🙏

PR #3758, opened 2021-03-12
Premultiply on Loading PNG Images

### LINKED ISSUE #3718: Tools do not work properly on straight alpha raster images
**Describe the bug**
OT handles raster images assuming they are premultiplied. When loading straight alpha images (like regular PNG format images), setting `Level Settings > Premultiply` enables to display them properly.
However, the tools are not capable of it. They always assume that the image is premultiplied and behaves improperly.

**To Reproduce**
1. Create a PNG image with some drawing and transparent background by using external software like Photoshop.
1. Load the image on OT. 
1. OT automatically sets `Level Settings > Premultiply` by default. (This behavior can be modified in the Preferences)
1. Use the tools (Brush, Geometric, Fill or Selection) on it. 

For instance, with Brush tool, a white edge appears in the stroke only when you draw outside of the existing drawing.

**Expected behavior**
The tools should work properly on the straight alpha raster images as well.

**Screenshots**
image
Note that the red lines are drawn with OT tools.

**Device Information:**
 - OS: Win10
 - OpenToonz Version: latest

shun-iwasawa:
This PR will fix #3718 by changing loading strategy of PNG format.
From this PR, PNG files will always be premultiplied on loading.

After investigating the issue, I found that there is two problems as follows:

1. Tools are designed to work on premultiplied-alpha image. On straight-alpha image (like PNG format) they don't work properly. (as written in #3718)
2.  OT has inconsistency in I/O of PNG file format. i.e. when loading, OT loads the straight-alpha image as-is from the file. OTOH when saving [OT un-premultiplies the raster data](https://github.com/opentoonz/opentoonz/blob/45bac5a15c26f3cdd0ff4e624a3722d2cc063715/toonz/sources/image/png/tiio_png.cpp#L931) before saving it to the file, assuming that raster data is always premultiplied. This works OK when rendering in PNG format - since the rendered image is premultiplied - but overwriting the PNG image by `Save Level` command will cause wrong result by un-premultiplying straight-alpha image.

To resolve both of the problems, I decided to make the premultiply operation to be done in the loading process which is corresponding to the un-premultiply operation in the saving process.

In order to make the change not to affect the rendering result of scenes made with the previous versions, I also added the following:

- The option `Preferences > Loading > Level Settings by FIle Format > PNG` will be **silently** removed. 
- When loading the scene created by the previous version of OT, OT will search for PNG levels and automatically unset `Level Settings > Premultiply` option. The result is notified in the Message Center.
- When loading the scene as sub-xsheet level, OT will search for PNG levels and automatically and **silently** unset `Level Settings > Premultiply` option. 
- tcomposer will also automatically modify the level settings before rendering.

------

Related to PNG format I/O, I also fixed the following problem:

- OT fails to save dpi of the PNG image.

I found that `png_write_info()` is called before `png_set_pHYs()` , which I think should be swapped in order to write dpi information properly.

### DIFF
--- a/toonz/sources/image/png/tiio_png.cpp
+++ b/toonz/sources/image/png/tiio_png.cpp
@@ -204,7 +204,8 @@ public:
       png_set_palette_to_rgb(m_png_ptr);
 
       // treat the image as RGBA from now on
-      m_info.m_samplePerPixel = 4; // there are 4 channels per pixel (R, G, B, and A)
+      m_info.m_samplePerPixel =
+          4;  // there are 4 channels per pixel (R, G, B, and A)
 
       // if there is no alpha channel, then fill it with "255" (full opacity)
       png_set_filler(m_png_ptr, 0xFF, PNG_FILLER_AFTER);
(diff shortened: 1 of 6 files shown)


RodneyBaker:
The Appveyor build seems to be in a never ending cycle of building...

RodneyBaker:
The Appveyor build finally finished.  Yay!
Thank you @shun-iwasawa for the PNG fix.

PR #3763, opened 2021-03-15
Fix MacOS Crash on Recording Audio

### LINKED ISSUE #3762: macOS crash on record audio
**Describe the bug**
In the latest nightly build, pressing the record button on the "Record Audio" window causes an immediate crash. This is probably due to the changes in the build environment. 

This has been an issue in Tahoma as well (https://github.com/tahoma2d/tahoma2d/issues/158) but since Jeremy & @manongjohn both do not have macs, they have been unable to find a fix. The crash doesn't seem to be replicated in a VM. At this point, we think it is most likely a permissions issue.

**To Reproduce**
1. Open the latest nightly build on a computer running macOS
2. Open the record audio window
3. Press record = immediate crash

**Expected behavior**
The first time you record, there should be a pop-up notifying that OT wants to use the microphone. After granting permission, audio should record normally.

Right now, there is no request to give permission to use the microphone. It looks like there are a couple pieces to this - a key in info.plist which needs to be added. Then a test to check if permission has been granted before starting a capture session & prompting the user to grant permission if necessary.
https://developer.apple.com/documentation/avfoundation/cameras_and_media_capture/requesting_authorization_for_media_capture_on_macos

**Crash Report**
[audio_crash.txt](https://github.com/opentoonz/opentoonz/files/6138090/audio_crash.txt)

**Device Information:**
 - OS: macOS Big Sur
 - OpenToonz Version: Nightly (2021-03-14)

shun-iwasawa:
This PR fixes #3762

### DIFF
--- a/toonz/cmake/BundleInfo.plist.in
+++ b/toonz/cmake/BundleInfo.plist.in
@@ -34,5 +34,7 @@
 	<string>This Open Source Program is developed from Toonz, a software originally created by Digital Video, S.p.A., Rome Italy</string>
 	<key>NSCameraUsageDescription</key>
   <string>OpenToonz needs access to the camera in order to use Camera Capture</string>
+  <key>NSMicrophoneUsageDescription</key>
+  <string>OpenToonz needs access to the microphone in order to use Record Audio</string>
 </dict>
 </plist>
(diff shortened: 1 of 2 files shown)


artisteacher:
This works great! Thanks so much for the super quick fix.

PR #3764, opened 2021-03-15
Actions: Fix MacOS Library Paths

### LINKED ISSUE #3748: Recent macOS builds crash on startup
Recent macOS Actions (master) and nightly builds crash when starting on 10.15 (Catalina)

For GitHub Actions, it crashes for this:

> Dyld Error Message:
>   Library not loaded: @rpath/libopencv_video.4.5.dylib
>   Referenced from: /Users/xxx/Desktop/OpenToonz.app/Contents/Frameworks/libopencv_gapi.4.5.dylib
>   Reason: image not found

For Nightly, it crashes for this:

> Dyld Error Message:
>   Library not loaded: @executable_path/../Frameworks/EDSDK.framework/Versions/A/EDSDK
>   Referenced from: /Applications/OpenToonz.app/Contents/MacOS/OpenToonz
>   Reason: image not found

shun-iwasawa:
This will fix MacOS artifacts output from GItHub Actions, and will resolve #3748 .
Please note that now `macdeployqt` is called twice, the first call is for introducing dependent libraries and the second is for create dmg after modifying library paths.

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -50,12 +50,12 @@ jobs:
         cd build
         cmake ../sources -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DQT_PATH='/usr/local/opt/qt@5/lib' -DWITH_TRANSLATION=OFF
         ninja -w dupbuild=warn
-
-    - name: Create Artifact
+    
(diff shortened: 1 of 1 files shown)


artisteacher:
I can confirm that the artifact build now runs properly. Thank you!

PR #3767, opened 2021-03-16
Fix Batch Rendering on Mac

### LINKED ISSUE #2940: Batch Rendering fails
### Issue Summary

On macOS, batch rendering through the tasks window fails.

tasks

### Steps to Reproduce

1. At the top of the Tasks window, click on "Add Render"
2. Select a scene
3. Click on the start button or right-click on a task & select "start"
### Expected Results

The render to be successful.
### Actual Results

The render fails. If you look at the path in the screenshot, the path uses OpenToonz_1.3.app instead of OpenToonz.app. Since the app no longer uses the version number as part of the name, that is likely the cause of failure.
### System Information
* **OpenToonz Version**: latest nightly, most recent OTX
* **Operating System**: macOS 10.15.1
* **CPU**: 
* **Memory**: 
* **Graphics Card**: 
* **Graphics Tablet**:

shun-iwasawa:
This PR will fix #2940 . Not only `tcomposer` , but also `tcleanupper` and `tconverter` are (hopefully) fixed.

### DIFF
--- a/toonz/sources/tcleanupper/tcleanupper.cpp
+++ b/toonz/sources/tcleanupper/tcleanupper.cpp
@@ -491,6 +491,7 @@ int main(int argc, char *argv[]) {
   // questo definisce la registry root e inizializza TEnv
   TEnv::setRootVarName(rootVarName);
   TEnv::setSystemVarPrefix(systemVarPrefix);
+  TEnv::setApplicationFileName(argv[0]);
 
   QCoreApplication::setOrganizationName("OpenToonz");
   QCoreApplication::setOrganizationDomain("");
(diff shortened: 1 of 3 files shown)


artisteacher:
Works great except for FFmpeg formats! You are on a roll!

shun-iwasawa:
@artisteacher 
Hmm, in my environment I could render mp4 via tcomposer.
Can you please check if FFmpeg path is properly specified in Preferences? (As OT does not bundle FFmpeg yet.)

artisteacher:
Oh, that's probably it! I've gotten so used to Tahoma bundling it. I'll confirm that it works tonight.

artisteacher:
FFmpeg formats do render when FFmpeg is installed.

PR #3769, opened 2021-03-16
Fix full screen feature

shun-iwasawa:

- Toggle full screen mode of the Viewer
- Use `Save All` command (via shortcut key)
- OT becomes uncontrollable, since the screen stays on top of the modal dialog


RodneyBaker:

PR #3770, opened 2021-03-17
Fix Main Window Full Screen

### LINKED ISSUE #3734: Bug in Full screen mode
**Tools are invisible in full screen mode (im using single monitor)**
They are working, But They are invisible

**To Reproduce**
1. Install the last nightly version
2.Go to      Windows ->Toggle Main Windows Full screen mode
3.Try to save the file (for example)

**Screenshots**
The problem is that when i was trying  to rec my screen with OBS in video it is  working and visible. But in real its invisible. So when i rec it i dont see the tools , but in rendered video they are visible .I create 2 videos. 
1: i rec my screen with OBS- https://youtu.be/RqkFEdc9pns   (When i rec this i dont see tools, only aftrer rendering video they are shown)
2 I rec my screen with my phone _https://youtu.be/dVU0u_AkDEs  (There u can see that they are invisible in full mode)

**Device Information:**
 - OS: Windows 7
 - OpenToonz Version: [nightly build 2021-02-26]
 

 Sorry for my  english )

shun-iwasawa:
This will fix #3734 
It needed the same modification as the viewer's full screen feature to work properly on Windows.

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -53,6 +53,9 @@
 #include <QPushButton>
 #include <QLabel>
 #include <QMessageBox>
+#ifdef _WIN32
+#include <QtPlatformHeaders/QWindowsWindowFunctions>
+#endif
 
 TEnv::IntVar ViewCameraToggleAction("ViewCameraToggleAction", 1);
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #3774, opened 2021-03-17
Fix undo when removing entire cells in the column

### LINKED ISSUE #3709: Crashing when opening a scene. Please HELP.
**Describe the bug**
Hi, I was working on a project, and I decided to use the plastic tool. I then applied the mesh fine, it worked, But I did not like the way it looked. So I deleted the Plastic and mesh layer, and saved, and exit. But then, Whenever I try to open that project, It crashes opentoonz. So I cannot get into the project. And I need that project ASAP.
**To Reproduce**
1.Apply Plastic and mesh covering and Rig it all up.
2.Save. Exit.
3.Try and get back into the project.

**Expected behavior**
Whenever you use the plastic tool and delete the layer, I expected for it to just go back to what it was before, not crash it again and again.

**Screenshots**
https://user-images.githubusercontent.com/73257686/106827134-21bced80-663d-11eb-8c4f-182c12a88d64.png

**Device Information:**
 - OS: [Mac]
 - OpenToonz Version: [1.4]
 - GPU: [Intel HD Graphics 6000 1536 MB]
 - GPU Driver Version: [NA]

shun-iwasawa:
Resolves #3709 , in the sense that this will fix the cause of the scene-corrupting problem.

This PR fixes undo behavior when cutting or deleting the xsheet cells, especially when the column becomes empty after the operation.

I've tried the resolution tahoma2d/tahoma2d#575 by @manongjohn but I decided to take different way with the purpose of recovering all information of deleted column such as keyframes, visibility info etc.

Instead of adding codes to `DeleteCellsUndo` , I decided to exploit `DeleteColumnsUndo` to take care of column recovering. They are stacked in the undo block when deleting cells will empty columns.

From this PR deleting cells may delete columns, so the warning popup will open if the preference option `Automatically Modify Expression On Moving Referenced Objects` is ON and the operation will lose expression reference.

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -14,6 +14,7 @@
 #include "tapp.h"
 #include "xsheetviewer.h"
 #include "levelcommand.h"
+#include "columncommand.h"
 
 // TnzTools includes
 #include "tools/toolutils.h"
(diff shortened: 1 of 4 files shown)


RodneyBaker:

PR #3775, opened 2021-03-17
Fix XDTS popup duplicative fields

### LINKED ISSUE #3773: Duplicative fields appears when loading XDTS including multiple columns with the same name.
**Describe the bug**
When loading XDTS, a popup opens for specifying the levels associated with each column.
However, if multiple columns have the same name, duplicative fields appears with the same level name, which is unnecessary and confusing.

**To Reproduce**
1. Load XDTS file including multiple columns with the same name.

**Expected behavior**
Unify the fields.

**Screenshots**
image

**Device Information:**
 - OS: Win10
 - OpenToonz Version: latest

shun-iwasawa:
This fixes #3773


RodneyBaker:

PR #3781, opened 2021-03-18
Fix Palette Glitches

### LINKED ISSUE #3689: Two glitches when merging Global Palettes with the current palette.
**Glitch 1:**
1. Create new level (any type)
2. Drag a palette from the Global Palettes into the current palette (or use the Merge to Current Palette command) so that the palette will automatically create a new page.
3. Draw something using the styles of this palette.
4. Save All and restart the scene.
5. The palette you added (and the page) will be lost and the drawing will be red.
6. If you load the same palette again, the styles color will return to its place. 

ezgif-7-064a54b7042f

**Glitch 2:**
1. In the Level Strip, select the frame and clear it (by pressing Delete).
2. Together with the drawing, the loaded palette (which is located on another page) will disappear.
3. If you have several frames drawn and you delete one, the loaded palette will not disappear, however, after restarting the scene, the drawings will be red.

ezgif-7-121b2189307f

The palette disappears the moment you move from the Level Strip to the Xsheet. If at the same time you have selected any style of the disappeared palette, then the Brush Tool will draw red strokes.

**Device Information:**
 - OS: Windows 8.1

shun-iwasawa:
This PR fixes #3689 

- Merging to the current palette command had wrongly set the dirty flag to the source palette, instead of the destination palette where the source palette is dropped.
- Using the Clear command to the level with one frame had wrongly initialized the palette.

### DIFF
--- a/toonz/sources/toonz/filmstripcommand.cpp
+++ b/toonz/sources/toonz/filmstripcommand.cpp
@@ -404,8 +404,11 @@ std::map<TFrameId, QString> clearFramesWithoutUndo(
                  "-" + QString::number(it->getNumber());
     TImageCache::instance()->add(id, sl->getFrame(frameId, false));
     clearedFrames[frameId] = id;
+    // empty frame must be created BEFORE erasing frame or it may initialize
+    // palette.
+    TImageP emptyFrame = sl->createEmptyFrame();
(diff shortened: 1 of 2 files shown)


RodneyBaker:

PR #3782, opened 2021-03-18
Fix Alt-drag Fx Node Insertion

### LINKED ISSUE #3600: Crash when binding / unbinding an FX node to multiple columns using ALT-drag
**Describe the bug**
This crash was hiding for a long time, but I caught it.
A crash occurs when you bind (and then unbind) an effect to multiple columns using ALT-drag.

**To Reproduce**

1. Create several levels (columns) and select one of the effects to which you can connect many columns (for example Over).
2. In the FX Schematic window, hold down the ALT button and move the effect so that all speakers are connected to it. Don't let go of the buttons!
3. As soon as there is a connection between all the columns and the effect, first release the ALT key, then the left mouse button (if you do it differently, then only one column will join the effect)
4. Now, in the same way, tear the effect out of the chain: hold down the ALT button and move the effect so that all links move away from the effect.
5. Click with the left mouse button on an empty space in the FX Schematic. A crash will occur.

ezgif-7-ff6ff7ede30f

**Device Information:**
 - OS: Windows 8.1
 - OpenToonz Version: latest

shun-iwasawa:
This PR fixes #3600 

Making sure to remove all simulation-links before adding new link to prevent the previous simulation-links remain connected to the port.

### DIFF
--- a/toonz/sources/toonzqt/fxschematicscene.cpp
+++ b/toonz/sources/toonzqt/fxschematicscene.cpp
@@ -490,7 +490,7 @@ void FxSchematicScene::updateEditedGroups(
     const QMap<int, QList<SchematicNode *>> &editedGroup) {
   QMap<int, QList<SchematicNode *>>::const_iterator it;
   for (it = editedGroup.begin(); it != editedGroup.end(); it++) {
-    int zValue = 2;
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #3783, opened 2021-03-19
Fix Crash on Inserting New Frame

### LINKED ISSUE #3613: Crash (and freezing) when creating new Frame
**Describe the bug**
OpenToonz starts freezing and then crashes, when drawing the inbetween onto a new Frame

**To Reproduce**

_1.Create a new Level_
_2.Draw 2 Keyframes on 2 seperate frames (every keyframe lasts 2 frames)_
image
image

_3. Draw the inbetween on Frame 2 ( which overwrites the copy of Frame 1), and also use Ctrl+Z_
image
image

_5. Watch as the Program freezes for a moment and then crash without any Messages or notifications_

**Expected behavior**
Inbetween Frame should be drawn as often as needed whilst also using Ctrl+Z/Undo, without the program Crashing

**Device Information:**
 - OS: Windows 10 Pro 1909
 - OpenToonz Version: Nightly Dec 1 2020 10:18:55 as well as the last Release (Feb 1)
 - GPU: GTX 1050Ti
 - CPU: Ryzen 7 3700x
 - RAM: 16GB

shun-iwasawa:
This PR fixes #3613 
This fixes erasing level frame to remove icon cache as well.
After investigation I found that icon images of the removed frames were remained in the cache, which cause malfunction when remapping the cache. ( In particular, `TImageCache::Imp::m_itemsByImagePointer` gets duplicative entries )

Cache remapping occurs when undoing insertion of new frame. Which causes malfunctions in the cache, followed by the crash.

### DIFF
--- a/toonz/sources/toonzlib/txshsimplelevel.cpp
+++ b/toonz/sources/toonzlib/txshsimplelevel.cpp
@@ -840,10 +840,15 @@ void TXshSimpleLevel::eraseFrame(const TFrameId &fid) {
   getHookSet()->eraseFrame(fid);
 
   ImageManager *im = ImageManager::instance();
+  TImageCache *ic  = TImageCache::instance();
   {
     im->unbind(getImageId(fid, Normal));
     im->unbind(getImageId(fid, Scanned));
     im->unbind(getImageId(fid, CleanupPreview));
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #3785, opened 2021-03-19
Actions: install libomp for Linux job

shun-iwasawa:
This will fix recent failure in Linux job of GitHub Actions by installing necessary library.


shun-iwasawa:
confirmed that Linux job completed

PR #3786, opened 2021-03-19
Fix OpenGL Line Smoothing

shun-iwasawa:
This PR will fix OpenGL line smoothing to work by making sure to apply proper alpha blending function.

The change can be seen, for example, how the mesh level is displayed in the viewer:
image

### DIFF
--- a/toonz/sources/common/tgl/tgl.cpp
+++ b/toonz/sources/common/tgl/tgl.cpp
@@ -66,7 +66,7 @@ double tglGetPixelSize2() {
   glMatrixMode(GL_MODELVIEW);
   glGetDoublev(GL_MODELVIEW_MATRIX, mat);
 
-  double det                      = fabs(mat[0] * mat[5] - mat[1] * mat[4]);
+  double det = fabs(mat[0] * mat[5] - mat[1] * mat[4]);
   if (det < TConsts::epsilon) det = TConsts::epsilon;
   return 1.0 / det;
 }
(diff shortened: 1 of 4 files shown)


RodneyBaker:
Thank you very much for the OpenGL fix @shun-iwasawa!

RodneyBaker:
@shun-iwasawa This is likely unrelated to any OpenGL issue or perhaps even this PR but...

I note that in testing this Plastic Tool after this PR is merged I get a crash if immediately after creating a plastic mesh I do not save and then click in the next cell down from the previously drawn in cell.   If I immediately save after creation of the plastic mesh... I can click into the next cell, draw something, create another mesh... all manner of things without a crash.

shun-iwasawa:
@RodneyBaker 
Thank you for testing. Unfortunately I couldn't reproduce the crash yet.
Sorry for the trouble but can you please post an issue with the detailed reproducible steps ? Thanks!

papio-pl:
Hi,

This change (I think) caused problems. Should be like this: 

image

And in this build preview shows something like this:

image

I use "External Palette" FX to remove some lines. I just made one color in my "Pallete" vector level fully transparent. It worked until now. 

Here's part of my FX table:

image
  
image
^this palette is replaced with this one:
image

shun-iwasawa:
@papio-pl 
I could reproduce the problem and will fix it right away. Thank you for the quick feedback!

PR #3788, opened 2021-03-19
Fix pasting raster to cell

### LINKED ISSUE #3571: undo fails on rasters levels
**Describe the bug**
If you cut/copy a part of canvas on a raster level or toonz raster level, and you past it directly on a empty cell and paste again for the image, undo will fail for the session.

**To Reproduce**
1.import or create a raster image by using new level command (both levels failed)
2.cut or copy a part of raster canvas
3.paste on empty cell to create a cell
4.paste again to insert the image
5.undo

shun-iwasawa:
This fixes #3571 

- Fixed behavior of pasting raster image to selected cell so that the image is pasted as soon as the new frame is created.
- Fixed undo of pasting operation to properly remove the cell. The Undo `PasteFullColorImageInCellsUndo` inherited `TToolUndo` , which means that you need to explicitly register the created cell to `TTool::m_cellsData` in order to make the undo to remove the cell properly.

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -639,6 +639,8 @@ bool pasteRasterImageInCellWithoutUndo(int row, int col,
       app->getCurrentLevel()->setLevel(sl);
       app->getCurrentLevel()->notifyLevelChange();
       sl->save();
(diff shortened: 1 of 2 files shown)


RodneyBaker:

PR #3789, opened 2021-03-19
feat: FxParticles preset simulating the Flock effect

andeon:
I saw that there wasn't a preset particle to simulate a large horizontal movement effect, so I'm creating this smooth flock effect (birds, fish, bees) providing a direction to users.

flock_bird


shun-iwasawa:
How about setting the following parameters to create a sense of depth?

- `Birth Params > Speed > Linked to Scale`
- `Birth Params > Size, Mass and Orientation > Perspective Distribution`
- `Birth Param > Top Layer` to `Bigger`

The above settings will make the nearer particles to be bigger, faster and sparse.

Original Flock
flock_original

With the above options
flock_new

andeon:
Yeah, it just works better.
 
flock_bird_depth

I've already changed the parameters, thanks!

RodneyBaker:
Thanks @andeon 
I'll delay merging to allow additional testing/feedback but this looks very good as is.
This will be a welcome addition to fx particles.

shun-iwasawa:
Although this is somewhat a feature addition I think it's OK to merge since it will not change the software binary itself.

@RodneyBaker I'll merge this as I'm going to release V1.5RC soon.

Thank you @andeon . LGTM

RodneyBaker:
Thank you @shun-iwasawa!

RodneyBaker:
Flocked dusting without the overlaid birds:
Flock

andeon:
@RodneyBaker ParticleFx is a fantastic resource, there are several possible combinations of parameters just by changing small range values to create many variations for the same effect purpose or not.

bee

PR #3791, opened 2021-03-20
Refactor toggle palette toolbar

shun-iwasawa:
This PR refactors toggle palette toolbar feature, which was recently merged ( #3757 ) .
Please note that the toolbar position status are now saved in integer (`1` and `0`) instead of boolean texts ( `true` and `false` ) in order to keep consistency with other options. 

@konero Please check if the feature still works as expected. Thanks!


konero:
@shun-iwasawa Works perfectly, thank you for the refactor!

RodneyBaker:
Thank you for the refactor @shun-iwasawa (and the confirmation @konero).

PR #3792, opened 2021-03-20
Fix Cleanup to Set Image Data

### LINKED ISSUE #3634: "Savel level as" failed after a cleanup conversion
**Describe the bug**
After a raster level cleanup if using save levels as command, levels turn to red and save fails. It seems that affect only the rasters levels made in OT, from outside OT that seems to work as expected.

**To Reproduce**
1.from a raster level made inside OT
2.use cleanup command
3.use save levels as

**Expected behavior**
No save failed

**Screenshots**

**Device Information:**
- w10
- OpenToonz Version: [nov 15 2020]

shun-iwasawa:
This PR will (hopefully) fix #3634 .
I replaced the line invalidating frame in the result level after cleanup by the one updating the frame to the result image.

@openanim
Sorry for the trouble, but can you please test this fix? I'm a little afraid of unexpectedly enbugging with this..

### DIFF
--- a/toonz/sources/toonz/cleanuppopup.cpp
+++ b/toonz/sources/toonz/cleanuppopup.cpp
@@ -310,6 +310,9 @@ CleanupPopup::CleanupPopup()
   m_imgViewBox->setChecked(false);
   m_imageViewer->setVisible(false);
   m_imageViewer->resize(406, 306);
+  ImagePainter::VisualSettings settings;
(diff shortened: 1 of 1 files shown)


openanim:
On my side I confirm that the problem is fixed, thank you @shun-iwasawa !

RodneyBaker:
Thank you @shun-iwasawa for the fast turnaround fix.

PR #3800, opened 2021-03-22
"Normalize Intensity" Option for PN Perspective Iwa Fx 

shun-iwasawa:

This PR will add `Normalize Intensity` option to the `PN Perspective Iwa Fx` with `Fresnel reflectivity` mode.
When the camera tilts downward (by setting `Eye Level` to the higher position), the `Fresnel reflectivity`  image becomes dark.

This PR will normalize intensity based on the brightness range of the lower-center and the upper-right pixel in the camera frame.
Users can also specify `Normalize Margin` to expand the normalize range in consideration of the normal angles are diffused due to waves.

pnperspectiv_normalize


RodneyBaker:

RodneyBaker:
Thank you @shun-iwasawa for the new fx option.
Quite a difference it makes!

PR #3801, opened 2021-03-22
New Fx : Rainbow Fx Iwa

shun-iwasawa:

This PR will add a new fx; Rainbow Iwa for rendering rainbow.

Numerical computation of scattering Intensity from water droplet for each wavelengh of light and scattering angle is precomputed, done with [MiePlot](http://www.philiplaven.com/mieplot.htm) by Philip Laven.

rainbow1

### Usage
RainbowFx Iwa is in Light category.
Generated rainbow image is expected to be clipped (using `Matte In` fx ) with some reference image indicating the density of water droplet and then add-composed on the original background.

### Parameters
- **Center** specifies antisolar point.
- **Radius** specifies radius of the rainbow. Physically speaking, it should be defined according to the camera fov.
- **Intensity** specifies the light intensity.
- **Width Scale** specifies the width of the primary rainbow. It is physically incorrect but may be needed for production.
- **Inside Intensity** can darken the supernumerary rainbow inside the primary rainbow. It is physically incorrect but may be needed for production.
- **Secondary Rainbow Intesity** can darken/lighten the secondary rainbow. It is physically incorrect but may be needed for production.
- **Alpha Rendering** can add fake alpha channel so that the image can also be composited with any modes other than `Add` . It is physically incorrect but may be needed for production.

Note that `Center`, `Radius` and `Width Scale` can also be adjusted using the Fx gadgets on the viewer.


RodneyBaker:

RodneyBaker:
Thank you @shun-iwasawa for the new fx!

konero:
@RodneyBaker You might want to hold off merging features for a bit, I'm not sure if a patch release will be published since a few bugs were recently fixed, if OT does them.

RodneyBaker:
@konero In looking at the calendar and considering the time users have had to test and provide feedback to the v1.5 release I'd say we very likely should be looking at an interim release in June/July timeframe.  We are even now entering the window for such an interim release whether that be a v1.5.1 or a v1.6.   With the number of features available I would suggest an interim v1.6 be targeted for July release.

I mainly suggest this because of the move to custom Qt and the benefits therein as well as potential regressions etc. that need to be accounted for will take time to explore.

cc:  @shun-iwasawa

RodneyBaker:
Thanks for the refresh.  Conflicts resolved.
LGTM
Thank you @shun-iwasawa for the new effect!

shun-iwasawa:
@konero
Thank you for your suggestion. I understand your proposal to hold introducing new features before the patch release.
A rational way would be separating a branch for the patch release (v1.5.1) which will cherry-pick only the bug fixes from the master branch (toward v1.6). The reason for not doing so is ~~my laziness~~ to simplify the maintenance task. I also believe especially introducing new fxs pose a minimal risk to other features compared to other changes in the software.

PR #3807, opened 2021-03-23
Fix Crash on Launch with Shrinked ImageViewer

shun-iwasawa:
This PR will fix the following problem occurred in our studio:

To reproduce:
- Enable 3DLut color calibration feature
- Open Color Model window and resize it to the minimum like this:
   shrinked_cm
- Quit OpenToonz and restart.
- OT crashes without launching.

The problem was due to calling initialization of Lut Caliburator (which uses OpenGL related functions in it) before the OpenGL context is created.


RodneyBaker:

PR #3810, opened 2021-03-23
Fix color model pick mode

### LINKED ISSUE #3809: Color Model Fails to Switch Picking Mode
**Describe the bug**
Even if you are choosing `Areas` mode in the Fill tool, both the line & area styles are picked in the color model.

**To Reproduce**
1. Load some level and the color model for it.
2. Choose Fill tool, select `Areas` mode.
3. Try picking the lines. Line styles are unwantedly picked.

**Expected behavior**
The picker should pick with the same mode as the Fill tool.

**Device Information:**
 - OS: Win10
 - OpenToonz Version: V1.5RC

shun-iwasawa:
This fixes #3809 

I modified the style picker function in #3652 but forgot to modify Color Model accordingly. Sorry for the trouble!


RodneyBaker:

PR #3822, opened 2021-03-25
fix pressure value sketch.myb, optimized images and file sizes

andeon:
Small adjustments to the files.


andeon:
I added 4 new brushes, once this V1.5RC version of opentoonz was released, I created another separate commit, not sure if it is correct.

RodneyBaker:
Thanks @andeon for the pressure fix.
The additional brushes shouldn't be a problem.

Issue #3813, opened 2021-03-24
Arc & multiarc weird behavior with pen


openanim:
**Describe the bug**
It seems there is issues with pen using theses tools

**To Reproduce**
List steps to reproduce the behavior.
1.create a level
2.use arc

 or multi arc on canvas

with mouse thats seems works normal

**Screenshots**
![arc   multi arc issues](https://user-images.githubusercontent.com/29790512/112236842-abeb0080-8c41-11eb-87fe-b1b06804d909.gif)

W10 cintiq companion 2 OT 1.5 RC

shun-iwasawa:
Fixed by #3823

Issue #3812, opened 2021-03-23
Lazy Nezumi Pro - Cannot hook viewer or use pen tablet through LNP.


melieconiek:
Lazy Nezumi Pro no longer works with OpenToonz. There are two separate issues. 

As of #3713 lazy Nezumi no longer works properly with my pen tablet (Wacom Cintiq Pro 24) The mouse works fine. There first stroke here is with the pen, the second with the mouse. Pen lag was set high so it's easier to see:

![_demo](https://user-images.githubusercontent.com/42298081/112211363-30159780-8bf2-11eb-9fe9-61937e7e0d98.gif)

**To Reproduce**
1. Click in the viewer window to insure it has focus.
2. Trigger 'Hook Window' in Lazy Nezumi.
3. Attempt to draw in the viewer with the pen.

**Expected Behavior**
Lazy Nezumi should work with the pen, it does not.

---

As of #3730 Lazy Nezumi can no longer hook the viewer and instead hooks the entire window. Lazy Nezumi will not work if the entire window is hooked. 

**To Reproduce**
1. Click in the viewer window to insure it has focus.
2. Trigger 'Hook Window' in Lazy Nezumi.

**Expected Behavior**
Lazy Nezumi should hook just the viewer, it instead hooks the entire window.

melieconiek:
To be clear I was first able to replicate these issues in these nightly builds: https://github.com/opentoonz/opentoonz_nightlies/releases/tag/nightly_19bf10004f7036feeab1ad3bed2e0772564a57d9 and https://github.com/opentoonz/opentoonz_nightlies/releases/tag/nightly_3fb02e0b8a873fe5ff96b81c58eee77a100ea6c2. So if something else changed with these builds that could be the cause.

RodneyBaker:
@melieconiek 
Just in case this applies to your case...

From the Lazy Nezumi site there is this note:

> Important: if you're using Win10, you should use your mouse to hook windows. If you use your tablet pen, the window might not be hooked properly

melieconiek:
This is not the cause of these issues @RodneyBaker -- even though your suggestion didn't seem likely, I went back and checked.

melieconiek:
[...]

PR #3823, opened 2021-03-26
Add WinTab Support Option by using Customized Qt

### LINKED ISSUE #3814: Select multiples strokes issue vector level with pen tablet
**Describe the bug**
It seems there is an issue using the pen and select multiples strokes.  Most of the time it fails

**To Reproduce**
1.create a vector level
2. draw severals strokes
3.use selection tool and select  multiple strokes using shift key

**Screenshots**
select multiple strokes issue.

W 10 cintiq companion 2 , OT1.5 RC

### LINKED ISSUE #3817: Wacom Tablet Issues due to Qt 5.15 using Windows Ink
**Describe the bug**
Recent issues related to the tablet behavior (like #3812, #3813 and #3814) seems to be related.
Qt seems to have started to support Windows Ink [from 5.12](https://bugreports.qt.io/browse/QTBUG-60437).
Unlike WinTab API used in Qt 5.9 the tablet behaviors are different  and are (at least, from our viewpoint,) problematic:

- `mousePressEvent()` is not called when the stylus touches the tablet, but when it hover from the tablet.
- Every time you press modifier keys like Shift, Ctrl and Alt , a tooltip with the name of key appears next to the cursor. (annoying)

I'm planning to try some measures to resolve the problem.

1. The problem had already been [reported](https://bugreports.qt.io/browse/QTBUG-83218) and Qt [provided](https://codereview.qt-project.org/c/qt/qtbase/+/295623) a way to switch APIs between WinTab and Windows Ink from Qt v6.0. I'm try to see if I can build customized Qt 5.15 including this feature.
2. Try to add some event filter to correct the timing of event to behave the same as WinTab.
3. Or, go back to good-old Qt 5.9.x 

The change may need RC2 trial period for verification.
OT v1.5 release will be postponed until this problem will be resolved.

**Device Information:**
 - OS: Win10
 - OpenToonz Version: V1.5RC
 - Tablet : Wacom Intuos 5

shun-iwasawa:
This PR fixes #3817 and (hopefully) fixes #3814 , #3813 and #3812 .

This PR will add a new option for enabling WinTab API instead of Widows Ink API.
To achieve the fix, I built the [customized version of Qt 5.15.2](https://github.com/shun-iwasawa/qt5/releases/tag/v5.15.2_wintab) from the source introducing the WinTab feature to be officially introduced from 6.0.

When being built with the customized Qt and with checking a `WITH_WINTAB` CMake option, a new Preference option `Use Qt's Native Windows Ink Support` will be added in the "Touch / Tablet" category which is DISABLED (= use WinTab) by default.
Enabling the check box will switch back to the Windows Ink API. It is for maintenance purpose and currently not recommended for users.

This PR will modify and redistribute the LGPL libraries so I added a necessary information in the `stuff/doc/LICENSE` folder.

I put the trial version [here](https://github.com/shun-iwasawa/opentoonz/releases/tag/v1.5.0rc_wintab) . 

Please try it and check if it will fix the issues and if there are any critical problems. Thank you!

### DIFF
--- a/doc/how_to_build_win.md
+++ b/doc/how_to_build_win.md
@@ -1,40 +1,37 @@
 
 # Building on Windows
 
-This software can be built using Visual Studio 2015 or above and Qt 5.9 (later Qt versions still not working correctly.)
+This software can be built using Visual Studio 2019 and Qt 5.x
 
 ## Required Software
 
-### Visual Studio Express 2015 or higher for Windows Desktop
-- https://www.visualstudio.com/vs/older-downloads/
-- Make sure the target platform is "for Windows Desktop", not "for Windows".
-- Community and Professional versions of Visual Studio for Windows Desktop also work.
+### Visual Studio Community 2019
+- https://www.visualstudio.microsoft.com
 - During the installation, make sure to select all the Visual C++ packages.
 
 ### CMake
 - https://cmake.org/download/
-- This will be used to create the `MSVC 2015` project file.
+- This will be used to create the `MSVC 2019` project file.
 
 ### Qt
 - https://www.qt.io/download-open-source/
 - Qt is a cross-platform GUI framework.
-- Install Qt 5.9 or higher (64-bit version, tested up to 5.9.9) with the [Qt Online Installer for Windows](http://download.qt.io/official_releases/online_installers/qt-unified-windows-x86-online.exe).
+- Install Qt 5.9 or higher (64-bit version, tested up to 5.15.2) with the [Qt Online Installer for Windows](http://download.qt.io/official_releases/online_installers/qt-unified-windows-x86-online.exe).
+
+#### Customized Qt v5.15.2 with WinTab support
+- Qt have started to support Windows Ink from 5.12. Unlike WinTab API used in Qt 5.9 the tablet behaviors are different and are (at least, for OT) problematic.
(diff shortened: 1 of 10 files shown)


RodneyBaker:
I've started to test this but it would be good to get other testers feedback as well... especially those that have had issues with tablet support in the past.

melieconiek:
Thank you for taking a look at this. Lazy Nezumi is now working properly. 

There is an issue in this build where the pen will not draw if the window did not have focus when you started drawing. This is not happening in the currently release candidate. It does this on all level types, even when Lazy Nezumi is not enabled.

_The first stroke here is without window focus, the second one, after the window has been selected, works properly._

_demo

The window hooking issue with Lazy Nezumi is still present, but this wasn't meant to fix that.

openanim:
For my part, I don't have the issues mentioned before now. Thank you !

shun-iwasawa:
Thank you for testing the fix!

> There is an issue in this build where the pen will not draw if the window did not have focus when you started drawing. This is not happening in the currently release candidate. It does this on all level types, even when Lazy Nezumi is not enabled.

The problem seems to have existed from OT v1.4, when activating the `Enable Windows Ink Support* (EXPERIMENTAL)` option.
It can be separated from this PR and be tackled in the future.

I think I'll introduce this and gather further feedback in RC2. 


**Please note that the nightly build after introducing this PR won't include the feature for a while as I need to migrate Qt for Jen.kins build.**

PR #3828, opened 2021-03-27
Fix Geometric tool / Type tool first point offset

### LINKED ISSUE #3820: Geometry tool: First point offset, when drawing on an empty cell.
I've recently noticed, when trying to draw (any type of shape) on an empty cell/column, that there's and offset in the first clicked point of the shape. Further points are positioned in the expected place.
It seems to happen in any type of level.

PC: Windows 7, Geforce, drawing using mouse.

shun-iwasawa:
This PR fixes #3820 , also fixes the same issue on Type tool.

### DIFF
--- a/toonz/sources/tnztools/geometrictool.cpp
+++ b/toonz/sources/tnztools/geometrictool.cpp
@@ -1022,19 +1022,23 @@ public:
     }
   }
 
+  bool preLeftButtonDown() override {
+    if (getViewer() && getViewer()->getGuidedStrokePickerMode()) return false;
+
+    if (!getApplication()->getCurrentObject()->isSpline())
+      // NEEDS to be done even if(m_active), due
+      // to the HORRIBLE m_frameCreated / m_levelCreated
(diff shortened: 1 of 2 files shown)


RodneyBaker:
I might be misunderstanding the full intent of this PR but it seems to me the Multi-Arc mode still gets the first point offset.
Other modes tested appear to have proper first point offset with this PR.

andeon:
Tested on Windows 10, this PR fix type tool and all Shapes except **Arc** and **Multi-Arc** when starting a Default level Type (Toonz Raster and Raster) by drawing a shape. 

The vector level wasn't affected by this problem, Arc and Multi-Arc working there.

shun-iwasawa:
Thank you @RodneyBaker and @andeon for testing the fix.
I fixed the offset of start position of the Arc and the Multi Arc primitives.
Also fixed that the undo fails to remove the created frame when undoing the Polyline and the Multi Arc drawing.

RodneyBaker:
LGTM

PR #3834, opened 2021-03-29
Expanded toolbar by default

### LINKED ISSUE #3804: Using V1.5: Can the toolbar expander be expanded by default
I thought that there was a fix to have all of the toolbar buttons shown by default. I see lots of comments/confusion from users who can't find a tool, simply because they have to click the expander. The expander button looks like just another button, so it's not obvious that you can expand it. It would save a lot of support time if all of the tools were shown from a fresh install.

Is there any chance that this could this be changed before 1.5 is released?

shun-iwasawa:
This PR is a quick patch to resolve #3804 .
Now the toolbar will be expanded on the first launch of the software.

### DIFF
--- a/toonz/sources/toonz/toolbar.cpp
+++ b/toonz/sources/toonz/toolbar.cpp
@@ -29,7 +29,7 @@
 #include <QVBoxLayout>
 #include <QObject>
 
-TEnv::IntVar ShowAllToolsToggle("ShowAllToolsToggle", 0);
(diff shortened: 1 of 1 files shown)


RodneyBaker:
LGTM
Just waiting for all the checks to complete...

DarrenTAnims:
Thank you so much @shun-iwasawa. This, as someone said in the original post, is only a small change, but will really help lots of new users. Thanks again.
Darren

Issue #3835, opened 2021-03-29
Reset to default room doesn't work


F-Burning:
I was trying to reorganize a room, but the interface panels froze and now can't move them (the room is unlock)

https://user-images.githubusercontent.com/74098734/112886308-a8d09400-90c9-11eb-8682-a48b4ed1a808.mp4

The thing is, when I try to reset to default, nothing happens.

DarrenTAnims:
A full screen video could help more. I was hoping you'd show how you tried to reset your rooms. If you used the menu command, you have to restart OpenToonz for it to take effect.

F-Burning:
> A full screen video could help more. I was hoping you'd show how you tried to reset your rooms. If you used the menu command, you have to restart OpenToonz for it to take effect.

I restart too, but this error was still there.

Yes, I use the window menu, and then Workspace: reset to default rooms.

shun-iwasawa:
Please let me confirm which is actually the problem:

1. (As shown in the video) You tried to drag the title bar of the Style Editor panel to pop it out as a floating window, but you couldn't, even if you are unlocking the workspace.
2. (As written in the comment) You used `Reset to Default Rooms` command and restarted OT, but the room layout were not reset.

Regarding the former problem;
- Please make sure again, the option `menu bar > Windows > Workspace > Lock Room Panes` is unchecked. It is NOT the lock icon in the top-right corner of the main window.
- I wonder what are the blank areas on the top and left side of the Style Editor. It would be helpful if you could upload the files under `OpenToonz stuff\profiles\layouts\personal\Default.username` .

Regarding the latter problem;
- Open the folder, `OpenToonz stuff\profiles\layouts\personal` and make sure the folder `Default.username` is removed once you use the `Reset to Default Rooms` command (it is what the command actually do).

konero:
@shun-iwasawa To reproduce:

1. Use the default room (Drawing)
2. Drag out the viewer (it will leave behind an empty void)
3. Try resize the panel next to the style editor, it will cause a lock up and then an eventual crash, you also can no longer undock other panels (if you close the viewer, it seems to cause a faster crash when you try to resize)

F-Burning:
@konero was faster than me 🏃

RodneyBaker:
@F-Burning This should be working.
Can you confirm you are still having a problem resetting to default rooms?

RodneyBaker:
@F-Burning Can you confirm this is still an issue with v1.5
[...]

PR #3836, opened 2021-03-30
Fix Undocking Window

shun-iwasawa:
This PR fixes an issue reported in [the comment](https://github.com/opentoonz/opentoonz/issues/3835#issuecomment-809782822) in #3835 by @konero .

Added Tool Bar to the items to be with fixed width when another item is undocked from the main window.

Hope this fix will resolve the issue #3835 itself as well.


RodneyBaker:

F-Burning:
Thank you so much @shun-iwasawa.

PR #3841, opened 2021-04-01
Improve Tablet Response (for Windows version only)

### LINKED ISSUE #3573: Broken lines and Rough drawings
Hello,

_**Describe the bug**_
I am using Tahoma2D for traditional animation. So the first thing I do is do a quick sketch to have my rough animation. I create a Smart Raster Level to draw then I start to make circles and Tahoma2D break all my curves. Circles look more like hexagons or polygons than fluid circles. I have tested on other software and this problem is not present so it is not my cintiq or my stylus.

_**To Reproduce**_
1. Create a Smart raster level
2. Use the brush tool
3. Make a circle

_**Expected behavior**_
I would like the lines that I draw to remain faithful to the gesture. It is very annoying that the lines are broken and angular. If I draw gently the angles no longer appear but it is not logical and natural to draw this way when it comes to sketches and rough animation.

_**Screenshots**_
2020-10-29 14_34_43-untitled6_  Mentoranim  _ Tahoma 1 0 1

_**Device Information:**_
 - OS: Windows 10 64 bits
- RAM : 16 Go
 - Tahoma2D Version: Tahoma 1.0.1 Built : Sep 3 2020
 - GPU: AMD Ryzen 5 2400G with Radeon Vega Graphics 3,60 GHz
 - GPU Driver Version: I don't know.

_**Others informations**_

I work on a Cintiq 22HD with an other screen (Benq 21'')

Thank you.

shun-iwasawa:
This PR will hopefully resolve #3573 .

This PR uses the `Qt::AA_CompressTabletEvents` application attribute for compressing frequent tablet event calls instead of using our own "delay timer".
In my environment, it seems to improve the tablet response, especially when drawing on Toonz Vector Levels.

Please note that currently the modification is only applied to Windows version since I was not sure if the `Qt::AA_CompressTabletEvents` attribute works on macOS.

### DIFF
--- a/toonz/sources/toonz/main.cpp
+++ b/toonz/sources/toonz/main.cpp
@@ -399,6 +399,12 @@ int main(int argc, char *argv[]) {
 
   // Enable to render smooth icons on high dpi monitors
   a.setAttribute(Qt::AA_UseHighDpiPixmaps);
+#if defined(_WIN32) && QT_VERSION >= QT_VERSION_CHECK(5, 10, 0)
+  // Compress tablet events with application attributes instead of implementing
+  // the delay-timer by ourselves
+  a.setAttribute(Qt::AA_CompressHighFrequencyEvents);
+  a.setAttribute(Qt::AA_CompressTabletEvents);
+#endif
 
   // Set the app's locale for numeric stuff to standard C. This is important for
   // atof() and similar
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -357,6 +357,13 @@ void SceneViewer::tabletEvent(QTabletEvent *e) {
     }
 #endif
     QPointF curPos = e->posF() * getDevPixRatio();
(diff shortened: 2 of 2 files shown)


artisteacher:
I'm happy to test a Mac version, if you do one. I can test with a Wacom Intuos tablet and with a Sidecar/ipad setup.

gelatinbomb:
It works. Good job!
2021-04-01 23_58_36-Window

RodneyBaker:
Thank you @shun-iwasawa!
And thank you @gelatinbomb for the feedback,

shun-iwasawa:
@artisteacher 
I tried applying the fix to all OS. Here is an macOS artifact for trial:
https://github.com/shun-iwasawa/opentoonz/actions/runs/711216625

However, according to [the documentation](https://doc.qt.io/qt-6/qt.html), it says;
```
Qt::AA_CompressTabletEvents
... Currently supported on the X11 windowing system, Windows 8 and above. ...
```
So unfortunately the option seems not to be supported.

artisteacher:
I'm not noticing any obvious differences.

Issue #3837, opened 2021-03-30
Menu icons color issue in macOS light and dark modes


artisteacher:
**Describe the bug**
Icons are too light for the native menu on macOS Mojave. I haven't had a chance to check out other versions of macOS yet.

**Screenshots**
<img width="212" alt="Screen Shot 2021-03-30 at 12 12 46 PM" src="https://user-images.githubusercontent.com/21159116/113021694-d7924d00-9151-11eb-9238-9d09d2776963.png"><img width="289" alt="Screen Shot 2021-03-30 at 12 12 22 PM" src="https://user-images.githubusercontent.com/21159116/113021709-da8d3d80-9151-11eb-93d1-97715940c6fb.png">

**Device Information:**
 - OS: [macOS Mojave]
 - OpenToonz Version: [Nightly (2021-03-30)]

konero:
I think this might be why I disabled them before, a number of things to consider:
- **[USER]** Switch to dark mode on macOS if these are important
- **[USER]** Switch to light theme in OT with dark icons
- **[DEV]** Disable these showing in menus for macOS only as quickfix (maybe a pref option)

I know it's not ideal, as for an actual solution I think maybe one way would be to separate the main menu and command bar so they create unique icons each if they don't already, create a way to detect macOS dark mode, then where the main menu is created and calls `createQIcon()` pass a QColor as an argument into the `recolorPixmap()` calls, and pick black or white depending if macOS has dark mode active.

That's above my skill level though! Not even sure if detecting macOS dark mode is possible or not. 😟

artisteacher:
You are right that this is a light/dark mode issue. The white menu icons are too light in light mode. The black icons aren't ideal in dark mode.
![Screen Shot 2021-03-30 at 2 11 43 PM](https://user-images.githubusercontent.com/21159116/113036187-f7317180-9161-11eb-8e40-778a8ecb9104.png)

I don't have strong feelings about the icons being visible or not in the menu for macOS. They are a little busy, yet helpful. 

I do feel strongly that the user shouldn't have to change the system theme to give the menu icons better contrast. Users also shouldn't have to experiment with theme/icon combinations to find one that displays properly for the light or dark mode.

artisteacher:
Now that the icons are optional in the menu, this is no longer an issue.

PR #3842, opened 2021-04-01
Make Menu Icons Optional

shun-iwasawa:
This PR adds a new Preferences option, `Show Icons In Menu` which enables users to choose whether to display icons in menus.

In order to hide menu bar icons, 16pixel-sized icons will be replaced by transparent pixmap for Windows and `setIconVisibleInMenu(false)` will be called for other platforms.

To quick-patch #3837 the default value of this option is OFF (= hide icons) for macOS.
(I tried to get information from OT whether the macOS is in dark mode, but in vain.)

### DIFF
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -40,6 +40,7 @@ enum PreferencesItemId {
   interfaceFontStyle,
   colorCalibrationEnabled,
   colorCalibrationLutPaths,
+  showIconsInMenu,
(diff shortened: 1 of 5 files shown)


RodneyBaker:
Thank you @shun-iwasawa for making the menu icons optional!

artisteacher:
Even if you could identify dark mode, the menu icons would need to be able to styled based upon the mode while the command bar icons would be based on the theme - or whatever is set in preferences. Hiding the icons is probably much simpler.

PR #3845, opened 2021-04-02
Added BSD compile instructions

jointri:
Added BSD compiling instructions because it should be made. Let me know how the instructions can be improved. I decided that it was only necessary to point out the initial dependencies and then have the user continue with the Linux guide. Let me know if there are any problems.


RodneyBaker:
Will merge as this is readme/build documentation.
Request BSD users let us know what (if anything) needs to be further revised in the instructions.
Thanks everyone!

Issue #3832, opened 2021-03-28
Rotate geometric shapes


F-Burning:
Something like this:

https://user-images.githubusercontent.com/74098734/112756503-63cf3380-8fdd-11eb-8f37-4d68b28b7740.mp4

artisteacher:
It would definitely be helpful to be able to rotate after creating without needing the Selection tool.

gab3d:
...with the possibility of pressing Shift to do a stepped rotation (45° perhaps?)

F-Burning:
> ...with the possibility of pressing Shift to do a stepped rotation (45° perhaps?)

and hold space to move the entire shape like in photoshop while you keep the drag.

I put it on top!

pojienie:
current progress: https://www.youtube.com/watch?v=_U1fv0_dCWo

I implemented the first two stuff. However, the 3rd one was unexpectedly hard, lol. Mostly due to the functionality needs to be activated by pressing space. And I can't access space key status in mouseMove event...

RodneyBaker:
Nice work @pojienie 
Impressive as always. :)

manongjohn:
Probably should avoid using space since that is the default shortcut for panning the canvas which a number of users use.

F-Burning:
Amazing!!

Grateful with this implementation ❤️

What about to use control or shift to move the shape? Is this possible?

manongjohn:
SHIFT, ALT and CTRL have special functions depending in the geometric shape being used.

gab3d:
@pojienie I think any free key would be useful for the movement part.
thank you for implementing it so fast! 🌈

pojienie:
PR created!

move was done when holding ctrl: https://www.youtube.com/watch?v=jaWoUmNsrRU

@manongjohn yeah, all those 3 keys are used by individual shape. However, my implementation uses shift and ctrl only when the user is rotating and/or moving the shape. Also, if the user is doing so, I don't even call the functions of those shape. So, at least for now, this feature will live in peace with all shapes.

F-Burning:
Fantastic @pojienie 😍 thank you.

This now need to be reviewed by someone to be implemented in Opentoonz or something? 

I would love to test this function.

KaseyFarron:
Awsome!

RodneyBaker:
@F-Burning  There are automated builds/artifacts for Mac, Windows and Linux created for every Pull Request.
[...]

PR #3846, opened 2021-04-02
Geometric tool rotation/move

pojienie:
This implements #3832.

I put Do Not Merge in the title because it's almost 1.5 release

demo:
1. https://www.youtube.com/watch?v=_U1fv0_dCWo 
2. https://www.youtube.com/watch?v=jaWoUmNsrRU

How to use:
1. select "geometric tool"
2. check the "rotate" checkbox
3. draw anything
4. once you're done drawing, you can move your mouse around to rotate the shape.
hold shift to force the rotation angle to be multiple of 45
hold ctrl to move it
5. once you're done with all those rotation and move, press left mouse button.

concern:
This will make `class GeometricTool` itself a lot more complex. Before this PR, `class GeometricTool` is rather simple: it just passes down the key/mouse events to the current geometric shape. However, after this PR, it will depend on its own state(namely:  `m_isRotatingOrMoving`) to decide if it wants to pass them down or not.

### DIFF
--- a/toonz/sources/tnztools/geometrictool.cpp
+++ b/toonz/sources/tnztools/geometrictool.cpp
@@ -45,6 +45,7 @@ TEnv::DoubleVar GeometricRasterSize("InknpaintGeometricRasterSize", 1);
 TEnv::StringVar GeometricType("InknpaintGeometricType", "Rectangle");
 TEnv::IntVar GeometricEdgeCount("InknpaintGeometricEdgeCount", 3);
 TEnv::IntVar GeometricSelective("InknpaintGeometricSelective", 0);
+TEnv::IntVar GeometricRotate("InknpaintGeometricRotate", 0);
 TEnv::IntVar GeometricGroupIt("InknpaintGeometricGroupIt", 0);
 TEnv::IntVar GeometricAutofill("InknpaintGeometricAutofill", 0);
 TEnv::IntVar GeometricSmooth("InknpaintGeometricSmooth", 0);
@@ -376,6 +377,7 @@ public:
   TDoubleProperty m_hardness;
   TEnumProperty m_type;
   TIntProperty m_edgeCount;
+  TBoolProperty m_rotate;
   TBoolProperty m_autogroup;
   TBoolProperty m_autofill;
   TBoolProperty m_smooth;
@@ -404,6 +406,7 @@ public:
       , m_opacity("Opacity:", 0, 100, 100)
       , m_hardness("Hardness:", 0, 100, 100)
       , m_edgeCount("Polygon Sides:", 3, 15, 3)
+      , m_rotate("rotate", false)
       , m_autogroup("Auto Group", false)
       , m_autofill("Auto Fill", false)
       , m_smooth("Smooth", false)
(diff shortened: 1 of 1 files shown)


F-Burning:
Hope this feature will be in the future, it's very, very helpful :)

thank you @pojienie!

RodneyBaker:
Since we are waiting to merge this until after v1.5 release...
First let me say 'thank you' for adding this feature @pojienie 
You are a master at these small but significant improvements!

Something I note is that when rotating a circle that is drawn with this option... very obviously... the circle appears not to rotate because of course it is just a circle.  This makes me wonder if there isn't an opportunity here to allow for some other process to happen during the rotation process that would benefit the user.  And this very likely could be another PR/feature so please just consider this discussion and not an actual request and also just the note that rotating a circle doesn't particularly do anything.

In fact, rotating a circle does do something and in particular it rotates the control points around the center point.
One thought I had was that perhaps the control points could be visible during the rotation to allow the user to see where the CPs are and in particular to see where the origin and the end of the spline that comprises the circle join.

Why might this be of importance?
Continuity is always a concern when dealing with splines and I've found it useful to know where the starting and ending of a shape are especially for processes such as auto-inbetweening. 

Perhaps in this case viewing the control points is not ideal.
It might be that some other process might be invoked to allow offset rotation or manipulation of a circle specifically as rotating it doesn't do anything particularly useful.   This of course unlike other shapes where the results are obvious.

Just random thoughts thrown into the mix.
Keep up the excellent work!

openanim:
@pojienie, again thank you very much for this feature. 

While trying it, I had some display problems but that could be related to a bad understanding on my part. In doubt...you can see it in the video.

When I draw with a geometrical shape no matter what level, the beginnings of the geometrical shape is not visible. It seems that the color blends with the background.  But if you overlay a shape on top of it, you can see that it is there.

Same thing with the level strip happens except for the vector level where I can clearly see the shape on the level strip.


https://user-images.githubusercontent.com/29790512/113536878-70ecb380-95d7-11eb-9bf1-71feecdad173.mp4

pojienie:
@RodneyBaker I sure never thought about the need to rotate a circle and displaying control points will be useful in this case. I tried copy-paste Control Point Editor Tool's `draw()` code and found out that it also draws speed-in and speed-out points. Which means I'll have to either copy-paste another part of code or put the function in to a common part of the codebase in order to filter those points out... So yeah, this is definitely in another PR. (Unsure if I'll do it, though, as I currently see no need for it despite the use case you just described)

@openanim it's because I didn't specify the color when drawing the preview line. Thus, in your case, it draws the preview line but it is white so you can't see it. I've created another commit that will use the exact same color as the color palette you're currently using.

openanim:
Thanks for these quick additions and explanations, everything is displayed now !

F-Burning:
> Puse Do Not Merge en el título porque es casi la versión 1.5.

Does this means that it will not merge with the official software?

Thank you for the new addition!

RodneyBaker:
@F-Burning   No, it just will not be merged before the initial release of v1.5 which could be as soon as tomorrow.
Feature freeze for the initial release of v1.5 was (technically) back in December.
This feature and more will be added to v1.5 but after the initial release.
Look for it in the nightly releases.  :)

RodneyBaker:
@pojienie I have removed the pre-release 'Do not merge' notice in the title.
This just to bring the PR back into the post release queue.

pojienie:
@RodneyBaker Gotcha, thanks! Didn't realize 1.5 is already now :)

RodneyBaker:
Time to move this feature enhancement forward.

PR #3850, opened 2021-04-03
Fix Hidden Icons

shun-iwasawa:

- When you deactivate `Show Icons In Menu` Preferences option, the "squared pixel" button in the camera settings fails to show "=" icon.
image


RodneyBaker:

Issue #3858, opened 2021-04-04
Creating SystemVar.ini


jointri:
Is opentoonz at the point that the "Creating SystemVar.ini" step in the GNU/Linux install steps no longer needed? Does the code discover it itself now?

pojienie:
I... think so? I seriously don't remember if I had done that step at all, not even in 1.4 era.

I just tested it by removing $HOME/.config/Opentoonz and copied `stuff` directory from the codebase to $HOME/.config/Opentoonz. After launching Opentoonz, I saw that SystemVar.ini was automatically generated.

My Linux distro is Artix Linux and codebase is latest master (27bd0ac41a10f5d57a4639c43cddd3bb06c38a54)

jointri:
The pull request has been merged, the step is no longer needed, I'm going to close this issue now.

callybug:
Why was this closed? I have to manually create the SystemVar.ini file after building OT from source in Linux. By the time I install OT again on a new machine I forget I need to do this, so it's a bit of a nuisance.

PR #3861, opened 2021-04-04
Update how_to_build_linux.md

jointri:
As someone else confirmed [that doing the SystemVar.ini step is no longer necessary](https://github.com/opentoonz/opentoonz/issues/3858) and it does it automatically now. It can be removed from the documentation. If anyone else wants to test it to double check, go ahead.


RodneyBaker:
Thanks @jointri for asking the question and following up!

PR #3878, opened 2021-04-09
[Test needed] Attempt to Fix Tool Option Slider

### LINKED ISSUE #3851: Vector level doesn't work with number pad
When I change the thickness with the mouse it works, but when I use the keyboard, the line width is the same below one.

https://user-images.githubusercontent.com/74098734/113490720-e8312300-94c3-11eb-8c63-bc27afa41955.mp4

shun-iwasawa:
This PR hopefully fixes #3851 .
- Made `decimals`  can be specified in `DoubleLineEdit` ctor in order to prevent the slider range to overflow integer limit in `DoubleValueField::setRange()` .
- Replace `pow(double, int)` to `std::pow(double, double)` just to be safe.

@F-Burning can you please try the GitHub Actions artifact once it is built and see if the problem is resolved?

### DIFF
--- a/toonz/sources/include/toonzqt/doublefield.h
+++ b/toonz/sources/include/toonzqt/doublefield.h
@@ -189,7 +189,7 @@ class DVAPI DoubleLineEdit final : public DoubleValueLineEdit {
   QDoubleValidator *m_validator;
 
 public:
-  DoubleLineEdit(QWidget *parent = 0, double value = 1);
+  DoubleLineEdit(QWidget *parent = 0, double value = 1, int decimals = 8);
   ~DoubleLineEdit() {}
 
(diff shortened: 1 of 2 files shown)


F-Burning:
Sure @shun-iwasawa 

Thank you so much! 👍

F-Burning:
@shun-iwasawa 

Still not working

The bar moves if I type commas instead of periods, but even if it moves, the line goes from 1 to 0, it does not recognize intermediate thicknesses, such as 0.5.

Once I go down the bar below 1 with the keyboard and I try to go up, the thickness of the line is 0, until I finally type the 1 key

This however works perfectly for me in Tahoma. Maybe you could check that code if you are interested.

PR #3882, opened 2021-04-09
Update toonz.ts

pafri:
Czech translation update. Greetings


shun-iwasawa:
Thank you @pafri for updating Czech translation!

PR #3895, opened 2021-04-13
Fix Brush Smoothing

### LINKED ISSUE #3893: Very slow working brush
**I cant draw anything because my brush is working very slowly and buggy**
When i try to draw my brash start to work unnormal (sometimes OpenToonz even crash). I saw this in last 1.5 r.c and 1.5 versions. Before that this problem never seen by me.

**To Reproduce**
1. Create new Toonz raster level
2.start to draw

**Screenshots**
I create a video to show it ----> https://youtu.be/J10cBu8fAj8

**Device Information:**
 - OS: [ Windows 7]
 - OpenToonz Version: [V 1.5]
 
 My tablet - GAomon M106K.

### LINKED ISSUE #3894: Smoothing stops working with stylus
When drawing with a stylus, smoothing only seems to work on the first stroke.  Each following stroke has little to no smoothing applied.  Smoothing may get applied to the first stroke of each new drawing/frame.  Turning Windows Ink on or off doesn't seem to have any effect on this behavior.  Smoothing with mouse seems normal.

**To Reproduce**
1. Turn smooth up to max
2. Draw multiple strokes with tablet pen

**Expected behavior**
Smoothing should be applied at the set level to every stroke.

**Screenshots**

https://user-images.githubusercontent.com/60277069/114323347-9ac34000-9ad9-11eb-8144-0efc6072837a.mp4

**Device Information:**
OpenToonz Version: OpenToonz V1.5
Operating System: Windows 10 Home 64-bit
CPU: Intel Core i7-4710HQ
Memory: 16 GB
Graphics Card: Nvidia GeForce GTX 970M
Graphics Tablet: Wacom Cintiq Pro 24

shun-iwasawa:
This PR fixes #3893 , and fixes #3894 .

Before the fix, the smoothing feature had smoothened all trajectory points in the stroke every time the cursor moves. Computing the points which are already plotted on the canvas is unnecessary and taking extra time.

Such excessive time had not been critical in the previous version, since the tablet movement was not caputured so frequent. However in the current version it causes the brush lag and stutters.

So, the modification in this PR are as follows:

-  Made smoothing computation to be done for only points where the output stroke is not appeared in the canvas.
- Made subsampling amount between trajectory points to be adaptive according to the distance of the points. Before the fix the amount had been fixed to 8.

### DIFF
--- a/toonz/sources/tnztools/toonzrasterbrushtool.cpp
+++ b/toonz/sources/tnztools/toonzrasterbrushtool.cpp
@@ -655,7 +655,8 @@ static void CatmullRomInterpolate(const TThickPoint &P0, const TThickPoint &P1,
 
 //--------------------------------------------------------------------------------------------------
 
-static void Smooth(std::vector<TThickPoint> &points, int radius) {
+static void Smooth(std::vector<TThickPoint> &points, const int radius,
+                   const int readIndex, const int level) {
   int n = (int)points.size();
   if (radius < 1 || n < 3) {
     return;
@@ -665,7 +666,10 @@ static void Smooth(std::vector<TThickPoint> &points, int radius) {
 
   float d = 1.0f / (radius * 2 + 1);
 
-  for (int i = 1; i < n - 1; ++i) {
+  int endSamples = 10;
+  int startId    = std::max(readIndex - endSamples * 3 - radius * level, 1);
+
+  for (int i = startId; i < n - 1; ++i) {
     int lower = i - radius;
     int upper = i + radius;
 
@@ -692,21 +696,23 @@ static void Smooth(std::vector<TThickPoint> &points, int radius) {
     result.push_back(total);
   }
 
-  for (int i = 1; i < n - 1; ++i) {
-    points[i].x     = result[i - 1].x;
-    points[i].y     = result[i - 1].y;
-    points[i].thick = result[i - 1].thick;
+  auto result_itr = result.begin();
+  for (int i = startId; i < n - 1; ++i, ++result_itr) {
+    points[i].x     = (*result_itr).x;
+    points[i].y     = (*result_itr).y;
+    points[i].thick = (*result_itr).thick;
   }
 
   if (points.size() >= 3) {
     std::vector<TThickPoint> pts;
-    CatmullRomInterpolate(points[0], points[0], points[1], points[2], 10, pts);
+    CatmullRomInterpolate(points[0], points[0], points[1], points[2],
+                          endSamples, pts);
     std::vector<TThickPoint>::iterator it = points.begin() + 1;
     points.insert(it, pts.begin(), pts.end());
 
     pts.clear();
     CatmullRomInterpolate(points[n - 3], points[n - 2], points[n - 1],
-                          points[n - 1], 10, pts);
+                          points[n - 1], endSamples, pts);
     it = points.begin();
     it += n - 1;
     points.insert(it, pts.begin(), pts.end());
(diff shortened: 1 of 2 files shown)


RodneyBaker:
@shun-iwasawa  I'm curious if this can be rolled into the v1.5 release as a patch rather than have it available via nightly release only.  As users encounter the v1.5 release it won't be good if they don't have this fix included.

RattusaurusRex:
Smoothing seems to work again, and there is no brush lag/stuttering.  Thank you for the fix!

RodneyBaker:
In testing, the only thing I cannot quite account for is something that I can only seem to reproduce by drawing C shapes (or reverse C shapes quickly).  I'm investigating further to see if recent tweaks to my Wacom Intuous might be responsible.

The problem can be seen in that the C shapes when drawn quickly close themselves with a line connecting the start and end points of the stroke.

image

This seems to occur more frequently near the center of the canvas but isn't confined to that location.

I can only get this to occur with toonz raster levels.

openanim:
Thanks for the fix!
Small aside, it seems that the current behavior is more pronounced than in previous versions. Especially in relation to the available value range. A very strong smooth is reached quite quickly compared to before, without really being close to the maximum value.

RattusaurusRex:
The smoothing is still inconsistent with Wintab, although the lag is greatly reduced on strokes that have smoothing turned on but not applied.  Strokes that appear to have smoothing applied are still lag-free.  This setting + turning on Windows Ink in tablet drivers seems to be the only way to get normal smoothing: 
image

RodneyBaker:
@shun-iwasawa My apology for delay in feedback. I seem to be having some trouble with my pen tablet that is unrelated to recent changes in Opentoonz and that makes it hard to test. 

 Some things appear to be better with current PR and some worse but on my system it seems that aspects of that relate to my pen settings even outside of Opentoonz.  Of late, I have found that restarting my computer has helped in that for awhile I had lost all pressure sensitivity but it came back after a reboot.  Between recent updates to Windows, Wacom tablet driver and Opentoonz it is harder determine what issues relate to the lastest Opentoonz updates.    

At any rate, I will try to separate the three primary elements (of operating system, pen/tablet and OT) and test each in better isolation to provide more useful feedback.

RattusaurusRex:
Couple of videos to show the difference for me between using Wintab and Windows Ink with this PR:


https://user-images.githubusercontent.com/60277069/115128737-ca1ff400-9f94-11eb-949c-d0ca0d17eaff.mp4

Smoothing works on the second stroke, but not the third using Wintab.  The third stroke starts to lag after a while, although the interval is much longer than prior to this PR.


https://user-images.githubusercontent.com/60277069/115128707-67c6f380-9f94-11eb-96d8-12572fa49b01.mp4

Smoothing seems to work consistently using Qt's Windows Ink Support.

RodneyBaker:
@RattusaurusRex  If you use the regular Windows Ink setting in OT preferences rather than the new maintenance option you are getting the same thing?

RodneyBaker:
As a follow up...

It seems this setting of Windows Ink relates to why my quickly drawn C strokes where closing:

RodneyBaker:
I have been staring at this PR with the intent of merging but want to make sure others have adequate time to test and provide feedback.   The primary reasons for wanting to merge now are:

- Performance is improved over previous releases especially with regard to Wintab operations
- Merger will allow for broader testing to assist with further identification of remaining issues 

It will be good to also see what (if any) differences there are (concerning this PR) relative to github action builds versus the nightly build.

RodneyBaker:
Will merge to get this out into the world for broader testing and feedback.

Aside:  Currently, without this PR, I am experiencing the best brush control that I recall ever experiencing in Opentoonz.
This was after experiencing considerable difficulties.  It is not entirely clear what changed in order for this to happen.

PR #3897, opened 2021-04-14
Fix PSD Loading

### LINKED ISSUE #3896: PSD files cannot be loaded on saved scene
**Describe the bug**
PSD file containing multiple layers cannot be loaded with `LoadAs:Columns` option on the saved scene.

**To Reproduce**
1. Launch OT and save the empty scene.
2. Prepare PSD file with multiple layers.
3. Try loading the PSD file, with `LoadAs:Columns` option.

It fails with `PSD code name error` .

**Expected behavior**
Load PSD files properly.

**Device Information:**
 - OS: Win10
 - OpenToonz Version: latest

shun-iwasawa:
This PR fixes #3896 

The problem was due to [the changes](https://github.com/opentoonz/opentoonz/pull/3758/files#diff-0ee8edc63b37effb1882f7e0ee2497d82c3ee78a7579664cbc7f08c26bcf4369L348-R352) in #3758 , the scene's version number started to be kept in the scene after loading.

When loading PSD files, some convertion of file path replacing layer ids to layer names is needed ONLY when loading the level. OTOH when loading PSD levels while loading the scene, the convertion is not needed as the file path is already converted.
OT had distinguished whether the `TXshSimpleLevel::load()` is called from `Load Level` command or from `Load Scene` procedure by scene's version number. So, after #3758 the necessary conversion was skipped and caused the error.

To resolve the problem I introduced `isLoading` flag to `ToonzScene` and evaluate it instead of the version number.

### DIFF
--- a/toonz/sources/include/toonz/toonzscene.h
+++ b/toonz/sources/include/toonz/toonzscene.h
@@ -259,6 +259,9 @@ If \b scene is in +scenes/name.tnz return name,
   // if the path is codable with $scenefolder alias, replace it and return true
   bool codeFilePathWithSceneFolder(TFilePath &path) const;
 
+  bool isLoading() { return m_isLoading; }
+  void setIsLoading(bool isLoading) { m_isLoading = isLoading; }
+
 private:
   TFilePath m_scenePath;  //!< Full path to the scene file (.tnz).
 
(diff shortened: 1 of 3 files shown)


RodneyBaker:
LGTM

PR #3898, opened 2021-04-14
Fix Missing Checkboxes in Menu Item

shun-iwasawa:

- When the preferences option `Show Icons in Menu`  is disabled, toggle commands like `Preview Cleanup` , `Transparency Check` , etc. do not show checkboxes. It is hard for users to know the commands' state.


RodneyBaker:

PR #3910, opened 2021-04-19
Fix Memo Popup Background Color

shun-iwasawa:
This PR fixes the memo popup's background color to be the same as the selected color.

This problem is probably a regression with migration of Qt.


RodneyBaker:

PR #3914, opened 2021-04-20
Particles Fx : Use 16bpc Control Image for Gradient

shun-iwasawa:
This PR will enhance the Particles Fx.

Unlike other parameters, `Environment > Gravity` and `Birth Params > Speed > Speed Angle` (with `Use Gradient Angle` option) use gradient (i.e. brightness difference between neighbor pixels) of the control image.
Since the gradient is computed by 2D finite difference method, the control image should be with sufficiently large color depth in order to obtain proper gradient. However, currently the Particle Fx computes the control image in 8bpc regardless of the output settings.

This PR make the Particles Fx to compute control images only for such parameters in 16bpc to obtain better result.

Here is an example. I use the following radial gradient image for `Speed Angle` parameter:
particles_ref

The center of the gradient is at the same position as the center of the Raylit Fx.

Result before this PR:
particles_old

Result of this PR:
particles_new

Note that in the result before this PR some particles move wrong direction since they fail to compute the proper gradient from the control image with small color depth.

FYI the `Tiled Particles Iwa` is already using 16bpc control image for all parameters. 💪

### DIFF
--- a/toonz/sources/stdfx/particles.cpp
+++ b/toonz/sources/stdfx/particles.cpp
@@ -150,12 +150,14 @@ Particle::Particle(int g_lifetime, int seed, std::map<int, TTile *> porttiles,
 
   /*- Speed Angleの制御。RangeモードとGradientモードがある -*/
   if (values.speeda_ctrl_val &&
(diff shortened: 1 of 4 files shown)


RodneyBaker:

PR #3917, opened 2021-04-21
Fix Raster Deformation Slowness

### LINKED ISSUE #3902: Lag / slow editing with selection tool
**Describe the bug**
It seems that some selections with the selection tool make OT lag on the raster levels. This seems to happen a lot when selecting the entire drawing. 

**To Reproduce**
1.create a  raster / toonz raster level
2. select the entire drawing
3.try to edit with selection tool

**Screenshots**
slow editing with some selections

W 10 Cintiq companion 2 1.5 official

shun-iwasawa:
This PR fixes #3902 .
I added following 2 modifications to enhance the response when deforming raster image with the selection tool:

- **Enhancement for using Windows + Tablet case** : For now the tablet event is called too frequent for relatively time consuming operation like raster deformation. So I re-introduced the "delay timer" to Windows so that the deformation will be calculated at 20msec interval (50fps) in maximum. Note that the delay is currently applied only to the raster deformation. Other operations like drawing with the brush tool still have no interval.
- **Enhancement for all cases** : Deforming with bilinear resampling is bit slower than with nearest-neighbor resampling. So I made the operation to use nearest-neighbor resampling when the mouse cursor moves in large distance (< 3pixels per frame). When the cursor moves slowly, it calculates with bilinear resampling in order to show an accurate result.

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -97,6 +97,7 @@ public:
   QPointF m_mousePos;  // mouse position obtained with QMouseEvent::pos() or
                        // QTabletEvent::pos()
   bool m_isTablet;
+  bool m_isHighFrequent;
 
 public:
   TMouseEvent()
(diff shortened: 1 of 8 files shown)


RodneyBaker:
That appears to be working nicely.
Well done.

I will delay merging slightly to allow those that were having the problem to test and provide feedback.
but... LGTM!

RodneyBaker:
Moving forward with this.

Issue #3918, opened 2021-04-21
BUGS when copy/pasting with keys


Al532:
Here's a short video showing several bugs when copy/pasting with keys (drag+ctrl).
https://youtu.be/Vabx3TzBP0Y

0:06 it doesn't paste at all (until I change the selected cell)
0:11 it doesn't paste the first key (which stays at frame 119)
(0:18 my mistake, I probably forgot to press ctrl)
0:28 same as 0:06
0:32 after pasting there's a key at frame 133 that shouldn't be there
0:34 after undoing, there's a key at 118 that shouldn't be there

OpenToonz 1.5 stable

shun-iwasawa:
> 0:06 it doesn't paste at all (until I change the selected cell)

I think I could reproduce this problem. Thanks!

Tartarin8:
Concerning the other bugs, it may be when trying to copy the first non-empty frames of the column, and paste before, within the empty cells.

RodneyBaker:
Closing as likely resolved via #3922
@Al532 Please let us know if that is not the case.

PR #3922, opened 2021-04-22
Fix Cell Selection on Click

shun-iwasawa:
This PR fixes the first problem of #3918 . In particular;

1. Select cells together with keyframes by Ctrl + dragging.
1. Copy them.
1. Click (without Ctrl) at a cell for selecting the paste destination. 
1. Try to paste. Paste does not work.

The cause of the problem is that the clicked cell fails to be selected at the step 3.
This was because the switching of selection types (from Cells+Keys selection to Cells only selection) is done AFTER selecting the clicked cell and it clears the selection.

I fixed the order of the procedures to switch the selection types BEFORE selecting the clicked cell.

### DIFF
--- a/toonz/sources/toonz/xsheetdragtool.cpp
+++ b/toonz/sources/toonz/xsheetdragtool.cpp
@@ -128,6 +128,12 @@ public:
     int col          = pos.layer();
     m_firstCol       = col;
     m_firstRow       = row;
+    // First, check switching of the selection types. This may clear the
+    // previous selection.
+    if (m_modifier & Qt::ControlModifier)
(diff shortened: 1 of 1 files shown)


RodneyBaker:

RodneyBaker:
Something of a post mortem note here...

I'm not sure if it relates to this PR or not but I note that when using the nightly built with this PR I seem to get an interference with shortcut keys.   I specifically try to use the Type Tool and then say typing the letter L locks the current column and then after unlocking I can't seem to get tools to work right.

Will test more to see if it actually relates to this PR or is something else.

PR #3927, opened 2021-04-23
Fix tool options' shortcuts to work without opening tool options bar

### LINKED ISSUE #3912: "Toggle Maximize panel" bug
**Describe the bug**

You can't change tools, colors, etc, with "Toggle Maximize panel" activated.

**To Reproduce**

1. Go to Windows menu.
2. Select "Toggle Maximize panel"
3. Try to change tools or colors with shortcuts.

**Expected behavior**

Work normally.

**Device Information:**

 - OS: Windows 10
 - OpenToonz Version Nightly 2021-04-19 
 - GPU: NVIDIA GeForce GT 730

shun-iwasawa:
This PR fixes #3912 .

Before this PR, shortcuts for changing tool options were made to activate the associated controls in the tool options bar. 
The major disadvantage of such approach was that the shortcut does not work when the tool option bar is hidden, as seen in the case of #3912 .

To resolve the problem I introduced in this PR a new class `ToolOptionsShortcutInvoker` to receive the shortcuts and change the tool properties on behalf of the tool options bar.

Cursory illustration:
tosi
 
- All implementations of the "combination" shortcuts - for switching tools and modes at the same time - are also moved from `MainWindow` to `ToolOptionsShortcutInvoker`.
- As an **disadvantage** of this change, the preference option `Preferenes > Tools > Dropdown Shortcuts` will be removed. Using the shortcut will always cycle through the available options.


RodneyBaker:
For the first time I got a virus notification on an artifact... with the Github Actions artifact for this PR.
Interestingly enough I did not get a notification when I tried the Appveyor artifact.

I'm currently looking into why I might have received the notification at all.

RodneyBaker:
Pressing on...

LGTM (and defaulting to cycling through the tool options seems to be the ideal,,, it will be good to know if anyone has use cases where activating the dropdown is preferred)

gab3d:
Default cycling sounds good to me too.

PR #3930, opened 2021-04-26
Fix Shortcut Input

### LINKED ISSUE #3926: Keyboard Shortcuts: Shifted number shortcuts inaccessible.
While testing issue #3923 I tried to create temporary shortcuts for interpolations and found out that while trying to create shifted number shortcuts (i.e. Shift+1, Ctrl+Shift+2, etc.) the editor showed them as: Shift+! and Ctrl+Shift+@

So, when trying to use those shortcuts they do not work, presumably because you can't double shift a key (i.e. Shift+Shift+1, Ctrl+Shift+Shift+2, etc.)

shun-iwasawa:
This PR fixes #3926 .

The input field of the Shortcut popup now inherits `QKeySequenceEdit` .

- You need to wait one second after inputting the key to finish editing.
- Qt accepts up to 4 sequential key inputs (e.g. Pressing `Ctrl+A` then `Ctrl+B` to trigger some command) . However currently OT supports only one key input. Multiple key input will be automatically truncated.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -1669,6 +1669,17 @@ QDialog #dialogButtonFrame QPushButton:focus:pressed {
   width: 16;
   margin-right: -1;
 }
+#ShortcutViewer {
+  font-size: 17px;
+}
+#ShortcutViewer QLineEdit {
+  color: black;
+  border-color: #b8bc7f;
+  background-color: white;
+}
+#ShortcutViewer QLineEdit:focus {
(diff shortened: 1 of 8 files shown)


RodneyBaker:
My apology for the delay.
LGTM

Aside:  I'm not a huge user of shortcut keys but look forward to the day when OT will accept the additionally complex shortcut key combinations allowed by Qt.

artisteacher:
Just an FYI, this causes a crash on macOS when trying to open the shortcut window. There's a fix in Tahoma.

PR #3944, opened 2021-05-10
Xsheet Zoom Control

shun-iwasawa:
This PR will enable the zoom control in the Xsheet view as well as in the Timeline view.

xsheet_zoom

- The slider is at the rightmost of the column headers.
- The range of scale is from 50% to 100% . The default value is 100% which is the same as the conventional height (20 pixels / cell) .
- Right clicking the slider will show a menu to quickly adjust the zoom scale fitting visible frame range to nice round number (1, 1.5, 2, 3, 4 or 6 seconds per page. Only available options will be displayed.)
- Added a button to the spread sheet view of the Function Editor for toggling synchronization the zoom scale with Xsheet.


gab3d:
@shun-iwasawa 
great! 😊
this will make Xsheet so much useful!

won't "Expand Function Editor Header to Match Xsheet Toolbar Height" option be turned on by default make sense, in light if this synchronization feature?

shun-iwasawa:
@gab3d 
To make the Function Editor to be fully synchronized with Xsheet, you need to prepare 2 things:

- Enable the `Expand Function Editor Header to Match Xsheet Toolbar Height` option.
- Get the top edges of the panels lined up.

Currently in the `Default` room theme the Function Editor panels are not lined up with the Xsheet panels. So I think only altering the default value of the `Expand Function Editor Header to Match Xsheet Toolbar Height` option may not effective.

beeheemooth:
This is great news! 
But judging by the fact that the width of the columns has remained unchanged, now we will more often scroll the sheet to the right-left than up and down :). Are you planning on making a preference for column widths (at least entering a numeric value in Preferences)? Look at Toonz Premium or Retas, the columns are narrower and the user has more columns in the field of view. 
Screen_0217

RodneyBaker:
Echoing the thanks for this improvement!
Like @beeheemooth I'm still looking forward to when users can adjust the column width of the xsheet.   This to allow for fuller viewing of notelevel cell content, wide headers, etc.  In my estimation even if it initially is just a setting in Preferences that applies a mullultiplier to the default/hardcoded value that would be sufficient.

In many user cases I'll guess they'd like the columns to be narrower but from my POV users need to be able to make them larger to allow for textual entries for storyboarding, notetaking and long titles/descriptions.  When using notelevels as the source with Text FX Iwa for instance it can be impossible to read what is entered into the cells of the column.  

If given the multiplier I might suggest the hard coded default value be reduced to 50 (for easy math purposes) and the default multiplier set to 1.5 (to bring it back to the current default... which I think is 70 or 75).   There should be a lower limit for the width although I forget what that was but it was somewhere around 5 to 10 pixels... or just wide enough for the current icons.  I forgot what I tested as the largest width but that can be fairly wide (perhaps constrained by the width of the xsheet size).  I can see why no one has made this change to column width... there is quite a lot to consider in allowing users to change that value.  

At any rate, thank you for this addition!
I'll finish my round of testing with it and press forward with merging.

shun-iwasawa:
@gab3d 
I came up with an idea that we can remove the `Expand Function Editor Header to Match Xsheet Toolbar Height` preference option and integrate its feature with "Synchronization the zoom scale with Xsheet" button, since they are supposed to be used at the same time.

RodneyBaker:
Thank you @shun-iwasawa for the useful addition!

PR #3945, opened 2021-05-10
Camera capture : Connect to any camera on start

shun-iwasawa:
This PR modifies the initial behavior of the Camera Capture feature.

Before this PR, if OT could not find the previously connected camera, it does not activate any camera even if there is an another available one.
This PR modifies such behavior. OT will try to connect any available camera if the previously connected one is not found in order to cut out the need of selecting the camera in many cases.


RodneyBaker:
Thank you @shun-iwasawa for the camera capture improvement.

PR #3947, opened 2021-05-12
Fix mesh file naming

### LINKED ISSUE #3946: OT deletes mesh file when using default names for multiple sub-xsheets
When using meshes on multiple sub-xsheet levels with default names (sub, sub_1, sub_2, etc), the first mesh level file is deleted on disk.  When reloading the scene, the first mesh level will be not be found.  

**To Reproduce**
1. Create a level
2. Collapse the level into a sub-xsheet
3. Create a mesh on the sub-xsheet
4. Save All (sub.0001.mesh is created on disk)
5. Follow steps 1-3 to create a sub-xsheet and mesh with a second level
6. Save All (sub.0001.mesh is deleted and sub_0001.mesh is created on disk)
7. Reload the scene (mesh level for the first sub-xsheet is missing)

**Expected behavior**
Mesh level files should not be deleted when creating mesh levels on separate sub-xsheets with default names.

**Screenshots**
After creating the first mesh:
image
After creating the second mesh:
image
Creating subsequent meshes causes the "level already exists" dialog to pop up:
image
Before reloading the scene:
image
After reloading the scene:
image

**Device Information:**
OpenToonz Version: OpenToonz Nightly built 2021-05-10 23:30:41 UTC
Operating System: Windows 10 Home 64-bit
CPU: Intel Core i7-4710HQ
Memory: 16 GB
Graphics Card: Nvidia GeForce GTX 970M
Graphics Tablet: Wacom Cintiq Pro 24

shun-iwasawa:
This PR fixes #3946 by teporarily disable underscore for the file path in order to get proper mesh file path for the level name with underscore like "sub_1" .

### DIFF
--- a/toonz/sources/toonz/meshifypopup.cpp
+++ b/toonz/sources/toonz/meshifypopup.cpp
@@ -248,9 +248,13 @@ TXshSimpleLevel *createMeshLevel(TXshLevel *texturesLevel) {
       codedOrigPath = TFilePath(
           "+drawings/a");  // parent directory. Store them in "+drawings".
 
+    // temporarily disable underscore in order to get proper file path when the
+    // level name includes underscore like "sub_1"
+    TFilePath::setUnderscoreFormatAllowed(false);
     codedOrigPath = codedDstPath =
         codedOrigPath.withName(pathName).withType("mesh").withFrame(
             TFrameId::EMPTY_FRAME);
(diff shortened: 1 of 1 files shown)


RodneyBaker:

RodneyBaker:
As a post mortem follow up on this PR...

There are reports of this fix not making it into the nightly build (or subsequent build).
I'm not sure why but looking into it.

PR #3969, opened 2021-06-08
Fix and refactor Line Blur Fx Ino

shun-iwasawa:
This PR will fix and refactor the `Line Blur Ino Fx`:

- Fixed the problem that the result becomes incorrect when applying this fx just before `Bloom Iwa Fx` . The Line Blur Ino Fx had assumed that the `wrap` of the raster image is the same as the image `width` , but it is not always so.

  

- Removed the redundant part of the source in `ino_line_blur.cpp` which was identical to `igs_line_blur.cpp` .


RodneyBaker:
Thanks also for the Gitthub Actions update as well!

PR #3974, opened 2021-06-10
Modify sync scroll between Xsheet and Function editor panels

shun-iwasawa:
This PR modifies the synchronization between Xsheet and Function editor panels, especially when they are placed next to each other with the same height.

Before this PR scrolling the one panel may cause flaw in synchronization with the other, i.e. scroll positions of the panels sometimes unwantedly offset by several pixels. The offset seemed to be occured due to the error when scaling the scroll value to standard size and back to the scaled size again.

This PR will resolve such problem. 
If the sender and the receiver scroll bars have the same maximum size, OT will assume that they are in the neighbor panels with the same height & scale. In such case just set the same value as the sender to the receiver without zoom adjusting.

### DIFF
--- a/toonz/sources/include/toonzqt/spreadsheetviewer.h
+++ b/toonz/sources/include/toonzqt/spreadsheetviewer.h
@@ -70,8 +70,9 @@ private:
   void connectScrollbars();
   void disconnectScrollbars();
 
-  void handleScroll(QPoint &offset);
+  void handleScroll(QPoint &offset, int senderMaximum, int senderValue);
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Nice.
LGTM

RodneyBaker:
Posting a follow up image so that those that look into this PR can better know what to look for with regard to the sync of Xsheet and Function Editor.  

image

Note also the setting in Preferences that is required to be on in order for the panels to be aligned:

image

PR #3988, opened 2021-06-21
Compass Gadget for Radial and Spin Blur Fxs

shun-iwasawa:
This PR will add a new "Compass" gadget to `Radial Blur` , `Radial Blur Ino` , `GPU Radial Blur` , `Spin Blur` , `Spin Blur Ino` and `GPU Spin Blur` . The gadget is located at the center of the table and shows the direction to the center point of the fx.
It had been originally developed for the `Raylit` Fx but we found that it would be also useful for some other fxs in which the center point can be distant from the camera frame.

radial_spin_fx_gadgets

N.B. Currently the guide lines does not show proper direction of the blur if you set some value to the `Twist` parameter of `Radial Blur Ino` .

### DIFF
--- a/stuff/library/shaders/radialblurGPU.xml
+++ b/stuff/library/shaders/radialblurGPU.xml
@@ -85,3 +85,13 @@
     center
   </Parameter>
 </Concept>
+
+<Concept>
+  compass_ui
+  <Name>
+  Center
+  </Name>
+  <Parameter>
+    center
+  </Parameter>
+</Concept>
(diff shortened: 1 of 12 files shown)


DarrenTAnims:
This look really useful. When you say that the gadget is located at the centre of the table, is that the centre point that you set with the animate tool? Can you move that centre point? What is the effect centered on?

RodneyBaker:
@DarrenTAnims The center point is more of a pivot that can be changed/animated via the gizmo itself (and via expressions via Curve Editor).

image

RodneyBaker:
Thank you @shun-iwasawa for the additions!

PR #3991, opened 2021-06-24
New Fx: Bokeh Advanced Iwa (and Bokeh Fxs overhaul)

shun-iwasawa:
This PR will introduce a new Fx: `Bokeh Advanced Iwa`, which is just like a hybrid of `Bokeh Iwa` and `Bokeh Ref Iwa` .

This Fx can compose multiple layers like the `Bokeh Iwa` Fx. Unlike the `Bokeh Iwa` Fx, this fx can use depth reference images for dividing layer into several sub-layers within specified range of depth. Also, this fx can apply individual hardness value for each source for adjusting bokeh brightness.

bokeh_advanced_result

In the above example, the background image is divided into sub-layers using the depth reference image (rendered with the Corridor Gradient Fx) and composited with the characters at different depths.

### Input Ports
- **Iris** : Connects the image to define the shape of the Iris. The image luminance values will be used for the effect. If no image is connected to the Iris port, no calculation will take place. It's possible to connect 8bit or 16bit RGBA images.
- **Source [1～5]** : Connects images for the layers. The order in which they are connected here doesn't define the layer stacking order. If nothing is connected to any Source port, no calculation will be performed.
- **Depth [1～]** : Connect the Depth reference image for the layer to be divided into sub-layers. The brightness of each pixel corresponds to the depth. The higher (brighter) the value, the farther away from the camera.

#### Common
- **On-Focus Distance** : If there's a layer or sub-layer in this position, it will look on focus and it will be composed normally. 0 represents the camera position. (Range 0.0-1.0)
- **Bokeh Amount** : Maximum size of blur (in scene units). When the focus position and the layer position are 1.0 away from each other and Bokeh Adjustment is 1, the Iris image is enlarged until the width of this image reaches this value.
- **Master Hardness** : The gamma value of the film. Used to convert between RGB value and exposure. This is equivalent to the increase in RGB value (0.0 to 1.0) when the exposure is increased 10 times. The lower the value, the more the highlights are emphasized. (Range 0.05-3.0)
- **Hardness per Source** : Specify whether to use individual hardness value instead of the master hardness for converting the layer RGB values to exposures.
#### Sources
- **Distance** : The distance of the layer from the camera. The layer stacking order is automatically sorted according to this value. (Range 0.0-1.0)
- **Bokeh Adjustment** : Bokeh size correction value. The size of the bokeh is multiplied by N, keeping the order of layer stacking. If this value is 0, the layers will be composited normally without blurring regardless of the distance of the layer. (Range 0.0-2.0)
- **Hardness** : Individual hardness used for converting this layer's RGB values to exposure.
- **Depth Image** : Specifies the Depth port number. If some available Depth port is set, this layer will be divided into sub-layers at different depths in the specified range.
- **Depth Range** : Specifies the range of depth where the sub-layers will be distributed.

### Notes
- Separating a layer into sub-layers is just like the Bokeh Ref Iwa Fx except that the following parameters are hard-coded:
  - Distance Precision = 10
  - Fill Gap = ON
  - Use Median Filter = OFF

## Other Changes
- The Bokeh Fxs are now sharing the same base class and functions. You can say that the `Bokeh Iwa` and `Bokeh Ref Iwa` are simpler variations of `Bokeh Advanced Iwa` with limited parameters.
- The `Bokeh Iwa` Fx is modified as follows:
  - Modified a composition method when layer is at the focus distance. Before this PR the layer is just over-composited in RGB values. Now the RGB values will be converted to exposure value just like other layers at off-focus distance.
  - Removed the `Layer* Premultiply` options which is for specifying whether the input image is premultiplied or not. This is for sake of simplicity, and because most of other Fxs in OT assume that input images are always premultiplied. If you would like to input input which is not premultiplied, apply `Level Settings > Premultiply` option or the `Premultiply` fx beforehand.
  - Altered the parameter names `Layer* Distance` and `Layer* Bokeh Adjustment` to `Source* Distance` and `Source* Bokeh Adjustment` respectively in order to gain consistency with the port name. 
    **ATTENTION** : This change will lose expression reference to the parameters with conventional names. If you have expression references to these parameters in the scene, please replace `layer` by `source` to recover the reference. (e.g. `fx.bokehiwa01.layer1distance` > `fx.bokehiwa01.source1distance`)
- Enhanced resizing behavior of Fx Settings when switching the current fx to be more precise.


RodneyBaker:
This is a very deep and complex FX and I won't pretend to be able to test it all out.
Instead we'll merge and get this into the hands of users that can explore and run through what is available.

Thanks you @shun-iwasawa not just for the new FX, and overhaul but the very detailed explanation of options and usage.

DarrenTAnims:
This looks like an excellent addition to the fx library.
Q. Would it be possible to work out the depth using the Z value of each column instead of having to specify it with depth images?

RodneyBaker:
@shun-iwasawa while investigating the use of expressions as a solution to @DarrenTAnims query about using the Z value of each column to drive depth in the Bokeh Advanced Iwa FX I am having problems entering any expression using the SO channel (i.e. col1.so).  The expression does work with the z channel so perhaps that can at least partially address Darren's thoughts.   I will continue to investigate use of the so channels with expressions but it appears that may not be working properly with expressions.

Also, via discord there was a query about having a default shape automatically set for use with the iris, for instance a circle.  I thought I would pass the idea along here via commentary.

PR #3993, opened 2021-06-25
Bokeh Advanced Fx Quick Fix

shun-iwasawa:
This PR is a quick patch for fixing a following problem in the Bokeh Advanced Iwa Fx:

- When a layer is separated into sub-layers using a depth reference image and the individual hardness is different from the master hardness, semi-transparent of blurred contour becomes incorrect.


RodneyBaker:
Thank you @shun-iwasawa for the quick fix!

PR #4003, opened 2021-07-09
Fix crash on clicking shortcut tree

shun-iwasawa:
This PR fixes the following problem in the Shortcut popup:

- When clicking the category name in the shortcut tree, OT crashes.

This PR also enables to expand the category by single-clicking of the category name.


RodneyBaker:
Thanks @shun-iwasawa!
I confirm the crash and that this PR fixes the issue.
Also, thank you for the UI enhancement.
LGTM

PR #4004, opened 2021-07-09
Fix Paste Color command not to set the "edited flag" to normal styles

shun-iwasawa:
This PR fixes a problem reproducible as follows:

1. Copy a style in a studio palette and paste it in a level palette. (The pasted style will become "linked" style.)
2. Modify color of the linked style. (The "edited" flag is set and a diagonal line appears.)
3. Copy the edited linked style.
4. Select another normal style and Paste Color.
5. Not only color, but also the edited flag is pasted. It is unwanted behavior.


RodneyBaker:
That you @shun-iwasawa for the fix!

PR #4025, opened 2021-07-28
Fx schematic node placement fix

shun-iwasawa:
This PR modifies a placement logic of fx schematic nodes as follows:
- Modified the fx node location when right-click the fx node and use "Add fx" command. Before the fix, the fx was placed unexpected position and sometimes it was too far from the parent fx. With this PR the fx will be placed more consistently.
    place_fx_fix

-  When using "Create Linked Fx" command, the duplicated fx nodes will be placed lower-right positions from the original ones. The command will no longer displace the original fx nodes' position.


RodneyBaker:
Thank you @shun-iwasawa for the FX Schematic node placement fix!

PR #4021, opened 2021-07-26
Fix text color style inside some windows

andeon:
The change fix the text color inside the windows:

Canvas Size | Export Level | Convert to Vectors
---               | ---                | ---          
(Current Size, New Size and Anchor) | (Vectors Export Box and Vectors Thickness) | (Full color non-AA images)

 ## Preview - Linux
<p align="center">  fix style Linux</p>

## Preview - Windows 
fix_style Windows 10

I found this problem when using window dark decoration with dark themes and light decoration with light themes. I tested it on Linux and Windows.

PR #4030, opened 2021-08-03
Fix text color style inside some windows by andeon

shun-iwasawa:
This PR is based on #4021 , just rebasing to the current master in order to resolve the error in github actions.


shun-iwasawa:
Thank you @andeon . LGTM

PR #4032, opened 2021-08-03
Change behavior of the Layer Blending Ino fxs when the background layer does not exist

shun-iwasawa:
This PR modifies the behavior of all fxs in `Layer Blending Ino` category, when the background layer does not exist.

Here is an example:
blending_ino_fix

Before this PR, in such case OT renders the foreground layer in full opacity regardless of the blending mode or the opacity value.
With this PR, OT will blend the foreground layer over full-transparent background, with specified blending mode and the opacity value in order to make the fx more consistent through frames.


RodneyBaker:
Thank you for the change in FX @shun-iwasawa 
LGTM

PR #4036, opened 2021-08-06
Redesign the Pass Through Fx Node

shun-iwasawa:
This PR modifies an appearance of the Pass Through Fx Node like as follows:

The arrow icons of input / output ports are hidden when they are connected:

Additionally, this PR modifies the highlighting behavior when clicking the fx node.
When the clicked node is connected to the pass through node, now all the link lines from such node will also be highlighted like as follows:


RodneyBaker:
Very nice redesign @shun-iwasawa 
Looks very good to me.

gab3d:
I like it very much! 🤩  
thank you!

WolfInABowl:
I love this redesign; it's so much more compact and distinct from other nodes and makes the graph so much easier to parse visually.

One thing I would like to mention is that there does not appear to be a way to rename PassThroughs now. As Macros use the names of nodes for their inputs, it is now not possible to give Macro inputs more purposeful names.

For example, this node setup with a renamed PassThrough
image
becomes this Macro:
image

Without renaming the PassThroughs, you get this Macro with far less useful input labels:
image

KaseyFarron:
I like the redesign but I agree we need an option to rename the node, it's so useful for macros! 🙏

DarrenTAnims:
I use the pass-through node only to create nicely-named inputs for macros, so the renaming is an essential element of this node. Could renaming be brought back @shun-iwasawa ?

konero:
Just chiming in to say I also think naming ability is useful for Macros.

PR #4037, opened 2021-08-06
Enable Inputting Frame with Suffix in Camera Capture

shun-iwasawa:
This PR enables to input frame number with suffix (such as `0001a`) in Camera Capture popup, which is already available in most of other part of the software. 
Please note that the Stop Motion feature (i.e. camera capture without opening the popup) is not yet modified. 

cameracapture_framesuffix


RodneyBaker:
Thank you @shun-iwasawa for the update to Camera Capture.
LGTM

PR #4042, opened 2021-08-16
Fix margin bottom slider-v-handle

andeon:
Added @slider-v-handle-margin to pass the value as @slider-handle-margin in controls.less

fix_slider

Fixed margin bottom of slider-v-handle.svg to Neutral and Light theme, respectively.


RodneyBaker:
Simple and effective.
LGTM

PR #4046, opened 2021-08-19
Backlog of PR #4037, fix for 32bit build

shun-iwasawa:
This PR will introduce the same modification as #4037 to `penciltestpopup_qt.h /.cpp` .
The fix will not directly affect the software since currently the modified files are used only for 32bit build.


RodneyBaker:
Thanks @shun-iwasawa for the 32bit fix!

PR #4048, opened 2021-08-24
Fix Bokeh Fxs Artifacts

shun-iwasawa:
This PR fixes artifacts appear in rendered result of Bokeh fxs (Bokeh Iwa, Bokeh Ref Iwa and Bokeh Advanced Iwa). 
As shown in  the following  image,  when the `Hardness` parameter is relatively small (like 0.2),  bright dots appears at the edge of the image.  The dot is more easy-to-find at the layer at the focus distance.
The artifacts appear only when rendering in 8bits per channel.

bokeh_artifacts

The problem seems to be due to calculation accuracy. In the fx computation, the layer images are un-premultiplied (i.e. RGB values are divided by Alpha value). Calculation error becomes noticeable when rendered in 8bpc, especially at semi-transparent part with small alpha channel value. Small hardness value makes the error more prominent.
In order to fix the problem, this PR will make the fx to compute layer images always in 16bpc, regardless of the preview / render settings.

### DIFF
--- a/toonz/sources/stdfx/iwa_bokeh_advancedfx.cpp
+++ b/toonz/sources/stdfx/iwa_bokeh_advancedfx.cpp
@@ -232,11 +232,12 @@ void Iwa_BokehAdvancedFx::doCompute(TTile& tile, double frame,
 
   // compute the input tiles
   QMap<int, TTile*> sourceTiles;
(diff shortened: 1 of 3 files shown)


RodneyBaker:
LGTM

Issue #3776, opened 2021-03-17
Additions to External Palette


WolfInABowl:
Though this is all currently achievable with an FX Macro, I believe it would be more convenient to include these features in External Palette.

1. Matte input
This would allow for limiting the influence based on a separate image, with the ability to choose the channel.
This could be used to create more artist-controlled lighting, e.g. by combining a bespoke 'shadowed' or 'lit' palette and using a separate level as a matte.

2. Mix amount
This would allow for a smooth blend between the Source's palette and the provided palette.

3. Limit to specified styles
This would work in the same manner as effects such as Texture where the user can limit the influence of the effect to the specified styles.

PR #4055, opened 2021-09-06
Add Fx Global Controls

shun-iwasawa:
This PR will add a new parameter "Fx intensity" to several fxs for controlling amount of intensity.

Some fxs (listed below) now inherits `GlobalControllableFx` class which has the "Fx intensity" parameter.
When rendering, the Cross Dissolve Fx will be inserted just after the fx and the "Fx intensity" parameter will work as mixture amount between the result of the fx and its input. The Cross Dissolve Fx is created internally just for rendering computation and is invisible from users.

The parameter will be added to the following fxs:

**Image_Adjust**
- Adjust Levels
- Brightness Contrast
- Channel Mixer
- Curves
- Despeckle
- HLS Add Ino
- HLS Adjust Ino
- HSV Add Ino
- HSV Adjust Ino
- HSV Scale Ino
- Level Auto Ino
- Level Master Ino
- Level RGBA Ino
- Multi Tone
- Negate Ino
- RGBA Cut
- RGBA Scale

**Matte**
- HSV Key
- RGB Key

**Noise**
- HLS Noise Ino
- HSV Noise Ino

**Toonz_Level**
- External Palette ( This resolves the second request in #3776 )
- Palette Filter

### DIFF
--- a/stuff/profiles/layouts/fxs/STD_externalPaletteFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_externalPaletteFx.xml
@@ -1,0 +1,4 @@
+<fxlayout>
+  <page name="External Palette">
+  </page>
+</fxlayout>
--- a/toonz/sources/include/tfxattributes.h
+++ b/toonz/sources/include/tfxattributes.h
@@ -36,6 +36,9 @@ class DVAPI TFxAttributes {
   // to maintain backward compatibility in the fx
   int m_fxVersion;
 
+  bool m_hasGlobalControl;
+  double m_globalIntensity;
+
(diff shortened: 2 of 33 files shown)


RodneyBaker:
The (new) Mac Mojave automated build appears to be taking an excessive time to build and as the additional WIndows build does allow opening of the Details link.  Perhaps that relates to a localized build on your end @shun-iwasawa?

At any rate, this PR looks good from this end and very useful.

RodneyBaker:
And now, even the Mac Mojave checks finished.
LGTM

PR #4059, opened 2021-09-10
Fix Fx Gadget Offset

### LINKED ISSUE #4058: Fx Gadget Offsets According To The Current Column
**Describe the bug**
Fx Gadgets of non-zerary fx offset according to the current column's transformation.
If the current column is not the "parent" column of the current fx, the gadgets may appears at wrong position.

**To Reproduce**
1. Build a scene like as follows:
    fxgadget1
2. Translate the column 1's position.
3. In the Fx Schematic, select the Spin Blur Ino Fx. The current column will be switched to the column 2 as it is the "parent" of the fx. (Correct behavior)
    fxgadget2
4. In the Xsheet, select the column 1.

The fx gadget moves to the column 1 position. (Wrong behavior)
fxgadget3

**Expected behavior**
It should be stayed at the proper position regardless of the current column.

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR will fix #4058 , setting the transformation of proper column to the fx gadgets.

### DIFF
--- a/toonz/sources/tnztools/edittoolgadgets.cpp
+++ b/toonz/sources/tnztools/edittoolgadgets.cpp
@@ -2084,6 +2084,15 @@ EditToolGadgets::DragTool *FxGadgetController::createDragTool(int gadgetId) {
 //---------------------------------------------------------------------------
 
 TAffine FxGadgetController::getMatrix() {
+  TFx *fx = m_fxHandle ? m_fxHandle->getFx() : 0;
+  if (fx) {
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I stared at this one for quite awhile trying to think of a case where this change would not be optimal.
It seems that even if there was such a case there are a number of workarounds that can be used to adjust the offset to another desired location.

Having the FX Gadget offset inherit the location of the parent does appear to be the correct implementation.   Thank you @shun-iwasawa for the fix!

PR #4066, opened 2021-09-13
Fix channel clamping behavior in Layer Blending Fxs 

### LINKED ISSUE #4065: "Linear Color Space" option of the Layer Blending Fxs clamps blue channel
**Describe the bug**
When the "Linear Color Space" option in the Linear Blending Ino Fxs is activated, blue channel of the rendered image is clamped. This makes the white pixels to be yellow-ish.

**To Reproduce**
1. Create a white Color Card Fx and connect both ports of the Over Ino Fx.
    clamp_bug1

2. Preview the scene. Compare the results with / without the "Linear Color Space" option. Although it just composites solid white images, the result with the "Linear Color Space" option becomes unexpectedly yellow-ish.
    clamp_bug2

**Expected behavior**
The white pixels should be white.

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #4065 .
Each Layer Blending Fx uses the same kernel function regardless off the state of "Linear Color Space" option.
In the end of the kernel function every channels are clamped between 0.0 - 1.0.
When the "Linear Color Space" is activated, RGB channel values are converted to XYZ color space which can have more than 1 channel value. Thus clamping the maximum value to 1.0 causes wrong result in bright pixels.

Modified the clamping to be done only when the kernel function is applied to RGB channel values.

### DIFF
--- a/toonz/sources/stdfx/igs_color_blend.cpp
+++ b/toonz/sources/stdfx/igs_color_blend.cpp
@@ -200,12 +200,11 @@ double overlay_ch_(const double dn, const double dn_a, const double up,
       dn, dn_a, up, up_a, up_opacity);
 }
 double soft_light_(const double dn, const double up) {
-  return (
-      (up < 0.5)
-          ? (dn + (dn - dn * dn) * (2 * up - 1))
-          : ((dn < 0.25)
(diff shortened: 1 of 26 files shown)


RodneyBaker:

PR #4068, opened 2021-09-14
Fix PN Clouds Ino Fx

### LINKED ISSUE #4067: PN Clounds Ino Fx malfunction
**Describe the bug**
PN Clouds Ino Fx malfunctions under some conditions.

**To Reproduce**
1. Create or load a Toonz Raster Level. Draw something on it. (In this example, an yellow circle is drawn.)
2. Build schematic like as follows.
    pnclouds1
3. Enable the "Auto Gain" option in Bloom Iwa Fx in order to clarify the result. Keep other parameters unchanged from the default.
4. Preview the scene. An upper  side of the yellow circle is cut off and the noise pattern is shrunk. White artifacts appear on the both side. (The artifacts may disappear after updating the preview)
    pnclouds2
5.  For comparison, this is a result of the same scene only replacing the PN Clouds Ino Fx by Fractal Noise Iwa Fx. (Correct result)
    pnclouds3

**Device Information:**
 - OS: Win10

shun-iwasawa:
This PR fixes #4067 . 
The raster image data to be computed in the fx may not tightly-packed. According to the Fx Schematic arrangement, there can be a "gap" between a pixel data at the end of scanline and the one at the beginning of next scanline.
PN Clouds Ino Fx assumed that the raster image data is always tightly packed (i.e. raster wrap = raster width) which was not always correct.

Fixed the fx to consider the raster wrap.

### DIFF
--- a/toonz/sources/stdfx/igs_perlin_noise.cpp
+++ b/toonz/sources/stdfx/igs_perlin_noise.cpp
@@ -9,7 +9,7 @@ double perlin_noise_3d_(const double x, const double y, const double z,
                         ,
                         const double persistence  // Not 0
                         // 1/4 or 1/2 or 1/sqrt(3) or 1/sqrt(2) or 1 or ...
-                        ) {
+) {
   double total = 0;
   Noise1234 pn;
   for (int ii = octaves_start; ii <= octaves_end; ++ii) {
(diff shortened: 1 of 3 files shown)


RodneyBaker:

PR #4072, opened 2021-09-22
fix a few typos

alexmyczko:


toonz/sources/translations/chinese/image.ts
```diff
@@ -101,7 +101,7 @@ Please try raising the FFmpeg timeout in Preferences.</source>
     </message>
     <message>
         <location filename="../../image/sgi/filesgi.cpp" line="1235"/>
-        <source>Endianess</source>
+        <source>Endianness</source>
```
shun-iwasawa, line 104: Thank you @alexmyczko for the typo fixes.
.ts files (the _translation source_ files) are automatically generated from the source.
Unless the typos are fixed in the original source, they will be overwritten by wrong words at some points in the future.

Can you please find the corresponding typos in the source and fix them as well?
For instance, this case of the word  `Endianess` , the original source is located in [the line 1235 of toonz/sources/image/sgi/filesgi.cpp](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/image/sgi/filesgi.cpp#L1235) as you can see in the `location` tag above this line.

alexmyczko: I tried, but I'm not sure if I edited too much or not, can you check? I wasn't able to find the other typos anywhere else though.

toonz/sources/image/sgi/filesgi.cpp
```diff
@@ -1136,7 +1136,7 @@ void SgiWriter::open(FILE *file, const TImageInfo &info) {
       (TBoolProperty *)(m_properties->getProperty("RLE-Compressed"));
   assert(bp);
   bool compressed = bp->getValue();
-  p               = (TEnumProperty *)(m_properties->getProperty("Endianess"));
+  p               = (TEnumProperty *)(m_properties->getProperty("Endianness"));
```
shun-iwasawa, line 1139: The term "Endianess" here is saved in the scene file and used as a key for loading the file format property.
It's fine to fix the typo, but I think we need to prepare some way to accept both the terms "Endianess" and "Endianness" in order to keep backward compatibility.
It will be quite cumbersome work and I think it is out of scope of this PR, so I recommend to keep this typo unfixed.

toonz/sources/image/sgi/filesgi.cpp
```diff
@@ -1210,7 +1210,7 @@ void SgiWriter::writeLine(short *buffer) {
 
 Tiio::SgiWriterProperties::SgiWriterProperties()
     : m_pixelSize("Bits Per Pixel")
-    , m_endianess("Endianess")
+    , m_endianess("Endianness")
```
shun-iwasawa, line 1213: This line is also used as the key term of file priperty. I think it should not be fixed, either.

toonz/sources/image/sgi/filesgi.cpp
```diff
@@ -1232,7 +1232,7 @@ void Tiio::SgiWriterProperties::updateTranslation() {
   m_pixelSize.setItemUIName(L"48 bits", tr("48 bits"));
   m_pixelSize.setItemUIName(L"64 bits", tr("64 bits"));
   m_pixelSize.setItemUIName(L"8 bits (Greyscale)", tr("8 bits (Greyscale)"));
-  m_endianess.setQStringName(tr("Endianess"));
+  m_endianess.setQStringName(tr("Endianness"));
```
shun-iwasawa, line 1235: This line is OK to be fixed.
The function `tr()` is for registering a word to the translation source file (.ts file). It is usually used for the terms in user interface like labels, command names and messages.
Fixing the typo of the words in the `tr()` function would be much appreciated as they are visible to users 👍

shun-iwasawa (changes requested):

luzpaz (commented):
Not worth merging this PR. it modifies upstream code, unnecessarily modifies translation files, and breaks the API without introducing backward compatible logic to remedy the issue. 

There is one valid typo. I'll create a new PR, add it and reference this issue.

PR #4073, opened 2021-09-24
Fix Export Xsheet PDF

shun-iwasawa:
This fixes an obvious bug in the `Export Xsheet PDF` feature as follows:

- OT fails to export PDF if the current scene is untitled. (Showing an error message "Please specify the file name.")


RodneyBaker:
I confirm the bug and that this PR resolves the issue.

PR #4075, opened 2021-09-27
Fix Fx placement

### LINKED ISSUE #4074: Fx placement offsets when the first input column contains empty cell
**Describe the bug**
Fx position is according to its "parent" column. i.e. Starting from the fx node, track back the first input of each nodes in the fx schematic tree and the firstly found column node is its "parent".
However, if the parent column has an empty cell at certain frame, the parentness is transferred to the second input.
If the first and the second input columns have different placement, the "child" fx unexpectedly offset at such frame.

**To Reproduce**
1. Load or create 2 levels, one is a background and the other is a character.
2. Expand the background cell to the second frame.
   fx_placement1
3. Build the schematic like as follows:
    fx_placement2
4. Select Animation tool and transfer Column2 position. Leave the Column1 position unchanged.
    
5. Preview the scene. Compare rendered frames 1 and 2. The center position of the spin blur unexpectedly moves to the center. 
    fx_placement4

**Expected behavior**
Fx placement should not change even if the cell is empty in the "parent" column. 

**Device Information:**
 - OS: Win10
 - OpenToonz Version: current master

shun-iwasawa:
This PR fixes #4074 
Now fx will always use transformation of the parent column, even if the column contains empty cells.

### DIFF
--- a/toonz/sources/toonzlib/scenefx.cpp
+++ b/toonz/sources/toonzlib/scenefx.cpp
@@ -770,7 +770,8 @@ PlacedFx FxBuilder::makePF(TLevelColumnFx *lcfx) {
   assert(m_scene);
   assert(lcfx);
   assert(lcfx->getColumn());
-  if (!lcfx || !lcfx->getColumn()) return PlacedFx();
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #4077, opened 2021-09-28
Fix Fx intensity malfunction

### LINKED ISSUE #4076: Fx Intensity does not work when there are multiple output from the Fx
**To Reproduce**
1. Load or create some level. In this example, the following photograph (a level `bg`) is used.
    fx_intensity2
2. Build fx schematic like as follows:
    fx_intensity1
3. Set `Shift -> Hue` of the HSV Adjust Ino Fx to `180` and preview the scene. (No problem)
    fx_intensity3
4. Set `Fx Intensity`  of the HSV Adjust Ino Fx to `0.5` and preview the scene. (No problem)
    fx_intensity4
5. Set `Opacity` of the Over Ino Fx to `0` and preview the scene. (Wrong result)
    fx_intensity5

**Expected behavior**
In the above example, changing the `Opacity` value should not change the rendered result as the both input ports are connected to the same node.

**Device Information:**
 - OS: Win10
 - OpenToonz Version: current master

shun-iwasawa:
This PR fixes #4076 
The Fx intensity is achieved by inserting a cross dissolve fx which blends the fx result and its input.
When rendering, I set some flag for handling such insertion in order to prevent infinite loop.

However it did not work properly when the fx has more than one output.
Now I modified the flag and made it to work fine. Sorry for the trouble!

### DIFF
--- a/toonz/sources/toonzlib/scenefx.cpp
+++ b/toonz/sources/toonzlib/scenefx.cpp
@@ -551,7 +551,12 @@ public:
   // (at least) of a particle Fx
   int m_particleDescendentCount;
 
-  QList<std::wstring> m_globalControlledFx;
+  // fxid and pointer to the correspondent blend fx for the global
+  // control. Boolean is the flag indicating that the makePF is just called
+  // from the blend fx. If the flag is true then just compute the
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #4082, opened 2021-10-02
Add Raster Lock Alpha tool setting 

manongjohn:
A requested port of my Lock Alpha PR from Tahoma2D.

This PR adds Lock Alpha toolbar option to the following raster based brush tools

- Paint Brush Tool
- Toonz Raster Brush Tool
  - Standard brush (Hardness 100%)
  - Standard brush (blurred - Hardness < 100%)
  - MyPaint brush
- Raster Brush Tool
  - Standard brush
  - MyPaint brush - no change already has Lock Alpha

### DIFF
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -458,6 +458,7 @@ class PaintbrushToolOptionsBox final : public ToolOptionsBox {
 
   ToolOptionCombo *m_colorMode;
   ToolOptionCheckbox *m_selectiveMode;
+  ToolOptionCheckbox *m_lockAlphaMode;
 
 public:
   PaintbrushToolOptionsBox(QWidget *parent, TTool *tool,
--- a/toonz/sources/include/toonz/rasterstrokegenerator.h
+++ b/toonz/sources/include/toonz/rasterstrokegenerator.h
@@ -37,6 +37,7 @@ class DVAPI RasterStrokeGenerator {
   bool m_doAnArc;
   bool m_isPaletteOrder;  // Used in the Draw Order option of Brush Tool,
                           // use style order to define line stacking order
+  bool m_modifierLockAlpha;
   QSet<int> m_aboveStyleIds;
 
   // Ricalcola i punti in un nuovo sistema di riferimento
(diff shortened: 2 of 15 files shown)


RodneyBaker:
This looks to be a github actions build issue but there are reports of some mypaint brushes not working with Alpha Lock either on or off (on TLV levels).

Its not clear to me why it would effect some brushes but not others however.

I believe this was reported already.

RodneyBaker:
As the only reason to hold back this PR appears to be the testing environment of the github action builds I will press on with merge so that the nigthly can be used for user testing.

Any objections please let me know.
If none then I'll plan to merge this in approx. 1 hour.

beeheemooth:
Screen_0326

RodneyBaker:
Something to consider, perhaps with the next UI scrub, would be setting the text of Lock Alpha to the color red when on.  I note that it can be easy to forget that option is on and then wonder why we cannot draw.

RodneyBaker:
Thank you @manongjohn for this new feature.
LGTM

RodneyBaker:
alphalockA

Example of shapes worked on separately with Alpha Lock feature in TLV levels.
Final pass of blending in a few areas was done after rendering on the separate layers via Alpha Lock in standard raster (due to restriction of blending with mypaint brushes in TLV levels).

beeheemooth:
@RodneyBaker Hmm, your technique implies that the colors of different elements are better filled at different levels in order to be able to use the features described in this topic. Cool! How many interesting approaches open up before us! This needs to be pondered. Although I like to keep the colors of the character on one level and the shadows on another. So far, the only downside to your technique is fiddling with palettes. It is unknown what will happen to the style indexes if the user decides to merge the TLV levels.

RodneyBaker:
@beeheemooth  It seems to me a logical progression of this feature would be as follows:

1) Inversion
Create additional option to Invert the area being locked by the Lock Alpha option.
This would reverse the area the user is allowed to interact wth... allowing drawing/painting everywhere except where there is a value of 0 (black) in the Alpha Channel).  An option to be able to quickly reverse the inversion would make painting on or off of objects quick... such as using the shift key to switch back and forth

2) Offset/Tolerance (min/max range)
As you suggest gaining the abilty to set a tolerance level for the value in the alpha channel so that grayscale levels of a channel would provide an offset (min and max) where drawing is allowed.

3) Extend range of Alpha Lock intto RGB channels (RGBA Lock).
Allow the ranges of each RGBA channel (in standard vector levels) to dictate where pixels can be drawn.  Values outside this range would be locked to prevent drawing.
This option could get very complicated in that the process could start by simply preventing applcation of color but also might dictate which colors could be applied.
Some of the code for this likely exists already in the Adjust Levels FX.  That FX would likely need to become a gizmo/panel that could hover over the canvas.

4) Extend the RGBA Lock to TLV levels
The conversion from RGBA to TLV's unique formatting would take some time.
This assuming RGBA Lock didn't start off with smart/toonz raster and then be ported to standard raster.

5, 6, 7 etc.) Ink/Trace
Another complex undertaking, this would take something like John's current Reference Fills feature (in Tahoma2D) and apply it to where a user can draw/paint...
eventually with vector strokes.  This option would essentially guide a user in where they can draw based on previously drawn areas in other columns/levels referenced on frame.  I'll guess this would first be done with raster and then advance to vector.
An additional level of sensitivity would be set by the user to control how much influence the reference exerts on the placement of the new strokes/drawing.

Look for all of this to be fully inplemented in the 2026 release of OT/T2D.  10th anniversary edition!  :)

beeheemooth:
Rhodney, you have a very rich imagination, but it needs to be filtered, hehe. I figured out what you mean: you want to use a raster as a stencyl for airbrush. I can't imagine what it would be like to paint the lower layer through semi-transparent pixels.
However, we are here sharing the skin of an unkilled bear.  :D

RodneyBaker:
That's with the filter on!  ;)

We now return you to the original channel you were viewing.

PR #4087, opened 2021-10-05
Fix Undo Malfunction

shun-iwasawa:
This PR fixes a problem that undo stops working after one of the following operations:
1. 1. Select some region of a raster level image with the Selection tool and copy. 
    1. Select a cell containing either Toonz Raster or Toonz Vector level and try to paste. (An error message opens.)
2. 1. Select the Animate tool. Start dragging to change a column transformation.
    1. While pressing the left mouse button, switch tools via shortcut key.
3. 1. On Function Curves panel, start dragging some point in a curve.
    1. While pressing the left mouse button, press either middle or right mouse button.
4. 1. Open the Level Settings popup.
    1. Try to change the level name to the one of another existing level. (An error message opens.) 

As you can see in the History panel, after doing one of the above operations, OT stops to register subsequent operations.

The problem was due to missing of calling the function `TUndoManager::endBlock()` .
OT can combine more than one undo-able operations as one undo block. The operations registered between the functions `beginBlock()` and `endBlock()` are combined into one block. However, if the call of `endBlock()` is bypassed somehow, all subsequent operations are wrongly registered in one block and fails to work.

### DIFF
--- a/toonz/sources/tnztools/edittool.cpp
+++ b/toonz/sources/tnztools/edittool.cpp
@@ -123,7 +123,11 @@ public:
         ->setCenter(m_objId, m_frame, m_center);
   }
   void leftButtonUp(const TPointD &pos, const TMouseEvent &) override {
-    if ((m_lockCenterX && m_lockCenterY) || m_firstPos == pos) return;
+    if (m_firstPos == pos) return;
+    leftButtonUp();
+  }
+  void leftButtonUp() override {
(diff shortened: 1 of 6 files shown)


RodneyBaker:

In addition to the undo fix this PR seems to address a copy/paste crash I've long had (from raster to raster) but I will do more testing to be sure.  That and I am experiencing a considerable speed up in drawing.  Not sure how this fix would have accomplished either of these.

PR #4098, opened 2021-10-12
Fix Rendering Fx After Xsheet Node

### LINKED ISSUE #4096: Effects after XSheet node don't work anymore
Hello,

In few recent Nigthly versions of opentoonz one thing seems to be broken.

I used to add nodes after the main XSheet node like in the picture below. Mostly to add some layer over the whole scene or for example blur the whole picture:
image

It worked as planned - the effect after the XSheet node affected whole scene. 

Lately Xsheet node output is invisible to effects placed between XSheet and Output node.

I hope it's a bug not a feature:)

shun-iwasawa:
This PR fixes #4096 which had occurred due to my recent modifications.
Sorry for the trouble!


RodneyBaker:
Thanks to @papio-pl for the report and thank you @shun-iwasawa for the quick fix!

PR #4124, opened 2021-10-27
Xsheet Minimum Layout

shun-iwasawa:
**UPDATE** I modified the row area and note area to be narrower, by putting buttons in the note area into a hamburger menu.

minimum_layout_mod

This PR introduces new 2 options to display Xsheet simpler & smaller suitable for pencil testing workflow.
minimum_layout
- Added a new option `Minimum` to `Preferences > Xsheet > Column Header Layout` .
     - This layout has the minimum-sized column header (no thumbail, parent object name or lock button) and narrower column width.
     - In the Minimum layout the lock button moves in the column option popup.
        column_option
- Added a new preference item `Level Name Display` to `Preferences > Xsheet` category.
    - The item `Preferences > Interface > Display Level Name on Each Marker` is removed instead.
    - `Level Name Display` has 3 options:
        - **Default** : corresponds to the former `Display Level Name on Each Marker = OFF`
        - **Display on Each Marker** : corresponds to the former `Display Level Name on Each Marker = ON`
        - **Display on Column Header** : a new option. If the column contains single level, the level name will be displayed in the column header instead of cells. If the column contains multiple levels, columns displays level name just like the same way as the `Default` option. In the above image, the column 6 contains 2 levels so the level names are displayed in the cell area.

### DIFF
--- a/toonz/sources/include/orientation.h
+++ b/toonz/sources/include/orientation.h
@@ -199,7 +199,8 @@ enum class PredefinedFlag {
   THUMBNAIL_AREA_BORDER,
   THUMBNAIL_AREA_VISIBLE,
   CAMERA_ICON_VISIBLE,
-  VOLUME_AREA_VERTICAL
+  VOLUME_AREA_VERTICAL,
+  NOTE_AREA_IN_POPUP
 };
 
 // Knows everything about geometry of a particular orientation.
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -112,6 +112,12 @@ public:
     ToggleBetweenGraphAndSpreadsheet
   };
 
+  enum LevelNameDisplayType {
+    ShowLevelName_Default = 0,
+    ShowLevelNameOnEachMarker,
+    ShowLevelNameOnColumnHeader
+  };
+
   //--- callbacks
   // General
   void enableAutosave();
@@ -213,14 +219,8 @@ public:
   bool isActualPixelViewOnSceneEditingModeEnabled() const {
     return getBoolValue(actualPixelViewOnSceneEditingMode);
   }
-  bool isLevelNameOnEachMarkerEnabled() const {
-    return getBoolValue(levelNameOnEachMarkerEnabled);
-  }
   bool isShowRasterImagesDarkenBlendedInViewerEnabled() const {
     return getBoolValue(showRasterImagesDarkenBlendedInViewer);
-  }
-  bool isShowFrameNumberWithLettersEnabled() const {
-    return getBoolValue(showFrameNumberWithLetters);
   }
   TDimension getIconSize() const { return getSizeValue(iconSize); }
   void getViewValues(int &shrink, int &step) const {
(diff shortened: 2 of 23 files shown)


shun-iwasawa:
I moved the buttons in the note area in a hamburger menu in the minimum layout.

RodneyBaker:
Looks great.

I still see a minor issue with regard to Notelevels as the user cannot see in real time what they have entered into the cell due to inadequate width but I am confident this will be addressed in the future.  Perhaps some column width multiplier for Notelevels that would take the current hard coded value of the selected layout  and multiply it by some value entered into Preferences.  A minimum and maximum range would likely need to be set to keep that value in a safe zone.

That you @shun-iwasawa for adding this new Xsheet Minimum Layout.
I can easily see it becoming the default xsheet layout.

adamearle:
> I moved the buttons in the note area in a hamburger menu in the minimum layout.

Have you seen the xsheet / timeline parenting in Tahoma it's super handy? 

Also, how do you feel about a campaign for OpenToonz to upgrade the viewport to be 3D and for the function editor to be updated as well? Would love to talk to you more about this if I can so we can push forward with it in the right direction.

shun-iwasawa:
@adamearle 
> Also, how do you feel about a campaign for OpenToonz to upgrade the viewport to be 3D and for the function editor to be updated as well? Would love to talk to you more about this if I can so we can push forward with it in the right direction.

Sorry but I think I don't get your point. What does "upgrading the viewport and the function editor to be 3D" mean?

gab3d:
I think he said: "upgrade the viewport to be 3D" and "the function editor to be updated as well".

RodneyBaker:
Not sure anything related to viewport really applies to this PR.
Perhaps a new issue can be opened to explore those ideas.

RodneyBaker:
Some post mortem feedback...
When using the xsheet minimum layout it doesn't appear that we can select columns and collapse them into a subxsheet.
At this point I'm not sure if this is something on my end or not but wanted to add this note in case other users observed the same.

beeheemooth:
Collapse works for me.

RodneyBaker:
Thanks @beeheemooth 
It must just be some mouse/pen issue on my end.

adamearle:
> @adamearle
> 
> > Also, how do you feel about a campaign for OpenToonz to upgrade the viewport to be 3D and for the function editor to be updated as well? Would love to talk to you more about this if I can so we can push forward with it in the right direction.
> 
> Sorry but I think I don't get your point. What does "upgrading the viewport and the function editor to be 3D" mean?

Hiya mate, Sorry about the confusion.

I meant to upgrade the viewport to a proper 3D viewport. This would mean:

1. Gaining the extras 3D rotation axis that is missing, which means there could be better integration of 3D applications such as Blender working with cameras and 3D scenes. Perhaps the ability to bring 3D geometry. A game engine-styled 3D viewport has been recommended here.

At the moment we have a small slice of code in the works to get camera data from OpenToonz to Blender 3D and back again. We would like to make it more open so people can set up their own export presets based on the 3D package that people are using ie Blender, Natron, After Effects, Godot, Unreal, Fusion, Cinema 4D, Maya, 3D Max,  Soft Image, and so on.

Fixing up the function editor would involve
1. Add the missing rotation axis.
2. Being able to select and manipulate more than 1 keyframe at a time. (This will speed up re-timing animations when supervisors give retakes).
3. General clean-up of how the data is viewed.
4. adding the ability to Mute, and  Solo channel switched.
5. Click and drag across switches.
6. Viewing only selected data.

Miscellaneous items:
1. We would also seek to add the much-needed masking column back into OpenToonz. This always gets talked about.
2. Updating the vector so that they can be tweened similar to Natron's rotoscoping tools
3. Try and get Open Fx compatible with OpenToonz.

The above items are what many have discussed that would attract more development to Opentoonz. Out of the 10 Items on the list the most important.

1. Being able to select and manipulate more than 1 keyframe at a time. (This will speed up retiming animations when supervisors give retakes).
2. We would also seek to add the much-needed masking column back into OpenToonz. This always gets talked about.
3. Gaining the extras 3D rotation axis that is missing, which means there could be better integration of 3D applications such as Blender working with cameras and 3D scenes.

adamearle:
> Not sure anything related to viewport really applies to this PR. Perhaps a new issue can be opened to explore those ideas.

Where can we put this so we can start talking about this?

PR #4128, opened 2021-10-29
Updating the libtiff library (WIP)

jointri:
This plan isn't going to work. I'm closing this PR now.

In the scenario that @shun-iwasawa doesn't want to use tinytiff, I dropped in the tif_getimage_64.c and had to change stuff like uint16 to uint16_t and put the header file #include <stdint.h> to be able to compile with only a few warnings. This is still the libtiff tif_getimage_64.c from 2012, so it may have vulnerabilities inside it, so it might be necessary to convert the 4.3.0 tif_getimage.c to tif_getimage_64.c. This is if shun doesn't want to use tinytiff, I'll leave it up to his discretion.


RodneyBaker:
Tagging as 'Help Wanted' per discussion with @jointri who stated this PR will need assistance to move forward.

@shun-iwasawa Thank you for your assistance thus far and for considering the various options available.

jointri:
Thanks for the tag rodney. We can boil this down to 3 possible options to fixing this libtiff problem:

1. Remain static linked, but update tif_getimage_64.c to each new libtiff release
2. move tif_getimage64.c out of the libtiff directory and compile it as a separate thing independent from libtiff.
3. Abandon libtiff entirely and switch to tinytiff.

I'll leave it to @shun-iwasawa to make the final call on what we do.

jointri:
If I can offer a 4th suggestion @shun-iwasawa why not compile "tif_getimage_64.c" as a shared object file (.so/.dll/whatever macos uses) and still use the system libtiff? You definitely can compile a single file as a shared object file, I managed to do so on my linux machine.

jointri:
Just an observation @shun-iwasawa but it seems that opentoonz has been relying on tiff-4.0.3 for so long, it can't compile with tiff 4.1.0 or later. I don't know which of these two would be less work, but I suggest we either backport fixes from tiff-4.3.0 to tiff-4.0.3 or we update opentoonz to work with later versions of libtiff.

RodneyBaker:
A 5th option would appear to be to leave things as is and contrinue R&D.

It seems the vfx industry itself with all the major players all struggle with use of libtiff.
The VFX reference platform was designed to address issues such as this but they haven't included Libtiff in their listing as of yet (nor libraries such as libpng, etc.).  They have had some discussions related to libtiff however.   Here's a snippet of one response from Larry Gritz:

>  ...libtiff 4.0 dates from 2012. So people writing plugins for Maya, or libraries that might be called by Maya plugins, if they use libtiff directly or indirectly, are probably, and likely without being aware that they are doing it, using the more modern version, only to get strange crashes or link errors when their plugin is loaded into Maya. But if they build their library against libtiff 3.9, to make it work with Maya, it won't work with any other apps that the library might be used in conjunction with.
> 
> It would be really fantastic if you guys could link against a STATIC libtiff.a, and use symbol hiding so that plugins don't see whatever version you build with at all. Then we wouldn't be particularly sensitive to what version Maya was using internally. (except for issues of file compatibility...)
> 
> But in addition to the linkage issue, there's also the matter that TIFF writers using 4.x may put Exif information into the TIFF header, whereas 3.x doesn't understand that tag and will spew warnings when it sees that. (Yet another solution there is to intercept the warnings so they don't spill all over the console.)

TLDR:  Most companies/software apparently use custom versions of libtiff to meet their needs.
In short, it seems the major problems with libtiff not with Opentoonz but upstream.

jointri:
@RodneyBaker @shun-iwasawa 
I am starting to agree with you that it's just too much work to move to modern libtiff and might be just as much work moving to tinytiff. However, can we look into backporting security patches (For CVEs) and bug fixes only from libtiff git to thirdparty/tiff-4.0.3?

PR #4132, opened 2021-11-04
Fix wheel scrolling on the output settings popup

### LINKED ISSUE #4131: Scrolling with the mouse wheel in the output settings should not change any settings
When you scroll down the output settings using your mouse wheel, if your pointer goes above any dropdown boxes, it can change it accidentally. I've seen lots of users asking why their output appears as separate files and that's because their pointer was accidentally over the Multiple Rendering dropdown box. Unless you click on the dropdown it shouldn't select it. This is causing lots of confusion and annoyance with the program.

shun-iwasawa:
This PR fixes #4131 .
Now all combo boxes in the output settings ignore mouse wheel input unless they have the input focus.

### DIFF
--- a/toonz/sources/toonz/outputsettingspopup.cpp
+++ b/toonz/sources/toonz/outputsettingspopup.cpp
@@ -50,6 +50,7 @@
 #include <QScrollArea>
 #include <QPropertyAnimation>
 #include <QSpacerItem>
+#include <QEvent>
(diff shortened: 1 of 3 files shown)


DarrenTAnims:
Thank you so much for fixing this. This will stop many problems for users.

RodneyBaker:
Nice.

Issue #3637, opened 2020-12-15
Requested adjustments from Tahoma2D


ghost:
X-Sheet Level Parenting.
![image](https://user-images.githubusercontent.com/74626650/102279078-11637e80-3ee8-11eb-985b-2f2c5a276a02.png)

Displayed TimeLine Effects Icons.
![image](https://user-images.githubusercontent.com/74626650/102279142-2b04c600-3ee8-11eb-81bf-907fbc07783e.png)

Diamond Keys  to match the ui icons design.
![image](https://user-images.githubusercontent.com/74626650/102279270-57204700-3ee8-11eb-87e4-dfb1ddc4ac19.png)

All of these changes are all available and can be ported from Tahoma2D.?????
I believe these Adjustments would be  Great for OpenToonz 1.5

Links
https://github.com/tahoma2d/tahoma2d/pull/68
https://github.com/tahoma2d/tahoma2d/pull/275
https://github.com/tahoma2d/tahoma2d/pull/288
https://github.com/tahoma2d/tahoma2d/pull/130
https://github.com/tahoma2d/tahoma2d/

JPUser1:
I think these are nice Adjustments and only need someone to port it.

JPUser1:
@shun-iwasawa Please consider some of these Adjustments for OpenToonz 1.5

RodneyBaker:
Closing as the original reporter has ghosted themself.
Note that there may still be useful information contained in this report and where appropriate that information should be transferred to a new report.

Note that portions of this request may have been implemented.

PR #4134, opened 2021-11-05
Tahoma2D porting: Allow Changing Parent of Column from XSheet by jeremybullock

shun-iwasawa:
This PR will port the following modification:
- tahoma2d/tahoma2d#68 by @jeremybullock
    I ported the modifications except the one done in `styleeditor.cpp` .

The further modifications (tahoma2d/tahoma2d#275 and tahoma2d/tahoma2d#288) seemed not to be necessary to introduce since OT still displays handle name along with the parent object name as of now.

This PR resolves one of the requested features in #3637 and [this comment](https://github.com/opentoonz/opentoonz/pull/4124#issuecomment-961523858).

Thank you @jeremybullock for the modification !

### DIFF
--- a/toonz/sources/toonz/xshcolumnviewer.cpp
+++ b/toonz/sources/toonz/xshcolumnviewer.cpp
@@ -339,6 +339,7 @@ void ChangeObjectWidget::show(const QPoint &pos) {
   }
   setGeometry(pos.x(), pos.y(), m_width, itemNumber * 16 + 2);
   QListWidget::show();
+  raise();
   setFocus();
   scrollToItem(currentItem());
 }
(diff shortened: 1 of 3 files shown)


shun-iwasawa:
@RodneyBaker 
When you merge this PR please select `Create a merge commit` instead of `Squash and merge` so that the author name of the commit will be maintained. Thanks!

RodneyBaker:
Thanks @shun-iwasawa , I hadn't even realized that was an option.
Very good to know.

RodneyBaker:
This PR as is works great.
There is some additional interest in seeing the option of color coding based on level type but we will save that for a subsequent PR.

Thank you @manongjohn for bringing this feature back and @shun-iwasawa for the port! 
LGTM

JPUser1:
Thanks! for the port @shun-iwasawa would it also be possible to add the color of the parented level in the X-sheet  which u can turn off/on in preferences like in Tahoma?

DarrenTAnims:
This is great. Thanks for merging this into OT and thanks to Jeremy for the original work.
Could this also be added to the dropdown for when the xsheet is in minimal mode? I think it would be a good addition there too.

gab3d:
@DarrenTAnims 
mmm... wasn't the purpose of the xsheet minimal layout to aid when doing pencil tests?
if that's the case, I don't see how having the parenting dropdown would be of much use in that scenario.
perhaps you're intending to use this layout for other uses?

TomDoingArt:
Thank you for the work on maintaining OT.
I was seeing this feature demonstrated by Darren T on a recent livestream. I commented that it would be good to have a tooltip explaining this feature and he suggested to add it as a comment here.
This would be similar to the tooltip that shows up when hovering over the top of the column. Attached is a screenshot from Tahoma2D 1.2 because that is what I am using at the moment, but it shows what I am trying to describe.
image

PR #4146, opened 2021-11-10
Add Shader Fx: HSL Blend GPU

justburner:
This PR adds a shader Fx to perform component blending modes such as: Hue, Saturation, Color and Luminosity.
Those exists in programs like Photoshop, Krita, CSP... but does not natively exists in OpenToonz or Tahoma2D.
It will be found under "Shaders" as `GPU HSL Blend`.

Settings to perform each blending mode:

Hue Blending:
☑️ Hue
⬜ Saturation
⬜ Luminosity

Saturation Blending:
⬜ Hue
☑️ Saturation
⬜ Luminosity

Color Blending:
☑️ Hue
☑️ Saturation
⬜ Luminosity

Luminosity / Brightness Blending:
⬜ Hue
⬜ Saturation
☑️ Luminosity

Note: *Feature already exists in Tahoma2D since Nov 2020, thanks everyone who shown interest for this to be ported into the upcoming OpenToonz 1.6 :)*

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -531,6 +531,12 @@
   <item>"SHADER_spinblurGPU.radius"		"Safe Radius"	</item>
   <item>"SHADER_spinblurGPU.blur"		"Blur"	</item>
 
+  <item>"SHADER_HSLBlendGPU"			"GPU HSL Blend"    </item>
+  <item>"SHADER_HSLBlendGPU.bhue"			"Hue"    </item>
+  <item>"SHADER_HSLBlendGPU.bsat"			"Saturation"    </item>
+  <item>"SHADER_HSLBlendGPU.blum"			"Luminosity"    </item>
+  <item>"SHADER_HSLBlendGPU.balpha"			"Blend Alpha"    </item>
+
 <!---------------------------------------------------------------------------------------->
 
 
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -535,7 +535,8 @@
   <item>"SHADER_HSLBlendGPU.bhue"			"Hue"    </item>
   <item>"SHADER_HSLBlendGPU.bsat"			"Saturation"    </item>
   <item>"SHADER_HSLBlendGPU.blum"			"Luminosity"    </item>
-  <item>"SHADER_HSLBlendGPU.balpha"			"Blend Alpha"    </item>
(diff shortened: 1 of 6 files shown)


RodneyBaker:
Congratulations on your first PR @justburner!

justburner:
No need to have my name there, specially for a general use effect because I didn't wrote the original algorithm for the blending (the heart of the effect), I've ported it from a pseudocode in a OpenGL ES Extension documentation so it should be unfair for me to get credits as author, original reference link is credited on the source code of the shader.

Well, I have to admit that i don't fully understand the license behind it, the license covers a "Specification" but that pseudo-code is used to demonstrate how that OpenGL ES Extension should work under that "Specification", yet this PR has nothing to do with any OpenGL Specification or Extension... it gets worse that many links in the document itself are dead so likely there's more into it but i can't track it.

justburner:
I've tracked an e-mail of the original author of that document and i'm asking further information, for now please put this PR in standby.

RodneyBaker:
There is a blanket document in the shaders directory that should cover the basic shader information.
If the other parts fall into the Kronos groups realm (and aren't someone's personal property) then that alone might be sufficient.
Other shaders distributed with Opentoonz rely on that document.
Exceptions (if any) would usually be included in the shader itself.

https://github.com/opentoonz/opentoonz/blob/master/stuff/library/shaders/readme.txt

justburner:
Just an update of the situation, I've waited 2 weeks for an e-mail response but didn't receive anything sadly (also checked the spam folder just in case but nothing there either).
Anyway I'll trust your judgement of how to handle this situation @RodneyBaker.

Also added "Clipping Mask" and renamed "Blend Alpha" to "Opacity" to be more similar to Layer_Blending_Ino effects, port names remain unchanged to avoid OT crashing on reading scenes with old version of this FX.

RodneyBaker:
Thanks @justburner
It appears to me the current comment in the code as well as the general purpose statement (and links) should cover this FX as well as it does the other FX included with Opentoonz.

Will wait for your latest updates to build and proceed with merger if there are no other issues to consider.

Thanks not only for this shader FX but also for the additional due diligence with regard to authoring and licensing.

RodneyBaker:
@shun-iwasawa The Nightly build appears to have failed for this PR.

PR #4160, opened 2021-11-26
File Path Processing Using Regular Expression

shun-iwasawa:
suffixes
 
This PR introduces regular expression for processing file path. It will make naming rules for frames more flexible. As the benefit of this feature, this PR also enables to remove restrictions of the suffixes after drawing numbers.

Before this PR only one alphabet had been allowed for the suffix (like 1**a**, 2**b**, etc.).
With this PR users can use any type and number of letters like (1**abc**, 2**あいうえお**) . The rules can be modified in the project settings.

This feature was demanded by Japanese animation industry, especially key frame animators. When they insert a drawings between existing frames, they tend to name drawings with numbers + special suffixes based on their own naming rules.

### How To Change the File Path Rules ###
1. Select `Menu bar > File > Project Management > Project Settings` . (The Project Settings popup will open.)
1. Select a `File Path Rules` tab.
1. Select a radio button `(Experimental) Custom` .
1. Specify the rules for suffixes.
    - **Accept Non-alphabet Suffix** enables to use non-alphabetical letters for suffixes (such as Japanese Hiragana). 
    - **Maximum Letter Count For Suffix** specifies the maximum number of letters for suffixes. (For instance setting this value to `3` enables to make a drawing named `1abc` .)
1. Note that the rules are part of the project settings. The rules will be shared to all scenes under the same project.

### CAUTION ###
1. This feature is experimental. Please make sure to keep backup before use.
1. When the `(Experimental) Custom` option is selected, the file format version of the Toonz Raster Level (.TLV) and Toonz Vector Level (.PLI) will be altered when saved (even if the level does not contain frames with suffixes). They will not be loaded properly by the previous versions of OT. 
1. To use this feature you need to turn off the preference option `Show "ABC" Appendix to the Frame Number in Xsheet Cell` .

### DIFF
--- a/toonz/sources/common/tcontenthistory.cpp
+++ b/toonz/sources/common/tcontenthistory.cpp
@@ -73,7 +73,7 @@ inline QString blanks(const QString &str, int count = 15) {
 //--------------------------------------------------------------------
 
 inline QString getStr(const TFrameId &id) {
-  if (id.getLetter() != 0)
+  if (!id.getLetter().isEmpty())
     return QString::number(id.getNumber()) + id.getLetter();
   else
     return QString::number(id.getNumber());
@@ -159,7 +159,7 @@ const QString TContentHistory::currentToString() const {
     dateSorted.insert(pair<QDateTime, TFrameId>(it->second, it->first));
 
   std::multimap<QDateTime, TFrameId>::const_iterator it1 = dateSorted.begin();
-  QDateTime currDate = it1->first;
+  QDateTime currDate                                     = it1->first;
 
   while (it1 != dateSorted.end()) {
     set<TFrameId> frames;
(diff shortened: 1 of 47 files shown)


shun-iwasawa:
I need to take care the mesh file frames with special suffixes.

RodneyBaker:
YIKES!  That has some pretty scary (good!) potential @shun-iwasawa  :)
I like!

RodneyBaker:
@shun-iwasawa You mentioned needing to work on mesh file frames with special suffixes.

I assume you prefer this PR wait to be merged until after that addition.

shun-iwasawa:
@RodneyBaker 
I already fixed the insufficiency in mesh file handling (and other bugs). 
I think it is ready to be merged. Thanks!

RodneyBaker:

PR #4163, opened 2021-11-28
Fix crash on creating Macro Fx from the Fx Browser

### LINKED ISSUE #4162: Macro FX added from Browser FX causes crash.
I am getting stable crash when loading MacroFX from FX Browser. However, if I load the same MacroFX from the context menu, then it loads without problems. 

**To Reproduce**
1. Create a MacroFX from several effect nodes (mine is "Matte Visible In", "Palette Filter", "Over") and save it as a preset.

2. Load the saved MacroFX into the FX Schematic window from the context menu. It will load successfully.

3. Load the saved MacroFX into the FX Schematic window from the FX Browser. A crash will occur.

**Screenshots**

ezgif-2-1e14111f1cf3

The same crash occurs in Tahoma2D. In the video you can see the Tahoma interface, I accidentally deleted a similar video recorded in OT (it's a pity, the eye was prettier there). 

**Device Information:**
 - OS: Windows 8.1

shun-iwasawa:
This PR fixes #4162 .
The crash had occurred when the macro fx  contains a fx with "port dynamic group" (= a group of ports which increases when the last port is connected).

I modified the crash by making the macro fx creation procedure to be the same as the one called from the context menu in the Fx Schematic.

### DIFF
--- a/toonz/sources/toonz/insertfxpopup.cpp
+++ b/toonz/sources/toonz/insertfxpopup.cpp
@@ -88,52 +88,51 @@ TFx *createPresetFxByName(TFilePath path) {
 }
 
 //-----------------------------------------------------------------------------
+// same as createMacroFxByPath() in addfxcontextmenu.cpp
 
 TFx *createMacroFxByPath(TFilePath path) {
-  TIStream is(path);
-  TPersist *p = 0;
-  is >> p;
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thank you @shun-iwasawa for the fix (and fast turn around!).

(Will watch the nightlies to see if it properly builds)

Issue #730, opened 2016-08-12
Feature Request: ability to export for game (asset)


2dvision:
I use [Spriter](https://brashmonkey.com/), but in Spriter you can not drawing directly. toon boom can export animation for game too, you can read [here](https://www.toonboom.com/industry/games) about that.

at the moment in OT you can export animate GIF for use in your game, but animate GIf do no support transparency and trim now. #729

I am beginner in game developing, I hop professional game developer expand this request. :)

2dvision:
https://github.com/opentoonz/opentoonz/issues/729

TheoXD:
related to this https://github.com/opentoonz/opentoonz/issues/173

blurymind:
This is not the first time it's raised as a duplicated request - seems to be a highly desired feature :D 
Too bad opentoonz devs are not into game development like us

PR #4164, opened 2021-11-29
Improvements to Audio Recording

justburner:
This PR is targeted to improve Audio Recording Popup compatibility between platforms and other things:

- Fixes Windows 10 recording at very low sample rate like 8KHz instead of 44.1KHz ([Tahoma2D #730](https://github.com/tahoma2d/tahoma2d/issues/730))
- Fixes bug of duplicated devices in the list ([Tahoma2D #730](https://github.com/tahoma2d/tahoma2d/issues/730))
- Fixes a bug that causes the program to crash when pressing Record in a room without XSheet or Timeline
- Adds `Default Audio Input` to record from default device in the OS
- Add ability to change sample rate, bit depth and number of channels to record
- Allows to refresh devices list without completely reopening the program
- Additional tool-tips that provides more information to the user
- Slightly bigger record, playback and pause buttons
- Moved Sync checkbox up so the Save button is now on the very bottom
- Timers now have accuracy in cents of a second

Windows and Mac platforms were tested and seems to be working. (Mac tested by @artisteacher, thanks)
I'll greatly appreciate if this is tested on Linux before accepting the merge.

### DIFF
--- a/toonz/sources/toonz/audiorecordingpopup.cpp
+++ b/toonz/sources/toonz/audiorecordingpopup.cpp
@@ -64,29 +64,47 @@ AudioRecordingPopup::AudioRecordingPopup()
   m_saveButton           = new QPushButton(tr("Save and Insert"));
   m_pauseRecordingButton = new QPushButton(this);
   m_pausePlaybackButton  = new QPushButton(this);
-  // m_refreshDevicesButton = new QPushButton(tr("Refresh"));
-  m_duration           = new QLabel("00:00");
-  m_playDuration       = new QLabel("00:00");
+  m_refreshDevicesButton = new QPushButton(this);
+  m_duration           = new QLabel("00:00.000");
+  m_playDuration       = new QLabel("00:00.000");
   m_deviceListCB       = new QComboBox();
   m_audioLevelsDisplay = new AudioLevelsDisplay(this);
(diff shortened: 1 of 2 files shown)


justburner:
Look in Windows 10

Original:
Old look

This PR:
New look

Oh forgot to mention that the peak meter is only visible during recording and playback.

artisteacher:
It's really nice having control of more settings!

justburner:
Commit jokes aside, I've noticed that QAudioRecorder did wrote directly to disk so I've changed code to mimic the same.
Previous implementation of writing to memory buffer is still there so it can be reverted or something else in the future, each implementation have their own pros and cons.

Additionally added a peak line that show the maximum peak level reached, if it exceeds 99,5% it will show a red bar at the right to tell the user that sound was likely clipped or distorted.

andeon:
The record button is working fine now.

The play button doesn't work, but it has never worked on any version that I can remember.. This is the message if you want to know:

```
Warning: "No decoder available for 'audio / x-wav' type."

Error: "The GStreamer installation is missing a plugin."
```
Anyway you can push this update!

PS:  I'm gonna to try to test a VM by changing gstreamer plugins to see if it's something missing or maybe an incompatible OS version. Has a lot  of plugins gstreamer1.0-plugins-ugly, gstreamer1.0-plugins-bad, gstreamer1.0-plugins-good

Issue #3994, opened 2021-06-25
.psd import OpenToonz 1.5 (Windows)


rabe01:
Importing .psd files OpenToonz 1.5
I saw that this problem was worked on - and closed - back in 2017. But even today, in 2021, I cannot properly import .psd files - neither from Krita, Clipstudio Paint or Photoshop CS6.
Tools like Moho or Cartoon Animator naturally import characters with many layers into different layers.
Although OpenToonz offers to import layers into multiple columns, I constantly have difficulties:
- psd code name error
- Alpha channel doesn't seem to be working (Alpha with "halo")
- Imported images must first be "combined" with a raster-layer in order to be editable ...
Am i too stupid or can someone help? Or should that be improved in OpenToonz?
Greetings

RodneyBaker:
@rabe01 Can you post/attach a .PSD file that you are finding to be particularly troublesome.
That will help greatly with testing.  

The Alpha with halo issue seems related to Premultiplied Alpha.

![image](https://user-images.githubusercontent.com/2576432/123530038-d377bc00-d6bb-11eb-9691-7409e5472238.png)

If you can supply any more information on the PSD code name error that will be appreciated.  Perhaps some screenshots or video capture?

Thanks!

shun-iwasawa:
> Imported images must first be "combined" with a raster-layer in order to be editable

OT currently cannot write images in PSD format, thus the file is loaded with a read-only flag.

Regarding the `psd code name error` , there are some lines ([this](https://github.com/opentoonz/opentoonz/blob/5aa5de6505d2612661466a707fdfd8392082cae5/toonz/sources/common/psdlib/psd.cpp#L1230) and [this](https://github.com/opentoonz/opentoonz/blob/5aa5de6505d2612661466a707fdfd8392082cae5/toonz/sources/common/psdlib/psd.cpp#L1254)) showing such error message under some conditions. As @RodneyBaker pointed, posting the problematic psd file would be helpful to find the cause in detail.

rabe01:
Sorry for my late reaction.
@shun-iwasawa thanks for the tipp to convert .psd-file in raster-layer. But in most cases I cannot import a .psd file at all.
I can not attache an example of a simple .psd-file here. I have sent a mail to Rodney. Hope it helps.

Greetings

rabe01:
Hi,
I am here one more time. Maybe someone knows a reason why I can't load .psd files into OpenToonz?
[...]

PR #4167, opened 2021-12-01
Fix PSD loading and level format preferences

shun-iwasawa:
This PR fixes two problems (the missing premultipy option and the insertion of empty columns) reported by @artisteacher in [this comment](https://github.com/opentoonz/opentoonz/issues/3994#issuecomment-982102520).


RodneyBaker:

artisteacher:
This works great. Thanks so much for the quick fix!

PR #4175, opened 2021-12-06
Fix shortcut loading from file

shun-iwasawa:
This PR fixes a following problem regarding the Configure Shortcut popup.
- `Loading from file..` does not work in non-English UI (such as Japanese).


RodneyBaker:

PR #4176, opened 2021-12-06
Allow Renaming Pass Through Nodes

konero:
This is a port from Tahoma2D, it will restore the ability to rename Pass Through nodes, this was useful for naming ports for MacroFx that are used in many scenes, the regression came about due to #4036.

renamepassthru1

The node has no name displayed by default, double-click to rename and it will show after hitting enter, thanks and credit to @manongjohn for developing it.

I love the redesigned node, just needed the port naming. 😄 


RodneyBaker:
This works great @konero.
Will merge after the appveyor checks complete successfully as I know quite a few are waiting for this capability to be restored.

PR #4178, opened 2021-12-07
Cell Mark feature

shun-iwasawa:
This PR introduces a new feature "Cell Mark", a versatile mark to be put on xsheet cell.

#### To add / remove the mark to cell ####
- Use commands in the right click menu of the xsheet cell > `Cell Mark` , or
- Use shortcut `Set Cell Mark` .
    - `Cell Mark` shortcut keys are registered in `Right-click Menu Commands` category in the Configure Shortcuts popup. 

#### To change name and color of cell marks ####
- Scene Settings > `Edit Cell Marks`

#### Utility of the cell marks as of now ####
- When loading / exporting XDTS file, cell marks can be corresponded to the `Inbetween Symbol` and the `Reverse sheet symbol` which had been ignored in the previous versions. 
- When exporting xsheet PDF, cell marks can be corresponded to the "inbetween symbols" and the "keyframe symbols" which are common in Japanese animation industry.
    

#### NOTICE ####
- If you set the mark to xsheet cell or change color or name of the marks from the default, such scene won't be opened properly by the previous version of OpenToonz.
- The default colors and names of the marks are consistent with the [Image Marks](https://www.tvpaint.com/doc/tvp11/index.php?id=lesson-tradigital-animation-additional-other-options-image-marks&lang=en), a similar feature implemented in TVPaint.

### DIFF
--- a/toonz/sources/include/orientation.h
+++ b/toonz/sources/include/orientation.h
@@ -67,6 +67,7 @@ enum class PredefinedRect {
   BEGIN_EXTENDER,           //! top / left extender
   KEYFRAME_AREA,            //! part of cell dedicated to key frames
   DRAG_AREA,                //! draggable side bar
+  CELL_MARK_AREA,           //! cell mark
   SOUND_TRACK,              //! area dedicated to waveform display
   PREVIEW_TRACK,            //! sound preview area
   BEGIN_SOUND_EDIT,         //! top sound resize
--- a/toonz/sources/include/toonz/sceneproperties.h
+++ b/toonz/sources/include/toonz/sceneproperties.h
@@ -50,6 +50,11 @@ class DVAPI TSceneProperties {
 public:
   typedef std::vector<double> Guides;
 
+  struct CellMark {
+    QString name;
+    TPixel32 color;
+  };
+
 private:
   Guides m_hGuides, m_vGuides;
 
@@ -74,6 +79,9 @@ private:
 
   //! Xsheet Note Color, color number = 7.
   QList<TPixel32> m_notesColor;
+
+  // Cell Mark colors and names
+  QList<CellMark> m_cellMarks;
 
   bool m_columnColorFilterOnRender;
   TFilePath m_camCapSaveInPath;
@@ -280,6 +288,12 @@ and height.
   QList<TPixel32> getNoteColors() const;
   TPixel32 getNoteColor(int colorIndex) const;
   void setNoteColor(TPixel32 color, int colorIndex);
+
+  QList<CellMark> getCellMarks() const;
+  CellMark getCellMark(int index) const;
+  void setCellMark(const CellMark &mark, int index);
+  bool hasDefaultCellMarks()
+      const;  // check if the cell mark settings are modified
 
 private:
   // not implemented
--- a/toonz/sources/include/toonz/txshcolumn.h
+++ b/toonz/sources/include/toonz/txshcolumn.h
@@ -9,6 +9,7 @@
 
 #include <QPair>
 #include <QString>
+#include <QMap>
 
 #undef DVAPI
 #undef DVVAR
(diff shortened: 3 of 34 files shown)


RodneyBaker:
Everything looks good here @shun-iwasawa 

Before proceding I want to ask if it would be possible to include the ability to select more than one cell and assign a cell mark.
Specifically I'm thinking of being able to select across columns or Control click say... every 4th frame in one column... then assign a cell mark which would propogate to all the cells that are selected.

I note that currently if we select more than one cell and apply a cell mark. only the first cell gets the mark.

beeheemooth:
It looks like it is not possible to pre-place markers in the cells of empty Xsheet columns if they are not displayed in the Timeline. Is this a bug?

konero:
I've been using this since yesterday, it's really nice! I tested exporting/importing back and forth between OpenToonz, Toei Animation Digital Exposure Sheet viewer and Clip Studio Paint with no problems.

shun-iwasawa:
Currently the mark does not move according to cell operation (e.g. stretching cells by using the handle). I'll modify it in the future development, referring to the implementation of the Image Marks in TVP for consistency.

@RodneyBaker 
> Before proceding I want to ask if it would be possible to include the ability to select more than one cell and assign a cell mark to all cells selected.
> Specifically I'm thinking of being able to select across columns or Control click say... every 4th frame in one column... then assign a cell mark which would propogate to all the cells that are selected.

Currently the mark is set to a cell where you right-click, not to all selected cells. This is because I thought that the mark will be used for marking a single cell , instead of certain range of cells. 
The cell selection is defined as a rectangular region and does not allow to select discrete cells. Applying marks to cells in specified interval in the selection would be a good idea. I'll consider on it for the future development.

@beeheemooth 
> It looks like it is not possible to pre-place markers in the cells of empty Xsheet columns if they are not displayed in the Timeline. Is this a bug?

It's not a bug. Currently you can't set marks to empty columns because there is no column data where the marks information is registered.

RodneyBaker:
@shun-iwasawa Just to clarify, I'm not seeking a method to set a range but rather just wondering if more than one mark could be set at the same time.  As an example:

image

Here in the scene file we can see the cell mark ID has three cells assigned.
Each one would need to be created separately.

In this example however I've manually added the 8 (via text editor) to flag cell 7 with a white cell mark.
If setting a large number of cell marks at known intervals I might resort to this manual text editing rather than individually selecting cells and assigning them to the cell mark.

All this for future consideration of course.  
I have not tried the cell marks in other programs (such as @konero refers to) but must assume they also must be set individually one at at time.

RodneyBaker:
A note of useful workflow here...

If we want to clear a column of cell marks we can use what @beeheemooth describes.
We move all exposures out of a column and the cell marks automatically disappear (due to the fact as @shun-iwasawa states there is no cell data in that column any more).

If on the other hand we wish to preserve the cell marks at their current location we want to make sure we keep at least one exposed frame in the column.  This will retain the cell marks even in empty cells.

A word of caution however, any column that does not have an exposure remaining in it will no longer retain its cell marks.
Undo will not restore the cell marks once they have been dropped.

beeheemooth:
@RodneyBaker Markers also disappear if two TLV levels with markers are merged.

RodneyBaker:
Posting intent to merge.
If anyone has a reason to hold off on the initialization of the core Cell Mark feature please make that interest known.

RodneyBaker:
Thank you @shun-iwasawa for the new Cell Mark feature!

DarrenTAnims:
I've just gotten round to looking at this (I know. I'm late :-) ), but this looks really useful @shun-iwasawa. I'm asked about marking frames a lot!

But I'd like to add my voice to the feature to move the markers with the drawings. I see them being used to mark a key drawing, then as you add your breakdowns/inbetweens, the mark moves with the key, so you can still see where the key is. Also, I'd expect them to move with the insert/remove frame. This would really finish them off for me (and the folks I speak with).
Thanks again,
Darren

PR #4182, opened 2021-12-10
Enable to Cleanup Without Line Processing

shun-iwasawa:
This PR modifies the cleanup feature as follows:
- When the `Line Processing` is set to `None` , now OT will take into account all of the other cleanup settings (autocenter, rotate, flip and camera settings)  for cleanup. Before this PR only the autocenter settings had been used.
- When the `Line Processing` is set to `None` , now user can choose file format for saving cleaned up images. (in either of tif, png, jpg or tga)
- Separate the `Brightness / Contrast` values between in the `Greyscale` and in the `Color` line processing modes. This change was requested by ink & paint staffs in our studio in order to prevent inputting wrong values by mistake.

### DIFF
--- a/toonz/sources/include/toonz/cleanupparameters.h
+++ b/toonz/sources/include/toonz/cleanupparameters.h
@@ -160,6 +160,12 @@ public:
   /*--- オフセットを軸ごとにロックする ---*/
   bool m_offx_lock, m_offy_lock;
 
+  // hold brightness and contrast values for each line processing modes (grey
+  // and color).
+  double m_altBrightness, m_altContrast;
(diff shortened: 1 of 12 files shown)


RodneyBaker:
As I have not been a really big user of Opentoonz clean up process I'm not sure if I trust myself to thoroughly test this PR but from what I'm seeing thus far... this is looking very good.  I may even have to start using Cleanup in my regular workflow.  :)

RodneyBaker:
Thank you @shun-iwasawa for this update to the clean up process.
Very much appreciated.
LGTM

PR #4183, opened 2021-12-10
Enable To View Palette Files From the File Browser

shun-iwasawa:
This PR enables to view contents of palette files by using the `View...` command in the right-click menu of File Browser.


RodneyBaker:
LGTM
Thank you @shun-iwasawa for the enhancement to the File Browser.

PR #4185, opened 2021-12-13
Fix TVP JSON export

shun-iwasawa:
This small PR fixes the problem as follows:
- The layer order goes wrong when loading JSON data on TV Paint


RodneyBaker:

PR #4186, opened 2021-12-13
Fix FrameIds Inconsistency in Level Strip Commands

shun-iwasawa:
This PR fixes the following problem regarding the right-click menu commands of Level Strip :
#### To Reproduce
1. Prepare raster image file named `A.3.jpg` and load the image to a scene.
1. Open Level Strip.
1. Right click the frame `0003` and select `Renumber...` command.
1. Set `Start =1` and `Step = 1` then click `Renumber`. (The frame 3 is renumbered to 1)
1. Save the level. 
1. The image file name becomes `A.0001.jpg` losing zero-padding information. It should be `A.1.jpg` instead.

    Other Level Strip commands such as `Duplicate Drawing`, `Add Frames` and `Swing` have the same problem (frame numbers become inconsist in zero-padding) . This PR will resolve such problems.

### DIFF
--- a/toonz/sources/toonz/filmstripcommand.cpp
+++ b/toonz/sources/toonz/filmstripcommand.cpp
@@ -48,7 +48,8 @@
 //=============================================================================
 
 TFrameId operator+(const TFrameId &fid, int d) {
-  return TFrameId(fid.getNumber() + d, fid.getLetter());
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I could reproduce the problem following the steps you provided.

beeheemooth:
Is this one of the reasons as a result of which raster levels lost drawings and frames were displayed in red after restarting the scene?

PR #4190, opened 2021-12-16
Note level column enhancement

shun-iwasawa:
This PR will enhance the note level column as follows:
- The appearance is changed, making it look like the dialogue column in the paper xsheet. Neighbor cells containing the same text are combined to display the whole contents.
    
    

- When editing text in the cell, pressing Enter key with empty field will now repeat the previous cell, instead of creating a blank text cell. 
- When editing text with selecting multiple cells, the input text will now be applied to all selected cells, instead of only one current cell.
- Locking the column will now prevent to edit cells.
- The note column will be saved in TNZ file more compact format, the neighbor cells with the same text will be shrunk as range of cells, instead of saving each cell one by one.
    - **CAUTION:**  Due to this change, the scene saved after this PR won't be opened by the previous version of OT if the scene contains the note level column.
- In `Export Xsheet to PDF` feature, the note level column can now be exported to the dialogue column.

### DIFF
--- a/toonz/sources/toonz/exportxsheetpdf.cpp
+++ b/toonz/sources/toonz/exportxsheetpdf.cpp
@@ -23,6 +23,8 @@
 #include "toonz/tstageobject.h"
 #include "toonz/preferences.h"
 #include "toonz/toonzfolders.h"
+#include "toonz/txshsoundtextcolumn.h"
+#include "toonz/txshsoundtextlevel.h"
 
 // TnzCore includes
 #include "tsystem.h"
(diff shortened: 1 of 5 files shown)


RodneyBaker:
Very nice @shun-iwasawa 
I've been testing this in conjunction with your Text FX and the level strip in horizontal orientation.  :)

image

adamearle:
Can notes take Emoji? Would be a quick way to mark an animation for a note and the write?

adamearle:
Pop-overs may have been better than this implementation. Thinking about it now. It's a tough space to work within.

RodneyBaker:
@adamearle Good news on the emoji front, Notelevels can present emojis.
I'm pretty sure they have to be added via copy and paste BUT it can be best to copy paste entire text anyway rather than type it all in at the cell level.

image

RodneyBaker:
Its madness I tell you!

image


Using Text Iwa FX to push emojis to the canvas.

RodneyBaker:
Posting intent to merge.
The assumption being that if someone has additional updates they would like to add those can be added in subsequent PRs.

RodneyBaker:
Thank you @shun-iwasawa for the notelevel enhancement.
LGTM

PR #4192, opened 2021-12-21
OpenEXR I/O

shun-iwasawa:
This PR enables to reading / writing [EXR file format](https://www.openexr.com/).

I'm working on implementing a new feature to render the scene in floating point / linear color space which supports high-dynamic-range image data. Supporting EXR file format is one of the necessary feature for that.

Currently OT does not support any of the above, so this feature will not utilize the advantages of EXR format alone:
- On loading, channel values will be converted to nonlinear color space (gamma=2.2), then clamped to 0.0-1.0 range, then normalized to requested channel depth (0-255 or 0-65535). Values in high dynamic range will be omitted.
- On saving, each channel value will be normalized to 0.0-1.0 and converted to linear color space (gamma=2.2).

This PR will support following formats:
- Storage Types 
  - Scanline format
  - Tiled format
    - Tile format w/o LoD (save/load).
    - Tile format with LoD (load only).
- Compression Types
  - No compression
  - Run Length Encoding (RLE)
  - ZIP compression per Scanline (ZIPS)
  - ZIP compression per scanline band (ZIP)
  - PIZ-based wavelet compression (PIZ)
- Pixel Format
   - HALF FLOAT (save/load)
   - UINT (load only)
   - FLOAT (load only)
 
* Multipart format or deep image are not supported.
* Images with more than 4 channels cannot be loaded.

This PR uses [tinyexr](https://github.com/syoyo/tinyexr), tiny OpenEXR image loader/saver library by Syoyo Fujita (@syoyo). The library is released under 3-clause BSD license.

### DIFF
--- a/stuff/doc/LICENSE/LICENSE_tinyexr_openexr.txt
+++ b/stuff/doc/LICENSE/LICENSE_tinyexr_openexr.txt
@@ -1,0 +1,62 @@
+tinyexr : Tiny OpenEXR image loader/saver library
+https://github.com/syoyo/tinyexr
+
+Copyright (c) 2014 - 2021, Syoyo Fujita and many contributors.
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+    * Redistributions of source code must retain the above copyright
+      notice, this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright
+      notice, this list of conditions and the following disclaimer in the
+      documentation and/or other materials provided with the distribution.
+    * Neither the name of the Syoyo Fujita nor the
+      names of its contributors may be used to endorse or promote products
+      derived from this software without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL <COPYRIGHT HOLDER> BE LIABLE FOR ANY
+DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
+(INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
+LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
+ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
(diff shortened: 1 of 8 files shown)


RodneyBaker:
Thank you @shun-iwasawa 
This is an exciting development.

Initially I was having issues both saving out (rendering) exr format as well as importing (using some exrs from external programs) but I am now able to load some files.  It seems that those that were problematic in loading must have had more than 4 channels.

I'm still not sure why I haven't been able to render out to exr format.

BUT... this is very early in testing.

Given that this is the introduction of EXR to Opentoonz perhaps it might be appropriate to name this PR "Initial OpenEXR I/O" to emphasize the limited support for this introduction to the format?  

Having said all of this... I am enjoying the process of using EXR files inside of Opentoonz.
This is a really unexpected and appreciated addition.  :)

I'll post more as soon as I get a chance to explore more deeply.

EXRtesting

(This gif animation doesn't have anything leveraging the EXR format other than the fact the content is all EXR files working inside of Opentoonz.  That alone is great stuff!)

RodneyBaker:
@shun-iwasawa 
After attaching FFMPEG to Opentoonz now I can render to EXR but if attempting to render a sequence I get a crash.
I seem to be able to render one image at a time however.

My initial feel is that ffmpeg doesn't have anything to do with this... and I don't see anything ffmpeg related in the code.  Perhaps it is only the sequence of images that is problematic.

syoyo:
@RodneyBaker @shun-iwasawa Nice work! BTW you may also have an interest in TinyDNG : https://github.com/syoyo/tinydng 

It can load/save DNG files(but no built-in RAW develop feature(e.g. demosaicing))

shun-iwasawa:
Thank you @RodneyBaker for testing this. I'll check the crash on sequential rendering.

Hello @syoyo san, thank you very much for sharing the great work!
As written in the above, the feature for rendering the scene in linear color space and in high dynamic range are demanded in our studio. Adding EXR format I/O was necessary for that and your library was so handy and helpful to implement it.
TinyDNG looks great as well. Thank you for the information!
By the way, I was excited to find your name as I once attended (as an audience, not as a speaker) an event hosted by you, "SBR" - SIGGRAPH ronBun implementation Race - when I was in graduate school. :-)

RodneyBaker:
@shun-iwasawa Is it possible there is a DLL file related to EXR missing in the github actions build?

I note that I also get a crash when I attempt to convert.
The option for EXR appears in the conversion list and a zero byte file is created just prior to the crash.

gab3d:
@shun-iwasawa great addition! thanks!

> On saving, each channel value will be normalized to 0.0-1.0 and converted to linear color space (gamma=2.2).

maybe I didn't followed your explanation well, but shouldn't it be (gamma=1.0) at this stage?

shun-iwasawa:
I think I could fix the crash bug.

@gab3d 

> maybe I didn't followed your explanation well, but shouldn't it be (gamma=1.0) at this stage?

Yes, you can say that the linear color space has 1.0 gamma.
Here I just meant the gamma value specifying a curve used for color space conversion from non-linear to linear.

RodneyBaker:
You got it @shun-iwasawa 
No more crashing in any of the cases I observed before (both rendering and conversion).  :)

Posting intent to merge so any feedback or request for delay to accomodate further testing will be appreciated.

RodneyBaker:
This is a very welcome present @shun-iwasawa 
I hope other developers with interest in the OpenEXR format will consider pressing into the other features of OpenEXR 3.0 and perhaps even assist @syoyo (where applicable) in the effort to maintain tinyEXR upstream.  Thank you!
LGTM

PR #4194, opened 2021-12-22
Fix pressing Shift key loses focus when renaming cell

shun-iwasawa:

#### To reproduce ####
- Activate 'Preferences > Xsheet > Enable to Input Cells without Double Clicking' .
- Load or create some level in the xsheet.
- Click a cell in the column. (Renaming field opens.)
- Try to type '1A' (type "1" then "Shift+A") .

The field loses focus when typing the Shift key and the input is unexpectedly canceled.


RodneyBaker:

PR #4195, opened 2021-12-24
Project Settings-aware Frame Number Field Width

shun-iwasawa:
This PR modifies the frame number field in the Pencil Test popup.
It is now aware of the file path properties of the project settings and changes its width according to the maximum letter count for suffix.

framenumberfield


RodneyBaker:
LGTM

PR #4196, opened 2021-12-24
Xsheet PDF Export: Tickmark breaks continuous line

shun-iwasawa:
This PR fixes the `Export Xsheet to PDF` feature as follows:

When the `Continuous Line` option is set to `More Than 3 Continuous Cells` , tick marks (`Inbetween mark` and `Reverse sheet mark`) will break the continuous line as if the cell number is input.

tickmark


RodneyBaker:
Thank you @shun-iwasawa
LGTM

PR #4197, opened 2021-12-24
Enhance Flipbook playback

shun-iwasawa:
This PR overhauls `PlaybackExecutor` in order to achieve smoother playback in precise fps.

Compared to the conventional one, changes are as follows:

- `PlaybackExecutor` inserts sleep between frames to achieve specified fps. In the conventional one, sleep length had been calculated NOT from the current frame, but from the statistics of previous frames. It can be inaccurate since the computational time for each frame varies. For instance, playing a footage with 1440x779 pixels takes computational time from 15 to 30 miliseconds per frame in my enviroment. Thus the playback had lacked smoothness.
    In this PR, the sleep will be inserted at the end of the `paintGL()` function and the sleep length is calculated based on the current frame. The time for swapping OpenGL buffers (which is at just before displaying frame on the screen) is only estimated from the statistics of previous frames.
- The timer was in miliseconds order in the conventional one. This PR computes them in nanoseconds order.
- This may not be true in every environment, but I found that calling `repaint()` does not call `paintGL()` immediately. This messed up the timing of drawing. In this PR I replaced `repaint()` by `update()` and call `qApp->processEvents()` in order to make sure to redraw the frame immediately.

I'm aware of a problem with this PR, that dragging and moving the Flipbook window while playing causes "wobbling" of the window for unknown reasons.

### DIFF
--- a/toonz/sources/include/toonzqt/flipconsole.h
+++ b/toonz/sources/include/toonzqt/flipconsole.h
@@ -19,6 +19,7 @@
 #include "toonz/imagepainter.h"
 #include "tstopwatch.h"
 #include <QThread>
+#include <QElapsedTimer>
 
 #undef DVAPI
 #undef DVVAR
@@ -52,6 +53,7 @@ class PlaybackExecutor final : public QThread {
 
   int m_fps;
   bool m_abort;
+  QElapsedTimer m_timer;
 
 public:
   PlaybackExecutor();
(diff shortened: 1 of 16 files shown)


RodneyBaker:
I see what you refer to about the wobbling when moving the flipbook window while it is playing.  Very odd.

Looking good!

RodneyBaker:
This doesn't apply to this PR but will mention it as it came to my mind some time back.
Currently for FPS we have both positive and negative integers (whole numbers) to choose from.
This is very useful for standard playback as well as playing back in reverse.

I am curious if we might consider an option to allow decimal/float values to achieve even greater fidelity of playback.
As an example, let us imagine we want to spend a good amount of time reviewing every frame of a sequence very carefully before it moves on to the next frame.  We might be looking for flaws but also might be reading text or visually tracking some changing aspect in the frame.  Allowing values such as .25 or .050 would give very fine control over playback speed.  This would allow finer values when speeding up playback as well but the primary purpose I can see is to allow for fractional values of playback less than 1 without stopping or reversing the playback.

I am not sure how complex this change would be but it seems the Qt timer (and nanoseconds) will allow for this to the greatest extent local hardware will support.

RodneyBaker:
@shun-iwasawa I will assume the desire is for this PR to proceed despite the wobbling which can be resolved in subsequent PR.
Posting intent to merge otherwise.

RodneyBaker:
Thank you @shun-iwasawa for the Flipbook enhancement!
LGTM

Aside:  Some recently posted a question related to the flipbook.
They were wondering if there is a direct way to clear a flipbook.
I don't see any direct way to do that.

PR #4202, opened 2021-12-28
Fix Flipbook wobbling when dragging while playing

shun-iwasawa:
This PR fixes the known problem after introducing #4197 : Flipbook and Viewer wobble when dragging floating window while playing.


RodneyBaker:

PR #4205, opened 2022-01-03
Multi-Thread FFMPEG and responsive finalizing window

justburner:
This PR allows to enable Multi-Thread FFMPEG rendering and have finalizing window responsive.

Multi-Thread FFMPEG will be **disabled by default** and must be enabled in `Preferences... > Import/Export`.
Enabling multi-thread rendering will render significantly faster but a random crash might occur, it's even more unstable when plastic tool is involved... in order to use multi-thread rendering the option `Dedicated CPUs` must be either `Half` or `All`.
Note: Is possible to Multi-Thread render in Tasks regardless of this patch.

Anyway in my opinion the speed gain is worth the risk as long the scene is always saved before rendering... I've got some users reporting gains between 6x to 8x with Dedicated CPU as All, also since the option can be enabled it will encourage developers to make multi-threading more stable.

Last but not least, the PR also makes the finalizing window responsive to the user instead of freezing OT window completely.
Progress bar will also perform a busy indicator while waiting for FFMPEG to finish.

### DIFF
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -4,6 +4,8 @@
 #include "tsound.h"
 
 #include <QProcess>
+#include <QEventLoop>
+#include <QTimer>
 #include <QDir>
 #include <QtGui/QImage>
 #include <QRegExp>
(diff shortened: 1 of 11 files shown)


RodneyBaker:
LGTM @justburner 
Thank you for the update to FFMPEG rendering!

RodneyBaker:
@shun-iwasawa The nighttly build for this PR appears to have failed.

PR #4207, opened 2022-01-07
Fix margin of Directional Blur Fx Iwa

### LINKED ISSUE #4206: Tile Iwa Fx + Directional Blur Iwa fails to fill gap at camera boundary
**Describe the bug**
Tile Iwa Fx can be utilized to fill gap at camera boundary when blurring the image. However it does not work properly in the following case.

**To Reproduce**
1. Create a new Toonz Raster Level with the same size as the current camera. Draw something touching edge of the image boundary like as follows:
    
2. Build the Fx schematic like as follows;
   
    with the fx settings as follows:
    - Tile Fx Iwa
        
    -  Directional Blur Iwa
        
    Note that the Tile Fx Iwa is used for filling gap caused by blurring the pixels.
3. Preview the scene. Despite the existence of the Tile Fx Iwa, gap appears at the boundary.
    
4. Oddly enough, the gap disappears when you turn on the `Bidirectional` option of the Directional Blur Fx Iwa and then turn it off again.
   

**Expected behavior**
The gap should always be filled.

**Device Information:**
 - OS: Win10
 - OpenToonz Version: latest

shun-iwasawa:
This PR fixes #4206 , modifying the bounding box of computation when the `Bidirectional` option is off. Now the Fx compute in the bounding box with the same size regardless of the `Bidirectional` option.

### DIFF
--- a/toonz/sources/stdfx/iwa_directionalblurfx.cpp
+++ b/toonz/sources/stdfx/iwa_directionalblurfx.cpp
@@ -143,38 +143,12 @@ void Iwa_DirectionalBlurFx::doCompute(TTile &tile, double frame,
              ,
              TDimensionD(/*- Render画像上(Pixel単位)のサイズ  -*/
                          tile.getRaster()->getLx(), tile.getRaster()->getLy()));
-  /*- 上下左右のマージンを得る -*/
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #4208, opened 2022-01-07
Tile Fx Iwa : a new "Image Size" option for the Input Size parameter

shun-iwasawa:
This PR will add a new option `Image Size` for the `Input Size` parameter. The new option will repeat image at the interval of the  image size of the input. Please note that it takes effect when you input Toonz Raster or Raster Levels. Other level types will return the same size as the `Bounding box` option.

This option is demanded for utilizing the Tile Fx Iwa for filling gap at camera boundary after blurring the image.

#### Example

Input image

The result when the `Image Size` is set to `Bounding Box` (existing option)

The result when the `Image Size` is set to `Image Size` (new option)


RodneyBaker:
Works as advertised!
LGTM

PR #4209, opened 2022-01-07
A new option for the Fractal Noise Fx Iwa : Conical Transform

shun-iwasawa:
This PR adds a new option to the Fractal Noise Fx Iwa: Conical Transform. It can generate radially propagating noise pattern. 

  
**Left** : Conventional noise pattern    **Right**:  (new) Conical noise pattern

The concept of this option is as follows: The conventional, regular noise pattern is sampled from flat plane. OTOH the conical noise pattern is sampled from the cone shape in the +1 dimensional noise space, with the "conical evolution" axis.

### New Parameters
- **Do Conical Transform** : enables conical transform of the noise. Unchecking this option will generate conventional noise.
- **Conical Angle** : specifies the slope angle of the sampling cone. (in degree)
    conical_angle
- **Camera Fov** : specifies the vertical field of view of the camera. (in degree) 
    conical_fov
- **Scale Depth** : specifies the exponential scale of the noise pattern along direction of the `conical evolution` axis.
    conical_depthscale

- **Conical Evolution** : specifies the noise evolution in the `conical evolution` axis.

Please note that when the Conical Transform is active, the `Offset Turbulence` parameter acts completely different: It offsets the peak position of the cone from the center of the camera, like this:

fractal_noise_offset

Please be forewarned that the parameter may altered before releasing the next version according to the feedback from our studio.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1450,6 +1450,11 @@
   <item>"STD_iwa_FractalNoiseFx.cycleEvolution"	"Cycle Evolution"	</item>
   <item>"STD_iwa_FractalNoiseFx.cycleEvolutionRange"	"Cycle (in Evolution)"	</item>
   <item>"STD_iwa_FractalNoiseFx.dynamicIntensity"	"Dynamic Intensity"	</item>
+  <item>"STD_iwa_FractalNoiseFx.doConical"	"Do Conical Transform"	</item>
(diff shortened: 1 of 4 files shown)


RodneyBaker:
Thank you @shun-iwasawa for the new Fractal Noise option.
That's sure to be useful.

PR #4156, opened 2021-11-21
Fix various typos

luzpaz:
Found via `codespell -q 3 -S *.ts,thirdparty, -L appy,ba,chunck,datas,forse,inbetween,inly,inout,pevent,possibile,upto`

PR #4213, opened 2022-01-10
Fix various source typos

luzpaz:
Follow-up to #4156

### DIFF
--- a/toonz/sources/common/tcore/tstopwatch.cpp
+++ b/toonz/sources/common/tcore/tstopwatch.cpp
@@ -24,8 +24,8 @@ extern "C" long sysconf(int);
 #endif
 #endif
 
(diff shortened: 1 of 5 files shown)


RodneyBaker:
Thanks @luzpaz 
It appears you got all instances of these typos.
It will still be proper to close the loop and validate this on a 32bit build running up against the 2GB limit.
But.. LGTM

Aside:  As you are into fixing typos perhaps you might be willing to assist with #4072?
If the legit typos identified in that PR's review are fixed we can close that PR.
I'm sure @alexmyczko would be open to the assist.

luzpaz:
@RodneyBaker I've created #4215 in favor of #4072. I've also left comments that suggest not to merge #4072 for specific reasons.

PR #4214, opened 2022-01-11
"Implicit hold" move by Shift+dragging cells

### LINKED ISSUE #4078: A way to slide a cell over its neighbouring hold cells, while moving it.
@shun-iwasawa @RodneyBaker 

While animating frame by frame digitally in OT, an usual procedure is to first draw key frames and then start drawing inbetweens in the intermediate space between the already drawn key frames.
As currently the default method of creation in the Xsheet automatically places "holding cells" inbetween exposed cells, it gets pretty uncomfortable when you need to move one of those newly drawn inbetweens to expose it a little early/late in the Xsheet to adjust timing.

An elegant solution to this problem could be to have a "Slide" function, that allowed to automatically adjust nearby "holding cells", while moving the first cell in which a drawing is exposed.

Example:
11**2**222333  (slide 1 frame to the left)
111**2**22333  (original)
1111**2**2333  (slide 1 frame to the right)

Activation of this new mode could simply be by holding some modifier keys (Ctrl + Shift) prior to dragging a cell.

What do you think about it?

shun-iwasawa:
This PR resolves #4078 .

When you Shift + dragging cells within the same column, cells will be moved as if it is implicitly held.
Please note that the conventional behavior of Shift + dragging cells can still be performed by Shift + Alt + dragging.

### DIFF
--- a/toonz/sources/toonz/xshcellmover.cpp
+++ b/toonz/sources/toonz/xshcellmover.cpp
@@ -89,7 +89,7 @@ TXsheet *CellsMover::getXsheet() const {
 
 void CellsMover::getCells(std::vector<TXshCell> &cells, int r, int c) const {
   for (int i = 0; i < (int)cells.size(); i++) cells[i] = TXshCell();
-  TXsheet *xsh                                         = getXsheet();
+  TXsheet *xsh = getXsheet();
   for (int i = 0; i < m_colCount; i++) {
     TXshColumn *column = xsh->getColumn(c + i);
     if (column && column->isLocked()) continue;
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Very nice @shun-iwasawa.
I will delay merging a little while to allow @gab3d time to test if he wishes to do so.
Over to you @gab3d.

manongjohn:
@shun-iwasawa Just something to note, if you want to do something about it with this PR...

Normally if you hold just CTRL and click-drag the dragbar on an unselected cell and move...it selects all cells linked to the dragbar + keys , and moves according to the behavior of CTRL cell move (copy), ALT+drag overwrites cells and SHIFT+drag shifts/inserts

If you do ALT+CTRL, it copies and overwrites cells as one would expect.

However, if you do SHIFT+CTRL and click the dragbar, it immediately does an insert+copy of the selection without keys and without an UNDO action.  Here is an animated gif of me using CTRL + click by itself first, a small drag, resetting then doing SHIFT+CTRL+click:

shift_ctrl_drag_click_issue

gab3d:
@shun-iwasawa @RodneyBaker 
It works great. Thank you so much! ❤️

shun-iwasawa:
@manongjohn Wow. It seems to be a bug. Thank you for letting me know!

RodneyBaker:
Will merge.
Thanks @shun-iwasawa for the new capability.

manongjohn:
@shun-iwasawa I found an odd shift+drag bug using this artifact

shift_drag_issue

In the animate gif, when I Shift+drag Frames 1-3 to the right and completely covered the last 3 (which happen to be the same, Scene Frame 3 shows as a Frame 2 Hold when it should be Frame 3...when I drag it back, now Frame 6 shows as a hold instead of the original Frame 3

Even worse is this gif, it deletes the cells when I drag it past the last 3

shift_drag_issue_2

gab3d:
@shun-iwasawa 
I've just found some weird behavior when Shift+dragging the first/last cells of a contiguous chunk of cells.
1. when shift+dragging the last cell and there's nothing after it, instead of extending the level it just moves that cell.
2. when there is a gap and then another chunk of cells, the last cell "glues" to them while moving.
3. if you shift+drag past the last cell of that other chunk of cells, then it acts again as if moving the last cell of the chunk.
4. when shift+dragging the first cell of a contiguous chunk of cells, it appears to behave as expected, unless there are other discontiguous cells before the dragged cell.


https://user-images.githubusercontent.com/18104091/153699714-b488db30-e137-4a6d-8eb5-75542117dd21.mp4

shun-iwasawa:
@gab3d 
As long as I can see in the movie, it works as expected.
In the implicit hold, only the first cells of contiguous chunks (with numbers) matter. A chunk of blank cells is also handled as the "contiguous chunk". The first cell of the blank (with "X" mark) also act like the numbered cells.
The "weird behaviors" you pointed seems to be due to the dragging cells were switched places with the X-marked blank cell.

PR #4216, opened 2022-01-12
Fix undo of moving cells with Shift+Ctrl

shun-iwasawa:
This PR will fix the problem reported in [this comment](https://github.com/opentoonz/opentoonz/pull/4214#issuecomment-1010285445) by @manongjohn .
When clicking the side bar holding Shift+Ctrl (i.e. Copy+Insert options), it seems to be not a problem that the copy of the selected cells are immediately duplicated. The problem is that such duplication is not undoable.


RodneyBaker:

PR #4217, opened 2022-01-13
Zoom in/out and fit floating panel geometry commands

shun-iwasawa:
This PR adds new commands `Zoom In/Out And Fit Floating Panel` , which will resize the top-level floating panel.
- If the top level floating panel is Flipbook or Color Model, these commands will zoom in/out the content and then fit the window geometry to it.
- If the top level floating panel is another type, these commands will just resize the window.
- If there is no floating panel (or the top level window is not a floating panel), these commands will do nothing.

This feature is just like the same feature in Photoshop. The default shortcuts are `Ctrl+Alt++` and `Ctrl+Alt+-` respectively (which are also consistent with Photoshop).

### DIFF
--- a/toonz/sources/toonz/flipbook.cpp
+++ b/toonz/sources/toonz/flipbook.cpp
@@ -2117,7 +2117,7 @@ QPointF newViewerGeomCenter(
 
 //-----------------------------------------------------------------------------
 
-void FlipBook::onDoubleClick(QMouseEvent *me) {
+void FlipBook::adaptGeometryToCurrentSize() {
   TImageP img(m_imageViewer->getImage());
   if (!img) return;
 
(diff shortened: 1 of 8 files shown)


RodneyBaker:
Even more reason to use Flipbook!

The main thing I see as being slighly problematic with this connection to floating panel is that the floating panel may not have the active focus.  In those cases the use of shortcut keys will do nothing (which is ideal of course).  My concern is that an additional step often has to be taken to get the focus and then settle into use of the zoom in/out options.  This might be more of a workflow issue where I'm just not using the ideal workflow but I make mention of this because it has caught me several times and I've wondered why use of shortcut wasn't working.  Clicking on the flipbook panel and giving it focus resolved the problem.

I really like using the Snapshot/Comparison tool in conjunction with this.

RodneyBaker:
Thank you @shun-iwasawa for the new zoom option.
LGTM

PR #4218, opened 2022-01-13
Fix backtab key to work for moving styles shortcut scope

shun-iwasawa:
- When the preferences option `Use Numpad and Tab keys for Switching Styles` is ON, pressing backtab (Shift+tab) key fails to move the styles shortcut scope upward.


RodneyBaker:
Thanks for the fix @shun-iwasawa 
LGTM

PR #4222, opened 2022-01-16
RGB sliders behaving like CSP/Krita, Hex editbox and various bugfixes

justburner:
This PR is mainly to give a new look at RGB sliders and color preview.

Image

- Changed background colors in RGB slides to behave like CSP/Krita.
- Changed handle to look more modern, similar to Tahoma2D (Thanks @konero for the SVGs).
- Fixed weird behaviors of the sliders by removing QSlider hack and implementing custom handling.
- Changed how color preview looks, now MyBrush icons are centered with current color as the background. No more ugly stretching!
- Added Hex edit-box to quickly get and set colors from text, must be enabled by clicking the bottom-right 3-lines icon, the placement of this widget is different from Tahoma2D.
- Hex edit-box accepts typing HTML colors and more! It can be configured with `colornames.txt` files.
- Clicking on the current style color preview will apply color.
- Bugfixes in regard to style index handling.

**Accepted Hex input formats:**
- `#RRGGBBAA` - Tahoma2D supports this
- `#RRGGBB` - Tahoma2D supports this
- `#RGBA` - Shorthand hex color
- `#RGB` - Shorthand hex color
- `#VV` - Non-standard, useful for gradients
- `#V` - Non-standard, useful for gradients

It will outputs `#RRGGBB` on opaque color or `#RRGGBBAA`.

**Custom color names tables:**
The default table is stored in `<OT_STUFF>/configs/colornames.txt` and is a XML format file.
If you want to add your own names is recommended to write them to user profile folder so it won't be affected by any future updates: `<OT_STUFF>/profiles/layouts/settings.<USERNAME>/colornames.txt`

Important: Name entries in file must have no spaces!
Anyway spaces in the Hex edit-box are ignored so you can write `Sky Blue` to mention `skyblue` color.
Unicodes are supported for names, make sure you save the file as "UTF-8" only, this can be tested by using `レッド` resulting in red color.
OpenToonz must be restarted if changes to the files are made while the app it's running.

Example `colornames.txt` in user settings folder:
```XML
<colors>
  <!-- color name is not case-sensitive and must have no spaces, items on bottom takes priority -->
  <color name="test">#808080</color>
  <color name="anothertest">#C0C0C0</color>
</colors>
```
User will be able to use `test`, `anothertest` and even `another test` in the Hex edit-box.
Sadly there's no editor build-in in OpenToonz to manage user entries.

**Performance implications:**
[...]

PR #4224, opened 2022-01-18
RGB sliders behaving like CSP/Krita, Hex editbox and various bugfixes by justburner (with modifications)

shun-iwasawa:
This PR is based on #4222 by @justburner , with some modifications as follows:

- Enabled to make the color sliders to be conventional one by choosing option from hamburger menu in the Style Editor.
    styleeditor2
    
    - The appearance setting will be stored in the user's .env file and is shared by all Style Editor panels.
    - Default value is `Relative colored + Triangle handle` (i.e. the new one)
- The style preview chips are now with double height. When the style with color parameters (sub colors) is selected, the color parameter selector becomes visible and the style preview chips shrink to the conventional height.
  styleeditor4

-  Related to the above changes, the MyPaint style chip in the Palette Viewer is modified so that the style icon is shown without stretching.
   styleeditor5


shun-iwasawa:
@RodneyBaker 
Please do not use `Squash and merge` when merging this PR to make the commits separated in the history. Thanks!

RodneyBaker:
LGTM!

WIll give @justburner a chance to wake up and double check to see if anything is missing.

RodneyBaker:
If someone could double check drawing with standard raster levels.
I seem to be having an issue where no matter what color on the palette I use I get a dark almost black color.
I've check to make sure I don't have any of the Ink and Paint checks toggled on.

justburner:
> 

I see you picked the Tahoma2D approach by moving the icons to the left. lol
Overall all the changes looks good to me. 👍 

> 

This is one of those settings the user pick own preference once and forget about it, so, really didn't expected this approach. OwO
At least it makes it easier to find it so that's good. 

> 

Nice approach, i like it. 😄
Anyway I'm surprised you didn't stretched the "Auto" & "Apply" buttons vertically similar to how they were in OpenToonz 1.0

> 

This was a oversight from my PR, really cool addition. 💯

gab3d:
@shun-iwasawa  nice work!

- Could there be a little less space destined to the slider arrows? especially in lower res screens they seem unnecessarily big.
- Perhaps if HEX characters could share space with the Auto Apply buttons, the panel could be functional while being narrower than with the current layout. (it also could have the benefit of having all the values aligned in the panel).
- if the two lower right buttons get stacked, we gain space for having the chips bigger also.

imagen
imagen

konero:
@shun-iwasawa I'm getting a crash by doing the following:
1. Go to the 'Drawing' room
2. Try resize the Style Editor to the smallest it allows you to go by dragging the dock handle

If I let go of the mouse button quickly I can avoid a crash, but if I keep the mouse button held as if making precise adjustment I can reproduce the crash over and over.

Otherwise, this is nicely done! The sliders feel much nicer to use now.

shun-iwasawa:
@gab3d 

> Could there be a little less space destined to the slider arrows? especially in lower res screens they seem unnecessarily big.

At least for the first request you can customize the button size by applying the additional style sheet like as follows:

```
#colorSliderAddButton,
#colorSliderSubButton {
  width: 14;
  padding: -2;
}
```
image

shun-iwasawa:
@konero Thank you for reporting the crash, I'll check it tomorrow!

toonz/sources/toonzqt/styleeditor.cpp
```diff
@@ -1020,79 +1258,68 @@ void ColorSlider::paintEvent(QPaintEvent *event) {
...
+    w -= 3;
+  } else {
+    x += s_chandle_size / 2;
+    w -= s_chandle_size;
+    h -= 3;
+    if (isLineHandle) {
+      y += 1;
+      h -= 2;
+    }
+  }
```
justburner, line 1277: Seems i should had tested both orientations, not just the default one alone.
This is one way to solve the crash:
```suggestion
  if (w < 2) w = 2;
  if (h < 2) h = 2;
  
```
it seems `makeLinearShading` doesn't like when size is 1 or lower.

shun-iwasawa: Confirmed that the suggested change resolves the crash. Thank you @justburner !

RodneyBaker:
@shun-iwasawa @justburner  Just doing final checks but all is looking good.
Thank you for the update.
Also thanks to @konero and @gab3d for testing and feedback.

LGTM

PR #4225, opened 2022-01-19
Fix crash after starting a pen stroke from inactive window

### LINKED ISSUE #4184: Crash after a moment as inactive window.
**Describe the bug**
I leave Opentoonz as an inactive window, for example while I consult a YouTube reference video. If after a brief moment I press the stylus in the Opentoonz interface it closes automatically. It is systematic, and I can reproduce it as many times as I want. Also, it crashes if I use the stylus while it is launching.

**To Reproduce**
1. Open Opentoonz, and draw something.
2. Activate other Window, for example Youtube, and wait a moment.
3. Activate Opetoonz directly with stylus.

**Expected behavior**
Don't crash.

**Video**
There is a video on #3941 that I recorded months ago **with first 1.5 release** that shows exactly the same problem that occurs with **nightly builds**.

**Device Information:**
 - OS: Microsoft Windows 10 Home.
 - OpenToonz Version: Nightly (2021-12-10)]
 - GPU: NVIDIA GeForce GTX 1050
 - GPU Driver Version: 399.01
 - DDR4 16Gb
 - HUION KAMVAS GT-191

-------
Thanks a lot.

justburner:
Fixes #4184 

Got report from @DarrenTAnims that this patch stop crashes.

**Result of crash:**
Always calling `processEvents` together with `emit` could cause infinite recursive calls resulting in a Stack overflow.

### DIFF
--- a/toonz/sources/toonz/tapp.cpp
+++ b/toonz/sources/toonz/tapp.cpp
@@ -748,7 +748,7 @@ bool TApp::eventFilter(QObject *watched, QEvent *e) {
     // if the user is painting very quickly with the pen, a number of events
     // could be still in the queue
     // the must be processed as tabled events (not mouse events)
-    qApp->processEvents();
+    if (m_isPenCloseToTablet) qApp->processEvents();
 
     m_isPenCloseToTablet = false;
     emit tabletLeft();


RodneyBaker:
LGTM

DrawToonZ:
Hello. I opened thread #4184. I don't know if I should answer there, since it's closed, or I can do it here.

First of all I wanted to say **thanks for the help** and specially to @justburner for **his effort** and the patch.

In the last hour I uninstalled all traces of OpenToonz and deleted the "OpenToonz stuff" folder, as recommended when version 1.5 was released. I then installed the **nightly 2022-01-20 01:18:25 UTC** to check if the issue was removed.

Unfortunately, in my case I checked on the third try that **it still happens**. But now **it doesn't happen 100% of the time**, nor as easily as before (it's not enough to just leave OpenToonz inactive for a few seconds and then touch the screen with the stylus… or yes, it depends), so **it's an improvement**. Of several attempts, it closes sometimes, but it has even closed a couple of times **without even touching** the screen with the stylus, but hovering over it. I'm trying to find a pattern, because surely there is a condition that is met every time it closes.

I think if I wait a little longer, for example by consulting the internet browser tabs for a few moments, when I return with the OpenToonz stylus it is more likely to close. I also think it's more likely to close when contacted on the canvas, but not in some empty area of ​​the interface. However, the times that OpenToonz crashed with the stylus just a few millimeters from the screen (without contact), the cursor was over an empty area below the timeline. All very confusing.

I've been really doing a lot of testing, opening the program over and over again when it closed, but since it doesn't always close or do exactly the same thing, I'm sorry I can't provide more useful information at the moment.

Again, **thank you very much**, because it must be very difficult to know what causes that very specific problem. I hope that OpenToonz will get better and better, and above all, get rid of these crashes.

Greetings.

justburner:
It seems the issue you're facing goes way deeper than the one me and @DarrenTAnims were experiencing. 😞 

I recommend you to try Tahoma2D, It's a fork of OpenToonz with some beginner friendly features.

Since it contains bugfixes that aren't in OT (and vise-versa) maybe it will work better for you.

Even if Tahoma2D still does the same, that one have CrashRpt that creates a mini dump at the moment of a crash, extremely valuable for bug tracking, something that OpenToonz doesn't support yet.

DrawToonZ:
Hi @justburner. Yes, of course I know Tahoma2D, and I use it. I believe that its version 1.2 is, at the moment, the most stable version of OpenToonz/Tahoma2D that I have tried, and that problem does not happen to me. Of course, it also fails from time to time for other reasons. I didn't know about CrashRpt, and it seems interesting.

In any case, I'm glad it's not a widespread problem, from what I see. Maybe if I have another computer at some point it won't happen to me.

The important thing is that OpenToonz/Tahoma2D is improving a lot, largely thanks to people like you. I love a lot of its tools, and it's becoming much more stable than it was when it started, but it can still play tricks on us. I think its Achilles heel is stability, which doesn't do justice to the quantity and quality of tools and features it offers. Let's see how the future version 1.6 goes.

Thanks for your help.

PR #4227, opened 2022-01-21
Fix implicit hold cell dragging

shun-iwasawa:
This PR fixes bug reported by @manongjohn in [this comment](https://github.com/opentoonz/opentoonz/pull/4214#issuecomment-1017104945).

Now the original cell arrangement will be stored when start dragging so that the modification on-the-go won't affect the subsequent results.


RodneyBaker:
LGTM

PR #4232, opened 2022-01-23
Hex color names editor

justburner:
Continuation of #4222 and #4224 for a feature that was left incomplete.
Now the user can modify the hex editbox color names inside the program itself.

- Refactored code, now all Hex implementation have its own C++ file.
- Hex editboxes now supports auto-complete of color names.
- Editor can be started by clicking the hamburger and choosing `Hex color names...` on the very bottom.
- Only personal entries can be managed, default colornames.txt can be viewed but can't be modified inside OT.
- All personal entries can be imported and exported, when importing the user can choose to replace or merge existing entries.
- Some characters in name aren't allowed: `\` `#` `<` `>` `"` `'`

### DIFF
--- a/stuff/config/colornames.txt
+++ b/stuff/config/colornames.txt
@@ -12,6 +12,10 @@
   <color name="1">#FF0000</color>
   <color name="2">#00FF00</color>
   <color name="3">#0000FF</color>
+  <color name="4">#00FFFF</color>
+  <color name="5">#FF00FF</color>
+  <color name="6">#FFFF00</color>
+  <color name="7">#FFFFFF</color>
   <!-- HTML Color Names -->
   <color name="AliceBlue">#F0F8FF</color>
   <color name="AntiqueWhite">#FAEBD7</color>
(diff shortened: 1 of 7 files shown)


RodneyBaker:
Works great @justburner 
Really like the ability to import and export the custom colornames files.
And the separation of default (protected) versus custom (editable) color sets is a great idea.

For those wondering...
Setting a custom color to the same name as what is in the defaults will allow users to override the default.
So the default for red might be #FF0000 but we can refine that as an override. (ex:  #7A1712 aka berry)

Psoting intent to merge.

RodneyBaker:
Thank you @justburner for the addition of the Colornames Editor.
LGTM

artisteacher:
Sorry for the late feedback. Changing semesters here! This is a fantastic addition. So helpful!

gab3d:
@justburner  great job! thanks!

Could default color onscreen names be made translatable? 
I mean, that when you select (for example) Spanish for the UI, those names could be shown in Spanish too. It's not terribly important, but it would be a nice touch.

justburner:
@gab3d I don't think it will be easy since translations are kinda magically handled in source code itself with the `tr( )` macro, and of course translations needs to be added into some kind of database file (.QM) or something.
Anyway you can do one of those two things:

A. Open colornames.txt in config folder with a text editor and translate them, anyway beware that any OT update might override this file.

B.  Import colornames.txt from config folder into your user defined ones, translate them inside OT itself and export to share to others if needed, anyway using this way you have both defaults in english and your spanish ones in user-defined on auto-complete.

PR #4233, opened 2022-01-24
Stop motion play sound on capture and deciseconds resolution

justburner:
This PR improves the Stop Motion a bit by the following:

- Added option to play a click sound when an image is being captured
- Time lapse interval setting is now saved
- Interval and review times are in deciseconds resolution (1/10 of a second)
- Added `Use Numpad Shortcuts When Active` tooltip to show which and what the shortcuts do
- Now `Use Time Lapse` checkbox is a checkable groupbox.

Warning: There's a chance of ears bleed if click sound is played on headphones ;)


RodneyBaker:
Thank you @justburner for the addition to Stop Motion control panel.
LGTM

PR #4234, opened 2022-01-24
Fix conical fractal noise with offset

shun-iwasawa:
This PR fixes Fractal Noise Fx Iwa with the conical transform option + offset as follows:
- Fixed the equation for computing the intersection between the cone and the ray from the camera. Now the offset position will always points the originating point of noise regardless of the conical angle and the camera fov.
- Double-sized the available range of the offset position.


RodneyBaker:
Looks like most of the changes are beautification.  (It helps to turn off changes of white space)
Thanks much for the comments in the code.

LGTM

PR #4235, opened 2022-01-24
Tahoma2D Motion Path data compatibility

manongjohn:
I try to ensure scene compatibility between Tahoma2D and OpenToonz, however new scene tags were added for Motion Paths that causes OT to fail to load the scene if it was saved by T2D. 

The following `pegbarspline` child tags were added to T2D: `color`, `active`, `steps`, `width`, `interpolationStroke`

This PR restores Motion Path compatibility by doing the following:
- T2D Motion Path data from the scene file will be read and stored, even though OT does not use or display it.
- OT will write out T2D Motion Path data using whatever data was previously read.
  - For new paths or if no data was read, OT will default the information similarly to T2D.

NOTE: This change will make scenes containing Motion Path data saved with this fix incompatible with prior versions.

### DIFF
--- a/toonz/sources/toonzlib/tstageobjectspline.cpp
+++ b/toonz/sources/toonzlib/tstageobjectspline.cpp
@@ -125,19 +125,6 @@ TStageObjectSpline::TStageObjectSpline()
   points.push_back(TPointD(d, 0));
   points.push_back(TPointD(2.0 * d, 0));
   m_stroke = new TStroke(points);
-
-  // Needed for T2D compatibility
-  m_interpolationStroke.push_back(TPointD(-40, 0));
-  m_interpolationStroke.push_back(TPointD(-20, 0));
-  m_interpolationStroke.push_back(TPointD(-20, 0));
-  m_interpolationStroke.push_back(TPointD(0, 0));
-  m_interpolationStroke.push_back(TPointD(65, 65));
-
-  m_interpolationStroke.push_back(TPointD(935, 935));
-  m_interpolationStroke.push_back(TPointD(1000, 1000));
-  m_interpolationStroke.push_back(TPointD(1020, 1000));
-  m_interpolationStroke.push_back(TPointD(1020, 1000));
-  m_interpolationStroke.push_back(TPointD(1040, 1000));
 }
 
 //-----------------------------------------------------------------------------
@@ -171,13 +158,6 @@ TStageObjectSpline *TStageObjectSpline::clone() const {
 //-----------------------------------------------------------------------------
 
 const TStroke *TStageObjectSpline::getStroke() const { return m_stroke; }
-
-//-----------------------------------------------------------------------------
-
-// Needed for T2D compatibility
-QList<TPointD> TStageObjectSpline::getInterpolationStroke() {
-  return m_interpolationStroke;
-}
 
 //-----------------------------------------------------------------------------
 
@@ -270,8 +250,6 @@ void TStageObjectSpline::loadData(TIStream &is) {
 
 void TStageObjectSpline::saveData(TOStream &os) {
   const TStroke *stroke = getStroke();
-  QList<TPointD> interpStroke =
-      getInterpolationStroke();  // Needed for T2D compatibility
   os.child("splineId") << (int)m_id;
   if (!m_name.empty()) os.child("name") << m_name;
   os.child("isOpened") << (int)m_isOpened;
@@ -291,13 +269,15 @@ void TStageObjectSpline::saveData(TOStream &os) {
   }
   os.closeChild();
   if (m_externalDataFound) {  // Needed for T2D compatibility
-    os.openChild("interpolationStroke");
-    n = interpStroke.size();
-    os << n;
-    for (auto p : interpStroke) {
-      os << p.x << p.y;
+    n = m_interpolationStroke.size();
+    if (n) {
+      os.openChild("interpolationStroke");
+      os << n;
+      for (auto p : m_interpolationStroke) {
+        os << p.x << p.y;
+      }
+      os.closeChild();
     }
-    os.closeChild();
   }
 }
 
--- a/toonz/sources/toonzlib/tstageobjectspline.cpp
+++ b/toonz/sources/toonzlib/tstageobjectspline.cpp
@@ -113,6 +113,11 @@ TStageObjectSpline::TStageObjectSpline()
     , m_id(-1)
     , m_idBase(std::to_string(idBaseCode++))
     , m_name("")
+    , m_active(false)             // Needed for T2D compatibility
+    , m_color(0)                  // Needed for T2D compatibility
+    , m_steps(10)                 // Needed for T2D compatibility
+    , m_width(1)                  // Needed for T2D compatibility
+    , m_externalDataFound(false)  // Needed for T2D compatibility
     , m_isOpened(false) {
   double d = 30;
   std::vector<TThickPoint> points;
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Thanks @manongjohn for the near term fix.  Also for the quick turn around on this from only yesterday.
@shun-iwasawa over to you for the look at the breaking change.

My primary concern is the breaking change and your original thoughts on future proofing for unrecognized tags in the .tnz scene files by ignoring those tags seems the ideal for a full fix.

Note:  I have not yet tested this PR but will and have also considered the like scenario Opentoonz has with Toonz Premium with regard to compatibility.   Namely, that Toonz Premium balks at current .tnz files created with Opentoonz.  Thus the problem introduced with T2D is a problem also introduced in a different way via Opentoonz with regard to the original .tnz file format.  From what I can tell Toonz Premium notifies the user of malformed frames from a current Opentoonz produced .tnz file and then loads the scene thubnail into the xsheet. (In other words it fails somewhat gracefully)

More testing needs to be done to shine a fuller view on the underlying problem.

In other news... I see the Mac checks failed for @shun-iwasawa over to you for the look at the breaking change.this PR.  Perhaps this relates to the planned deprecation of Mac dependencies?

@shun-iwasawa over to you for the look at the breaking change.

Other than what I've noted, thus far this the PR itself looks good.

toonz/sources/toonzlib/tstageobjectspline.cpp
```diff
@@ -214,10 +288,15 @@ void TStageObjectSpline::loadData(TIStream &is) {
...
   const TStroke *stroke = getStroke();
+  QList<TPointD> interpStroke = getInterpolationStroke(); // Needed for T2D compatibility 
   os.child("splineId") << (int)m_id;
   if (!m_name.empty()) os.child("name") << m_name;
   os.child("isOpened") << (int)m_isOpened;
   os.child("pos") << m_dagNodePos.x << m_dagNodePos.y;
+  os.child("color") << (int)m_color; // Needed for T2D compatibility 
+  os.child("active") << (int)m_active; // Needed for T2D compatibility 
+  os.child("steps") << (int)m_steps; // Needed for T2D compatibility 
+  os.child("width") << (int)m_width; // Needed for T2D compatibility
```
shun-iwasawa, line 299: You load these parameters to OT instead of just ignoring them. Is it because, there may be a case that the scene saved with T2D is opened and overwritten by OT, and then loaded by T2D again and you still need to preserve these data, right?

Even if that is the case, how about saving these tags to the file only when the values are different from the default ones? Which means that these tags won't be saved unless you are using only OT.

manongjohn: > You load these parameters to OT instead of just ignoring them. Is it because, there may be a case that the scene saved with T2D is opened and overwritten by OT, and then loaded by T2D again and you still need to preserve these data, right?

Yes.  Users may be working together using either application, or may be doing something for a scene in T2D and then transfer to OT for other purposes and may want to come back. Without this, information is lost if it comes back to T2D.

> Even if that is the case, how about saving these tags to the file only when the values are different from the default ones? Which means that these tags won't be saved as long as you are using only OT.

That's fair. T2D defaults this information if not found in the file so OT doesn't need to write it if it doesn't need to.  I'll make a change that if it reads it in, it will write it back out, but if it was not read in, it won't write anything out.

manongjohn:
> My primary concern is the breaking change and your original thoughts on future proofing for unrecognized tags in the .tnz scene files by ignoring those tags seems the ideal for a full fix.

As far as future proofing goes: Yes...both applications should be tolerant enough to ignore unrecognized tags and continue loading as much as possible instead of immediately stopping in the middle and ignoring the rest of the file as it does now. This usually causes a crash when you try to interact with the application.  It should really only fail to load in the middle in an extreme case, which to me would be if an existing tag's number of data points changes causing it to misread the remainder of the file.

I would rather have a mostly loaded scene that I can still work with rather than a partially loaded scene that I can't.

There are cases where one may want to downgrade versions due to some issue in the latest version, but now the scene cannot be opened due to unrecognized tags in the older version. Of course it would only work for older versions that have the "skip unrecognized tags" logic.  While it may not be helpful for the current version, it would be good for future versions.

toonz/sources/toonzlib/tstageobjectspline.cpp
```diff
@@ -255,30 +263,31 @@ void TStageObjectSpline::loadData(TIStream &is) {
...
-        m_interpolationStroke.push_back(TPointD(1020, 1000));
-        m_interpolationStroke.push_back(TPointD(1020, 1000));
-        m_interpolationStroke.push_back(TPointD(1040, 1000));
-        m_active = false;
-        m_steps = 10;
-        m_color = 0;
-        m_width = 1;
+  // try loading the old type
+  if (m_externalDataFound &&
+      points.size() == 0) {  // Needed for T2D compatibility
```
shun-iwasawa, line 268: I don't think this block is needed.
The original commit in T2D (https://github.com/tahoma2d/tahoma2d/pull/420/), this condition was added replacing [the block for handling old version of Toonz scene files](https://github.com/tahoma2d/tahoma2d/pull/420/files#diff-455d1a948001440c636d62a3294f844e47eb8e5258c07936bc4a0bc5a9908c80L178-L184) .
In OT the conventional version check is still there.

manongjohn: Removed.   I was skeptical about including it myself. Not even sure why it was added to T2D as it seems an unlikely scenario.

I squashed my commits and forced pushed because I realized I forgot to add a co-authored-by Jeremy since I took these changes from his Motion Path PR.

toonz/sources/toonzlib/tstageobjectspline.cpp
```diff
@@ -113,13 +113,31 @@ TStageObjectSpline::TStageObjectSpline()
...
+  m_interpolationStroke.push_back(TPointD(-20, 0));
+  m_interpolationStroke.push_back(TPointD(-20, 0));
+  m_interpolationStroke.push_back(TPointD(0, 0));
+  m_interpolationStroke.push_back(TPointD(65, 65));
+
+  m_interpolationStroke.push_back(TPointD(935, 935));
+  m_interpolationStroke.push_back(TPointD(1000, 1000));
+  m_interpolationStroke.push_back(TPointD(1020, 1000));
+  m_interpolationStroke.push_back(TPointD(1020, 1000));
+  m_interpolationStroke.push_back(TPointD(1040, 1000));
```
shun-iwasawa, line 140: I think setting the default values of `m_interpolationStroke` here is not needed, since the values will always be replaced when loading the T2D file.

toonz/sources/toonzlib/tstageobjectspline.cpp
```diff
@@ -226,6 +290,15 @@ void TStageObjectSpline::saveData(TOStream &os) {
     os << p.x << p.y << p.thick;
   }
   os.closeChild();
+  if (m_externalDataFound) {  // Needed for T2D compatibility
+    os.openChild("interpolationStroke");
+    n = interpStroke.size();
```
shun-iwasawa, line 295: I think you can simply use `m_interpolationStroke` directly instead of assigning the values to `interpStroke`.
(i.e. you can remove the function `TStageObjectSpline::getInterpolationStroke()` )

RodneyBaker:
The Mac check for this PR appear to be taking an extraordinary amount of time.

RodneyBaker:
I've had some success in testing but still need to spend some time.
Initially I was crashing in Tahoma2D just setting up motion paths (not sure if that relates to the action builds only).
When I was successful and saved a scene with two motion paths (each connected to a different column) when Opening the scene in Opentoonz only one motion path appeared in the schematic.

My initial take is that this is working pretty good but there is perhaps another problem more on the surface for users in that .tnz scene can very easily not load when not properly assigned to a project file.  Both OT and T2D do present the error message that the scene cannot be opened because it doesn't belong to a project.  Will have to investigat this workflow... that of easily just opening a tnz scene file from anywhere and having it properly open as a separate issue.  The experiment File path rules does seem to make a difference although I need to test this more thoroughly.

image

At any rate, concerning this PR the main areas I would suggest to test would be:  Create a scene (starting with OT) with multiple columns/levels and multiple motion paths and see if how they make the transition back and forth between T2D and OT.  Do the same but this time create the multiple motion paths in Tahoma.

manongjohn:
> When I was successful and saved a scene with two motion paths (each connected to a different column) when Opening the scene in Opentoonz only one motion path appeared in the schematic.

I'm not having this issue. I can create multiple paths linked to different nodes and open in either direction.

RodneyBaker:
I'm stil needing to do a few tests on this PR that that I haven't been able to set up yet to include trying to find out why I was running into a crash on Tahoma2D.  That issue might not relate to this PR at all but need to figure out a test that would isolate and confirm that.  

On the OT side the changes of this PR seem to work as advertised.

manongjohn:
Not sure which version/nightly of T2D you using, but there is a different but related issue where T2D doesn't fail to load but crashes when selecting a Motion Path loaded from OT.  Make sure you use that PR's artifact (or wait til it gets to the nightly) when testing.

RodneyBaker:
Pressing forward with this as it works as advertised on the Opentoonz side.
I'll follow up with the Tahoma side tests and it's nightly for that side of the equation.

PR #4236, opened 2022-01-25
Fix dynamic conical fractal noise

shun-iwasawa:
This PR fixes the Fractal Noise Fx Iwa with the conical transform option (again). Now it can generate better result when `Fractal Type` is set to `Dynamic` or `Dynamic Twist` .


RodneyBaker:
Thank you @shun-iwasawa   again   for the fix to conical fractal noise!

PR #4241, opened 2022-02-04
Raster level frame number format & PNG for new Raster level 

shun-iwasawa:
This PR enables to specify the frame number format of raster sequential files generated with OT.
The frame number format consists of the following:
- **Separate Character** : specifies a character between level name and frame number. (either "." period or "_" underscore) 
- **Padding** : specifies character length for frame number section.

Default values are `Separate Character = "." (period)` and `Padding = 4` . (like `A.0001.tif`) 

The frame number format can be modified via combo boxes newly added in the format option popup.
frameformat

Each scene contains two frame number formats, one is for output and the other one is for input.
These formats are shared by the following features:
**Frame format for output** (with blue box) : `Output Settings` and `Convert`
**Frame format for input**  (with red box) : `Camera Capture`, `Export Level`, and creating a new Raster level.

Note that the formats can be default by using `Save Default Settings` command.

-------

This PR also enables to use PNG file for new Raster level created in OT.

- Default file format for new Raster level can be specified in `Preferences > Drawing > Default Raster / Scan Level Format` .
The option replaces `Scan File Format` option. Now the option takes effects both Raster and Scan levels. (Scan level is not used much as of now.)
- The file format can also be specified in the New Level popup.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -1727,6 +1727,15 @@ ProjectPopup QLabel {
   padding-right: 5px;
 }
 /* -----------------------------------------------------------------------------
+   Frame format settings box in FormatSettingsPopup 
+----------------------------------------------------------------------------- */
+#FrameFormatBoxInput {
+  border-color: rgba(135, 22, 22, 0.82);
+}
+#FrameFormatBoxOutput {
+  border-color: rgba(21, 22, 136, 0.82);
+}
+/* -----------------------------------------------------------------------------
    Schematic Viewer
 ----------------------------------------------------------------------------- */
 SchematicViewer {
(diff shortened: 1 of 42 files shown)


gab3d:
It's a great option! Thanks for that @shun-iwasawa  😊

Just as a minor comment, I'd say that in the Output Settings panel I'd expect this kind of option to be visible in the main panel, near the File Name input field and not inside a specific file format Options window.

I realize that this should be ghosted when output is not an image sequence, and perhaps that's the reason why you chose to "hide" it inside the Options window, but still, if there is the chance to do it, I think users will be much more aware of the existence of the option if it were visible right away.
(another good reason to keep it where you put it would be if this option was tied to each individual file format, which I don't know, as I haven't had the chance to try it yet)

Anyway, it's a very much welcomed addition! 💪

RodneyBaker:
Thanks @shun-iwasawa for the new options and PNG as a Raster Level type.
LGTM

RodneyBaker:
@shun-iwasawa Nightly builds are failing to build.

PR #4242, opened 2022-02-04
Style Editor: Move "toggle orientation" button in the menu

shun-iwasawa:
This PR moves "Toggle orientation of the Color page" button into the hamburger menu in order to gain extra width for style chips when the hex color field is hidden.
toggleorientation


RodneyBaker:
Thanks for the modification @shun-iwasawa 
Just a few days ago I ran into the limited for style chips in the Style Editor.
I was pleasantly surprised to discover we could scroll the hidden chips into view.
Having more visible is always welcome.

LGTM

PR #4244, opened 2022-02-06
Stop motion patch for non-windows platforms

justburner:
Small fix for #4233 that made non-windows platforms crash when opening the Stop Motion window.

Thanks @artisteacher for crash report and @manongjohn for finding the cause of the issue. 👍


RodneyBaker:
Thanks for the patch @justburner 
Nightly builds have been failing to build so we'll see if this one builds.

cc:  @shun-iwasawa

shun-iwasawa:
I restarted jenkins . Thank you for your patience.

PR #4247, opened 2022-02-08
Fix saving tlv

### LINKED ISSUE #4246: Fail to save PNG and TPL palette on latest nightlies
On latest nightlies (last two) I'm getting an error where TLV levels seem to be getting confused with raster levels.  This relates to palette saving and that is the error that message that pops up.  This generally leads to a crash later if continuing to work with the scene.

I'll upload some screenshots but am having some issue replicating the problem without crashing.

Of note:  The error message states that a PNG image cannot be saved and yet I'm not using any PNG image... leaving the default option as TIF for the standard raster level.  

I am using the option 'Use White as Transparency' in the Level Settings.
This to show through the TLV level underneath that has key'd color.
Added:  This doesn't appear to related to the specific TLV saving issue.

This seems mostly unrelated but a few days ago a fix was posted to Tahoma2D relating to palette saving.
REF:  [Fix palette saving issues](https://github.com/tahoma2d/tahoma2d/pull/895#)
#895

Update:
I was able to salvage a test project by specifically saving the levels as .tlv levels whereas it was apparently trying to reference them as PNG.

shun-iwasawa:
This PR fixes #4246 . I deleted a necessary line by mistake when  developing #4241 .

### DIFF
--- a/toonz/sources/toonzlib/toonzscene.cpp
+++ b/toonz/sources/toonzlib/toonzscene.cpp
@@ -1386,6 +1386,7 @@ TFilePath ToonzScene::getDefaultLevelPath(int levelType,
     break;
   case TZP_XSHLEVEL:
     levelPath = TFilePath(levelName).withType("tlv");
+    break;
   default:
     levelPath = TFilePath(levelName + L"..png");
   }


shun-iwasawa:
I'll merge this as the fix is quite obvious.

RodneyBaker:
Thank you @shun-iwasawa   Much appreciated!

PR #4248, opened 2022-02-08
Fix crash on saving scene containing missing palette level

### LINKED ISSUE #4237: Saving scene with audio causes opentoonz to close
**Describe the bug**
When saving a particular scene, it saves the scene (no work is lost) and then closes the program. Other, similar scenes (similar length of 5000-6000 frames, and with a similar audio file) do not cause the issue.

**To Reproduce**
1. Open the problematic scene
2. ctrl-s, or file/save all
3. The program saves, then closes.

**Expected behavior**
The program should not close after saving. 

**Screenshots**
https://i.gyazo.com/8889ca5bd91c95d2641b2496034282a6.mp4

**Device Information:**
 - OS: Windows
 - OpenToonz Version: Nightly (Jan 28 2022) (Updated specifically to see if the new version would fix it, it did not.)
 - GPU: NVIDIA GeForce GTX 1080
 - GPU Driver Version: 27.21.14.5671

shun-iwasawa:
This fixes #4237 .
The crash occured when opening a scene which contains a palette level whose file does not exist and then trying to save the scene.

### DIFF
--- a/toonz/sources/toonzlib/sceneresources.cpp
+++ b/toonz/sources/toonzlib/sceneresources.cpp
@@ -407,7 +407,9 @@ void ScenePalette::rollbackPath() { m_pl->setPath(m_oldPath); }
 
 //-----------------------------------------------------------------------------
 
-bool ScenePalette::isDirty() { return m_pl->getPalette()->getDirtyFlag(); }
+bool ScenePalette::isDirty() {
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
I'll merge this as the fix is obvious.

PR #4250, opened 2022-02-08
Camera calibration for the Camera Capture feature

shun-iwasawa:
This PR will add a camera calibration feature to the Camera Capture, which undistort the captured images by using [OpenCV's calibration feature](https://docs.opencv.org/4.x/d4/d94/tutorial_camera_calibration.html).

### How to calibrate camera

1. Prepare the checkerboard
    - Print out `checkerboard.tif` in `OpenToonz Stuff/library/camera calibration` to A4 sized paper.
    - It would be better to paste the checkerboard pattern on some flat panel like cardboard.

2. Capture the pattern
    - In the Camera Capture popup, enable `Camera Calibration` box.
    - Click "Start calibration".
    - Take 10 snapshots of the checkerboard pattern in various positions, angles, and distances.
       calib_checkers
    - Once the capturing is done, calibration will be automatically applied.
          

3. Export & load calibration settings
    - Calibration settings will automatically be saved under the folder `OpenToonz Stuff/library/camera calibration` with name `[MACHINENAME]_[CAMERANAME]_[RESOLUTION].xml`
    - The settings file will be overwritten after finishing the new calibration.
    - You can export the file for backup, or load it afterwards.


artisteacher:
Will this work ok using 8.5x11? That's standard in the US.

gab3d:
@shun-iwasawa Thanks! It's a very good feature.

Once calibrated, will that be applied when capturing from the stop-motion controls panel too?

RodneyBaker:
Will merge this so users can experiment and fully test various assumptions via the nightly release.

shun-iwasawa:
@artisteacher 
>Will this work ok using 8.5x11? That's standard in the US.

Yes. Actually the size of pattern does not matter (as the computation is done in arbitrary unit space). The conditions of the checkerboard for proper calibration are as follows:
- Whole pattern should be captured in every snapshots.
- The amount of squares should not be changed since it's [hard coded](https://github.com/opentoonz/opentoonz/pull/4250/files#diff-a13dfb1c802732b95f73cfe23aedf7ad553aeb58397b0e3bbfe9bc567f635b5aR281).
- There should be sufficient amount of white margin around the pattern for better recognition.

When you print the image in A4 (or Letter-sized) paper it would be sufficient with usual setup for pencil testing.
As long as you satisfy the conditions you can even use a newspaper-sized pattern. :-)

@gab3d
> Once calibrated, will that be applied when capturing from the stop-motion controls panel too?

Unfortunately currently the camera calibration feature is only for Camera Capture popup, as I'm not yet familiar with the stop-motion code.

PR #4252, opened 2022-02-10
Fix crash on undoing tool

### LINKED ISSUE #4249: incorrect undo behavior after fill
**Describe the bug**
If we undo server times, after using fill-tool, OpenToonz will act incorrectly.

**To Reproduce**
take a look at the attached screen [recording](https://user-images.githubusercontent.com/417336/152914234-ccf0969e-4d0c-407b-9ae9-a7b4fe91849b.mp4)

**Expected behavior**

**Screenshots**

**Device Information:**
 - OS:  tested both macOS and Windows
 - OpenToonz Version: 1.5 release
 - GPU: Radeon Pro 555 2 GB / Intel HD Graphics 630 1536 MB
 - GPU Driver Version: from macOS 11.6

shun-iwasawa:
This PR fixes #4249 . Step to reproduce the crash is described in [this comment](https://github.com/opentoonz/opentoonz/issues/4249#issuecomment-1033309414).

Actually the crash can occur not only with the Fill tool but also with many other tools such as Paint Brush, Eraser, Tape, etc.
All of them does not call `TTool::touchImage()` therefore the flags like `TTool::m_isLevelRenumbererd` does not reset to false.

I added the function to reset all flags on adding the undo so that they won't affect to subsequent operations.

### DIFF
--- a/toonz/sources/include/tools/toolutils.h
+++ b/toonz/sources/include/tools/toolutils.h
@@ -182,6 +182,8 @@ public:
         .arg(QString::fromStdWString(m_level->getName()))
         .arg(QString::number(m_frameId.getNumber()));
   }
+
+  void onAdd() override;
 };
 
 //================================================================================================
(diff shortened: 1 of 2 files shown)


RodneyBaker:

PR #4256, opened 2022-02-14
Fix Stage Schematic Column (Right click) crash

### LINKED ISSUE #4255: Stage Schematic Column  (Right click) Bug
when in the Stage Schematic and you select a Column then double click to rename and simply Right Click you get a instant crash every time.
Screenshot (203)

justburner:
Fixes #4255


RodneyBaker:
Thanks for the fix and the quick turnaround @justburner 
LGTM

PR #4257, opened 2022-02-15
Fix output file name for movie format

shun-iwasawa:

#### To reproduce
1. Load / create some level to a brand-new scene.
2. Open the Output settings.
3. Set output format to `png` (or any sequential image file format).
4. Open format option and set the separate character to `_ (underscore)`.
5. Set output format to `mp4` (or any movie format).
6. Render the scene.

In my environment, the output file becomes either `_.mp4` (when the file name is the same as the scene name) or `filename_.mp4`  (when the _filename_ is different from the scene name). Which are both wrong. It should be `filename.mp4`.


RodneyBaker:

Issue #4081, opened 2021-10-01
v1.6 Release Schedule


RodneyBaker:
v1.6 Release Schedule  

- [x] Feature Freeze (15 February) 
- [x] Two weeks for bug tracking and translations
- [x] Release Candidate
- [x] Two weeks for urgent bug fix (as required) - releasing RC2 (as needed)
- [x] Full Release

RodneyBaker:
Placeholder

RodneyBaker:
Placeholder

RodneyBaker:
Placeholder

RodneyBaker:
Placeholder

RodneyBaker:
Placeholder

RodneyBaker:
Placeholder

RodneyBaker:
(early draft change log)

Viewers/Flipbook
#4093 30 bit display feature by shun-iwasawa
#4202 Fix Flipbook wobbling when dragging while playing by shun-iwasawa 
#4197 Enhance Flipbook playback by shun-iwasawa
#4108 Modify Flipbook Histogram by shun-iwasawa
#4105 Fix crush on showing snapshot in flipbook by shun-iwasawa

Xsheet / Timeline
#4194Fix pressing Shift key loses focus when renaming cell by shun-iwasawa
#4178 Cell Mark feature by shun-iwasawa
#4050 Fix Autorenumber of Raster Level by shun-iwasawa
#4124 Xsheet Minimum Layout by shun-iwasawa
#4141 Fix undo pasting raster selection to cell by shun-iwasawa
#4134 Tahoma2D porting: Allow Changing Parent of Column from XSheet by jeremybullock ported by shun-iwasawa
#3944 Xsheet Zoom Control by shun-iwasawa 
#3922 Fix Cell Selection on Click by shun-iwasawa 

Rendering / Image formats 
#4205 Multi-Thread FFMPEG and responsive finalizing window by justburner

Image formats
#4192 OpenEXR I/O by shun-iwasawa
#4082 Add Raster Lock Alpha tool setting by manongjohn 
#4167 Fix PSD loading and level format preferences by shun-iwasawa
#3897 Fix PSD Loading by shun-iwasawa

Notelevels
#4190 Note level column enhancement by shun-iwasawa

Level Strip 
#4186 Fix FrameIds Inconsistency in Level Strip Commands by shun-iwasawa
 
Export
#4185 Fix TVP JSON export by shun-iwasawa 
[...]

Issue #2368, opened 2018-11-06
Issues with Inverse kinematics and Pinned center


psychoanima:
### Issue Summary
Pinned center with IK bones is not locking bones properly, although lock option is activated

### Video Reference
https://1drv.ms/v/s!AtMv1SrHrjT38m2lnCUmt6kDz7ZO

### Steps to Reproduce
Lock at least two bones at different frames 

### Expected Results
Pinned bone should stay locked while other bones will be able to move

### Actual Results
Pinned bone is not locked (reverse effect)

### System Information
* **OpenToonz Version**:  1.2.1 (built jun 28) 
* **Operating System**: Windows 10 Pro
* **CPU**: i7 2630QM
* **Memory**: 20 gb
* **Graphics Card**:  NVidia Quadro 1000m

beeheemooth:
I have about the same problem with two legs.

https://groups.google.com/forum/#!category-topic/opentoonz_en/troubleshooting/Yf5sL_mbgQ8

![ezgif-4-cd6d8dca54](https://user-images.githubusercontent.com/25724133/48094927-14291b00-e21c-11e8-83e5-edcc108cff07.gif)

DarrenTAnims:
I've not been able to use IK because of this. It's never worked for me. I meant to log a bug, but never found time and I just used other methods to animate with.

DarrenTAnims:
I've not been able to use IK because of this. It's never worked for me. I meant to log a bug, but never found time and I just used other methods to animate with. So thanks for logging these, with examples.

psychoanima:
I hope someone will assign to this issue and fix it so we can use IK normally :)

turtletooth:
@psychoanima @beeheemooth Thanks for posting examples here.  I would personally love for IK to be usable, but I don't touch it because the results are so unreliable.

artisteacher:
I agree - using IK is very unpredictable. Adding constraints to limit the range of motion would help in some areas - though probably not with pinning.

PR #4258, opened 2022-02-16
revert changing pinned center

### LINKED ISSUE #3616: Where is option to change bone pinned center?!
I was using an older version of OT and I decide to download nightly build just to check what is new. I was surprised to discover that option to change skeleton pinned center is gone! Only temporary pinned center is available! Guys am I crazy, is this just a bad dream or someone seriously took away this option from IK? Long time ago I even asked Mr. Iwasawa to create indicator in xsheet, to see on which keyframe pinned center is changed (visible as a blue square in xsheet, next to keyframe number). And of course he did it, because it sounded as a logical thing to have, and that option is right now useless cuz it's impossible to change pinned center.

Please tell me that I am wrong and that there is still an option to change pinned center for bones in IK mode.

### LINKED ISSUE #3617: How to prevent pinned center bone bug (Inverse Kinematics)
After many hours of experimenting and crashing things I found a solution, how to prevent well known bug with Inverse Kinematics and Pinned center

Please take a look:

https://1drv.ms/v/s!AtMv1SrHrjT3gZt3ZWF9IRoSE7GMRg?e=CzrWVr

shun-iwasawa:
This PR will revert a feature for changing pinned center in Inverse Kinematics mode of the Skeleton tool.
The feature was once removed in #3275 . According to the discussion starting [here](https://github.com/opentoonz/opentoonz/issues/4081#issuecomment-1031560318) , I reverted it but with ctrl modifier needed for the operation in order to prevent mishandling.

**Usage:**
- Ctrl + Click to toggle pinned center
- Shift + Click to toggle temporary pin

This PR will resolve #3616 and may resolve #3617 as well (as the workaround will no longer be necessary) .

@rabe01 @gab3d 
Can you please check if this will (at least) recover the functionality which was available in v1.4 and older versions?

### DIFF
--- a/toonz/sources/tnztools/skeletontool.cpp
+++ b/toonz/sources/tnztools/skeletontool.cpp
@@ -413,19 +413,22 @@ void SkeletonTool::leftButtonDown(const TPointD &ppos, const TMouseEvent &e) {
 
   // lock/unlock: modalita IK
   if (TD_LockStageObject <= m_device && m_device < TD_LockStageObject + 1000) {
-    Skeleton *skeleton = new Skeleton();
-    buildSkeleton(*skeleton, currentColumnIndex);
     int columnIndex = m_device - TD_LockStageObject;
     int frame       = app->getCurrentFrame()->getFrame();
-    if (skeleton->getBoneByColumnIndex(columnIndex) ==
-        skeleton->getRootBone()) {
-      app->getCurrentColumn()->setColumnIndex(columnIndex);
-      m_device = TD_Translation;
-    } else if (e.isShiftPressed()) {
+    if (e.isCtrlPressed() || e.isShiftPressed()) {
+      // ctrl + click : toggle pinned center
+      // shift + click : toggle temporary pin
       togglePinnedStatus(columnIndex, frame, e.isShiftPressed());
       invalidate();
       m_dragTool = 0;
       return;
+    }
+    Skeleton *skeleton = new Skeleton();
+    buildSkeleton(*skeleton, currentColumnIndex);
+    if (skeleton->getBoneByColumnIndex(columnIndex) ==
+        skeleton->getRootBone()) {
+      app->getCurrentColumn()->setColumnIndex(columnIndex);
+      m_device = TD_Translation;
     } else
       return;
   }


RodneyBaker:
That seems to work quite well (with regard to changing pinned center).
Seems a good revert to me.

leg

It did seem to be a little harder to set up the connectivity between columns but that might just be me in need of more experimentation.

RodneyBaker:
I'm starting to feel a bit more comfortable with Opentoonz's IK... AND with this revert
IK

psychoanima:
@RodneyBaker I am a bit dumb for github, can I download that build and where, would like to test it out?

RodneyBaker:
Here is the Windows artifact (a portable):

https://github.com/opentoonz/opentoonz/suites/5324816993/artifacts/165910104

beeheemooth:
Unfortunately changing the pinned center still doesn't count if you change it during the key animation. :(

RodneyBaker:
> Unfortunately changing the pinned center still doesn't count if you change it during the key animation. :(

Yes.  Understood.  But is a revert of the change better than not reverting?

beeheemooth:
Why do these pink things (manipulators) display the correct behavior when the pinned center changes, but the skeleton (with the changed pinned center) does not use this data? The skeleton moves all parts as if the pinned center has not been changed.

@RodneyBaker Why return something that doesn't work? For the user to rack their brains trying to figure out how it works? 

Screen_0462

psychoanima:
@shun-iwasawa I can confirm from my side that functionality is recovered. 


Anyone interested in using this option to lock bone/ankle in some position I can record a video how it can be done.

beeheemooth:
@psychoanima There is no hatred in me, but I want to be objective. When @turtletooth suggested removing the "pinned center" I was the first to oppose it. But later I realized that he did this so that the work with inverse kinematics was more stable. The above examples of using IK are purely formal (demonstrative) in nature. In practice, we have frequent switching between IK and FK. Imagine that you have done a rather complex key animation using IK and accidentally changed the pinned center - the whole animation will shift and you will need to redo it from the very beginning, since the "undo" operation does not work.

RodneyBaker:
One thing that makes this IK process hard to test is a visual glitch that sends things (apparently) off canvas.
Moving back to earlier frame, then back to the current frame can restore everything to appear in proper place.

I can definitely see why Jeremy made the change though as it can be hard to navigate the various inconsistencies.

psychoanima:
I didn't have much time to create something special with it but here is the workflow, as shown in video bellow:

https://youtu.be/AsJ_bvNZRiU

Now, you might wonder why do I have five bones for three moving elements? Good question, not quite sure but it works. Trough many failed attempts I realized that I need to have a "root bone", something similiar like in 3D rigs or 2D moho rigs (one bone to move them all). However, even root bone is not enough, cuz it's stiff, and when you pin that bone, first "ankle" won't budge. For that reason I add Main bone, which is basically invisible one, but it's the top bone in hiararchy that should be pinned. Bottom bone that can be pinned is the last bone (foot in this example) and doesn't need extra "invisible" bones to make it locked. So far in my testing, only the top of bone hiararchy need one Root (to hold the chain) and one Main bone (to act as first ankle that can be pinned/locked) to make everything else work. I know how confusing this text sound but with video above maybe it makes sense.

gab3d:
@psychoanima 
I think it'd be of much help to add this video (or a similar one) to the corresponding manual section.

RodneyBaker:
Interesting @psychoanima   I too was finding that I need to create a NULL (so to speak) to parent everything to to get best results.

As I'm the likely one to merge this PR and find myself in favor of the change I hesitate a little to merge BUT... just as this is a revert to the earlier PR should we discover additional problem those can be addressed in subsequent changes.

As such...

Posting intent to merge and very interested in any and all discussion related to this PR.

RodneyBaker:
Thank you all for your feeback!
And thank you @shun-iwasawa for the revert and adjustment.
LGTM

rabe01:
It's nice to see how many people get involved here - and that something is moving.
It looks to me like IK is at least possible again.
thank you

psychoanima:
>    I am finding that rather than take the time to place each Skeletal bone into place that if I move the elements manually via Animate Tool/Position that when I activate the Skeleton Tool everything is automatically rigged. That can save a lot of time in searching for bones and dragging them so they will connect with the desired columns. 
> 


That's really cool.

Next days I will dive deeper into IK and pinned bones, to see how far it can go. 

For now what I see really necessary is an option to move blue square pinned center keyframes on x-sheet. Once placed they (blue squares) are locked and can't be moved, neither they are following keyframes related to pinned bone if they are being repositioned on x-sheet (keyframes in same row as blue square pinned keyframe) 

Meaning, you can't tweak and re-time key positions once pinned center is defined. In-between's don't affect pinned bones, so that's a plus.

beeheemooth:
You can call me Unbelieving Thomas, but I don't believe this thing works. :)  In all cases, you demonstrated a simple IK chain with no problems (@RodneyBaker regarding "nulls" - we discussed this approach back in 2018). Show me a simple line spider that has IK branches and a common center so I can believe.

RodneyBaker:
@beeheemooth   While understandably not very impressive I have delivered on your request for an example of character walking up stairs (with additions demo'd via discord).  Perhaps you can provide a setup to test for your line spider so we can focus more closely on your requirements.

There is much that still needs to be done... obviously... but at least we have an important capability back with this revert.

beeheemooth:
Screen_0465

gab3d:
@psychoanima 
Agree on being able to move the pinned center mark in the Xsheet is something necessary to be able to do basic adjustments in animation timing.

psychoanima:
> You can call me Unbelieving Thomas, but I don't believe this thing works. :) In all cases, you demonstrated a simple IK chain with no problems (@RodneyBaker regarding "nulls" - we discussed this approach back in 2018). Show me a simple line spider that has IK branches and a common center so I can believe.

Screenshot 2022-02-18 015806

Webp net-gifmaker

Back four legs are IK with pinned bones, front two legs are FK. Do you believe now, Unbelieving Thomas? :)

beeheemooth:
Hmm. It seems that you are no longer confused by the problem of shifting pinned bones that you wrote about in #2368. 
In fact, I also really want to believe that it will work if I do at least one phase of the spider's walk, fixing the bones of different limbs in turn.

psychoanima:
My confusion didn't stop, I still don't know what is happening under the hood and there is no doubt that pinned bones are not working as intended. Right now I just found the way around it until error is fixed.

If we could only have some programmer willing to dedicate his time on this tool, we could do wonders. With proper IK tool and working pinned option this tool could become really handy in making animated characters for 2D games - so much potential!

PR #4261, opened 2022-02-17
Rhubarb integration (Auto Lip-sync)

justburner:
Port of Tahoma2D Rhubarb integration, with a few changes:
- No UI changes on OT part, the new auto lip-sync has it's own window and menu: `Lip Sync > Apply Auto Lip-Sync to Column` over right-clicking the first cell to access it, or the default `Ctrl+Alt+L` shortcut.
- List in `Audio Source` have the actual columns names instead of "Column 1" and so on.
- All preview audio stops when swapping source or leaving window, instead of mysteriously still playing on background.
- Opening Auto-Sync on an invalid column now auto closes it after the warning (OT lip-sync data window behavior is still unchanged).

**How to setup Rhubarb:**

The steps are very similar to how to setup FFMpeg.

1. Go to https://github.com/DanielSWolf/rhubarb-lip-sync and download the latest release.
2. Extract all into somewhere safe on your drive, make note of the path where the `rhubarb` executable is.
3. In OpenToonz, go to `File > Preferences...` and select `Auto Lip-Sync` category.
4. Put the full path from step 2 into the `Rhubarb Path` box.
5. Close the Preferences window.

If all done correctly you can now use auto lip-sync feature right away (yep, no restart required).

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -284,6 +284,8 @@ public:
   int getFfmpegTimeout() { return getIntValue(ffmpegTimeout); }
   QString getFastRenderPath() const { return getStringValue(fastRenderPath); }
   bool getFfmpegMultiThread() const { return getBoolValue(ffmpegMultiThread); }
+  QString getRhubarbPath() const { return getStringValue(rhubarbPath); }
+  int getRhubarbTimeout() { return getIntValue(rhubarbTimeout); }
 
   // Drawing  tab
   QString getDefRasterFormat() const { return getStringValue(DefRasterFormat); }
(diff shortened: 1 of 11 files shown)


RodneyBaker:
@shun-iwasawa This is the PR from @justburner I had mentioned before in anticipation of it getting into the v1.6 release.

RodneyBaker:
An option for Opentoonz users who don't want to search out and download Rhubarb (and also ffmpeg)  is to download Tahoma2D (T2D) which comes with both of those third party programs.  Then just point Opentoonz to those appropriate locations via Preferences.  

For testing purposes I tend to download T2D and place it in the Opentoonz stuff directory where all three programs can easily be found.  This works particular well because all three programs are portable and do not need to be installed.  

Then if I move or archive my stuff directory those supporting programs move right along with it.

RodneyBaker:
Will merge this as it all checks out.
Thank you @justburner for the useful addition.
LGTM

PR #4264, opened 2022-02-21
Fix viewer playback

### LINKED ISSUE #4253: Playback slow with Xsheet
I'm noticing inconsistent playback slowness issues that I've traced back to PR #4197.

- Case 1:
I loaded an older scene I had done that was at 30FPS (11 vector columns + 2 meshes).  In both Timeline/Xsheet, it plays back 17-19 FPS. In the original release and in the 12/24/2021 nightly prior to Enhanced Flipbook playback was merged, it plays at 29-30 FPS.

- Case 2:
I loaded another older scene at 24FPS (14 vector columns + 4 sub-xsheets).  On the Timeline, it played more consistently at 24 FPS, but in Xsheet, it flips quite a bit between 22-25 FPS.  In the original release and in the 12/24/2021 nightly, it plays consistently at 24FPS in either.

- Case 3:
I was using a 1/27 nightly playing back a 24FPS scene (various types of columns). It would play back anywhere between 12-15 FPS when using an Xsheet.   Switching the Xsheet to Timeline mode, the playback was a bit faster, but still slower (20-22 FPS) than expected.  

It seemed the more columns/rows that are visible in the Xsheet the slower it got.  This does not appear true for Timeline which seems odd considering it's the same code.

Another oddness about this case is it suddenly started playing back relatively normal. As I was testing this issue out trying to see what conditions might be impacting it, the problem was consistent for a least an hour or more, no matter how many times I restarted OT or reloaded the scene.  Then all of a sudden it was fine.  

No idea what  I might have done to correct the problem but couldn't reproduce this specific scenario again with the same scene.  However, similar to other scenes, Timeline played back more consistently at 24 FPS, but in Xsheet it would vary quite a bit between 21-23 FPS

shun-iwasawa:
This PR (hopefully) fixes #4253 .
Modified Viewer / Combo Viewer not to call  `qApp->processEvents()` on drawing frame.

#### Advantage
- The playback becomes faster. 
   Unlike Flipbook, playing on Viewer / Combo Viewer advance the current frame. Calling `qApp->processEvents()` on drawing frame caused repaint of other panels (such as Xsheet) which slows playback. Now the Viewer / Combo Viewer panels just call `update()` in specified interval and let the event handler repaint the frame at certain timing.

#### Disadvantage
- The playback may become uneven, just like before applying #4197 .

@manongjohn @melieconiek 
Can you please try this PR and see if it resolves the issue? Thanks!

### DIFF
--- a/toonz/sources/toonz/comboviewerpane.cpp
+++ b/toonz/sources/toonz/comboviewerpane.cpp
@@ -288,10 +288,9 @@ void ComboViewerPanel::onShowHideActionTriggered(QAction *act) {
 
 void ComboViewerPanel::onDrawFrame(int frame,
                                    const ImagePainter::VisualSettings &settings,
-                                   QElapsedTimer *timer, qint64 targetInstant) {
+                                   QElapsedTimer *, qint64) {
   TApp *app = TApp::instance();
   m_sceneViewer->setVisual(settings);
-  m_sceneViewer->setTimerAndTargetInstant(timer, targetInstant);
 
   TFrameHandle *frameHandle = app->getCurrentFrame();
 
@@ -327,10 +326,6 @@ void ComboViewerPanel::onDrawFrame(int frame,
 
   else if (settings.m_blankColor != TPixel::Transparent)
     m_sceneViewer->update();
-
-  // make sure to redraw the frame here.
-  // repaint() does NOT immediately redraw the frame for QOpenGLWidget
-  if (frameHandle->isPlaying()) qApp->processEvents();
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


manongjohn:
For some reason when I use the artifact from this PR, it crashes loading the 30 fps scene I was using to show the issue.  I can load other scenes fine though.

However, when I built OT master with your changes pulled in, the scene loaded fine and plays back closer to 30 fps, similar to how it plays in the official 1.5 release.

I'll have to see if the crashing issue persists when this gets merged into the nightly.

RodneyBaker:
@manongjohn For unknown reason the github action artifacts build with a number of issues especially related to standard raster levels.  This can especially be seen in use of raster brushes such as mypaint but also elsewhere.  This same problem isn't seen in the appveyor artifacts... from which the nightly builds are generated.

Will merge this to get this PR into circulation for confirmation.

PR #4265, opened 2022-02-21
Fix fx gadget picking

shun-iwasawa:
This PR fixes logic for picking the fx gadget (i.e. finding the gadget under the cursor position) , especially for the gadgets with more than one handles.

Before the fix, the current handles did not switch without switching the current gadget. You cannot select different handle in the same gadget without moving the cursor off the gadget once. 

The change would be prominent, for instance, in the gadget for `Corridor Gradient Fx Iwa` .
With this PR you can dynamically switch the highlighted handles while hovering on the gadget.


RodneyBaker:

PR #4268, opened 2022-02-21
Fix GIF rendering on older versions

justburner:
Port of tahoma2d/tahoma2d#893 but re-implemented with a less hacky way, plus now properties can be invisible.
Also restored '-r' flag to avoid FFmpeg messing with the number of frames.


justburner:
tiny force push to get rid of redundant `setQStringName(tr("Generate Palette"));` specially since translations were already generated to v1.6 RC and i don't wanna mess with that.

RodneyBaker:
Thank you for the updated fix @justburner

PR #4269, opened 2022-02-22
Fix Stage Schematic Column (Right click) now w/Context Menu

justburner:
While I've provided a quick fix in #4256 for #4255, it prevents opening any context menu inside rename widget in the node.

This PR try to simulate the most common commands such as Cut, Copy, Paste and Select as a workaround for the Qt 5.15 bug.

### DIFF
--- a/toonz/sources/include/toonzqt/schematicnode.h
+++ b/toonz/sources/include/toonzqt/schematicnode.h
@@ -19,6 +19,7 @@ class SchematicName final : public QGraphicsTextItem {
   Q_OBJECT
   double m_width;
   double m_height;
+  bool m_noAllSelect;
 
 public:
   SchematicName(QGraphicsItem *parent, double width, double height);
@@ -35,11 +36,17 @@ protected:
   void keyPressEvent(QKeyEvent *ke) override;
   void contextMenuEvent(QGraphicsSceneContextMenuEvent *cme) override;
 
+  void reFocus();
+
 signals:
   void focusOut();
 
 protected slots:
   void onContentsChanged();
+  void onCopy();
+  void onPaste();
+  void onCut();
+  void onSelectAll();
 };
 
 //========================================================
--- a/toonz/sources/toonzqt/schematicnode.cpp
+++ b/toonz/sources/toonzqt/schematicnode.cpp
@@ -15,9 +15,16 @@
 #include <QMenuBar>
 #include <QPolygonF>
 #include <QDesktopWidget>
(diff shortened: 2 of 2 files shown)


justburner:
I'm not entire sure 100% that compiler will initializes all members to zero in classes but better be safe than sorry.

justburner:
@shun-iwasawa Just a heads up, I've dropped more info in #4255 in regard to the initial bug.

shun-iwasawa:
Note to self: The issue had been reported in [QTBUG-88309](https://bugreports.qt.io/browse/QTBUG-88309) and [QTBUG-89563](https://bugreports.qt.io/browse/QTBUG-89563). The fix had been posted in [this commit](https://github.com/qt/qtbase/commit/4df5f93018344f6cdc6cd5a08a084b1c61e0c076).

@justburner 
Thank you for tackling with the further modification.
It's just an idea, but how about using `QGraphicsProxyWidget` containing `QLineEdit` instead of inheriting `QGraphicsTextItem` ?

justburner:
Oh my!
Where do i start? xD

Not only the popup is inside the graphics but also is affected by scaling (very amusing).
`QLineEdit` is rendered below the node's buttons, this issue also exists `QGraphicsTextItem` but isn't so obvious.
If too much zoomed in the popup won't show up but QLineEdit lose focus.

No contextMenuEvent override:
- Easy to get into the "zombie state", Tahoma2D is also affected by this.

Use createStandardContextMenu and menu->exec:
- Since popup scale is incorrect it also points to incorrect position when invoking it at different zooms other than 100%.
- Forcing setFocus fails to clear the "zombie state" a few specific times.

shun-iwasawa:
Oh.. I confirmed the huge context menu here as well ! 


Please forget the idea. Sorry for the trouble!

justburner:
I've found a bug related to `menu->exec`, i'm trying to work around atm...

justburner:
Okay, that issue is now gone by using `QMenu::popup()`.
I've refactored the code a bit so likely need a new review.

Added `Delete` on context-menu and actions are now enabled based of availability.

Additionally pressing Escape on node name restore the previous name before editing.

Clicking the name somewhere or pressing Enter auto accept name as normal.

Accepting an empty name now result in resetting back into the initial name from `setName()` (unless i misunderstood the role of setName() wrong? ... if so I can revert that back).

shun-iwasawa:
Confirmed the fix. Thank you @justburner ! LGTM

PR #4276, opened 2022-02-24
fix cell mark default names to be translatable

### LINKED ISSUE #4275: Cell Mark colors translation
@shun-iwasawa 
While translating for v1.6 I've noticed there are new strings for the Cell Mark default colors, but then they don't really show translated.

shun-iwasawa:
This fixes #4275 . There is no change in the translation files. 
Thank you @gab3d for reporting the issue!


RodneyBaker:

PR #4277, opened 2022-02-24
Fix crash on switching scenes while selecting raster level frame

### LINKED ISSUE #3940: Creating a new scene with a shortcut will crash the program when you are in the Level Strip of a Raster Level.
**To Reproduce**
1. Create a new scene and a new raster layer.
2. Enter the Level Strip of the raster level.
3. Close the scene (more precisely, create a new scene) using the shortcut (I have Ctrl + W).
4. After closing the dialog box, the program crashes.

The steps described do not always lead to a crash, for me it happens with a probability of 50%.

ezgif-2-42db33881f40

This glitch is missing in Tahoma2D. 

**Device Information:**
 - OS: Windows 8.1

shun-iwasawa:
This PR fixes #3940 .
The crash occurs when switching scenes the Toolbar tries to access invalid current level to update tool availability. 
In most of the cases except the one reported in the issue, the current level is properly updated before using it.
To avoid crash, I added lines to temporary clear the current level before switching scenes. 
Thank you @beeheemooth for reporting the crash!

@manongjohn 
I've tested the current nightly build of Tahoma2D and confirmed the same issue.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1316,6 +1316,10 @@ void IoCmd::newScene() {
 
   if (!saveSceneIfNeeded(QApplication::tr("New Scene"))) return;
 
+  // temporary clear the current level to prevent UI to access deleted level
+  // while switching scenes
+  app->getCurrentLevel()->setLevel(nullptr);
+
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #4278, opened 2022-02-24
Fix file browser assertion failure

shun-iwasawa:
This PR resolves an assertion failure when switching a scene while opening the File Browser.
The failure is just skipped when running in the Release mode (so this has not been a problem in the released software) but was quite troublesome for me when developing in the Debug mode.

The failure had been occurred in `DvDirModelRootNode::updateSceneFolderNodeVisibility()` .

According to [the c++ manual](https://www.cplusplus.com/reference/vector/vector/erase/) :

> **iterator std::vector::erase (iterator _first_, iterator _last_)**
> Iterators specifying a range within the vector to be removed: [_first_,_last_). i.e., the range includes all the elements between _first_ and _last_, including the element pointed by _first_ **but not the one pointed by _last_**.

And, according to [the Qt manual](https://doc.qt.io/qt-5/qabstractitemmodel.html#beginRemoveRows) : 

> **void QAbstractItemModel::beginRemoveRows(const QModelIndex &_parent_, int _first_, int _last_)**
> The _parent_ index corresponds to the parent from which the new rows are removed; **_first_ and _last_ are the row numbers of the rows to be removed.**

Therefore, the _last_ positions specified in the above functions should not be the same.

### DIFF
--- a/toonz/sources/toonz/filebrowsermodel.cpp
+++ b/toonz/sources/toonz/filebrowsermodel.cpp
@@ -194,7 +194,8 @@ DvDirModelFileFolderNode::DvDirModelFileFolderNode(DvDirModelNode *parent,
 //-----------------------------------------------------------------------------
 
 bool DvDirModelFileFolderNode::exists() {
-  return m_existsChecked ? m_exists : m_peeks
+  return m_existsChecked ? m_exists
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #4279, opened 2022-02-24
Fix crash on creating a zerary fx in the camera column

### LINKED ISSUE #3153: Crash when creating zerary node after cleaning up camera column cell
Repro Steps:

1. Select a cell in the camera column, or the camera column's header
2. Execute the `Scan & Cleanup -> Cleanup` command, nothing will appear to have happened
3. Attempt to create a new zerary node

OT will crash.

OS: Win 10
OT Ver: Mar 5 2020 Nightly

shun-iwasawa:
This PR fixes #3153 .
It seems that the crash always occurs when creating zerary fx while selecting the camera column, regardless of the cleanup execution.
Thank you @WolfInABowl for reporting the crash!


RodneyBaker:

PR #4280, opened 2022-02-24
Fix crash on cleaning up zerary column

### LINKED ISSUE #3152: Cleanup Crash with zerary column header selected
Repro Steps:

1. Create a new zerary level, e.g. Checkerboard
2. Select the column's header (where its name is shown)
3. Execute the `Scan & Cleanup -> Cleanup` command

OT will crash.

OS: Win 10
OT Ver: Mar 5 2020 Nightly

shun-iwasawa:
This PR fixes #3152 .
The crash occurs not only when selecting the zerary fx column, but also other types of the columns which cannot be cleaned up (e.g. Palette level column).
Again, thank you @WolfInABowl for reporting the crash!


RodneyBaker:

PR #4281, opened 2022-02-24
Fix crash on rendering PaletteFilter Fx (and other fxs) with hyphen

### LINKED ISSUE #3111: Texture, PaletteFilter, and ColorBlending - crash when using a hyphen
gif example:
PaletteFilter crash

Repro Steps;
1. Create a Texture or PaletteFilter node, 
2. Create a ToonzRaster or ToonzVector level and connect to the created node's Source input, this is needed so the node computes. Otherwise nothing will happen.
3. In the node's Indexes box, enter a hyphen
4. Attempt to render

OT will crash

I encountered this when accidentally pressing - instead of 0

Edit:
This does not seem to happen with the ArtContour or CalligraphicLine nodes

Interestingly, Texture and PaletteFilter crash no matter the type of level connected, but ColorBlending does not crash with zerary levels.

OS: Win 10
OT Ver: 1.4 Feb 20 Nightly

shun-iwasawa:
This PR fixes #3111 .

### DIFF
--- a/toonz/sources/common/tfx/ttzpimagefx.cpp
+++ b/toonz/sources/common/tfx/ttzpimagefx.cpp
@@ -14,7 +14,7 @@ void parseIndexes(std::string indexes, std::vector<std::string> &items) {
   char seps[] = " ,;";
   char *token;
   if (indexes == "all" || indexes == "All" || indexes == "ALL")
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #4282, opened 2022-02-24
Fixes toward Xsheet parenting + enhancements

### LINKED ISSUE #4273: Parent from Xsheet, hangs when there's only 1 column in the scene
imagen

the image is pretty much self explanatory.

1. create New Scene.
2. draw something in the Viewer.
3. in the Xsheet, click on the "Table" word in the column header. menu opens showing two entries: "Table" and "Col 2" (which is empty).
4. select "Col 2".
Crash.

justburner:
Fixes #4273 crash by removing extra invalid column on parent list if scene only have 1 column.
Additionally this PR does:
- User can click the current parent/handle to close the list, previously it only closed after a change.
- `Table` is now translated.
- All cameras now show on parent's list.
- Removed object type and index to simplify reading, e.g. `Steve` instead of `Col 1 (Steve )`.
- Show a tiny color bar next to names to see the object type.
- Correct list height is now calculated.
- Handle B acts as default and won't show up in the column.
- Show a separator on parent handle location, dashed means default handle.

Other bugs were solved by borrowing code from Tahoma2D, additionally also does:
- Have parent color box match the column type of the parent. (Optional, can be toggled)
- Don't show sound and note columns into the parent list.
- Don't show any childs on the parent list.
- Don't show parent list on note column.
- List offset positioned right below the column header.
- List in scrolled Xsheet shows up in correct spot.

Column parent colors is disabled by default, to toggle go to `File > Preferences...`, `Xsheet` category and `Show Column Parent's Color in the Xsheet`.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -2293,6 +2293,8 @@ XsheetViewer {
   qproperty-MeshColumnColor: #594d75;
   qproperty-MeshColumnBorderColor: #423956;
   qproperty-SelectedMeshColumnColor: #656692;
+  qproperty-TableColor: #62628c;
+  qproperty-PegColor: #9f6e3c;
   qproperty-SoundTextColumnColor: #a7a7a7;
   qproperty-SoundTextColumnBorderColor: #8e8e8e;
   qproperty-SelectedSoundTextColumnColor: #adb9c0;
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -2293,6 +2293,8 @@ XsheetViewer {
   qproperty-MeshColumnColor: #594d75;
   qproperty-MeshColumnBorderColor: #423956;
   qproperty-SelectedMeshColumnColor: #656692;
+  qproperty-TableColor: #62628c;
+  qproperty-PegColor: #9f6e3c;
   qproperty-SoundTextColumnColor: #a7a7a7;
   qproperty-SoundTextColumnBorderColor: #8e8e8e;
   qproperty-SelectedSoundTextColumnColor: #adb9c0;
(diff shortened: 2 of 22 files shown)


RodneyBaker:
Just a note to say... still testing but looking good thus far!

RodneyBaker:
The only thing I notice is perhaps unrelated to this PR in that parenting assignments don't seem to stick for cameras when collapsing to and exploding from subxsheets.

Is that more of a 'new features' element we likely should not delve into this close to release candidate?
Not sure.

RodneyBaker:
Thank you @justburner for the fixes and xsheet parenting enhancement!

justburner:
sorry for not commenting earlier, i got distracted with other things...
@RodneyBaker can you create an Issue with those steps so it doesn't get lost and forgotten?

gab3d:
@justburner 
@shun-iwasawa 
Using 1.6RC I keep seeing the **Table** string untranslated... :thinking: 
Could it be just that a new string was generated for that, instead of reusing one of the existent ones? (of course if that´s the case, as this PR happened after I sent the translations file to Shun, it is untranslated).

imagen

justburner:
That's odd, i did manually translate that entry in particular in some of the languages, Spanish was included.

gab3d:
@justburner 
Thank you for checking it out.
If you upload here the TS files I could generate the QM ones.

justburner:
@gab3d You can always get up to date TS files directly from the source, here's the path for Spanish language: https://github.com/opentoonz/opentoonz/tree/master/toonz/sources/translations/spanish - this PR only touches `toonz.ts` so you only need to generate `toonz.qm`.

Btw, `Show Column Parent's Color in the Xsheet` remains untranslated so you might as well translate that.

gab3d:
@justburner 
Thanks, but it seems the TS files are not updated automatically, so those TS files from the link are really the previous ones.
Perhaps @shun-iwasawa could update them so we can translate that latest strings before release?

manongjohn:
@justburner Also a reminder on something I noticed when I was migrating this to T2D...

You updated the `.qss` files directly with new entries, but did not update the `xsheet.less` file with appropriate entries. The next time someone regenerates the `.qss` files from the `.less` files (preferred method), your entries will be lost and will result in black boxes.

gab3d:
@shun-iwasawa @justburner 
I've already translated the latest TS files, but apparently there's still an elusive "Table" string out there:
imagen

PR #4283, opened 2022-02-25
Fix crash on rendering ArtContour fx

### LINKED ISSUE #2663: ArtContour + Particles crash
Repro steps:

1. Create a new level of any type and draw something on it so it is not blank
2. Create the following setup in the FX Schematic
image
3. Attempt to render - crash

shun-iwasawa:
This PR fixes #2663 
ArtContour Fx could not handle the Controller's bounding box when it has infinite size.


RodneyBaker:

PR #4289, opened 2022-02-26
Fix audio: mixing, 24 bits playback. Add 32 bits WAV support.

justburner:
- Fixes stereo mixing by not processing samples in mono when converting between stereo formats.
- Fixes 24-bits playback.
- Add support for 32-bits (float) WAV playback.
- Audio Recorder: Added 192Khz frequency.
- Audio Recorder: Added 24-bits and 32-bits formats.
- OGG and FLAC can be imported now.

Only Windows platform tested, I'll appreciate to know if Linux and/or MacOS will handle 24-bits and 32-bits formats just fine.


justburner:
GCC didn't liked my cute pseudo 24-bits integer, oh well >_>'

It's not work-in-progress because it's pretty much done :p

justburner:
Since i wasn't aware that some of this changes were already implemented in Tahoma2D there's some conflicts with the structures.
My aim is for both forks have a unified code database so this needs serious refactoring, i'm closing this one.

PR #4272, opened 2022-02-22
Revise Spin Blur Fx Ino

shun-iwasawa:
**Please review & merge this PR after releasing V1.6**

This PR will revise the Spin Blur Fx Ino according to the request from our studio.

1. Added a new option `Uniform Length` to the `Type` parameter. 

    Here is an example :
    
    Source Image

    spin_blur1
    - **Accelerator** :  The spin angle is zero at `radius` position and becomes larger outward. 
    - **Uniform Angle** : Former `Uniform` . The spin angle is the same in all distance.
    - **Uniform Length** : A new option. The spin angle becomes smaller outward so that the blur length becomes the same in all distance.

2.  Added new parameters; **Ellipse Aspect Ratio** and **Ellipse Angle** to make elliptical blur.

    

    - **Ellipse Aspect Ratio** : specifies the ellipticity of the blur direction.
    - **Ellipse Angle** : specifies how the ellipse is inclined in degrees.

3. Revised the Fx gadgets.

    

    - Added a handle for changing Ellipse Aspect Ratio and Ellipse Angle.
    - Added a dashed-line field in the entire window which indicates the blur direction. The field is displayed only when the mouse cursor overs the gadget.
    
      spin_blur5

    - Modified the `Center` gadget shape so that it can be easily grabbed when the `Radius` is set to 0.

PR #4302, opened 2022-03-07
Revise Radial Blur Fx Ino

shun-iwasawa:

This PR will revise the Radial Blur Fx Ino.
This PR is based on #4272 (thus, merging this PR will automatically merge #4272) .

1. Added a new parameter `Type` with 2 options as follows:
    
    - **Accelerator** : The blur intensity is zero at `radius` position and becomes larger outward. (Conventional behavior)
    - **Uniform Length** : A new behavior. The blur intensity becomes the same in all distance.

2. Added new parameters; **Ellipse Aspect Ratio**, **Ellipse Angle** and **Intensity Correlation** to make elliptical blur.
    

    - **Ellipse Aspect Ratio** : specifies the ellipticity of the blur direction.
    - **Ellipse Angle** : specifies how the ellipse is inclined in degrees.
    - **Intensity Correlation** : specifies relation between the ellipse shape and the blur intensity. When more than 0, the blur intensity along the long axis becomes bigger and vice versa.
        

3. Revised the Fx gadgets.

    

    - Added a handle for changing Ellipse Aspect Ratio and Ellipse Angle.

    - Added a handle for changing Twist.

    - Added a dashed-line field in the entire window which indicates the blur direction. The field is displayed only when the mouse cursor overs the gadget.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1172,6 +1172,8 @@
   <item>"STD_inoSpinBlurFx.alpha_rendering"	"Alpha Rendering"	</item>
   <item>"STD_inoSpinBlurFx.anti_alias"	"Anti Alias"	</item>
   <item>"STD_inoSpinBlurFx.reference"	"Reference"	</item>
+  <item>"STD_inoSpinBlurFx.ellipse_aspect_ratio"	"Ellipse Aspect Ratio"	</item>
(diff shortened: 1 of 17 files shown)


RodneyBaker:
Although I recall reading about it I forgot about merging this PR in order to merge the other.

PR #4303, opened 2022-03-07
Custom dpi option for Camera Capture

shun-iwasawa:

This PR will add an option to specify the dpi value to the captured image by the Camera Capture feature.

Now the DPI option can be selected from the following items:
- **Auto** : Conventional behavior. If the subcamera is not active, apply the current camera dpi. If the subcamera is active, compute the dpi so that the image will fit to the camera frame.
- **Custom** : A new option. Always use the custom dpi specified in the field.

The new option was introduced based on a situation in some animation studio; they draw characters on the paper which is smaller than the actual camera frame size so that they can handle the papers easier. (For instance, the camera frame size is in "cinema scope" but the characters are drawn on "vista vision" sized paper.)


RodneyBaker:
LGTM

PR #4292, opened 2022-02-28
Fix high dpi multimonitor glitch

shun-iwasawa:
**This PR needs to be checked carefully if it won't slows the performance.**

This PR fixes #3687 , the long-long-time bug with working on the multi monitor environment with different device-pixel ratio.
Now the device pixel ratio will be obtained from the screen where the widget is located.

PR #4320, opened 2022-03-13
Fix crash on Fx Schematic

shun-iwasawa:
This PR will fix the problem [reported](https://github.com/opentoonz/opentoonz/pull/4292#pullrequestreview-908090680) by @manongjohn , introducing the proposed fix (removing unused line).
Thank you @manongjohn !


RodneyBaker:
Thank you @manongjohn for the fix!

PR #4322, opened 2022-03-14
Fix tile fxs & bounding box handling

### LINKED ISSUE #4321: Crash with Tile Fx Iwa in Macro Fx
**Describe the bug**
Tile Fx Iwa causes crash under some (very) specific condition. 

**To Reproduce**
(I made the example as simple as I can so it looks quite odd.)
1. Load some levels (`BG` and `A`) and build Xsheet like as follows:
    tile_crash1

1. Build Fx schematic like as follows:
    tile_crash2

1. Make macro fx like this:
    tile_crash3

1. Preview frame 3.  OT crashes.

**Expected behavior**
No crash. Just render empty frame.

**Device Information:**
 - OS: Win10
 - OpenToonz Version: latest

shun-iwasawa:
This PR fixes #4321 , adding check if the source input of the Tile Fx Iwa is empty.
Also modified `TRasterFx::getBBox` to convert the bounding box only when the bounding box is not empty in order to prevent unnecessary further computation.

### DIFF
--- a/toonz/sources/stdfx/iwa_tilefx.cpp
+++ b/toonz/sources/stdfx/iwa_tilefx.cpp
@@ -134,6 +134,11 @@ void Iwa_TileFx::transform(double frame, int port, const TRectD &rectOnOutput,
                            TRectD &rectOnInput, TRenderSettings &infoOnInput) {
   infoOnInput = infoOnOutput;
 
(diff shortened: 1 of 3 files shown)


RodneyBaker:
LGTM

PR #4323, opened 2022-03-14
Deactivate Edit Shift mode when deactivating Shift and Trace

### LINKED ISSUE #4315: Shift and Trace: Disable when Edit Shift enabled.
When using the **Edit Shift** mode of Shift and Trace, and then disable the **Shift and Trace** option, the Edit Shift mode remains active.
I think it should get disabled or hidden in some way.

shun-iwasawa:
This PR resolves #4315 .
Now the `Edit Shift` mode will be released when the Shift and Trace mode is deactivatied, with keeping the `Edit Shift` checkbox state unchanged.


RodneyBaker:
Thank you for the adjustment @shun-iwasawa!

PR #2289, opened 2018-09-11
Fix antialiasing

blackwarthog:
This PR fixes anti-aliasing for strokes.
Issues #1166 #2285
Driver independent AA for regions not implemented yet.

PR #4333, opened 2022-03-15
Fix vector levels to render aliased line

### LINKED ISSUE #3101: Disable antialiasing on vector levels
By default, the program smooths out vector lines when rendering them, but it provides no option to keep them pixelated.
In (Render->Output Settings), there is no setting to control antialiasing. 
In (Level->Export Level), there is a "No Antialias" option, but it seems to have no effect.

### System Information
* **OpenToonz Version**: 1.4

shun-iwasawa:
This PR fixes #3101 .
Now the `Export Level` command will output aliased line when `No Antialias` option is set.
I think the problem was due to my oversight when introducing the modification #2289 . I hadn't imagined that there is a case that aliased line is needed.
In this PR I reverted the previous code for drawing "aliased" outline of vector stroke.

And also I made OT to render jaggy image from vector levels if `Output/Preview Settings > Resample Balance` is set to `Closest Pixel (Nearest Neighbor)` .

**N.B. Please note that currently the fix is only applied to "normal" style. Vector image drawn with special styles will still be antialiased.**

### DIFF
--- a/toonz/sources/common/tvrender/tsimplecolorstyles.cpp
+++ b/toonz/sources/common/tvrender/tsimplecolorstyles.cpp
@@ -311,6 +311,21 @@ void drawAntialiasedOutline(const std::vector<TOutlinePoint> &_v,
   outline.finish();
 }
 
+void drawAliasedOutline(const std::vector<TOutlinePoint> &_v,
+                        const TStroke *stroke) {
+  static const int stride = 2 * sizeof(TOutlinePoint);
+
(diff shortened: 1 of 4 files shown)


RodneyBaker:
This is a very welcome fix and one much anticipated by many users.

RodneyBaker:
I was hoping we might get a little more verbal feedback on this one as it has been a subject of much interest in the past.
Everything looks to be working as advertised here.

RodneyBaker:
Will merge as this LGTM

Issue #4349, opened 2022-03-20
White pixels along .tga files when move camera?


F-Burning:
I have this problem when import .tga files, do camera work and then the preview.

![o](https://user-images.githubusercontent.com/74098734/159160108-51b2acf4-0c6b-439a-9247-880d537aa638.jpg)

I've tried with "premultiply" and "white as transparent" in levels settings, but nothings happens.

RodneyBaker:
It certainly looks to be related to Premultipled Alpha.
Perhaps for testing you can share the image (or one similar to it that behaves in similar fashion)?

F-Burning:
Hi @RodneyBaker, thank you for help me.

I was trying now again and discover the error. It was because the loaded files did not have the same DPI of the camera, and I did change it on Preferences -> Loading -> Use camera DPI for all imported images. I also check the "premultiply" option in the level settings just in case, and now looks fine.

Before and after

![1](https://user-images.githubusercontent.com/74098734/159163222-ef8a26b1-11b4-4941-bb9d-d003da6d8892.jpg)

![2](https://user-images.githubusercontent.com/74098734/159163228-e00b72a3-6dc9-4724-a3ca-8be88a2be31f.jpg)

The image was not created in Opentoonz by the way.

Now this is solved I don't know if keep the issue opened to people who could has the same question, or close it. I leave it in your hands Rodney, sorry for the inconvenience, and once again thank you.

RodneyBaker:
If we keep issues open that have been resolved we'll soon have millions of open reports.
Better to close.  

That's what search is for... to find relevant topics related to issues we are having.
A good search will turn up results both open and closed depending on the criteria used.

So, yes... recommend closing.

F-Burning:
You're right.

Before I close it, I would like to report another "bug" I just noticed.

I have this aliased level. It's a .tga format and was loaded into Opentoonz (it wasn't created on it)...

![01](https://user-images.githubusercontent.com/74098734/159167188-0bc20923-40a7-4360-8751-e0ad65c1b808.jpg)

...and this is the preview result before compositing with camera work. I have activated "Use camera DPI for all imported images" in preferences.

![03](https://user-images.githubusercontent.com/74098734/159167256-1b5d417b-e616-4716-85d4-fc31a219d470.jpg)



The steps I following was:

1. I loaded the .tga into Opentoonz and check "Premultiply" option.
2. I expose the level in the Xsheet and first column was created.
[...]

PR #4351, opened 2022-03-20
Fix RGB key fx affine transformation handling

shun-iwasawa:
This PR will resolve a problem reported by @F-Burning in [this comment](https://github.com/opentoonz/opentoonz/issues/4349#issuecomment-1073268304) .
Modified a return value of the function `RGBKeyFx::canHandle()` so that the effect is applied before the column's transformation.

### DIFF
--- a/toonz/sources/stdfx/rgbkeyfx.cpp
+++ b/toonz/sources/stdfx/rgbkeyfx.cpp
@@ -51,7 +51,7 @@ public:
   void doCompute(TTile &tile, double frame, const TRenderSettings &) override;
 
   bool canHandle(const TRenderSettings &info, double frame) override {
-    return true;
+    return false;
   }
 };
 


F-Burning:
That was fast!

Thank you so much Mr. @shun-iwasawa 😄

RodneyBaker:
Thanks for the fix quick turn around fix @shun-iwasawa!
And thanks @F-Burning for the report.

F-Burning:
Hi.

The issue persist applying RBG Key, but this time there is not camera work.



Here is the drawing if you wanna try.

A1

RodneyBaker:
There is something strange about this PNG image...

F-Burning:
Yes, I've tried with another TGA image and now works fine, so looks like the problem is in that PNG...or could be the cyan color? I don't know.

Anyway, the fix was fine. Thank you again.

shun-iwasawa:
The uploaded PNG image was indexed color image which is not available in OT. Please convert the image to RGB color.

F-Burning:
It's strange, because I loaded the image in GIMP, and it's in RGB.

RGB

F-Burning:
This is new 😅

https://user-images.githubusercontent.com/74098734/159265262-a47cf04b-47a5-44f8-89ed-18bbe1dc806e.mp4

PR #4361, opened 2022-03-24
Display blank frames in camera stand view

shun-iwasawa:

This PR will enable to display blank frames between each viewer playback even if the viewer is in the camera stand view (i.e. non-preview) mode.
When the blank frame is set & activated, the whole viewer will now be filled with the blank frame's color instead of "frozen" at the last frame.


RodneyBaker:
A note to anyone testing this...
Initially I was having trouble testing this PR but it was because I was unfamiliar with the blank frames feature and had forgotten how to activate the frames.  Once activated everything works as expected.

@shun-iwasawa Is there an option I am missing to have the blank frames at the beginning rather than the end of the sequence?  I forget.

As for this PR... LGTM.

PR #4362, opened 2022-03-25
Fix the parent object label in xsheet column header translatable

shun-iwasawa:
This PR resolves a problem pointed by @gab3d in [this comment](https://github.com/opentoonz/opentoonz/pull/4282#issuecomment-1078676677) .

Now the column header and the context menu share the same string (obtained by `ChangeObjectParent::getNameTr()`) with the same translation.


RodneyBaker:

gab3d:
I've just tested the artifact, and it finally works! 🥳 
Thank you Shun-san!!

p.s. I'll upload the updated translations files in another issue.
p.s. 2  I also discovered, columns can't be parented to the Camera from the drop-down box.

PR #4363, opened 2022-03-25
Fix Preferences popup combobox size adjustment

shun-iwasawa:
This PR modifies size adjustment policy of the following comboboxes in the preferences popup so that they always adjust to their contents.
- `Interface > Font Style`  (Adjust size when the current font family is switched)
- `Loading > Level Settings by File Format` (Adjust size when a new entry is added or existing entry is removed)


RodneyBaker:
Thank you @shun-iwasawa for the fix,

PR #4371, opened 2022-03-27
Fine tune Chinese translation

jcome:
- Update Chinese translation according to a local animation artist's suggestions  
- Unify "Toggle" translation
- Fix typos


RodneyBaker:
LGTM (Not that I can understand the language!)
Thanks for the translation fine tuning @jcome

PR #4374, opened 2022-03-28
Fix parent object field in the xsheet column header

### LINKED ISSUE #4365: Can't choose any Cameras from parent drop-down in Xsheet.
@shun-iwasawa 
As mentioned in #4362 it seems we can't currently choose any cameras from the parent drop-down menu in the Xsheet.

shun-iwasawa:
This PR fixes #4365 .
Changed the field to directly keep `TStageObjectId` (which is actually unsigned int) as an item data instead of the text for it.


RodneyBaker:

PR #4376, opened 2022-03-28
Camera capture: Subcamera preset feature

shun-iwasawa:
subcamera_preset

This PR will enhance the Subcamera option in the Camera Capture popup as follows:
- Added the position (X and Y) fields.
- Enabled to save / load preset from the context menu of the Subcamera button. 
    - The presets will be saved in `$TOONZPROFILES/layouts/settings.username/camera_capture_subcamera.ini` .

**EDIT**
- Now the preset will save not only the subcamera geometry but also the dpi as well.
- Subcamera geometry are now stored in the user .env file and it reproduces the previous position & size on launch.

### DIFF
--- a/toonz/sources/common/tapptools/tenv.cpp
+++ b/toonz/sources/common/tapptools/tenv.cpp
@@ -172,9 +172,9 @@ public:
     m_applicationFullName = m_version.getAppName() + " " + m_applicationVersion;
     if (m_version.hasAppNote())
       m_applicationFullName += " " + m_version.getAppNote();
-      
-    m_moduleName          = m_version.getAppName();
-    m_rootVarName         = toUpper(m_version.getAppName()) + "ROOT";
+
+    m_moduleName  = m_version.getAppName();
+    m_rootVarName = toUpper(m_version.getAppName()) + "ROOT";
 #ifdef _WIN32
     // from v1.3, registry root is moved to SOFTWARE\\OpenToonz\\OpenToonz
     m_registryRoot =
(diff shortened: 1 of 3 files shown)


RodneyBaker:
@shun-iwasawa This PR looks to need a rebase/update perhaps related to merging the other PR related to camera capture.

RodneyBaker:
I've been testing this PR and it is a great addition to camera capture.
A very nice companion to the DPI feature as well.

RodneyBaker:
@shun-iwasawa for some reason when I attempt to save additional presets/subcameras all the presets seem to save to the center of the screen and do not maintain their x and y coordinates.  I'm still testing and this may be the result of something I am doing wrong.

To test, try to set up 4 subcamera presets (I named them 1, 2, 3 and 4 and have each occupy one quarter of the camera frame.  In my tests the size seems to be maintained but the position reverts to being centered for all but the first preset (named 1 in my case).

The current ini file I'm testing contains:

[1280x720]
1="0,0,725,432,300"
2="556,0,725,432,300"
3="0,289,725,432,300"
4="556,289,725,432,300"

shun-iwasawa:
Fixed the subcamera range limit not to over the camera resolution.

Thank you @RodneyBaker for pointing the bug!

RodneyBaker:
That's got it.  Thanks for the fix.
I'll continue testing while the final checks continue but looking very good thus far.
This will be a useful addition to camera capture/line testing in Opentoonz.
Much appreciated.

RodneyBaker:
The Mac Mojave checks appear to be frozen.
Will merge as this should not effect the nightly release.

PR #4380, opened 2022-03-31
Fix missing mouse release

### LINKED ISSUE #4355: animation tool drags without clicking
**Describe the bug**
Found a bug where when you select a column through control point editor tool (auto select drawing) and there are multiple vector columns on top of each other, you click and quickly go to animate tool. your cursor will drag the column without you pressing the button

**To Reproduce**

1.click with conttrol point editor where 2 vectors are in the same place in different columns with the "auto select drawing" on
2.change to animation tool through shortcut on keyboard
3.

 - OpenToonz Version: 1.6-RC

OpenToonz_NE1OiUP5jc

shun-iwasawa:
This PR fixes #4355 .
When the menu opens with the left click, `mouseReleaseEvent` is not called properly which causes OT to misunderstand that the mouse button is not released.

I added a check in `SceneViewer::onMove()` if the mouse button is actually pressed.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -509,6 +509,14 @@ void SceneViewer::mouseMoveEvent(QMouseEvent *event) {
 void SceneViewer::onMove(const TMouseEvent &event) {
   if (m_freezedStatus != NO_FREEZED) return;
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thanks @shun-iwasawa for the fix (and @age3rcm for the report with reproducable steps).

PR #4385, opened 2022-03-31
Modify default workspace layout

### LINKED ISSUE #4319: Can the room layout be quickly tidied before the release of 1.6?
I've just installed the 1.6RC and I hoped the room layouts would have changed. They're still the same as 1.5 (as far as I can see). I've not got a big issue with the content of the rooms (it's always a bone of contention, so I don't want to open that box again), but some of the panels in these rooms are oversized. If we're going to keep the content, could we set them to a better default size before the final release? I know some of these comments will be small changes, but it won't affect the release (no code changes), but will help new users.
My resolution is 1920x1080. I've included screenshots of the initial layout of each room that I think could be tidied and my suggestion:

**Drawing room:**

- The timeline could be slightly shorter. It only needs enough initial space to show the camera column and 2 or 3 other columns.
- The palette and style editor column could be narrower
 
Current:
image
To change to:
image

**Palette Room**

- The style editor column could be much narrower

Current:
image
To change to:
image

**Ink n Paint**

- The level strip is far too wide. The width will never be used because it's restricted to the thumbnail size.

Current:
image
To change to:
image

**Animation room**

- The level strip is too wide

Current:
image
To change to:
image

Everything else is set to a fair size.

Thanks for the consideration in advance.

shun-iwasawa:
This PR resolves #4319 

- Modified the default workspace layout according the feedback by @DarrenTAnims 
- Made the main window to be maximized on the first launch


RodneyBaker:
The main layout that still seems to be pushing windows into too small spaces appears to be in the InknPaint room.
I need to test more however and also wait for appveyor built to test on.

Here's an example of what I'm seeing for the InknPaint room:

image

Note that many have suggested it would be best for the initial layout to be locked down as well (see red arrow).
In the attached image I performed a Layout Reset, restarted and then locked down the layout.
The reset returned the windows back to their thin location on the right after I had made them larger.

shun-iwasawa:
@RodneyBaker 
I updated the layout with FHD (1920x1080) monitor size.
It seems that you are using FWXGA (1366x768) monitor, which causes some panels to be shrunk.

Probably in the future OT should prepare several types of default room settings adapting any kind of monitor size.
For now, at the last minute of releasing the next version, I think all I can do is to decide the most typical monitor size for which the default layout will be made.

DarrenTAnims:
Thank you for updating this @shun-iwasawa. I think this will really help new users to feel comfortable with the software from first load.
Did you manage to include the rooms being locked when first running, too? From the Windows > Lock Room Panes. As Rodney says, it's easy for new users to try to resize panels and pull them away from their docked position.

PR #4388, opened 2022-03-31
[WIP] Fix "Visualize Vector as Raster" behavior

### LINKED ISSUE #4153: "Visualize Vector As Raster" missing behaviour.
When using the "Visualize Vector As Raster" option the selected (vector) level doesn't visualize as vector, as the [manual states here](https://opentoonz.readthedocs.io/en/latest/drawing_animation_levels.html#adjusting-the-viewer-visualization).

imagen

I think that missing behaviour of the function potentially makes it way less powerful, as a user wanting to edit a vector level will have to switch the option OFF to be able to see it with good quality. But in doing so, the Viewer can potentially get slow (supposing the reason to activate it in first place was to speed it up, because of a slow performance, most probably due to many complex vector levels displayed).

Perhaps the code is still there, and could be reactivated?
As always, thanks for listening!

shun-iwasawa:
This PR will fix #4153 ... but not finished yet.
Currently this PR is WIP as there is a problem on refleshing the image when switching the current columns.
Will continue after releasing V1.6 .

### DIFF
--- a/toonz/sources/toonzlib/stage.cpp
+++ b/toonz/sources/toonzlib/stage.cpp
@@ -428,11 +428,12 @@ void StageBuilder::addCell(PlayerSet &players, ToonzScene *scene, TXsheet *xsh,
     bool isCurrent = (subSheetColIndex >= 0)
                          ? (subSheetColIndex == m_currentColumnIndex)
                          : (col == m_currentColumnIndex);
-    player.m_isCurrentColumn     = isCurrent;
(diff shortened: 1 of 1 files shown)


F-Burning:
Hi.

This PR will be in the 1.7, or in other future releases?

RodneyBaker:
#5793 should also resolve #4807

RodneyBaker:
Closing this PR as implement via #5793 
Thanks @shun-iwasawa for gettting the ball rolling and @Abel032 for implementing.

PR #4390, opened 2022-04-06
Fix QTabWidget Pane Border in Stylesheets

konero:
Seen in the New Project window popup, currently has no style, this adds a style. It's not perfect but nicer than before.

OTfix


RodneyBaker:
Thank you @konero for the stylesheet fix.
LGTM

PR #4393, opened 2022-04-14
Make style editor textfield expandable

konero:
This makes the textfield in the style editor expandable like the sliders.

I also increased the padding on the buttons on the bottom for extra height since there's a lot of unused space. I think it was like this before and honestly can't recall why they were shrunk. Let me know if there's any problem.

Basically, I did this to increase the 'hit box' when using a small screen, like Cintiq 16 (we mostly use those now without second displays). Since we all don't work in-house, trying to communicate how to change the sizes using custom stylesheet function is not easy to communicate.

Before_StyleEditor After_StyleEditor

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -1512,7 +1512,7 @@ QStatusBar #StatusBarLabel {
   padding: 3 2 8 3;
 }
 #StyleEditor #bottomWidget QPushButton {
-  padding: 3 5;
+  padding: 13 5;
 }
 #HexagonalColorWheel {
   qproperty-BGColor: #414345;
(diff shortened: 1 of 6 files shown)


RodneyBaker:
Initial review for this PR looks good to me.

It would be good to hear from those that previously were involved in adjusting the Style Editor prior to this to make sure it meets their needs.

RodneyBaker:
We'll have to fight for feedback via actual usage.  :)

Thanks for the adjustments to the Style Editor @konero.
LGTM

A note regarding the Preferences Additional Style Sheets approach.
Perhaps in the future we can improve that feature with ability to open, edit and save directly via that process.
Then the user could be pointed to a file to load as a style sheet and they would just load/install it.
Another option being to perhaps just place the style sheet file into proper place in the stuff directory and OT would recognize it's presence and load the customized style preferences.  There is more to say about this but... some other time and place.  :)

PR #4394, opened 2022-04-15
Fix xsheet pdf export tick-marking bug

shun-iwasawa:
This PR fixes following problems in the `Export Xsheet to PDF` feature:
- If the pdf has more than one page, tick marks in the first page duplicately appear at the same positions in subsequent pages. (For instance, when using 3-seconds template, a tick mark at the frame 4 wrongly appears at the frame 76 in the second page as well.)
- The tick mark does not appear overriding the x-mark indicating the beginning of empty cells.


RodneyBaker:
@shun-iwasawa The nightlies don't appear to be building...

RodneyBaker:

PR #4396, opened 2022-04-16
Change issue templates to forms

konero:
This will use the new forms template that supports validation when making new issues. You can test it out on [my repository here](https://github.com/konero/opentoonz/issues/new/choose). I'll delete all issues if you post them, no worries.

The user will need to fill out some system information before they can post, this would I hope prevent the replies "what version are you using?" or "what OS are you running?".

**Example:**
image


RodneyBaker:
I like it.

For operating system there should probably be an 'other' entry or 'See remarks'... something to capture the inevitable report that doesn't fit the main three (win, mac and linux).

konero:
@RodneyBaker Thanks for checking.

> For operating system there should probably be an 'other' entry or 'See remarks'... something to capture the inevitable report that doesn't fit the main three (win, mac and linux).

We can add it later, for now I think only the three main operating systems are officially supported here in some capacity.

> It might be good to at least mention that optional attachments to include crash reports can be added to reports. OT doesn't currently have crash reports like Tahoma2D but some users will know how to get those.

Fair enough, I reworded the screenshots textarea to include crash logs.

I made it a requirement to fill out OpenToonz Version information and GPU, that's actually a really annoying thing to have to assume when offering help. I also trimmed some paragraphs, much leaner now.

RodneyBaker:
We'll proceed with this move to forms instead of templates and look forward to feedback on how well it works for reporters.

Thanks @konero for actively improving the reporting process.

PR #4398, opened 2022-04-18
Reword "Reverse sheet symbol" to "Inbetween symbol 2"

shun-iwasawa:
This PR will stop to use the word `Reverse sheet symbol` in the XDTS-related features and replace it by the `Inbetween symbol 2` . This change is made reflecting the voices from the production sites of Japanese animation industry.

### What was the "Reverse sheet symbol"?

The term `Reverse sheet symbol` had came from page 4 in the [XDTS file format specification](http://vd.clipstudio.net/clipcontent/paint/app/ToeiAnimation/XDTSFileFormat_en.pdf) .

It defines the instruction for `SYMBOL_TICK_2` as "Reverse sheet symbol（●）" (逆シート記号 in Japanese) . When I implement XDTS-related feature, I simply referenced them for UI labels in OT.

However, after hearing the voices from the production site, I learned the fact that **there is no "Reverse sheet symbol"** in the practical rules in Japanese animation industry.
As a matter of fact, when writing Xsheet, there is a direction for "Reverse sheet", but it is expressed with no special symbol. And the symbol "●" is used as a symbol for "_DO NOT_ Reverse sheet"!

### What actually is the "Reverse sheet"?

The "Reverse sheet" (逆シート) is one of the directions from key-frame animator to inbetweener.
When animate some swinging movement such as chewing mouth, key frame animators sometimes "reuse" the previous key frame. And write a "keyframe-xsheet" like as follows:

The dots "・" are the "inbetween symbols" where the inbetween drawings are to be inserted.
Please note that the symbols are all the same, even at the ”reverse playing” part.
Based on the keyframe-xsheet, inbetweeners will clean-up keyframes and insert the necessary inbetween drawings. Then write a "inbetween-xsheet" as follows:

As you can see, the inbetween drawings are reused at the ”reverse playing” part. Inbetweeners look at order of the previous and following keyframes and decide whether to insert a new inbetween drawing or to reuse the previous drawing.
And the part where the inbetween drawings are reused is called the "Reverse sheet".

### What actually is the filled circle symbol "●"?

Movements in which key frames are reused but different inbetweens are needed at the ”reverse playing” part (such as waving hair), key frame animators use the filled circle symbol "●" to let the inbetweeners to insert new drawings. The key and inbetween sheets are like as follows:

Unfortunately there seems to be no popular name for this symbol.

To recap, calling the filled circle symbol as the "Reverse sheet symbol" would be not collect from the point of view of the practical usage in Japanese animation industry. So I renamed the labels for the symbols to `Inbetween symbol 2` to avoid the confusion.


RodneyBaker:
It is always interesting to learn more about animation workflow in the Japanese Anime Industry.

Aside:  I'm not sure if nightlies are building yet but this seems a good PR to test that on.  :)

shun-iwasawa:
@RodneyBaker 
Thank you for looking and merging this PR. 
I'll reboot the Jen.kins machine tonight. Thank you for your patience!

PR #4400, opened 2022-04-19
Fix blank frames malfunction on loaded scene

shun-iwasawa:
This PR fixes a bug related to #4361 as follows:

- After loading a scene, blank frames in the camera stand view does not work properly. It works only on the initial scene after launching the software.


RodneyBaker:

PR #4401, opened 2022-04-19
Spin Blur & Radial Blur fx gadgets: Lock AR / angle with modifier keys

shun-iwasawa:
This PR is an addition to the recent modifications in Spin Blur Fx Ino and Radial Blur Fx Ino (#4272 and #4302).
Now the `AR and angle` handle in the fx gadget can move one parameter with locking another, by pressing modifier keys while dragging.
- +Shift key for locking the aspect ratio and only move the angle.
- +Ctrl key for locking the angle and only move the aspect ratio.


RodneyBaker:
Thank you @shun-iwasawa for the enhancements!

PR #4402, opened 2022-04-19
Patch Stylesheet (Tabs, Light Theme)

konero:
This is the second part to patching the tabs (#4390), so they don't look so broken in the New Project window, since we already had a rounded tabs class I swapped this to be the default, the flat style is then applied to the custom `TabBarContainer` widget which is utilized by Palette, Style Editor etc.

Also some very minor color balancing for the Light Theme only (*) I might as well include here.
- Darkens palette numpad group border so it's easier to see (like Retas PaintMan)*
- Makes text field disabled color look more faded*
- Darkens dock border (gap between panels) to give it more contrast against the viewer gray color*
- Add some styling for QHeaderView, QTreeWidget (these are found in Hex name editor panel)

image

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -124,43 +124,51 @@ QGroupBox {
   qproperty-BottomBelowLineColor: #262728;
 }
 .tab-flat,
-QTabBar::tab {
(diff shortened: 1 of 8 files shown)


RodneyBaker:
Thank you @konero for part 2 of your update.
LGTM

PR #4414, opened 2022-04-26
Fix missing PSD Layer id 

### LINKED ISSUE #4415: Error with PSD files without Layer ID
### Description

As Identified via https://github.com/tahoma2d/tahoma2d/issues/982
Some programs to include Photoshop and Krita can create PSD files without proper layer IDs.
@manongjohn created logic to address this issue and I confirm the problem does exist in Opentoonz.
Notice:  An important assumption in this fix is that PSD files will either have all Layer IDs or none.

### Steps to Reproduce

1.  Load example PSD file provided in discussion at https://github.com/tahoma2d/tahoma2d/issues/982
2.  Compare results of this file with that of other PSD files created in variety of different programs.

### Expected Behavior

Opentoonz should recognize and process PSD files appropriately from all common graphics programs that produce the PSD format.

JPUser1:
This port from Tahoma2D resolves #4415

Co-authored-by: manongjohn <19245851+manongjohn@users.noreply.github.com>

### DIFF
--- a/toonz/sources/common/psdlib/psd.cpp
+++ b/toonz/sources/common/psdlib/psd.cpp
@@ -220,8 +220,8 @@ bool TPSDReader::doLayersInfo() {
     m_headerInfo.layersCount = -m_headerInfo.layersCount;
   }
   if (!m_headerInfo.linfoBlockEmpty) {
-    m_headerInfo.linfo = (TPSDLayerInfo *)mymalloc(
-        m_headerInfo.layersCount * sizeof(struct TPSDLayerInfo));
+    m_headerInfo.linfo = (TPSDLayerInfo *)mycalloc(
+        m_headerInfo.layersCount, sizeof(struct TPSDLayerInfo));
     int i = 0;
     for (i = 0; i < m_headerInfo.layersCount; i++) {
       readLayerInfo(i);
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Of importance in this fix is John's statement about how the layer data initialization process is modified to work.
From from his original PR:  

```Modified the logic to effectively initialize all layer data elements to 0 and assign a layer Id based on layer read order when reading the file. **_If the file contains layer ids, it will replace it with the layer id found.`_**``

(Emphasis added)

I want to point this out so that no one thinks this fix discards the layer IDs already in place in a PSD file where they are found.

RodneyBaker:
Where I am with testing...

Confirmed the problem.
Confirmed this PR fixes the problem.
Investigating a crash shortly after successfully importing the test PSD file.

RodneyBaker:
Aside:

While testing some outliers I did run into a problem with PSD files but it is an issue well outside the scope of this PR.
Briefly, dynamic object attributes (for example:  editable text) must be rendered to object in the program creating the PSD file.
If not Opentoonz will not be able to open that part of the PSD file... but does apparently open everything else.
Programs such as Corel Photoshop warn the user to 'render to object' to ensure compatibility.

manongjohn (changes requested):
Not credited properly in commit

RodneyBaker:
> Not credited properly in commit


Yes,  looks like @JPUser1 attibuted you in #4413 but not here in this PR.

PR #4421, opened 2022-04-26
Fix missing PSD layer

JPUser1:
Whatever created the PSD file might not have written out Layer Ids and certain layer information when saving.

Due to the missing Layer Ids, the software was effectively assigning random values to layer ids and several other layer data elements every time it read the file for information. This is a problem for consistency between file reads and also if the layer id was negative, the layer would not be loaded into T2D.

Modified the logic to effectively initialize all layer data elements to 0 and assign a layer Id based on layer read order when reading the file. If the file contains layer ids, it will replace it with the layer id found.

Assumption: PSD file will either provide layer ids for all layers or none at all.
this was authored by @manongjohn https://github.com/tahoma2d/tahoma2d/pull/984
Thanks @manongjohn for the fix.


JPUser1:
Sorry about all the mess I am very new at this and currently still learning.

PR #4422, opened 2022-04-26
Fix missing PSD layer IDs

JPUser1:

Whatever created the PSD file might not have written out Layer Ids and certain layer information when saving.

Due to the missing Layer Ids, the software was effectively assigning random values to layer ids and several other layer data elements every time it read the file for information. This is a problem for consistency between file reads and also if the layer id was negative, the layer would not be loaded into T2D.

Modified the logic to effectively initialize all layer data elements to 0 and assign a layer Id based on layer read order when reading the file. If the file contains layer ids, it will replace it with the layer id found.

Assumption: PSD file will either provide layer ids for all layers or none at all.

Thanks @manongjohn for the fix.

### DIFF
--- a/toonz/sources/common/psdlib/psd.cpp
+++ b/toonz/sources/common/psdlib/psd.cpp
@@ -220,8 +220,8 @@ bool TPSDReader::doLayersInfo() {
     m_headerInfo.layersCount = -m_headerInfo.layersCount;
   }
   if (!m_headerInfo.linfoBlockEmpty) {
-    m_headerInfo.linfo = (TPSDLayerInfo *)mymalloc(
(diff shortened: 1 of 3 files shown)


JPUser1:
Again sorry for all the mess I am very new at this. : )

manongjohn:
> Again sorry for all the mess I am very new at this. : )

Well...looks like you finally figured out how to give proper credit and also combine all file changes under 1 commit. 👏 Now need to work on a more meaningful commit title. 😁  No need to redo it again, but something to think about in the future.

RodneyBaker:
You'll have this mastered yet @JPUser1

Besides a few rough edges such as titles and such the core of this port... LGTM.

Thank you @manongjohn for the fix and @JPUser1 for the port!

PR #4428, opened 2022-04-28
Update colornames.txt

JPUser1:
Just a few new extra default Hex color codes with English titles to come stock with the new Hex Editor feature.


RodneyBaker:
You should describe these changes (and additions) and why they are being made.

The current description suggests additional colornames are being added BUT there are a number of colors being removed entirely and OTHERS altered to different hex values.  A quick sampling of a number of these changes suggest the changes are not correct.  For example Teal and Silver.

JPUser1:
the ones I changed were to more common hex values they are all still there.

JPUser1:
Its ok if u do not care to merge it I just figured the more the merrier.

PR #4430, opened 2022-04-28
Auto close and auto fill options for vector brush

JPUser1:
This adds the option to automatically close or fill a vector stroke. If the stroke is filled, it will automatically be self grouped.

Thanks to @jeremybullock


JPUser1:
this is more of a test.

PR #4432, opened 2022-04-29
Fix Lazy Nezumi Pro hooking Bug

JPUser1:
This PR fixes the issue with Lazy Nezumi Pro incorrectly hooking to the main window instead of the canvas.

this should fix: https://github.com/opentoonz/opentoonz/issues/3812

Thanks @manongjohn again!

### DIFF
--- a/toonz/sources/toonzqt/imageutils.cpp
+++ b/toonz/sources/toonzqt/imageutils.cpp
@@ -895,6 +895,10 @@ FullScreenWidget::FullScreenWidget(QWidget *parent) : QWidget(parent) {
   layout->setSpacing(0);
 
   setLayout(layout);
+  
+#ifdef _WIN32
+  this->winId();
+#endif
 }
 
 //---------------------------------------------------------------------------------
@@ -1035,7 +1039,6 @@ bool FullScreenWidget::toggleFullScreen(
             this->window()->windowHandle()->setScreen(ptrScreenThisWindowIsOn);
 
             // http://doc.qt.io/qt-5/windows-issues.html#fullscreen-opengl-based-windows
-            this->winId();
             QWindowsWindowFunctions::setHasBorderInFullScreen(
                 this->windowHandle(), true);
 


RodneyBaker:
I'm not seeing any issue with Lazy Nezumi Pro and Full Screen here in Opentoonz.
If anyone is experienceing issues with this PR that would be very good to know.

manongjohn:
As @shun-iwasawa presumably saw the full screen issue in his build environment, I would suggest waiting for his feedback.

I suspect it might be some difference in the build environment.

Based on that I would recommend immediately checking the nightlies after this is merged in case the PR artifact doesn't exhibit the issue but the main build environments might.

RodneyBaker:
This does appear to fix the hooking bug with Nezumi Pro.
I do note that on occassion the hook is lost and the LN icon has to be dropped into the target window again.
Once dropped everything works again as expected.

It isn't clear to me what is allowing the hooked window to lose its connection.
At first glance it seems to be when drawing really fast without concern for where the pen tip is interacting with the screen.

What caused a relook at this issue for me was the release of a new version of Lazy Nezumi Pro.

RodneyBaker:
Call for more testing of this PR so if you are a Lazy Nezumi Pro user...

Code-wise the change here is simple and straightforward.
With regard to fixing the bug it seems the bug is fixed especially when using the lasted release of Lazy Nezumi Pro.
I personally need to do more testing to make sure I'm not missing anything.

When testing please be sure to use the update from 14 Oct 2022; Link:  https://lazynezumi.com/news

RodneyBaker:
Posting intent to merge as there are some brush drawing anomolies that are likely related to the github action artifacts.
Testing the nightlies will be the best way to determine what issues remain.

This PR very obviously addresses the basic hooking issue although I suspect there will still be issues with Full Screen mode that need to be addressed in later PR.

RodneyBaker:
Pressing forward with this PR as it requires deeper testing via the nightly release.

I have heard that some users have experienced white screens in full window mode.
This is likely with older release of Lazy Nezumi and is something we definitely need to verify if that is the case as some users will still be using older versions (of potentially both OT and Lazy Nezumi).

RodneyBaker:
Currently investigating report of lag being introduced with this PR which would appear unrelated to the Full Screen issue.

RodneyBaker:
Reverting this PR as that is perhaps the best way to get to the bottom of the question of whether the change leads to the noticeable lag (and artifacting) in drawing.

PR #3041, opened 2020-01-19
Remove unused toonzrle.cpp

otakuto:

PR #4443, opened 2022-05-11
Fix loading TZP

shun-iwasawa:
This PR fixes loading TZP files (Old level format used in Toonz4.x) by reintroducing the source file `toonzrle.cpp` which was removed in #3041 by my mistake.

I merged #3041 since the file was not linked any other sources, but now realized that it registers the decompression scheme for TZP files to LibTiff and is necessary for the TZP loading feature. So sorry for the trouble!


RodneyBaker:
@shun-iwasawa Windows builds are currently failing.
I intially thought it related to a typo but appears to be something else.

RodneyBaker:
Looks like everything is back on track.
Thank you @shun-iwasawa for the fix ...fixes!.

PR #4452, opened 2022-05-22
Github Actions: Update Windows Runner

shun-iwasawa:
This PR will resolve recent failure in the Github Actions, updating vcpkg version to the latest.
Note that packaging DLLs are also modified according to the update of OpenCV.


shun-iwasawa:
[nasm.us](https://www.nasm.us) seems to be down. I'll re-run the action as soon as it gets fixed.

RodneyBaker:
Thank you @shun-iwasawa!

PR #4454, opened 2022-05-25
Fix Margin of Spin Blur Ino Fx

shun-iwasawa:
This PR fixes Spin Blur Ino Fx, modifying the margin computation so that the blurred result won't cut off at the border of the source image's bounding box.


RodneyBaker:

PR #4455, opened 2022-05-25
Update Bokeh Advanced Iwa Fx

shun-iwasawa:
This PR will update the Bokeh Advanced Iwa Fx, enabling to specify whether to expand the border of the separated images when the reference depth image is used in some layer. The check boxes `Fill Gap`  and `Use Median (Filter)` are added to each layer. They work just like as the ones in the Bokeh Ref Iwa Fx.


RodneyBaker:
Mergining this but... the Windows build has not been propagating to the nightly listing.
We'll see if that is still the case.

PR #4200, opened 2021-12-26
FFMPEG GIF Export enhancements

justburner:
This PR mainly enhances GIF Export with new options but also provides related fixes.

Image

- Allow generating GIF with different modes and ditherings, setting up the maximum number of colors for the palette is now possible.
- Unchecked `Generate Palette` is now just one of the modes.
- Fixes tahoma2d/tahoma2d#831 and maybe fixes #2628 by calling FFMPEG only once during final stage, updating to latest FFMPEG is also recommended.
- Exported GIF resolution is now exactly as expected no matter how much transparent padding the animation has.
- When loading a scene, properties in file are now merged instead of discarding the program ones. *¹
- Adds `Close` button into file settings window for all formats.
- Adds "double" type in property for any future use (was planned to be used but currently unused).

*¹ When testing make sure you don't rewrite any important scene that will later be used in the unpatched version, the original behavior of OpenToonz will cause `Generate Palette` to be missing as side-effect of program options being discarded, it can be solved by deleting GIF formatProperties inside scene file.

PR #4456, opened 2022-05-25
Fix Mov Settings Updating

shun-iwasawa:
This PR fixes a problem which appears after introducing #4200 as follows:

- In the Output Settings, changes in the file format options of the Quicktime (.mov) are not saved. The options seems to be reset every time you open the format options popup.


RodneyBaker:

PR #4467, opened 2022-06-08
Make ruler marks based on pixels

JPUser1:
this PR will will make rulers marks based on pixels.

Co-Authored-By: jeremybullock <79284068+jeremybullock@users.noreply.github.com>

### DIFF
--- a/toonz/sources/toonz/ruler.cpp
+++ b/toonz/sources/toonz/ruler.cpp
@@ -58,7 +58,7 @@ double Ruler::getGuide(int index) const {
 //-----------------------------------------------------------------------------
 
 double Ruler::getUnit() const {
-  double unit = getZoomScale() * Stage::inch;
+  double unit = getZoomScale() * (Stage::inch / Stage::standardDpi);
   return unit;
 }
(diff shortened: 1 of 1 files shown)


toonz/sources/toonz/ruler.cpp
```diff
@@ -58,7 +58,7 @@ double Ruler::getGuide(int index) const {
 //-----------------------------------------------------------------------------
 
 double Ruler::getUnit() const {
-  double unit = getZoomScale() * Stage::inch;
+  double unit = getZoomScale() * (Stage::inch / Stage::standardDpi);
```
shun-iwasawa, line 61: In this implementation the ruler will display correct scale only when the DPI of the current output camera is set to `Stage::standardDpi` (=120 dpi).

How about making it to be preference-aware, i.e. making the ruler to be pixel based when the current unit is set to pixel?  In the pixel unit environment, the dpi of the camera and all images are always set to 120 in the viewer.

JPUser1 (commented):
sounds good to me.

PR #4475, opened 2022-06-29
Fix SoapBubble Iwa fx

shun-iwasawa:

**To reproduce:**
- Create a SoapBubble Iwa Fx. Load or create some level node.
- Connect the level node to the both the `Thickness` and the `Shape` ports of SoapBubble Fx.
- Set the `Preview Settings > Channel Width` to `16bit` .
- Preview the SoapBubble Fx. Fx fails to render the result.

Thickness port will always request the image with 8bit channel width (since it will be bynalized afterwards).
Before the fix, 8bpc image had been stored in the cache with the same alias as the 16bpc image, which caused an exceptional error.


RodneyBaker:
I believe the nightly builds (for Windows) are still having issues but will press forward with this PR.

PR #4483, opened 2022-07-24
OCA Export

justburner:
Support for exporting Open Cel Animation (OCA) format, well, to some extend...
There are missing features and limitations mostly due to how OT pipeline works compared to OCA capabilities.

**What's supported:**
☑️ Layers
☑️ Layer Groups - _As sub-xsheets only_
❌ Layer Labels - _OT doesn't support labels, cell marks or color filters aren't alternatives_
☑️ Layer Visibility- _OT cam stand and Opacity for whole layer_
❌ Blending Modes - _OT uses Nodes, implementation is extremely complex_
❌ Layer Alpha options (inherit alpha) - _OT doesn't support it as far I know_
☑️ Keyframes and their duration (animation exposure) - _Duration of a cel, unrelated to OT actual keyframes_
❌ Opacity Keyframes - _OT uses Nodes, implementation somehow complex_
☑️ Layer Sizes and Coordinates
☑️ Document background color
☑️ Document color depth - _U8 for PNG or U16 for EXR_
☑️ Duration
☑️ Framerate

**Limitations:**
- OT keyframes information can't be stored in OCA so they do nothing.
- OT FXs won't work, obviously.
- OCA doesn't support 2D or 3D transformations, all animation keyframes are discarded.
- OCA doesn't support palette or similar information so styles are discarded.
- All levels on a column need to match same dimensions.
- Sub-xsheet needs to be on it's own column.
- Anything related to nodes is very complex to solve as nodes can be wired and combined in all possible ways, best alternative is if OT support the features directly on a Level?

### DIFF
--- a/stuff/profiles/layouts/rooms/Default/menubar_template.xml
+++ b/stuff/profiles/layouts/rooms/Default/menubar_template.xml
@@ -28,6 +28,7 @@
         <menu title="Export">
             <command>MI_SoundTrack</command>
             <command>MI_ExportXDTS</command>
+            <command>MI_ExportOCA</command>
             <command>MI_StopMotionExportImageSequence</command>
             <command>MI_ExportTvpJson</command>
             <command>MI_ExportXsheetPDF</command>
--- a/toonz/sources/include/timage_io.h
+++ b/toonz/sources/include/timage_io.h
@@ -227,6 +227,7 @@ private:
   TImageWriter &operator=(const TImageWriter &src);
 
 public:
+  void setFilePath(const TFilePath &path) { m_path = path; }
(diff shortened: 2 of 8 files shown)


RodneyBaker:
Initial look at this has everything working as advertised.
I do not have access to After Effects so would appreciate someone who does testing with that.
I do not anticipate any issues there.

justburner:
I don't have After Effects either but from all my testing with Blender it worked as expected.

RodneyBaker:
Opentoonz (and Tahoma2D) now have both OCA import and export capability.

For information on the import feature please see: #5315

PR #4484, opened 2022-07-26
Existing scenes list in Startup Popup

justburner:
Modify the Startup Popup to list all existing scenes of a project as thumbnails.
Additionally allows to open a project from any folder and explore the folder content of the current project.

StartupExistingScenes

... also I think it makes a nice modern looking welcome after opening OpenToonz, and it looks cute to see all the thumbnails when looking inside a large project. :)

### DIFF
--- a/toonz/sources/toonz/startuppopup.cpp
+++ b/toonz/sources/toonz/startuppopup.cpp
@@ -16,6 +16,7 @@
 #include "toonzqt/menubarcommand.h"
 #include "toonzqt/gutil.h"
 #include "toonzqt/doublefield.h"
+#include "toonzqt/icongenerator.h"
 
 // TnzLib includes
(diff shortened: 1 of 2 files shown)


RodneyBaker:
It appears to me that appeveyor needs to be restarted.
Currently I believe @shun-iwasawa may be the only one that can do that (short of perhaps @manongjohn).

RodneyBaker:
Thank you @justburner for the enhancement to the Startup popup menu!
LGTM

As discussed on discord there are a few things I personally would like to see adjusted (or adjustable) for the previews:
- Size of preview (as details can be hard to see.  Of course the same can be said for Browser and Scene Cast previews)
- Background of preview (as black ink on black showing through transparency defeats the purpose of the preview)

PR #4485, opened 2022-07-26
Fix memory corruption caused by rapid project/scene change

justburner:
This one is really weird... 😕 

Each time a project changes, a new scene is created and the old scene gets immediately deleted from memory.

From what I undestand the issue seems that Qt signals can occur after that and OT will be accessing and/or modifying memory that is supposed to be now free and reused for something else, causing memory corruption that could lead to weird behaviors or ultimately crashing OT.

**To reproduce the crash:**
Have a couple of projects in your stuff directory, the more the easier for the crash to occur.
In Startup Popup select the Current Project ComboBox and hold UP and DOWN arrows on keyboard to cause OT to rapidly change projects as fast as it can.
After so many project changes in quick succession there's a chance for OT to crash.

This PR delay the deletion of the old scene by a few secs, enough for the large spam of signals to finish in time. This is more than a hack than a proper solution sadly, on my computer at least the crash stopped with this patch but mileage may vary.

### DIFF
--- a/toonz/sources/toonzlib/tscenehandle.cpp
+++ b/toonz/sources/toonzlib/tscenehandle.cpp
@@ -1,4 +1,4 @@
-
+#include <QTimer>
 
 #include "toonz/tscenehandle.h"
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I'm unable to reliabably produce the stated crash but that might be because I need larger project files to test with.
I'll look into that.

Other than that (lack of testing on my part) this PR looks straightforward.
It might be a bit of a hack but unless others can suggest a better way forward it appears reasonable to implement.
I've tagged a few ghosts from the past to request review as well as @shun-iwasawa

RodneyBaker:
I am of a mind to merge this 'hack' even though it might not be fully optimal.
My deeper interest is in Qt itself and whether any more recent changes/updates to Qt might address the issue... making this hack unnecessary.
Given that this is a fairly simple change and easy to revert I will plan to merge this unless there are objections.
For those inclined to offer such feedback, please let me know at earliest convenience.

TLDR:  Posting intent to merge.

RodneyBaker:
LGTM

PR #4486, opened 2022-07-26
Fix Starting frame Multi-thread FFMPEG

justburner:
This fix the miss of few starting frames on exported FFMPEG movie formats

The amount lost was related to the number of cores of the CPU (and RNG mostly!), this is caused by race condition with logic that was designed for single threading.

Patch only affects when Multi-thread FFMPEG is enabled, does not affect non-FFMPEG formats.


RodneyBaker:
Interestingly, while the nightlies are not being generated at present the automated appveyor builds are still building correctly (even the PR for the Startup Popup update which failed the appveyor test).

As for this PR... LGTM

PR #4488, opened 2022-07-26
Audio fixes and enhancements

justburner:
Preform audio related fixes and enhancements:

- Fixes stereo mixing by not processing samples in mono when converting between stereo formats.
- Fixes 24-bits playback.
- Added floating-point 32-bits playback.
- Add support for 32-bits (float) WAV.
- Add support for 32-bits (float) AIFF.
- Audio Recorder: Added 192Khz frequency.
- Audio Recorder: Added 24-bits and 32-bits formats.
- OGG and FLAC can be imported now.
- `Info...` in Browser now reports more info about audio, additionally some missing formats were added.

Note: This patch was created before 1.6 release but it seems to work fine on my up-to-date copy of OT.

### DIFF
--- a/toonz/sources/common/tsound/tsop.cpp
+++ b/toonz/sources/common/tsound/tsop.cpp
@@ -394,8 +394,12 @@ dstSample += tmp;
       // assert(dstSample.getValue(0) == dstChannel[0]);
     }
 
+    // Removed tround() since it was destroying 32-bits float data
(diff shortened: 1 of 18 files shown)


RodneyBaker:
Thank you @justburner for the fix (and additional improvements).

@shun-iwasawa  The nightlies don't seem to be generating...

PR #4489, opened 2022-07-28
Modify hiding and showing controls in fx settings

shun-iwasawa:
This PR will modify the fx settings popup as follows:

#### Modification in the visibility tag
Consider there is an fx parameter which only have meanings when the fx is in some state specified with another parameter.
( For instance, the `Gamma` parameter of  `Over Ino` is only used when the `Linear Color Space` checkbox is activated. )
In such case, unused fields under the current state should be hidden in order to keep the fx settings simple.

Currently there are two ways to realize such behavior; using the `modeSensitive` tag attribure and using the `visibleToggle` tag.
This PR will fix the `modeSensitive` tag attribure to work even if the "controller" and the "controlled" fields are not in the same layout.

#### Recycling the unused `isHidden` flag of the fx parameter
When there is an update in some fx, such update sometimes alters an fx version. It will enable the fx to work in different ways, the updated and the convintional ways, according to the version so that the old scene containing the fx will render the same result as before.
Consider that there may be a case that a new parameter will be introduced which is needed only for version 2.
Such parameter  should be hidden from the fx settings when using the same fx version 1.

In order to achieve such behavior, I decided to recycle a flag `TParamVar::m_isHidden` which seems to not be used in anywhere in the source.
Now, setting `m_isHidden` to `true` will hide the parameter from the fx settings.
Setting the flag in `TFx::onFxVersionSet()` will enable parameter to change its visibility by versions.

### DIFF
--- a/toonz/sources/common/tfx/tfx.cpp
+++ b/toonz/sources/common/tfx/tfx.cpp
@@ -1002,7 +1002,10 @@ TFx *TFx::getLinkedFx() const {
 
 //--------------------------------------------------
 
-void TFx::setFxVersion(int v) { m_imp->m_attributes.setFxVersion(v); }
+void TFx::setFxVersion(int v) {
+  m_imp->m_attributes.setFxVersion(v);
+  onFxVersionSet();
+}
 
 //--------------------------------------------------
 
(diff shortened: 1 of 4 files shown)


RodneyBaker:
LGTM

P.S.  One of these days we need to have an online session where we explore the depths of FX.  Much of it is pretty straightforward but a lot is easy to miss!  Many new options have been added (thanks to you Shun!) since the documentation was written so it's easy to miss the new additions.

PR #4491, opened 2022-07-29
Particles Fx Motion Blur

shun-iwasawa:
This PR will add a new option `Motion Blur` to the Particles Fx which shows particles blurred along their movement.
Additionally, this PR will move the `Trail` parameter category from `Birth Params` page to `Animation` page just below the `Trail Opacity` parameter as they are closely related.

particles_no_motionblur
Result with no motion blur
particles_motionblur
Result with motion blur
particles_motionblur_trail
Result with motion blur + trail

### New Parameters
- **Motion Blur** (`Animation` page > `Trail` category) : When turned ON, particles will be rendered with motion blur according to its movement.
- **Gamma**  (`Animation` page > `Trail` category) : Film gamma value used for the motion blur.

### Usage
- Turn on the `Motion Blur` .
- Shutter speed will become (1+`trail`) frames. Without trail, It will be 1 frame (from previous frame to the current frame) . 
- Density of the blur can be specified with the minimum and the maximum values of `Trail Opacity` parameter.
- Note that the particle won't be rendered at the frame of birth when the `Motion Blur` option is activated.
- The `Trail` option of Particle Fx seems to be buggy. It sometimes fail to render proper afterimages. The `Motion Blur` option is affected by the same problem when using with trail.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -821,6 +821,8 @@
   <item>"STD_particlesFx.source_gradation"		"Use Control Image Gradation"		</item>
   <item>"STD_particlesFx.pick_color_for_every_frame"	"Pick Control Image's Color for Every Frame"	</item>
   <item>"STD_particlesFx.perspective_distribution"	"Perspective Distribution"	</item>
+  <item>"STD_particlesFx.motion_blur"	"Motion Blur"	</item>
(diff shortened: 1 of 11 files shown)


RodneyBaker:
Thank you @shun-iwasawa for the additions!
LGTM

PR #4494, opened 2022-08-04
Sampling Size for Gradient Warp Iwa Fx

shun-iwasawa:
This PR will introduce a new parameter `Sampling Size` to the Gradient Warp Iwa Fx, which specifies a distance of the sampling point for computing gradient of the Warper image.

Before this PR, sampling distance had been always 1 pixel regardless of the camera size. Which is to say, the rendered results had been not identical when you set different shrink values in the Preview or Render Settings.

This PR will advance the fx version to 2. Version 1 fxs (made in the previous versions) won't show the `Sampling Size` control in the fx settings and will render the same result as before.


RodneyBaker:
@shun-iwasawa It appears ninja has issues with Linus and Mac...

RodneyBaker:
Thank you @shun-iwasawa for the upgrade to Gradient Warp Iwa Fx.
LGTM

PR #4495, opened 2022-08-09
Custom Panel

shun-iwasawa:
This PR enables to create a custom panel window which can be arranged and customized by users.

Users can:
- Create a new panel template (.UI file) with any button arrangement by using Qt Designer.
- Set commands to the buttons in the template and register as a new custom panel.

The custom panel can be floated or docked in the main window just like other normal panels.

        

#### To create a new custom panel

1. Click `Menu bar > Windows > Custom Panels > Custom Panel Editor...`
    The custom panel editor window will open.
1. Choose a layout template from the `Template` combo box on the top-left.
    - Currently there are two default templates, `drawing toolbox` and `eight buttons`.
    - The `drawing toolbox` template is inspired by the ui of [Rough Animator](https://www.roughanimator.com/).
    - Users can create and add a new layout template. (to be described later)
1. Drag & drop the command from the `Command Tree` (on the right) to the buttons or fields (on the left) to set.
    - Currently the template consists of two types of UI: a button and a "**scroller**" (new widget).
    - Each button can set one command to be executed on clicking.
    - Each scroller can set two commands. A scroller is a rectangle-shaped widget with either horizontal or vertical direction, which will execute either command according to the direction of mouse dragging. It is supposed to be used with setting `Previous / Next Frame` or `Decrease / Increase Brush Size` commands.
1. After settings commands to all buttons, type a panel name in the `Panel Name` field on the bottom-right and press the `Register` button.
1. Your custom panel can be used from a command listed under `Menu bar > Windows > Custom Panels` submenu.
    - The custom panel settings will be saved in `Stuff Folder/profiles/layouts/settings.username/custompanels`

#### To create a new layout template

1. Install Qt Designer. Qt Designer (designer.exe) is bundled in Qt5.15.2 package or can be tiny-downloaded [here](https://build-system.fman.io/qt-designer-download) if you don't need rest of the package.
1. Run Qt Designer and create your own layout.
    - Currently OT will recognize `QToolButton` or `QPushButton` as the button type and `QWidget` (with no layout is set) as the scroller type.
    - Please refer to the [manual](https://doc.qt.io/qt-5/designer-widget-mode.html) for detailed instruction.
1. Save .UI file in `Stuff folder/library/custom panel templates`


RodneyBaker:
Works great @shun-iwasawa 
This will be a great addition.

The things that stand out at the moment is:
1.  Opentoonz appears to need a restart to get the custom panel to appear.  This seems reasonable but it would be nice to be able to more dynamically adjust custom panels without a restart.  
2.  I see no direct way to remove a custom panel as in a delete or remove option.  A user will need to know to go into their profile directory to remove the custom panel manually.  I do note that in some cases this might be a good thing as giving users a delete option could allow them to delete their favorite custom panel inadvertently and...  tears will fall like rain.

I have not built a new template via Qt Designer but will do that soon.

So far so good!

shun-iwasawa:
@RodneyBaker 
Thank you for testing the PR.

>Opentoonz appears to need a restart to get the custom panel to appear. This seems reasonable but it would be nice to be able to more dynamically adjust custom panels without a restart. 

Hmm I couldn't reproduce the problem. The custom panel appears in the `Custom Panels` submenu as soon as I finish editing it.

RodneyBaker:
The issue with menu items disappearing might relate to the build I'm testing.
I'll look into that a little more.

By way of testing the scroll capability of custom panels... I did create a Undo History Scroller.  
That has buttons for moving back and forth through time one step at a time via the buttons on left and right or scrubbing through time via the scroll area.   Obviously I have to work on the styling of the scroll area.  ;)

image

RodneyBaker:
I see no reason to further delay this PR especially given that additional PRs will very likely follow to update icons and such... which is separate requirement that also relates to other toolbards.

Thank you @shun-iwasawa for yet another improvement to Opentoonz customizablilty.
LGTM

RodneyBaker:
@shun-iwasawa Do you have any suggestion for removing the outlines in the scroll area.  I thought this was something I inadvertently added to custom panels but I see that outline in your examples as well.  At this point it looks like that might not yet be implmented.

Several people have been diving into this new feature set and they are curious about this as well.

shun-iwasawa:
@RodneyBaker 

>Do you have any suggestion for removing the outlines in the scroll area.

Sorry but I'm not sure what actually is "the outlines in the scroll area". You meant the cyan border lines around the buttons in the preview area of Custom Panel Editor?

RodneyBaker:
Here is hopefully a better idea of what I refer to:

image

shun-iwasawa:
Ah, it is the scroller widget.
Probably I need to enhance the look, making it as much like side view of a scroll wheel.

openanim:
@shun-iwasawa 

I have play around this very useful feature, thank you. It reminds me of the wacom on-screen controls.  It would be nice if in the future we could also use a shortcut to call up some of our custom panels even faster. Like the wacom controls. 
shortcut custom panel

RodneyBaker:
Just a note to link to @konero's update that resolves the issue with scroll bar borders.  REF:  #4661
It may be ideal to make transparent borders (or semi-transparent) the default.

konero:
@RodneyBaker Hmm, I read the posts here but not sure on what the issue was, you wanted to hide the border?

Just some ideas... if they need to be created by Qt.

Could construct a skeuomorphic scroll wheel from three parts like the one in DaVinci Resolve.

image

Or a flatter version like from RoughAnimator, the rounded edges could be an ellipse cut in half using arc. I haven't done anything like this in donkey's years though, it might be outdated.

```
path->arcMoveTo(etc);
path->arcTo(etc);
```

Then draw 4px lines inbetween with 4px spacing (or what looks good). Could also think about constricting the wheel resizing its width to a certain point but placed inside a hit box that actually handles the user action, or could just let it rip.

image

RodneyBaker:
Yes.  Until your PR the border for the assigned areas given to scroll bar was visible. Making it just look like generic boxes.
Shun then talked of plussing the look up even more to actually look like a dial.

Waiting in the wings is also the radial dials which currently work but do not look in any way like radial dials.

I really like the look of your gradient scroll wheel.  Nice.

PR #4496, opened 2022-08-09
Camera Capture : Show subcamera center on mouse hover

shun-iwasawa:
This PR will improve the `Camera Capture` popup, showing a small cross mark at the center of subcamera when the mouse cursor is inside of the subcamera frame. It will enable users to adjust the subcamera position more precisely. 

penciltest_subcamera_center_cross


RodneyBaker:
Thank you @shun-iwasawa for the update to Camera Capture.
(exciting times with the pending opening of Ghibli Park!)

LGTM

PR #4498, opened 2022-08-10
Expand xsheet frame area

shun-iwasawa:
frame_area

This PR is to resolve a request by some Japanese animation studio as follows:
- Regarding the frame number in the Xsheet row area, when the display mode is in "3sec sheet" or "6sec sheet", it should have more spacing between the page number and the frame number for better visibility.

To achieve this, this PR will expand the frame number region in the Xsheet row area when the onion skin and the shift&trace features are both disabled.

Please note that the appearance of Timeline is kept unchanged.


RodneyBaker:
Thank you @shun-iwasawa 
LGTM

PR #4499, opened 2022-08-10
Fix preview range not to be negative

shun-iwasawa:
This PR will fix a problem that the preview range markers can be set to negative frames (therefore, the markers disappear from the xsheet).

**To reproduce:**
1. Right click frame1 in the xsheet row area.
2. Drag the mouse pointer upward with keeping the right mouse button pressed.
3. Release the button and select `Preview This` from the context menu.
4. The preview range markers will disapper.


RodneyBaker:
LGTM

PR #4500, opened 2022-08-12
Crash Handler

justburner:
Finally OpenToonz can catch crashes and even create a report! Something I've felt that OT was lacking for a loooong time.

CrashReport

Tahoma2D uses CrashRpt to generate reports but this PR adds no dependency into OT, I've added support for all OSes but only fully tested functionality on the Windows platform.

Since there's no specialized libraries the collected info is rather limited, but it's better than nothing, right?

**Current limitations:**
- No report of current loaded modules for Linux/Mac.
- No generation of MiniDump for Linux/Mac (is this Windows only anyway?).
- Probably some more important information to help track bugs that I've forgot. 🤔 
- No send report automatically, there's no server to collect and/or manage that info.

**Current issues:**
- Windows: Artifacts seems to be compiled with `Release` instead of `RelWithDebInfo` so there's no source file and line information.
- Linux: I suspect there's no `-rdynamic`, it means that addr2line can't figure out source file and line info from the app ELF.
- Mac OS X: Completely untested, may not even work. 🤷 
- Resuming work likely always causes an unrecoverable crash.
- Implementation isn't perfect, OpenToonz can still crash without generating a report.

### DIFF
--- a/toonz/sources/common/tcore/texception.cpp
+++ b/toonz/sources/common/tcore/texception.cpp
@@ -3,15 +3,7 @@
 #include "texception.h"
 #include "tconvert.h"
 
-static TString s_lastMsg;
-
-TException::TException(const std::string &msg) {
-  m_msg     = ::to_wstring(msg);
-  s_lastMsg = getMessage();
-}
-
-TString TException::getLastMessage() { return s_lastMsg; }
-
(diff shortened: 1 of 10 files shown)


RodneyBaker:
Thank you @justburner for this initial Crash Handler which is sure to be useful in tracking down and fixing bugs in Opentoonz.  Much appreciated.

LGTM

PR #4506, opened 2022-08-20
Style Selection .isEmpty logic fix

justburner:
Mainly fixes Palette Gizmo not being able to modify color if there's no styles selected.

Now the user can open Palette Gizmo anywhere on the Level Palette and the current style will be affected.

### DIFF
--- a/toonz/sources/toonzqt/styleselection.cpp
+++ b/toonz/sources/toonzqt/styleselection.cpp
@@ -568,7 +568,7 @@ void TStyleSelection::selectNone() {
 //-------------------------------------------------------------------
 
 bool TStyleSelection::isEmpty() const {
-  return m_pageIndex < 0 && m_styleIndicesInPage.empty();
+  return m_pageIndex < 0 || m_styleIndicesInPage.empty();
 }
 
 //-------------------------------------------------------------------


RodneyBaker:
Interesting that the appveyor tests all passed but the Windows test (via github actions) did not.

justburner:
Seems like a permission denied in 'D:/a/opentoonz/opentoonz/thirdparty/glut/3.7.6/include'

RodneyBaker:
Initial testing looks good.
The only thing that seemed odd... but somewhat expected... is that if a new palette page is created and the user is on that new page (with not style on that page)... and the Palette Gizmo is opened... nothing is technically selected so nothing changes.  This would seem to be desireable although I can also see where if nothing is selected on that new (empty) page then an error might inform the user to create/select a color style.

(In case it isn't obvious... I'm testing the appveyor here)

justburner:
Actually the style on the other page remains highlighted, you'll notice that when tabbing on different style pages.

RodneyBaker:
Yes, sorry, I did notice the selection remained on other page but didn't mention it because even though selected it isn't adjusted by the changes made in the Palette Gizmo.  Is that something that should happen?  Initially I thought it was changing that original selection but apparently it is not... as if nothing is selected.

justburner:
During my debugging session to track the problem with Palette Gizmo I've seen it can silently change a style... in another words, it changes but doesn't show up on Level Palette until you move your mouse near it to "refresh"... is strange.

RodneyBaker:
(This is something of an experiment to unearth additional information on what  the github action build lacks)

Thank you @justburner for the Palette Gizmo/Palette fix.

RodneyBaker:
Merging this as the nightlies are built via appveyor.

Thank you @justburner  for the fix.

RodneyBaker:
@shun-iwasawa unfortunately, the nightlies are again not building for Windows.

PR #4507, opened 2022-08-20
Palette Level Toolbar: Menu to hide buttons, Palette Gizmo button, Name Editor button and missing icons

### LINKED ISSUE #4083: Add Palette Gizmo Button back to Level Palette.
I was wondering if you could include the Palette Gizmo Button to the Level Palette next to the Freeze Button?  Because I use it a whole lot in my coloring and also the original Toonz has it there too, many people do not know it exist at current and those who do usually forget because its so well hidden I think it would be a good idea to bring it out of hiding for V1.6 maybe even a way to show or hide it like the Style button has?

justburner:
Resolves #4083, adds a tiny Palette Gizmo button on the bottom of the Level Palette.

In addition it removes Palette Gizmo menu on palette viewers other than Level Palette because Palette Gizmo is limited to current level palette manipulations.

Last but not least, it provides a very tiny visual difference when all styles are deselected in thumbnails view.

UPDATE:
Ok, this PR now do a bit more... it allows to choose which buttons to hide on the toolbar, useful for narrow Level Palette window/dock.

Added Name Editor button ... because... why not?

Some of the missing icons have been added too.


RodneyBaker:
With the exception of the Windows build issue (with github action build) this PR appears to be working as advertised (via appveyor build).  As the nightlies are created via appveyor I'm tempted to merge...

justburner:
Maybe I should add an option to hide some buttons... today I've tried to make a narrow level palette and all those squished buttons bothered me a bit. 🤔

RodneyBaker:
While the (second) Windows build is still being problematic all other build tests have passed and everything checks out.
Will press on with this PR.

Thank you @justburner for this new toolbar, icons and options.
LGTM

PR #4508, opened 2022-08-22
Current frame indicator for Xsheet

shun-iwasawa:
This PR will show the current frame indicator in Xsheet.
In the Xsheet mode, the indicator will be top and bottom border lines at the current row instead of the middle line, in order to keep the cell numbers visible.
current_frame_indicator

Also made the color of indicator editable in the style sheet ( with property name `qproperty-CurrentTimeIndicatorColor` ).
For instance, type the additional style sheet as follows to make the indicator semi-transparent yellow:
````
XsheetViewer {
    qproperty-CurrentTimeIndicatorColor: rgba(255, 255, 0, 0.7);
}
````
current_frame_indicator_yellow


RodneyBaker:
@shun-iwasawa when you get a spare moment can you look into the build environments and see what needs to be refreshed?  Thanks!

RodneyBaker:
Thank you @shun-iwasawa for the addition (and for making it customizable via stylesheet.
LGTM

PR #4511, opened 2022-08-23
Actions: temporary disable clcache

shun-iwasawa:
This PR will temporary disable clcache used in Windows runner of GitHub Actions in order to resolve recent failure.
I've tried to fix the problem but unfortunately could not find a good way so far.

Disabling clcache slows the build but at least it can complete the build.
(Please note that the build triggered by this PR will take long time than usual to complete since I tried removing all caches, some of them are not related to clcache and will be reused in subsequent builds.)


RodneyBaker:
Thank you @shun-iwasawa!

PR #4512, opened 2022-08-24
Fix crash on replacing level's parent directory while opening preview

shun-iwasawa:
This PR fixes following 2 crash scenarios while opening preview flipbook:

#### Scenario 1

1. Load some level to column1.
2. Preview the scene. (the Flipbook opens)
3. Select cells containing column1 and 2.
4. Right click an empty cell at column 2 in the cell selection then select `Replace Level > Replace Parent Directory...` .
5. Select some folder containing a level with the same name. Click `Replace` .
6. OT crashes.

#### Scenario 2

1. Load some level to column1.
2. Preview the scene. (the Flipbook opens)
3. Right click a level and select `Level Settings` . (The level settings popup opens)
4. Select cells containing column1 and 2, by dragging from a cell at column2 to a cell at column1. (i.e. the current column will be column 2)
5. Close the level settings popup.
6. OT crashes.

Both operations will emit `LevelChanged` signal.
`PreviewFxManager` receives the signal and try to updat the preview if needed. `PreviewFxManager` tries to access the current level to obtain the file path of changed level. If there is no current level, OT crashes as it tries to access the null pointer.

### DIFF
--- a/toonz/sources/toonz/levelsettingspopup.cpp
+++ b/toonz/sources/toonz/levelsettingspopup.cpp
@@ -747,7 +747,7 @@ LevelSettingsValues LevelSettingsPopup::getValues(TXshLevelP level) {
       values.doAntialias = (sl->getProperties()->antialiasSoftness() > 0)
                                ? Qt::Checked
                                : Qt::Unchecked;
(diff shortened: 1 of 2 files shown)


RodneyBaker:
LGTM

PR #4516, opened 2022-08-26
Update Grass Particle fx.

JPUser1:
this fixes the start value for the Grass  Particle FX.
seen in this Tahoma2D Bug report here: https://github.com/tahoma2d/tahoma2d/discussions/1030

### DIFF
--- a/stuff/fxs/presets/STD_particlesFx/Grass.fx
+++ b/stuff/fxs/presets/STD_particlesFx/Grass.fx
@@ -1,7 +1,7 @@
 <dvpreset fxId="STD_particlesFx" ver="1.0">
   <params>
     <source_ctrl>
-      0 
+      -1 
     </source_ctrl>
     <multi_source>
       0 0 


manongjohn:
This is not properly crediting the person who created the original PR for T2D that this is based on.

JPUser1:
> This is not properly crediting the person who created the original PR for T2D that this is based on
sorry about that. Lol

RodneyBaker:
@shun-iwasawa As you find some time can you look to see if updating the underlying FX itself would be better to alllow the current defaults to work as expected?  While on the surface changing the default value does fix the problem it seems to me that fixing the underlying code that works with those defaults would be the better approach.

RodneyBaker:
Going back to square one in testing...
I'm not seeing the original problem with the default Grass FX.

image

More testing required...

JPUser1:
> Going back to square one in testing... I'm not seeing the original problem with the default Grass FX.
> 
> image
> 
> More testing required...

I was under the impression there was a problem form my own testing but maybe I was testing it wrong? I still cant figure out how u ar4e getting it to work however I will just close the PR.

PR #4522, opened 2022-09-01
solving linker issue on customized libtiff

mufafa:
If you install libtiff through brew before, linker tool couses fail. In order to solve this problem use install_name_tool.
If libtiff library was installed through brew tool before, linker tool causes fail. In order to solve this problem customized libtool.dylib's id should be changed by install_name_tool

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,3 @@
+{
+    "cmake.sourceDirectory": "${workspaceFolder}/toonz/sources"
+}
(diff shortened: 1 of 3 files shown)


RodneyBaker:
It seems to me checks are failing because this relates to building with brew.
If others building with brew can investigate that will be appreciated.

RodneyBaker:
Pressing forward with this PR as the commit that failed checks consists only of reformattng.
The failures appear to have occurred back when @shun-iwasawa was last updating the build environment.

Thank you @mufafa for the fix for linker with customized libtiff.

PR #4523, opened 2022-09-01
Fix Bokeh Fx White Artifacts

shun-iwasawa:
This PR fixes the following problem occurred in our studio:

- When rendering in 16bit per channel, a white pixels artifact sometimes appears in the result of Bokeh Advanced Fx Iwa, especially at a layer using the Depth reference image.

I found that there is a case the RGB value becomes negative in the fx result. And the fx had no way to handle negative values on converting them to 16bpc channel values, which cannot be negative. In such case the pixels becomes 65535 (16bit max) and turned into white pixels artifacts.
I made an additional condition to make the negative RGB values to be clamped to 0.


RodneyBaker:
@shun-iwasawa it seems Appveyor is having issues again.

RodneyBaker:

PR #4525, opened 2022-09-08
Fix assertion failure on initializing Studio Palette

shun-iwasawa:
This PR removes the assertion macro in  the function `PaletteViewer::applyToolbarPartVisibility()` which was added in #4507 .
It assumes that the toolbar contains every types of buttons, but since the studio palette does not have a button for Palette Gizmo, it always failed on initializing the studio palette panel.

@justburner 
Please check if the removal would be OK. Thanks!

### DIFF
--- a/toonz/sources/toonzqt/paletteviewer.cpp
+++ b/toonz/sources/toonzqt/paletteviewer.cpp
@@ -234,7 +234,6 @@ void PaletteViewer::toggleToolbarOnTop() {
 //-----------------------------------------------------------------------------
 
 void PaletteViewer::applyToolbarPartVisibility(int part, bool visible) {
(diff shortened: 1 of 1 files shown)


justburner:
hmm... that made me realize that even if i do `assert(false)` my VS Windows Debugger won't trigger it because I'm not using  "Debug" solution anymore. ... interesting... but that actually makes complete sense.

This bug doesn't affect the "Release" target so Nightly is unaffected.

> @justburner Please check if the removal would be OK. Thanks!

Approved, that assert does more harm than good.

RodneyBaker:
Based on @justburner's feedback... posting intent to merge.
If there is anyone else with thoughts on the issue please let us know.

RodneyBaker:

PR #4542, opened 2022-09-22
Main Window See Through Mode

justburner:
This PR allow the main window to be semi-transparent so all windows below OpenToonz are partially visible, useful to work with references when the user has limited screen space.

Something like this already exists in Tahoma2D, note this PR is not a port but a re-implementation from scratch so features and behavior do differ.

- The command will toggle mode on and off, The dialog is only visible when the mode is on.
- Command shortcut is Alt+` by default.
- Slider go from 2% to 100% with steps of 2%, you can see the exact percentage from the slider tooltip.
- Opacity value, dialog position and size are saved in settings so OT will remember those.
- When the dialog is focused the arrow keys, page up/down, home and end will control the opacity and enter key closes the dialog.
- There's a button to quickly toggle semi-transparent and full opaque at the right side of the slider.
- Holding ALT while clicking the quick toggle button will toggle semi-transparent and full transparency instead, on Windows this allow to interact with stuff behind (as if the window isn't there) but I'm unsure if that behavior is true for all other OSes.
- When slider is at 100%, the quick toggle will act as ALT is held allowing to quickly toggle full opaque and full transparency with only a single mouse click.

Thanks to SodaPop and Rodney for beta testing.


RodneyBaker:
The only thing odd that I can see is that when the canvas (viewer) is full screen transparency doesn't appear to work.
This would seem to align with what was stated; "allow the main window to be semi-transparent so all windows below OpenToonz are partially visible"... where that full screened window is not below Opentoonz but rather above.

The workaround of a sort that I used to get maximum screen drawing space with transparency was to create a new custom room that only have a viewer in it and use shortcut keys for drawing.

It's possible I am missing something here but that is my observation.

This will be a welcome and useful feature.   :)

openanim:
thanks @justburner for this feature. I can reproduce what you said @RodneyBaker . Im using "toggle maximize panel command" instead full screen. Another thing, It's not a big problem actually, but when you shrink the window very compactly to save as much space as possible on the tablets, the slider is not usable in the best conditions. Maybe the slider should be below the button when it is so small or placed vertically, in the future.


slider_compact

justburner:
> The only thing odd that I can see is that when the canvas (viewer) is full screen transparency doesn't appear to work. This would seem to align with what was stated; "allow the main window to be semi-transparent so all windows below OpenToonz are partially visible"... where that full screened window is not below Opentoonz but rather above.

Full screen seems to work fine with me, or am I missing something?

https://user-images.githubusercontent.com/433868/192022882-073c029f-e600-4a9d-ba1e-4dcc47c702de.mp4

> The workaround of a sort that I used to get maximum screen drawing space with transparency was to create a new custom room that only has a viewer in it and use shortcut keys for drawing.

Panels outside of Main Window won't work, use @openanim suggestion of maximizing the viewer (you can also maximize a docked panel by double clicking the top title area too).

> Another thing, It's not a big problem actually, but when you shrink the window very compactly to save as much space as possible on the tablets, the slider is not usable in the best conditions. Maybe the slider should be below the button when it is so small or placed vertically, in the future.

Thanks for the suggestion, I'll work on making the dialog more compact :)

RodneyBaker:
Try Alt F instead of Alt '... although... I was at times using the same Alt ' as you and getting the full screen canvas rather than Opentoonz itself going full screen.

justburner:
See through window is now more compact by removing the label widget, the user can still verify that the slider is related to opacity from the tooltip.

@openanim Just to let you know what I've implemented your suggestion toward orientation, in order to make the slider vertical the window need to be taller than wider, give it a try. :)

> Try Alt F instead of Alt '... although... I was at times using the same Alt ' as you and getting the full screen canvas rather than Opentoonz itself going full screen. In other words, I was attempting to get maximum screen space for drawing because Opentoonz's windows can interfere with this feature.

I've chosen Alt+\` because it's both unused and it complements Ctrl+\` that is used for a feature related to main window also.
Just swapped to US keyboard mode (since \` shortcuts in PT keyboard doesn't work) and tested all 3 shortcuts, I didn't found any interference toward the shortcuts.

RodneyBaker:
Sorry, I wasn't clear.  None of my observations had anything to do with shortcut keys.
I was just trying to convey that I believe I was using a different 'Full Screen' option that you might have been using.
I will plan to record what I'm seeing but first I want to test on a different computer to rule out this particular computer's resolution etc.  It can cut off parts of screens...

What I was trying to convey is that when I go into a full screen mode where it is just the canvas at full screen... the See Through Mode dialogue box is not visible.  I can't confirm yet but I think it is still there just below that full-screened window.

I will test more and report back.

In other news, I'm not sure why the appveyor tests are now failing.

openanim:
Thanks a lot for the quick setup! I'll be able to try it out as soon as the appveyor is available. About alt+f @RodneyBaker  refers to the full screen of the viewer itself. If you right click on the viewer you can see it or alt+f from the viewer.

justburner:
Oh, I didn't knew that feature, interesting.

As for the Full Screen Mode of the Canvas (as opposed to Main Window's Full Screen Mode... confusing) it detaches the widget from the main window and put it on a top-most borderless "window" over the main window.

Going around that will be quite tricky but anyway I'll see what i can do...

openanim:
@justburner I feel a slight lag in my brush with the PR even without the full screen mode, I think it needs confirmation by others (with or without activating the see through mode) but it is noticeable when you draw.

JPUser1:
I am using a Windows build and I do not have any problems with lag at all. @openanim  what are u on Mac? also what type of Display or Tablet do are you using?

openanim:
STM_brush_issues 0001

In the first screenshot what my drawing lines look like when I draw on this PR.

 In the second one below , in the latest normal NB version.

normal_nb_comparison 0001



I wonder if it's just a matter of resources and hardware power for all the screen manipulations covered by this PR. My tablets are not very powerful, which may explain our different experiences. I think that this could be understandable especially in case the mode is active. It is more problematic if it happens when the mode is inactive. But maybe once this PR is merged it will be better.

Here are my settings:

Cintiq companion 2 -

proc Core i7-4558U - intel iris 5100 HD graphics - 8 go ram.

I use the following Windows 10 settings: resolution (2560 x 1440) at 150% for text size etc. I have also activated for better compatibility the high dpi settings for tablets from the opentoonz shortcut on "enhanced system". (right click on the shortcut => properties => compatibility)

konero:
Just tested, I'm having the same problem as @openanim, comparing the OT nightly I have installed with this PR there's a big difference in brush performance even without enabling the mode. It feels really floaty like there's some kind of brush smoothing enabled, my curves end up with sharp angles to them, especially at the start of a stroke, and there's no pen pressure at all.

I used a basic raster level with the basic brush, I enable pressure with opacity, except pressure isn't working here.

Tested with mouse, Cintiq and Intuos.

Specs are...
CPU: 4.6GHz, 16-Threads, Ryzen 5000 Series
RAM: 32GB 3200Mhz
GPU: Nvidia

JPUser1:
hmmm very bazaar that it works perfectly for some of us and then others have lag?  I guess this PR is not looking good to be merged @justburner  maybe we should just save it for a private fork?

RodneyBaker:
It would be good to validate specifically which build is being tested.

@konero said:

> ...the OT nightly I have installed with this PR...

There is technically no 'nightly build' until the PR is actually merged.
The closest to that is the appveyor build.
If testing with the github action builds there are known issues (and no doubt unknown) related to brushes that might account for features not working as designed and (presumably) the potential for such things as lags.

Because the builds themselves (nightlies that are built via appveyor) have not been building correctly this makes me question whether or not we are getting valide results from testing.

cc: @shun-iwasawa (regarding issues with appveyor/nightly building

RodneyBaker:
For those Windows users who know how to set the stuff directory up to make the download into a portable... here is the best build to test a present from what I can tell:

https://ci.appveyor.com/api/buildjobs/uybpi44ydmcyqbns/artifacts/toonz%2Fx64%2FOpenToonz.zip

Mac and Linux builds should theorectically be good to test as well.

With a quick test of this build I detect no lag (none introduced by this PR beyond the standard lag present since release of v1.6 that is).  There are a few brush oddities I have noted prior to this PR but I haven't been able to track them down as of yet.

DarrenTAnims:
This look really nice. Thanks @justburner.
One thing though, the tooltip comments on the little button say that pressing Alt when you click uses full transparency. But for me (Windows computer, using the build linked by Rodney), pressing Alt and clicking makes OpenToonz disappear and instead of drawing in the program, I can click on the program beneath and operate that. The next line says that when the slider is at 100%, it acts like the Alt is held. It doesn't. I can still draw in OpenToonz. So these 2 options are very different. The text should reflect that.

RodneyBaker:
Posting intent to merge this PR with the following goals in mind:

- Determine if the nightly actually builds 
- Determine if a lag can be verified in that nightly build
- Observe/Identify shortfalls in automated build processes with goal to fix/improve those that fall short
- Proceed with PRs in timely manner*

*While the goal is always to merge only code that works well across the board github has a host of tools designed to streamline this process to include automated checks.  In the future it may be ideal to add some form of user/usability checks that allow interested parties to run more customized checks and to faciliate specific testing.  

Bottom line:  If there is an issue with this PR then it is important to validate it and see it be addressed.
In a case where an issue cannot be resolved the PR can easily be reverted.

Feedback always appreciated.

konero:
> Posting intent to merge this PR with the following goals in mind:
> 
> * Determine if the nightly actually builds

I'd rather you didn't, the brush lag is not a good trade off.

What is the nightly build supposed to resolve exactly?

I will check an AppVeyor and Git build with another PR later to see if it also has brush lag, if @openanim has time to do the same we can know if it's caused by this PR or by whatever circumstance the builds are produced from, probably.

RodneyBaker:
> What is the nightly build supposed to resolve exactly?

Most of the following will be stating the obvious but I'll add it here for those interested.


The nightly build is a snapshot of the interim build of Opentoonz after each PR is merged.
As such there is a new 'nightly' for each PR merged.
Subsequent PRs then build based on what is in that interim (nightly) build.

I understand the reluctance to merge but what we end up with is a situation where most consider the nightlies unstable... and yet they tend to be more stable than the previous releases due to the accumulated bug fixes the incorporate.
These interim builds do potentially introduce new bugs but since it is a year between annual releases testing to resolve such relies heavily upon users who work with the nightlies reporting their experience.

To alleviate concern for potentially merging code before it reaches some level of thorough vetting the user can always move back one or more nightlies until they no longer experience a problem.  In Studios, this will often take on the form of sticking with one specific release unless a specific bug fix or new feature is required in which that portion of the production will use the (unstable) release pending the next (stable) version.

Ultimately, all the various releases are designed to allow users maximum flexibilty in risk assessment vs use of new developments which might not have covered all the bases; i.e. the law of unintended consequences.

In short, the nightlies provide a means to move a PR forward with reasonable assurance of operating as intended.
With the limited number of developers and testers available the nightlies provide a waypoint on the way to the next release that should maximize stability and functionality to the greatest extent possible.

The nightlies therefore serve a criitical purpose in getting new code into the hands of the maximum number of users that can validate the code as it moves toward the next (versioned) release.

Thanks for commitment to testing as that feedback is key to this process.
Also for voicing your concerns which are equally important.

Aside:  It will be good to lock down standardized and effective means of timely feedback and reliable testing in the future but it remains to be seen what form that will take on.  If a PR is solid (?) and concensus (?) is that a feature should be added we should not unnecessarily delay its merger.

openanim:
As suggested by @konero , I also tested with another PR. (histogram pop up) to compare. @RodneyBaker It seems indeed that in other PR the problem exists in my environment. 

This is visible when testing on the vector level (with smooth at 0) as we can see how the lines look before they are rendered, we can see that the curves on the both PR are going to be broken. Problem that I could'nt see on the normal NB build with the 3 levels.

Note that the STM mode is disabled in all my tests below since the problem is strongly accentuated when it is enabled
in the way described by @konero in his previous messages. I also think it's important to preserve a good drawing experience no matter what if we test the merger of this PR.

comparason PR

Some screenshots :


Histogram popup PR

standard raster level :

histogram popup_PR

Vector level :

histogram popup_PR_vector


STM PR :

standard raster level :

See t_mode_PR

Vector level :

See t_mode_PR_vector

Normal NB build :

standard raster level :

last_NB build

Vector level :

last_NB build_vector

RodneyBaker:
@openanim Your results suggest to me that the nightly build (for this Main Window See Through Mode) would likely be good while the PR build is not.

A question that still has not been answered by you or by @konero:  When testing the PR is it the appveyor build that is being tested?  This is the one that will require renaming of the stuff direcotory and placement inside the Opentoonz directory.
This is very important to get a valid test to compare against the nightly builds. as the Windows nightly.

That will be this build:  https://ci.appveyor.com/api/buildjobs/uybpi44ydmcyqbns/artifacts/toonz%2Fx64%2FOpenToonz.zip
Again, this requires  renaming of the stuff directory to 'portablestuff' and placement inside the Opentoonz directory at is also inside the zip file.

I am not suggesting the other builds are not useful but there are known problems (especially with mypaint brushes) that make testing with them impossible. 

Note that I assume the results you are getting from this and other PRs is largely due to using the github action builds.
This appears to be confirmed by the nightly build results.  As such as good test of say... the Historgram popup PR would be using the appveyor build.  Link to that build:  https://ci.appveyor.com/api/buildjobs/4kvn1rd342pqjd8y/artifacts/toonz%2Fx64%2FOpenToonz.zip

Can you please verify the builds you are testing with are the appveyor builds when you say you are testing this and other PRs?

openanim:
@RodneyBaker Sorry, in my previous tests I did not build in a standalone portable as you seem to suggest. I redid the test in standalone portable for each PR using your links. Although I get better results, unfortunately it seems that there is still this lag compared to the normal build on my side. I always get broken curves. And also in both PR building in same way. On STM PR it's worse when STM is enabled as before.

Note what i did if it's valid: I unzipped the zip and renamed "Opentoonz stuff" to "portablestuff" and put it in the Opentoonz folder before launching Opentoonz.exe

STM PR (portable)

STM_PR_Portable

Histogram PR (portable)
histogram_PR_Portable

RodneyBaker:
Thanks @openanim, it certainly reads like you used the correct (tm) build.
I do find it interesting that you also get the same (or very similar) results with the Histogram PR which doesn't have any code from this (Main Window See Through Mode) PR.

There is something afoot here that does not seem to be strictly speaking related to the code from the PRs but rather due to the build environment(s) but it may take an eye such as from @shun-iwasawa to get to the bottom of it.  This may be related to the trouble with building nightlies but that doesn't seem to be the case either.  The building and propogation to the nightlies listing appears to be a different although troublesome problem.

Given that the Histogram PR is unlikely to contain bad code... @shun-iwasawa is quite knowledgeable in this regard... I think a way forward might be to merge that PR and use it as a means to further identify the problem(s).  @shun-iwasawa is also very quick to correct any deficiencies in code so that is a reasonable risk to take in my estimation.

While an unfortunate delay for this particular PR it  is certainly a reasonable precaution to take to delay while we investigate further.  @justburner is known to be patient and of the mind to prefer delay over inviting errors to creep into the master branch and I don't see this as being any different.  Developer silence isn't always reassuring and I can only hope such delays don't discourage additional development.  Delays in reviewing and merger in the past led to external forks by Jeremy and John; which some see as a good thing while others do not.  (I personally like forks of Opentoonz as it allows for maximum experimentation).

At any rate, we'll press forward as best we can with expectation that the rough spots will be smoothed out.
It may well be that the custom Qt used for Opentoonz is not the same as the Qt used in creation of these PRs and something like that might account for the difference.

RodneyBaker:
The histogram popup PR is now merged and available as a nightly (except for the Mac build which hopefully will still propogate to the list).  As such the Windows nightly build should theoretically be good for testing.

RodneyBaker:
@konero I believe we are waiting on your results of testing in order to decide how to best proceed.
Based on @openanim's feedback merger appears to be a more than acceptable risk for this PR.
Should problems be encountered we will ask @justburner to assist.

My primary concern at this point is whether the nightly will actually build and propogate to the nightlies listing as that still seems to be having issues.

konero:
I tested another PR, it felt about the same as this one.

RodneyBaker:
The last nightly built (for WIndows) after about 3 days.
Think I'll press my luck and see if this one does the same.

For those interested in testing I'll plan to provide announcement and link to the nightly when it successfully builds.

Thank you @justburner for the new Window Transparency feature!

RodneyBaker:
The nightly didn't take long to build (although the Mac buildis absent in the listing).
Here's the link to the WIndows build:   https://github.com/opentoonz/opentoonz_nightlies/releases/download/nightly_877e17c02959e015a9648f92fee1f9821c626623/OpenToonzSetup.exe

Note that I didn't immediately see the new menu item appear for this build but after a clean install it appeared.

In initial testing of the build I am not experiencing any lag.
Hopefully that will be other users experience as well.

PR #4545, opened 2022-09-24
Thicken Xsheet marker lines

shun-iwasawa:
image

This PR will thicken marker lines in Xsheet when the preference option `Xsheet > Highlight Line Every Second` is ON.
Also enabled to apply different colors to the marker lines at every second and other marker lines.

This PR will also enable to apply different colors to;
- Text color for the current frame number in the xsheet row area (separated from other frame numbers)
- Text color in the selected cells (separated from non-selected cells)
- Marker line color in the selected cells (separated from non-selected cells)

These properties will not affect the existing themes since they have the same colors as before.
They will be useful when making a custom theme in which the current frame indicator and selected cells are with "inverted" background colors, as seen in an Xsheet UI of Celsys QuickChecker (see the image below).

image

These modifications were requested by some Japanese animation studio.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -2282,9 +2282,13 @@ Ruler {
 XsheetViewer {
   qproperty-TextColor: #d6d8dd;
   qproperty-ErrorTextColor: #ff7b7b;
+  qproperty-SelectedTextColor: #d6d8dd;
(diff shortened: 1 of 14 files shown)


RodneyBaker:
LGTM @shun-iwasawa 

Failed check due to appveyor apparently needing another refresh.

RodneyBaker:
That you @shun-iwasawa For the update.
All now appears to be working properly.

PR #4548, opened 2022-09-26
Fix Fx obsolete parameter handling

shun-iwasawa:
This PR fixes obsolete parameter handling when loading older version of fxs.
I missed to take over the obsolete status in some functions which made this mechanism to fail to work properly.


RodneyBaker:

PR #4549, opened 2022-09-26
Fix Custom Panel Registration

shun-iwasawa:

- After registering / updating the custom panel by the Custom Panel Editor, the `Custom Panel` submenu in the main menu bar fails to update with new items.


RodneyBaker:
The Mac test appears to be stalled.
Will merge this to see if there is any issue with the final build itself.

Thank you @shun-iwasawa for the fix to Custom Panel registration.
It looks like you fixed the problem I was trying to report but couldn't explain clearly.

PR #4551, opened 2022-09-26
Export Xsheet to PDF: Frame Length Field

shun-iwasawa:
image

This PR will add a `Frame Length` field to the Export Xsheet PDF popup which enables users to specify frame length to be exported. It will be useful especially when users would like to export xsheet after placing key drawings, since inbetween drawings can be placed below the last key drawing and in such case they need to set longer duration than the current scene length.


RodneyBaker:
Still anticipating the nightlies might be building correctly again.

As for this PR... LGTM
Thank you @shun-iwasawa for the addition.

PR #4557, opened 2022-10-06
Histogram popup enhancements

shun-iwasawa:
This PR will enhance the Histogram popup as follows: 
 - Now the histogram popup will open next to the viewer in order not to hover over the image.
 - Clicking the Histogram button or using the `Histogram` command  via shortcut key while the popup is visible will now hide the popup, instead of raise the popup on top. (*)
 - Alpha channel value will be shown in the `Picked Color` and the `Average Color` labels when the Alpha histogram is visible.
 - The `Histogram` command shortcut key now can open the Flipbook histogram as well as the Viewer histogram. (Which to open is depending on which console is active.)
 - Fixed picking average color (Ctrl + Drag) may shows wrong value when displaying 16bpc image (due to overflow).

(*)  **PLEASE NOTE** : I applied this behavior to ALL popup-opening commands such as `Scene Settings`, `Preferences`, `Configure Shortcuts`, etc.. Now the commands will hide the popups if the popups are already shown.

### DIFF
--- a/toonz/sources/include/toonzqt/combohistogram.h
+++ b/toonz/sources/include/toonzqt/combohistogram.h
@@ -51,6 +51,7 @@ private:
   QColor m_color;
 
   DisplayMode m_mode;
+  bool m_alphaVisible;
 
 public:
   ComboHistoRGBLabel(QColor color, QWidget *parent);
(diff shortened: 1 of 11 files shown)


RodneyBaker:
Please refer to #4542 for additional information on specifically why this PR is being merged before others.

Thank you @shun-iwasawa for the Histogram popup enhancement.
The histogram is becoming more and more useful which is much appreciated.
LGTM

PR #4568, opened 2022-10-18
Spin Blur Fx Ino : Spin gadget alternated coloring

shun-iwasawa:
This PR will modify the "spin field" gadget of the Spin Blur Fx Ino with alternated coloring for better viewability.

image

This change had been requested by a digital compositing chief in our studio.


RodneyBaker:
Pressing forward with hopes that nightly builds will start working again.

PR #4569, opened 2022-10-19
Fix Custom File Path Rule Bugs

shun-iwasawa:
This PR fixes the following problems occur when the `Project Settings > File Path Rules` are set to `Custom` :

1. `Output Settings > File Settings > Name` (i.e. output file name) becomes `.` (dot) when no file name is manually specified. It should automatically be synchronized with the current scene name.
2. When batch rendered using `tcomposer`, the File Path Rules of the `sandbox` project are always used regardless of the project which the target scene is belonging to.

### DIFF
--- a/toonz/sources/common/tsystem/tfilepath.cpp
+++ b/toonz/sources/common/tsystem/tfilepath.cpp
@@ -1,7 +1,7 @@
 
 
 #ifdef _WIN32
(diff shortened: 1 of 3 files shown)


RodneyBaker:

P.S. Nightlies still do not appear to be building correctly and propagating to the nightlies listing.

PR #4570, opened 2022-10-20
Convert TZP to TLV in folder

shun-iwasawa:
Sorry but this PR is quite tailor-made for Studio Ghibli, who had been used Toonz in animation production for over 20 years..
I post this for the sake of future maintenance and hoping this may be helpful to someone in the world :-)

This PR introduces a new command `Convert TZP to TLV in folder` , which converts all TZP (old Toonz level) files in all subfolders under specified path to TLV files for archiving purpose.

Furthermore, I (hopefully) managed to enable to read TZP created by using Toonz 4.1 in addition to the newer versions.
(In Studio Ghibli, Toonz 4.1 had been used for production of  _My Neighbors the Yamadas_ (1999) .)
  

### USAGE
image

**Folder to convert:** Specifies a path to the target folder.
**Skip Existing Files:** Specifies whether to overwrite or skip if TZP files already exist.
**Apply to Subfolder:** When ON, conversion will be recursively done in all subfolders.

- Converted TLV and TPL files will be saved in the same folder as the source TZP with the same level name.
- After conversion, report text will be displayed in the popup. The text can be saved as a log file.
- This command is not registered in the menu bar by default as it is quite specific feature. Please register the command to the menu bar manually before use.

### DIFF
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -228,7 +228,7 @@ void setPathsPermissions(const TFilePathSet &pathSet,
     f.setPermissions(permissions);
   }
 }
-}
+}  // namespace
 
 // gestire exception
 void TSystem::mkDir(const TFilePath &path) {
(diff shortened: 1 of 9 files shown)


RodneyBaker:
NIce!   I recall running into an old TZP file someone had shared and had quite a time getting it converted.
This will resolve problems like that.

RodneyBaker:
As I do not currently have TZP files to test with and must assume @shun-iwasawa can address any shortfalls introduced in the TZP to TLV conversion process... will proceed with merging.

Thanks @shun-iwasawa!

PR #4572, opened 2022-10-26
Misc. updates to style editor brush pages

justburner:
This PR provide several changes to the brush selection pages in the style editor:

- Selected brush box now match the current selected style.
Original implementation was extremely naive, now the box is always in sync with current style's brush instead of only being affected by the last mouse click.

- Box colors are now customizable in themes and styles.
It can be fully customized in Addition Style Sheet too:
```CSS
#StyleChooserPage {
  qproperty-CommonChipBoxColor: #000000;  /* Box color of all common chips */
  qproperty-PinnedChipBoxColor: #a0a0a0;  /* Box color of pinned chips */
  qproperty-SolidChipBoxColor: #e41000;  /* Box color of Plain Color */
  qproperty-SelectedChipBoxColor: #ffffff;  /* Outer box color of selected brush */
  qproperty-SelectedChipBox2Color: #c7ca32; /* Inner box color of selected brush */
}
```

- User can pin brushes to top.
Allows to access certain brushes easier and quicker, can be used to "favorite" the most used brushes.
To pin a brush, right-click the brush and check `Pin-to-top`, note that this action also changes the current style's brush!

- Search brushes that contains certain text.
Inspired by Krita! For example in Raster page typing `water` in will list all water color based mypaint brushes, alternatively typing `aotz` will list all mypaint brushes in the aotz directory.
The text it searches is the same as the ones shown in tooltip.
This feature is off by default, to enable it click the hamburger menu and check `Search`.

- Fixed some textures inside the library.
OT doesn't support indexed images so I've converted them to true-color except the weird sphere ones, those remains non-working.

- Refactored some code...
Several changes to make the code more reusable, I've tried my best not to bloat `styleeditor.cpp` so moved all managing implementation to `stylemanager.cpp`.
All chips icons are now cached into QImage, allowing Qt to directly draw them, this slightly increases performance when redrawing the pages.


JPUser1:
Nice job @justburner  now if @shun-iwasawa could please fix the Nightly builds to work again we can test it out?

RodneyBaker:
Looking very good.

A trivial question...
Do you know if there is a way to increase the border thickness also via the style sheets?
I haven't tried but I observe the custom colors can be hard to see (perhaps especially on white or opposite color to what the border colors are set to be).

As for observations on the implementation here:
I thought I would miss the ability to separate and compartmentalize brushes ala the implementation of brush management seen in Tahoma2D but I am pleasantly surprised to see that isn't the case.  This doesn't prevent us from doing that compartmentalizing and management of course via the directory structure itself.

That search capability... very obviously that will be useful!

Nice work.

shun-iwasawa:
@JPUser1 @RodneyBaker 
It seems that the Windows version of nightly build is created & uploaded but only the label (showing the release date) fails to be changed. 
You can try the latest feature like `Main Window See Through Mode` in the current nightly.

Regarding the mac os version, it seems to stop updating for weeks. (Its another problem.)

I will check and try to fix them this weekend. Sorry for the trouble!

Anyway, thank you @justburner for the nice work! :-)

RodneyBaker:
@shun-iwasawa Thank you for the information.
Can you check the latest nightly to see if Custom Panels is working for you?

When I attempt to click on the Custom Panels menu, first I see no options other than the main Custom Panel Editor but then when clicking on that nothing opens (although it 'feels' like something might be opening in the background behind everything else).

justburner:
> Do you know if there is a way to increase the border thickness also via the style sheets?

No, I could allow that to be configurable but I believe the boxes line thickness are best to be as 1 pixel considering the size of the smallest chips.

> I haven't tried but I observe the custom colors can be hard to see (perhaps especially on white or opposite color to what the border colors are set to be).

The pinned boxes colors are supposed to be very subtle/non-distracting but they can be customized to the user's taste.

> I thought I would miss the ability to separate and compartmentalize brushes ala the implementation of brush management seen in Tahoma2D but I am pleasantly surprised to see that isn't the case. This doesn't prevent us from doing that compartmentalizing and management of course via the directory structure itself.

The goal of this PR was avoiding drastic UI changes to prevent objections from any studio.
Brush managing inside the program is probably best to be a separated PR.

(Also nice sketch!)

> Anyway, thank you @justburner for the nice work! :-)

Appreciated :D

openanim:
Thank you for these improvements ! 

On the same topic the brush space is not always 100% filled when you change the size of the style editor window. You can see the problem on the drawing posted above by @rodneybaker.

RodneyBaker:
Thank you for these improvements ! 

> On the same topic (but not related to this PR) the brush space is not always 100% filled when you change the size of the style editor window. You can see the problem on the drawing posted above by @rodneybaker.

Sure can.  That has been present since 2016.

RodneyBaker:
I want to give testers time to test and provide feedback so... if you are testing this PR and feedback is pending please let us know.  Otherwise...

...posting intent to merge where the testing pool can expand to all users through use of nightly download.

RodneyBaker:
Pressing forward with the PR.
Thank you @justburner for the Update to the Style Editor!

openanim:
I like the news textures brushes and especially the chalk fill brush in the "generated" category! It would be nice if we had more news attributes (or the ones available in others brushes)  to animate theses brushes in their settings. Is it hard to create one yourself ?

RodneyBaker:
We can 'animate' certain brushes but only in the sense of resetting values.
In this way we could use one brush with specific settings on frame 1... move the exposure to frame 2 where the brush has other settings etc etc.   A bit too esoteric for this discussion though.   Play with Palette style keying and that will give you the idea.  That's just fodder for a ton of new feature requests though.  ;)

> Is it hard to create one yourself

Brsush creation (of all types) is something we as a community should document fully (with best practices, workarounds etc).
Most brushes** are (relatively) easy to create although I've not delved into the generated brushes yet short of just reviewing the basics.

**I mainly refer to MyPaint and custom vector brushes here.

RodneyBaker:
Something I'd like to look into is having additional brushes in Opentoonz but not (initially) active.
This approach would likely apply to other resources as well such as plugins, shaders, macros, LUTs, textures*,etc.
For example, the brushes might be contained in one (or more) zip files and then could be extracted for use to 'activate' them.
The core idea being that when Opentoonz sees the files they become active.

A few areas of interest/concern:
- Cross platform compatiblity
- Useful means for user to 'activate' the resource (how can the user best gain activation of a resource with very few clicks (optimally one click)
- This might be a good short term solution so how might we envision an even better solution fo the long term 

*While production ready resources is ideal a goal is to provide an adequate set of example files to demonstrate how the process can be extended to include any number of additional resources with minimal impact to productivity.

PR #4577, opened 2022-10-27
Fix Custom Panel Editor hiding

shun-iwasawa:
This PR fixes the hiding popup problem reported in [this comment](https://github.com/opentoonz/opentoonz/pull/4572#issuecomment-1292913504).

Thank you @RodneyBaker for the report!


RodneyBaker:
That addrssses the problem.

PR #4587, opened 2022-11-02
Animation tool option: select camera after switching it

### LINKED ISSUE #3228: Difficulty Selecting Camera with Animate Tool
### Issue Summary
It's difficult to select the camera from the drop down box when I'm using the animate tool, I had two cameras, trying to select one it instead changes to 'Col 1', I have to select it twice to make it selected.

selectCam

### Steps to Reproduce
1. Make a new level
2. Add a new camera
3. Change to Animate Tool
4. Select the table from the drop down
5. Now try to select camera 2... it changes to col 1 instead

### System Information
* **OpenToonz Version**: 1.4
* **Operating System**: Win
* **CPU**: i7-6700
* **Memory**: 16GB
* **Graphics Card**: Intel HD 530 iGPU

flurick:
Should fix #3228

### DIFF
--- a/toonz/sources/tnztools/tooloptions.cpp
+++ b/toonz/sources/tnztools/tooloptions.cpp
@@ -980,12 +980,12 @@ void ArrowToolOptionsBox::onCurrentStageObjectComboActivated(int index) {
     return;
   }
   // switch the current object
-  m_objHandle->setObjectId(id);
   if (id.isCamera()) {
     TXsheet *xsh = m_xshHandle->getXsheet();
     if (xsh->getCameraColumnIndex() != id.getIndex())
       m_xshHandle->changeXsheetCamera(id.getIndex());
   }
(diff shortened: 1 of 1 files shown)


flurick:
Whoops, this is so tricky to test that this patch might not actually do anything, it just looked like it worked hehe

RodneyBaker:
Thanks @flurick for looking into this.
The issue with the dropdown appears to be fixed but the underlying issue may be that the newly selected camera then does not become the active camera.

KaseyFarron:
It fixes the problem, thanks!

> but the underlying issue may be that the newly selected camera then does not become the active camera

I don't know what this means.

RodneyBaker:
Thanks @KaseyFarron for the confirmation.

I don't know what this means.

Not a problem.  That will be fodder for a future PR.  :)

Thank you @flurick for the update.
LGTM

PR #4588, opened 2022-11-02
Keep selection valid when cutting frames

### LINKED ISSUE #4392: Level Strip causes OpenToonz to crash when cutting nothing
Since the Delete button removes just the content of the frame and not itself, I have to cut it (this isn't intuitive at all). That being said, It's normal that I press Ctrl+X more than I intend. So in some of the cases in which I "delete" them all, I press Ctrl+X once more (not on purpose) while the level strip is empty and then the program crashes.

Reproducing the bug:
1. Select a random Level from the XSheet you have dtawn
2. Select all frames from the Level Strip and press Ctrl+X twice 

Expected behavior:
It shouldn't crash.

Device Information:
 - Windows 10
 - OpenToonz Version: 1.6
 - GPU: AMD Radeon(TM) Graphycs
 - GPU Driver Version: 26.20.14040.0

flurick:
Fixes #4392

### DIFF
--- a/toonz/sources/toonz/filmstripselection.cpp
+++ b/toonz/sources/toonz/filmstripselection.cpp
@@ -237,7 +237,10 @@ void TFilmstripSelection::cutFrames() {
   TXshSimpleLevel *sl = TApp::instance()->getCurrentLevel()->getSimpleLevel();
   if (sl) {
     int firstSelectedIndex = sl->fid2index(*m_selectedFrames.begin());
-    assert(firstSelectedIndex >= 0);
+    if(firstSelectedIndex < 0 || firstSelectedIndex > sl->getFrameCount()) {
+      selectNone();
+      return;
+    }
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@flurick  Initial testing confirms this PR works as expected.

I am curious about some edge cases related to this that would be good to validate.
Some of those cases would likely be resolved if we could assign/reassign multiple shortcut keys to the same process.
Then users expecting to use Control X to Cut (that is to say remove) a frame from the Level Strip could do that but then also get the same result by hitting Delete key (assuming it was also assigned to Delete).

For those testing we'd love to have your feedback.

Back to testing...

KaseyFarron:
Fixes the problem, thanks!

RodneyBaker:
Thank you @KaseyFarron for the confirmation.
Thank you @flurick for the fix.
LGTM

Issue #4556, opened 2022-10-03
Spacebar shortcuts, accidentally leaving "temporary tool" state


JBSedich:
### Description

New to this software, trying to use the spacebar navigation that is fairly universal across the stuff I've used (Clip Studio, Sai, etc)

It seems if you press and hold ctrl+space _in that order_, you can rotate the canvas as intended. If you press ctrl+space but you press space first, 1) it changes to hand tool 2) it switches the tool permanently not temporarily.

Similarly,
If you press and hold shift+space _in that order_, you can zoom as intended. If you press shift+space but you press space first, 1) it changes to hand tool 2) it switches the tool permanently not temporarily.

I am doing both of these accidentally, quite often

Just to be clear, my issue isn't that in this software all the spacebar navigation shortcuts are tool switches if you don't hold the input. That did take a bit of getting used to but I reduced the tool switch time delay in settings and it's okay. But the input order mattering is not something I can seem to adapt to.

### Expected Behavior

Would expect:
Order of key input not to matter
Secondary input not to break out of "temporary tool" state.

The spacebar keys are the most problematic for my workflow but it seems to be a general pattern of a secondary input exiting the "temporary tool" state. So if I am holding B to enter brush and draw a stroke temporarily but press N to change to actual pixels (or just hit another key by accident) my stroke is fine but I have switched tools.

### OpenToonz Version

1.6

### OpenToonz Version Information

Apr 4 2022

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Wacom

RodneyBaker:
Sounds similar to some previous issues (particularly with spacebar).  REF:  #1235

RodneyBaker:
Recommend reviewing this report in light of #6199 being merged (soon).

RodneyBaker:
@JBSedich This should be resolved in current nightly releases.
Please let us know if that is not the case.

PR #4590, opened 2022-11-05
Fill Tool - Picker+FH

### LINKED ISSUE #4404: Auto Pick Style when using Fill Selection Types

After using the paint bucket tool, usually there are small unfilled areas left over and if there are many different styles we need to select the style picker tool, then the rectangular fill tool and fill them. This is a lot of clicks!

I would like to see an option for the fill tool (rectangular, for ex.) that automatically picks the style under the cursor, so that when you drag over the area you want to fill, it fills with that style, then after releasing the mouse button it goes back to the style you had selected before.

- When you first click with the Freehand/Rectangular fill types it remembers what style you currently have selected and then selects the style under the cursor position **_temporarily_**
- You drag where you want to fill and release
- It switches back to the style that was selected before this command happened

I made a [AutoHotkey script](https://github.com/KaseyFarron/AutoPickColourAndFill) for this if you want to actually try what it feels like, the current version I have on GitHub will only work with a mouse.

### Example:
113505841-33d1e400-9539-11eb-9cc5-f84753b405cf

justburner:
Resolves #4404 <- Please check the link to understand how this tool works.

This new tool can be located in the Fill tool at "Type:" under the name of "Pick+Freehand"
By default it only sample the `Area`, by holding CTRL on first click it will sample both `Line+Area`.


KaseyFarron:
Wow thank you; it works just like my script! 😮 

I have a little feedback if it's OK.

When setting a hotkey to `Fill Tool - Picker+FH`, it works fine pressing the hotkey but if holding it to activate it temporarily then it makes it unable to make a selection, the cursor also flickers slightly and there's a bit of CPU usage spike.

When adding the `Type` command to the command bar, clicking it repeatedly cycles between `Normal` and `Pick+Freehand`, I'm not sure if that is intentional or not.

I think the names of the commands could be a bit clearer, for example:
|Current|After|
|-----|-----|
|Pick+Freehand|Freehand (auto pick style)|
|Fill Tool - Picker+FH|Fill Tool - Freehand (auto pick style)|
|Type - Picker+FH|Type - Freehand (auto pick style)|

Thank you again! 💞

justburner:
> When setting a hotkey to `Fill Tool - Picker+FH`, it works fine pressing the hotkey but if holding it to activate it temporarily then it makes it unable to make a selection, the cursor also flickers slightly and there's a bit of CPU usage spike.
> 
> When adding the `Type` command to the command bar, clicking it repeatedly cycles between `Normal` and `Pick+Freehand`, I'm not sure if that is intentional or not.

Please test all those with "Freehand" type since that's what the tool is based of.

It's been known that temporary on OT has some flaws: #4556

> I think the names of the commands could be a bit clearer, for example:
>
> Current | After
> -- | --
> Pick+Freehand | Freehand (auto pick style)
> Fill Tool - Picker+FH | Fill Tool - Freehand (auto pick style)
> Type - Picker+FH | Type - Freehand (auto pick style)

`Pick+Freehand` was a compromise of keeping the text small while still being clear enough, the longer the name the bigger the combobox will be and will push all widgets of the toolbar to the right, requiring more screen size to the toolbar overall.

`Picker+FH` was the initial name which i've forgot to replace to the final name... ops.

KaseyFarron:
> Please test all those with "Freehand" type since that's what the tool is based of.

Ah yes, seems all freehand tools are affected, shame... was not aware.

openanim:
@justburner @KaseyFarron  Thanks for this feature.

I tested it quickly on the vector level. I can't get it to work like on toonz Raster level. It seems that depending on the boundary I have a different result. My vector level is fillable normally.

pick+FH

konero:
Works nicely. Is there a reason to have this available for vector levels? I can't imagine the scenario.

justburner:
> I tested it quickly on the vector level. I can't get it to work like on toonz Raster level. It seems that depending on the boundary I have a different result. My vector level is fillable normally.

Doubt they are fillable normally... vector filling issues has been a very old problem that still persists on OT.
I've noticed that closed shapes are significantly more problematic than open ones... anyway all of this is unrelated to the PR.

> Works nicely. Is there a reason to have this available for vector levels? I can't imagine the scenario.

Not much really, since this tool at core behaves the same as "Freehand" so vector levels support came out for free.

RodneyBaker:
I hadn't even thought to test with vector levels.
Very happy to see it works there as well.

It's outside the scope of this PR but an option to 'Ungroup All' on a specific frame or selection of frames would likely resolve some of the issues observed here and elsewhere.  Grouping is something of a beast however as there is a lot of other features (and potential features) that connect to that.

I can see some uses for vector fill with pick+freehand but at present they are fairly fringe processes due to the nature of fill/grouping.

> By default it only sample the Area, by holding CTRL on first click it will sample both Line+Area.

Very nice.  Thanks for adding this option!

RodneyBaker:
Posting intent to merge (later tonight).

PR #4592, opened 2022-11-05
Fix selecting previous xsheet frames/layers with the arrow keys

flurick:
Previous behavior: Peek 2022-11-05 22-48

### DIFF
--- a/toonz/sources/toonz/xsheetviewer.cpp
+++ b/toonz/sources/toonz/xsheetviewer.cpp
@@ -1250,12 +1250,21 @@ void XsheetViewer::keyPressEvent(QKeyEvent *event) {
     stride.setFrame(cellSel->getSelectedCells().getRowCount());
 
     if (m_cellArea->isControlPressed()) {  // resize
-      if (r0 == r1 && shift.frame() < 0) return;
+
       if (c0 == c1 && shift.layer() < firstCol) return;
-      cellSel->selectCells(r0, c0, r1 + shift.frame(), c1 + shift.layer());
+
+      bool prevRow = r0 < getCurrentRow();
+      bool prevCol = c0 < getCurrentColumn();
+
+      if (!prevRow and !prevCol)  cellSel->selectCells(r0, c0, r1 + shift.frame(), c1 + shift.layer());
+      if ( prevRow and !prevCol)  cellSel->selectCells(r0 + shift.frame(), c0, r1, c1 + shift.layer());
+      if (!prevRow and  prevCol)  cellSel->selectCells(r0, c0 + shift.layer(), r1 + shift.frame(), c1);
+      if ( prevRow and  prevCol)  cellSel->selectCells(r0 + shift.frame(), c0 + shift.layer(), r1, c1);
+
       updateCells();
       TApp::instance()->getCurrentSelection()->notifySelectionChanged();
       return;
+
     } else {  // shift
       CellPosition offset(shift * stride);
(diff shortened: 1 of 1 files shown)


flurick:
It should do the job, but I put it up as a draft at first because there are many moving parts to this. The vertical-xhseet and horizontal-timeline modes. I'm not sure of it's use case but moving with the arrows moves the size of the selection. So this seemed the most obvious way make catch all 4 possible directions of selection.

flurick:
It should do the job, but I put it up as a draft at first because there are many moving parts to this. The vertical-xhseet and horizontal-timeline modes. I'm not sure of it's use case but moving with the arrows moves the size of the selection. So this seemed the most obvious way make catch all 4 possible directions of selection.

RodneyBaker:
This PR needs a refresh to get it to pass through the checks/tests again.

cc:  @shun-iwasawa

RodneyBaker:
@shun-iwasawa Due to the age of this PR I cannot relaunch the workflows.
Can you refresh?  Thanks!

flurick:
Should I just re-request this one? @RodneyBaker

manongjohn:
Closing and opening the ticket sometimes causes the check to be performed again.

FYI, I had fixed this issue in T2D a different way than you did because when I tried something similar to your solution, I was seeing some other problems that warranted a different way for me to do it.  Most likely the way I implemented my solution is overkill and/or a terrible solution to the problem but it worked. lol

I never got a chance to test to see if your implementation exhibited the same problems I ran into and would be curious to see once it's rebuilt.

RodneyBaker:
Closing and reopening did the trick and now all checks have passed/completed.
Thanks

manongjohn:
Ok, I remeber the specific reason I chose a different method for fixing this issue.

The T2D solution addressed a scenario in which the Current Frame indicator was not the same as the Current Selection indicator. This can happen if you used something like Next/Previous drawing or possibly other keyboard frame navigation shortcuts.

In this scenario, if you tried adjusting the selection and move towards the Current Frame indicator, the selection would shift/reset towards it.  This PR doesn't fix that behavior.

It is an odd case you can decide to handle or not. I never understood why the frame and selection indicator would be allowed to separate nor why you'd change selection size not based on the current frame indicator.  At the time this scenario looked like a bug so I changed my approach to fix it. 😄.

flurick:
@manongjohn 

Yeah, the active cell and the selection are very slippery. I also noticed they can get out of sync when pasting a range of cells.

One option could be to tweak commands that change the selection to keep them in sync, 
or if there are more maybe a quicker solution, a "select none" command?

I guess it's a "normal spreadsheet" thing to select multiple separate ranges, but less likely for animation - I hope :)

shun-iwasawa:
Thank you @flurick for tackling the issue.

As @manongjohn pointed, the current frame and the cell selection can be separate, especially when you set the preference option `Interface category > Move Current Frame by Clicking on Xsheet / Numerical Columns Cell Area` to OFF.
Disabling this option would be helpful when you work on "heavy" scene where a lot of large-sized image levels are placed, and you just want to build the xsheet. Because switching the current frame every time clicking on Xsheet triggers the viewer update frequently and it will slow the operation.

Thus, I think it would better to modify the selection size without referring to the current frame or column. I'll try to add some commit based on your PR. Thanks!

shun-iwasawa:
I think I will merge this PR and then post a separate modification PR later, so that I can safely squash the commits retaining authorship.

PR #4591, opened 2022-11-05
3rd party interface adjustments + more

justburner:
This is my breakdown of #4501 into 2 parts, this PR being the first:
- A) 3rd party interface adjustments
- B) Initialization Wizard

There's been some additional bug-fixes and features added since then.

Fixes #4384
Resolves #3911, resolves #4537
(Implements #1978)

PR #4596, opened 2022-11-09
Patch for #4591

justburner:
Small patch to workaround QuickTime state because `isQuicktimeInstalled` always reports false in Toonz:
```C++
bool Tiio::isQuicktimeInstalled() {
  // NOTE: This is *NOT* the same function as IsQuickTimeInstalled(), which is
  // implemented locally in the image lib and used there. This function here is
  // actually NEVER USED throughout Toonz, so we're placing a dummy
  // implementation here.

  assert(false);
  return false;
}
```

This causes OpenToonz to assume QuickTime wasn't installed and opened the wrong interface for the MOV options after #4591:

Thanks for @RodneyBaker for discovering this bug.


RodneyBaker:
That's got it fixed.
Thanks @justburner for the fast turnaround!
LGTM

PR #4599, opened 2022-11-14
Actions: Fix MacOS runner

shun-iwasawa:
This PR will fix recent failure in MacOS runner of GitHub Actions.
Removed a line unlinking gcc v9 as it seems to be no longer used.


RodneyBaker:

PR #4604, opened 2022-11-20
Update Icons: Types

konero:
Small update for Type icons corner frame, also adds icon for new "Freepick" tool and one for Segment type.

image

_The polyline tool icons I just tweaked to fix some pixel snapping, there's no visual difference really._


RodneyBaker:
Thank you @konero for the updated icons.
LGTM

PR #3665, opened 2021-01-01
Preproduction Board feature

morevnaproject:
This PR adds new panel - Preproduction Board.

The purpose of this new panel is to assist with managing scenes and help users handle preproduction activities - such as crafting storyboard and animatic. The final goal is to provide functionality similar to [WonderUnit's Storyboarder](https://wonderunit.com/storyboarder/) or [Story Planner](http://www.toonz.com/htm/products/prodSP.htm).

At current implementation (delivered in this PR) it is capable to display scenes and you can create new scenes using dedicated button.

screenshot_002

Quick demo - https://youtu.be/UjU7BgKYq58

Tasks to do / plans:

- [ ] 1. Allow to change order of scenes
- [ ] 2. New button "Add scene named..." which asks for scene name
- [ ] 3. Allow to playback scenes together, in sequence
- [ ] 4. Option to show scenes horizontally, like in [WonderUnit's Storyboarder](https://wonderunit.com/storyboarder/)
- [ ] 5. Option to show scene length (when horizontal mode enabled)
- [ ] 6. Allow to change scene length (it will change playback range of scene)
- [ ] 7. Allow to duplicate scenes
- [ ] 8. Add background soundtrack when scenes show in horizontal mode
- [ ] 9. Export storyboard as PDF

**IMPORTANT:** My coding skills are not very strong, so apologies for code quality. I am mostly animation production manager, not a developer. My main intention was to  create a prototype to show how this feature can look like in OpenToonz. I appreciate help/advices with achieving any of TODO tasks above or improving code quality. Thank you! ^__^

PR #4611, opened 2022-11-24
Preproduction Board feature 1.6

flurick:
Rebased #3665 and _slightly_ tweaked according to those comments

### DIFF
--- a/toonz/sources/toonz/CMakeLists.txt
+++ b/toonz/sources/toonz/CMakeLists.txt
@@ -82,6 +82,7 @@ set(MOC_HEADERS
     ruler.h
     savepresetpopup.h
     scanpopup.h
+    scenebrowser.h
     scenesettingspopup.h
     sceneviewer.h
     sceneviewercontextmenu.h
@@ -285,6 +286,8 @@ set(SOURCES
     runscriptcommand.cpp
     savepresetpopup.cpp
     scanpopup.cpp
+    scenebrowser.cpp
+    scenebrowserversioncontrol.cpp
     sceneviewercontextmenu.cpp
     scenesettingspopup.cpp
     scriptconsolepanel.cpp
--- a/toonz/sources/toonz/filebrowser.cpp
+++ b/toonz/sources/toonz/filebrowser.cpp
@@ -2423,3 +2423,6 @@ void FrameCountTask::onCanceled(TThread::RunnableP thisTask) {
 
 OpenFloatingPanel openBrowserPane(MI_OpenFileBrowser, "Browser",
                                   QObject::tr("File Browser"));
+
+OpenFloatingPanel openPreproductionBoardPane(MI_OpenPreproductionBoard, "PreproductionBoard",
+                                  QObject::tr("Preproduction Board"));
--- a/toonz/sources/toonz/filebrowser.cpp
+++ b/toonz/sources/toonz/filebrowser.cpp
@@ -2351,7 +2351,7 @@ calculateTask:
 
 //-----------------------------------------------------------------------------
 
-inline void FrameCountReader::stopReading() { m_executor.cancelAll(); }
+void FrameCountReader::stopReading() { m_executor.cancelAll(); }
 
 //=============================================================================
 // FrameCountTask methods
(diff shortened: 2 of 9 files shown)


RodneyBaker:
Thanks much for this @flurick 
Very much appreciated.
Initial testing... looking good.

The only problem I've ran into thus far is the Preproduction Board entry disappearing from the Windows menu.
Restarting the program didn't bring it back... odd.
The option to assign a shortcut key to Production Board window was still present so I'm using that to open the window.

konero:
My feedback for what it's worth (I know you're not the original developer, there's no issue I can post this to, so don't worry about it).

1. Bigger thumbnails
2. Change the transparent backdrop for the thumbnail to white
3. Remove the clapperboard stripes, since this panel only displays scenes

openanim:
@flurick  First of all, thank you for developing this feature, it was really one of the few work steps that could not be done in OT, it would be great to be able to do almost everything in OT.

If I may say so, here are my feelings.

While testing this PR, I was frustrated with the scene save pop up that kept appearing, and gives a rather different experience than in other software with stroyboarding.

So I see this panel more as a way to facilitate access to the scenes and play them. And I was wondering if we couldn't have some level strip features to complete, and (from my point of view) improve the experience.

Here are some suggestions in that direction:

- Allow different level strip thumbnail sizes per room. 

I tested the PR by creating a room with a level strip with a larger thumbnail size (from the interface tab in the preferences). But it would be better if you could choose the size individually for each room.

- Allow shortcuts or buttons to add frames before and after the one selected. Currently the insert key does not allow to choose the direction of adding the frame. And the add frames command neither. Below is an example from (the storyboard function of krita) (plus button on thumbnail). Or alternatively elsewhere on the level strip.

story ot

Like @konero, it would be nice to be able to change the size of the thumbnails of the scenes (like with the level strip). By the way you can choose the color you want of the thumbnails from the menu Xsheet => scene setting => camera BG => full alpha. In my suggestion, howover I am in favor of keeping the scene strips to differentiate them from the level strip and be able to know u can click on play icon on them.

artisteacher:
As far as the level strip size goes, it really should be responsive. So when you make the level strip taller/shorter (horizontal orientation) or wider/narrower (vertical orientation), the icons automatically grow/shrink to fill the available space. There shouldn't be a hardcoded size. Then it would easily work on a room by room basis.

RodneyBaker:
I know there are lots of plans and hopes for the Production Board moving forward but I want to post my intent to merge this PR in order to finally have closure on the initial Production Board which only had a few small things keeping it from moving forward.  With those now resolved thanks to @flurick, @morevnaproject's original contribution can be added and additional changes can start to be added as they become available.

Please let me know of any reason to delay this initial feature addition any further.
Thanks!

morevnaproject:
@flurick Thanks for taking on this PR. Will be happy to see it merged! ^__^

RodneyBaker:
And... merge time.

Thank you @flurick for moving the Preproduction Board to this finish line.
Thank you @morevnaproject for setting it in motion.
I'm confident that in time it will be even more useful.

Aside:  We'll have to find a good place to maintain suggestions/links to those things that testers provided feedback on.
In the meantime there are certainly the PRs themselves.

morevnaproject:
> Aside: We'll have to find a good place to maintain suggestions/links to those things that testers provided feedback on.
In the meantime there are certainly the PRs themselves.

Many open-source projects use GitHub issues for feature requests and suggestions. Those issues are just have special tag (i.e. "feature request") to separate them from regular bugs. 
When PR is merged and some suggestion remain, those suggestions are submitted as separate issues to be addressed later. This is very helpful for project management.

RodneyBaker:
So very nice to have this feature finally in Opentoonz.

image

RodneyBaker:
The disappearance of the Production Board option from the menu observed during testing appears to also be in the nightly releases.  I am unsure why the option isn't persistent but am starting to see reports of this so thought this notice would be appropriate.

The option does appear initially after installation.

flurick:
Yeah, looks like MI_OpenPreproductionBoard needs to be added to the menubar.xml and the different menubar_template.xml's in stuff?

PR #4615, opened 2022-11-27
Allow svn versions over 1.10

flurick:
SVN decimal versions are not zero padded and instead goes: 
1.8 
1.9 
1.10
1.11


RodneyBaker:
Nice catch @flurick

flurick:
> Nice catch @flurick Question: Are you intentionally making your PRs drafts or is that a side effect of how you are submitting? (I don't have any issue with it either way... just curious)

It's mostly just because its in draft mode by default

RodneyBaker:
Thank you @flurick for the update.
LGTM

PR #4626, opened 2022-12-01
Add new brushes and fixes

andeon:
Add news brushes and fixes. 

I'm trying to improve the sketch brushes by putting some delay to stabilize the strokes more. Adjusted the opacity on some textures to avoid abrupt posterization, a good tip is use a blackish color from the same texture to "blend" the color.

Kneaded eraser is good for creating highlights or creating other details by also using it with the smear.

Keep in mind that by default the opentoonz Bg camera is alpha, so Xsheet → Scene Settings sets a color. Or set a plain color at a below level to avoid color glitches. 

e.g. Kneaded eraser, water texture or another things that make it transparency.

[aotz_v6.webm](https://user-images.githubusercontent.com/11843239/205204136-90b699cd-43d9-464b-a153-6a3c56f01a1b.webm)

I'm thinking of removing or reducing the pressure of some brushes many times we don't need that, but since opentoonz let us to turn it off when it's convenient, I'm not sure it would be necessary.

PS, Test it on appveyor artifact;


RodneyBaker:
Thank you @andeon for the additions and fixes.
Much appreciated.
LGTM

PR #4628, opened 2022-12-02
Function Editor Curve Graph: Bigger Handles & Improved Contrast

konero:
Requested by some colleagues, they wanted improvements to the curve graph, better contrast and bigger handles. So, I made the lines lighter for the dark themes, and the speed in speed out handle is now drawn as a circle.

Screenshot 2022-12-02 000527

- Fixes an issue introduced in #4572, where `#StyleChooserPage` properties were misnamed compared to the ones set in the theme variables, this caused the light theme to break and become unreadable.


adamearle:
There is a lot of stuff that was done in Tahoma with the function editor that might help lead to a solution here.

RodneyBaker:
Thank you @konero for the update!
LGTM

PR #4633, opened 2022-12-05
New Fxs: Flow Fx series

shun-iwasawa:
This PR will introduce a series of new Fx making and using the "Flow Image".
It contains 4 new fxs, 1 new vector style and 5 types of brush tip textures as follows:
#### New Fxs
- **Tangent Flow Iwa**   ; for making the flow image from image.
- **Motion Flow Iwa**   ; for making the flow image from movement.
- **Flow Blur Iwa**   ; for using the flow image as blur direction.
- **Flow Paint Brush Iwa**   ; for using the flow image as brush tip texture's direction.
#### New Style
- **Flow Line style**   ; for converting vector stroke to wide multiple strokes which is suitable for inputting to Tangent Flow Iwa Fx.
#### Brush Tip Textures
- dry_brush (19 imgs)
- fluffy (7 imgs)
- single_line (10 imgs)
- streaky (13 imgs)
- thick (10 imgs)

## Tutorial Video
Part1: https://youtu.be/DS2kw88hVOs
Part2 is not yet released.

## What is the Flow Image ?
The Flow Image contains directional (or vector) data in each pixel. 
In each pixel, red and green channels represent horizontal and vertical component of the flow direction respectively.
The blue channel represents an intensity which is used as a supplementary information in some Fxs.

flow_image
_Left:_ Flow directions and correspondent colors in the Flow Image.  _Right:_ Flow Image overlapped with the flow directions.

## Tangent Flow Iwa Fx
Tangent Flow Iwa Fx generates the flow image from the source image.
This is based on an implementation of a paper ["Coherent Line Drawing" by H.Kang et al, Proc. NPAR 2007](http://www.cs.umsl.edu/~kang/Papers/kang_npar07_hi.pdf).
Intensity of the Sobel Filter will be stored in the blue channel.
 
tangentflow1

### Parameters
- **Iteration** specifies how many times the filtering will be applied. The larger the iteration amount, the more noises due to small features are reduced.
- **Kernel Radius** specifies the size of filter. The larger the filter size, the larger features will be taken into account.
- **Threshold** specifies the minimum difference of brightness the fx will recognize as feature.
- **Align Direction** when ON, the flow vector will be inverted if the vector faces backward to the Pivot Angle.
- **Pivot Angle** specifies the center angle for aligning the flow vector.

## Motion Flow Iwa Fx
Motion Flow Iwa Fx generates the flow image from the movement of specified object.
Normalized speed will be stored in the blue channel.

motionflow
_Up:_ Results of Motion Flow Iwa Fx with different types of movement of a column. _Down:_ Results of Flow Blur Iwa Fx using the above images as flow references.

### Parameters
- **Reference Object** and **Index** specify the stage object of which the movement will be sampled.
- **Shutter Length** specifies the offset from the current frame where the affine transformation will be sampled.
- **Normalize** specifies how to normalize the speed value (and store to blue channel of the result).
    - **Auto** uses the maximum value in the camera frame.
    - **Manual** uses the value of Maximum Length 
- **Maximum Length** specifies the speed value to be used for normalizing

## Flow Blur Iwa Fx
Flow Blur Iwa Fx will blur each pixel in the image along the directions specified by the Flow Image.
In contrast to the other blur fxs (like directional, radial, spin or motion blur), the direction of the blur can be arbitrarily specified.

flowblur
_Upper Left:_ Source image. _Upper Right:_ Vector image specifying the trajectory of motion. Note that the new Flow Line style is applied to the stroke. _Lower Left:_ The flow image generated by Tangent Flow Iwa Fx. _Lower Right:_ Results of Flow Blur Iwa Fx using the flow image.

### Parameters
- **Length** specifies the size of blur 
- **Filter Type** specifies how the intensity decays along the trajectory of the blur.
    - **Linear** / **Gaussian** / **Flat**
- **Linear Color Space** when ON, computation will be done in the linear color space.
- **Gamma** specifies the linear color space gamma.
- **Reference** specifies the source for the reference image which can adjust the blur length.
    - **Reference Image**: Brightness of the image connected to the Reference port (if any) will be used.
    - **Blue Channel of Flow Image**: Suitable for using the result of Motion Flow Iwa as the Flow Image.
  
## Flow Paint Brush Iwa Fx
Flow Paint Brush Iwa Fx will place textures of brush tip in the specified region, along the direction of the Flow Image.
It is originally intended to create effects of scratches, dirt or stains on animated characters.
In Japanese animation industry, such effects have been usually done in the workflow of 特効 ( [tokkou],  the abbreviation of "特殊効果" = special effect ). The 特効 staffs manually paint the effect frame by frame, by using PhotoShop.
This Fx can also be exploited for painterly rendering.

fpb1
Scratches and dirts on the character. Note that the pattern distorts according to the turning action.

fpb2
Example of the painterly rendering.

Please refer to the [tutorial video](https://youtu.be/DS2kw88hVOs) for usage in detail.

### Input Ports
- **Brush** : Brush tip textures
- **Flow** : Flow image. If not connected, all brush tips will be placed in a direction parallel to the vertical edge of the parallelogram.
- **Area** : Area image in which the brightness of the pixels is propotional to the brush tip density. If not connected, all the brush tip in the parallelogram range will be generated.
- **Color** : Specifies the color of the brush tip by picking the color at the tip position. If not connected, The Brush texture will be rendered as-is.

### Parameters
- **Tip Placement**
    - **Horizontal** and **Vertical Density** specify the amount of the brush tips to be placed.
    - **Position Randomness** specifies the random-offset ratio from aligned lattice position of the brush tips.
    - **Position Wobble by Frame** will make each tip randomly offset from the initial position frame by frame.
- **Source Area**
    - **Origin**, **Horizontal** and **Vertical** positions define a parallelogram region where the brush tips will be placed. The brush tip pattern will deform according to the parallelogram.
    - **Curve Anchor** will add a small bulge to the brush tips placement so that the pattern looks more natural when putting on the curved surface of the character.
    - **Fill Gap Size** will dilate then erode the Area image with specified size before the brush tip generation. It is useful when the area is specified by parallel hatching in the Area image.
- **Brush Tip Geometry**
    - **Tip Width** and **Tip Length** specify the size of the brush tips. They are expressed as in ratio of the size of the parallelogram region. The value range will correspond to the Area image value.
    - **Width** and **Length Randomness** specify the maximum ratio of random variation in the brush sizes.
    - **Sustain Width to Skew** will adjust width of each tip in order to keep them visible even if the parallelogram region is skewed.
    - **Angle Randomness** specifies the maximum degree of random variation in the brush direction from the original angle defined by the Flow image.
    - **Tip Alpha** specifies the opacity of the brush tips. The value range will correspond to the Area image value.
    - **Tip Joints** specifies the amount of polygons to render each brush tips. The larger the value, the smoother the strokes.
    - **Bidirectional** : When ON, brush tips direction will be randomly flipped against the flow direction.
    - **Prevent Jaggies by Texture Margin** : When the brush tip is very thin, this option will prevent jaggies by adding extra margins to the texture.
- **Reference Frame**
    - **Reference Frame** : If some frame is specified, the Area image will be the composition of the current frame and the reference frame. It is usefull to make the pattern shape consistent through frames.
    - **Reference Prevalence** specifies the mix ratio of the Area image of the reference frame.
- **Random Seed**
- **Sort By** specifies the stacking order of the brush tips.

## New Vector Style : Flow Line
The Flow Line style will convert a single stroke to multiple, wide and parallel lines. It is suitable for making the flow image from sparse "stream lines" by using the Tangent Flow Iwa Fx.

image
_Left_ : Original vector image. _Right_ : The same image with the Flow Line style applied.

## Brush Tip Textures
Brush tip textures are also included in this PR, suitable for the Brush input for the Flow Paint Brush Iwa Fx.
The textures are stored in `$TOONZLIBRARY/textures/brush tips` .
image
The textures were painted with water color and scanned by me.


RodneyBaker:
I'm going to press forward with merging this PR as it is complex enough with many variables that need to be tested more broadly.  Getting it into the hands of users via the nightly release is therefore appropriate.

Thank you @shun-iwasawa this highly beneficial addition to Opentoonz.
(BTW:  Thanks also for the IwaWarp application which also will take time to master!)
LGTM

PR #4634, opened 2022-12-05
Fix Radial and Spin Blur Ino offset problem

shun-iwasawa:
This PR fixes the following problems with Radial and Spin Blur Ino Fxs:

- The result of fxs offsets by 0.5 pixels to lower-left.

The problem appears prominent where the blur intensity is small. Especially when specify `Radius` to some value, the fxs cause irregular gap at the boundary of the non-blurred region.

radial
GIF showing the problem in Radial Blur Ino Fx. Note that only outside of the "Radius" circle slightly offset to lower-left.

### DIFF
--- a/toonz/sources/stdfx/igs_radial_blur.cpp
+++ b/toonz/sources/stdfx/igs_radial_blur.cpp
@@ -119,9 +119,8 @@ public:
 
     const QPointF center(this->center_.x, this->center_.y);
 
-    /* Pixel位置(0.5 1.5 2.5 ...) */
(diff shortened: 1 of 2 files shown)


RodneyBaker:
@shun-iwasawa The WIndows nightlies are failing to build again so I am hesitant to merge any PRs at this time.

shun-iwasawa:
@RodneyBaker 
I rebooted the Windows jen.kins machine. Thanks!

RodneyBaker:
Thanks @shun-iwasawa 
And... Thank you for this Fix to Radial and Spin Blur Ino FX!

PR #4639, opened 2022-12-09
Fix short-play only playing the last few frames

flurick:
From "[Make short play work universally #105](https://github.com/tahoma2d/tahoma2d/pull/105)"

### DIFF
--- a/toonz/sources/include/toonzqt/flipconsole.h
+++ b/toonz/sources/include/toonzqt/flipconsole.h
@@ -272,6 +272,7 @@ public:
     doButtonPressed(button);
     setChecked(button, !isChecked(button));
   }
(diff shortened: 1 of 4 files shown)


flurick:
> Even though Jeremy's github account no longer exists, he should still get Co-Authored By credit for his work.

I agree, what would be a good workflow for that in github?

manongjohn:
To give proper credit, you need to add this line in the actual commit message:

> Co-authored-by: Jeremy Bullock <jcbullock@gmail.com>

RodneyBaker:
Oh yeah.  This makes the short play feature actually usable.
Thank you @flurick for the port of Jeremy's fix!
LGTM

PR #4643, opened 2022-12-12
Fix Output Settings not to set the dirty flag when unchanged

shun-iwasawa:

- Just opening the Output Settings popup makes the scene dirty (i.e. the asterisk appears next to the scene name on the title bar) .


RodneyBaker:
The Mac Mojave checks stalled for this PR but I don't see any issue pressing forward.

LGTM

PR #4646, opened 2022-12-14
Timeline/Xsheet navigation tags

manongjohn:
This is a requested port of Tahoma2D's Navigation Tags for the Timeline/Xsheet

image

Command/Quicktool bar buttons:
image
(`Toggle Navigation Tag`, `Previous Tag`, `Next Tag`....If someone feels like making better icons, feel free.)

Navigation Tag features:
- You can tag the currently selected scene frame with the `Toggle Navigation Tag` right-click menu option or command
- Move quickly to a tag using the `Next Tag` and `Previous Tag` right-click menu options or commands
- A `Tags` list is maintained which allows you to see all tag and jump to any of them quickly 
- The default tag label is `Frame #`, but you can `Edit Tag` and add additional message to it
- The default tag color is Magenta, but you can `Edit Tag` and change it
- Hovering over the tag will show the tag's label as a tooltip
- Can click+drag an existing tag to move it
- Tags will shift with `Insert Frame` and `Remove Frame` accordingly

WARNING: Tag information is saved in the scene file.  If you save a scene with tags, the scene will not open in prior versions of T2D or OT that do not support Navigation Tags.  If you need to share with older versions or OT, use the `Remove Tags` options in the `Tags` list.


RodneyBaker:
Initial testing... everything appears to be working as expected.
Thanks for the detailed write up @manongjohn.

RodneyBaker:
Posting intent to merge this feature into Opentoonz.

Note:
Outside the scope of this PR, there is a broader discussion to be had related to the number of different navigational and notational features Opentoonz has that are all independent of each other and yet complimentary.  It will be good to consider workflow of each and how they can be used most effectively with one another. This to also gain capabilities such as use of these tags/labels to drive other processes such as selective viewing, rendering and animation.  

Also of note to users, the use of this and other similar features is enhanced by use of toolbars and custom panels.  With custom panels for instance we can scroll back and forth through the navigation tags quickly by setting the in field to Previous Tag and the out field to Next Tag.

RodneyBaker:
Will press forward with this as I know a few who are eagerly awaiting use.

Thank you @manongjohn for the feature.
LGTM

JPUser1:
Thank you! @manongjohn  😁

PR #4650, opened 2022-12-15
Refactoring: Unify Viewer and Comboviewer base classes

shun-iwasawa:
This PR wil unify the `SceneViewerPanel` and the `ComboViewerPanel` classes and make them to be the subclasses of a `BaseViewerPanel` base class, because most of the source code is identical.

### DIFF
--- a/toonz/sources/toonz/comboviewerpane.cpp
+++ b/toonz/sources/toonz/comboviewerpane.cpp
@@ -1,78 +1,30 @@
-
-// TnzCore includes
-#include "tconvert.h"
-#include "tgeometry.h"
-#include "tgl.h"
-#include "trop.h"
-#include "tstopwatch.h"
-#include "tsystem.h"
-
-// TnzLib includes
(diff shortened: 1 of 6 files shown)


artisteacher:
Does the viewer have any performance benefit over the combo viewer? The combo viewer just seems so much more flexible, since visibility of the tool options and toolbar can be toggled on and off. I guess I just wondered if there's a reason we have both?

RodneyBaker:
Posting intent to merge as everything looks good here.

RodneyBaker:
Thank you @shun-iwasawa for the refactoring of viewer base classes.
LGTM

RodneyBaker:
@shun-iwasawa I don't see it in the code for this PR but we recently lost the default setting that turns on the Frame Slider for the Viewer.  I plan to search it out but perhaps you'll know exactly where to look for it.

It does NOT appear to me that this PR turned the default option off but this is me starting the search...

PR #4670, opened 2022-12-22
Fix Particles Fx column placement bug

### LINKED ISSUE #4669: Placement of a Particles Fx column affects particles movement
### Description

When the Particles Fx column offsets from the origin, the particles movement becomes wrong.
This problem occurs after merging #4075 . (It's my PR.. So sorry for the trouble!)

### Steps to Reproduce

1. Create Particles Fx. Load some level for a particle texture. (In this example, a red triangle)
    particles_bug1

2. Set `Animation > Rotation > Rotation Speed` of the Particles Fx to 5. Keep other parameters unchanged from the default.
3. Render the scene. The result is like this:
   particles_no_offset

4.  Move the column containing the Particles Fx upward.
    particles_bug2

5. Render the scene again. The result becomes like this: 
    particles_with_offset

The particles movement is unexpectedly changed.

### Expected Behavior

The column placement should not affect the particles movement. It should just translate the particles as a whole.

_No response_

### OpenToonz Version

Other (Specify below)

Current master

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

None

shun-iwasawa:
This PR fixes #4669 , making not to include Particles Fx column's transformation to the Texture column.

### DIFF
--- a/toonz/sources/toonzlib/scenefx.cpp
+++ b/toonz/sources/toonzlib/scenefx.cpp
@@ -1018,12 +1018,13 @@ PlacedFx FxBuilder::makePF(TZeraryColumnFx *zcfx) {
   PlacedFx pf;
   pf.m_columnIndex = zcfx->getColumn()->getIndex();
 
-  // Add the column placement NaAffineFx
-  if (!getColumnPlacement(pf, m_xsh, m_frame, pf.m_columnIndex, m_isPreview))
-    return PlacedFx();
-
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I confirm the problem and that this PR addresses the issue.
Thanks @shun-iwasawa for the fix.
LGTM

WIll merge although one of the Mac builds is still not operational.

PR #4672, opened 2022-12-23
Actions MacOS runner: Remove symlinks before installing Python

shun-iwasawa:
This PR will fix recent failures in Github Actions MacOS runner.
When installing Qt 5 via homebrew, it failed when linking the dependent python (v3.10) library since there is a pre-installed newer version (v3.11) in MacOS runner environment.
This PR will remove existing symlinks before installing python.


RodneyBaker:
Thank you @shun-iwasawa!

PR #4680, opened 2022-12-29
Straight line mode on brush tool

justburner:
This is a Tahoma2D port.

Hold `Shift` before a stroke to activate Straight Line mode, holding `Control` will angle snap to 45 degrees.

The `Control` modifier can also be used to activate Straight Line mode allowing to angle snap without requiring `Shift`.

`Raster`, `Toonz Raster` and `Vector` levels are supported.

Changes compared to Tahoma2D:

Note that in Vector level it works differently, OpenToonz already has straight line on brush tool implemented for vectors, for this one you can hold `Shift` anytime during drawing a stroke so having it held before a stroke isn't required, this PR only adds the ability to angle snap to 45 degrees for vectors while keeping the original implementation intact.

End tip pressure is based of the maximum pressure value instead of the initial pressure value (on my tablet initial value is always near 0.0, this resulted lines on T2D to always having lowest pressure value).
This change should allow more pressure control on the line stroke, if uniform pressure is required then disabling Pressure is recommended (this was the only way it worked reliably on Tahoma2D for me anyway).
Start tip is always based on the initial pressure value as changing that behavior should be quite difficult.

### DIFF
--- a/toonz/sources/include/toonz/rasterstrokegenerator.h
+++ b/toonz/sources/include/toonz/rasterstrokegenerator.h
@@ -67,7 +67,7 @@ public:
   void add(const TThickPoint &p);
 
   // Disegna il tratto interamente
-  void generateStroke(bool isPencil) const;
+  void generateStroke(bool isPencil, bool isStraight = false) const;
 
   // Ritorna m_points
   std::vector<TThickPoint> getPointsSequence() { return m_points; }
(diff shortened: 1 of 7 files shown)


RodneyBaker:
Posting intent to merge.  Still doing some additional testing but if all is well this PR will likely be the last for 2022...

RodneyBaker:
Thank you @justburner for the update.
LGTM

PR #4681, opened 2022-12-29
Paste external clipboard images

### LINKED ISSUE #3503: Copy/Paste from external programs
The ability to copy and paste from external programs into Opentoonz will provide a good foundation and flow of image data into Opentoonz.   Up until this point copy/pasting has been restricted to internal processing only with the exception of (mostly) textual data.   A near equivalent is drag/dropping of images into Opentoonz but that does not quite meet the workflow needs of copy/pasting directly into Opentoonz.

This feature is currently operational in Tahoma and porting to Opentoonz master will be done after testing and consideration of whether the feature is best added 'as is' or modified.  The scope of copy/pasting in general is rather large but this basic copy/pasting capability should fill the immediate requirement of Opentoonz production.  Additional copy/pasting features will likely be added as more is understood about the requirements of OT user productions.

The feature should be cross platform as it uses Qt.

REF:  https://github.com/turtletooth/tahoma/pull/202

Tagged as Roadmap

justburner:
This is a Tahoma2D port with the following changes:
- Pasting on a `Raster` level cell or the empty X cell in front will automatically paste in the center instead of nagging the user by always asking if want to `Paste in place` or `Create new level` on each paste.
- Attempting to paste on non `Raster` level cell or empty column/location will ask the user if wants to create a new level instead of creating a new level automatically without user's consent, this is to avoid polluting the project with extra files as undo does not delete them.

More information about the T2D implementation:

- External images do require `Raster` level.
- Selection Tool must be active to be able to control and transform the pasted image, otherwise the image will be instantly pasted in the center.
- If the image is bigger than the level canvas then the paste will be scaled down to fit the canvas.

Resolves #3503

### DIFF
--- a/toonz/sources/include/tools/rasterselection.h
+++ b/toonz/sources/include/tools/rasterselection.h
@@ -56,7 +56,7 @@ class DVAPI RasterSelection final : public TSelection {
   bool m_noAntialiasing;
 
 private:
-  void pasteSelection(const RasterImageData *data);
+  bool pasteSelection(const RasterImageData *data);
 
 public:
   RasterSelection();
--- a/toonz/sources/tnztools/rasterselection.cpp
+++ b/toonz/sources/tnztools/rasterselection.cpp
@@ -28,6 +28,7 @@
 #include "toonz/tframehandle.h"
 #include "toonz/txsheethandle.h"
 #include "toonz/tstageobject.h"
+#include "toonzqt/gutil.h"
 
 #include <QApplication>
 #include <QClipboard>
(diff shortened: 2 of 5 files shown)


RodneyBaker:
Happy new year to all!
Posting intent to merge this PR later today.

RodneyBaker:
Thank you @justburner for the port of copy/paste from Tahoma2D and for additonal adjustments.
Thank you @jeremybullock for the original implementation!

F-Burning:
Thank you so much!

A suggestion is don't crop or resize the image, because if you have to re-scale the raster after the resize, lose a lot quality, and when a picutre is large in horizontal at least, it's cropped and lost information too.

PR #4700, opened 2023-01-10
Floating point & Linear color space rendering

shun-iwasawa:
This PR will enable OT to render in floating point pixel, i.e. each channel in pixel has 32bit float value.
Using this feature you can compute the scene in high-dynamic range (HDR), which is necessary to meet some premium cinema requirements such as [Dolby Cinema](https://en.wikipedia.org/wiki/Dolby_Cinema).

#### Output / Preview Settings
- Added a new option `32bit Floating point` to `Channel Width`.
    - HDR images can be rendered with this option. 
    - Note that some Fxs (as listed below) does not support floating point rendering.  When an unsupported Fx is used, the output from that Fx node will be clamped to 0-1.
- Added `Linear Color Space` check box.
    - This option is referenced when the `Color Space` combo box (see below) in Layer Blending **Ino** Fxs is set to `Auto`.
    - From Layer Blending Fxs, this option is always referenced.
- Added `Color Space Gamma` parameter, which is used to convert images between linear and nonlinear color spaces.

#### Layer Blending Ino Fxs
- Replaced the `Linear Color Space` check box by a `Color Space` combo box with options as follows:
    - **Auto** (default) : Same as the `Linear Color Space` option in Output/Preview settings.
    - **Linear** : Always compute in the linear color space. Color space gamma can be adjustable with the `Gamma Adjust` parameter.
    - **Nonlinear** : Always compute in the nonlinear color space. 

#### Layer Blending Ino Fxs, Adjust Exposure Iwa, Motion Blur Iwa, all Bokeh Fxs, and Bloom Iwa
- Replaced the `Gamma` parameter by `Gamma Adjust`. Which is an offset value from `Color Space Gamma` specified in Output/Preview settings.

#### All Bokeh Fxs
- Added `Linearize Mode` combo box which enables to choose how to convert color spaces, from Hardness-based (conventional way) and Gamma-based (newly added way).

#### EXR format I/O
- HDR is now supported.
- Added `Float` to the `Bit Per Pixel` output property, in addition to `Half Float`.
- Added `Color Space Gamma` to the output property to specify gamma value when saving.
- Enabled to specify loading gamma value in the Level Settings.

#### Flipbook
- Gain can now be changed when displaying a raster image in Floating Point format.
    image
    image

#### Histogram
- When displaying a raster in floating point format, the upper range of the histogram can now be expanded.
    

-----
### Effects that do NOT support Floating Point calculations
- Blur
    - Line Blur Ino
    - Local Blur
    - Motion Blur
    - Motion Blur Ino
    - Radial Blur
    - Spin Blur
    - Flow Blur Iwa
- Distort
    - Free Distort
    - Linear Wave
    - Random Wave
    - Floor Bump Iwa
- Gradient
    - Diamond Gradient
    - Four Points
    - Spiral
    - Square Gradient
- Image Adjust
    - Despeckle
    - HSV Scale
    - Multitone
    - Negate Ino
    - RGBA Cut
    - RGBA Scale
    - Sharpen
- Light
    - Backlit
    - Body Highlight
    - Cast Shadow
    - Fog Ino
    - Glow
    - Light Spot
    - Target Spot
- Matte
    - Erode/Dilate
- Noise
    - Dissolve
    - Median Ino
    - Median Filter Ino
    - Noise
    - Salt Pepper Noise
- Render
    - Clouds
    - Particles
    - Tiled Particles
    - TimeCode Iwa
    - Text Iwa
    - Flow Paint Brush Iwa
- Stylize
    - All Fxs other than Max Min Ino
- Toonz Level
    - All Fxs
- Shaders
    - All Fxs

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -822,7 +822,7 @@
   <item>"STD_particlesFx.pick_color_for_every_frame"	"Pick Control Image's Color for Every Frame"	</item>
   <item>"STD_particlesFx.perspective_distribution"	"Perspective Distribution"	</item>
   <item>"STD_particlesFx.motion_blur"	"Motion Blur"	</item>
-  <item>"STD_particlesFx.motion_blur_gamma"	"Gamma"	</item>
+  <item>"STD_particlesFx.motion_blur_gamma_adjust"	"Gamma Adjust"	</item>
 
   
  <!------------------------------ Function Editor ------------------------------------------->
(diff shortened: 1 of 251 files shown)


artisteacher:
I'll be excited to try this out when the macOS action build succeeds!

artisteacher:
I'm hoping this will help with the color shift that my students get in their renders. It can be subtle, but colors seem less intense. It might be more obvious in colors with a lot of red. This is definitely an issue on macs, but has been reported on Windows in one of the Discord groups. Possibly related to a difference in color profile/colorspace between OT and the device.

artisteacher:
Looks like there is still a color shift:
Color shift
The left shows what a PNG render looks like while the right is the original color in the Viewer.

shun-iwasawa:
@artisteacher 
I'm not sure, but does it appear to be related to Apple's gamma shift issue..?
https://www.youtube.com/watch?v=YwjLuUR51gE

artisteacher:
In the video description, an addendum says that 2021+ macs should use the HDTV Bt.709-BT.1886 preset. That does seem to fix the color shift outside of OT for renders with this build on my MacBook. 

The problem with just changing the display profile is that there's still the potential for the shift on other devices - so the color won't necessarily be consistent. Also at least one Windows user on Discord reported the same thing.

Would including the correct color profile in the render itself make the colors display more reliably across a broader range of devices/display profiles/players?

RodneyBaker:
@shun-iwasawa This is mostly unrelated to this PR but I can't help but wonder if in the near term use of the EXR format (importing) could be made to ignore additional channels rather than refuse to open in Opentoonz.  My thought being that getting some of the EXR image into Opentoonz (i.e. the core RGBA image) would be better than not getting it in at all.

I do understand that in the interim a conversion of the EXR file to another format will work just as well.
Perhaps that might be an option; to warn the user that the EXR files cannot be open due to unrecognized channels and give the user an option to convert/import into the PNG (or TIF) format.

Another point of interest is in the exposures added in this PR which mostly align with standard camera F-stops.
There I note that this PR adds values inbetween standard F-stops but also appears to round one of the values up.
I don't really know enough to ask a question here but find myself interested in that.

Thus far this PR is looking good and it's exciting to see this development.

RodneyBaker:
Posting intent to merge as all appears to be working as intended.

RodneyBaker:
Thank you @shun-iwasawa for foating point and liner color space rendering.
LGTM!

PR #4702, opened 2023-01-11
Actions MacOS runner fix: Remove symlinks before installing Python

shun-iwasawa:
This PR will resolve recent failures in GitHub Actions MacOS runner.

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -8,13 +8,26 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    - name: Remove symlinks
+      run: |
+        # remove existing symlinks before installing python@3.10 and 3.11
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@shun-iwasawa It seems that each time the Mac side of building gets fixed... the windows side breaks (and vice versa).
I'm not sure if this will continue but at present that appears to be the case.

shun-iwasawa:
The Jen.kins Windows virtual environment that generates the nightly builds gets periodically (at intervals that cannot be postponed beyond 35 days) Windows Updates, which require me to re-start Jen.kins each time.

Also, both the MacOS and Windows Jen.kins environments are on a single Mac mini, and if the Mac mini is turned off for any reason, both will stop.

I don't know about the cause of only the MacOS Jenkins environment stopping, but I don't think it has any correlation with the state of the Windows environment.

PR #4706, opened 2023-01-12
Straight Line Mode on Toonz Raster: Have maximum thickness when drawing with mouse.

justburner:
This is something missing on #4680 that was pointed by a user in Discord.

Making a straight line with mouse resulted in minimum thickness line instead of maximum, this whatever only affected regular brushes on Toonz Raster.

Additionally I've got rid of the first dummy stroke at the first press in Straight Line Mode.


RodneyBaker:
Thank you @justburner for the addition.
The speed to which you went from the missing element being discovered to submitting this PR is also much appreciated.

PR #4709, opened 2023-01-15
Triangle Brush Cursors

konero:
This adds triangle cursors for the brush, I tried to make them identical to the ones found in Clip Studio Paint. They're extremely useful for cases when the drawing tablet for whatever reason due to parallax or hardware design issues where the cursor is not completely aligned with the pen tip and depending on handedness or posture you can pick any angle to help.

brush_triangle_bottom_left brush_triangle_bottom_right brush_triangle_top_left brush_triangle_top_right brush_triangle_down brush_triangle_up brush_triangle_left brush_triangle_right


DarrenTAnims:
How do we pick the different cursors?

RodneyBaker:
Select the various cursors in Preferences via the dropdown menu:

image

Aside:
There has been a few requests for the ability of users to change cursors locally to customized icons but as of yet this is still not implemented.  Cursor images are contained in the 'resources' directory which get compiled at build time. As such the only way to customize cursors is to replace those images and then re-build Opentoonz.

RodneyBaker:
Thank you @konero for the additional cursors.
And for your continued diligence updating the Opentoonz UI.
(My personal favorite cursor is probably (triangle bottom left)  ;)

LGTM

PR #4712, opened 2023-01-18
Viewer preview feature enhancement

shun-iwasawa:
This PR will enhance the preview feature on the Viewer / ComboViewer i.e. the "eye" button on the top-right corner of the panel.

- Added options for the preview behavior which can be selected from the right-click menu of the eye button.
    image
    - **Current frame** : Renders only the current frame in single thread. ( Conventional behavior. Selected by default )
    - **All preview range frames** : Renders all frames in the preview range in multi-thread. It's just like the `Preview` command but displayed the result in the Viewer, instead of the Flipbook.
    - **Selected cells - Auto play** : When Xsheet cell is selected, OT renders frame range of such selection in multi-thread. Once the rendering is completed, OT automatically start loop-playing the previewed frames.
- Enabled to set shortcut keys for toggling the preview and the sub-camera preview. The shortcut key will take effect to the current active viewer. 
    - Also you can set the command to menu bar, command bar and custom panels.
    - The commands are in `Menu Commands > Render` category.
image
 
 - Enabled to set sub camera rectangle larger than the original camera size.
     - It is useful when previewing the animation which are drawn in larger size than the camera. This is often the case in scenes with camera work.
     -  When dragging the sub-camera preview frame, it now snaps when the cursor is close to the original camera frame.

### DIFF
--- a/toonz/sources/include/toonzqt/flipconsole.h
+++ b/toonz/sources/include/toonzqt/flipconsole.h
@@ -276,6 +276,7 @@ public:
     setChecked(button, !isChecked(button));
   }
   void setStopAt(int frame);
+  void setStartAt(int frame);
 
   // the main (currently the only) use for current flipconsole and setActive is
   // to
@@ -356,6 +357,8 @@ private:
   int m_from, m_to, m_step;
   int m_currentFrame, m_framesCount;
   int m_stopAt = -1;
+  int m_startAt =
+      -1;  // used in the "play selection" mode of the viewer preview
   ImagePainter::VisualSettings m_settings;
 
   bool m_isPlay;
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -467,6 +467,9 @@ centralWidget->setLayout(centralWidgetLayout);*/
   if (TSystem::doesExistFileOrLevel(TFilePath(ffmpegCachePath))) {
     TSystem::rmDirTree(TFilePath(ffmpegCachePath));
   }
+
+  connect(TApp::instance(), SIGNAL(activeViewerChanged()), this,
+          SLOT(onActiveViewerChanged()));
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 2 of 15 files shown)


RodneyBaker:
In testing this I've become a little distracted by features I didn't know existed in Opentoonz.
Specifically, the ability to save/render out selections based on what we preview.
image

With this approach it's fairly trivial (and apparently has been since 2016) to render out say... only odd frames... or only even... or sections of a sequence.  And, while currently not automated, we might preview only the keys or frames that with specific cellmarks, memos, labels...

All this to suggest that I'm distracted but trying to focus as I have barely begun to explore.  
The ability to preview (and then render out) selected areas larger than the original camera size is... incredibly useful.
The shortcutting is something of a surprise as I hadn't anticipated that and... all of this is the viewer.  :)

Looking very good.

RodneyBaker:
The subcamera extension makes that feature a very powerful cropping tool.
image

F-Burning:
Can you show me where I could download it to test? The new sub-camera option is a game changer.

RodneyBaker:
> Can you show me where I could download it to test? The new sub-camera option is a game changer.

At the bottom of every Pull Request (PR) there are a set of automated checks/tests that get ran and these are all builds of the specific code included at the time of the PR being generated.  Clicking on the Details button for the build of interest allows us to dril down and find the artifact/build that was generated by the automated build process.

In the case of the WIndows github build if we drill down we eventually get to this:

https://github.com/opentoonz/opentoonz/suites/10432940299/artifacts/516027036

This being a zipped up portable build for Windows.
Just download, extract and run.
Note that while downloading I will rename the file to something easier to find such as "Opentoonz (Viewer Preview Enhancement). That way when I extract the contents of the zip file it automatically creates a directory with that name containing the contents.  In this way I can easily step into any number of builds to test changes, make comparisons, etc.

More info than requested but hopefully some of this will be useful to folks reading at some later time.  :)

F-Burning:
Thank you as always @RodneyBaker 😃 👍

RodneyBaker:
Posting intent to merge.
There is a lot to explore in this enhancement.

RodneyBaker:
Thank you @shun-iwasawa for the enhancement.
Much appreciated.
LGTM

RodneyBaker:
@shun-iwasawa I have noted two issues that loosely relate to this PR.

1) Sub-camera framing
At times it is now almost impossible to see the frame/rectangle being drawn for the sub-camera.
In a simple test I have no problem seeing the frame but at times it is invisible.

I have not yet looked at where in the code the outline might be able to be enhanced.
Perhaps it can be enhanced via Preferences Additional Style sheet as well but this may need to be changed in the actual code.

2) Define Sub-Camera icon overly persistent
When the sub-camera is on it is easy to forget to turn it off and that can lead to moments where using other parts of the UI seem not to work... and they don't... because the icon for Define sub-camera is still on.

It's not quite clear to me what a fix might be here.  Having the Define Subcamera icon toggle itself off after drawing the sub camera frame doesn't seem optimal.  Perhaps if item one above is addressed the presence of the sub-camera frame will be more obvious.

Issue #2746, opened 2019-09-01
[Discussion] OpenToonz development


shun-iwasawa:
Dear OpenToonz community members,

I post this comment is to let you know my recent situation and to clarify a policy for reviewers to judge which PR can be merged to OT.

I had consulted on management of open-source community of OpenToonz as "paid work" since it had been released. The situation has changed a month ago, I started a new career and now I'm working on a voluntary basis.

Thanks to many motivated and highly-skilled contributors, many wonderful PRs are already waiting for review, but I worry about that it may need more time to be merged in the future because I can work on them during night-time and weekend.

It is not my intention to interrupt the progress of the software due to my delay. Therefore, here I would like to suggest a basic policy for development of OpenToonz so that other members can review and merge more PRs without hesitating.

The policy consists of one principle and two supplementary notes:

- PR can be merged if it will not interfere existing usages.

note, however, that

1. If there is the possibility that introducing PR may conflict OT's existing license, please wait for owner's review.
1. From several month before releasing stable release date (to be announced), please do not merge "feature PR"s.

OT has many features which enables various way of producing animation.
A modification for some workflow may become disadvantage for others. In such case please consider to make it optional (via user profiles) so that users can select whether to use it.

Here are examples:

- Bugfix : Fine to be merged.
- Refactoring : Fine to be merged.
- Enhancement / Speed-up : Fine to be merged.
- Translation : Fine to be merged.
- New Fx : Fine to be merged.
- New command : Fine to be merged.
- New panel : Fine to be merged.
- New preset (fx, camera settings, shortcut, etc.) : Fine to be merged.
- Altering default user profiles (preferences, panel layout and shortcut) : Fine to be merged.

- Altering UI which cannot be controlled user profiles (like adding buttons to the panel or adding commands to the context menu) : Basically it's fine to be merged, but please be careful if it will not disturb the conventional usage.
- Altering existing fx or rendering behavior : Basically it's fine to be merged, but please ensure that it will not change a rendering result of a scene which is made with the previous version.

As always, any contributions are much appreciated.
Please let me know if there are any opinions or questions to the above.

Thanks!

turtletooth:
@shun-iwasawa Is there a way to assign Jenkins privileges to other developers?

gab3d:
@shun-iwasawa 

> Please let me know if there are any opinions or questions to the above.

a million questions! 😄 
[...]

PR #4715, opened 2023-01-21
Add Foreground Alpha Option for HSLBlend Shader

konero:
Adds an option to use the alpha of the foreground. For example, currently using a color card connected to the blend port changing the alpha produces a strange result, enabling the alpha option produces a result I would expect.

I'm an absolute beginner at GLSL (literally just going by the ReadMe) so please don't merge until:
1. Someone checks the code properly
2. That this doesn't modify the original intent of the shader
3. That the original implementation isn't actually a bug

Based from: #4146

### DIFF
--- a/stuff/library/shaders/programs/HSLBlendGPU.frag
+++ b/stuff/library/shaders/programs/HSLBlendGPU.frag
@@ -11,6 +11,7 @@ uniform mat3      outputToInput[2];
 uniform bool bhue;    // Blend HUE?
 uniform bool bsat;    // Blend Saturation?
 uniform bool blum;    // Blend Luminosity?
+uniform bool matte;   // Use foreground alpha?
 uniform float balpha; // Blending Alpha
 uniform bool bmask;   // Base mask?
 
@@ -114,7 +115,11 @@ void main( void )
   // Perform blending
   if (fg_alpha > 0.0 && bg_alpha > 0.0) {
     vec3 o_pix = SetLumSat(bhue ? fg_pix : bg_pix, bsat ? fg_pix : bg_pix, blum ? fg_pix : bg_pix);
-    gl_FragColor.rgb = mix(bg_pix, o_pix, balpha);
+    if (matte) {
+      gl_FragColor.rgb = mix(bg_pix, o_pix, balpha * fg_frag.a);
+    } else {
+      gl_FragColor.rgb = mix(bg_pix, o_pix, balpha);
+    }
   } else if (fg_alpha > 0.0) {
     gl_FragColor.rgb = fg_pix;
   } else {
@@ -122,5 +127,9 @@ void main( void )
   }
 
   // Premultiplication
-  gl_FragColor.rgb *= gl_FragColor.a;
+  if (matte) {
+    gl_FragColor.rgb = gl_FragColor.rgb * gl_FragColor.a;
+  } else {
+    gl_FragColor.rgb *= gl_FragColor.a;
+  }
 }
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -535,6 +535,7 @@
   <item>"SHADER_HSLBlendGPU.bhue"			"Hue"    </item>
   <item>"SHADER_HSLBlendGPU.bsat"			"Saturation"    </item>
   <item>"SHADER_HSLBlendGPU.blum"			"Luminosity"    </item>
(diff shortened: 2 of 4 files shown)


RodneyBaker:
Thanks @konero 

@justburner initiated the HSLBlend Shader in Opentoonz so he may have thoughts to share regarding original intent etc.
Having said that...

- The code looks good and works as advertised.  Nice work.
- The change is optional which meets the criticial requirements outlined in #2746 (a key component of navigating your concern regarding original intent)
- I can't speak to the question of if the original implementation was a bug.  The closest thing to a bug I can find is that in some implementations of similar FX the order of the terms HSL are named HLS.  For that please refer to Image_Adjust > HLS Add Ino and HLS Adjust Ino.  Both have an alpha option via drop down menu.

As the go to expert on all things FX, @shun-iwasawa is the likely person to answer any deeper questions.

RodneyBaker:
@konero Can you elaborate on the 'strange result' you observed?

RodneyBaker:
There are some issues with this code as well as the original shader.
@justburner is currently looking into this.

stuff/library/shaders/programs/HSLBlendGPU.frag
```diff
@@ -114,13 +115,21 @@ void main( void )
...
+    }
   } else if (fg_alpha > 0.0) {
     gl_FragColor.rgb = fg_pix;
   } else {
     gl_FragColor.rgb = bg_pix;
   }
 
   // Premultiplication
-  gl_FragColor.rgb *= gl_FragColor.a;
+  if (matte) {
```
justburner, line 130: This code is completely redundant.

stuff/library/shaders/programs/HSLBlendGPU.frag
```diff
@@ -114,13 +115,21 @@ void main( void )
...
+    gl_FragColor.rgb = gl_FragColor.rgb * gl_FragColor.a;
+  } else {
+    gl_FragColor.rgb *= gl_FragColor.a;
+  }
```
justburner, line 134: This code is completely redundant and doesn't change the implementation in any way.

`gl_FragColor.rgb *= gl_FragColor.a;` is the shorthand of `gl_FragColor.rgb = gl_FragColor.rgb * gl_FragColor.a;`

stuff/library/shaders/programs/HSLBlendGPU.frag
```diff
@@ -114,13 +115,21 @@ void main( void )
   // Perform blending
   if (fg_alpha > 0.0 && bg_alpha > 0.0) {
     vec3 o_pix = SetLumSat(bhue ? fg_pix : bg_pix, bsat ? fg_pix : bg_pix, blum ? fg_pix : bg_pix);
-    gl_FragColor.rgb = mix(bg_pix, o_pix, balpha);
+    if (matte) {
+      gl_FragColor.rgb = mix(bg_pix, o_pix, balpha * fg_frag.a);
```
justburner, line 119: fg_alpha defined earlier in [#L106](https://github.com/opentoonz/opentoonz/blob/b1bffa51fefb348c2c9984798ec9010187786d0a/stuff/library/shaders/programs/HSLBlendGPU.frag#L106) holds the same value of `balpha * fg_frag.a` already.

justburner:
I've been doing some testing together with Rodney.

Despite the best efforts to solve the main issue, I don't think adding a new option is the answer, it's much better to fix the original implementation since we found issues with semi-transparency.

<hr>

Testing scenario, two soft lines intersecting each other, red on base(background) and green on blend(foreground):
LinesTest_Original
With all 3 boxes of HSL ticked it's expected no changes in the image, so the results should look the same
Note: The pixels are a little janky because of the odd zoom in the viewer... this snapshot was taken with preview off.

<hr>

HSLBlend messes up blending on alpha between 0% and 100%:
LinesTest_HSLOld
It was designed and tested with solid images containing either opaque or fully transparent pixels so that's why I've never stumbled upon this problem.

<hr>

This patch sadly still have issues.
Matte on and Clipping Masking off, background image seems to bleed out between crossing alphas:
LinesTest_Konero
... but with Clipping Mask active it works as expected:
LinesTest_ClipMask

<hr>

Today I've rewritten the mixing process to take account semi-transparency and here's the result:
LinesTest_Expected
Sadly this is not an one liner and since it doesn't need Matte option I think it's best to fill a new PR.

konero:
Thanks for looking into it!

PR #4718, opened 2023-01-23
Panel FixWidth mode + Level Palette option to not auto adjust width

justburner:
Both a bit of code refactoring and a new feature.

Level Palette panel now can have a fixed width while docked, if the panel is too narrow it will avoid mysterious stretching each time OT opens or main-window gets resized *(in one of my rooms i have a very narrow Level Palette for quick color select and the excessive stretching after reopening the program was driving me nuts, hence this PR)*.

This option can be found on Level Palette hamburger menu called "Auto Adjust Panel Width", it's checked by default and that's the original behavior *(Check state is opposite of this feature)*.

To use this new feature just uncheck this option and the main-window won't be able to auto resize the panel, note that manually resizing with the cursor still works.
Anyway having the panel width fixed has a side-effect! Shrinking the main-window will be limited by the width of the panel since... well... it won't be able to resize it automatically.

Lastly the refactoring did bring a few changes in behavior, panels now keep track of the minimum width properties of the widgets instead of overriding them to 0px, that means certain panels can't be resized to width of 0 anymore.

### DIFF
--- a/toonz/sources/include/toonzqt/paletteviewer.h
+++ b/toonz/sources/include/toonzqt/paletteviewer.h
@@ -140,6 +140,9 @@ protected:
   QAction *m_visibleGizmoAction;
   QAction *m_visibleNameAction;
 
+  bool m_variableWidth;
+  QAction *m_variableWidthAct;
+
 protected:
   void createTabBar();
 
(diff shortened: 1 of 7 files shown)


RodneyBaker:
Would like to recommend this new option be on by default and as such would like to hear from anyone who prefers otherwise.

We can also consider the state of various defaults as we move toward locking down defaults for release of v1.7.
It is anticipated that some Preference defaults may change but each change will require discussion.  Determining whether to make this behavior default now will be one less thing to consider.

IAW #2746, having this update on by default meets the intent that users can turn the option on/off.  :)

RodneyBaker:
Posting intent to merge as this PR works as advertised.

RodneyBaker:
Thank you @justburner for the fix/option.
Very much appreciated.
LGTM

PR #4723, opened 2023-01-24
Update xsheetviewer.cpp

flurick:
A strange calculation for scroll zooming that locks vertical xsheet zooming minimum 50% of the GUI zoom slider, would removing it break anything?

### DIFF
--- a/toonz/sources/toonz/xsheetviewer.cpp
+++ b/toonz/sources/toonz/xsheetviewer.cpp
@@ -1899,9 +1899,6 @@ void XsheetViewer::zoomToFramesPerPage(int frames) {
 
 //----------------------------------------------------------------
 int XsheetViewer::getFrameZoomFactor() const {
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I'm reviewing the PR that implmented the zoom in the xsheet (whereas before that was only available in the Timeline).
REF:  #3944

shun-iwasawa:
The `m_frameZoomFactor` variable was originally prepared for the zoom function in the Timeline.
The same variable was used to implement the zoom function in the Xsheet, but the same value range as in the timeline (20.0 - 100.0) caused the cells to become too small (from my perspective) at the smallest zoom factor.
Therefore, the value range was adjusted so that the cells would remain large enough to be manipulated even at the smallest size (50.0 - 100).

image

flurick:
The issue I was after here was that Ctrl+scrolling would only move the GUI zoom slider (top right) between 50%-100%

PR #4724, opened 2023-01-24
Tablet fixes

flurick:
I should really keep fixes in different branches, this is a couple of Qt tweaks and changes ctr+alt for setting the brush size to be less slippery (but still not implemented for mypaint brushes, which would be nice)


RodneyBaker:
Just a note to say that when first testing this PR I lost pen sensitivity.
Regained that after a computer restart.

I'm not entirely sure the loss of sensitivity was due to this PR and haven't dug deeper as of yet.
I did want to provide that initial feedback however just in case that is being reported elsewhere also.

Hopefully other tablet users (and linux users in particular) are testing and will provide their feedback as well.

flurick:
Will separate into smaller more testable PRs

PR #4731, opened 2023-02-01
Adjust Panel Width checkbox fix

justburner:
Auto Adjust Panel Width checkbox is always shown checked on startup regardless of internal state, this patch will fix it.


RodneyBaker:
Thank you @justburner for the fix.
LGTM

Issue #4738, opened 2023-02-06
Typedef declaration of size_t in tgc::hash class triggers error in gcc13 env.


grillo-delmal:
### Description

This bug was reported to me by the Fedora build system when rebuilding packages for the next version of the distro[0], which uses  gcc13.

Honestly, I don't know if they added new flags that makes builds more strict or if it's because of the gcc version change, but this is the error message that it shows when building v1.6.0 and the latest commit onf F38.

```
/opt/toonz/toonz/sources/include/tcg/hash.h:37:50: error: declaration of 'typedef typename tcg::list<tcg::hash<K, T, Hash_functor>::BucketNode>::size_t tcg::hash<K, T, Hash_functor>::size_t' changes meaning of 'size_t' [-fpermissive]
   37 |   typedef typename tcg::list<BucketNode>::size_t size_t;
      |                                                  ^~~~~~
/opt/toonz/toonz/sources/include/tcg/hash.h:28:5: note: used here to mean 'typedef long unsigned int size_t'
   28 |     size_t m_prev;
      |     ^~~~~~
```

I searched about this message and found the following Stack Overflow thread [1] that explains why it happens.

I already prepared a simple patch for Fedora for it though, so I'll upstream it soon.

[0] https://koji.fedoraproject.org/koji/taskinfo?taskID=96384858
[1] https://stackoverflow.com/questions/12187549/typedef-changes-meaning

### Steps to Reproduce

1. Boot up a Fedora 36+
2. `sudo dnf install fedpkg`
2. `fedpkg clone opentoonz`
3. `cd opentoonz`
4. `git checkout f37`
5. `fedpkg --release f38 mockbuild`

### Expected Behavior

I expect it to build without problems, as it did on Fedora 37

### Screenshots, Video & Crash Logs

```
[  6%] Building CXX object tnzcore/CMakeFiles/tnzcore.dir/__/common/trop/tdespeckle.cpp.o
[...]

PR #4739, opened 2023-02-07
Clarify size_t origin for `tgc::hash::BucketNode`

grillo-delmal:
## The problem

The problem with #4738 is the following:

`toonz/sources/include/tcg/hash.h:37`:
```cpp
template <typename K, typename T, typename Hash_functor = size_t (*)(const K &)>
class hash {
public:
  typedef K key_type;
  typedef T value_type;
  typedef Hash_functor hash_type;

  struct BucketNode {
    K m_key;
    T m_val;
    size_t m_next;
    size_t m_prev;

    BucketNode(const K &key, const T &val)
        : m_key(key), m_val(val), m_next(-1), m_prev(-1) {}
    BucketNode(const std::pair<K, T> &pair)
        : m_key(pair.first), m_val(pair.second), m_next(-1), m_prev(-1) {}
    ~BucketNode() {}
  };

  typedef typename tcg::list<BucketNode>::size_t size_t;
  typedef typename tcg::list<BucketNode>::iterator iterator;
  typedef typename tcg::list<BucketNode>::const_iterator const_iterator;
  ...
  ```

In the line that says `typedef typename tcg::list<BucketNode>::size_t size_t;` it's declaring that in the scope of this class, the `size_t` keyword should be treated as the `tcg::list<BucketNode>::size_t` type. This line was declared after the `BucketNode` struct was declared, and since this struct uses `size_t` type parameters, the compiler doesn't know if the programmer's intention was to declare them as the global `::size_t` or `typename tcg::list<BucketNode>::size_t`, which is in the scope of the class but declared later in the code. [0]

The truth is that this is not actually a problem, since the `tcg::list<BucketNode>::size_t` is declared in the following block:

`toonz/sources/include/tcg/list.h:177`:
```cpp
template <typename T>
class list_base {
public:
  typedef T value_type;
  typedef ::size_t size_t;
```

So then we can conclude that `tcg::list<BucketNode>::size_t` is actually the same as `::size_t`. But since I have to implement this cos the compiler it's complaining anyway, I thought it would be a good idea to upstream this change :).

[0] https://stackoverflow.com/questions/12187549/typedef-changes-meaning

## The solution

There are several ways to fix this, the easiest one being the following

```cpp
template <typename K, typename T, typename Hash_functor = size_t (*)(const K &)>
class hash {
public:
  typedef K key_type;
  typedef T value_type;
  typedef Hash_functor hash_type;

  struct BucketNode {
    K m_key;
    T m_val;
>   ::size_t m_next;
>   ::size_t m_prev;

    BucketNode(const K &key, const T &val)
        : m_key(key), m_val(val), m_next(-1), m_prev(-1) {}
    BucketNode(const std::pair<K, T> &pair)
        : m_key(pair.first), m_val(pair.second), m_next(-1), m_prev(-1) {}
    ~BucketNode() {}
  };

  typedef typename tcg::list<BucketNode>::size_t size_t;
  typedef typename tcg::list<BucketNode>::iterator iterator;
  typedef typename tcg::list<BucketNode>::const_iterator const_iterator;
  ...
```

But I think that it might be a better Idea to use the `tcg::list<BucketNode>::size_t` as the type of those variables in case a refactor to this part is needed in the future. This is what I implemented in the end:

```cpp
template <typename K, typename T, typename Hash_functor = size_t (*)(const K &)>
class hash {
public:
  typedef K key_type;
  typedef T value_type;
  typedef Hash_functor hash_type;

> struct BucketNode;
> typedef typename tcg::list<BucketNode>::size_t size_t;
> typedef typename tcg::list<BucketNode>::iterator iterator;
> typedef typename tcg::list<BucketNode>::const_iterator const_iterator;
>
  struct BucketNode {
    K m_key;
    T m_val;
    size_t m_next;
    size_t m_prev;

    BucketNode(const K &key, const T &val)
        : m_key(key), m_val(val), m_next(-1), m_prev(-1) {}
    BucketNode(const std::pair<K, T> &pair)
        : m_key(pair.first), m_val(pair.second), m_next(-1), m_prev(-1) {}
    ~BucketNode() {}
  };
  ...
```

I know that this is not a problem in the current backend considering that it's being built on ubuntu with an older version of gcc that ignores this case, but I still think that it will be better to clarify this part and improve consistency 😄 

Let me know if this implementation is ok, if you don't like it or if there is another way it should get implemented 😉

### DIFF
--- a/toonz/sources/include/tcg/hash.h
+++ b/toonz/sources/include/tcg/hash.h
@@ -21,6 +21,11 @@ public:
   typedef T value_type;
   typedef Hash_functor hash_type;
 
+  struct BucketNode;
+  typedef typename tcg::list<BucketNode>::size_t size_t;
+  typedef typename tcg::list<BucketNode>::iterator iterator;
+  typedef typename tcg::list<BucketNode>::const_iterator const_iterator;
+
   struct BucketNode {
     K m_key;
     T m_val;
@@ -33,11 +38,6 @@ public:
         : m_key(pair.first), m_val(pair.second), m_next(-1), m_prev(-1) {}
     ~BucketNode() {}
   };
-
-  typedef typename tcg::list<BucketNode>::size_t size_t;
-
-  typedef typename tcg::list<BucketNode>::iterator iterator;
-  typedef typename tcg::list<BucketNode>::const_iterator const_iterator;
 
 private:
   std::vector<size_t> m_bucketsIdx;


RodneyBaker:
@grillo-delmal Thank you for the submitted change and especially for your thoughts related to it and interest in upstreaming.  I don't see any specific concerns with your implementation but certainly invite others to discuss further.

As this is a small change and no adverse effect is apparent... will post intent to merge.
Linux users in particular are encouraged to join in.

RodneyBaker:
Thank you @grillo-delmal for your help improving Opentoonz.
Much appreciated.
LGTM

PR #4742, opened 2023-02-08
Export Camera Track feature

shun-iwasawa:
This PR will add a new feature `Export Camera Track` which will render a camera track image showing how the camera frame moves on the specified raster level column.
This image is used in the following cases:
- After determining the camera work with OpenToonz, use it as a reference for finalizing animation or background.
- After determining the camera work with OpenToonz, use it as a reference for compositing in AfterEffects.

image

### Usage
- You need to manually register the command to the menubar or use `Reset to Default Rooms` command.
- Exportable image formats are JPG, BMP, PNG, or TIF. (As it uses Qt's image I/O)

### Parameters
- **Target Column** : Specifies a column where the camera moves. The camera's trajectory will be drawn as a relative motion in the Target Column. Only the column containing raster or toonz raster level can be the target.
- **Background** : Specifies an opacity of the target column image to be placed as the background.
- **Line Color** : Specifies line color of the camera track.
- **Camera Rectangles**
    - **Draw On Keyframes** : When ON, the rectangles will be drawn on the keyframes of the Target Column and current camera and their parent objects.
    - **Draw On Navigation Tags** : When ON, the rectangles will be drawn on the frames where the navigation tags are set.
    - **Specify Frames Manually** : Type frame numbers (separated by comma) for drawing the rectangles.
- Track Lines
    - **Top Left**, **Top Right**, **Center**, **Bottom Left**, and **Bottom Right** : Specify the corners (or center) of the camera frame where you want to draw the track lines.
    - **Graduation Marks Interval** : Specifies interval of the graduation marks on the line.
- Frame Numbers
    - **Camera Rect Corner** : Specifies the corner of the camera frame where the frame number is to be drawn.
    - **Draw Numbers On Track Line** : When ON, numbers will also be drawn at the graduation marks of the track line.
    - **Font Family** and **Font Size** : Specify the font properties for the frame numbers.

### DIFF
--- a/stuff/profiles/layouts/rooms/Default/menubar_template.xml
+++ b/stuff/profiles/layouts/rooms/Default/menubar_template.xml
@@ -33,6 +33,7 @@
             <command>MI_StopMotionExportImageSequence</command>
             <command>MI_ExportTvpJson</command>
             <command>MI_ExportXsheetPDF</command>
+            <command>MI_ExportCameraTrack</command>
         </menu>
         <separator/>
         <command>MI_PrintXsheet</command>
--- a/stuff/profiles/layouts/rooms/StudioGhibli/room4_menubar.xml
+++ b/stuff/profiles/layouts/rooms/StudioGhibli/room4_menubar.xml
@@ -11,6 +11,7 @@
     <separator/>
     <command>MI_PrintXsheet</command>
     <command>MI_Export</command>
+    <command>MI_ExportCameraTrack</command>
     <separator/>
     <command>MI_Quit</command>
   </menu>
--- a/toonz/sources/toonz/CMakeLists.txt
+++ b/toonz/sources/toonz/CMakeLists.txt
@@ -129,6 +129,7 @@ set(MOC_HEADERS
     custompanelmanager.h
     custompaneleditorpopup.h
     convertfolderpopup.h
+    exportcameratrackpopup.h
 )
 
 if(PLATFORM EQUAL 64)
@@ -382,6 +383,7 @@ set(SOURCES
     custompanelmanager.cpp
     custompaneleditorpopup.cpp
     convertfolderpopup.cpp
+    exportcameratrackpopup.cpp
 # Tracker file
     dummyprocessor.cpp
     metnum.cpp
--- a/toonz/sources/toonz/exportcameratrackpopup.cpp
+++ b/toonz/sources/toonz/exportcameratrackpopup.cpp
@@ -1,0 +1,986 @@
+
+
+#include "exportcameratrackpopup.h"
+
+// Tnz6 includes
+#include "tapp.h"
+#include "mainwindow.h"
+#include "filebrowser.h"
+#include "menubarcommandids.h"
+//// TnzQt includes
+#include "toonzqt/colorfield.h"
+#include "toonzqt/filefield.h"
+#include "toonzqt/doublefield.h"
+//// TnzLib includes
+#include "toonz/txsheet.h"
+#include "toonz/tcamera.h"
+#include "toonz/txshlevel.h"
+#include "toonz/txshsimplelevel.h"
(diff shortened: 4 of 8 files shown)


shun-iwasawa:
@RodneyBaker 
Thank you for verfying the PR. I confirmed that the OT crashed when selecting a column containing toonz raster level as the target. Now I think I have fixed it.

RodneyBaker:
Thanks for the tip @shun-iwasawa 
I was indeed using toonz raster format and when that was failing... forgot to try again with raster.
I confirm the featuure is working with raster and anticipate it will work fine with toonz raster once your update completes building.

RodneyBaker:
This is likely outside the scope of this PR but a thought while testing...

I do understand that this feature is designed to work per scene with each scene most likely consisting of it's own singular background.  It would be useful to have a Target Frame field to allow for a column holding multiple background images.
This might be different versions of the same BG (rough, cleaned up, colored, etc.) or it might be BGs from different shots that are simply contained within the same camera movement.

A specific case I am thinking of is where one BG might fade away to another (as with walls disappearing or special effects where the BG is more dynamic than static).

Assuming I am simply not missing the capability already being there...
A workaround at present would be to simply expose the desired BG for output via Export Camera Track on the first frame... which I assume is always targeted by the current PR.

If possible perhaps to keep the code simpler it might be a frame offset... offsetting the target frame from the default of frame 1.

image

In the above I'm imagining a BG (overlay) with wall that is obscuring part of the shot were a character will move exposed on frames 1 -10.  On frame 11 a new BG without the overlaid wall is then exposed.  FX might then dissolve the one into the other or it might just cut directly to the new BG.

<End random thought>

The updated PR is working great with toonz raster levels.
Thanks for the quick fix on that!

RodneyBaker:
In hindsight I think another workaround to the multiple BGs in the same level is just to distribute the frames of the level across multiple xsheet columns (which is how I set up the image in my last post).  

This because it isn't the first frame that is targeted but I believe the first 'occupied' frame with an exposure in it.
So in this case the BG2 frame of a series of sequential BGs would just be exposed in the appropriate frame of that specific column.

RodneyBaker:
There are three things I have been looking at and just want to make note of...

1.  Refreshing - The preview window of the Camera Track updates on some specific updates... such changing target, toggling on/off a setting such as use of keyframes etc.   It would be nice for the dialogue to either have a 'Refresh' button or somehow know when the camera has changed in some way such as changing of keyframes.

2.  Line thickness - The user can change the color of camera rectangle but apparently not line thickness.  In some cases the lines appear quite thin and it would be ideal to be able to adjust that thickness.

3.  Track Lines - I note some oddness with track lines where sometimes multiple track lines appear when only one should. I believe I was able to get the lines to duplicate by closing the dialogue and then reopening.  Once duplicated the extra lines persist.  In one of my previously posted images we can see some jagged lines apparently going to frames that do not have camera rectangles displayed.  With rotation I was able to get nice arcing lines most of the time.  Nice!

Regarding the one aspect of the lines:
image

In the image I would think the Track lines would go from key 1 to key 2 to key 3.
It's not clear why there is a line from 1 to 3 (or is it 3 to 1?)
Perhaps this is just triangulation?

Some of the extra lines (apparently not related to above) might relate to the build and we can confirm more after release of the nightly.

Having said all this...
This feature is ready to be merged.
If anyone is testing and requires more time please make that desire known.

RodneyBaker:
Thank you @shun-iwasawa for the new Export Camera Track feature.
LGTM

RodneyBaker:
Here's a quick example that leverages the recently added subcamera capability to output camera selections larger than canvas; to supply Camera Track export with very large backgrounds:


https://user-images.githubusercontent.com/2576432/218298300-96545914-ab20-4ff8-9f42-89df328ef5ae.mp4

HorseOnHillCameraTrack

PR #4752, opened 2023-02-17
Stretch function curve keys with Alt+drag

shun-iwasawa:
stretch_curve
This PR enables stretching timings of the selected keys with Alt + dragging in the function curve panel.

### DIFF
--- a/toonz/sources/include/toonzqt/functionselection.h
+++ b/toonz/sources/include/toonzqt/functionselection.h
@@ -131,6 +131,7 @@ public:
   // overlapping the selection
   int getCommonSegmentType(bool inclusive = true);
 
+  QList<int> getSelectedKeyIndices(TDoubleParam *curve);
 signals:
   void selectionChanged();
 };
--- a/toonz/sources/toonzqt/functionpanel.cpp
+++ b/toonz/sources/toonzqt/functionpanel.cpp
@@ -1274,16 +1274,39 @@ void FunctionPanel::mousePressEvent(QMouseEvent *e) {
               getSelection()->deselectAllKeyframes();
             getSelection()->select(currentCurve, kIndex);
           }
-          // move selected point(s)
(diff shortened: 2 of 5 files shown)


gab3d:
@shun-iwasawa thank you!!
this very very good news!

Will stretching have its "center" on the opposite side of the key you drag onto, or is it hardcoded to be on the left side?

Also, here are some ideas for future expansion of the feature:
- the current time cursor could be used as the stretching center.
- initial direction of mouse movement could allow stretching on either horizontal or vertical directions (in case of vertical direction stretching, the mouse cursor height could be used as stretching center).

RodneyBaker:
> Will stretching have its "center" on the opposite side of the key you drag onto, or is it hardcoded to be on the left side?

The pivot/center is on the opposite side.

Additionally, in the code @shun-iwasawa has provided the following comments:
`          // stretch the selected keys IF;
          // 1) alt key is pressed, and
          // 2) more than two keys are selected, and
          // 3) clicked key is one of the end of the selection, and
          // 4) all keys between the both end of the selection are selected`

The clicked key (on one end or the other of a (continuous) sequence of selecteted keys) is stretched using the opposite end's key as the pivot for the stretch.  So we might select the right and stretch the selected keys and then select the left and stretch it again... with the right-most key staying in place.

gab3d:
@RodneyBaker thanks for checking it!
It's great to have the feature behaving as you described. 👍

RodneyBaker:
I'm not sure this will propagate as a nightly but we'll give it a shot.
The Mac build for the last PR did finally appear in the nightlies but not the Windows build.

Thank you @shun-iwasawa for the very useful addition to function curves.
LGTM

RodneyBaker:
Update to last:  The latest nightly release has properly posted and this feature is now included in Opentoonz.
Thanks @shun-iwasawa!

PR #4756, opened 2023-02-28
Customizable column color filters

shun-iwasawa:
This PR will enable to customize column color filters via the Scene Settings popup.
image

Please note that once the settings have altered from the default, such scene won't be opened from the previous versions of OT before this PR.

### DIFF
--- a/toonz/sources/include/toonz/sceneproperties.h
+++ b/toonz/sources/include/toonz/sceneproperties.h
@@ -58,6 +58,14 @@ public:
     }
   };
 
+  struct ColorFilter {
+    QString name;
+    TPixel32 color;
+    bool operator==(const ColorFilter &cf) {
+      return name == cf.name && color == cf.color;
+    }
+  };
+
 private:
   Guides m_hGuides, m_vGuides;
 
@@ -85,6 +93,9 @@ private:
 
   // Cell Mark colors and names
   QList<CellMark> m_cellMarks;
+
+  // Color Filter colors and names
+  QList<ColorFilter> m_colorFilters;
 
   bool m_columnColorFilterOnRender;
   TFilePath m_camCapSaveInPath;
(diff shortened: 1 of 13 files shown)


RodneyBaker:
@shun-iwasawa This may appear to be a small change but in my estimation it's huge.
You've anticipated a need that has been there since Opentoonz was first released and delivered it in an even better way that I anticipated.

A quick test suggests everything is working as advertised. 

Thanks also for the mention of the breaking change as users will definitely want to know that.

RodneyBaker:
@shun-iwasawa A question...

While we can set matte/alpha transparency in the custom color, from what I can tell this has no effect on the column filter.
That is reserved for the opacity slider only at this time, correct?

shun-iwasawa:
@RodneyBaker Thank you for reviewing the feature!

> While we can set matte/alpha transparency in the custom color, from what I can tell this has no effect on the column filter.
That is reserved for the opacity slider only at this time, correct?

That was correct, but there may be an advantage to including the alpha channel of the color filter in the column opacity calculation. I'll try to modify it a bit. Thanks for the idea!

shun-iwasawa:
The opacity of the displayed column is now the alpha channel of the column's color filter multiplied by the value of the column's opacity slider.

RodneyBaker:
That's got it.  Nice!  Thanks @shun-iwasawa 

The only thing I'm having difficulty with currently is setting colors to either fully white or fully black.
RGB values of 0,0,0 and 255,255,255 appear to me as potentially problematic.

In the case of white it seems that no matter then change to opacity or slider the resulting output is... completely white.
Perhaps that's as it should be but it seems odd.  Perhaps the opacity is being ignored at that value?

With black I'm finding that I cannot set complete black (0,0,0) which would make sense I suppose as well.
If changing just one value in any RGB channel (or say... using a value of 1,1,1 we can get the full spectrum from white (white replaces original image) to fully transparent (i.e. the image appearing as if there was no color tinting at all).  

If someone could test full white and full black and observe the results that will be appreciated.
(Note that use of pure white was also an issue prior to Shun's latest modification.  I can't recall testing pure black.)

RodneyBaker:
Here is an example of a frankenstein'd reverse color separation thingy:

All the images are black and white/grayscale with the color provided by Custom Column Color Filters (tm).

image

RodneyBaker:
Will press on with this PR and get it into the hands of Opentoonz users.

Thank you @shun-iwasawa for the customizable column color filters.
LGTM

RodneyBaker:
I meant to post a follow up to the (percieved) black and white colors issue I posted before.

That might have been related to the build I was testing.
I don't see any issues with black and white in the appveyor and nightly build.

RodneyBaker:
I did confirm that scenes created with adjusted column color will not open in earlier versions.
It will be good to consider this when troubleshooting.

customcolumnfilter animated

Project file:
[CustomColumnColor animated.zip](https://github.com/opentoonz/opentoonz/files/10871029/CustomColumnColor.animated.zip)

RodneyBaker:
This is to note the issue reporting column tinting is now resolved.
REF:  #4803

PR #4772, opened 2023-03-10
Fix notifyClick when mouse button is released rather than pressed

flurick:
To allow dragging scenes in browser views that are activated with a single click, previously this would cancel the drag action


RodneyBaker:
Thank you @flurick for the fix.
LGTM

PR #4773, opened 2023-03-11
Revert preprodcution board icon sizes to 80x60

flurick:
For larger thumbnails the quality should to be improved with some kind of smoothing


RodneyBaker:
This fix is obvious so will merge.
Thank you @flurick for the revert!

PR #4777, opened 2023-03-13
Paste cells command behavior option by Contrail Co.,Ltd.

shun-iwasawa:
In most cases in Japanese animation production, one xsheet column corresponds to one level. Therefore, there are few cases where the entire contents of a cell (i.e. level and drawing number) in one column are pasted into another column, and pasting of drawing numbers is more frequently required.

Therefore, this PR will add a preference item `Paste Cells Behaviour` to Xsheet category, with the options as follows;
- **Insert Paste Whole Data** (default option)  : The command will paste entire contents of copied cells.
- **Overwrite Paste Cell Numbers** : The command will execute `Paste Cell Numbers` instead.

When the latter option is selected, a command ~~`Paste Whole Cell Data`~~  `Paste Cell Content` will appear in the right click menu of xsheet, in `Paste Special` submenu. This will allow users to use the conventional paste behavior.

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -368,7 +368,7 @@ public:
   int getDeleteCommandBehaviour() const {
     return getIntValue(deleteCommandBehavior);
   }
-
+  int getPasteCellsBehavior() const { return getIntValue(pasteCellsBehavior); }
   bool isIgnoreAlphaonColumn1Enabled() const {
     return getBoolValue(ignoreAlphaonColumn1Enabled);
   }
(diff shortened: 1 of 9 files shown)


gab3d:
Would **Paste Cell Content** be an equivalent but shorter alternative to the **Paste Whole Cell Data** command name?

Also, what's the effect of this new preference option over assigned keyboard shortcuts?

shun-iwasawa:
@gab3d 

> what's the effect of this new preference option over assigned keyboard shortcuts?

With the latter option , using the shortcut key for the command `Paste Insert` (Ctrl+V by default) will act as `Paste Cell Numbers`. Assigning the frequent paste behavior to Ctrl+V is actually the main purpose of this modification.

RodneyBaker:
Thanks again to Contrail Co. ltd for the new options.
LGTM

PR #4785, opened 2023-03-18
Fix rendering loud output  + additional changes

manongjohn:
This PR contains 3 general changes ported over from T2D:

1. Add missing `break;` statements missing in logic executed only by Linux and macOS builds. Fixes #4779
2. Changed the max sample rate from 44100 to 48000 for Windows builds.
3. Use the appropriate copy overall sound functions based on sample type.

I forget why I did the last 2 changes in T2D's original 32bit implementation but figured it might be good for OT as well.

Issue #4779, opened 2023-03-13
Audio in Fast Render to MP4 Extremely Boosted


deathchimera:
### Description

Resulting files from Fast Render to MP4 have their audio maxed out, probably even more than what's allowed normally. This is on linux, though it's surprising I haven't heard anyone else address it, considering how long it's been like this in the nightlies.

### Steps to Reproduce

Fast render to MP4 with an audio level on linux.

### Expected Behavior

The volume to be set at the percent chosen on the level.

### Screenshots, Video & Crash Logs

(SERIOUS VOLUME WARNING)

https://user-images.githubusercontent.com/77942602/224632787-73a3a28f-2b1c-41a7-b642-21376d669fc2.mp4

### OpenToonz Version

Nightly (Specify below)

### OpenToonz Version Information

Latest, built Mar 13 2023

### Operating System

Linux

### GPU

Intel

### Graphics Tablet

Wacom

RodneyBaker:
Interesting.

I don't use  the Fast Render to MP4 feature often and when I have I have rarely done so with audio included.
As such I was fully expecting to run into the same issue.
But, in my tests I don't see anything amiss with audio using that means of output.

I'll tag as a Linux issue until we find out more about the problem.
There are several things you can do to assist in the problem solving.
For instance if you can Export an example scene that has the problem that will make sure we are seeing the exact thing you are.  If you have access to another computer I'd also suggest test on that to rule out something localized to your current system.  

I do at this point suspect this might be Linux related but hopefully we can get other on PC, Linux and Mac to confirm.

I think I'll try to incorporate use of Fast render to MP4 for my weekly drawing sessions with my niece.
Here a result adding audio to yesterday's session that exported fine (but on WIndows):

https://user-images.githubusercontent.com/2576432/224706806-31adbfd5-bf74-4906-af21-2c99e9bb5dd3.mp4

deathchimera:
This scene only has the audio used in the example given, though it still distorts on Fast Render. Getting others to test would seem necessary.

[OT_AudioIssue.zip](https://github.com/opentoonz/opentoonz/files/10960957/OT_AudioIssue.zip)

RodneyBaker:
Thanks for the sample file.
Yes, that renders fine for me via Fast Render to MP4.

I'm beginning to suspect this might be a Linux only problem but...
There are some other things you can check before we can conclude it is something in the source code itself.
[...]

Issue #4638, opened 2022-12-08
Audio tracks muted by default, and have no sound in preview


Lantau42:
### Description

I use Opentoonz for Mac Monterey 12.5, and every time I upload audio files, they appear muted by default. Each time I try to unmute them (using the 'click to play the soundtrack back' button) to play sound in the animation preview, there's no sound whatsoever. Even stranger, that track's mute button gets muted on its own again after a few seconds, even when I'm not pressing anything.
But for some bizarre reason, the sound does come out in the rendered video. It's just in the preview that's the problem, which makes it extremely difficult to make proper lip-sync animations. 

I've tested this using the official Opentoonz 1.6 build (from August, I think), and the nightly build from December 5 2022. 
I have ffmpeg installed (with the files in the ffmpeg folder all opened in terminal before opening Opentoonz), and I've tested this with both mp3 and wav files (both 8 bit and 16 bit). Even with multiple different audio files, the bug's still there. 

I've used Opentoonz for the previous version of Mac's OS (Mac Big Sur), and the audio tracks did work back then. So I don't know if it's just a problem with my OS. 
 
Frankly, I'm tearing my hair out, and I'm begging for a solution. Thank you.

### Screenshots, Video & Crash Logs

This is what the audio file looks by default. Note that the 'sound' icon in the toolbar is on, and I've tried turning it on-and-off several times to see if there's any difference (no change)
<img width="1026" alt="Screenshot 2022-12-08 at 2 27 24 AM" src="https://user-images.githubusercontent.com/120071891/206341478-bdd33e96-6244-4c63-9044-0c2cbc33eb4e.png">

How it looks when I try to unmute the audio track. There is no sound, even when scrubbing through. 
<img width="808" alt="Screenshot 2022-12-08 at 2 17 05 AM" src="https://user-images.githubusercontent.com/120071891/206341087-3d93c555-014c-4df3-bac2-38b40277bb6f.png">

Volume is completely at 100% the whole time. 
<img width="869" alt="Screenshot 2022-12-08 at 2 40 38 AM" src="https://user-images.githubusercontent.com/120071891/206342967-14dd4355-4f57-40ef-a6fc-2d0fa711457a.png">

How it looks seconds later. This can happen while I'm still playing the audio, and haven't even clicked anything. 
[...]

PR #4787, opened 2023-03-20
Fix rendering loud output + additional changes (Part 2)

justburner:
There were additional missing `breaks` not caught in #4785

Hopefully this should fix the Linux audio shenanigans in #4779 for good.


RodneyBaker:
Thank you @justburner for yet more fixes to the problems with audio on Linux systems.
(Addition of 3 "breaks" only in the change so will proceed with merger)
LGTM

RodneyBaker:
These fixes might also address some other reports related to audio.  For example:  #4638

PR #4792, opened 2023-03-26
Improve mypaint chip icon

manongjohn:
This was a requested T2D port to improve Mypaint chip icon in the Style Editor and Palette viewer.

Before:

After:

Note: The color triangle in the upper left corner only shows up in List View since there is no other color indicator in that chip.


RodneyBaker:
Satisfaction confirmed.

Thanks @manongjohn for the port!

PR #4797, opened 2023-03-30
Modify Convert TZP in Folder and register TZP icon 

shun-iwasawa:
Modified the `Convert TZP in Folder` command so that the progress bar popup shows the relative path from the target folder instead of only showing the level name.
image

Also registered the icon files for TZP and TZU (both are level file formats used in old Toonz versions). Missing registration had caused the `Run out of contiguous memory` error popup in File Browser.


RodneyBaker:

PR #4798, opened 2023-03-30
Fix rendering loud output + additional changes (Part 3) 

manongjohn:
I'm porting these additional changes I am making to T2D to fix 2 issues that also are in OT.

- Multiple audio levels not always mixing properly, caused by additional changes I added in #4785 that were actually unecessary. Sorry about that
- Rendered 32bit Fixed audio was garbled/noisy.  Corrected ffmpeg arguments to encode using `f32le` or `s32le` depending on the final audio format.


RodneyBaker:
Thank you @manongjohn for the (part 3) fix!
LGTM

PR #4799, opened 2023-03-31
Fix column transparency to work on rendering

### LINKED ISSUE #4791: New Custom Column Color breaks Preview/Render Transparency
### Description

@shun-iwasawa 
Since the addition of #4756 adjusting the Opaqueness slider no longer works with Preview and final rendering without also adding a color tint.  If a color is selected the opaqueness/transparency will kick in and work.

### Steps to Reproduce

1.  Draw with several colors on a toonz raster level (results should be the same on other levels as wel) 
2.  Adjust the column opaqueness slider (without Preview on) and note that transparency works
3.   Make sure the option to render column transparency and tinting is turned on in Scene Settings so that preview and rendering should be effected by the columns transparency (and color tinting).
4. Turn Preview on and note that column transparency no longer works (Final rendering should also not produce desired result)
5.  Select any color tinting from the dropdown and note that now the transparency will work.

shun-iwasawa:
This PR fixes #4791 

Thank you @RodneyBaker for reporting the problem and @flurick for pointing out the problematic line.

### DIFF
--- a/toonz/sources/toonzlib/scenefx.cpp
+++ b/toonz/sources/toonzlib/scenefx.cpp
@@ -976,7 +976,7 @@ PlacedFx FxBuilder::makePF(TLevelColumnFx *lcfx) {
          (column->isCamstandVisible() && column->getOpacity() != 255))) {
       TPixel32 colorScale = m_scene->getProperties()->getColorFilterColor(
           column->getColorFilterId());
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thank you @shun-iwasawa for the timely fix!
LGTM

PR #4802, opened 2023-03-31
Fix Style Editor HSV rounding

### LINKED ISSUE #4651: Colours are shifting their values when switching styles
### Description

When changing the values of a colour style in the style editor for HSV, after switching to another style and back again the digit has decreased by 1 or 2.

If I entered a lower number, then it would decrease by 2. If I entered a higher number it would decrease by 1.

### Steps to Reproduce

1. Make two styles
2. Edit one of the HSV fields
3. Switch to another style
4. Switch back to the style that was modified
5. Notice the value has changed

### Expected Behavior

For the values to remain when switching styles.

https://user-images.githubusercontent.com/55353161/207823330-4f0aa443-e5f4-4d77-ae2b-71bdf2002c12.mp4

### OpenToonz Version

Nightly (Specify below)

Latest

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Wacom

shun-iwasawa:
This PR will fix #4651

Now the converting between HSV and RGB becomes more precise so that the rounding error won't occurs.

### DIFF
--- a/toonz/sources/toonzqt/styleeditor.cpp
+++ b/toonz/sources/toonzqt/styleeditor.cpp
@@ -168,15 +168,16 @@ void ColorModel::rgb2hsv() {
   QColor converter(m_channels[0], m_channels[1], m_channels[2]);
   m_channels[4] =
       std::max(converter.hue(), 0);  // hue() ritorna -1 per colori acromatici
-  m_channels[5] = converter.saturation() * 100 / 255;
-  m_channels[6] = converter.value() * 100 / 255;
+  m_channels[5] = (int)std::round(converter.saturationF() * 100.);
+  m_channels[6] = (int)std::round(converter.valueF() * 100.);
 }
 
 //-----------------------------------------------------------------------------
 
 void ColorModel::hsv2rgb() {
-  QColor converter = QColor::fromHsv(m_channels[4], m_channels[5] * 255 / 100,
-                                     m_channels[6] * 255 / 100);
+  QColor converter =
+      QColor::fromHsvF((qreal)m_channels[4] / 360., (qreal)m_channels[5] / 100.,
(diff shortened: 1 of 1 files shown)


KaseyFarron:
Thanks for tackling the bug! I tested and unfortunately I can still reproduce the issue at least with Hue slider, it happens if I single click somewhere, click and drag, use the buttons or type a number manually. I tested using AppVeyor. I haven't reproduced it with Saturation or Value sliders so far.

https://user-images.githubusercontent.com/55353161/229275757-d07e4894-6c6c-4374-a39d-f2124470a49f.mp4

RodneyBaker:
Will hold off merging this for additional testing purposes and to verify results reported by @KaseyFarron

shun-iwasawa:
@KaseyFarron
Regarding the Hue slider, I think it would be difficult to fix the problem completely.

The color information of a style is stored in 8 bits (255 steps) in RGB channels.
Since the Hue slider is finer (360 steps), it is possible that the same RGB is calculated from two HSV that differ only by 1 Hue.
When the style is switched, HSV is reproduced based on the RGB channel values.
At this time, they will converge to one of the Hues. I think this is why only Hue sometimes shifts by one.

This problem is separate from the accuracy issues of the Saturation and the Lightness sliders.
If the Hue sliders were to snap to discrete positions, the problem of shifting Hue would be solved, but the movement of the slider would be unnatural.

KaseyFarron:
Oh I see, thank you for the explanation! I won't hold up merging this.

RodneyBaker:
Thanks for the feedback @KaseyFarron  Much appreciated.
And...

Thank you @shun-iwasawa for the fix to HSV color rounding!
LGTM

PR #4813, opened 2023-04-07
Fix File Browser to open network path

shun-iwasawa:

- When typing a network path in the `Folder` input field of the File Browser and the path is not a project folder, the error message `The input folder path was invalid.` appears and the folder cannot be displayed.

If there is no node in the file browser tree that matches the input path, now the file browser will search for network folders and will add a new node.


RodneyBaker:
LGTM

PR #4814, opened 2023-04-07
Fix raster caching not to set the modified flag

shun-iwasawa:

1. Do pencil testing. Capture many drawings (say, 50 frames) and save as frames in one raster level
3. Set the preference option `Loading > Raster Level Caching Behavior` to All Icons & Images
4. Save the scene & Revert the same scene 
5. Try to capture another drawing as a frame of the same level
6. OT freezes

This problem was caused by the `m_modified` flag being set on all frames when caching the raster level at load time.
As a result, when saving the additional frames captured, all other frames were being overwritten and saved.

### DIFF
--- a/toonz/sources/include/toonz/imagemanager.h
+++ b/toonz/sources/include/toonz/imagemanager.h
@@ -314,10 +314,7 @@ private:
   ImageBuilder(const ImageBuilder &);
   const ImageBuilder &operator=(const ImageBuilder &);
 
-  void setImageCachedAndModified() {
-    m_cached   = true;
-    m_modified = true;
-  }
+  void setImageCached() { m_cached = true; }
 };
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


RodneyBaker:
@shun-iwasawa I am not sure why I cannot reproduce the freeze you experience.
I get a short half second freeze some times but then I can capture a new frame.
Perhaps this is due to the low rez nature of my webcam?  Not sure.

I do think I saw the issue with overwriting of image files which I don't see via this PR.
This PR does appears to address that issue.
As such will post intent to merge but am still testing to see if I can get OT to freeze via the steps you have provided.

RodneyBaker:
Pressing forward with this...

LGTM

shun-iwasawa:
@RodneyBaker Thank you for testing the PR. Probably the problem had been more prominent in our studio as we store all images in file server, instead of in the local machine.

PR #4820, opened 2023-04-10
Fix Camera Capture to flip image after BG reduction

shun-iwasawa:
This PR fixes the problem regarding the Camera Capture as follows:

- When using the `Upside Down` option, BG reduction feature does not work properly as the white reference BG image is not flipped.


RodneyBaker:
Another self evident code fix.
Will merge as soon as the last PR gets into a nightly.

Thank you @shun-iwasawa for the fix,
LGTM

PR #4821, opened 2023-04-10
Fix MacOS color model picking with stylus

### LINKED ISSUE #4819: Unable to get style from color model using pen tab
### Description

When using a pen tab and clicking on a color model, there is no response and the style cannot be retrieved.
This problem does not occur when using a mouse or trackpad.
When the pen tablet is reconnected to the computer, it may temporarily respond normally, but after a certain period of inactivity, it will stop responding.
Operations on other windows are normal as far as I can tell.
(This problem also occurs in v1.6)

### Steps to Reproduce

1. Prepare the painted animation.
2. Load the appropriate color model.
3. Use a pen tab to try to get the style from the color model.

### Expected Behavior

In most cases, you will not be able to get the styles.
In a few rare cases, it may respond correctly, but after a short time, it will stop responding.
(Operations on other windows will respond normally even when the style cannot be retrieved from the color model.)

https://user-images.githubusercontent.com/126930364/230818602-bdf04ce2-3e5b-4174-8b18-65cdfe987d84.mov

This is a screen recording with QuickTimePlayer.
When QuickTimePlayer detects a click, a black circle is displayed around the cursor.

Any hints would be greatly appreciated.

### OpenToonz Version

Nightly (Specify below)

Mar 20, 2023

### Operating System

macOS

### GPU

Intel

### Graphics Tablet

Wacom

shun-iwasawa:
This PR fixes #4819 

Accepting the tablet event in `ImageViewer::tabletEvent` prevents to propagate the event to the parent widget `ColorModelViewer` and it seemed to be a probable cause of the problem.
Thank you @Kumotate for reporting the issue!

### DIFF
--- a/toonz/sources/toonz/imageviewer.cpp
+++ b/toonz/sources/toonz/imageviewer.cpp
@@ -1496,8 +1496,6 @@ void ImageViewer::tabletEvent(QTabletEvent *e) {
   } else if (e->type() == QTabletEvent::TabletRelease) {
     m_stylusUsed = false;
   }
-
-  e->accept();
 }
 
 //------------------------------------------------------------------


RodneyBaker:
Fix is self evident so...

Kumotate:
Thank you for the quick correction response!

PR #4823, opened 2023-04-11
Camera Capture: Save default image adjust parameters

shun-iwasawa:
**This is to be merged after releasing the next version.**

This PR will add `Save current image adjust parameter as default` button in the Camera Capture popup.

image

Clicking the button will save the all parameters of the Image adjust group box in `$TOONZLIBRARY/camera image adjust/[host name]_[camera name]_[resolution].ini`. If the white reference BG is captured, it will also be saved in  `$TOONZLIBRARY/camera image adjust/[host name]_[camera name]_[resolution].jpg` .

The saved default parameters will be loaded on the next launch of OpenToonz so that you can continue to capture with consistent brightness as long as you have the same lighting environment.

**PLEASE NOTE**
- The default image adjust parameters will be shared among users
- The default parameter will NOT be loaded on switching the current camera or the resolution for now. I will listen to user feedback and consider this in the future.


RodneyBaker:
I would love to be able to save out multiple 'defaults' and then choose from those as needed (perhaps from a dropdown menu).
But... we can swap out defaults easily by manipulating the file(s) saved to the directory in the library so... all good.

This additional does exactly what it is designed to do.
Thank you @shun-iwasawa for the new save default option.
LGTM

Added:
For those curious, the contents of the file saved stores the default parameters in the following fashion:

image

PR #4824, opened 2023-04-11
Fix deleting output node

### LINKED ISSUE #4822: The delete context menu option for Deleting an Output doesn't work
### Description

You can right click on an output on the FX Schematic to show a menu. It only has 1 item, Delete. But clicking it does nothing! You can select the node and press the delete key on your keyboard, but the command doesn't work.

### Steps to Reproduce

1. Open a scene
2. Open the FX schematic
3. Press the New Output button to create a new output (or press Alt+O or right click on the schematic and add one from there)
4. Right click on the new output and choose Delete from the context menu.
5. Nothing changes!

However - If you create a number of outputs, then try to delete them from their context menu. Notice nothing happens.
But if you then right click on another output and choose to Activate it, the outputs that you tried to delete now disappear!!
So it seems that the node might ACTUALLY be deleted and it's just a redraw issue.

### Expected Behavior

After choosing the delete option the node should disappear immediately.

ScreenRecording

### OpenToonz Version

Nightly (Specify below)

built Apr 9 2023 10.03:23

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

None

shun-iwasawa:
This PR will fix #4822 

- Deleting output node is done in `DeleteFxOrColumnUndo::redo()` .
- Updating schematic UI after delete is done only when `m_isLastInRedoBlock==true` .
- `m_isLastInRedoBlock` will be properly set when registering the undo. However, since the initialization was missing, calling `redo()` before registering the undo had been unpredictable.

I set the initial value to `m_isLastInRedoBlock` to fix the problem.

### DIFF
--- a/toonz/sources/include/tundo.h
+++ b/toonz/sources/include/tundo.h
@@ -29,9 +29,9 @@
 class DVAPI TUndo {
 public:
   // To be called in the last of the block when undo
-  bool m_isLastInBlock;
+  bool m_isLastInBlock = true;
   // To be called in the last of the block when redo
-  bool m_isLastInRedoBlock;
+  bool m_isLastInRedoBlock = true;
 
 public:
   TUndo() {}


RodneyBaker:
Nice fix.
Thank you @shun-iwasawa 
LGTM

DarrenTAnims:
Thanks @shun-iwasawa. Much appreciated. And amazingly fast!
I think these small, quality of life fixes, help everyone.
Cheers

DarrenTAnims:
I've just tried this in a nightly build and it works great for a single Output. But if you highlight more than one and try to delete them with the context menu, that doesn't work. Do you have time to take a look at that too @shun-iwasawa ?

PR #4833, opened 2023-04-17
Fix deleting output node (part 2)

shun-iwasawa:
This PR modifies `Delete` command in the context menu of Output node in Fx Schematic, so that the command will delete all selected (highlighted) nodes instead of deleting only the clicked node.

This PR will resolve the problem mentioned in [this comment](https://github.com/opentoonz/opentoonz/pull/4824#issuecomment-1510337106).


DarrenTAnims:
Amazing @shun-iwasawa.
I really appreciate this fix and completed so quickly too.
😍

PR #4838, opened 2023-04-18
Update Bug Report for 1.7

konero:
Merge after 1.7 stable has been released.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -1,6 +1,5 @@
 name: Bug report
 description: File a bug report to help us improve
-labels: 'bug'
 body:
   - type: markdown
     attributes:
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@konero Would you consider modifying the entry on line three so that the submitted report doesn't automatically get labeled as a bug?   While it might make the reporter feel better to see it labelled as a bug many reports aren't actually bugs and need to be altered to remove that label.  I'd personally rather promote a report to status as bug rather than demote one labeled as a bug to not being one.  

I'm very open to discussion on this but find myself removing the bug label from reports often which I think has the potential to suggest to the reporter their concern has been downgraded.  I'd rather start with something unconfirmed and unlabeled as a bug and then have it labeled as a bug upon validation.  

I personally don't have an issue with downgrading reports from being labeled as a bug to another label... that's why I've recently added the 'troubleshooting' label... but I know some reporters can and do take issue with such a change.

What should the label of a potential bug report be?  Not sure.

RodneyBaker:
Time to lean into v1.7 bug reporting.  :)
Thanks @konero for the report formatting update.

PR #4843, opened 2023-04-21
Unify preview and camstand visibility toggles option by Contrail Co., Ltd.

shun-iwasawa:
**This is to be merged after releasing the next version.**

This PR will add a preferences option `Unify Preview and Camstand Visibility Toggles` which will unify the camstand and the preview visibility to get a simple UI.

image

When this option is turned on, the camera stand and the preview visibility toggles in all columns will be automatically adjusted to the same states.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -2368,6 +2368,7 @@ XsheetViewer {
   qproperty-XsheetPreviewButtonOnImage: url('../Default/imgs/white/x_prev_eye_on.svg');
   qproperty-XsheetPreviewButtonBgOffColor: transparent;
   qproperty-XsheetPreviewButtonOffImage: url('../Default/imgs/white/x_prev_eye_off.svg');
+  qproperty-XsheetUnifiedButtonTranspImage: url('../Default/imgs/white/x_prev_eye_transp.svg');
   qproperty-XsheetCamstandButtonBgOnColor: #eb906b;
   qproperty-XsheetCamstandButtonOnImage: url('../Default/imgs/white/x_table_view_on.svg');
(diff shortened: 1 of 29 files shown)


gab3d:
But... so with this option active there will be no way of having a column hidden in render and visible in the viewer?
That sounds too limiting to me. (thinking of controls for rigs or matte layers used with FX)... 🤔 

Why don't, instead, sync-toggle both options while clicking on any of them, BUT, retaining the possibility of doing (say) CTRL+click for purposely toggle only one of the options, when needed?
I mean, there isn't any space savings either with this proposal, as it is now.

shun-iwasawa:
As you pointed, the ability to operate the two visibility toggles independently is useful in many aspects, such as displaying only the reference image on the camstand for positioning, or hiding only the other columns when manipulating one column, etc.

However, in the pencil testing workflow that Contrail is doing with OpenToonz, there are few occasions where such ability is necessary, and for them it is more important to keep the UI simple. It is not to free up screen space, but rather to avoid distraction by unnecessary features.

I think leaving the possibility to switch between one side by modifier key + click is a good idea, but for now they would prefer the simpler look... And some other users who use OpenToonz for pencil testing might prefer to use this option.

DarrenTAnims:
I think having a single button is great for most users anyway. I like it.
But for power users you should be able to right click on it and choose one of the 2 options: Show in render/show while editing. And then show half an eye (top & bottom or left/right). A left click on the button should then turn them both on. I think this would work for most users.

gab3d:
@shun-iwasawa  Thanks for the explanation, I understand perfectly.

RodneyBaker:
Thanks for the additional option @shun-iwasawa (and Contrail Co., Ltd!)
LGTM

PR #4844, opened 2023-04-21
Fix pen not working on Schematic Viewer

flurick:


RodneyBaker:
@flurick Sorry to trouble you here but can you provide additional information about the problem this PR addresses.
When I test in Schematic swatch viewer my pen works fine and it appears to work in other locations related to Schematic viewer as well.

Note that your change in the code looks fine but I always try to reproduce the initial problem also and cannot currently do so.  As such I'm convinced I'm doing something differently than I should.

My thought is that I'm just not understanding the original problem.

manongjohn:
Just realized this is a port from T2D. The original problem it fixed only impacted macOS/Linux builds.

RodneyBaker:
> Just realized this is a port from T2D. The original problem it fixed only impacted macOS/Linux builds.

Ah, that certainly would explain why I haven't been seeing the problem. Thanks!

shun-iwasawa:
I tested on my MacOS environment and confirmed that the PR fixes the problem.
Thank you @manongjohn for the fix and @flurick for porting it!

PR #4848, opened 2023-04-24
Fix update sound after editing cells

### LINKED ISSUE #4846: Sound Not Update Correctly on OT1.7RC
### Description

When you cut audio frames it still reproduces the audio as it was at the beginning although in the timeline noiselines have been updated.

### Steps to Reproduce

1. Cut audio from the timeline
2. 

### Expected Behavior

_No response_

_No response_

### OpenToonz Version

Other (Specify below)

1.7RC

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Huion

shun-iwasawa:
This PR fixes #4846 
Added emitting signal for updating sounds when sound cells are cut.

The same modification was applied to other edit cells commands as well, such as pasting, deleting, and inserting cells.

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -216,11 +216,14 @@ class PasteCellsUndo final : public TUndo {
   TCellSelection *m_newSelection;
   TCellData *m_data;
   std::vector<bool> m_areOldColumnsEmpty;
+  bool m_containsSoundColumn;
 
 public:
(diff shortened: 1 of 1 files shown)


RodneyBaker:
As a follow up to this PR there was a short discussion about concerns with the fix of this PR and larger audio files.
I believe specifically that it caused a slow down in processing due to the emitted signal.

I don't have specific details and haven't seen any slow down in my testing thus far but perhaps @manongjohn and @melieconiek will see this and share their concerns.

PR #2656, opened 2019-07-01
(WIP) Krita-Like Assistant Tool

blackwarthog:
Issue #1349

- [x] Change the pen color on the cells for the Assistant Level to Black so the level name can be read against the light level color, similar to the Note Level.
- [x] Add a thumbnail for the Assistant Level so it's not just an empty box in xsheet/timeline.
- [x] Issue when a frame is created in the Assistant Level, Undo/Redo of the new frame doesn't work.
- [x] DEL removes whole frame instead of selected assistant.
- [x] Delete assistant via configured shortcut, instead of hardcoded DEL.
- [x] Vector strokes does not creates when magnetism < 1
- [x] Too thin assistants guidelines
- [x] Unexpected thin tail at the end of each stroke
- [x] Lag drawing on Toonz Raster level
- [x] Remove code blocking seq 0 frames from loading
- [x] Fix ALT key not resetting on release in tools
- [ ] Raster brushes don't stay on ellipse when drawing too fast
- [x] Ellipse Assistant causes crash on macOS
- [x] Duplicate Drawing causes crash on Assistant Levels
- [x] Standard raster brush does not use max size when using mouse or pressure turned off
- [ ] Smooth lag is worse in this PR than nightly.
- [x] Flickering while drawing. Make the line between cursor and drawn line toggleable?
- [x] Plastic Tool's toolbar options not shown for Sub-Xsheet columns
- [ ] Hardness settings producing bad lines for Toonz Raster brush
- [ ] Autosave causes crash on Raster levels when drawing with pen

Issue #1349, opened 2017-07-29
Krita-Like Assistant Tool For Opentoonz


Orphanlast:
It'd be nice to be able to draw your backgrounds IN Opentoonz. As it currently stands, the program seems to lack the necessary tools to effectively draw in perspective. So It'd be nice to have something like the Assistant Tool in Krita. Especially their Vanishing point and Fisheye perspective guides.

Examples of what I'm talking about would be here:
https://www.youtube.com/watch?v=uByCdWXCyhs

https://www.youtube.com/watch?v=Zc7xatDNc4A

turtletooth:
The bounty for this has been created and is here: 
https://www.bountysource.com/issues/47719781-krita-like-assistant-tool-for-opentoonz

blurymind:
The perspective grid tool in krita is especially useful for drawing backgrounds:
https://docs.krita.org/Perspective_Grid_Tool

Here is a link to the documentation of the different assistants in krita:
https://docs.krita.org/Painting_With_Assistants

video demo:
https://www.youtube.com/watch?v=OhEv2pw3EuI

article
https://shivaramanaiyer.wordpress.com/2013/05/22/grid-assistant-awesome-perspective-assistant/

![assist_grid](https://user-images.githubusercontent.com/6495061/28769317-da08c3b4-75d2-11e7-9062-23898bd7775d.png)
![500px-krita-tutorial3-iii-2](https://user-images.githubusercontent.com/6495061/28769329-e89d92c4-75d2-11e7-8fd1-912bc2b7ce25.png)
![krita-tutorial1-iv 2](https://user-images.githubusercontent.com/6495061/28769341-ffffde4a-75d2-11e7-9ee5-2c231b4bca43.png)

Orphanlast:
@blurymind thank you for the elaboration and clarnfication. And thanks for the additional two donations. Soon, next week, I'll be making another donation

Orphanlast:
Thank you Richie.

jpturcotte:
Sorry for the bump, but the bounty is now at $600 of its $700 goal. Let's push it all the way through!

mason-bially:
I saw this on BountySource, is it open for anyone, or is it meant for developers already in the community? Specifically @blackwarthog who set the bounty.

Orphanlast:
I'm kind of fond of having Black warthog work on it, primarily because he's involved with the Morevna Project. 

[...]

PR #4852, opened 2023-04-25
Krita-Like Assistant Tool Alternative

blackwarthog:
This is a bit another way to implement the issue #1349
Previous was in PR #2656

Here assistants functionality is more separated from other OpenToonz code. So i don't need to solve the old bugs, and you may more easily to review the code.

In this PR assistants is implemented only for FullColorBrushTool.

I'll implement ToonzBrush and ToonzRasterBrush later, when this PR will be merged. I'm afraid that while i work on it the OpenToonz repo will flied forward and i'll need to solve a lot of merge conflicts.


RodneyBaker:
I'm not quite sure I know how to use this tool...yet.
Interesting to see this restart of a much requested toolset.

image

RodneyBaker:
Ah...  the github action builds fail to bring everything in.
Using the appveyor builds allows the addition of the dropdown menu (pretty important to have!)

image

blackwarthog:
> I'm not quite sure I know how to use this tool...yet.

At your screenshots is the test multi-touch modifier. It can be enabled by the painting with Alt key pressed.
I can disable it if it confuse somebody

> Ah... the github action builds fail to bring everything in.
> Using the appveyor builds allows the addition of the dropdown menu (pretty important to have!)

I don't understand. Need i to do something with it?

RodneyBaker:
We'd need to advise @shun-iwasawa on what updates are needed for the github action builds.
It's not clear to me what is lacking that keeps the menu items from not appearing.

BUT...

That isn't the real problem in my estimation.
I'm not getting any assistants for drawing with the appveyor from this PR other than the one with use of Alt key (that you describe).

RodneyBaker:
While we can't draw with vector yet with the Assistants.. we can convert what we draw to vector.  :)

image

RodneyBaker:
It's been so long since I looked at that previous feature I had completely forgotten.
It probably doesn't help that there have been a few other perspective tools that made their way into the Tahoma2D fork... and they had/have different requirements.

Once the basic premise is known it is relatively simple to expand upon that.
I mainly had forgotten that we needed to select an empty cell to create the initial Assistants column.

There are two main concerns that I have at this point:
1)  Ease of discovery
2)  Some Undo issues  (Undoing doesn't appear to undo what has been drawn but other things)

This second issue is not necessarily an issue of the Assistants Tool but perhaps just one of many Undo related issues.

Ease of Discovery... not sure how to overcome that short of doing what you are here... sharing a video of the process.

Having said all of the above... I will say this... I've been having fun.

image

beeheemooth:
> . It makes me sad that just few of people was interested.

In fact, most users did not understand how to use it. Using an additional level just for the assistant turned out to be quite cumbersome and time-consuming in practice (despite the fact that the assistant functionality is wonderful). The "Perspective Grid Tool" by manongjohn for Tahoma2D has more modest functionality, but it is very practical, as it can be turned on with one click, and perspective grid settings can be saved to presets and quickly switch between them.

RodneyBaker:
The additional level is indeed hard to discover... which is why I forgot it.
I do find some functionality in that extra level however that the perspective tool in Tahoma2D lacks.
Specifically, the ability to store a different assistant setup per frame.

Now, granted... I haven't fully explored any of the perspective tools... Jeremy's previous one was very simple (and although very hard to discover was quite useful).  John's is very easy to discover (and as you've stated @beeheemooth much more modest in features/functionality.  Not as customizable and versatile either.  For instance, it's really nice to be able to add Line Assistants, Ellipse Assistants and Vanishing Point Assistants all into the same space.  Powerful stuff that.

This is early in development yet and so it will be good to take a Macro view of how this feature set and others can coexist together.   It probably is a hint that porting should occur earlier rather than later to prevent conflicting efforts.

Of note, I am not observing any of the previous issues noted with slow down in drawing with this current Assistant.
Others more familiar with those issues may have additional feedback there.

blackwarthog:
> It would be nice if when the Assistant Tool icon is first pressed it automatically created the Assistants level

I'm not sure that a user will expect any changes in the project just while switching between tools.

In current version when you select at empty column and try to create a new assistant, then the new level will be created.

Of course we may discover some methods to make assistants usage more simple. For example i may place the "create assistants level" button at the top bar in case when the tool selected but assistants level does not exists.

But will be good to do it after merging of this PR.

RodneyBaker:
I just noticed that Assistants are saved as TZM level files.  Nice.
As a community we should be able to create and share a wide variety of Assistants which can be loaded and used at any time.

Also, I was looking for a good way to copy/paste Assistants and the Duplicate key (shortcut D) does a fine job.

RodneyBaker:
Layering of grids/assistants works great too:

image

blackwarthog:
Originally the feature was planned to be expanded to multi-brushes painting tool. Like mirrored drawing and something like this:

multipaint

So assistants look a bit more difficult then may be expected.

RodneyBaker:
When you get some time you should look at John (@manongjohn 's) current implementation of the Symmetry Tool in Tahoma2D.
What you just posted has some definite simularities and it seems reasonable that at some point that tool will be ported over to Opentoonz.  The more code these various features can share the better in my estimation.

REF:  https://github.com/tahoma2d/tahoma2d/pull/1131

manongjohn:
Try not to laugh to hard at me and/or the code if you look at T2D's Symmetry code. Just happy it works! 😄

I can say that it was very paintful integrating it into existing drawing tools without wanting to somehow completely rewrite or consolidate how the different drawing levels and drawing modes handled the various ways it tracked input.

If this PR moves you 1 step closer to that unification without compromising performance that's great.  I think that was one of the issues with the original implementation of the Assistant tool.

openanim:
It's good to see that there's still a chance that this feature will be in OT.

Couldn't re-test it in depth yet, but my quick feeling about it is that compared to the original version.

You can't use the other 2 levels Pli and tlv on this PR anymore but if I understood correctly, it's planned to bring back these features after the eventual merger of this PR?

- Another point that bothers me is that the preview that was present when the guides were set up on the assistant level has been abandoned. Now the guides are no longer animated, we have to move them manually... Couldn't this be an option? 

I find that previewing during the setting up is much better and fast to check notably the direction of our guides. For me the preview is one of the strengths of this feature.

- It has also become more difficult to select points compared to the original version of the assistant guides even if it is slight, the difference is noticeable in practice.

For me the fact that there is a dedicated level is also a CORE feature in the approach to this feature. It's less noticeable on a project with few guides to use but on a project with many guides it's much faster and clearer to enable and disable this way imo 

Some quick suggestions:

Since the assistant level is linked to the assistant tool. Why not automatically activate the tool when you are on the assistant level? And go back to the previous tool when you are on a different level?  This would save one click

Why not keep by default the last selected/used guide until it is changed?  There too we would gain click. For example to place several vanishing points in a row.

blackwarthog:
@openanim thanks for your feedback!

> You can't use the other 2 levels Pli and tlv on this PR anymore but if I understood correctly, it's planned to bring back these features after the eventual merger of this PR?

Yes, it will be re-implemented in a next PR. I think i'll create a new "work-in-process" PR for it soon.

> Another point that bothers me is that the preview that was present when the guides were set up on the assistant level has been abandoned. Now the guides are no longer animated, we have to move them manually... Couldn't this be an option?

Actually i thought that it has no matter. I'll restore the old behavior.

> It has also become more difficult to select points compared to the original version of the assistant guides even if it is slight, the difference is noticeable in practice.

I'll look

> Since the assistant level is linked to the assistant tool. Why not automatically activate the tool when you are on the assistant level? And go back to the previous tool when you are on a different level? This would save one click

Good idea. But i'm afraid that source code become more tangled... I'll think about it.

> Why not keep by default the last selected/used guide until it is changed? There too we would gain click. For example to place several vanishing points in a row.

I'm not sure. But i'll think about it too.

blackwarthog:
heh... merging conflicts fixed

openanim:
Thank you for your thoughts on these things. Here is a video where we can see the difference in behavior between the current PR and the original about preview during the set up and the issue while selecting the points, if that helps.


https://user-images.githubusercontent.com/29790512/235504092-155e1f37-7172-44ef-ac19-9982c1ec778b.mp4

blackwarthog:
I've fixed bug with points selection and restored the old guidelines behavior in the assistants editor

JParks123:
Since 1.6 I have been working with OpenToonz  developers researching and creating all new features also bringing in the majority of T2D ports via JB and MJ I have had much to do with a lot of the features PR's outside of Shuns stuff and I must say its very nice to see someone  actually doing a PR so please keep it up I like what has been done so far and may be willing to help u with this PR if needed?

openanim:
@blackwarthog 

Sorry if I confused you in my 1st post about re-enabling the preview. It seems we were not talking about the same thing. What I was referring to was the movement of the guides when editing the guides on the assistant level which I showed in my previous video. (And that you fixed  in your last commits, thanks).

I was not referring to the line that appears when drawing and that would be good to disable indeed again. Sorry if I gave you some extra work because I misled you with this preview term which also seems to be linked to another feature but indeed this "track effect" is not very pleasant to see when drawing (see the attached video). That was something already mentioned in the original version. Sorry for the misunderstanding.

track issue_and_pan need rede to work

Otherwise there is a problem present also in the previous versions (you can see in the above video also), it is the no response of the multitouch with the PR having the assistant levels on opentoonz. We have to try twice before the multitouch move works.  I dont know if we can see it correctly in the attached video but here steps to reproduce : 

With gesture touch enabled in preferences : 

- Draw something on a level
-  use finger to pan ( no works)
- redo twice to make it works until next draw and repeat it everytime

blackwarthog:
No problem, it is by bad. I actually did not want to restore this behavior. I've fixed this problem already.

RodneyBaker:
@blackwarthog 
I would like to target merger of this PR for the beginning of June.
There are some conflicts that need to be resolved but they shouldn't be overly difficult.
It is important this foundational PR is in place so further modifications can be made more easily.

artisteacher:
The first time I tried this today on macOS, it was super buggy for some reason. Getting the assistant visible while the brush tool was active was hit or miss. When it did display, I couldn't get figure out how to get my strokes to follow the guide - they wouldn't follow no matter what. Now I did make a toonz raster level after the assistant level and before making the raster level, so i don't know if that had an impact.

After a restart (and only having raster levels in the scene), it finally seems to be working as expected. 

Things I really like:
- being able to have different guides on a frame by frame basis
- being able to keyframe the guides - or not - completely independent of the other levels
- the ellipse behavior

Areas for improvement:
- when switching to the assistant level, the guides need to be visible, no matter what tool is active. We can use the animate tool to tween the guide - which is fantastic - but we can't actually see what we are keyframing - less helpful.
- vanishing points could be more intuitive - I really prefer them in T2D
- it's hard to tell different guides apart - in T2D you can give the guides different colors/opacities which is super helpful
- when a tweened drawing level is the parent of the guide level, they almost seem synced but it's inconsistent & seems buggy
- it could be helpful to have a duplicate option in the Xsheet cell context menu. The equivalent of "Duplicate Drawing" but for the guides. So that you could use easily use the pre-existing guide as a starting point and then adjust it.
- similarly, it might be nice to be able to copy/paste guides

shun-iwasawa:
@blackwarthog 
Can you please resolve the conflict? Or can you squash the commits so that I can post another PR appending another commit on top of your work for resolving the conflict?

blackwarthog:
Оh... I'm sorry... I saw the green check mark next to PR and thought it was okay. I had need to get kids to ready for school exams and rarely checked the comments. One moment

blackwarthog:
@artisteacher thanks for your feedback i'll try to implement your suggestions

blackwarthog:
conflicts are solved

RodneyBaker:
@shun-iwasawa This is a huge addition but one I'm prepared to move forward with especially given it is the first and major part of the Assistants Tool.  If you have any concerns I'd love to hear them.  

Otherwise, posting intent to merge.

RodneyBaker:
Thanks @blackwarthog for your patience and...
Thank you for adding this inaugural Assistants Tool to Opentoonz.

Looking foward to future developments that take the Assistants even farther!

RodneyBaker:
The windoes nightly has failed to build but I don't think that is related to this PR.
The build environment might need to be refreshed.

cc:  @shun-iwasawa

RodneyBaker:
After a slight delay due to nightly build issues...

The inaugural appearance of the new Assistants Tool is available in the latest nightly release(s).

Thank you @blackwarthog for your hard work (and patience) getting this feature into Opentoonz.
Much appreciated!

RodneyBaker:
@blackwarthog 
Now that Opentoonz has Assistants...  :)

Something that will need to be looked at early on is how users can experience delay/disconnect of drawing when using other level types (toonz raster) when those levels are present along with Assistant levels.  

To examine that delay (or loss of focus) try the following:
- Draw an image with Assistants Tool
- Clean up that drawing to create a toonz raster level (or create a new toonz raster level)
- Note how Opentoonz seems to have a memory leak or loss of focus for the Brush Tool and  user will not be able to interact properly with Opentoonz.

konero:
I did some quick testing:

Firstly, even without creating or activating the assistant tool, over time the UI becomes sluggish, this was a problem with the last attempt as well. Things like changing levels in the XSheet, using the color wheel in the Style Editor and changing styles. Not as bad as before, but it's still a problem. Note it's worse when the Brush tool is active and far worse on Windows.

Before:

https://github.com/opentoonz/opentoonz/assets/19820721/f487890d-0521-4efc-8147-ebe953dcec62

PR:

https://github.com/opentoonz/opentoonz/assets/19820721/a00ceb41-2bf5-4703-abb4-d0f62c2b46ab


I also noticed this extra "size" slider in Tool Options when Toonz Raster levels are selected with the Brush tool active which has a range of -3 to 3:

TRaster_ExtraSizeSlider

As for feedback on the feature itself:
- I don't think this needs to be its own level on disk, I fail to see the purpose of that. Instead why not bake it into scene data? Perhaps you are done with the ruler and delete the level from the scene, but you still need to delete the file on disk.
- It's cumbersome to use due to switching objects in the XSheet, instead the user experience would be improved greatly by breaking it out into its own widget like a layer stack panel with buttons that let you hide, add or remove an assistant. Clicking a layer in this panel automatically activates the Assistant tool and the adjustment handles and holding the Assistant tool shortcut key can do the same, right now it does nothing.
- When deactivating 'Assistant' checkbox in the Tool Options the assistant guidelines should be hidden, that's what I would have expected from a user point of view.
- The "twizzler"🐀 shape drawn when you hold alt I fail to see the purpose of and seems rather out of scope, I would suggest removing it and instead making it a vector brush for the effect, unless there was a good reason for it? It's a waste of an important modifier key in my view.
- The handles are nice, however they are very difficult to find when there is a lot of detail on the viewer, it would be easier to use if the handles were a filled blue circle with the current outline for example.
- When an ellipse ruler is added when you draw a stoke you need to move the cursor by some distance before you begin to see the line appear, however if you lift the mouse button the line will appear right away.

@shun-iwasawa I feel bad for suggesting but perhaps this should be reverted for now if the sluggishness can be replicated and it's not just from my environment (I tried Windows/MSVC and Ubuntu/GCC with raw Qt 5.15.2 but it is far worse on Windows).

RodneyBaker:
I've updated #4957 to reflect that the problems I was having with extreme lag have gone away with an update of wacom driver and restart of system (although note that a previous restart did not resolve the problem).   

I am currently guessing tat a bad install might be a key factor in some of what we are seeing, lag, missing UI elements... all suggest.  Not sure how this accounts for the local builds having similar issues however.


@Bracket-H By all means run any and all test you feel the need to run.  You've got me curious.

RodneyBaker:
I raretly use the Timeline myself but sometimes it's just there so I use it.  I prefer the vertical xsheet.  ;)

Initially, I couldn't get Assistants to work either as it isn't entirely intuitive.
Then suddenly I did something right and it all started to work.
A key to understanding is perhaps that Assistants are a type of level.
From there I'd say the next best thing to know is that in order to create an Assistant gizmo we need to use the dropdown menu to select one.
Currently the main thing that might keep OT users from discovering Assistants is that it only works (for now) with raster levels.
As OT defaults to toonz raster levels that can make it appear that Assistants don't work... and they don't... except with raster levels.

Now, having said all of this I'm not sure what (if any) difference there is with Linux and Mac versus Windows with regard to the Assistants...

I'm confident more information will be forthcoming as well as improvements to the feature set.
I don't think there is a big rush to get information out as it is still quite early in the tools development.

toonz/sources/tnztools/fullcolorbrushtool.cpp
```diff
@@ -173,6 +171,7 @@ void FullColorBrushTool::onCanvasSizeChanged() {
 //---------------------------------------------------------------------------------------------------
 
 void FullColorBrushTool::onColorStyleChanged() {
+  getApplication()->getCurrentTool()->notifyToolOptionsBoxChanged();
```
shun-iwasawa, line 174: This addition causes re-building the tool options bar every time you change the color of style. It seems to be one of the  causes of sluggish response.

shun-iwasawa:
I'll try to refactor the source so that the Assistance feature computation would not affect the other part of the software.

toonz/sources/tnztools/tooloptions.cpp
```diff
@@ -2883,6 +2849,19 @@ void ToolOptions::hideEvent(QShowEvent *) {
 
 //-----------------------------------------------------------------------------
 
+void ToolOptions::onToolOptionsBoxChanged() {
+  TTool *tool = TTool::getApplication()->getCurrentTool()->getTool();
+  std::map<TTool *, ToolOptionsBox *>::iterator it = m_panels.find(tool);
+  if (it != m_panels.end()) {
+    ToolOptionsBox *panel = it->second;
+    m_panels.erase(it);
+    layout()->removeWidget(panel);
```
shun-iwasawa, line 2858: @blackwarthog 
Why did you have to delete and recreate the tool options each time you changed the style? This appears to be a time consuming operation.
Is it not enough to hide unnecessary options with `filterControls()` ?

blackwarthog: Sorry for late answer. Yes, possibly it was not good idea... I've made this code before mypaint brushes was implemented for toonz raster level (it was implemented in parallel)

blackwarthog:
Oh... Sorry... I haven't guessed to check comments for the closed PR.
Seems most of them is not actual already.
Do still remain any troubles with this PR?

RodneyBaker:
@blackwarthog I think #4963 took care of the main issues.

Issue #4688, opened 2023-01-01
Resizing cursor stays when the mouse pointer passover


noelcurray:
### Description

the cursor doesn't change to original state if your mouse moves over to a panel. 

### Steps to Reproduce

1. Open Opentoonz
2.  Create a project
3. Try to move your mouse over a panel like color pallete
4. then move your mouse cursor to the canvas
5. the issue shows

### Expected Behavior

I expect that the mouse cursor changes according to the task that it should be doing. Like if you want to adjust the size of the panel, it will show the resizing cursor and change back to the original mouse cursor when you are not trying to adjust the panel. 

### Screenshots, Video & Crash Logs

https://user-images.githubusercontent.com/83384364/210168456-6951d7b4-8e3b-44b4-a0fb-67f39def8d0c.mp4

### OpenToonz Version

1.6

### Operating System

Arch Linux

### GPU

AMD

### Graphics Tablet

Wacom Cintiq 16

flurick:
Looks like the canvas Viewer doesn't get focused by pen events

flurick:
Could [possibly just be this]( 
https://github.com/opentoonz/opentoonz/compare/master...flurick:opentoonz:linux-restore-cursor) (similar to some other tablet input issues)

nukeCrayon:
I also notice that this just happen on vertical side. I test to adjust the horizontal panel and the issue doesn't show. 

one of my workaround is hit the shortcut key <S> and <B> and the cursor go back to original arrow.

andeon:
I belive it's an issue with Linux event mouseMove or tabletMove refresh. Unfortunately, this bug is very annoying, to avoid this I try to use a detached window like a toolbar or locator.

https://github.com/opentoonz/opentoonz/issues/2569

nukeCrayon:
> I belive it's an issue with Linux event mouseMove or tabletMove refresh. Unfortunately, this bug is very annoying, to avoid this I try to use a detached window like a toolbar or locator.
> 
> #2569

it's been 4 years and we still have the same issue.

noelcurray:
this is fixed now. thank you.

Issue #2569, opened 2019-03-17
Tablet pen input cursor


andeon:
I originally opened here: https://github.com/morevnaproject/opentoonz/issues/19 but I found more details about this problem and also here is the correct place to report this.

### Issue Summary
When you move / focus the cursor out of the drawing area with the tablet, it changes the cursor depending on where the cursor is pointing. It returns the correct if you point to the "locator" window,  use the mouse or the touchpad, sometimes press the shift or ctrl keys, and if have a window detached like toolbar and move cursor in front of this.

And I noted that the right click of the pen does not work as reported here https://github.com/opentoonz/opentoonz/issues/1790

I tested the version appimage 1.1.3 (https://sourceforge.net/projects/opentoonz-morevna-edition/files/stable/1.1.3/) and it really has no problem with cursor tablets, I did not find this version here to test, but I know they use the same master code, then. I'm not sure now which release this broke.

It's not just a visual cursor problem, I recorded a video, tools like "**Magnet**", "**Tape**" lose functions and generally work only if you click the mouse or the touch pad, then the tablet works another time with these tools, well until you move again the cursor out of the drawing area.

### Screenshot or Video Reference

Click to watch the video:

[![](http://img.youtube.com/vi/FApN4vil0QY/0.jpg)](http://www.youtube.com/watch?v=FApN4vil0QY "")
http://www.youtube.com/watch?v=FApN4vil0QY 
### Steps to Reproduce

With the graphics tablet just open the program and start painting, move the cursor out the drawing viewer. Back to drawing viewer to paint or switch to another tool, look that the cursor does not back to the correct state.

ps: Krita developers have frequent problems with MacOs versions, they may be related and you can help each other. Since you are using similar tools for development.

### System Information
* **OpenToonz Version**: 1.3, 1.2.0.3
* **Operating System**: Linux 64 bits (Kde Neon)
* **CPU**: Intel core i7-7500U
* **Memory**: 8GB
* **Graphics Card**: Nvidia
* **Graphics Tablet**: Wacom Intuos Draw CTL490DW, Genius MousePen 8x6 (UC-Logic Tablet WP8060U)

artisteacher:
This looks like it's related to known issues: #1951, #1796 and #1679.

andeon:
The issue is gone with this fix https://github.com/opentoonz/opentoonz/pull/4860, thanks!

Issue #1951, opened 2018-05-07
Cursor icon stuck on "change width" 


TGGeko:
When moving the cursor off the workspace, it changes to the "change width" cursor icon, and stays there when hovering over any toolbar or sidebar. It changes to the usual arrow cursor when hovering over the top menu bar. 
![20180507_194651 1](https://user-images.githubusercontent.com/6518222/39730611-a8f3e642-5230-11e8-939c-a75a21ea86c9.jpg)

Open Toons v1.2.0 Build May 7 2018

Forgive the photo, usual screenshots don't capture the cursor.

RodneyBaker:
I saw this briefly on a stripped down release of OT but it was very temporary and I had not seen it before nor have I seen it since.

@TGGeko Can you download the latest nightly release and confirm if this is still an issue?

Link:  https://github.com/opentoonz/opentoonz_nightlies/releases

artisteacher:
I definitely get that in OS X when I'm using a tablet - though I'm not quite sure what triggers it. It happens sporadically. I've assumed it had to do with the tool icons not updating properly since hovering is not currently recognized #1796, #1853 when using a tablet in OT with OS X.

TGGeko:
Issue still persists in Nightly build 5-25-2018

RodneyBaker:
As far as I know this is still a problem.
I've seen something similar in slightly different location but instead of the wrong cursor it temporarily appears as if there is none.

RodneyBaker:
Closing as likely resolved and/or abandoned.
@TGGeko Please let us know if you are still experiencing this issue.

Issue #1796, opened 2018-02-12
Tablet:  Hovering not recognized


ghost:
--- 
Status
---
- [x] Known Issue
- [x] Documented Workaround(s)
- [x] Documented Code Location(s) / Upstream depencies
- [ ] Drivers confirmed
- [ ] Priority Established
- [ ] Bounty Attached
- [ ] Developer Assigned

---
Report
---

The small yellow dot that appears with hover before we set an onionskin point doesn't appear anymore after I start to paint in the Viewer. It seems like the pen tablet hovering is not properly recognised. I'm on OSX 10.12.6, Wacom Tablet and OT 1.2.

It happens with the Timeline and also with the Xsheet. I have attached a small screen recording.

Thanks!

[Issue.mov.zip](https://github.com/opentoonz/opentoonz/files/1716650/Issue.mov.zip)

DotSilva:
Strangely enough, if I change the Wacom Tablet Settings 'Tip Feel' and open Open Toonz, it momentarily works before returning to the described issue.

artisteacher:
I’ve been having the same problem in 10.11.6 and 10.13.5. It seems that if OpenToonz is already open and a tablet is connected, it doesn’t register hover events. If the tablet is connected before OT opens, then hovering seems to work as expected - but only temporarily. For example, if the stylus leaves the proximity of the tablet, once the stylus is back in range, hovering no longer works. Disconnecting & reconnecting the tablet also stops hovering from working. If the tablet is connected before OT opens and the stylus does not leave the range of the tablet, all of the hover functionality and tool tip changes register as expected.

manongjohn:
Just an FYI,

This is not an OT issue. There is actually an open QT bug reported (QTBUG-65559) that is related to this issue.  It's an issue with pen and mouse tracking not working.

The problem is pen moments typically send 2 events, tablet and mouse events but now it's only receiving tablet events.  Tablet events are where we get pressure data from, but the mouse events are used for a number of things and can help to trigger cursor changes, mouse over button highlights, tooltip display and other mouse over effects.

I will go further and say I don't believe this is a QT version issue, but rather a macOS issue.  I've compiled OT using 5.9.2 and 5.9.7 on  macOS 10.11 (El Capitan) and the problem persists, only tablet events are generated.  I'm guessing it might have started with El Capitan.

Funny thing is, you can make it generate mouse or tablet events by doing the following:
[...]

PR #4860, opened 2023-04-29
Fix tablet tools

flurick:
(again on linux, maybe macos also?) 

This just removes two lines that blocked tablet input, so assuming they where there for some reason something "should" break, but seems to fix thing like: 
- setting the right cursor when re-entering the canvas
- allow editing with tape tool
- allow drawing arcs

### DIFF
--- a/toonz/sources/tnztools/vectortapetool.cpp
+++ b/toonz/sources/tnztools/vectortapetool.cpp
@@ -327,7 +327,7 @@ public:
     if (!vi) return;
 
     // BUTTA e rimetti (Dava problemi con la penna)
-    if (!m_draw) return;  // Questa riga potrebbe non essere messa
+    //if (!m_draw) return;  // Questa riga potrebbe non essere messa
     // m_draw=true;   //Perche'??? Non basta dargli true in onEnter??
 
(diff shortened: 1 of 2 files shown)


flurick:
#4688 #2569 #1951 #1796 #1679

RodneyBaker:
I haven't seen any adverse effects on Windows.  :)

toonz/sources/tnztools/vectortapetool.cpp
```diff
@@ -327,7 +327,7 @@ class VectorTapeTool final : public TTool {
     if (!vi) return;
 
     // BUTTA e rimetti (Dava problemi con la penna)
-    if (!m_draw) return;  // Questa riga potrebbe non essere messa
+    //if (!m_draw) return;  // Questa riga potrebbe non essere messa
```
shun-iwasawa, line 330: The function `VectorTapeTool::draw()` has the same condition and it occasionally block to draw the tool UI properly.
As far as I tested, I think we can just remove `VectorTapeTool::m_draw` and all lines related to it. I'll do so after merging this PR.
Thank you @flurick for the fix, LGTM

PR #4862, opened 2023-05-01
Remove Vector Tape Tool "m_draw" Flag

shun-iwasawa:
As a continuation of the PR #4860 by @flurick this PR removes the flag `VectorTapeTool::m_draw` and all related lines.
As far as I tested, this flag prevents from drawing the tool UI property on macOS. Removing the flag seems to be with no problem.


RodneyBaker:
@shun-iwasawa I can't test for the Mac side of things so please feel free to press forward with merger as you see fit.
Alternatively, if you say the word I'll be glad to merge.

PR #4866, opened 2023-05-02
Fix cell selection resizing with arrow keys not to refer to the current frame

shun-iwasawa:
This PR is a  small modification to #4592 by @flurick.
Instead of using the current frame, in this PR I introduced a variable `TCellSelection::m_resizePivotRow` and use it as an pivot frame when resizing the selected cells with arrow keys.


RodneyBaker:
Thank you @flurick for the initial fix and @shun-iwasawa for the modification.
LGTM

PR #4872, opened 2023-05-08
Fix the dialogs to be closed when clicking the buttons, instead of pressing.

### LINKED ISSUE #4864: Creating a motion path got stuck accepting the new path
### Description

After drawing a motion path, the dialog appeared, asking if I want to replace the motion path. Clicking yes just re-showed the same dialog.
I closed and re-opened OpenToonz and it happened again. So I had to find a different work around to move my matte.

### Steps to Reproduce

1. In the stage schematic, I selected a column, then pressed the Motion Path button to add a path to it.
2. Then I selected the brush and drew the path
3. The dialog appeared (Are you sure you want to replace the motion path), clicking Yes closed and reopened the dialog. Clicking No did the same! I had to press the X to get out of the cycle.

Notice that when I first pressed Yes, the Green filled rectangle moved (this is the image on the column), and subsequently clicking yes, the rectangle followed my click. It appeared that instead of the button getting the click, the viewer did, which recreated the path (with a single point) and that's why the question re-appeared.

### Expected Behavior

Clicking yes should accept my path.

ScreenRecording

### OpenToonz Version

Other (Specify below)

V1.7rc

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Huion

shun-iwasawa:
This PR fixes #4864 .
Changed the buttons of dialogs opened with `DVGui::MsgBox()` , to be triggered when clicked (i.e. press and release the mouse button while the cursor is on the same widget) , instead of when just pressed.
This will ensure that the mouse is released when the dialog is closed, preventing unwanted events in the scene viewer.

### DIFF
--- a/toonz/sources/include/toonzqt/dvdialog.h
+++ b/toonz/sources/include/toonzqt/dvdialog.h
@@ -261,7 +261,7 @@ public:
 
 public slots:
   void onCheckboxChanged(int checked);
-  void onButtonPressed(int id);
+  void onButtonClicked(int id);
 };
 
 //-----------------------------------------------------------------------------
--- a/toonz/sources/toonzqt/dvdialog.cpp
+++ b/toonz/sources/toonzqt/dvdialog.cpp
@@ -807,7 +807,7 @@ MessageAndCheckboxDialog::MessageAndCheckboxDialog(
 
 //=============================================================================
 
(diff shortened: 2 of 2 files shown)


RodneyBaker:

DarrenTAnims:
Superb fix. Thanks @shun-iwasawa for this, and for all your work on OpenToonz.

PR #4873, opened 2023-05-08
v1.7.0

shun-iwasawa:
This PR will be merged just before releasing the next version.
It removes `RC` from the version name.


shun-iwasawa:
🚀

PR #4885, opened 2023-05-10
qol: Prevent CTRL activating draw straight line angle snap

konero:
A QoL change, CTRL seems to fire or become sticky enough that it's causing issues with shortcuts, I think it's good to instead only activate angle snap when holding Shift & CTRL (before, it was just either), which this change does.


RodneyBaker:
Merging to fix the problematic use of Control key while drawing workflow with Undo.

Thank you @konero for the modification.
LGTM

PR #4903, opened 2023-05-17
Antialias xsheet

flurick:
No aa:
before
With aa:
after

The loop arrow icon maybe looks a bit too blury? Other than that things are slightly more round

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -1430,6 +1430,8 @@ void CellArea::drawExtenderHandles(QPainter &p) {
     selected.adjust(0, topAdj, 0, bottomAdj);
   }
 
+  p.setRenderHint(QPainter::Antialiasing, true);
+
   int x0, y0, x1, y1;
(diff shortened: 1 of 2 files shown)


konero:
Hmm, I think it looks too fuzzy. The range handles at least should be crisp, as well as the straight edges of the cell drag handle.

The straight line on the onion skins should technically remain 1px, but it's now 2px. I can't remember too well what the solution was to problems like this in QPainter, but I think drawing the shapes by shifting their x and y positions by 0.5 (half a pixel) helped to snap them properly to screen pixels.

image

As this PR is now, it makes it feel like I need to put some glasses on.😂

RodneyBaker:
Thanks @konero, I was hoping someone would add their thoughts/feedback.
There definitely needs to be some additional discussion.

I personally prefer the crisper no-antialias look.
( and my eyes aren't getting better as I get older so the entire world is getting blurrier  ;) )

RattusaurusRex:
I think I prefer the aliased version better in the example above.  There doesn't seem to be as much of a difference with the higher res icons: 
aa

Issue #4229, opened 2022-01-22
Help needed with custom Flipnote style pattern brushes


GoodbyeGhost:
## The Goal
I've been wanting to replicate the style of Flipnotes in my animation for a while now, but don't know how I would go about making custom brushes, if it's possible to do so with what I want. I'm aiming to replicate the 10 dither brushes found in this Krita package: https://sabs.like.sobs.moe/post/185842858590/hello-this-is-a-small-brush-pack-for-the-free-art. It's probably best if you use the pack to better understand what I'm wanting to make.

If making just the 10 dither brushes isn't too crazy of a thing, I wouldn't be apposed to the idea of recreating the entire package for Opentoonz either. I have tons of free time pitching in if needed, and really want this to happen, as I'm sure there are more people than myself who like this style and would welcome Opentoonz into their workflow if these brushes were available. 

I don't have first-hand access to the Flipnote software, so this Krita package is the most accurate parallel to the brushes on Flipnote that are readily available, to my knowledge.

## Concerns
I don't have much technical knowledge of what goes into making custom brushes, so I'm not sure how viable these dithering brushes are to make or use. I have tried to make custom raster brushes in the past but to no avail. 

## Enticement
On my path to try and replicate the style, I came across a like minded individual who also ran into this problem, CS Ghost Animation (Youtuber). Not that it adds much, but I'm sure he would appreciate the availability of these brushes as well, since he has to copy and paste in order to obtain the dithering effect seen in Flipnote.

RodneyBaker:
I downloaded the brush pack and am looking into it.
In the meantime, have you contacted the author of the brushes?
They would be the first best place to get advice on porting the brushes over to Opentoonz.
They might even know a little about how to get that done.

Disclaimer:  I am not at all familiar with Flipnote and have very basic understanding of Krita.

GoodbyeGhost:
True, I was looking at contacting them, I'll see if they can help port it to Opentoonz. It's not exactly clear how they implemented them into Krita, so I'm sure they'll have some good insight

RodneyBaker:
Any information they can provide will help.
If we are going to port the brushes it'd be best to do it in a way that the brushes meet their full potential.

GoodbyeGhost:
[...]

PR #4911, opened 2023-05-21
Feat: adding screentones

andeon:
Patterns to simulate screentone https://github.com/opentoonz/opentoonz/issues/4229

First things first, go to Preferences and in the General tab look for "Minimize Raster Memory Fragmentation*". The change won't go into effect until OpenToonz is restarted.

Create the Toonz Raster level and new plain colors, go to the texture tab and choose the screentone to each color. In "Settings" tab you should set some parameters for the style: 

1. Tick the 'Use As Pattern' 🗹 ; 

2. Set 'Contrast' to 10.

 ***Make sure the color in the tab color is correct, especially the alpha parameter.**

Use the textures with "Fill Tool (F)", "Brush Tool (B)" and "Paint Brush Tool". You can set a plain color with "Fill Tool" or "Paint Brush Tool" and overlap it with another color _texture_ "Brush Tool" on the same level. But you cannot overlap two color _texture_ on the same level.

To create this effect, you need to create multiple levels, or to speed up the workflow you can duplicate levels by cloning cells for a new level and change the current textures colors to anothers. Main menu "Cells > Clone Cells" select all the desired cells you want to clone.

***Setting the "Eraser tool" to "freehand typing" to erase areas or lines is a nice trick to work with this style.**

***Press the "Preview" button to visualize the texture effect, avoid working with it active all the time.**

image

The textures was optimized for 12x12 and 16x16 and I think that cover the most common effects.


RodneyBaker:
Very nice @andeon 
In my first test I forgot the requirement to use Preview and found myself wondering why the screentones weren't working.  
Always read and understand the directions no?  ;)

As this PR is just adding images to the OT Library will post intent to merge although there is no rush here.
I was planning to hold off on adding any new features until 1 June as that will be a good date to remember for development purposes (i.e. there would be only bugfixes prior to 1 June).

This will also allow anyone testing the process of using these screentones to provide feedback.

Hopefully this delay is acceptable to you.

andeon:
> Very nice @andeon In my first test I forgot the requirement to use Preview and found myself wondering why the screentones weren't working. Always read and understand the directions no? ;)
> 
> As this PR is just adding images to the OT Library will post intent to merge although there is no rush here. I was planning to hold off on adding any new features until 1 June as that will be a good date to remember for development purposes (i.e. there would be only bugfixes prior to 1 June).
> 
> This will also allow anyone testing the process of using these screentones to provide feedback.
> 
> Hopefully this delay is acceptable to you.

No problem @RodneyBaker, it's good to do more testing so I can change or add anything that's missing.

@Bracket-H Special settings can be a headache if you are working with a lot of textures, although this is mostly a minimalist style that uses few textures and colors. The biggest problem here is that the huge use of the combination of editing, painting, preview, undo and redo can make it unstable, even if you try to avoid this by activating the "Minimize raster memory fragmentation" function.

andeon:
Is there any possibility to enable the texture for use in the "Raster" level, so that we can get this effect in real time with Overlapping/Dithering and so on. It could work similarly to "mypaint brushes", so we will have a huge improvement.

PR #4916, opened 2023-05-24
Update how_to_build_linux.md

gelatinbomb:
It seems at least for Fedora 37 some extra packages are needed in order to compile Opentoonz so I decided to add those to the instalation guide for Linux:
- turbojpeg-devel
- opencv-devel
- qt5-qttools-static
- qt5-qtserialport-devel

I have never done a pull request before so if there is anything wrong you can tell me :)

### DIFF
--- a/doc/how_to_build_linux.md
+++ b/doc/how_to_build_linux.md
@@ -38,7 +38,7 @@ Notes:
 (it may include some useless packages)
 
 ```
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Looks straightforward enough to me.
Thank you @gelatinbomb for the update to linux build instructions.

PR #4920, opened 2023-05-29
Fix batch rendering via task panel

### LINKED ISSUE #4897: Ot 1.7.1 render issues - render doesn't start at all from task room
### Description

as in previous versions also in 1.7 it is very tiring to render scenes that contain assets that use the plastic tool. To generate a movie of about 180 frames I had to render 6 clips of 30 frames each. Normally the solution is to generate a sequence of images by launching the render from the task room, but in this version the render from the task room doesn't start at all, both in scenes that use plastic and in scenes that don't use it.

### Steps to Reproduce

1. create a scene, set the output settings and save
2. from the browser room select the scene, right click "add as render task"
3. in the task room start the render

### Expected Behavior

 the render icons will get immediately red.

Schermata 2023-05-16 alle 19 24 50
Schermata 2023-05-16 alle 19 24 57

### OpenToonz Version

1.7

_No response_

### Operating System

macOS

### GPU

AMD

### Graphics Tablet

None

shun-iwasawa:
This will fix #4897 in a different way than #4904 .
Modified all arguments to be put in `QStringList` , instead of a single-line `QString`.

@Bracket-H
Can you please try this fix?

### DIFF
--- a/toonz/sources/include/tfarmtask.h
+++ b/toonz/sources/include/tfarmtask.h
@@ -135,6 +135,7 @@ public:
 
   QString getCommandLinePrgName() const;
   QString getCommandLineArguments() const;
+  QStringList getCommandLineArgumentsList() const;
   QString getCommandLine(bool isFarmTask = false) const;
   void parseCommandLine(QString commandLine);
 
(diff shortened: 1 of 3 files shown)


artisteacher:
Works on the macOS side too!

RodneyBaker:
Thank you @shun-iwasawa for the fix
And thank you @Bracket-H and @artisteacher for the confirmation.

PR #4923, opened 2023-05-30
Fix Text Iwa Fx input field

### LINKED ISSUE #4922: Japanese Input Bug in Text Iwa FX?
### Description

Hello,

At the begining of this year, I posted a message on the English User Group. Since I did not receive any reply, and because the problem still occurs also in the new Opentoonz 1.7 version, let me mention it again here.

Whenever I use Text Iwa Fx and I try to input Japanese text inside the textbox of the popup window, I never get words written correctly in Japanese. 

For example, if I try to input the word さくら (sakura), 
the word is wrongly rendered like ｓあｋうｒあ.

On the other hand, in the following cases, the problem does not occur, and I can easily input any Japanese word correctly. I can input Japanese text in the Memo popup, or with the Type tool on the current scene, or inside a cell of a Note level, or whenever I use a name written with Japanese characters for the scene I am going to save.

Thank you in Advance,
Tony

### Steps to Reproduce

1. 
2. 

### Expected Behavior

The correct orthography should be さくら

_No response_

### OpenToonz Version

1.7

_No response_

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

None

shun-iwasawa:
This will fix #4922 
Added an event filter to ignore OpenToonz's shortcut commands.

### DIFF
--- a/toonz/sources/include/toonzqt/paramfield.h
+++ b/toonz/sources/include/toonzqt/paramfield.h
@@ -613,12 +613,12 @@ namespace component {
 class MyTextEdit : public QTextEdit {
   Q_OBJECT
 public:
-  MyTextEdit(const QString &text, QWidget *parent = Q_NULLPTR)
-      : QTextEdit(text, parent) {}
+  MyTextEdit(const QString &text, QWidget *parent = Q_NULLPTR);
 
 protected:
   void keyPressEvent(QKeyEvent *event) override;
   void focusOutEvent(QFocusEvent *e) override;
+  bool eventFilter(QObject *, QEvent *) override;
 
 signals:
   void edited();
--- a/toonz/sources/toonzqt/paramfield.cpp
+++ b/toonz/sources/toonzqt/paramfield.cpp
@@ -9,6 +9,7 @@
 #include "toonzqt/doublepairfield.h"
 #include "toonzqt/tonecurvefield.h"
 #include "toonzqt/checkbox.h"
+#include "toonzqt/menubarcommand.h"
 
 #include "tdoubleparam.h"
 #include "tnotanimatableparam.h"
@@ -1563,6 +1564,11 @@ void IntParamField::update(int frame) {
 
 namespace component {
 
+MyTextEdit::MyTextEdit(const QString &text, QWidget *parent)
+    : QTextEdit(text, parent) {
+  installEventFilter(this);
+}
+
 void MyTextEdit::keyPressEvent(QKeyEvent *event) {
   QTextEdit::keyPressEvent(event);
   if (event->key() == Qt::Key_Return) emit edited();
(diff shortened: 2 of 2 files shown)


TonyPolizzi:
Hi,

Thank you very much Iwasawa-san. I am new to this forum. Will #4923 fix be available for testing in the next nightly build?
Ijoo, yoroshiku o-negai-shimasu,
Tony Polizzi

RodneyBaker:
@TonyPolizzi Every PR is testable via a portable release of Opentoonz by accessing those automated builds either via the Details section at the bottom of each PR or via the Actions menu at the very top of github.  This is optimal for testing BEFORE the PR gets merged into the nightly where it can be further used/tested and then eventually (annually) added into the main release.

But yes, every PR is testable via the nightly releases.

And... a plus with the automated builds is also that the change is better tested in isolation of other PRs.

The main reason this PR isn't already in a nightly release is that no one has tested the automated builds and given their feedback.  As such it's mostly waiting on me to test and confirm.  Look for the fix very soon however in a nightly release.  :)

Here's a direct link to the WIndows build (and please note these builds are portable so don't need to be installed, just unzipped and ran):

https://github.com/opentoonz/opentoonz/suites/13236173740/artifacts/721338563

RodneyBaker:
LGTM

TonyPolizzi:
Thank you to all. I have downloaded the nightly build May 31, 2023, 10:03:25.
Unfortunately, I don't notice any improvement. The issue remains. In fact, previously, whenever I typed Latin alphabet characters, I would immediately see the correct result on the screen. Now it has become difficult to even type Latin characters within the text box of the FX Settings window in TextIwa.

Going in order:
1. I switch to Preview mode.
2. I double-click the left mouse button on the TextIwa01 node in the FX Schematic.
3. The 'FX Settings: TextIwa01' window opens correctly.
4. I click the left mouse button in the text box, and the blinking cursor appears.
5. The input language in my case (Windows 11 Home Italian 64-bit version) is set to Half-Width Alphanumeric Japanese. The Japanese input system is the one provided with the operating system (Microsoft IME). However, the problem happens also when I shift to the Italian IME.
6. As soon as I type any letter in the text box of the 'FX Settings: TextIwa01' window, no letters appear, but the icons in the toolbar gradually become highlighted in blue.
7. If the letter I type in the box doesn't correspond to any keyboard shortcut, for example, "d," only in this particular case, a "d" appears correctly in the text box. By holding down the D key, I am able to type dozens or hundreds of "d" in the box. The same applies to certain other letters or numbers.

If I switch to the Japanese Hiragana input mode, the situation doesn't change. It seems that priority is given to keyboard shortcuts related to various Opentoonz tools.

Apparently, the only solution I have at the moment is the one suggested by Bracket-H, which is to type the Italian or Japanese word in Windows NotePad and then copy and paste it into the text box of the 'FX Settings: TextIwa01' window.

RodneyBaker:
Yes, I think this is a shortcut issue.
For instance, if we try to type the letter 'Z' in into a cell of a Notelevel... we cannot.
To test this:
- Create a Notelevel in the xsheet/timeline
- Type in the word 'Opentoonz' and note that we cannot type the lower case letter 'z'.
- Note that we can enter the capital letter 'Z'
- Now change the shortcut for 'Set Key' from 'z' to 'alt +z' or other desired shortcut
- Note that now we CAN type the letter 'z' where we could not before.

PR #4924, opened 2023-05-31
Fix stylesheet variables

shun-iwasawa:
This PR fixes property variable names used in the style sheet.
As you can see, the two variable names were obviously reversed.


RodneyBaker:
Thanks @shun-iwasawa for (obvious!) fix.

PR #4925, opened 2023-05-31
Tweak skipping holds

flurick:
Previously **Shift + Arrows** would jump to the cell next to the "block" of drawings.

The idea is to use the same keyboard shortcuts as what a boring old text editing field does, but for navigating the xsheet.
**Shift +  Arrows** grows/shrinks selection.
**Ctrl + Arrows** move to end/beginning of block of drawings. 
**Shift + Ctrl +Arrows** select blocks.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -1357,8 +1357,7 @@ public:
 //-----------------------------------------------------------------------------
 
 bool changeFrameSkippingHolds(QKeyEvent *e) {
-  if ((e->modifiers() & Qt::ControlModifier) == 0 ||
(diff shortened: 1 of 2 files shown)


RodneyBaker:
There is a lot to like about this change.
I'd love to hear additinoal feedback from those willing and able to test.

RE:  Similar question as asked before... Is the draft status of this PR intentional because additional work or consideration is being done?

flurick:
> There is a lot to like about this change. I'd love to hear additinoal feedback from those willing and able to test.
> 
> RE: Similar question as asked before... Is the draft status of this PR intentional because additional work or consideration is being done?

This time it's intentional :)
For example, other columns could be taken into account for both frame and column jumping.

RodneyBaker:
Closing this PR as part of routine maintenance.
We can reopen at any time.

PR #4934, opened 2023-06-05
Fix Text Iwa Fx input field (again)

### LINKED ISSUE #4922: Japanese Input Bug in Text Iwa FX?

shun-iwasawa:
This will partially fix #4922 
The previous fix #4923 seems to be wrong. (So sorry for the trouble!)
I found one of the cause is the panel taking the focus while typing in the field.

Unfortunately, I found that this PR does not fix the problem completely. (It still sometimes input `sあ` instead of `さ`)
But the problems occur less frequently.

### DIFF
--- a/toonz/sources/include/toonzqt/paramfield.h
+++ b/toonz/sources/include/toonzqt/paramfield.h
@@ -618,7 +618,6 @@ public:
 protected:
   void keyPressEvent(QKeyEvent *event) override;
   void focusOutEvent(QFocusEvent *e) override;
-  bool eventFilter(QObject *, QEvent *) override;
 
 signals:
   void edited();
--- a/toonz/sources/toonz/pane.cpp
+++ b/toonz/sources/toonz/pane.cpp
@@ -34,6 +34,7 @@
 #include <QDesktopWidget>
 #include <QDialog>
 #include <QLineEdit>
+#include <QTextEdit>
 #include <QScreen>
 
 extern TEnv::StringVar EnvSafeAreaName;
@@ -123,7 +124,8 @@ void TPanel::enterEvent(QEvent *event) {
   if (w) {
     // grab the focus, unless a line-edit is focused currently
     QWidget *focusWidget = qApp->focusWidget();
-    if (focusWidget && dynamic_cast<QLineEdit *>(focusWidget)) {
+    if (focusWidget && (dynamic_cast<QLineEdit *>(focusWidget) ||
+                        dynamic_cast<QTextEdit *>(focusWidget))) {
       event->accept();
       return;
     }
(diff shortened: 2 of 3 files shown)


TonyPolizzi:
Thank you, Iwasawa-san

- Typing in Italian or English now seems to be back to normal, although slightly lagging.
- Typing in Japanese, on the other hand, still presents the same issues, although not with all words.

Some examples of Japanese words that I can type correctly:
ふじ (Fuji)
おまえ (Omae)
とうきょう (Tokyo)
おおさか (Osaka)
よなぐに (Yonaguni)
えひめ (Ehime)
ひろしま (Hiroshima)

Some examples of Japanese words that I cannot type correctly:
ｓあｋうｒあ (sakura)
ｓあっぽろ (Sapporo)
ｎあごや (Nagoya)
ｙおｋおはま (Yokohama)
ｋｙおうと (Kyoto)
ｆうｋうおか (Fukuoka)

However, in some cases, like さくら (sakura), for reasons I cannot understand (perhaps even typing speed affects the hiragana conversion), I sometimes get different results than ｓあｋうｒあ. In rare cases, I even get the correct result, さくら.

Additionally, in some cases, if I type the same k key twice, like in the initial letter of the word きょうと (Kyoto), I get ｋきょうと as a result, which allows me to correctly convert the hiragana part into kanji, resulting in k京都. At that point, I just need to remove the initial k. However, even in this case, if I change the typing speed, I can get a different result like ｋｋｙおうと, which does not allow for kanji conversion.

shun-iwasawa:
@TonyPolizzi 

Before typing, try shifting the mouse cursor up slightly so that the cursor does not overlap the pop-up of the conversion suggestions (orange square) .
In my case, that prevented the problem.
image

TonyPolizzi:
Dear Iwasawa-san,

Excellent solution! Thank you very much. It works perfectly.

PR #4937, opened 2023-06-06
Link Column Name with Level option by Contrail Co., Ltd.

shun-iwasawa:
This PR will add a new preference option `Link Column Name with Level` to `Preferences > Xsheet` category.

OpenToonz allows multiple levels in one column. This is a very flexible advantage, but in the Japanese animation industry, in most cases, only one level goes in one column.
This option "reserves" the levels that will be placed in a column by the name of the column, so that you can set cells in an empty column by simply entering a number.

image

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -421,6 +421,9 @@ public:
   bool isShowFrameNumberWithLettersEnabled() const {
     return getBoolValue(showFrameNumberWithLetters);
   }
+  bool isLinkColumnNameWithLevelEnabled() const {
+    return getBoolValue(linkColumnNameWithLevel);
+  }
 
   // Animation  tab
   int getKeyframeType() const { return getIntValue(keyframeType); }
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -140,6 +140,7 @@ enum PreferencesItemId {
   currentColumnColor,
   levelNameDisplayType,
   showFrameNumberWithLetters,
+  linkColumnNameWithLevel,
 
   //----------
   // Animation
--- a/toonz/sources/include/toonz/txshlevelcolumn.h
+++ b/toonz/sources/include/toonz/txshlevelcolumn.h
@@ -19,6 +19,7 @@
 // forward declarations
 class TLevelColumnFx;
 class TXshCell;
+class TXshLevel;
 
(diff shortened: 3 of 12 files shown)


DarrenTAnims:
So with this change does it mean that when you change the name of the column it changes the name of the level?

RodneyBaker:
There are some VERY interesting things going on with the option/change @shun-iwasawa.
I'm a user who loves using multiple levels in the same column but the benefits of this option may be well worth adjusting my workflow.  

Minor observation
The only thing I see that doesn't appear to quite work as well as it could relates to a user manually renaming a column or level.
When the column or level is renamed they don't automatically resync.
It takes moving the column or level (or presumably some other action that will force a refresh) to get the names back in sync.

Thanks for the comments in the code as well.

RodneyBaker:
> So with this change does it mean that when you change the name of the column it changes the name of the level?

From my testing if we change the name of the column the level name doesn't change.
But if we change the name of the level... the column name changes (although not always immediately... see my post above).

As mentioned above by @shun-iwasawa if we name an EMPTY column... or rename it... it becomes reserved for a level of that name.  Note that this doesn't keep us from moving or pasting a different level into that column and that will update the column name... but.. the reservation is so that we can type just the number and if the level already exists the cell will be populated by that frame of the reserved level.   (This may need to be experienced or demo'd to fully understand as reading this probably doesn't convey the full story)

DarrenTAnims:
That link didn't give me this feature. Adding levels just created a new column named the normal way, as you can see from the image.

image

RodneyBaker:
And you did turn the option on in Preferences > Xsheet?

RodneyBaker:
That sounds like a request for this feature to be on by default @DarrenTAnims as that would pretty much implement what you describe.

RodneyBaker:
A separte request will be appreciated as we don't want to delay the option being added.
Posting intent to merge (if anyone has additional feedback they wish to add)...

RodneyBaker:
Thank you @shun-iwasawa  (and Contrail Co. Ltd) for the new column/level option.
LGTM!

RodneyBaker:
I'm adding the following gif for two reasons:
1- to demonstrate we can still use multiple levels in a column with this option on (the column name is derived from the first level in the column similarly to how the column preview icon uses the first frame in the first level encountered)
2 - to demonstrate the slight nuance where column names don't always stay in sync which can require the levels to be moved to an empty column to refresh the name.

ColumnLevelNames

This option is definitely a great time saver.

PR #4947, opened 2023-06-12
Fix Soapbubble fx

shun-iwasawa:


RodneyBaker:
Thank you @shun-iwasawa for the gamma related fix!

Issue #4957, opened 2023-06-16
Sluggish response after addition of Assistants Tool


RodneyBaker:
This is just an initial report to let others know of issues appearing after the incorporation of the Krita-like Assistants Tool.
It's not entirely clear what is causing the problem nor why the problem is appearing now but not in original testing.

Hopefully the problem will be resolved speedily and we won't need to revert the Assistants Tool PR.

I did want to let others know the problem has an associated bug report opened.

### Steps to Reproduce

To test just install any of the latest nightly releases after the inclusion of Assistants Tool from #4852 

If the problem is present on your system it will be very obvious.

konero:
Can you give more detail?

RodneyBaker:
I'm currently on the road so can't provide additional detail.
I did want to get the word out though to be on the look for issues.
If you are using a nightly and NOT experiencing any issues then... very interesting.

konero:
> Are you having no issues with currently nightly while or after using the Assistants Tool @konero?

I didn't test nightly (I built myself) but I am having a few issues, I outlined them in #4852.

RattusaurusRex:
Drawing with stylus on ToonzRaster level: 

https://github.com/opentoonz/opentoonz/assets/60277069/7a4c3665-f34a-4e2b-b9aa-9df8271e0593

RodneyBaker:
An update here...

After noticing my entire system was being laggy, I decided to restart my system (again as I had done before with no change) but this time I decided to download the latest Wacom driver for my tablet and...

After a restart... no lag.
I'm currently playing with the Assistants to see when/if the lag starts afresh.

RodneyBaker:
Keeping this open for now to collect troubleshooting information for users that have issues so we can work through them.

andeon:
It increases usage memory very fast over time, I don't know why this leak, maybe it registers and calculates constantly, so it needs to be better optimized.

Wouldn't a [SQLite](https://www.sqlite.org) database work well with this concept?

RodneyBaker:
Will close this as #4963 resolves the issue with Assistant Tool slowness.  (Thank you @shun-iwasawa!)
For other issues likely related to addition of Assistants Tool please see:  #4968

blackwarthog:
Sorry for the late answer. I worked on implementation of assistants for other tools (toonz brush and toonz raster brush), and did not visit github for a while. I'm trying to reduce my presence on github since MS bought it. Now i've enabled email notifications to be in touch.

[...]

PR #4963, opened 2023-06-19
Fix Assistant Tool Slowness

shun-iwasawa:
This PR will partially fix the slowness mentioned in #4852 and #4957 .

- Reverted the update method of tool options bar, from re-building whole options from scratch to show/hiding the relevant fields. Which will
  - remove unwanted "Size" field in the tool option of Toonz Raster Brush, and
  - will resolve the slowness when dragging on the color wheel in the style editor.
- Made the twizzler shape to be available only in the debug builds.
- Reverted the `SceneViewer::tabletEvent` to modify the slowness when using tablet

### DIFF
--- a/toonz/sources/include/tools/modifiers/modifiertest.h
+++ b/toonz/sources/include/tools/modifiers/modifiertest.h
@@ -1,3 +1,5 @@
+
+#ifndef NDEBUG
 #pragma once
 
 #ifndef MODIFIERTEST_INCLUDED
@@ -5,7 +7,6 @@
 
 // TnzTools includes
 #include <tools/inputmanager.h>
-
 
 #undef DVAPI
 #undef DVVAR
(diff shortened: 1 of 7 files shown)


shun-iwasawa:
I think it would be better if the Assistant tool becomes more like the Shift and Trace, which will activate a special tool when active.
And if the assistant column can be narrower and placed next to the camera column..? Well, I think I need more investigation.

RodneyBaker:
Thank you so much for addressing this current problem.
Your fix does indeed appear to address the extreme slowness brought on by the addition of Assistants Tool.

The appveyor build failed and checks didn't pass and I see that as critical especially as when testing prior things worked fine under github build but then not after appveyor build of the nightly release.  While I see many errors reported, the main culprit appears to be failure to link up with a DLL file.

shun-iwasawa:
UPDATE:
- Fixed the debug build error
- Reverted the `SceneViewer::tabletEvent` to modify the slowness when using tablet

RodneyBaker:
Thank you @shun-iwasawa for the fix... update... and the refresh.
LGTM

PR #4966, opened 2023-06-20
Fix missing initialization

shun-iwasawa:
I found a missing initialization when I was implementing another feature.


RodneyBaker:
This fix is such a simple change that I don't think the appveyor checks failure should prevent proper merging.
But merging without it completing all checks is more of a curiosity than anything else.
Perhaps best to be patient and wait for a refresh of this PR.

Over to you @shun-iwasawa and please feel free to merge as you see fit.

RodneyBaker:

PR #4979, opened 2023-06-28
Add info how to install OpenToonz as portable and without admin rights required.

jointri:
It works so far, no issues loading OpenToonz. It just needs a (also portable) application called innoextract to fully extract the contents of the installer and then one directory has to be moved. That's all it is. I think this should be added so more people can use OpenToonz where they cannot because they don't have admin rights.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,6 +24,8 @@ Please download and install OpenToonz from the latest installer at <https://open
 
 Older versions and unstable nightly build are also available at <https://github.com/opentoonz/opentoonz/releases>.
 
+If you cannot install because you lack sufficent permissions, read [here](./doc/how_to_extract_opentoonz.md)
+
 ## How to Build Locally
 
 - [Windows](./doc/how_to_build_win.md)
--- a/doc/how_to_extract_opentoonz.md
+++ b/doc/how_to_extract_opentoonz.md
@@ -1,0 +1,12 @@
+This document is assuming you're on windows, but can be modified to work on other operating systems as well.
+
+First, download [innotextract](https://constexpr.org/innoextract/) from the website.
+
+Make a temporary folder where you downloaded it. Ensure the innoextract binary is in the same location as the OpenToonz installer.
+Type the following in your terminal or command prompt:
+.\innoextract.exe OpenToonzSetup.exe --output-dir <folder you made>
+
+cd to the extracted folder.
+
+You should see two directories, app and one (most likely) called code$GetGeneraldir. Rename that to portablestuff.
+Move portablestuff into the app folder. You can now rename app as OpenToonz or whatever you like and move it anywhere you desire.


RodneyBaker:
Thanks @jointri I was unaware of the Innoextract utility.
Very interesting... and useful.

jointri:
> Thanks @jointri I was unaware of the Innoextract utility. Very interesting... and useful.

Thanks, glad I could be of help. You think it's a possibility to have OpenToonz in the future ship portable zip releases right from the releases download page? Either way, having a option to easily make your own portable version is good.

RodneyBaker:
Anything is possible.  Someone just has to build it (i.e. automate the process) and the Opentoonz github would have to present it in the releases page.  I don't think innoextract would be needed for this because the files are already in their pre-installation condition so packaging them together only to extract them later would probably be an unnecessary process.

Having said that...
I believe Innoextract is licensed lippng and could be included with OT so that in the installation process the user could opt to create a portable rather than install.  It isn't clear to me if the same restrictions where users run into issues with admin rights might not also apply to running the installer to get to that process.

Here's the license:  

> Copyright (C) 2011-2020 Daniel Scharrer <daniel@constexpr.org>
> 
> This software is provided 'as-is', without any express or implied
> warranty.  In no event will the author(s) be held liable for any damages
> arising from the use of this software.
> 
> Permission is granted to anyone to use this software for any purpose,
> including commercial applications, and to alter it and redistribute it
> freely, subject to the following restrictions:
> 
> 1. The origin of this software must not be misrepresented; you must not
>    claim that you wrote the original software. If you use this software
>    in a product, an acknowledgment in the product documentation would be
>    appreciated but is not required.
> 2. Altered source versions must be plainly marked as such, and must not be
>    misrepresented as being the original software.
> 3. This notice may not be removed or altered from any source distribution.

konero:
I'm personally against this change, it would make more sense to adjust the installer and add a "install for this user only" option and only trigger UAC when installing for "all users".

But you can't avoid situations where Microsoft Smart Screen blocks the exe because it's unknown, you'll still need Admin rights to bypass it. If a user doesn't have Admin rights, it's probably for good reason?

jointri:
> I'm personally against this change, it would make more sense to adjust the installer and add a "install for this user only" option and only trigger UAC when installing for "all users".
> 
> But you can't avoid situations where Microsoft Smart Screen blocks the exe because it's unknown, you'll still need Admin rights to bypass it. If a user doesn't have Admin rights, it's probably for good reason?

By microsoft smartscreen you mean this?
runanyways
This is easy to bypass because it's only for unrecongized applications. Just click run anyways. That never failed me. It has nothing at all to do with admin rights or anything. The issue is that windows **thinks** you must have admin rights with the UAC prompt just to run the installer at all. That's the problem. But you are right, adding a user only install like how GIMP and other free software programs have as an install option might be a better idea.

RodneyBaker:
@konero Just to clarify, as this PR only adds text to the readme documentation, what specifically is the change you are against?
You would prefer the proposed text not be added to the readme.md file?

(While I'm very interested in the process and use of innoextract (and similar things related to portables)... I'm neither for nor against this specific change in readme text.  While it was submitted here, it could also be added to Opentoonz documentation and/or just tagged as 'resource')

konero:
It's unnecessary, it would make more sense to add a task to the installer script with an option to either:

1. Install for current user only (doesn't require admin, HKCU/{appdata})
2. All users (requires admin, HKLM/{app}).

A single checkbox option will suffice.

RodneyBaker:
Will close with antiipation that the installer may one day be updated.
Thanks @jointri for outlining the optional process which users may use in the meantime.

We don't have many innosetup experts that frequent this arena but it looks pretty straightforward.

GradientGamer-XD:
Innoextracter works, but running the program has issues. i get the error "SOFTWARE\OpenToonz\OpenToonz\TOONZROOT" empty error. Is there a way to manually write it

GradientGamer-XD:
> Innoextracter works, but running the program has issues. i get the error "SOFTWARE\OpenToonz\OpenToonz\TOONZROOT" empty error. Is there a way to manually write it

The issue is, i believe that the app only reads from HKLM not HKCU, im hoping the next version says otherwise. Ive attempted to write the necessary keys in the HKCU root, but it didnt work
image

PR #4982, opened 2023-06-29
Fix unsigned warning in IDE

konero:


openanim:
By any chance, will this solve the problems with Avast? I mean when OT is recognized as a virus and blue screen?

konero:
Oh, no this just fixes a warning in the IDE about comparing `int` vs `DWORD`, specifically `'<': signed/unsigned mismatch`.

RodneyBaker:
Thanks @konero for the fix!

PR #4988, opened 2023-07-01
refactor: Icon stuff

konero:
This removes the hacky way in which the icon theming was implemented, although still far from perfect this is much easier to understand and debug, unfortunately I had to touch many files but most of it is just replacing function calls.

All icons will be loaded into a QMap<QString, QString> using the file base name (without extension) as the key and storing the full file path as its value.

Icons are still created with `createQIcon()`, however I wanted to reduce duplication as much as possible so I combined the `addSpecifiedSizePixmapToIcon` function which I think was meant for custom panels into `createQIcon()` as a size argument, it does the same thing, it will use `svgRenderer` to render the image at the size you pass, not upscale a bitmap.

You can also return a QPixmap of an icon image quickly using `generateIconPixmap()`, it will be automatically themed and acts the same way as `createQIcon()`. I also found it's still useful setting an absolute path in places, for that use `getIconPath()` and it will return the path stored in the QMap.

I also changed all the image manipulation stuff to stay working with QImage and then only convert to QPixmap once, because currently it converts back and forth between QPixmap and QImage multiple times for each icon, although not itself an issue it seems a waste regardless.

For that temp workaround that was implemented to hide icons in Windows menus by adding a transparent pixmap for 16x16 due to Qt5.15.2 I've marked all the places I believe handles this with a ID number that can be searched in the IDE when this needs to be removed: `#20230627`.

Change the painting events for TPanelTitleBarButton() as it was causing tens of thousands of image recomputations it seems just by moving my mouse around the GUI for a short time each time a panel got focus or was moved. I made the images members of the classes instead of computing stuff in QPaintEvent (it was my bad).

I completely refactored `LayerHeaderPanel()` (the bar that holds the visibility icons in timeline) to be constructed from QToolButtons instead of using QPainter, this was for no other purpose other than that it shrinks the code by 50%.

Made some icons use high dpi such as Xsheet columns and updated the mimetypes. The panel close button will now highlight red (same color as lock/freeze) to improve the UX when trying to move a floating ToolBar or ToolOptionsBar.

And lastly, I swapped the logo in startup window to be the official logo seen on the website and when OpenToonz launches, just because someone pointed out the font in the current one doesn't match.

All in all, I measured about a 40-60% speedup.


RodneyBaker:
@konero I believe if you refresh your initial post it will trigger a new round of checks and... the Mac check should pass now that Shun's boost PR has been merged.  Not sure but surely worth the try.

konero:
Well, all checks succeeded just this time Appveyor timed out but it succeeded before, since I haven't committed anything since, I'm not worrying about it.

RodneyBaker:
My main concern with this PR related to how the code might stay compatible with Tahoma2D (or further diverge)  but I've seen John's response to your PR to port this refactoring over to T2D so that takes care of that concern.

A secondary concern is that of how users with custom themes of the past might expect to have to tweak their older theme files but I believe that won't be overly problematic.  Any issues encountered are easily countered by the number of improvements the refactoring adds into the mix.

Still poking at the code... and appreciating the description of changes provided as that takes care of other concerns... but posting general intent to merge.

For those testing, now is a great time to provide feedback.

RodneyBaker:
Thanks for this refactoring @konero 
LGTM!

PR #4993, opened 2023-07-05
Fix straight line regressions

### LINKED ISSUE #4968: Mypaint Geometry tool and  TLV Raster shift for Straight line Bug.
### Description

Mypaint Geometry and TLV Raster and Raster Brush shift for Straight line is Broken in the Latest nightly build due to the new Assistant Perspective tool merge.

### Steps to Reproduce

1. try to use the Geometry rectangle tool with Mypaint.
2. try to use the shift for straight line Brush feature with Mypaint in Raster TLV, also Mypaint Raster brush has the same shift for straight line problem.

### Expected Behavior

the Mypaint Brush shift should let u maker straight lines and the Geometry tool should work with Mypaint brushes. 

_No response_

### OpenToonz Version

1.7

current Nightly build

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Wacom

shun-iwasawa:
This PR fixes #4968 .
Now the class `MyPaintToonzBrush` seems to need the 4th argument of the constructor `interpolation` being set to `true` in order to work properly.

### DIFF
--- a/toonz/sources/tnztools/geometrictool.cpp
+++ b/toonz/sources/tnztools/geometrictool.cpp
@@ -1460,7 +1460,7 @@ void GeometricTool::addRasterMyPaintStroke(const TToonzImageP &ti,
 
   m_workRaster = TRaster32P(dim);
   m_workRaster->lock();
-  MyPaintToonzBrush toonz_brush(m_workRaster, *this, mypaintBrush);
+  MyPaintToonzBrush toonz_brush(m_workRaster, *this, mypaintBrush, true);
   m_lastRect.empty();
   m_strokeRect.empty();
   toonz_brush.beginStroke();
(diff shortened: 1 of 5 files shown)


RodneyBaker:
Thanks for the timely fix @shun-iwasawa 
Much appreciated!

openanim:
@shun-iwasawa I don't know if it's intentional, but in case it hasn't been brought up, there's been a difference in the behavior for creating lines with modifiers on the standard raster level since this fix was merged. Isn't too annoying but just to mention :

Now, the line is affected by the same behavior found on the guide assistant. This gives an effect of very slight offset when u look it from a distance, although I suppose there isn't any. 

Also, a different observation when activating "camera box" from the view menu for color when on the xsheet.

With toonz raster level the level seems to work as expected. But after testing this PR #5039  it seems that it will resume this new behavior after the merge.

https://github.com/opentoonz/opentoonz/assets/29790512/a5715a47-1d18-4505-9aa8-edcc58d58072

blackwarthog:
Hm... I thing it is not shift. Just line pained with two colors - black and white (black at left, white at right). So line will be visible on any back ground. On the white background you see the left part of line.

openanim:
@blackwarthog 

Thanks for the detailed explanations :)

PR #5006, opened 2023-07-09
Fix pane buttons hit area on hdpi

konero:
Caused by me in #4988, on display scales above 100% the pane buttons had an oversized hit area, this fixes it. Thanks @manongjohn for spotting the issue.


RodneyBaker:
Thank you @konero for fix... and the fast turnaround.
LGTM

PR #5024, opened 2023-07-27
Fix 32bit preview

shun-iwasawa:
- Even if `Preview Settings > Channel Width` is set to `32bit Floating Point`, the preview result is not HDR. Only when `Preferences > Interface > 30bit display` is set to ON, the preview result becomes correctly HDR.

The dithering process was incorrectly applied to the 32bpc preview results.


RodneyBaker:
(and apologies for the delay!)

PR #5039, opened 2023-08-07
Assistants for ToonzRasterBrush

blackwarthog:
Implementation of assistants for the ToonzRasterBrush.
Partially solves this issue #1349
Related PRs: #2656, #4852

### DIFF
--- a/toonz/sources/common/tcore/tmathutil.cpp
+++ b/toonz/sources/common/tcore/tmathutil.cpp
@@ -738,6 +738,114 @@ int rootFinding(const std::vector<double> &in_poly, std::vector<double> &sol) {
 
 //-----------------------------------------------------------------------------
 
+int solveEquation2(std::complex<double> *roots, double a, double b, double c) {
+  if (isAlmostZero(a)) {
+    if (isAlmostZero(b)) return 0;
+    roots[0] = -c/b;
+    return 1;
+  }
+
+  double D = b*b - a*c*4;
+  double k = 0.5/a;
+  std::complex<double> d = D < 0
+                         ? std::complex<double>(0, sqrt(-D))
+                         : std::complex<double>(sqrt(D));
+  roots[0] = (-b - d)*k;
+  roots[1] = (-b + d)*k;
+  return 2;
+}
+
+//-----------------------------------------------------------------------------
+
+int solveEquation3(std::complex<double> *roots, double a, double b, double c, double d) {
+  if (isAlmostZero(a))
+    return solveEquation2(roots, b, c, d);
+
+  // x = y - b/(3*a)
+  // y*y*y + p*y + q = 0
+  double p = (3*a*c - b*b)/(3*a*a);
+  double q = (27*a*a*d - 9*a*b*c + 2*b*b*b)/(27*a*a*a);
+
+  double Q = p*p*p/27 + q*q/4;
+  std::complex<double> Qs = Q < 0
+                          ? std::complex<double>(0, sqrt(-Q))
+                          : std::complex<double>(sqrt(Q));
+  std::complex<double> A = pow(-q/2 + Qs, 1.0/3);
+  std::complex<double> B = pow(-q/2 - Qs, 1.0/3);
+
+  // choose complimentary B for A (A*B must be equal -p/3)
+  std::complex<double> rot(-0.5, sqrt(3.0)/2);
+  if (!isAlmostZero(A*B + p/3)) B *= rot;
+  if (!isAlmostZero(A*B + p/3)) B *= rot;
+
+  std::complex<double> Y = (A - B)*std::complex<double>(0, sqrt(3.0));
+  std::complex<double> y0 = A + B;
+  std::complex<double> y1 = (-y0 - Y)/2.0;
+  std::complex<double> y2 = (-y0 + Y)/2.0;
+
+  double dd = b/(3*a);
+  roots[0] = y0 - dd;
+  roots[1] = y1 - dd;
+  roots[2] = y2 - dd;
+  return 3;
+}
+
+//-----------------------------------------------------------------------------
+
+int solveEquation4(std::complex<double> *roots, double a, double b, double c, double d, double e) {
+  if (isAlmostZero(a))
+    return solveEquation3(roots, b, c, d, e);
+
+  // x = y - b/(4*a)
+  // y^4 + p*y^2 + q*y + r = 0
+  double dd = b/(4*a);
+  double p = (8*a*c - 3*b*b)/(8*a*a);
+  double q = (8*a*a*d - 4*a*b*c + b*b*b)/(8*a*a*a);
+  double r = (256*a*a*a*e - 64*a*a*b*d + 16*a*b*b*c - 3*b*b*b*b)/(256*a*a*a*a);
+
+  if (isAlmostZero(q)) {
+    // biquadratic equation
+    // y^4 + p*y^2 + r = 0
+    // handling this special case will give us more accurate results
+    std::complex<double> y[2];
+    solveEquation2(y, 1, p, r);
+    y[0] = sqrt(y[0]);
+    y[1] = sqrt(y[1]);
+    roots[0] = -y[0] - dd;
+    roots[1] =  y[0] - dd;
+    roots[2] = -y[1] - dd;
+    roots[3] =  y[1] - dd;
+    return 4;
+  }
+
+  // solve cubic equation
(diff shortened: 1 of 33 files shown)


shun-iwasawa:
1. GitHub Actions MacOS runner's build failure had been fixed in #5041 which is already merged to the master. Please rebase your commits onto it. In addition, it would be helpful if you could squash the commits into one using `git rebase -i`.
2. When the Pencil option is enabled, the red circle of the brush tip seems to stay around the stroke. Perhaps the size of the `invalidate()`-ed area is not large enough?
image

blackwarthog:
Done.
Thank you Shun!

RodneyBaker:
Something I'm still trying to validate here... and still need to test with the appveyor build as I've only tested with the github action artifact...

In the release version (that supports raster only) if we create multiple assistants either in the same Assistants level or in different levels/columns, we can only draw using one of them.   If in different columns we can hide one and draw with influence of the other... switch which one is on/off and then draw with the other... but not more than one.  If we have multiple assistants it appears only the closest one can be used.

Will test more but wanted to post this in case others who are testing can verify.

I have a few other things to report but those are more general in nature regarding the Assisants and not directly related to this PR.  I'll probably post those other observations in #1349.

RodneyBaker:
Here's a similar setup with two ellipse guides inside one assistants level.
Only the first guide is recognized/usable:

image

This is the same with both toonz raster and raster.
Compare this to nightly where (using raster) all guides can be used.

blackwarthog:
Thanks @RodneyBaker 
I've fixed this bug

RodneyBaker:
Two toonz raster Assistants tests , the first one rather untraditional:

image

andeon:
It's using much less memory now since I last tested it. Don't you plan to change the graphic design of the columns to something like the camera, since it doesn't use the same functions or the filter and opacity could be used soon?

artisteacher:
Getting a crash on open for the action build on macOS.
Screenshot 2023-08-11 at 1 58 57 PM
T2D has the same issue with both nightly and action builds. Haven't tested the nightly here, since I don't particularly want to go through the install process.

andeon:
It introduced an offset bug when drawing in "Toonz Raster Level"

Create a Toonz Raster level, zoom in as far as possible and try clicking precisely in the center of the image (center of the red X "camera box") with the brush tool so you can replicate this issue.

bug

Maybe need to check the pointed regressions to fix it.

konero:
@andeon I haven't time to test this right now but I don't think that's a bug (correct me if I'm wrong) that's just how the pixel matrix is laid out for your output resolution, the center point being (w / 2) and (h / 2) which can't physically be a pixel in a resolution of even numbers, the center dot would need to be 2x2 here. Try increasing your resolution by 1 pixel, for example 1921x1081 to see what I mean.

blackwarthog:
@andeon thank you. I've fixed it

Also i've improved detection of the best guideline for ellipses (when more than one guideline available)

blackwarthog:
Was a bug while compilation for windows. I've fixed it

andeon:
@konero I see, but the problem was that it pushing the brush strokes to another position, but I admit that I could put an image with the 'Pencil' parameter unchecked too to better show how it was working. 

@blackwarthog  It's gone now!

openanim:
1.The track follows 2 guides 

I'm not sure whether this is a bug or a feature, although I'm tending to think it's a bug because its behavior doesn't seem to be uniform on all levels. 

On the raster level, the stroke follows a single guide, but on the TLV and PLI levels (for pli tested from an old PR with the 1st version), the path follows 2 guides at once.

https://github.com/opentoonz/opentoonz/assets/29790512/f00b9f49-07f4-4222-bde8-d9adbf1e5e6c

https://github.com/opentoonz/opentoonz/assets/29790512/c3403e6d-c76d-436a-a42f-2c47ad45c18a

2.problem touch appears on screen

When I use a tool with my stylus, the touch spot appears on the screen, for example here with the selection tool

I don't know if it's my tablet, but it's only visible with this PR. It may also disappear once merged, as was the case with another problem I had that disappeared after the PR was merged with the master.

https://github.com/opentoonz/opentoonz/assets/29790512/faa8ca7a-c9d7-41f6-8810-890734ba58e2


3. Assistant guides when the "grid" option is enabled do not always appear at a certain % zoom. 

triggering seems random depending on the initial reference zoom. I think @konero mentioned this problem too.

4. For some guides, such as the line or ellipse, when using a certain grid mode, the lines are too small and it's difficult to distinguish the location.

It's could be good to use them as measuring lines in a traditional approach, but it would be easier if you could see them better and distinguish them. Either by size, or by a different color.

https://github.com/opentoonz/opentoonz/assets/29790512/8c43b6f3-1a12-492b-887b-7cfd647244e7

Maybe we could have something similar like Krita, where it's easy to distinguish between the big strokes and the intermediate ones. But for now if could just have them appears more clearly, that will be already very helpfull.

https://github.com/opentoonz/opentoonz/assets/29790512/e11abaac-d50a-4482-9f67-88cd60ad1975

andeon:
@openanim Do you use windows artifact to test, right? You could use appvveyor it's gonna be like nightly build.

Downloaded the artifact from "continuous-integration/appveyor/prcontinuous-integration/appveyor/pr" link

Go to **Configuration: Release** > **Artifacts** and download the file "toonz\x64\OpenToonz.zip" 

Extract all and rename OpenToonz Stuff to portablestuff and put it inside the extracted Opentoonz folder.

> - Opentoonz
>   - portablestuff

This will work as a portable version too and in the same way it's merged with the master. So it give you more accuracy to found bugs, maybe some disappear or not.

openanim:
@andeon i confirm im use already the appveyor built as portable like you described it

blackwarthog:
@openanim thanks for your feedback. I've fixed bugs with "assistants mixing" (it is not the feature) and with grid disappearing. Also I've improved ellipse and line ruler marks.

Some note about grids. Grids should disappear when you zoom out too far. If else grid lines will be drawn too often and they will hide the picture behind. So it is a feature when grids disappear on zoom out.

But i've fixed bug with disappearing of line grid on zoom-in.

Also i do not know why a touch bubble appears on the screen and what it means. May be it is related with CI auto-builds only, may you check this behavior on compiled executables from other PRs?

openanim:
@blackwarthog 

Thanks  a lot for the changes and explanations. Concerning the problem of the touch on the screen, I tried with another PR this one "fix use default constructor instead" and the problem is present. I wasn't able to test an older artifact to see if it had any impact. But it could be solved once merged, as is sometimes the case.

Otherwise, sometimes there's a stroke delay at times with some guides, it's easier to see when you go slowly.


https://github.com/opentoonz/opentoonz/assets/29790512/9fbda04d-bfc7-481b-bfe7-19846f7e46db

blackwarthog:
Program needs to gather approximately 20 pixels of mouse path to make a decision about chosen guideline. But there was a bug and this path was much longer. I've fixed it now.

RodneyBaker:
I'll add this here although it's not directly related to toonz raster implementation of Assistants.

The problem:
When we attempt to animate an Assistants level there is no visual of the Assistant or Guides that can tell us where  we are moving things to so we are operating blind.  We could draw something and attach it as a child of the Assistant and that would allow us to see something but otherwise we just have the center point of the column to guide us.

This is also the case when we assign a motion path to an Assistants column as we use Animate Tool to adjust positioning on the path. (It is VERY cool that we can animate Assistants and assign motion paths to them however!)

Here's a video that I hope demonstrates the problem:

https://github.com/opentoonz/opentoonz/assets/2576432/7d6bef15-4f30-4af3-a54c-56ee922bab6c

blackwarthog:
@RodneyBaker thanks. I understand. I think we need to draw assistants with any tool selected. It may cause some new bugs or artifacts. Will be better to implement this feature in the next PR (together with the toonz vector brush assistants).

blackwarthog:
Hm... Is all right? Why are we not merging this PR?

RodneyBaker:
> Hm... Is all right? Why are we not merging this PR?

In addition to not knowing if you were finished adjusting...
@shun-iwasawa has this self assigned so as a courtesy we are giving him time to review the PR and its changes.

blackwarthog:
Thanks. Ok, If is no actions from me required i'll be wait patiently.

RodneyBaker:
Posting intent to merge.
Additional fixes and updates can be found in subsequent PR #5066.

RodneyBaker:
Thank you @blackwarthog for bringing us to this point with the Assistants.
Very much appreciated.

PR #5041, opened 2023-08-08
Actions: fix MacOS runner removing link to qt6

shun-iwasawa:
This PR will resolve recent failure in GitHub Actions MacOS runner.
The error message was `"Qt requires a C++17 compiler"`. This occurs when Qt6 and Qt5 are installed at the same time and the two are mixed up.
After investigation, the vtk package in Homebrew was updated on August 1 so that it now depends on Qt6.
Since vtk is used by OpenCV, it is assumed that Qt6 is to be installed when installing OpenCV via Homebrew.
This PR removes the link to Qt6 when installing the library.

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -37,8 +37,11 @@ jobs:
           done
         }        
         brew update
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thank you @shun-iwasawa!

pravdomil:
this needs to be also reflected in https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_build_macosx.md

PR #5042, opened 2023-08-08
Fix multi-monitor icons

konero:
Ports some fixes by @manongjohn (thanks!).

When using multiple monitors with different scale factors, icons were only being generated for the primary size and weren't showing for other sizes and fixes active icons hover states not showing icons.


RodneyBaker:
LGTM

PR #5047, opened 2023-08-15
fix color screentones

andeon:
Change the screentone color to match opentoonz default themes like neutral.

screentonefix

Set only 'Use As Pattern' in 'Settings' or adjusting the contrast should be enough to create a screentones effect most of the time, mixing them gives more options.

ref: https://github.com/opentoonz/opentoonz/pull/4911


RodneyBaker:
Will press forward with this as it isn't changing any code... just updating the original screentone images.
Thank you @andeon!

PR #5049, opened 2023-08-17
Fix Use default constructor instead

otakuto:
Fix this warning.
```
/usr/include/x86_64-linux-gnu/qt5/QtCore/qflags.h:123:80: note: declared here
  123 |     QT_DEPRECATED_X("Use default constructor instead") Q_DECL_CONSTEXPR inline QFlags(Zero) noexcept : i(0) {}
      |                                                                                ^~~~~~
```


RodneyBaker:
Thanks for the fix @otakuto 
LGTM

PR #5069, opened 2023-08-28
Brush strokes replication feature

blackwarthog:
As continue of assistants idea here i've implemented a multi-painting feature.

Most code for was already done while implementing of assistants.

Related PRs #2656, #4852, #5039 and issue #1349

Here is the demo:

https://github.com/opentoonz/opentoonz/assets/5143743/7f91065e-b73c-4e4c-be58-3e4205c66c61

Notice that order of replicators is significant. I've tried to show the difference it in the demo.

The "Jitter" adds noise to brush strokes, and may keep specified number of first tracks unchanged. With this feature allows you make copies different. Order of Jitter is also significant.

The "Affine" replicator is a custom power full feature that can replace Mirror, Grids and Star functionality. But it a bit difficult to configure.

### DIFF
--- a/toonz/sources/include/tgeometry.h
+++ b/toonz/sources/include/tgeometry.h
@@ -202,6 +202,16 @@ inline TPointD normalize(const TPointD &p) {
 
 /*!
 \relates TPointT
+This helper function returns the normalized version of the specified point
+or zero if it is not possible
+*/
+inline TPointD normalizeOrZero(const TPointD &p) {
+  double n = norm2(p);
+  return fabs(n) > TConsts::epsilon*TConsts::epsilon ? p*(1/sqrt(n)) : TPointD();
+}
+
+/*!
+\relates TPointT
 This helper function converts a TPoint (TPointT<int>) into a TPointD
 */
 inline TPointD convert(const TPoint &p) { return TPointD(p.x, p.y); }
@@ -1426,9 +1436,9 @@ public:
     inline Range range() const
       { return Range(a0(), a1()); }
     inline int size() const
-      { return (m_end - m_begin)/2 + 1; }
+      { return (int)(m_end - m_begin)/2 + 1; }
     inline int index() const
-      { return (m_current - m_begin)/2; }
+      { return (int)(m_current - m_begin)/2; }
     inline int reverseIndex() const
       { int i = index(); return i == 0 ? 0 : size() - i; }
     inline bool lapped() const
--- a/toonz/sources/include/tmetaimage.h
+++ b/toonz/sources/include/tmetaimage.h
@@ -181,8 +181,10 @@ protected:
   virtual void onFixData() { }
 
 public:
-  void setDefaults()
-    { onSetDefaults(); }
+  void setDefaults() {
+    { LockEvents lock(*this); onSetDefaults(); }
+    data().touch();
+  }
   void dataChanged(const TVariant &value)
     { if (m_locks == 0) onDataChanged(value); }
   void fixData()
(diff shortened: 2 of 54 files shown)


RodneyBaker:
I've been having a lot of fun with this.

image

blackwarthog:
@RodneyBaker wow!

blackwarthog:
@beeheemooth I'm already working in it. I've planned to create additional assistants - full perspective (three vanishing points in one), full fish eye (two elliptical grid) and cylinder (an advanced ellipse assistant)

blackwarthog:
> But what do we do with drawing patterns in a spiral, drawing in reverse perspective, and random jitter in fish-eye space?
> (ok, I think you got it, it was a joke)

Actually all of these are in OpenToonz already

RodneyBaker:
There is no way all the options of this feature set can be documented short of the person putting the Assistant together demonstrating their setup and process.

I understand the need that @beeheemooth outlines but the same might be said of complex features such as Shaders... or LUTs.   New users won't understand them right?  No, and neither will advanced users until they dive in and explore.

There are features in here that certainly won't be documented because they aren't intended parts of the feature set; example using this feature to create on the fly custom brushes...  where is the documentation for that?  (Answer:  In my pea sized brain)

@beeheemooth mentioned the jitter etc... and I was rather excited to find the magnetism settings that allow us to break free of the assistant guides.  That produces some excellent opportunity to draw with brush feel much like Lazy Nezumi etc.  Conforming to the guides but allowing us to move away from their influence as we desire.  Nice.

As with Opentoonz itself it will take considerable time to document Assistants and while I'm certain the basic will be well documented I doubt they can ever be fully documented.  THAT is why I find it so excellent that these Assistants are Level types.  Level Types that can be shared and easily examined in a text editor.  (My turn to say "Wow!")

Keep up the excellent work.
And make no mistake, we really do appreciate your documentation!  :)

Here's a guy that is eager to expore more Assistants:
mirror

RodneyBaker:
A feature I would love to see added to Assistants would be a reference to an adjacent vector level to use as a guide.
This could be a motion path but I believe it would be better suited as an easily editable vector level in a nearby column.
Some code for that adacency is likely in features such as TextIwa FX.

What that would allow the user to do is quickly draw a guide (even store it for later use) and call it up where needed.
Then Opentoonz users could create tools such as french curves, custom shapes, etc. and that would in turn drive the assistant.

Adding this as an Assistant would be good also in that a new separate feature for using vector guides would not need to be implemented somewhere down the line.  It can instead be a part of the current Assistants feature set.

Note that I believe this might be limited to one single vector stroke at the outset.
This because currently there is no way for OT to internally consider more than one stroke as being continuous.
This is something that needs to be addressed in time but far outside the scope of any Assistants PR.
The example case for this might be a motion path consisting of multiple vector strokes where all those lines are considered continuous for the sake of motion but can be selected (and presumably grouped) independently.


At any rate, an Assistant option that referenced a known vector line whether on disk, in column or elsewhere would elevate the customization capability of Assistants by the user  considerably.  

Certainly a tall order but something to ponder along the way and if/where the code does fall into place...  wala!

RodneyBaker:
Posting intent to merge...
@blackwarthog Do you have any additions or modifications planned for this this PR?
My understanding is that we are ready to move on to the next phase/PR.

blackwarthog:
Yes, this PR is complete and i think we may merge it.

RodneyBaker:
Thank you @blackwarthog for your continued work with the Assistants Tool.
The brush strokes replication feature is a very useful addition.
LGTM

PR #5071, opened 2023-08-29
Improve assistant tool selection

blackwarthog:
I've changed behavior of tool selection to allow to make aliases for tools (via TTool::bind)

Now tool can be assigned to multiple toolbar buttons (actions)
We already had ability to assign different tools to a single tool button (for example, three different brush tools share the same button)

I've assigned the EditAssistantsTool to the "brush" button.
And now for editing of assistants you do not need to switch the tool  (while drawing). You just need to choose a cell in Xsheet/timeline.

Also i've made more detailed message about level/tool imcompatibility.

msg

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -288,7 +288,7 @@ public:
 
     EmptyTarget = 0x80,  //!< Will work on empty cells/columns
 
-    MetaImage   = 0x100, //!< Will work on mets images
+    MetaImage   = 0x100, //!< Will work on meta images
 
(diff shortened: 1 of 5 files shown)


blackwarthog:
This PR in independent from other my PRs and can be merged before or after them

blackwarthog:
ok

RodneyBaker:
Posting intent to merge.  
I want to run a few additional tests but also want to give a timely heads up.

JParks123:
Would it be possible to add a way to change the colors of the Assistant perspective rulers? because when building Perspective rulers with multiple Assistants I find it pretty difficult to differentiate which Assistants ruler u are using.

PR #5083, opened 2023-09-04
Update sceneviewerevents.cpp remove identical include

maddin200:


RodneyBaker:
I'm going to press forward with this one as it is obvious.
The fail in Mac checks is apparently a build issue that will need to be addressed (assuming it hasn't been already).

Thanks @maddin200 for the PR and identification of other typos and nitpickables.
LGTM

PR #5093, opened 2023-09-11
Fix build on FreeBSD 13+

rozhuk-im:
Small fixes to build on modern FreeBSD.


RodneyBaker:
Will press forward with this PR as self explanatory.
Thank you for the fix/update for building on FreeBSD 13+ @rozhuk-im!

PR #5100, opened 2023-09-17
Improve assistants

blackwarthog:
To improve usability of assistants i've added two "combo" assistants: "perspective" and "fish eye".

The perspective assistant is a combination of three vanishing points.
Fish eye is a combination of two ellipses and one vanishing point.

These things make creation of an assistants environment much easy.
I've attached videos with demonstration of new assistants.

https://github.com/opentoonz/opentoonz/assets/5143743/d07707e2-a42b-4831-9949-c60abe5ec5f3

https://github.com/opentoonz/opentoonz/assets/5143743/02848223-8f91-43f6-beb9-a57938845b44

Also i've added two new options to each replicator and to the jitter: "skip first track" and "skip last tracks".
These options allow to exclude some tracks from replication/jittering.

And new options has been added to the jittering: "fix first point" and "fix last point".
These options allow to exclude first and/or last points of a track from jittering and keep them at place. Jittering will fade smoothly at tails.

### DIFF
--- a/toonz/sources/common/tgeometry/tgeometry.cpp
+++ b/toonz/sources/common/tgeometry/tgeometry.cpp
@@ -61,9 +61,8 @@ TAffine TAffine::inv() const {
         (a22 == 0.0 ? std::numeric_limits<double>::max() / (1 << 16)
                     : 1.0 / a22);
     return TAffine(inv_a11, 0, -a13 * inv_a11, 0, inv_a22, -a23 * inv_a22);
-  } else if (a11 == 0.0 && a22 == 0.0) {
-    assert(a12 != 0.0);
-    assert(a21 != 0.0);
+  } else
+  if (a11 == 0.0 && a22 == 0.0) {
     double inv_a21 =
         (a21 == 0.0 ? std::numeric_limits<double>::max() / (1 << 16)
                     : 1.0 / a21);
@@ -235,8 +234,79 @@ TScale::TScale(const TPointD &center, double s) {
 
 //==================================================================================================
 
-TPoint4D TAffine4::operator*(const TPoint4D &b) const {
-  return TPoint4D(
+T3DPointD TAffine3::operator*(const T3DPointD &b) const {
+  return T3DPointD(
+    b.x*a11 + b.y*a21 + b.z*a31,
+    b.x*a12 + b.y*a22 + b.z*a32,
+    b.x*a13 + b.y*a23 + b.z*a33 );
+}
+
+TAffine3 TAffine3::operator*(const TAffine3 &b) const {
+  return TAffine3(
+    *this * b.rowX(),
+    *this * b.rowY(),
+    *this * b.rowZ() );
+}
+
+TAffine3 TAffine3::operator*=(const TAffine3 &b)
+  { return *this = *this * b; }
+
+TAffine3 TAffine3::inv() const {
+  TAffine3 r;
+  r.a11 = a22*a33 - a32*a23;
+  r.a12 = a32*a13 - a12*a33;
+  r.a13 = a12*a23 - a22*a13;
+
+  double det = r.a11*a11 + r.a12*a21 + r.a12*a31;
+  det = fabs(det) > TConsts::epsilon ? 1.0/det : 0.0;
+  r.a11 *= det;
+  r.a12 *= det;
+  r.a13 *= det;
+
+  r.a21 = (a31*a23 - a21*a33)*det;
+  r.a22 = (a11*a33 - a31*a13)*det;
+  r.a23 = (a21*a13 - a11*a23)*det;
+  r.a31 = (a21*a32 - a31*a22)*det;
+  r.a32 = (a31*a12 - a11*a32)*det;
+  r.a33 = (a11*a22 - a21*a12)*det;
+  return r;
+}
+
+TAffine TAffine3::get2d() const {
+  return TAffine(
+    a11, a21, a31,
+    a12, a22, a32 );
+}
+
+TAffine3 TAffine3::translation2d(double x, double y) {
+  TAffine3 r;
+  r.rowZ().x = x;
+  r.rowZ().y = y;
+  return r;
+}
+
+TAffine3 TAffine3::scale2d(double x, double y) {
+  TAffine3 r;
+  r.a11 = x;
+  r.a22 = y;
+  return r;
+}
+
+TAffine3 TAffine3::rotation2d(double angle) {
+  TAffine3 r;
+  double s = sin(angle);
+  double c = cos(angle);
+  r.a11 =  c;
+  r.a12 =  s;
+  r.a21 = -s;
+  r.a22 =  c;
+  return r;
+}
+
+//==================================================================================================
+
+T4DPointD TAffine4::operator*(const T4DPointD &b) const {
+  return T4DPointD(
     b.x*a11 + b.y*a21 + b.z*a31 + b.w*a41,
     b.x*a12 + b.y*a22 + b.z*a32 + b.w*a42,
     b.x*a13 + b.y*a23 + b.z*a33 + b.w*a43,
@@ -288,6 +358,13 @@ TAffine TAffine4::get2d(double z) const {
   return TAffine(
     a11, a21, z*a31 + a41,
     a12, a22, z*a32 + a42 );
+}
+
+TAffine3 TAffine4::get2dPersp(double z) const {
+  return TAffine3(
+    T3DPointD( a11       , a12       , a14       ),
+    T3DPointD( a21       , a22       , a24       ),
+    T3DPointD( a31*z+a41 , a32*z+a42 , a34*z+a44 ) );
 }
 
 TAffine4 TAffine4::translation(double x, double y, double z) {
--- a/toonz/sources/include/tgeometry.h
+++ b/toonz/sources/include/tgeometry.h
@@ -26,7 +26,8 @@ inline double logNormalDistribuition(double x, double x0, double w)
 
 //=============================================================================
 
-template <class T> class TPoint4T;
+template <class T> class T3DPointT;
+template <class T> class T4DPointT;
 
 /*
 * This is an example of how to use the TPointT, the TRectT and the TAffine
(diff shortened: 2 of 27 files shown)


blackwarthog:
Related PRs #2656, #4852, #5039, #5069 and issue #1349

blackwarthog:
I've fixed windows compilation bug. Now all compilers work fine

RodneyBaker:
This is looking very good and... the feedback I'm seeing thus far from others is very positive as well.
Nice work!

There is one aspect of this particular improvement... specifically the perspective assistant improvement that I hope can be further expanded a little more and it relatest to the view that we get while setting up the box for perspective.

image


This box is such a great tool in and of itself for helping artists get a proper sense of the space in which they are drawing that I hope a means to be able to see the box while drawing can be added.  To my mind this really cements that space which the drawn object will occupy all the way and it also emulates many instructors approach to actually drawing the box (outlining at least) prior to committing to drawing inside that space.

Random image grabbed from the internet to demonstrate that 'visible' box around the character:

image
Image copyright the original artist

blackwarthog:
@RodneyBaker I've added this option

RodneyBaker:
Not a particulary good example but... wanted to try the change.
Here's I've tested having two perspective boxes in the assistants level but then duplicated it by exposing it again in a new column where it is moved to the side a little to add new horse and rider.
The Assistants are off in the Brush Tool for this as I'm just testing how well the boxes provide a general guide to drawing inside of the box spaces.


https://github.com/opentoonz/opentoonz/assets/2576432/7af6f4dc-3d15-4b8d-b857-11ef96aa6da9



I've got some crazy ideas to test but... first some rest.
Thanks much for improving the Assistants.

Ahl76:
Thanks for all the improvements (I'm formerly openanim's account, by the way). Here's a feedback at some of the tests I've done, and my feelings on a few points :

1) The selection improvement is really nice ! And in addition to that, I still think it would be better 

- To leave the last assistant guide active, since if we have to change it, it won't change the current process, but if we have to use the same one several times in a row, it will save us time.

- To have an option to auto select the assistant tool when we're on an assistant level.

2) Modify column selection behavior.

This request isn't actually related to this PR but a problem with opentoonz itself except that playing with assistant guides highlight it when you use multiple assistant columns like I usally do.

Honestly, I don't know of any other software with OT behavior that seems so counterproductive for me in this way. The column should only change when you click on the column body itself and not the attributes/properties of its header. Note that Harlequin neither dont have this behavior. So maybe it's something intentional but i'll dont really understand why in this case.

But again, not a problem directly linked to this PR so you can just as well ignore it although in my opinion it would make the use smoother and more common to any software.

https://github.com/opentoonz/opentoonz/assets/145405932/76aee78a-487f-400c-af23-3ffdb567deef

3) Smooth has an unusual behavior on this PR as you can see with or without the use of an assistant level.

https://github.com/opentoonz/opentoonz/assets/145405932/8cea384e-d52c-4218-a2d7-50602db770b5

4) Difficulty in fine-tuning certain guides, notably the measuring line (with many little lines) or the angle with certain replicators such as affine. The sensitivity is too high in my opinion. Perhaps using a modifier can make it less sensitive ?

https://github.com/opentoonz/opentoonz/assets/145405932/c1a3656a-7ce3-47e1-8837-286f02cdf0c5

5) In my opinion, when a guide is selected for the 1st time, it shouldn't shift and stay in place (I systematically use the undo key after selecting a guide because of this behavior). And On the menu side, I find it more common to add on the right and subtract on the left same for symbols + an -. At present, it's the opposite. (eg count +- ..inver .count +-....etc).

https://github.com/opentoonz/opentoonz/assets/145405932/d134c8ae-09ba-4e88-89d1-df2011171ea7

6) The perspective guide is very interesting, some observations :

- I don't think it can completely replace a preset interface, although it does make it easier to set up common guides, so really hope we could have an interface for that in future as well.

- Compared with the krita version, what im missing is the ability to delimit/restrict the grid to a specific area. whether for the vanish point guide or the perspective guide and increase/decrease lines. Maybe something like "restrict" option ?

- Difficult to use when setting up from a model or sketch, especially for auto VPs.
I find that is useful feature for automatically transcribing perspective lines based on a sketch, image or model, on the same principle as the vanish point guides allow to do.

https://github.com/opentoonz/opentoonz/assets/145405932/5eea1968-8771-42c1-a704-734a7dc3b022

The problem I've encountered is the difficulty of being precise. I have the feeling that a 4th point could help stabilize angles ? Or i dont fullyunderstood his behavior because that make me time to setup currently.

https://github.com/opentoonz/opentoonz/assets/145405932/e7786533-97ab-49c8-96c7-95df4687a029

8) Perspective ellipse guide

As perspective guide this will be another usefull guide to have 

https://github.com/opentoonz/opentoonz/assets/145405932/d27deac1-e93d-4410-ab38-b9cdc66093b4

9) have the various functions and guides available as shortcuts or in the custom panel

10) Custom panel are powerfull and will help to speed up selection if we have entries for them . I feel like guides are very oriented to this kind of use.


11) In future I'd like a way to trace paralell in fly. Maybe a "ruler" version as an independent tool from assistant tool (we already have a ruler tool, but it's not very useful since we cannot trace from this) to quickly draw parallel lines in fly, using only a parallelepiped with 4 points that can also retract or extend.

I've tried to provide many details as possible in relation to my workflow, but please feel free to consider these requests at your own convenience. And thanks for the promising work already done!

blackwarthog:
@Ahl76  thanks for your feedback!

1. Hm... OpenToonz already remember last selected assistant. Also since PR #5071 you don't need to switch between brush and assistant tool - just edit assistants with the brush tool selected.

https://github.com/opentoonz/opentoonz/assets/5143743/ab494f86-0d39-4fe8-b4ee-31e87527ad43

2. I've updated this PR to disable column switching when clicking on "lock", "eye", "camera stand" and "configure" column buttons. Also i've switched + and - buttons in spinner properties. @RodneyBaker I hope these changes was not significant for the reviewing process.

3. It's a feature. In some cases we cannot paint the track until we know points from the future. Smoothed track is one from these cases. Also spline interpolation of track requires to know one point from the future. Assistants also need to accumulate several points to make decision which guideline user want to use in a current stroke. So in these cases track is painting with a delay.

With the vector brush you may see the final look of the stroke immediately (without any delays). We can implement this feature for raster brushes too when they will be fast enough.

5. I'll think about it. May be we can do some kind of "fine tune mode", for example with dragging with shift pressed reduce drag distance by ten times. But may be we will better use shift and control modifiers for other purposes.

Also you don't need to rotate the affine replicator, just moving the origin will be enough.


https://github.com/opentoonz/opentoonz/assets/5143743/e0812c2f-42a7-4eba-af2e-7ae37059d881

6. I've switched + and - (see point 2). I'm not sure that we need to disable point dragging on the first click. It may be confusing. I think we need more feedbacks about it.

I don't understand what the "presets interface". Did you mean dedicated vanishing points with 4 helping handles?
I'll think about the grid restrictions. The current implementation of infinite grids have no place to add such restrictions quickly. In you example you've chosen a wrong origin point for the perspective. You need to choose a point where all three axis (x,y,z) connect each other.

https://github.com/opentoonz/opentoonz/assets/5143743/47611099-2422-48a9-a505-585daeb7d70a

7. We have no the point number 7 :)

8. I understand, i'm still working on ellipses improvement

9-10. I don't know where a custom panel in OpenToonz and how to use it... If we need to implement it, i think it will not be in near future.

11. I don't understand... heh. May you try to use the line-assistant with the "parallel" option checked, or perspective assistant with all the "parallel" options checked?

Ahl76:
@blackwarthog Thanks for tips  and sorry for the long post !

1. Regarding point 1, my suggestion was not about the recent and very welcome improvment to the selection, but about the fact that you can add the last selected guide several times in a row without having to search for it again from the dropdown assistant type menu..

https://github.com/opentoonz/opentoonz/assets/145405932/a9f6c5d2-1371-492d-9f67-cd27568cd955

2. That's great! I also hope it won't be a problem. Maybe separate it in another pr if it requires shun iwasawa's approval, but frankly this change solves a lot of annoying situations. Thanks ! Hope could be merged whitout problem.

3. About smooth feature Isn't it possible to have this behavior only when using assistants, or to have 2 types of smooth, a normal smooth and an assistant smooth ? Smooth is also very useful for making inking easier. The workaround seems to be to use a low smooth value to minimize the current effect behavior.

4. In your 5th point, I think you've answered my 4th point and yes not necessary to be a modifiers :).
I have the similar situation using the "measuring line" with many littles lines. When I need to stop my cursor at a very precise point/position I have to fumble a bit... But nothing insurmountable in itself. And your explanation and tips gave me better results for the Affine guide part, Thanks !

https://github.com/opentoonz/opentoonz/assets/145405932/8e1670c7-e760-476b-948c-3f0aba74e63e

By the way, is this possible already currently to Make my affine assistants react like a vanish point ? What i mean is not to follow my other vanish point guides but to be himself like a vanish point ? not sure if it's understandable : 

https://github.com/opentoonz/opentoonz/assets/145405932/6fa3a175-cc2f-432a-aa96-290e7c1a8bdd

I just noticed that the +- inversion was already present in OT. And that make think your initial implementation was more consistent in this respect, although I think it was more natural in the current layout with the change. I guess need more feedback about that too like could be confuse things.

Like  "the point dragging on the first selection", to have more feedback on this, but I just want to make sure we're talking about the same thing: here's another example: each time I select a guide, for example, if I want to change its parameters on my current layout, I'll have to undo it after selecting any guide.

https://github.com/opentoonz/opentoonz/assets/145405932/8283a970-3fad-45e1-9a7e-e13d9358541e

What I meant by "preset interface" was the ability to save our own "models guide" with quick access. We can already save them and reload them, but I'm thinking of something like the current brushes preset for fast and load, access. 

https://github.com/opentoonz/opentoonz/assets/145405932/5c26c2db-4ed2-4872-9257-6063f1e9e5c8

What I meant by custom panel is the recent functionality to create panel layout with existing commands. It's more like make the differents guides available for them in OT like this:

https://github.com/opentoonz/opentoonz/assets/145405932/3214c5f1-c1f6-417c-b370-c27efe24d532

Regarding the idea for drawing parallel lines.  At the moment I'm already following the parallel lines approach from assistant guide, but sometimes I don't feel the need to reuse the guide and it overloads. That's why I was advocating in the future for a perspective paralelle ruler tool separate from the assistant tool, to be able to draw parallel lines in the fly whitout setup them.

Although maybe there's a better solution, but the idea would be to have a 4-point ruler that can be extended or shortened in the same way that your perspective tool but only with 4 point like a parallelepiped's face. and that allows us to trace on it like a ruler.

Here's an example, although not exactly as I've described it, but inspired by it.

https://github.com/opentoonz/opentoonz/assets/145405932/69c2b1ec-0bde-435c-bf2c-17cd322808d8

But with some properties from the perspective tool

https://github.com/opentoonz/opentoonz/assets/145405932/3fb693b0-feb2-42ee-b833-3914cafdac68

So yes currently workaround could be use perspective guide.

Note that all these ideas are just suggestions on my part ! Otherwise very excited by the perspective ellipse !

blackwarthog:
1. I dont thing that batch creation of assistants is very useful. You may use perspective assistant instead of three vanishing points. I think that unwanted creating of new assistant will be more bugging.

3. Smooth feature was not been changed. You just now see a full stroke earlier while painting. In previous version you was see the white void instead of line. Now you see the line in form exactly the same to the stroke "finalized" in current mouse position. The final results of "old smooth" and "new smooth" are the same. You may compare with Toonz Raster Brush Tool.  I don't know but for me a seeing to a stroke tail faraway from cursor is more confusing than seeing a stroke with "flexible" tail.

4. Sadly but affine transform cannot simulate the perspective behavior. It is math. We need special 3d grid tool that i haven't made yet.

"the point dragging on the first selection". Yes we talk about the same thing. I'll recheck. The point should not move by simple click without cursor jittering. But i thing that user usually expecs that point will be draggable immediatelly since first touch.

Preset interface. Assistant level is just a file and you free to load it into any project in any column and in any frame. May be we can simplify the procedure of adding existing levels from files. I don't know.

Custom panels. Ok, i know about them for now. Thank you :) What commands we can add where? Hm.. "activate the assistants tool in mode: create vanishing point"... Looks a bit tricky... may be convert drop down box into toolbar with small button for each assistant... I don't know the graceful solution for this yet.

"Parallel lines". We need to add the some handles from assistants into brush tool. It can be useful for some features, but here i also do not know the good solution.

blackwarthog:
@RodneyBaker what do you think about [-] [+] buttons. Should we to swap them back to common OpenToonz order (+,-)?
Also i can create the separate PR for column selection fix if it need.

RodneyBaker:
> @RodneyBaker what do you think about [-] [+] buttons. Should we to swap them back to common OpenToonz order (+,-)? Also i can create the separate PR for column selection fix if it need.

Hmmmm... first impression... without delving too far into the question I'm personally in favor of having the plus on the right side... minus on the left.  

It's definitely beyond the scope of this PR to fix all the other instances in Opentoonz where the order is reversed however.

Now you've got me curious if this is a Studio Ghibli change or if that order is reversed in Toonz software as well.
Off hand, I can't think of too many cases in software where +, - is in that reversed order.

For now you could leave a comment in the code to suggests it corrects the order of other cases in Opentoonz.

blackwarthog:
@RodneyBaker  I've created separate PR for columns selection fix #5108. And also i've added comment in code about a spin-buttons sequence.

RodneyBaker:
I'm confident in this PR moving forward but a bit apprehensive the nightly will build and propagate to the nightly listing as the past few PRs didn't succcessfully make it (for Windows).

As such I'm a bit hesitant to merge at this time.

But otherwise... posting intent to merge.  :)

Ahl76:
After checking, the +- inversion is also in Toonz Harlequin, so it doesn't seem to be linked to a Studio Ghibli wish...For the inversion cases I've seen (brush tool, plastic tool, gizmo palette).

@blackwarthog 

About the smooth, I agree that it's better to see our line without delay, but what bothers me a little (compared to before) is more in the curves when I go fast, it tends to shorten. When I go slowly, it's better. My eyes may just have to adapt.

"the point dragging on the first selection".  I've tested it with the mouse and I don't have any dragging. I think it's not a problem specific to this PR but to selection with a stylus. The same thing happens with the hook tool. So just ignore it.

I like the approach to make parrallel from the brush without another intermediate tool that could be make things faster, although I can't really see how it could be set up. We can let the question mature for now, i guess.

For custom panels, I suppose the simplest solution would be to have the commands/shortcuts for each guide (vanish point, line, ellipse, etc.) available in the custom panel editor and use one of the existing panels. Although this PR #4495 could help with the creation of a more customized layout, I don't know where the commands available in the custom panel are added ?

By the way Thanks a lot for the PR for the selection fix and its merge, very nice :) !

RodneyBaker:
@Ahl76 You may want to transfer some of your thoughts to the main Assistants topic so they don't get buried in this PR once it gets merged which is about to happen.

Posting intent to merge.

RodneyBaker:
Thank you @blackwarthog for this latest round of improvements to Assistants.
Much appreciated.

PR #5108, opened 2023-09-24
Tune column selection

blackwarthog:
This PR prevents selection of a column when clicking on the "eye", "camera stand", "lock" and "options" buttons of the column. 

See also discussion here https://github.com/opentoonz/opentoonz/pull/5100#issuecomment-1728063568

### DIFF
--- a/toonz/sources/toonz/xshcolumnviewer.cpp
+++ b/toonz/sources/toonz/xshcolumnviewer.cpp
@@ -2343,10 +2343,7 @@ void ColumnArea::mousePressEvent(QMouseEvent *event) {
       TApp::instance()->getCurrentXsheet()->notifyXsheetChanged();
       return;
     }
-    // set the clicked column to current
-    else
-      m_viewer->setCurrentColumn(m_col);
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Going to press forward with this change.
Thanks @blackwarthog for the modification.

beeheemooth:
@blackwarthog , since you touched on this issue, please make it possible to turn off the visibility of the current column without header highlighting, so that you can turn it on and off using a shortcut. Please!

ezgif-1-052ea76e03

blackwarthog:
@beeheemooth i've created PR #5109

beeheemooth:
@blackwarthog Oh man! Thank you so much! 
I believe that these three lines of code will help us! If only you knew how difficult it is to make sure which column you are in in order to draw (after all, a column in an Xsheet can often be out of sight)! The only way to do this is to toggle the visibility of the current column. But in order to do this, you need to find it and select it. But what was the point of switching if you found it? hehe.

Thanks!

PR #5136, opened 2023-10-19
Update toonzqt.ts

Roosterington:
Added some translations into French.


RodneyBaker:
Will merge this as it shouldn't be effected by the appveyor checks.

Thank you @Roosterington for the French language update!

Issue #5002, opened 2023-07-08
Problem copying values from the Style Editor context menu


beeheemooth:
### Description

The HEX code in the input window for a color cannot be copied from the context menu, because when the context menu is called, the focus of the selection is lost.

Calling the context menu with the system button "Menu" also leads to loss of focus.

### Steps to Reproduce

1. Highlight the code in the input window
2. Call the context menu with a right click
3. Code highlighting will disappear, copying is impossible

### Expected Behavior

Copying from the context menu in my case is the only way to copy, since I use a macro for Ctrl + C, which at the same time introduces extraneous characters into the field.

### Screenshots, Video & Crash Logs

![ezgif-1-7abb783bc5](https://github.com/opentoonz/opentoonz/assets/25724133/f9d32b55-6c9c-43cd-8da3-a7b3d1df78ad)

### OpenToonz Version

1.7

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Huion

RodneyBaker:
I can replicate the problem.
As you've stated a workaround is to use Control + C to copy the values.
It is unfortunate that you can't use that standard shortcut.
Perhaps you can share your Autohotkey script and we can find a way to allow both your customized use and the standard copy/paste.

The bug itself appears to be (as you suggest) a loss of focus (or failure to release focus).
The problem exists in Tahoma2D as well.

RodneyBaker:
Closed by #5185

PR #5185, opened 2023-11-21
Fixed deselect of hexlineedit after right click

Funinja:
This PR attempts to fix issue #5002, loss of focus on right click of the hex code in the style menu. 

I believe its due to the shift of focus from the style editor to the right click menu. To fix this I just disabled deselect upon editing the hex code.


Funinja:
It seems like I made a mistake for the windows build, but the windows build failed its vcpkg download which I don't think is related to my change. I'm new to contributing to this repo, so maybe I made a mistake, is there anything I can do to fix this?

RodneyBaker:
It may be that for the WIndows automated github actions build you have to be whitelisted to be included in the build.
We'll may need to invoke @shun-iwasawa to confirm that however.

RodneyBaker:
You are now added to the whitelisting @Funinja so PRs should now correctly process.
Thanks for the fix to the hexlineedit
LGTM

Issue #4887, opened 2023-05-11
Crash when holding Ctrl+Z while drawing.


beeheemooth:
### Description

I discovered this crash while testing #4885, but the crash can be reproduced in earlier versions.
In Tahoma 2D, such a crash is not observed.

### Steps to Reproduce

1. Create a new vector level.
2. Start drawing a line (keep the stylus on the canvas)
3. While drawing, press and hold Ctrl+Z
4. At the end of the cancellation of all operations, the program will crash.

(btw, cool crash window! I haven't tested OT in a while)

### Screenshots, Video & Crash Logs

![ezgif-2-990963defa](https://github.com/opentoonz/opentoonz/assets/25724133/42dd713e-2f5f-4be9-8b5a-1ee401283169)

### OpenToonz Version

1.7

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Huion

RodneyBaker:
I can't reproduce the crash.
Perhaps I'm missing an important step however.

RodneyBaker:
@beeheemooth @Bracket-H Can you confirm that #5187 actually fixed your problem?
I can't produce the crash but I couldn't last year either so it'll be good to get a confirmation either way.

RodneyBaker:
Will work with that then.  Thanks
@beeheemooth Please let us know if you still have thie problem and we'll reopen.

PR #5187, opened 2023-11-23
Fixed crash when holding Ctrl+Z while drawing.

Funinja:
This PR fixes issue #4887, crashing after creating layer and holding Ctrl+Z while drawing.

To fix this, I just deactivated the tool before the undo of the Ctrl+Z. Maybe it's better to just disable Ctrl+Z while drawing, any thoughts??

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -936,13 +936,11 @@ Room *MainWindow::getCurrentRoom() const {
 void MainWindow::onUndo() {
   ToolHandle *toolH = TApp::instance()->getCurrentTool();
 
-  // end tool use to avoid applying tool on removed layer on undo
-  if (toolH->isToolBusy()) {
-    toolH->getTool()->onDeactivate();
+  // do not use undo if tool is currently in use
+  if (!toolH->isToolBusy()) {
+    bool ret = TUndoManager::manager()->undo();
+    if (!ret) DVGui::error(QObject::tr("No more Undo operations available."));
   }
-
-  bool ret = TUndoManager::manager()->undo();
-  if (!ret) DVGui::error(QObject::tr("No more Undo operations available."));
 }
 
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


RodneyBaker:
So many people use Control + Z as a drawing tool (yeah I know) that removing the ability to use such would not be welcomed by those users.  Perhaps the Undo functionality used while drawing could be evolved into something more dedicated to the process of drawings... with more of a focus on history than erasing of mistakes but... insert shrug here... not sure.

Funinja:
Thanks for the reply @RodneyBaker! Yeah my bad, what I meant to say was to disable the undo ability while drawing not the Ctrl+Z. 

Essentially the problem is that when a user is using a tool, their tool action doesn't get added to the undo manager stack until the tool finishes. Meaning, if you create any new level, begin drawing on that level, and use an undo while drawing, the undo will remove the layer the user is drawing on since the user hasn't finished drawing.

Right now my fix forces the user to stop using the tool so it gets added to the undo manager, like so:
ezgif com-video-to-gif

But it still leaves behind a line since the user would still be drawing while using the undo command, so it might be better to disable the undo command when a tool is being used.

RattusaurusRex:
Other programs (Photoshop, CSP, Fire Alpaca, etc) don't let you use Undo until you lift your pen, so that would probably be preferable.

Funinja:
> Other programs (Photoshop, CSP, Fire Alpaca, etc) don't let you use Undo until you lift your pen, so that would probably be preferable.

Alrighty, I'll change it to disable undo while a tool is being used instead.

RodneyBaker:
Not sure this will work if you don't have permissions but...

If not we'll need to invoke @shun-iwasawa to get you on the list.

RodneyBaker:
Thank you @Funinja for the fix.
LGTM

Thank you for contributing to Opentoonz development!

RodneyBaker:
Tracking this from @manongjohn's fix/revert over at Tahoma2D.
@Funinja Both this fix as well as that one needs to be reviewed to see if there is a deeper fix that will address all sides of the issue.

REF:  https://github.com/tahoma2d/tahoma2d/pull/1374

Funinja:
> Tracking this from @manongjohn's fix/revert over at Tahoma2D. @Funinja Both this fix as well as that one needs to be reviewed to see if there is a deeper fix that will address all sides of the issue.
> 
> REF: [tahoma2d/tahoma2d#1374](https://github.com/tahoma2d/tahoma2d/pull/1374)

My apologies, my implementation was too naive for the original fix. I'll get started on it right away.

PR #5213, opened 2023-12-09
fixing workflow_macos.yml

Funinja:
This PR fixes the brew installation of the macos test workflow file.

I believe that the GitHub macos workflow may have been updated to include python 3.12, which would break the brew installation if the python 3.12 symlinks are not removed. Therefore, I just added their removal in the symlink remove step in the macos workflow file to fix the build test.

Successful build after my fix: https://github.com/Funinja/opentoonz/actions/runs/7152992464

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -10,7 +10,7 @@ jobs:
 
     - name: Remove symlinks
       run: |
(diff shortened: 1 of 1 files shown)


RodneyBaker:
LGTM
Thank you @Funinja for the workflow fix.

Issue #4737, opened 2023-02-04
Lost Vector Fill Issue - simple file and steps to reproduce.


gab3d:
### Description

The attached A.pli file allows to reproduce the loss of the fill color in a simple closed stroke.
I know this issue is a long standing one, but since it's still an elusive one, I thought more data on how to reproduce it could be useful.

[A.zip](https://github.com/opentoonz/opentoonz/files/10609861/A.zip)

### Steps to Reproduce

1. Load the A.pli level into an empty cell of the Xsheet. The drawing shows two closed and filled shapes formed by a single stroke each.
2. Select the stroke at the center of the drawing.
3. Right click and select: Remove Vector Overflow.
4. Undo.
5.  ...the fill is lost.

Curiously enough the other similar stroke in the drawing can pass the Undo step without any problems.

### Screenshots, Video & Crash Logs

https://user-images.githubusercontent.com/18104091/216793632-bbc86ca3-d2e7-44d6-9522-2ede8457d5ed.mp4

### OpenToonz Version

1.6

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

None

RodneyBaker:
I believe the underlying problem relates to a single stroke being used to define an area for color fill.
To test this theory...

- Take your problematic stroke and anywhere along the stroke cut it with one cut using the Cutter Tool.
- Select the stroke(s)... as there will now be two instead of one to trigger the required option to Remove Vector Overflow from both ends of the original stroke
- Undo
- Redo

The problem should go away.

Added:  There is an interesting and related problem in that if the vector level is imported instead of loaded the fill can remain slightly longer although it too will disappear... but... even when returning to the original vector file (the one provided via this report... the fill will not return.  This suggests that OT is storing some data in memory that remains until a refresh allows the file to be see as fillable again.  The surest way to get the original to fill again is to remove the imported copy and then extract the original again which apparently OT sees as a new file (?)  Not sure.

RodneyBaker:
If we make a cut in both shapes we can then select the entirety of the level, Remove Vector Overflow, and Undo/Redo without losing the color fill in either shape.

RodneyBaker:
@gab3d Can you provide additional detail on how the shapes in your A.pli file were created?
Were they part of a larger image... or randomly doodled... etc.

gab3d:
@RodneyBaker  thank you for your response.
Both shapes were randomly drawn, using the Brush Tool. The centered one in first place, and then the other one.

[...]

PR #5228, opened 2023-12-22
Known vector fill glitch fix

Funinja:
This PR fixes issue #4737.

A region is defined by edges. The edge class contains a stroke with m_w0 and m_w1 values that define the beginning and end of the stroke that creates the region. Thus, when the vector overflow is removed m_w0 = 0.0 and m_w1=1.0 to signify that the entire stroke is used to create a region. When the user tries to undo a vector overflow removal, Opentoonz needs to calculate the m_w0 and m_w1 values of the undo vector by calculating the offset of the stroke, ie the length from the start of the stroke to m_w0. This is where the problem lies.

The offset is calculated by getting the point on the old stroke closest to the beginning of the new stroke. This is because if you draw a loop the vector will overflow at the beginning and end from naturally leaving behind a stroke, meaning the start of the new stroke would be the point where m_w0 begins for the old stroke. 

However, there are two points closest to the start of the new stroke, the beginning and end of the intersection. This means Opentoonz will potentially define the start of the region as the end of region, and shift the region to the end of the stroke where there is no region at all:

image

To fix this, when the user uses an undo, I transfer the offset calculated from the vector removal overflow method to the undo method to use the correct offset.

Correct functionality:
ezgif com-crop (1)

This fix also fixes broken symlinks to remove brew warnings in the macos workflow file.


gab3d:
Awesome!!
Thank you!

RodneyBaker:
This works great.
Thanks @Funinja for the fix!

Posting intent to merge.

PR #5235, opened 2023-12-28
Sub-Xsheet Navigation Bar

manongjohn:
This was a requested T2D port to add the following feature:

This new feature will make it easier to visualize and navigate multiple nested sub-xsheets.

image

The panel in the timeline/xsheet shows you where you are in nested sub-xsheets (Xsheet Depth), relative to the main. You can click on any prior sub-xsheet breadcrumb to return to it quickly. Breadcrumbs will also show what sub-xsheets are in the current timeline/xsheet you can open. Click on the breadcrumb to enter it. In cases of multiple sub-xsheets in the current timeline/xsheet, a dropdown will be shown. Select the one you want to enter.

This new panel can be toggled on from column header's context menu, a defined shortcut or in Preferences -> Xsheet

image

image

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -2268,7 +2268,7 @@ Ruler {
   padding-right: 3;
 }
 #XsheetBreadcrumbs {
-  padding: 0;
+  padding: 0 5px 0 5px;
   margin: 0;
   border-bottom: 1 solid #262728;
 }
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -2267,6 +2267,14 @@ Ruler {
 #RenameCellField {
   padding-right: 3;
 }
+#XsheetBreadcrumbs {
+  padding: 0;
+  margin: 0;
+  border-bottom: 1 solid #262728;
+}
+#XsheetBreadcrumbs::separator:horizontal {
+  margin: 0 0 0 2;
+}
 /* xsheetColumnHeader (Context Menus)
 ----------------------------------------------------------------------------- */
 #xsheetColumnAreaMenu_Preview {
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -2268,7 +2268,7 @@ Ruler {
   padding-right: 3;
 }
 #XsheetBreadcrumbs {
-  padding: 0;
+  padding: 0 5px 0 5px;
   margin: 0;
   border-bottom: 1 solid #111111;
 }
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -2267,6 +2267,14 @@ Ruler {
 #RenameCellField {
   padding-right: 3;
 }
+#XsheetBreadcrumbs {
+  padding: 0;
+  margin: 0;
+  border-bottom: 1 solid #111111;
+}
+#XsheetBreadcrumbs::separator:horizontal {
+  margin: 0 0 0 2;
+}
 /* xsheetColumnHeader (Context Menus)
 ----------------------------------------------------------------------------- */
 #xsheetColumnAreaMenu_Preview {
--- a/stuff/config/qss/Default/Default.qss
+++ b/stuff/config/qss/Default/Default.qss
@@ -2267,6 +2267,14 @@ Ruler {
 #RenameCellField {
   padding-right: 3;
 }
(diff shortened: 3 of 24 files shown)


RodneyBaker:
I ran into a crash while testing this that echoed a crash I got earlier today while working with this feature (in T2D).
The only thing that stands out is that I switch the Selection Tool just prior to the crash.
Not a lot of info to go on at this point but will delve a little more deeply.

Here's the playthrough of the workflow right up to the moment of crashing:


https://github.com/opentoonz/opentoonz/assets/2576432/45a39217-773e-4f04-8954-afd1d9220375

RodneyBaker:
I'll add this also.
It appears that just navigating to different levels of subxsheets makes the scenen dirty (flagged as unsaved).
Not sure why it is doing that.

manongjohn:
> I ran into a crash while testing this that echoed a crash I got earlier today while working with this feature (in T2D).

Please provide crash log from T2D.

RodneyBaker:
> > I ran into a crash while testing this that echoed a crash I got earlier today while working with this feature (in T2D).
> 
> Please provide crash log from T2D.

Sorry John, I don't think I have that crash log to share.
It may be stored on harddrive and I'll see if I can locate that and post it.

My current attempts to do the same process in T2D that crashes in OT doesn't result in a crash with T2D.
But... I'm having stranger things happen currently in T2D that I need to investigate and report that have nothing to do with crashing.

I'm attaching a crash report from OT although I'm not sure that will be particularly helpful.
Currently I'm waiting to investigate if the crash could relate to having multiple monitors as others (on discord) can't reproduce the crashing.

[OTcrash during subxsheet breadcrumbs.txt](https://github.com/opentoonz/opentoonz/files/13787303/OTcrash.during.subxsheet.breadcrumbs.txt)
 
I'm also attaching another video that shows how easy it is to crash OT when using this basic process on my system

https://github.com/opentoonz/opentoonz/assets/2576432/24005d64-f1b3-43c0-890f-6fc751ab973f

:

manongjohn:
Crash logs for Windows builds can be found here: `C:\Users\youruserid\AppData\Local\xxx\xxx\crash`.  Replace `xxx` with Tahoma2D or OpenToonz.

manongjohn:
I tried to duplicate your 2nd video but I had no crash.

This new feature uses existing logic for switching in/out sub-xsheets. It just creates a UI to replace the need for using shortcuts, menu commands or customized toolbars. I made no change to that so likely these crashes are result of something else but may be exacerbated by switching in/out of sub-xsheets in general.

manongjohn:
I found the issue.  I need to put in a fix in both T2D and OT for this.

konero:
That seems to fix the crash, can no longer reproduce it. 👍 

Maybe two small quirks:

1. When changing orientation (using the button) the breadcrumb bar doesn't update position until you click on an item
2. (really very minor) The x-sheet orientation could do with 5px of left-padding to move it off the panel edge, but setting this in stylesheet makes the timeline version move out of alignment a bit. If it's worth tackling this, maybe hard-code padding then adjust the timeline position to compensate, or maybe just for OT move the timeline version to the left edge like x-sheet if it simplifies it

But nice job! It's gonna be one of those things where you can't imagine it not being there. 😂

RodneyBaker:
That indeed does appear to have solved the crash.
Thanks @manongjohn (and @konero!)

manongjohn:
@konero 

1.  I have a fix for this and will push up shortly
2. What do I need to put in the stylesheet to put padding on left and right, but not top and bottom?  Code adds spaces for the Timeline orientation so I can easily move it back to account for the padding.

manongjohn:
Nevermind, I figured it out

RodneyBaker:
Everything looks good here except the MacOS Mojave checks which I'm not quite sure why we even have that automated build...

As always thank you @manongjohn for all you do.
This Sub-xsheet Navigation Bar (Subxsheet Breadcrumbs) will be highly useful.
LGTM

gab3d:
@manongjohn  Thank you for this feature and port!! It's very useful.

Looking at the breadcrumbs, I see they show names, so I have a question: 
Do sub-Xsheets have a name property now? (one that can be referenced in expressions: i.e. xsheet.column.property)
That is one of the limitations of OpenToonz for building more complex (and useful) rigs.

manongjohn:
@gab3d The breadcrumbs just show the column name from the xsheet. I didnt add any new properites related to sub-xsheets.

gab3d:
@manongjohn 
ok, thanks for answering.

JParks123:
@shun-iwasawa  Can you please fix the nightly builds so we can test them?

PR #5272, opened 2024-01-23
Porting some fixes

manongjohn:
Here are 4 Tahoma2D fixes that were requested to be ported to OT:

-----
### 1. Fix undo selection paste into new frame + undo renumbering

While trying to reproduce the crash mentioned in a report, I found an issue with Undo when selecting part of a Raster/Smart Raster level and then pasting into a held frame w/ `Creation In Hold Cells` toggled `On`. A new blank frame is created with the paste, but an Undo doesn't remove the newly created frame. There after, the undo history becomes messed up.

Corrected the logic to record the creation of the blank frame as part of the paste so it is undone correctly.

Also fixed an issue with undo and cell renumbering. After removing the newly created frame as part of the undo, an update to the level was triggered. This reset the old frame ids list which can cause the renumber to crash if the list was empty. This may be why the crash occurred, but I could not duplicate the original issue to confirm. Modified the logic to update the level after the renumber is complete.

-----
### 2. Fix actions that shouldn't happen while autosaving

While reviewing a reported issue, I noticed there is a potential for overlapping save operations under certain scenarios which could potentially result in file corruption. Additionally, the report suggests an undo operation was performed while an autosave was in progress. This could cause a crash when it tries to access something that may no longer exist.

I've added logic to internally flag when a save is in progress so that any additional calls to save while actively saving will wait until the flag is cleared.

Scenarios that should be covered by this fix

Timed autosave + suspended autosave due to active drawing or something
Timed autosave + manual save by user
Timed autosave + Undo/Redo
NOTE: It's hard to duplicate overlapping saving issues naturally. I mostly regression tested for normal saving and autosaving operations as well as a simple saving error (file locked) and ensuring I didn't completely freeze up.

-----
### 3. Fix palette loss when caching images

Fixed a crash caused by the system running low on memory. When it detects the system is low in memory, it will compress and cache raster images in memory or onto disk in order to gain more space. When it does this, it loses palette information. Later, when the palette information is accessed, it may either do nothing or crash as was the case here.

Modified logic to store/restore palette information for compressed images.

-----
### 4. Fix insert column crash without a column selected

Some column operations like `Fold Column` and `Delete` cause the selected column list to be cleared. When you attempt an Insert column without a selected column, the logic didn't account for this and crashed.

Added logic to find the current column focus, outlined in cyan, and use that as the insert point. Still added logic to ignore the insert action if for some reason a current column still could not be found.

### DIFF
--- a/toonz/sources/common/tcache/timagecache.cpp
+++ b/toonz/sources/common/tcache/timagecache.cpp
@@ -31,6 +31,7 @@
 #include "tsystem.h"
 
 #include "traster.h"
+#include "tpalette.h"
 
 //#include "tstopwatch.h"
 #include "tbigmemorymanager.h"
@@ -95,14 +96,16 @@ public:
       : m_cantCompress(false)
       , m_builder(0)
       , m_imageInfo(0)
-      , m_modified(false) {}
+      , m_modified(false)
+      , m_palette(0) {}
 
-  CacheItem(ImageBuilder *builder, ImageInfo *imageInfo)
+  CacheItem(ImageBuilder *builder, ImageInfo *imageInfo, TPalette *palette)
       : m_cantCompress(false)
       , m_builder(builder)
       , m_imageInfo(imageInfo)
       , m_historyCount(0)
-      , m_modified(false) {}
+      , m_modified(false)
+      , m_palette(palette) {}
 
   virtual ~CacheItem() {}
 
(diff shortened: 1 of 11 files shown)


RodneyBaker:
@shun-iwasawa Can you look into why appveyor is failing?  Thanks!

RodneyBaker:
@manongjohn If you can please refresh this PR the appveyor refresh (opencv update) should allow it to complete the check.
Thanks!

RodneyBaker:
Thank you @manongjohn for these very important bug fixes!
LGTM

PR #5295, opened 2024-02-11
 Add new brushes and fixes

andeon:
Add news brushes and fixes.

The sketch and ink brushes are reworked, now have more varieties, a brush has been added to simulate rake effect.
brushes
For painting using many raster brushes I recommend use the "**Settings tab**" to set the "**Radius**" for each **style color**. The style color keep the configuration unlike "**Size**" in the "**Tool options bar**", even using  **preset** wouldn't be efficient as there.

Repository Pack : https://github.com/andeon/aotz/releases/tag/v0.7

PS, If gonna test use the  appveyor artifact;


RodneyBaker:
Pressing forward with this addition... with grateful thanks to @andeon for the additional brushes and update.

Issue #5288, opened 2024-02-04
Undo permanently disabled on MacOS when using drawing tablet


OwlOfAvery:
### Description

On MacOS (Sonoma 14.3 - Intel CPU), undo does not work when using drawing tablet. Neither the shortcut command nor manually clicking from edit > undo. However, you can undo by manually using the history panel. Tested multiple nightlies, appears to be introduced in the 12-06 build, likely due to PR #5187. If I manually draw using mouse or trackpad, undo works as expected

### Steps to Reproduce

1. Open new or existing project
2. Grab any drawing tool
3. Using drawing tablet, draw on canvas
4. Attempt to undo

### Expected Behavior

Expect undo to undo the last action

### OpenToonz Version

Nightly (Specify below)

### OpenToonz Version Information

2023-12-06 build through current

### Operating System

macOS

### GPU

AMD

### Graphics Tablet

Huion Kamvas

andeon:
@RodneyBaker Since the issue appears to have been resolved by the latest PR linked to it, it can be closed. 
@OwlOfAvery Please test with the latest version of [OpenToonz](https://github.com/opentoonz/opentoonz/releases), and if the issue persists, provide additional details.

PR #5301, opened 2024-02-14
Fix permanently disabled undo in MACOSX

Funinja:
This PR aims to fix the previous PR #5187 and issue #5288. 

It turns out the variable I used in the previous PR fix, m_toolIsBusy, is permanently true on macosx and linux. Since the software always believes a tool is busy it prevents itself from using undo to prevent other bugs. 

**The reason for the variable being permanently true is as follows:**

image
This function keeps being called in the tablet move event, setting the value to be true for macosx without checking if a stroke is being created.

I'm not completely sure if I am correct about this fix because **I don't have a mac, so if someone with a mac could please check if they can undo a drawn line using my PR** that would be great. The variable is also used for the autosave feature and could potentially fix other long term bugs. Also, Tahoma2d has the same code structure for the tablet event, so if all goes well the fix could be imported into Tahoma2d.

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -343,11 +343,6 @@ void SceneViewer::tabletEvent(QTabletEvent *e) {
 #endif
   } break;
   case QEvent::TabletMove: {
-    // for Windowsm, use tabletEvent only for the left Button
-    if (m_tabletState != StartStroke && m_tabletState != OnStroke) {
-      m_tabletEvent = false;
-      break;
-    }
 
 #ifdef MACOSX
     // for now OSX seems to fail to call enter/leaveEvent properly while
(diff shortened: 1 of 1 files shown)


manongjohn:
Please note, this issue also appeared to have impacted at least 1 windows user that i know off, though it was not all the time.  I dont know what exact conditions it might have triggered on Windows systems since i could not reproduce it myself. I suspect it was a random move event that was being processed after a pen lift.

Funinja:
> Please note, this issue also appeared to have impacted at least 1 windows user that i know off, though it was not all the time. I dont know what exact conditions it might have triggered on Windows systems since i could not reproduce it myself. I suspect it was a random move event that was being processed after a pen lift.

Yeah, looking at the file there can be different orderings in events that could cause that that I don't want to change. I'll probably just extend the tool class to have a field to check if its in use instead of relying on the tool handle busy flag.

PR #5302, opened 2024-02-14
Macos undo freeze

Funinja:
This PR is a continuation of PR #5301, attempting to fix PR #5187 and issue #5288.

The original issue, is that a mac user cannot use undo after drawing a line with their tablet. This is because sometimes, the busy tool flag is sometimes permanently true preventing the undo function from working.

The new implementation adds an undo flag on the tool class, thereby separating the need to check the busy tool flag used for the different implementations of macos and windows. Moreover, the extension attempts to get away from the event handling bugs that cause the variable to be permanent. Therefore, the user should be able to undo after using a tool with a tablet for any OS with this fix.

I already tested it on windows and it seems to be okay.

**I don't have a mac, so we need someone to test this PR on a mac to verify my implementation.**

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -585,7 +585,7 @@ protected:
   int m_targetType;  //!< The tool's image type target.
 
   bool m_enabled;  //!< Whether the tool allows user interaction.
-  bool m_canUndo; //!< Whether the tool allows the user to undo while the tool is selected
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Pressing forward with this so that a few Mac testers can have more easy access to verify.
I don't see anything amiss in the code itself.

Thank you @Funinja for the fix.

PR #5315, opened 2024-02-28
New OCA Import feature + Export fixes

manongjohn:
This is a requested port from Tahoma2D to add the following feature and fixes:

### OCA Import
This OCA Import PR is based on the early commits of @janimatic's original OCA Import PR in Tahoma2D's repo.

This is a cleaned up and simplified version of the OCA Import logic that was being introduced. I had to revise it for various reasons and chose to undertake it myself due to the complexity of trying to explain how image files needed to be loaded and handled. Additional corrections were made.

OCA Import notes:

- OCA Layer Types
  - `paintlayer` - Supported, except for `blendingModes`.
  - `vectorlayer` - Supported via automatic conversion of SVG to PLI format
  - `grouplayer` - sub-layers will be imported without grouping

- User will be prompted to Import or Load
  - Importing places PNG files in +extras and SVG files in +drawings
  - Loading (instead of Importing) vector layers will create the PLI file in the source directory.
- Layers without image files will not be imported.
- When importing, if the level already exists, user will be prompted to either keep, overwrite or rename.
- In case of import issues, warnings and errors will be displayed as it imports each layer
  - User can opt to ignore all future warnings and errors to speed up loading.

### OCA Export fixes
- Fixed an issue adjusting start/end time using output settings when it shouldn't have
- Fixed an issue labeling vector exports as `paintlayer` when it should be `vectorlayer`
- Removed framenumber offset logic. Spec has framenumbers starting at 0, which we will auto translate to start at 1
- Added OCA Export icon


RodneyBaker:
@manongjohn I believe @shun-iwasawa may have removed the Mojave Mac check.
If you refresh this PR it should remove that check and pass all current checks.

I'd just merge this PR but... better to have a clean pass of all checks.

RodneyBaker:
Thanks.  It looks like the check is still there and will plan to press forward with this merge tomorrow.

Posting intent to merge.

RodneyBaker:
Thank you @janimatic and @manongjohn for this feature (and fixes!).
LGTM

janimatic:
congrats, and thanks to you two...

jorgedelcampo:
Hi guys. Where exactly should we find this feature in the UI? I managed to import an OCA file via "Level > Load Level", but I'm not sure if that is the intended workflow. 

Considering that "File > Export" explicitly includes an option for OCA, wouldn't it be more intuitive to have a corresponding "Import OCA" option under "File > Import"? I think aligning both actions in the same menu would make the feature much easier to discover.

btw: I'm using new 1.8 RC version.
image
image

PR #5317, opened 2024-03-01
Refactor FFmpeg format check on launch

shun-iwasawa:
This PR refactors the file i/o availability checking procedure on launch.
Before this PR, OT calls ffmpeg.exe for each format type. This PR makes OT to keep the first returned value (a text containing all available format extensions) and reuse it for the rest of the formats.
This modification shortens the launching time for about 10 seconds in our environment.


RodneyBaker:
Thank you @shun-iwasawa for the refactoring.
Much appreciated!

PR #5333, opened 2024-03-12
Fix OCA import/export frame range

ChengduLittleA:
This changes OCA Import/Export's frame range as follows:

Export frames based on timeline/xsheet frame count instead of using Output Setting's frame range so the entirety of the scene is always exported.
This also applies to sub-scenes which can have more or less frames than the main.
Fixed an issue with using the wrong frame range for folders within sub-scenes.

Do not update scene's Output Settings frame range using Import's startTime and endtime.
This will allow the scene's Output Setting frame range to remain dynamic, changing as frames are added to extend the scene after import.

### DIFF
--- a/toonz/sources/toonz/ocaio.cpp
+++ b/toonz/sources/toonz/ocaio.cpp
@@ -180,12 +180,9 @@ bool OCAData::buildGroup(QJsonObject &json, const QList<int> &rows,
 
   // Build a list of child rows
   QList<int> crows;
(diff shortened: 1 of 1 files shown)


RodneyBaker:
All my testing suggests this port is looking good.
Posting intent to merge later today.

RodneyBaker:
Thank you @ChengduLittleA for this port and @manongjohn for the fix.
Thanks also to @JParks123 for your behind the scenes wrangling in support of Opentoonz and the OCA format.
LGTM

PR #5334, opened 2024-03-12
Fixes to SVG Import and Export

ChengduLittleA:
This is ported from https://github.com/tahoma2d/tahoma2d/pull/1395

### DIFF
--- a/toonz/sources/image/CMakeLists.txt
+++ b/toonz/sources/image/CMakeLists.txt
@@ -199,4 +199,4 @@ elseif(BUILD_ENV_UNIXLIKE)
     endif()
 endif()
 
-target_link_libraries(image Qt5::Core Qt5::Gui Qt5::Network ${Z_LIB} ${GLUT_LIB} ${GL_LIB} ${JPEG_LIB} ${TIFF_LIB} ${PNG_LIB} ${EXTRA_LIBS})
+target_link_libraries(image Qt5::Core Qt5::Widgets Qt5::Gui Qt5::Network ${Z_LIB} ${GLUT_LIB} ${GL_LIB} ${JPEG_LIB} ${TIFF_LIB} ${PNG_LIB} ${EXTRA_LIBS})
--- a/toonz/sources/image/svg/tiio_svg.cpp
+++ b/toonz/sources/image/svg/tiio_svg.cpp
@@ -4,6 +4,8 @@
 #include <stdlib.h>
 #include <math.h>
 
+#include <QApplication>
+#include <QDesktopWidget>
 #include <QTextStream>
 #include <QFile>
 
@@ -14,6 +16,7 @@
 #include "tregion.h"
 #include "tcurves.h"
 #include "tpalette.h"
+#include "tstroke.h"
 
 //=------------------------------------------------------------------------------------------------------------------------------
 //=------------------------------------------------------------------------------------------------------------------------------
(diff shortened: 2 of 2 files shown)


RodneyBaker:
This is a very important set of fixes that many have been looking forward to.
Posting intent to merge.

RodneyBaker:
There isn't a particular rush here if #5338 needs to be added to this PR.
@ChengduLittleA As this is your port we will leave it to you to decide if you wish to leave this port as is or add the additional line from #5338.

konero:
Yeah don't worry about letting pull requests stay open for awhile, no need to merge them so fast before testing.

RodneyBaker:
Just to be clear...
As a port, this code has already been tested pretty thoroughly via Tahoma2D.
For Opentoonz, this is mostly just the success of the port itself that is being tested.

Having said that, the more testing the better!

ChengduLittleA:
Hi guys, I just added the last line to this port as well so it's more complete :D

RodneyBaker:
For those curious about the fixes involved that don't want to follow the link to the original PR over at Tahoma2D, here's John's description quoted:   

>I'm sure there are a number of issues surrounding SVG Import, and to a degree SVG Export. This PR fixes only some of the more common issues people have run into.
> 
> SVG Import (SVG -> PLI converison)
> Fixes fill and stroke opacity not being applied
> Opacity was being read in but was not making palette entries for them, resulting in default fills, usually black, being used
> Default fill and stroke color to black when not specified
> In some cases, fill/stroke information was provided but not color so it used whatever style it successfully used before. It now defaults to Style 1, which is usually black.
> Forces default fill if there is no stroke information or stroke width = 0. It is likely a fill-shape
> Forces default fill if it's a closed shape without fill information
> Fixes incorrect application of stroke scaling information
> Was applying scale information to stroke width incorrectly.
> The scaling is now applied to the overall stroke for proper placement and sizing
> Fixes unnecessary grouping of every stroke.
> If there is a fill associated with a stroke (i.e. closed shapes), the stroke is grouped and filled
> If a fill is applied on a single stroke, the stroke will be ungrouped.
> If a fill is applied on a multi-stroke shape, the strokes will remain grouped.
> If no fill is associated with the stroke, no grouping is performed.
> Fixes an issue where there was minimal fill information (i.e just opacity), but nothing else.
> This now applies the default fill with the information provided
> Fixes stroke width importing wider than indicated in the file
> This is likely an issue where device pixel ratio > 1 (usually 2) but the information in the file only assumes 1, causing stroke width to be doubled.
> When building the stroke, it now adjusts for device pixel ratio > 1 to maintain proper stroke width
> Fixes imported SVG image displaying in the lower-right area of the canvas
> Now centers the image on the canvas if width/height or viewBox information is provided
> If none of the above is found, it will use the bounding box of all the strokes in the image to center
> Fixes importing filled shapes with cutouts
> These shapes are multi-stroke shapes. Added some logic to handle these properly
> Subject to how the write records the outer path as well as our sometimes problematic fill process.
> Fixed closed shapes having an extra point causing circles to be deformed.
> Fixed moveTo logic not properly moving relative after moving absolute
> Fixed missing fills on shapes made by single strokes. (strokes with fills).
> Added 0 width stroke to connect endpoints so fill can take place.
> SVG Export (PLI -> SVG conversion)
> Corrects how it writes stroke-width information to the file so it can be read by Inkscape.
> Now writes out fill-opacity, stroke-linecap, stroke-linejoin, stroke-miterlimit and stroke-opacity
> For closed shapes, now writes out the 'Z' in the path instruction

RodneyBaker:
Pressing forward with this.

Thank you @ChengduLittleA for the port (and additional port) and @manongjohn for the fixes.
LGTM

RodneyBaker:
@ChengduLittleA I'm investigating a problem with the currrently nightly build and find myself wondering if the change to CMake list might be the cause.

Basically what is happening is the creation of a SRV directory with contents is being placed in the installation directory and this is causing the program not to run until that directory is removed.

I've opened #5351 in the meantime while investigating.

PR #5338, opened 2024-03-12
Fix fillable regions after SVG import 

JParks123:


manongjohn:
This doesn't do anything other than remove spaces.  This should be closed.

You shouldn't be trying to port this particular fix from T2D separately since this change was built on the initial SVG Import/Export fix.  If anything the followup PR you are trying to port should just be added as a new commit to #5334.

PR #5353, opened 2024-03-21
Revert the Qt5::Widget change

ChengduLittleA:
Suspected to have caused OTZ to not launch with this change. Reverting.

### DIFF
--- a/toonz/sources/image/CMakeLists.txt
+++ b/toonz/sources/image/CMakeLists.txt
@@ -199,4 +199,4 @@ elseif(BUILD_ENV_UNIXLIKE)
     endif()
 endif()
 
-target_link_libraries(image Qt5::Core Qt5::Widgets Qt5::Gui Qt5::Network ${Z_LIB} ${GLUT_LIB} ${GL_LIB} ${JPEG_LIB} ${TIFF_LIB} ${PNG_LIB} ${EXTRA_LIBS})
+target_link_libraries(image Qt5::Core Qt5::Gui Qt5::Network ${Z_LIB} ${GLUT_LIB} ${GL_LIB} ${JPEG_LIB} ${TIFF_LIB} ${PNG_LIB} ${EXTRA_LIBS})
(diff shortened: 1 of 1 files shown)


manongjohn:
I had added Qt::Widget to the image/CMakeLists.txt file in the SVG fixes to correct a macOS build error since it could not find QApplication.

Not sure why it's not needed for Windows/Linux builds but it was a quick fix for macOS builds.  My original change doesn't impact T2D because we don't have tsvr32 anymore.

Try moving Qt::Widget to the EXTRA_LIBS definition for macOS builds only.

RodneyBaker:
It isn't clear to me why we are seeing a return of the old problem of t32bitsrv.exe.
Perhaps a regression in the build enviroment itself?

manongjohn:
It's likely because the packing of OT isn't including the 32bit version of Qt5Widgets.dll in the svr directory since the image.dll is being compiled with it.

You wouldn't see this issue in the PR artifacts since t32bitsvr is only included in the non-github builds.

Making it so the library is included in macOS builds will probably fix the issue.

ChengduLittleA:
@manongjohn I'm not sure what I'm supposed to do here to make it run correctly, should this PR be merged or actually ignored?

RodneyBaker:
For my part, I think merging will fix the WIndows build but break the Mac fix.
I am tempted to merge but would prefer the full fix.

manongjohn:
> @manongjohn I'm not sure what I'm supposed to do here to make it run correctly, should this PR be merged or actually ignored?

@ChengduLittleA If you take out the Qt5:Widget from the list, it causes macOS builds to fail.

You need modify this file so there is a separate `target_link_libraries()` list for macOS builds that includes Qt5::Widget and one for others platform builds.

manongjohn:
I updated the file (in your branch) to add the Qt5::Widgets for macOS builds only.   This PR build successfully now.

manongjohn:
I confirmed my change works, but macOS builds are failing.

Seems there is an issue with the macOS build environment.  It's failing for the following:
```
-- Build files have been written to: /Users/runner/work/opentoonz/opentoonz/toonz/build
ninja: error: unknown warning flag 'dupbuild=warn'
Error: Process completed with exit code 1.
```

RodneyBaker:
As @shun-iwasawa should now be aware of problems with the Mac build I will plan to press forward with this to get us back to the status quo.  Mac nightly builds likely won't work until that gets addressed however.

Thank you @ChengduLittleA for moving this forward and @manongjohn for the necessary refinement.

PR #5356, opened 2024-03-24
[ui] Decouple x-sheet text colors

konero:
No front-end change, this just adds some new qcolors so some stuff can be styled with stylesheets in the x-sheet separately.


RodneyBaker:
Thank you @konero for the new UI customizeable options.
Much appreciated.

Issue #44, opened 2016-03-26
TFarm:  Fix signal handler (Code Typo?)


elfring:
- [The variable "Sthutdown"](https://github.com/opentoonz/opentoonz/blob/7e185111a8e317ccea5fd8060cc15a7d1527d660/toonz/sources/toonzfarm/tfarm/ttcpipserver.cpp#L36) should [use the data type "sig_atomic_t"](https://www.securecoding.cert.org/confluence/display/c/SIG31-C.+Do+not+access+shared+objects+in+signal+handlers) (instead of bool).
- Does this name contain a typo?

RodneyBaker:
Instance of possible typo appears twice:

https://github.com/opentoonz/opentoonz/search?q=Sthutdown&unscoped_q=Sthutdown

RodneyBaker:
It appears we may be keeping this report open so it can be closed as part of the 10th Anniversary Extravaganza.  I believe it is officially the oldest report still not addressed since Opentoonz historic release.

PR #5416, opened 2024-04-23
Correct Minor Typos and Text

RodneyBaker:
Collecting here a few long standing typos and textual changes that don't effect the code itself but have been reported over the past few years.


RodneyBaker:
There are some typos located in code as well but ideally we'll make that a separate pull request assuming we want to modify them.  For example:  #44 that refers to: https://github.com/search?q=repo%3Aopentoonz%2Fopentoonz+Sthutdown&type=code

In almost all cases such as these encountered the typo does not effect the code as the spelling is consistent throughout the scope of usage.  It is not generally recommended that typos be fixed just for the sake of the typo as that can lead to unintended consequences (or more typos) but where those typos reside should be considered for review so that the typo can be fixed when the code itself is updated or modified.

PR #5419, opened 2024-04-24
windows configure ThirdParty

janimatic:
* configure the third party libs (only once) to avoid manual copies


RodneyBaker:
Pressing forward with this as it should address the build issue

PR #5420, opened 2024-04-24
Fix camera capture saying no camera available because of a Qt bug...

janimatic:
Finally OT can use a camera !

simple workaround because Qt fails to auto-detect the camera resolutions https://forum.qt.io/topic/68904/how-to-get-the-supported-resolution-of-a-qcamera/5

Ideally we should fix the Qt bug and read the resolution from the device instead !
Or in my case, use Blackmagic Decklink SDK and avoid WDM capture... (that would only be usefull for those using a decklink card like me).

### DIFF
--- a/toonz/sources/toonz/penciltestpopup.cpp
+++ b/toonz/sources/toonz/penciltestpopup.cpp
@@ -2196,6 +2196,24 @@ void PencilTestPopup::onCameraListComboActivated(int comboIndex) {
   // refresh resolution
   m_resolutionCombo->clear();
   QList<QSize> sizes = m_currentCamera->supportedViewfinderResolutions();
+  // simple workaround because Qt fails to auto-detect the camera resolutions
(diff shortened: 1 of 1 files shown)


RodneyBaker:
This fix look solid.  It's just the current checks that are being problematic.
It would be good for users with various cameras to test this and report back.

manongjohn:
Since the mac build fix was fixed and merged, any PRs that were opened before it can be updated by the the PR owner by doing one of the following

 - Rebase the PR branch against OpenToonz master and force push to update the PR.
- Pull the OpenToonz master into the PR branch and push up the changes to update the PR

janimatic:
I will do the update, but I am not certain that the camera needs any fix outside of windows WMF :
Apparently, only WMF is missing the Qt camera backend
https://wiki.qt.io/Qt_5.13_Multimedia_Backends
Is anyone using a camera in OT under linux and/or mac, that could do the test ?

RodneyBaker:
Pressing forward with this...
With thanks to @janimatic for the camera capture fix!

PR #5425, opened 2024-04-25
Fix qt mac workflow action

janimatic:
Fixed mac build (github workflow action).
A distraction because seeing the mac build fail everytime distracted me... (i don't use mac os)
* Fixed Qt instal on mac using https://github.com/jurplel/install-qt-action
see this usefull command generator :
https://ddalcino.github.io/aqt-list-server/
* Added qtscript module
* Removed deprecated ninja argument 
* Raised c++ version to 17
* There maybe some redundant informations but it works fine...

PR #5427, opened 2024-04-25
Fix mac build v2 : staying on c++11, macos-12, x86-64...

janimatic:
Minimal change to fix the macos build

* removed the obsolete ninja argument

* All the arm64 issues might be due to 
`runs-on: macos-latest` :
https://github.blog/changelog/2024-01-30-github-actions-introducing-the-new-m1-macos-runner-available-to-open-source/

replacing with
`runs-on: macos-12`
fixes it...
That would explain inconsistancies while the workflow_macos.yml didn't change, since latest behavior change over the time...

* FindQt5 needs a minimal fix (I didn't retry without that change since i switched to macos-12) :
`        cmake ../sources -G Ninja -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DQT_PATH=$(brew --prefix qt@5)/lib -DQt5_DIR=$(brew --prefix qt@5)/lib/cmake/Qt5 -DCMAKE_PREFIX_PATH=$(brew --prefix qt@5)/lib/cmake/Qt5  -DWITH_TRANSLATION=OFF
`

For a deeper fix (wip) with mac arm64 & c++17 support see https://github.com/opentoonz/opentoonz/pull/5425

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -4,7 +4,9 @@ on: [push, pull_request]
 
 jobs:
   macOS:
-    runs-on: macos-latest
+    # runs-on: macos-latest
+    # avoid the headache of arm64 builds?
+    runs-on: macos-12
     steps:
     - uses: actions/checkout@v2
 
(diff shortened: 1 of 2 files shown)


janimatic:
Even though its not really usefull for me, that's an humble satisfaction to finally see a pr that is all green...

RodneyBaker:
I'll try this out on my Mac later today but that system is so old it might not be able to validate much.

I agree that seeing all that green is very satisfying.  :)

artisteacher:
The action build opens without issue on my M1 mac.

janimatic:
good news! thanks

RodneyBaker:
The nightlies currently aren't building so... we are still in a state of largely waitingfor @shun-iwasawa to arrive on the scene and save the day.  The good news is that Shun should be available in May.   The 'gooder' news is that obstacles such as these, whiile frustrating, do often lead to better understanding.

Because this is the first PR to actually pass appveyor checks in a very long time I am going to press forward with merging.
We have a number of PRs currently open however that I'm not as comfortable pressing forward with.

PR #5441, opened 2024-04-30
Update deprecated github artifact actions to version 4

RodneyBaker:
Version 1, 2 and 3 of github artifact actions are being deprecated.

Please see action error logs and:  https://github.blog/changelog/2024-04-16-deprecation-notice-v3-of-the-artifact-actions/


RodneyBaker:
Pressing forward with this which should not effect appveyor in any way.

PR #5444, opened 2024-05-05
Allow multiple key stroke shortcuts

manongjohn:
This requested Tahoma2D port enhances shortcuts by allowing you to use up to 4 key strokes to trigger an action.

For example, you could set `Play` to `P,L,A,Y` if you wanted to.  Only when you hit those 4 keys in that order will the Play action be triggered.

**Limitation:**

The beginning part of the sequence (1st key, 1st 2 keys, 1st 3keys) cannot match another shortcut using the same number of keys since the actions with the fewer key strokes will take priority and trigger.

Examples of bad assignments:
- 1-key matches
  - `P` = `Play`
  - `P,L` = `Short Play` => would never be reached because 1st `P` would trigger Play
- 2-key matches
  - `P,L` = `Play`
  - `P,L,A` = `Short Play` => would never be reached because `P,L` would trigger Play
- 3-key matches
  - `P,L,A` = `Play`
  - `P,L,A,Y` = `Short Play` => would never be reached because `P,L,A` would trigger Play

A warning will be provided if you try to do this.
  
You could instead do:
- 2-key
  - `P` = Not a shortcut
  - `P,P` = `Play`
  - `P,S` = `Short Play`
- 3-key 
  - `P` = Not a shortcut
  - `P,L` = Not a shortcut
  - `P,L,A` = `Play`
  - `P,L,S` = `Short Play`
- 4-key
  - `P` = Not a shortcut
  - `P,L` = Not a shortcut
  - `P,L,A` = Not a shortcut
  - `P,L,A,Y` = `Play`
  - `P,L,A,S` = `Short Play`

### DIFF
--- a/toonz/sources/toonz/shortcutpopup.cpp
+++ b/toonz/sources/toonz/shortcutpopup.cpp
@@ -73,7 +73,7 @@ public:
 //-----------------------------------------------------------------------------
 
 ShortcutViewer::ShortcutViewer(QWidget *parent)
-    : QKeySequenceEdit(parent), m_action(0) {
+    : QKeySequenceEdit(parent), m_action(0), m_keysPressed(0) {
   setObjectName("ShortcutViewer");
   setSizePolicy(QSizePolicy::Expanding, QSizePolicy::Expanding);
   connect(this, SIGNAL(editingFinished()), this, SLOT(onEditingFinished()));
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Looking good!

There are a few commands I struggled with assigning a shortcut to but I think that is because I needed to reassign that lower level key (1st key) to get the subsequent key assignments to work.  (i.e. User error on my part)

Hopefully we'll have some folks share their favorite shortcut key combinations expecially as those can be loaded and saved via the shortcut dialogue.  

Fodder for the future:  I'd love to see the ability to assign multiple keys to trigger the same command.  For instance, we might have number keys for all the tools but still retain other common keys for use.  Then we just use whichever one is closest to where are fingers are at the time the command is needed.

beeheemooth:
One of the disadvantages of multiple shortcuts is the loss of the ability to temporarily switch while holding the button. For example, if you have shortcuts configured for the Eraser Tool:
-----for regular Eraser
- E,E
-----for Eraser with interpolations
- E,L
- E,I
- E,O
----for Eraser types
- E,F
- E,R
- E,P
- E,S

 You cannot use E,E as a temporarily held switch to return to the previous tool or option when released.
In theory, is it possible to make holding the last key of a shortcut temporarily toggle the desired option?

RodneyBaker:
The Eraser with interpolation feature hasn't yet made its way to Opentoonz but I'm following what you are suggesting.
It would be nice if perhaps holding the last key (or any key for that matter for an extended period of time acted as if resetting the keystroke sequence to the first letter again.  (not sure if that is quite what you are suggesting here @beeheemooth).

beeheemooth:
> The Eraser with interpolation feature hasn't yet made its way to Opentoonz but I'm following what you are suggesting. It would be nice if perhaps holding the last key (or any key for that matter for an extended period of time acted as if resetting the keystroke sequence to the first letter again. (not sure if that is quite what you are suggesting here @beeheemooth).

In multiple shortcuts, the buttons are pressed sequentially, so you can call up any command with one finger (which is important for people with limited hand capabilities). Therefore, it is logical that in order to temporarily switch, you need to hold down exactly the last button of the shortcut for a time greater than that specified in the “Temporary Tool Switch Shortcut Hold Time” field. (if, of course, what we are talking about is possible)

RodneyBaker:
I do like the idea of being able to hold down a key for extended timeframe to activate a switch.

RodneyBaker:
Pressing forward with this...

Thanks @manongjohn for the code and the port.
Thanks @Sodapop for the interest and developmental funding.

JParks123:
@shun-iwasawa  can you please fix the nightly builds so I can test this PR I requested?

PR #5461, opened 2024-05-16
Add TONE Mypaint Brushe pack

JParks123:
this brush pack was created by: Luovatone

image


manongjohn:
The `stuff\library\mypaint brushes\Licenses.txt` should be updated with the correct copyright information if this brush pack is to be automatically included as part of OT's distribution.

I would also suggest putting the brushs in a sub folders like the other brush packs.

konero:
I think it would be better to have an import brush pack command that individuals can use, with a few included brushes that are tested and fully compatible the way Krita does it, as many aren't, quality over quantity, can’t really add every pack into the release, the brush page is already a bit bloated.

Just my opinion.

PR #5515, opened 2024-07-05
Fix Exporting/Importing OCA folder

manongjohn:
This is a requested port from T2D which corrects OCA Export and Import as follows:

- Export:  Per OCA specs, the actual OCA file and directory containing the image files are now created inside a folder of the same name as the actual OCA file (i.e `C:\OCA Exports\myscene.oca\myscene.oca`)
  - Change the name of the subdirectory containing the level files to be the name of the export + "_files" (i.e `C:\OCA Exports\myscene.oca\myscene_files`)

- Import: Handles importing an OCA folder or an actual OCA file.
  - In case of folder, it looks for the actual OCA file of the same name within the selected folder.

### DIFF
--- a/toonz/sources/toonz/ocaio.cpp
+++ b/toonz/sources/toonz/ocaio.cpp
@@ -568,6 +568,9 @@ void ImportOCACommand::execute() {
     importEnabled = (ret == 1);
   }
 
+  // Check if this is a directory or the actual file
+  if (TFileStatus(fp).isDirectory()) fp = fp + fp.withoutParentDir();
+
   QString ocafile           = fp.getQString();
   OCAInputData ocaInputData = OCAInputData(0, false, false);
   QJsonObject ocaObject;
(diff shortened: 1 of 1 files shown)


RodneyBaker:
There was a question someone posted regarding the subdirecttory inside the OCA folder and if it also fits into the OCA spec.
The answer is 'yes, it complies with the spec':

```TIP  All layers may be stored in a single subfolder with an arbitrary name (e.g. “content”, “data”, “layers”, “Document”…) instead of directly at the root.```

REF:  https://oca.rxlab.guide/specs/folder-structure.html  (at thee very bottom of the page)

manongjohn:
It appears Appveyor updated the VS 2019 Build environments last month to now have Boost 1.85.0 installed by default.  OT's current Appveyor scripts need to be updated to use that instead.

Updating the appveyor biuld script in a separate PR will not help existing PRs unless open PR's pulling the change in after its merged into master.

RodneyBaker:
Thanks @manongjohn I believe that should fix a number of issues with perhaps the most important being the failure of Mac nightly builds.

RodneyBaker:
If it doesn't fully seat in the nightly release it should do so upon the next PR merge.

PR #5516, opened 2024-07-05
Exploration of workflow_windows.yml

RodneyBaker:
Investigating some changes to workflow to improve reliability.

### DIFF
--- a/.github/workflows/workflow_windows.yml
+++ b/.github/workflows/workflow_windows.yml
@@ -5,8 +5,6 @@ on:
     branches:
       - master  # Specify branches where the workflow should run
   pull_request:
-    branches:
-      - master  # Specify branches where the workflow should run
 
 jobs:
   Windows:
--- a/.github/workflows/workflow_windows.yml
+++ b/.github/workflows/workflow_windows.yml
@@ -3,10 +3,10 @@ name: Windows Build
 on:
   push:
     branches:
-      - main  # Specify branches where the workflow should run
+      - master  # Specify branches where the workflow should run
   pull_request:
     branches:
-      - main  # Specify branches where the workflow should run
+      - master  # Specify branches where the workflow should run
 
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow_windows.yml
```diff
@@ -1,6 +1,12 @@
 name: Windows Build
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+      - main  # Specify branches where the workflow should run
```
manongjohn, line 6: OT's main branch is `master` so this workflow will never run when merged. 

When specifying  specific branches this way, this workflow will no longer run against PRs.

Not sure what you are trying to fix, but i dont recommend this change.

manongjohn: The windows one that is having trouble is not a Github action and is specific to an external build process set up by Dwango/DV? We have 0 control over that as far as I've known.  Same with the 1 macOS build.

The only other external build process we have any control over right now is the Appveyor windows builds.

Appyveyor builds tend to have issues if it takes 1 hour ro build.  Recent Appveyor builds are failing due to omething is going on with their build environments, i think.

RodneyBaker: I've sent a message to @shun-iwasawa so hopefully he will have things sorted soon.
He will likely have insight into the specific Dwango/DV build that is separate from what we experience here.

I don't personally have much control over the appveyor build although perhaps there is more that we can do that I am aware of.  For example, I don't have the same ability to restart Appveyor builds that you do as I don't have that permission.  I'm quite sure that extends to other aspects of the build system as well.

manongjohn: Appveyor build steps are defined in the appveyor.yml found in the root of this repo

RodneyBaker:
I expect to close this PR shortly but want to see all the current PRs running complete their checks.

PR #5517, opened 2024-07-05
Update bug_report template.yml

RodneyBaker:
Minor update of wording and link to known issues.
A note is added about incomplete reports as crash dumps by themselves are generally not sufficient to troubleshoot.


RodneyBaker:
Will press forward with this bug report template update as it doesn't factor into the Opentoonz build.

PR #5519, opened 2024-07-06
Update Boost in workflows

RodneyBaker:


RodneyBaker:
Pressing forward with this boost reference update.

manongjohn:
You only updated the github workflow. The Appveyor workflow still fails

RodneyBaker:
The appveyor was updated as well but I don't have proper permissions to restart/refresh appveyor.
As for the appveyor.yml... please see direct changes made to Appveryor.yml here**:  https://github.com/opentoonz/opentoonz/blob/master/appveyor.yml

Not sure why that was a direct merge.  Thought I was adding the commit to the patch.

manongjohn:
The next PR will trigger an Appveyor build which should use the new settings.

@shun-iwasawa To avoid unintended pushes directly to master, please add branch protection rules so that all pushes to `master` require a PR except by a few individuals (you).  There really should be no one who can push directly to master other than the owner/administrator.

shun-iwasawa:
@manongjohn I think I could add a branch protection rules to require a PR before merging. Thank you for your suggestion!

PR #5523, opened 2024-07-10
Modify CMakeLists.txt for Windows

shun-iwasawa:
This PR slightly modifies #5419 so that it works on my environment.

- Changed to use `if(NOT EXISTS ...)` condition arguments for checking the file existence.
- Changed to use `file(COPY_FILE ...)` instead of `file(RENAME ...)` in order to keep the original files and prevent git from detecting file deletion.


RodneyBaker:
Will plan to merge this as it appears to resolve the core of the problem.

Issue #4962, opened 2023-06-18
Dialogs action buttons not working properly 


andeon:
### Description

Dialog with buttons "Save All", "Save Scene Only", "Discard Changes" , "Cancel"  etc. not working

I'm testing with appimage, I think only linux users are affected by this, not sure about MacOS.

I believe the problem was introduction after this fix patch: https://github.com/opentoonz/opentoonz/pull/4872

Edit:
Weird I decided to test the opentoonz flatpak 1.7.1 and not found this problem.

### Steps to Reproduce

1. Try closing opentoonz without saving by menu or keyboard shortcuts after create some level.

### Screenshots, Video & Crash Logs

[dialogbug.webm](https://github.com/opentoonz/opentoonz/assets/11843239/68d17065-7d0b-436f-8d8a-fc71aa65a9b1)

### OpenToonz Version

1.7

### Operating System

Linux

### GPU

NVIDIA

### Graphics Tablet

Wacom

andeon:
I need to investigate more, but the last appimage that worked without this issue was this https://github.com/opentoonz/opentoonz/actions/runs/4859247003

Perhaps something self builted is slightly different from what is generated by the actions.

andeon:
@Bracket-H Could you tell me which version of Qt are you using to build?

andeon:
I tried building appimage with the most up-to-date qt, qt-5.15.4-focal and it didn't work. ref: https://github.com/andeon/opentoonz/actions/runs/5352120940

So I rolled back the fix patch: #4872 and it's back to working as I thought. ref: https://github.com/andeon/opentoonz/actions/runs/5371114137

Now I just can't understand why this fix works on traditional build and Flatpak and not AppImage maybe something missing or deprecated in the appimage script.

morevnaproject:
The problem is related to use of `idClicked()` signal.
This signal is available starting from QT  5.15 - https://doc.qt.io/qt-5/qbuttongroup.html#idClicked

The AppImage was built with older version of QT, this is why problem appears.

The issue is also mentioned here - https://github.com/opentoonz/opentoonz/issues/5508#issuecomment-2198465778

morevnaproject:
...so in this commit you probably should revert only places when `idClicked()` signal is used - https://github.com/andeon/opentoonz/commit/d7612d2880dd31649d70b6198c05133d998d9d1d

Possibly, we can use `buttonClicked()` signal instead of `idClicked()`?

andeon:
[...]

PR #5535, opened 2024-07-17
Update workflow_linux.yml

andeon:
Change [GitHub-hosted runner](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories) to Ubuntu 22.04, consequently Qt will be updated to v5.15 and compile correctly solved some issues like https://github.com/opentoonz/opentoonz/issues/4962 for those using Linux artifacts.

_(I think it's better to update the runner instead of using Stephan Binner's PPA and having to adjust the packages to Qt 5.15, but if someone is against it I can try to keep the same runner and make these adjustments.)_

Replaced qt5-default with qtbase5-dev as it is no longer used in 22.04

added libomp-dev package to resolve compilation lib/opentoonz/libtoonzlib.so

added libfuse2 package to create appimage

update workflow to use actions v4 

Some packages may no longer be needed, but need to be checked over time to be safely removed.

### DIFF
--- a/.github/workflows/workflow_linux.yml
+++ b/.github/workflows/workflow_linux.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   Ubuntu:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-22.04
     strategy:
       matrix:
         compiler: [gcc, clang]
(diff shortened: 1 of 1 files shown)


RodneyBaker:
No surprise that the appveyor failed here.
As this is about as 'official' as any linux build will press forward with this change.

Issue #5508, opened 2024-06-28
Version 1.7.1 requires QT >= 5.10


morevnaproject:
### Description

In build instruction (https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_build_linux.md) mentioned that minimal QT version is 5.9, but OpenToonz 1.7.1 fails to build with QT 5.9 -
```
[  3%] Building CXX object tnzcore/CMakeFiles/tnzcore.dir/__/common/tvectorimage/tstroke.cpp.o
/build/packet/linux-64/opentoonz-me/build/opentoonz/toonz/sources/common/tsystem/tsystem.cpp: In member function 'QDateTime TFileStatus::getCreationTime() const':
/build/packet/linux-64/opentoonz-me/build/opentoonz/toonz/sources/common/tsystem/tsystem.cpp:90:21: error: 'const class QFileInfo' has no member named 'birthTime'
   return m_fileInfo.birthTime();
                     ^~~~~~~~~
```

birthTime() function in QFileInfo class is available since QT 5.10 - https://doc.qt.io/qt-5/qfileinfo.html#birthTime

Please update requirements in documentation.

### Steps to Reproduce

1. 
2. 

### OpenToonz Version

1.7

### Operating System

Linux

### GPU

NVIDIA

### Graphics Tablet

None

morevnaproject:
In fact, building with QT 5.10 not helps.
It gives next error:

```
[  8%] Building CXX object tnzcore/CMakeFiles/tnzcore.dir/__/common/tcore/moc_tthreadp.cpp.o
/build/packet/linux-64/opentoonz-me/build/opentoonz/toonz/sources/common/tvrender/tfont_qt.cpp: In member function 'TPoint TFont::drawChar(QImage&, TPoint&, wchar_t, wchar_t) const':
/build/packet/linux-64/opentoonz-me/build/opentoonz/toonz/sources/common/tvrender/tfont_qt.cpp:177:43: error: 'class QFontMetrics' has no member named 'horizontalAdvance'
     int w = QFontMetrics(m_pimpl->m_font).horizontalAdvance(chars[0]);
                                           ^~~~~~~~~~~~~~~~~
/build/packet/linux-64/opentoonz-me/build/opentoonz/toonz/sources/common/tvrender/tfont_qt.cpp: In member function 'TPoint TFont::getDistance(wchar_t, wchar_t) const':
[...]

PR #5538, opened 2024-07-18
Update how_to_build_linux.md

andeon:
OpenToonz requires Qt 5.15 to support new code functions and ensure proper functionality.

This will avoid problems when the user build OpenToonz locally from source on Linux.

ref: https://github.com/opentoonz/opentoonz/issues/5508


RodneyBaker:
Thanks @andeon !

PR #5539, opened 2024-07-18
Update workflow_linux.yml

andeon:
Update workflow to generate cache correctly

I'm checking **Post Run actions/cache@v4** in workflow runs and it showed this:

> 
> Post job cleanup.
> Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.

Apparently actions/v4 requires explicit directory creation to ensure the path is available.

So I set the path:

```
 - name: Set up cache
 run: mkdir -p /home/runner/.ccache
```
I changed 'restore keys' to provides a more comprehensive set of restore keys to maximize the chances of finding a hit in the cache. These keys progressively fall back to broader matches if a specific cache key is not found.

```
 ${{ runner.os }}-${{ matrix.compiler }}-
 ${{runner.os}}-
```
and I configured 'ccache' to intercept compiler calls and cache the results, ensuring that builds benefit from caching, reducing build time by avoiding redundant builds.

```
 export CCACHE_DIR=/home/runner/.ccache
 export CC="ccache ${{ matrix.cc }}"
 export CXX="ccache ${{ matrix.cxx }}"
```

It will now store the cache properly and speed up the process of upcoming Linux builds artifacts.

### DIFF
--- a/.github/workflows/workflow_linux.yml
+++ b/.github/workflows/workflow_linux.yml
@@ -33,7 +33,6 @@ jobs:
         restore-keys: | 
           ${{ runner.os }}-${{ matrix.compiler }}-
           ${{ runner.os }}-
-          ${{ runner.os }}
 
     - name: Build libtiff
       run: |
--- a/.github/workflows/workflow_linux.yml
+++ b/.github/workflows/workflow_linux.yml
@@ -17,16 +17,23 @@ jobs:
             cxx: clang++
     steps:
     - uses: actions/checkout@v4
+
     - name: Install libraries
       run: |
         sudo apt-get update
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Will press forward with this @andeon
Please advise if the workflow_linux.yml needs additional adjustment as I cannot validate on my end.

andeon:
The cache is working again for Linux https://github.com/opentoonz/opentoonz/actions/caches now the next pulls with this configuration should generate artifacts in much less than half the time it would take.

I'm not sure if Macos is as optimized as Linux, but it should also save some time on the other pull request I submitted later, of course if everything is working well.

PR #5541, opened 2024-07-20
Update workflow_macos.yml

andeon:
Updated actions/checkout and actions/cache to v4

Created a loop to remove symlinks more efficiently. The _checkPkgAndInstall_ function has been simplified for better readability.

Changed `make -j $(nproc)` to `make -j $(sysctl -n hw.ncpu)` this command is specific to macOS.

This workflow must work with ccache support and create the required artifact correctly. I'd appreciate it if MacOs users could test the artifact.


RodneyBaker:
Pressing forward with this change.

PR #5549, opened 2024-07-28
Fix workflow_windows.yml

andeon:

Fix the workflow to generate a portable version and make mypaint brushes work correctly, this is based "windows_build.yml" and Batch Script files from tahoma2d repository create by @manongjohn .

Since this isn't working properly, I tried adjusting it another way, now the "mypaint brushes" are working as they should. I'd appreciate it if Windows users could test the artifact.

Edit: I used the GitHub web interface, I believe there could not be message text in the commit only author, so I'll use an application to add git commit --amend --author= to fix the co-author.

### DIFF
--- a/.github/workflows/workflow_windows.yml
+++ b/.github/workflows/workflow_windows.yml
@@ -9,180 +9,136 @@ on:
 jobs:
   Windows:
     runs-on: windows-2019
-    env:
-      vcpkg_ref: cd5e746ec203c8c3c61647e0886a8df8c1e78e41
-      BOOST_ROOT: ${{github.workspace}}/3rdparty/boost
-      BOOST_URL: https://sourceforge.net/projects/boost/files/boost/1.85.0/boost_1_85_0.tar.bz2/download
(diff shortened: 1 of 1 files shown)


andeon:
@RodneyBaker  Is there a way to rerun the job https://github.com/opentoonz/opentoonz/actions/runs/10133321074, this failed because it ran multiple jobs at the same time because I resubmitted --amend right after to fix the co-author.

RodneyBaker:
Yes, I can restart the github workflows.
I can't do the same with Appveyor though.

The github action build is restarted.

andeon:
I tried to update some code just to make a commit, it's strange because It's worked on the previous artifact https://github.com/opentoonz/opentoonz/actions/runs/10132928231

andeon:
Thanks @RodneyBaker, it's live after restarting on the last commit, now I believe the portable Windows Artifact is working.

I tested it extensively yesterday and today. We now know that this can happen due to a connection issue: "`Unable to connect to source 'https://community.chocolatey.org/api/v2/':`", it's rare, but it can happen now that this was observed is mapped ;)

Issue #4899, opened 2023-05-16
how do i get big pixels without loosing quality


home-gihub:
### Ask a question related to OpenToonz

I've been wanting a flipnote style animation in OpenToonz i know i just need to make the pixels bigger but when i change the dpi it results in a blurry export, i want the effect of having big pixels but on having blur maybe not changing the resolution at all.  CS gost animation is doing this.

home-gihub:
i don't want to mess with the resolution

home-gihub:
ok this is what im talking about

zoomed out:
![nozoom](https://github.com/opentoonz/opentoonz/assets/124263428/1c4ed3f3-2edf-4d10-8586-9c60f6ba56fe)
(looks not pixelated)

home-gihub:
> For the brush, check the pencil tickbox, that will make it pixelated.
> 
> As for the resolution, you don't have to change the resolution, only the resample balance.

already did that

home-gihub:
I don't know where to find all of theses setting but I would like a aspect ratio of 16:9

home-gihub:
btw i have a 3440x1440 display it might just be my screen

andeon:
Try using Raster Level instead of Toonz Raster Level


https://github.com/opentoonz/opentoonz/assets/11843239/8e2bcde6-2686-488b-a2b2-d0dbe9a8af06


Choose in Style Editor "Raster guide" 1 pixel brush, avoid using eraser, set another "1 pixel brush" to white or bg color to avoid transparency.

home-gihub:
> Try using Raster Level instead of Toonz Raster Level
> 
>  untitled.mp4 
> Choose in Style Editor "Raster tab" the 1pixel brush, avoid using eraser, set another "1pixel brush" to white or bg color to avoid transparency.

![still](https://github.com/opentoonz/opentoonz/assets/124263428/0db02d9f-2e2d-4fd7-b86d-2a0b3ed76c8e)

its still blurry

home-gihub:
i  close this now

elixsmax:
@Bracket-H Hi, do u have discord, I'd love to ask u some questions about this

elixsmax:
> Try using Raster Level instead of Toonz Raster Level
> 
>  untitled.mp4 
> Choose in Style Editor "Raster tab" the 1pixel brush, avoid using eraser, set another "1pixel brush" to white or bg color to avoid transparency.

[...]

PR #5564, opened 2024-08-11
Fix mypaint brushes

andeon:
This fixes the gray pixel displayed when using the "1pixel" brush from the experimental set https://github.com/opentoonz/opentoonz/issues/4899 and fixes the "Watacolor_Swirl*" brush from the Mojo set that was not showing the effect. 

*Use the Raster Level.


RodneyBaker:
As this change only effects the textual content of mypaint brushes and no program code will press forward with this although the Appveyor checks are still failing.

Hopefully @shun-iwasawa can get that back online and working again soon.

Thanks @andeon for the mypaint brush fix.

PR #5615, opened 2024-10-14
Run clang-format for Each Pull Request

kotauchisunsun:
I have implemented linting using clang-format to be executed every time a PR is created.
Due to the use of GitHub Actions, I have changed the location of the .clang-format file.

You can check the results of the execution at the following URL:
https://github.com/kotauchisunsun/opentoonz/actions/runs/11325031036

[summary_raw.txt](https://github.com/user-attachments/files/17363118/summary_raw.txt)

As a result of the execution, I found that many parts of the source code do not pass the lint check.
Please let me know if you have any policies regarding how these should be linted.


RodneyBaker:
This PR cannot move forward unless/until conflicts are resolved.
@kotauchisunsun has requested feedback for those willing to look into the subject of linting Opentoonz source code.

RodneyBaker:
This is an attempt to lean toward a 'policy' as @kotauchisunsun refers to above.
Some thoughts on potential policy:

1.  For saftely and to limit persmissions to those the repo specifically allows we should add:
```
permissions:
  contents: read
```

2.  We likely do not want to check everything but only those files that have changed.  I assume this would be through path filtering. 

3.  As written clang-format will run on every PR.  I believe this can be done by only running for files that have changed. (or is that the default?)

Example (not all inclusive):  
```
on:
  pull_request:
    paths:
      - "**/*.cpp"
      - "**/*.h"
      - "**/*.hpp"
      - ".clang-format"
      - ".github/workflows/workflow_lint.yml"
```

5. As a policy we use SHA pinning over tag pinning where possible.  SHA should only need to be updated when the workflow_lint.yml is updated and is best for third party actions. `actions/checkout@v4` is fine for those controlled by github.

What @kotauchisunsun has in this PR looks to fall in line with that policy with minor exceptions.

RodneyBaker:
@kotauchisunsun I am closing this PR for the present as not likely to move forward in its current state.

Amongst other issues the proposed process will run through OpenToonz source tree on every pull request.  It is my assessment that every PR may likely fail because of unrelated legacy formatting.

One possible way forward would be to implement clang-format for each PR after current source code is modernized (circa v2.0).

Short of that, adjusting the current PR's scope might allow it to move forward.
A starting place might be to target only files that have changed and exclude non-program realated documentation only sources.

RodneyBaker:
See attached development document for considerations for this and related PRs.

Note:  The link in original post returns 404.
@kotauchisunsun's repo is located at:  https://github.com/kotauchisunsun/opentoonz

[clang_format_ci_path_forward.pdf](https://github.com/user-attachments/files/31239854/clang_format_ci_path_forward.pdf)

RodneyBaker:
I am reopening this PR as it is still of importance.
The exact way forward is still to be determined.

OpenToonz’s CONTRIBUTING.md explicitly tells contributors to apply clang-format with toonz/sources/.clang-format, then run either beautification.sh or beautification.bat from toonz/sources. That .clang-format file sets BasedOnStyle: Google and then applies OpenToonz-specific overrides such as AccessModifierOffset, alignment rules, pointer alignment, and include sorting.

The practical hierarchy is:
- Google style is the base.
- OpenToonz .clang-format is authoritative.
- Beautification scripts apply that formatting across the intended source set.

While I personally do not think running through the entire codebase to beautify is something we should do we can and should consider carefully what new PRs introduce.

The question this PR asks is what we should do with clang-format with regard to each Pull Request, at what stage we should do taht and to what extent, etc.

@kotauchisunsun Your continued feedback is appreciated.
It seems to me we can just automate the process that is already in place which perhaps is the core of your request.  The secondary question being what to do with code outside the scope of any given PR.

RodneyBaker:
@kotauchisunsun I have implemented https://github.com/opentoonz/opentoonz/pull/7082 based on one aspect of your PR.  I believe the remaining work now more fully relates to your question regarding policy related to linting/beautification of code outside of a PRs scope. (Generally speaking what we might refer to as 'legacy code')

RodneyBaker:
Closing as superceded by https://github.com/opentoonz/opentoonz/pull/7082
@kotauchisunsun Please let us know about any additional concerns you have regarding clang0format and linting.  I'm quite confident we have not resolved them all.

PR #5617, opened 2024-10-15
Update macOS workflow

andeon:
Update the GitHub Actions workflow to operate on macOS 13 (Ventura) due to the discontinuation of macOS 12, which caused build times of several hours and job interruptions when building from source code and other libraries.

see: https://github.com/actions/runner-images/issues/10721

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -93,10 +93,26 @@ jobs:
             done
           done
 
+      - name: Unmount Existing Disk Images
+        run: |
+          if mount | grep "/Volumes/OpenToonz"; then
+            echo "Unmounting existing volume..."
+            hdiutil detach /Volumes/OpenToonz || true
+            echo "Volume unmounted."
+          else
+            echo "No volume to unmount."
+          fi
+
       - name: Create Artifact
         run: |
           cd toonz/build/toonz
-          /usr/local/opt/qt@5/bin/macdeployqt OpenToonz.app -dmg -verbose=1
+          echo "Creating DMG..."
+          if /usr/local/opt/qt@5/bin/macdeployqt OpenToonz.app -dmg -verbose=1; then
+            echo "DMG created successfully."
+          else
+            echo "Failed to create DMG."
+            exit 1
+          fi
 
       - uses: actions/upload-artifact@v4
         with:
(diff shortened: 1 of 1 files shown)


andeon:
I need check why it not generate the artifact 

macOS
No files were found with the provided path: toonz/build/toonz/OpenToonz.dmg. No artifacts will be uploaded.

andeon:
After digging into it, I found out that GitHub's macOS 13 runner often throws errors when making DMG files, and it’s a pretty common issue: https://github.com/actions/runner-images/issues/7522.  I took a tip from that issue to kill the `XProtectBehaviorService` process, and also added a timeout  `timeout-minutes: 90` to avoid the build running for hours. I set it to 1:30 based on the average times of previous runs, which should be plenty to get the job done.

Basically, the added script block tries to create a DMG using `macdeployqt`. If it runs into a "Resource busy" error, it goes ahead and kills the `XProtectBehaviorService` process and then tries to make the DMG again, giving it three more attempts.

I hope this fixes it for now. If any macOS users could test the artifact, I’d appreciate it!

RodneyBaker:
@andeon 
@manongjohn posted this fix to similar problem moments ago.
REF:  https://github.com/tahoma2d/tahoma2d/pull/1620
I haven't had a chance to look at it but wanted you to know.

andeon:
Thanks @RodneyBaker 

I have been testing extensively by running the jobs, and my conclusion aligns with John's: the problem is usually resolved simply by rerunning the job.

Unfortunately, killing the XProtectBehaviorService didn’t seem to solve the problem; it felt more like a placebo effect. I tested macOS multiple times in parallel using a matrix strategy, which sped up the testing process. I've now modified the code to focus solely on rerunning and increased the retries from 3 to 10. In one of the tests, it took me 7 reruns, pretty close to John, who took 6.

This ensures that users will still have an artifact to test new versions of OpenToonz on macOS.

RodneyBaker:
Because moving the runner to Mac-os 13 is important I'm prepared to merge this PR,
I had noted that @shun-iwasawa did a similar change to the IwaWarper Mac runner some weeks back so was curious when we might do the same here.

If there are no objections I'll be happy to merge.

RodneyBaker:
Pressing forward with this update.

PR #5619, opened 2024-10-18
Stylesheets Contrast Boost

konero:
No C++ changes, this completely refactors the stylesheets, rewritten from scratch. It's more robust, flexible and easier to debug. Going forward it is highly recommended to avoid using things like Less's `@extend` directive as this causes problems with the cascade, instead simply just include mixins normally. We also shouldn't point to non-existent URL's, this causes some kind of temp slowdown on Windows, in cases this was done as a workaround to hide icons or images simply use the `misc\transparent_16x16.png` image.

The main goal was to fix some contrasts issues, such as text readability, highlights and such. I had a lot of discussions over some months about all kinds of software we use and how there is this 'low-contrast' trend which looks nice on the surface but seems to be frustrating after long-term use. We'd run into issues where text was difficult to read (Cintiq's are usually angled enough that studio lights bounce right off, washing out part of the screen) and not just with OpenToonz either, but 3DSMax, Windows 11, etc..

I imagine some users might dislike the small aesthetic difference with this update, it's very small but it makes the themes more punchy, more contrasty and more saturated and defined.

I've also produced a few other themes along the way when we was doing tests, but they're very low maintenance so hopefully that's not a problem.

I didn't record a changelog, but this fixes a few issues we came across:
- text on raster level was difficult to read, it seemed the blue lightness was too high so this has been darkened a bit
- I spun the node colors back to what they were originally from an early version of OpenToonz to be darker
- highlights are brightened with boosted saturation so users shouldn't have to go 'finding' things
- the highlighted cels in Function Editor are much more pronounced, it should be easier to see selected ranges now
- viewer frame slider was changed to make it pop out, I used a bit of skeuomorphism in the design here, it seemed that despite telling users you don't need to click on the marker to scrub through frames they do it anyway and go searching for the marker to click on it, so hopefully now it's more pronounced

Files marked with an underscore like `_name.less` are files meant not to be compiled.

Try them here, close OpenToonz and download the zip, rename the `stuff\config\qss` folder from your OpenToonz stuff directly, then copy the zip contents.

📁 [qss.zip](https://github.com/user-attachments/files/17426229/qss.zip)

---

Light:
light
Neutral:
neutral
Neutral Gold:
neutral-gold
Default:
default
Default Green:
default-green
Blue (Cool Tones):
blue
Clay (Warm Tones):
clay
Synthwave (High-Contrast Specific):
synthwave


RodneyBaker:
Pressing forward wth this PR.
No program code is changed.

This is also a test to see if nightlies are back to properly generating.

PR #5624, opened 2024-10-23
(WIP) Port fix enabling tool options on Vector Select All

RodneyBaker:
A bug fix for Tahoma2D which also applies to Opentoonz to address issue using select all on a vector level not activating the attribute boxes in the tool option bar.

REF:  https://github.com/tahoma2d/tahoma2d/pull/1626

Note:  This is mostly a test to confirm that a direct transfer of that code in Tahoma2D works in Opentoonz.

I'm not seeing the co-author attributed here  so perhaps am pull requesting the wrong branch.

### DIFF
--- a/toonz/sources/tnztools/strokeselection.cpp
+++ b/toonz/sources/tnztools/strokeselection.cpp
@@ -495,6 +495,10 @@ void StrokeSelection::selectAll() {
   StrokeSelection *selection = dynamic_cast<StrokeSelection *>(
       TTool::getApplication()->getCurrentSelection()->getSelection());
   if (selection) selection->notifyView();
+
+  TTool::getApplication()
+      ->getCurrentTool()
+      ->notifyToolChanged();  // Refreshes toolbar values
 }
 
 //=============================================================================


andeon:
If you are using web editor on Github, I think you would have to put co-author in the text box below commit message, that is the Extended Description.

Therefore, in the "Commit message" field, a short description of the changes.

In "Extended Description" Co-authored-by: name <name@example.com>

RodneyBaker:
Yes, that's what I did but not in the first commit but in the second commit.
I'll spare you the details but... there is no second commit in this PR so that's likely why.
The second commit apparently didn't go through.

konero:
You shouldn't be making branches on the OT repo, they should be on your fork. You can fork [SpoonKnife](https://github.com/octocat/Spoon-Knife) and practice GitHub branches/pull requests a bit (it's what the repo is for, so don't worry about public changes).

image

andeon:
It's a simple change. You can use git commit --amend to add the co-author, but I think it's best to close this one.

Instead, create a fork of OpenToonz and rename it in the settings. Your renamed fork will stay in sync, so then create the branches to work on the functions you need.

Since you have access to the main project, this will help avoid confusion when working between the fork or when you need to work on the main project for administrative functions. Although, I think I've seen you already use the fork before to do this, so you might already have it.

PR #5639, opened 2024-11-03
Change empty shared_ptr ctor to make_shared

simstim-star:
Fix for https://github.com/opentoonz/opentoonz/issues/5598. 

Currently, it's creating a shared_ptr calling a constructor without arguments. This will result in a shared_ptr with no managed object, causing a nullpointer, as it tries to use the pointer in updateProjectFromFields right away.
Check https://en.cppreference.com/w/cpp/memory/shared_ptr/shared_ptr to see how shared_ptr() is equivalent to shared_ptr(nullptr) (methods (1) and (2) have the same behaviour).

Simply changing it to `auto project = std::make_shared<TProject>();` should be enough.


RodneyBaker:
Will press forward with this fix although appveyor and Mac checks likely will still be failing.

PR #5659, opened 2024-12-01
Export all levels

### LINKED ISSUE #5658: Export All Levels (导出所有层级)
# Want a command to Export All Levels to save time.
图片

## 0. "Export All Levels..." does not export levels which NOT VISIBLE in camera
图片

## 1. While exporting the File Name would be automatically fill with the wait-to-export level name(use part of column name if exist)
图片

## 3. "Export All Levels..." exports level one by one,until all levels have exported.
And with a skip buttom to skip the level that doesn't need to export.
图片

## 4. "Export All Levels..." provides a checkbox that enable to export all levels at one time.
图片

#### Add Create Folder Option,which create a folder named by the name of "File Name",and saves the output files
图片
#### Add ext tlv if only have Toonzvector and Toonzraster)

### LINKED ISSUE #5662: Wrong behavior of "Export Level" when selecting mutiple columns
### Description

The command **ExportLevel** is incomplete

### Steps to Reproduce

1. Select more than one columns
2. Export level
3. the mainwindow columns may be dis-selected AND
4. could not export if not typed something

https://github.com/user-attachments/assets/d9a2157f-5c50-4953-ae83-0528a2f94a5d

### Expected Behavior

_No response_

_No response_

### OpenToonz Version

1.7

_No response_

### Operating System

Windows

### GPU

AMD

### Graphics Tablet

Wacom

### LINKED ISSUE #5669: [File Browser For Save]Wrong File name update behavior when select a folder
### Description

https://github.com/user-attachments/assets/73e19429-3d76-49a9-a669-3f9eec7a2dca

### Expected Behavior

Do not update File name field if not selecting a file.

Abel032:
This PR resolves #5658 
And fixes #5669 fixes #5662

# About this feature
You can find this command from here:
图片

This feature allows user to export all Raster,ToonzRaster and ToonzVector Level in the xsheet,
but only levels that could be seen in camera can be exported.
a6e4d265f382c550bbacb92a26ce57b6

You can choose to create folder for every level.
图片

You can export all levels at one time or export them one by one.
图片
图片

The level's file name would be automatically filled with level name by default , no matter export all one by one or export all at one time.But if you export one by one,you can change every level's filename.
图片

If a column name is specified , while exporting ,that column's level's filename would be automatically filled with the words before first underscore ,and remove all "#" from column name.
图片
图片

If column name includes a number 1 or 2 level's filename would add "_1gen" or "_2gen",which means “第一原画” or "第二原画".
图片

Export would start from a Vector Level (if exist),so you can adjust vector options in export options tab before exporting.
QQ20241225-13222-HD

If you choose to export with folder to a folder which doesn't have any folder, the file browser won't update.
But the levels are already exported.


RodneyBaker:
@Abel032 Thanks for your efforts thus far.  It's good to have you with us!
I understand this is still very much Work in Progress/Draft so disregard the following as necessary:

Unless there is a specific reason to have code/comments in another language... please use English.

Not that this is anything official but in general the vintage of code can be determined by the language of commentary in the source code:  Italian (Digital Video original Toonz commentary), Japanese (Studio Ghibli interim development), English (modern usage/commentary).  The primary exception to this is of course the translation files where each language is represenative of common content; specifically for the GUI).

We probably do need to extend the areas for building and customization in various languages so that is something we should consider moving forward.  Here I'm specifically references files such as .bat/batch files which would be used locally by the user/developer.

Abel032:
https://github.com/Abel-8/opentoonz/blob/master/toonz%2Fsources%2Ftoonz%2Fexportlevelpopup.cpp#L414-L419

Abel032:
This PR is ready for review and test.
It would be nice if anyone sends test report to me.

The codes and comment will be further sorted in the future. (maybe after a month)
Go ahead merge if needed.

RodneyBaker:
The scope of this feature addition PR might be a bit broad touching 5000+ files.
The additional in this PR of carotene as a new third party library might have been unintentional?

Abel032:
> The scope of this feature addition PR might be a bit broad touching 5000+ files. The additional in this PR of carotene as a new third party library might have been unintentional?

Yes ,it includes third party files.
didn't use git in VS before
Trying to fix it...

Abel032:
still problem……

Abel032:
Seem to be OK ,but why MacOS Build faild

Abel032:
BUG: Don‘t update browser when exporting with folder

Abel032:
- [] Wrong level counting when export levels one by one

Abel032:
### About Can't update node when create new folder in an empty folder.

This problem should be solved in another PR.

When new folder button pressed,the folder_node refreshes twice,one is because folderchanged,another is by button's slot.

Abel032:
Is it better if I move the create folder function from exportlevelcommand to filebrowser?
That would be better...

Abel032:
- [ ] Refactor: move move the create folder function from exportlevelcommand to filebrowser.

Abel032:
Sorry for the mess…
I need to do all tests done and squash before commit...
And split the PR...
learned from this

Abel032:
Test List:

### 1. Export All Levels without Folder
- [ ] at one time 
- [ ] one by one

### 2. Export All Levels with Folder
- [ ] at one time 
- [ ] one by one

### 3. Export All Levels with Folder
- [ ] with empty column
- [ ] with empty cells before column
- [ ] with not level column(audio, assistants, note,scan)

### 4. ...

toonz/sources/toonz/exportlevelpopup.cpp
```diff
@@ -111,50 +112,13 @@ struct MultiExportProgressCB final : public IoCmd::ProgressCallbacks {
 
 }  // namespace
 
-//********************************************************************************
```
Abel032, line 114: Move to exportlevelpopup.h

toonz/sources/toonz/exportlevelpopup.cpp
```diff
@@ -418,7 +411,6 @@ void ExportLevelPopup::showEvent(QShowEvent *se) {
   //          filebrowserpopup.cpp...
 
   updateOnSelection();  // Here the selection is used
-  updatePreview();
```
Abel032, line 421: moved to line 571

RodneyBaker:
Congrats on getting the Export All Levels process to this point.
I'm able to successfully export levels.

I'm running into 2 problems, one of which is completely unrelated to your PR.
That is a failure of the browser listing to update so that Opentoonz must be reopened in order for exported files to be recognized as being present.  This is outside the scope of your project but will tend to make it seem like the export process has failed.

The second problem is a bit more elusive and doesn't happen all the time.
The end result is that Opentoonz is frozen as if there is a dialogue window still open and the program must be manually shut down.  This might relate to sometimes where the Export All dialogue sometimes remains open (which I like BTW as that allows for exporting to multiple formats easily) and at other times closes.  I think... not sure... it might relate to exporting vector levels.

At any rate, this is just general feedback.  Keep up the good work!

Abel032:
> The second problem is a bit more elusive and doesn't happen all the time. The end result is that Opentoonz is frozen as if there is a dialogue window still open and the program must be manually shut down. This might relate to sometimes where the Export All dialogue sometimes remains open (which I like BTW as that allows for exporting to multiple formats easily) and at other times closes. I think... not sure... it might relate to exporting vector levels.

To see if that happen in my machine.

Abel032:
Feature description added.
And ensured dialogue to be NonModal when it hides.

Abel032:
New Bugs:

- [ ] Stuck by Modal windows after export
- [ ] Can't show alpha slip other than choosing png again

Abel032:
Hi @RodneyBaker   ,

Will the modal bug occur in the new build?

RodneyBaker:
I'm not familiar with the modal bug but my sense is that it will but only if the fix from ```[5ff3b07](https://github.com/opentoonz/opentoonz/pull/5659/commits/5ff3b0756e394f94c77e64d64dd7999ea809b346)``` didn't entirely fix it.

Abel032:
Let’s see if it still occurs...
I would use this feature in recent two weeks frequently.

RodneyBaker:
Some feedback...

I have tested current PR as of this date and like the ability to export to directories/levels named after the columns.  This not only is useful but helps to encourage users to organize and name their assets so that's a great move in the right direction workflow-wise.

I did have some initial difficulty getting the Export dialog to appear.
This happened several times as I questioned whether I had simply forgotten if I needed to select the levels in the xsheet or the columns.  
I resorted to trying the regular Export Level and the dialogue appeared.
When going back to selecting multiple levels and trying Export All Levels the dialogue then appeared and worked as expected.

I confess I was saddened to see TLV wasn't in the dropdown options but understand the difficulty in converting from full color raster to TLV and why it needed to be dropped.  Hopefully in the future we'll be able to streamline and improve that full color to TLV process.  Definitely outside the scope of this PR however.  For a moment I enjoyed seeing that TLV option as I caught a glimpse of the future.

I'm looking forward to seeing this move forward as it does add new capabilities beyond that of other OT processes such as OCA export.  As it adds an option to the menu users would expect to find that's also nice.

I do want to see a little more testing to confirm if/why the dialogue doesn't always appear.
I assume this likely relates to your observation relative to a modal dialogue windows bug.

Abel032:
Thanks for the feedback!

I thought Modal Setting is not so necessary, so removed it.
This may solve problem.

Abel032:
> Hopefully in the future we'll be able to streamline and improve that full color to TLV process.

I use cleanup function to do this,although it crashes frequently.

But what you said might be different thing with the cleanup?

RodneyBaker:
I use cleanup function to do this,although it crashes frequently.

But what you said might be different thing with the cleanup?
Like converting without losing colors(because we use cleanup to keep only few colors)

Yes, cleanup is limited to few colors but there are also other limitations.
My take is that cleanup is for transferring... linework.
As such that gives us a good framework for when to use cleanup versus other methods such as conversion.  Other limitations include the difficulty of potentially having millions of colors whereasa TLV and PLI palettes are limited to a little over 4000 individual colors.

With conversion a full color conversion requires two images... one for the lines and one for the fills/color.  An updated option might use thresholding to extract the lines out of a color image and use that as the second (linework) image.

At any rate, more research is needed to determine what the challenges and shortfalls actually are and what the best way forward should be.  If it had been straightforward I expect they would have already added that streamlined full color conversion option.

konero:
Menu icons should be 16x16, here's a fixed version: [export_all_levels.zip](https://github.com/user-attachments/files/18311903/export_all_levels.zip)

toonz/sources/toonz/exportlevelpopup.h
```diff
@@ -86,52 +94,55 @@ class ExportLevelPopup final : public FileBrowserPopup {
...
+//    ExportOptions  definition
+//********************************************************************************
 
 class ExportLevelPopup::ExportOptions final : public QFrame {
   Q_OBJECT
 
 public:
   ExportOptions(QWidget *parent = 0);
-
+  bool pliOtionsVisable;  // be used when updateonselection,ExportOptions show,
```
konero, line 124:
```suggestion
  bool pliOptionsVisible; // indicates visibility of PLI (vector) export options based on current selection
```

Would also need to be updated elsewhere.

toonz/sources/toonz/exportlevelpopup.cpp
```diff
@@ -387,6 +350,37 @@ ExportLevelPopup::~ExportLevelPopup() {
   QWidget().setLayout(layout());  // Reparents all layout-managed child widgets
 }  // (i.e. all) to a temporary
 
+void ExportLevelPopup::GetSelectedSimpLevels() {
```
konero, line 353:
```suggestion
void ExportLevelPopup::collectSelectedSimpleLevels() {
```

I'd change this from `simp` > `simple` to match `TXshSimpleLevel` wording for readability.

toonz/sources/toonz/exportalllevelspopup.h
```diff
@@ -0,0 +1,39 @@
...
+private:
+  bool isexport_all;  // set to false when "Export All" be dis-checked
+  int level_exported = 0;
+  std::map<std::wstring ,std::wstring> level_to_foldername;
+  std::wstring backfoldername(std::string colname, std::wstring levelname);
+  bool isAllLevelsExported();
+
+  // Widgets
+private:
+  DVGui::CheckBox *m_exportall;
```
konero, line 25:
```suggestion
  DVGui::CheckBox *m_exportAll;
```

Abel032:
Thank you @konero

Abel032:
Works well in my work.
I think it could be merged.
Thank you for reviewing,@konero @RodneyBaker

Abel032:
The newest commit [adcddaa](https://github.com/opentoonz/opentoonz/pull/5659/commits/adcddaa6c7c941cc635acedfd7634fb9185f3ae8) is the implementation of PR #5678.
I can do it simply by add a Option,so added to this PR.

RodneyBaker:
This PR appears to be working as expected.

I'm still testing here and did run into a crash although I haven't been able to reproduce it.
At the time I was checking to make sure export from vector to tlv was working (which it is) and was also checking to see that the option to export to tlv was not available for raster and toonz raster levels.  I don't think that particularly related to the crash as I had a lot of other things going on at the same time.  

I believe all of @konero 's suggestions were implemented.
We'll defer to @konero for confirmation.

In my estimation the real superpower of these feature (besides the obvious one of being able to export all levels) is the name transfer from the level to the directory.  That will help a lot in organizing project assets and adds additional benefit to workflows that rely on level naming.  In short, it encourages good practices which is beneficial to all users

Additional thoughts
There is a fairly wide range of differences between Level 'rendering' (Exporting) and Scene rendering (merger of all levels though Rendering) and this PR has helped me appreciate those differences even more.

Add to that export options like OCA Export/Import and Multiple Image rendering options and we have a lot of similar output options that vary just a little but in significant ways.  It would be good to document those differences and perhaps even consolidate the options/GUI so the user can more easily know about those options and use them as appropriate when needed.

Finally, I don't think I fully realized before this PR the benefits of the Export Options settings for vector levels.  This PR's addition should help others discover those benefits too.

Abel032:
> run into a crash although I haven't been able to reproduce it.

My OT stuck after I delete the folder just created(No matter what kind of export ext),
wait it about half minute OT returns to work,
but if open Export All Level Popup again,OT crashes.

This can be reproduced 90%

RodneyBaker:
That reads like another problem with the refresh of Opentoonz file management where it constantly checks for valid files and (if found) updates any related data).

I am curious if the crash you indicate will happen less often if that option is turned off in Preferences:

image

Of course, if turned off... Opentoonz likely won't see changes as that is the whole point of having it turned on.  As such, files will update but Opentoonz browser and dialogue windows won't know that because the Opentoonz File Watch System is off.

Abel032:
Well...back to normal after turning it off.
Althouth user need to refresh folder by hand without it.

RodneyBaker:
Thanks for confirming my suspicion.
What this suggests is that we need to take a deeper look at Opentoonz File Watch System as it has a number of issues, particularly with refreshing but... as you've encountered... also that contributes to freezing and crashes.  It may be that some minor tweaks can resolve the vast majority of these issues.  Perhaps particularly by delaying other processes during critical data refreshes.  How to do this with minimal interuption of the user is a primary concern.

Abel032:
> My OT stuck after I delete the folder just created(No matter what kind of export ext),
> wait it about half minute OT returns to work,
> but if open Export All Level Popup again,OT crashes.
> 
> This can be reproduced 90%


I found it is caused by #5717.
Not related to this PR.

Abel032:
These merge order of these PRs doesn't matter
#5730 #5723  #5718 #5717 #5659

and they should not conflict with each other.

Abel032:
Some bug of Export Level would also be fixed in this PR,
because Export All Levels based on it, so just share same PR.

RodneyBaker:
Curious if you have plans to adjust this PR further @Abel032 

No rush to merge here but also no desire to leave PRs waiting when they can move forward and be shared by everyone.

Abel032:
@RodneyBaker 

It's okay to merge.
I think the features have already implemented, and bugs also fixed.

RodneyBaker:
Consider it done.

Thanks again for your effort to improve Opentoonz.

PR #5685, opened 2024-12-24
Modify XDTS I/O to Support Keyframes and Reference Frames Symbols

shun-iwasawa:
This PR will allow OpenToonz to correspond the keyframe and the reference frame symbols to cell marks when importing/exporting XDTS.

Please note that the Keyframe and the Reference Frame symbols will be exported in an [unofficial format](https://github.com/shun-iwasawa/XDTS-uext), which may not be displayed correctly or may cause errors in applications other than [XDTS Viewer](https://github.com/opentoonz/xdts_viewer).


RodneyBaker:
Ready to proceed with merging but I'm not confident the Windows nightly will build.
I guess there is one way to find out.

PR #5709, opened 2025-01-12
Update how_to_build_linux.md

andeon:
Removed qt5-default it's which has been dropped upstream in Debian, the base libraries qtbase5-dev qttools5-dev-tools should be enough to pull in the other dependencies, otherwise we will add some more after.


RodneyBaker:
Thanks @andeon

PR #5717, opened 2025-01-14
Fix Failure when updating the folder node

Abel032:
## Issue Description
If you create new folder in a folder which has no other folder,then file browser won't refresh properly.
This PR would solve this issue.


RodneyBaker:

Abel032:
This PR may cause crash when delete all folders from a folder..
But I can't understand why now.

Abel032:
The updated commit would avoid crash when deleting folder from outside.

But the function **new folder** don't show entire feature,
now layout would return to parent folder before user type new folder's name.

Abel032:
TODO:
- [ ] Recovery full **New Folder**
- [ ] Avoid long wait time when refreshing

RodneyBaker:
Thank you @Abel032 for these PRs updating and improving Opentoonz response to files and folders.  Very much appreciated.

PR #5718, opened 2025-01-14
Optimize Adjust Current Level to This Palette

### LINKED ISSUE #5676: disk tlv data won't update after studio palette be adjusted to level
### Description

disk tlv data won't update when studio palette be adjusted to level

### Steps to Reproduce

https://github.com/user-attachments/assets/440e70b5-17fb-42c8-bef6-005a75dfca57

### Expected Behavior

https://github.com/user-attachments/assets/6ce98c1e-e017-4137-ae18-0b2026a4f1fc

Abel032:
This PR would Resolves #5676
Set current level to dirty so that line/area which have different style index but it's color is included in studio palette can be saved properly.

### DIFF
--- a/toonz/sources/toonzqt/studiopaletteviewer.cpp
+++ b/toonz/sources/toonzqt/studiopaletteviewer.cpp
@@ -753,6 +753,11 @@ that the last operation is the icon refresh...
                                            m_currentLevelHandle,
                                            apd.getTolerance());
 
+  m_currentLevelHandle->getSimpleLevel()->setDirtyFlag(true);
+
+  //TApp::instance()->getCurrentLevel()->notifyLevelChange();
+  //No need to notice user
+
   TUndoManager::manager()->add(new InvalidateIconsUndo(m_currentLevelHandle));
 
   TUndoManager::manager()->endBlock();


RodneyBaker:
Thanks for refreshing of your current PRs @Abel032 
Automated checks for all appear to be passing.

I'm currently testing all PRs in the queue but need to refine that testing process more to make sure we don't miss anything in the process.  It's always likely something will be overlooked but better if we catch it first time around.

This PR appears to be straightforward enough.
At first I didn't think the comment necessary but actually appreciate it more now.
Technically, although trivial, the following:

`//Not necessaryto notice user`

should read:

`//Not necessary to notify user`

It could take on other forms as well such as:

`No need to notify user`

or, the comments could be dropped entirely.

For the record, I personally would not delay merger for commentary alone.
As others look through the history of the code, comments such as these are useful for learning why we prefer one thing versus another.  Is it nice to have even the comments perfect?  Sure.  But we don't live in a perfect world.  Someone that sees an opportunity can submit a PR to continue to move the code and commentary forward.  And in that process we learn even more.

A possible suggestion for those using github might be to consider using copilot (or similar systems) to scrub the code not specifically for the code's sake but for accurate (and terse/precise) commentary.  It tends to be quite good at commenting code using standardized terminology.  Kind of like autocorrect in a word document.  Useful for translation as well.

Abel032:
Agree.
Testing and reviewing are always important.
It’s advisable to allocate more time for these steps before moving PRs forward
As time passes by, we continue to learn more and test more, which ultimately enhances the quality of PRs.

(PS) I've opened multiple PRs recently, and it's fine to hold them for further testing and review.

Abel032:
Corrected comment，and thank you @RodneyBaker !

RodneyBaker:
Close enough on the comment @Abel032 
LGTM

PR #5719, opened 2025-01-15
Export Xsheet to CSV

### LINKED ISSUE #5675: Export timesheet(xsheet) to csv(just like csp do)
Add a command to export timesheet to csv.
similar to "export xsheet to pdf"

Abel032:
This feature is implemented basing on existing function **Export Xsheet to PDF**.
图片

And would Resolves #5675


RodneyBaker:
Works great.
Simple.  Effective.

RodneyBaker:
I should add that this exported CSV frame information can be reimported as keyframe values and use as File Interpolation (in the Function Editor) if the CSV file has the following two changes made:

- Rename the file from .csv to .dat (So that File Interpolation can see the file)
- Remove all of the quote marks in the file (File interpolation requires raw data like:  1, 2, 3 without quotes).

PR #5723, opened 2025-01-18
Fix update folder delay

Abel032:
## Issue Description
If you delete folder outside the file browser or click on a .lnk file in the file browser,
it would cause OT to an over 15s freeze(depends on your machine and network)

This PR would resolve the issue by avoiding unexpected refresh of Network Node.

### DIFF
--- a/toonz/sources/toonz/filebrowsermodel.cpp
+++ b/toonz/sources/toonz/filebrowsermodel.cpp
@@ -1209,17 +1209,20 @@ DvDirModelNode *DvDirModelRootNode::getNodeByPath(const TFilePath &path) {
 
   // it could be a network folder
   if (m_networkNode) {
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I am still testing this PR with a focus on determining if the delay still exists if the directory is deleted from inside the Browser.

RodneyBaker:
Pressing forward with this.
We'll need to do a scrub of related areas to see what (if anything) is still not operating optimally.
This set of PRs related to file access and refresh of the UI tackle some long standing issues and is greatly appreciated.

RodneyBaker:
@Abel032 I did mean to follow up on previous comment where I said I was doing additional testing based on some delay I saw deleting files/folders via Browser.
I'm happy to say this PR resolved that.

PR #5729, opened 2025-01-20
update Adobe Animate keybind presets to be closer to Animate/Flash

ninjamuffin99:
- `V` was changed from the simple selection tool, to `Rectangular` selection specifically
- Added `L` to be bound to `Freehand` selection, which is equivalent to Flash/Animate "lasso"
- Added `Shift+L` to be bound to `Polyline` which is equivalent to "Polygon Tool" selection
- `F5` will now increase step of the current cell, similar to how F5 will extend current frame. 
- `F6` will create a duplicate cell, similar to flash


RodneyBaker:
@ninjamuffin99 
Appveyor check appears to have just timed out.
The change to the Adobe Animate keybind preset is trivial enough and doesn't touch the program proper so...

Posting intent to merge although it isn't cool that your first addition will have a failed appveyor check.  I'll delay for a moment to see if you want to refresh the PR (close and reopen usually does the trick).

ninjamuffin99:
seems like it timed out again on the Release build! however Debug build on AppVeyor seems to be green

RodneyBaker:
It was worth a shot.
Let's proceed.

Thanks for updating the Adobe Animate keybind presets @ninjamuffin99

PR #5730, opened 2025-01-20
Fix Focus Lost when renaming new created folder

Abel032:
## Issue Description
When **Watch File System and Update File Browser Automatically** is enabled in Preference.
Open File Browser and click New Folder Button ,
would see LineEdit appears but remains no time for user to type.


RodneyBaker:
This has been a pesky problem and will be glad to see it gone.
Posting intent to merge.

Issue #5724, opened 2025-01-18
Rework the set of vector brushes


andeon:
I'm looking at the brushes in the Vector Brush section and I don't see much use for most of them, as they don't work well, even as stamps. So I'm proposing to remove most of them and keep a few that can demonstrate the features, like intertwined_waves, scratch, otlogo and simple_lace.

Also, I created a few with tapered line effects to help users, especially those who use a mouse. Maybe I'll add 3 or 2 , so they don't run into the same problem I'm reporting. I also reworked the scribble_chalk brush; now it doesn't overload  and crash OpenToonz when editing the brush file.

![Image](https://github.com/user-attachments/assets/641a8765-d205-437e-9ed0-4de96a362c78)

It may seem drastic, but honestly, I can't see any use for most of them. So I'd like to discuss this before submitting the pull request, maybe gather some ideas and rework it. But avoid suggesting brushes that are covered by generated

PS: Editing and creating these types of brushes can be tricky. Sometimes it can cause memory overflow, resulting in a blue screen and requiring a computer restart. There appears to be a problem with reading or creating the file, depending on how it is baked, and it needs to be investigated.

RodneyBaker:
An update of the default vector brushes in Opentoonz is long overdue.
Thanks for getting the discussion going.

The only vector brushes I tend to use are those that I create myself.
The current defaults tend to be a little niche for specific (drawing) uses.

Legacy Brushes
If removing current default brushes my suggestion would be to zip up the current brushes and place that zip file in the PR.  In this way users can download those to support any legacy usage.  I'd say just place the zip file in the directory but it may be best to make a clean break/distinction between the old and new.

Creation of new brushes
There are a number of factors making creating custom vector brushes difficult.
Other than that it can be easy to create a malformed brush which will crash Opentoonz I'd say the main issue is that brushes are too small in scale when applied to default size vector strokes.  As such when applied they aren't very useful unless/until the stroke is scaled up considerably.

File size
The good news is that vector brushes tend to be quite small with most being under 10KB.
Only 2 brushes are above 50kb and one of those is Skribble Chalk.
Interestingly, the Skribble Chalk brush is the largest at 107kb
In the past I have wondered if size was a determining factor is difficulty uses custom vector brushes but I've been creating entire characters and character parts as brushes without issue so that isn't the determining factor.

Malformed brushes
Where the ends of the vector strokes are placed (margins at the end of the brush) are important as well for when vector lines are edited later via Control Point Editor.
[...]

Issue #3863, opened 2021-04-05
Is possible a Fx Smooth node?


F-Burning:
Could be possible to create a Fx copying and pasting the code of this option?

![image](https://user-images.githubusercontent.com/74098734/113578275-75868b80-961a-11eb-9b4b-4f3d9f7340e8.png)

This would work like OLM plugins

![image](https://user-images.githubusercontent.com/74098734/113578424-b7afcd00-961a-11eb-94a9-36ea344ba7ce.png)

Why create a node like this?

-Because you could add a simple node instead go level by level change each configuration
-Because apply the smooth before the compositing is like building the house from the roof
-Because you could add several smooth to a compositing and get a better control of jaggy

WolfInABowl:
I'd love an anti-aliasing node, there are many FX techniques that purposefully create aliased edges then smooth them out. 
Also, some nodes only produce aliased results such as `Posterize`, you need to render an image sequence to have the ability to smooth it out. You could resort to using `Blur`, but that doesn't do a very good job.


On a related note:
After Effects is the industry standard compositing software used in the anime industry; it works on a much more granular level as it is designed to work with regular pixel data. There are plugins to blur colors together, as well as numerous other useful things such as generating lines between color borders. These are applied to aliased images which are then smoothed via anti-aliasing.
Compared to OpenToonz, this allows for a much greater degree of creativity as you are able to work with dynamically generated image data, rather than being restrained to the palette defined in the level. 
E.g. you can easily use `PaletteFilter` to key out areas of color, but if you wanted to dynamically generate new areas with new colors, then `PaletteFilter` is unusable for that purpose.

As OpenToonz applies its anti-aliasing *before* any FX nodes, effects designed to work in the same manner as they do in After Effects would not be possible as they need to work on aliased images. 
The addition of an anti-aliasing node would open the door for future nodes designed to work on any kind of image instead of just ToonzRaster/ToonzVector as the user can choose when the anti-aliasing is applied.

andeon:
I agree that this workflow is much more cleaner, makes more sense for many other types of work or with slightly more complex scenes.

The Fx mode could be good way since you don't need to change the workflow where the Level is used, as I believe that many artists already use this process for anime:

https://youtu.be/lTwMjbD2adQ?t=1247
![Dong Chang - Head Turn](https://i.ytimg.com/vi/lTwMjbD2adQ/maxresdefault.jpg)

F-Burning:
[...]

PR #5733, opened 2025-01-21
Refactor Vector Brushes and Cleanup

andeon:
Tapered line effects are very useful, especially for users who use a mouse, as they allow for quick application of these effects to lines, streamlining the workflow. I am proposing to remove most of the brushes, as they have little practical application, while keeping a few that create effects useful as demonstrations, so users can understand what is possible to achieve. The brush tool lacks parameters to offer sufficient variety, limiting the ability to refine settings and create variations through parameterization.

The brushes have been moved to the 'toonz vector brushes' subdirectory to support any legacy usage. They may be permanently removed in the future.

I redesigned the scribble_chalk.pli and renamed it to sketch chalk.pli (just to preserve legacy). It now has smoother, less jagged lines, and OpenToonz only experiences slowdown when there are dozens of lines. Before, it would overload and crash OpenToonz when editing the brush file. Just so you know, in the 'Generate' tab, the Gouache brush can better simulate a Scribbled Chalk effect, which can be animated (https://opentoonz.readthedocs.io/en/latest/managing_palettes_and_styles.html#animating-palettes-1). Not only the color but also what you see in the 'Settings' tab can be animated as well.

Ref: https://github.com/opentoonz/opentoonz/issues/5724


RodneyBaker:
Note:  This will be a breaking change for any projects that reference the old vector brushes as they will no longer be located in the same place.  Options will be to either restore the files to that location or point the reference to a proper location.

Users should keep this in mind as they move from older releases to new.

RodneyBaker:
I think we do need an arrow of some sort in the new brushes.  This would make it trivial to add a pointer to some element of the canvas.

Arrows can easily be animated as well by moving the vector strokes... erasing one end or both... and the stroke can be autoinbetweened.

image

andeon:
This arrow looks good to me, could you add it to set?

RodneyBaker:
I don't have that arrow any more.  
That's from a few years ago.

andeon:
Okay, I'm gonna try to recreate this. Or do you have another example you think is better?

RodneyBaker:
No, I don't have anything I'd feel comfortable adding to the brush set at this time.
If necessary I can create something and let others decide.

For the record, I'll always feel more at ease if resources are coming from you or other Opentoonz users as I'll likely be the one merging the PR.

andeon:
The style is black by default, but it allows you to have different fills and outlines by altering the little square boxes available beside the Auto and Apply buttons.

ot_vector

@RodneyBaker, is AppVeyor unavailable? If it's not working, I’m going to revert the last config change regarding the timeout.

RodneyBaker:
Yes, let's revert and see if that helps.

Thanks for the arrow(s).   That's exactly what I had in mind.

RodneyBaker:
Posting intent to merge although I'm a little concerned that the appveyor checks were dropped.
Given that this is just staic file replacement it shouldn't be an issue but if the nightly doesn't build the change won't be available until the next  PR is merged.

andeon:
I'm going to close and reopen it to check if it starts.

RodneyBaker:
LGTM

RodneyBaker:
@andeon  Just a heads up that I'm looking into a gap at the end of the vector brushes that I have not yet confirmed weren't/aren't there in the originals.

This can be seen expressly when drawing a circle or closed shape as the ends will not meet.
This is somewhat to be expected with tapered lines but brushes that have both ends the same size should meet as if joined.

I didn't see the gap while testing and I don't recall the gap being there before this PR.
I expect that testing earlier releases the vector brushes will have this gap and I just hadn't noticed it before.

I'll post images showing the gaps.

andeon:
This problem has happened before. One way to avoid it is to remove the outline, but this makes the line look rough and can also cause issues at the ends of some lines. However, I think I can work around that last one. Since the line will mainly be used for inking, I didn’t worry too much about it. But what do you think about this? 

What is the best output configuration to make the line very clean and smooth, so I can check and test if the issue I’m seeing is related to the viewer or not?

RodneyBaker:
> What is the best output configuration to make the line very clean and smooth, so I can check and test if the issue I’m seeing is related to the viewer or not?

I've never been able to lock down specifics although I lean toward a few 'rules' that tend to produce better results.  Some of those (with varying degree of potential merit) include:

(Unconfirmed) Rules for Creating Custom Vector Brushes
- The margins around a brush image are important.  If imagery goes too far outward adjusting control points and biases will break the brush image.  If not far enough outward gaps will appear
- Do not use the color styles 0 and 1 for any brushes.  Better to create new styles and use those.
- Set UI guides to allow exact placement of end points and to establish middle of brush stroke
- Keep names of brushes simple
- Keep the size of the brushes small
- Consider using 2 control points in a simple brush to attempt to achieve the desired look and expand in complexity from there.   Adjust the bias handles to effect curvature rather than add additional control points.**
- It is better to use a freehand shape than a space generated by Geometry Tool***
- Closed shapes (using Tape Tool) are better for shapes containing color than open shapes whose ends just happen to be in the same location allowing the shape to fill with color.
- Use a different style (and not style 0 or 1) for outlines versus areas/fills.
- When creating Text brushes (Text on a Path)... use guidelines to place the text
- Consider using Tahoma2D's Align and Distribute Tools to get 'perfect' distribution and alignment (That feature set is not yet ported to OT)
- Be aware that anything that potentially will corrupt a PLI level will just as easily corrupt creation and use of brushes


There is more.
I'm likely forgetting some of the more important/best practices.
Will add them and adjust these as necessary.
Some listed are very like just superstition and have little or no merit.  It just seemed to work (or not work) at some given time.

** Really need to test this and get factual feedback
*** Some problems with brush creation potentially stem from use of Geo Tool.  While unconfirmed its easy to bypass this potential by simply creating the shape from scratch.  Other issues encountered with Geo Tool shapes suggest calculations of end points might be different.

Regarding testing itself... consider using the Appveyor build rather than the Github Action build s it will (potentially) behave the most like the nightly as the Windows nightly is created via Appveyor.  You (@andeon) are familiar with some differences with mypaint brushes and that's a good example of how results of builds an vary.  The Appveyor build is now portable which saves time manually turning it into a portable.  That portable tweak is thanks to @Abel032 if I recall correctly.

RodneyBaker:
Here's a quick video to show the current variable width (breadth?) of vector brushes.
The small width of the drawing brushes likely accounts for the gaps when drawing (and joining) closed shapes.

Obviously, frame by frame this.  ;)


https://github.com/user-attachments/assets/7ab9f0a9-a4f5-468e-8c10-b043e84e49c8

andeon:
I've removed the outline from the 'ink spike' brush."

brush

However, the line appears a bit rough. I could adjust the entire ink set to work in the same way if needed.

> What is the best output configuration to make the line very clean and smooth, so I can check and test if the issue I’m seeing is related to the viewer or not?

I'm referring to the output in rendering. Technically, the result won’t be a vector but a raster, so I wanted to test different configurations to see if there would be a noticeable difference between brushes in the output.

RodneyBaker:
Yes, I've long wanted to look at the process whereby the vector line and fill gets 'rasterized' but still assigned to a PLI vector file.  I'm confident we should be able to find some optimizations in that process that keep the result sharper and crisper.  

Is this mostly just an antialiasing result of the process?
That would be my initial guess without knowing more of that process.

Regarding outline/lines of zero thickness...
There are several issues to contend with here.
I'd say one of the most interesting to me personally is that if a vector line has any variable thickness then it cannot be zero.  I'd share more but at this exact moment I'm testing in a build that doesn't allow pen pressure...   always something.  ;)

I guess what I'm saying here is that if we have a vector line that is zero thickness and thicken any part of that line (via pump tool)... no part of that line is allowed to be zero thickness any more.  I find this to be problematic.

This alone might account for shifting of pixels during rasterization and/or antialiasing especially in extreme edge cases such as when creating vector brushes (and use of Plastic Tool meshing).  It would at a minimum set the stage for rounding error during the conversion process.

RodneyBaker:
Just a random thought...

If the vectors are being 'rasterized' anyway as part of the process of assigning the images to vector files that are useful as brushes... why not also allow raster images to be used/converted as vector brushes?

andeon:
It'd be a cool feature, similar to the ToonBoom Textured Brush, like in this video:  https://www.youtube.com/watch?v=-Alk8tlvrgo

When MyPaint was introduced to work with the geometric tool, I imagined it could function at a vector level, similar to how texture work with vector. I'm not sure if this would be difficult to implement.

andeon:
@RodneyBaker I'm doing some more testing. I think I can get around this by adjusting the brush size while keeping the thickness low, so when it's rendered, there won't be any gaps like is showing in the viewer when zoomed in. I'll test at resolutions higher than 1920x1080 to see the render stays consistent. But it won't work for all brushes as the thickness needs to increase or decrease along brush to keep their the shape like the scratch.

RodneyBaker:
Thanks @andeon!

Future work:  Something I hope to investigate is how we can automatically trigger an increase in size (or pass a desired size) so that pretty much every vector brush we draw with doesn't have to be upscaled in order to see the details of the brush.  This is more important for non-brush uses where the user is just leveraging the fact that they can lay down vector imagery by drawing a line... or applying a different style to a line.

There's a lot more to develop of course.  Currently, we can have vector brushes that store multiple brushes but the only way to 'activate' one of those brushes is to make sure it is the first frame in the sequence.  We can manually reorder the level's frames but it would be ideal to be able to point to a 'poster frame' and use that as the current brush.  This would be a very nice way to present vector brush sets as they could all be stored in and referenced from one PLI file.  Trail brushes ALMOST get this done but there is no way to stop at any given frame of the multi-frame brush.

andeon:
@RodneyBaker I'm testing some FX nodes. Do you know of any that work better than Blur? Maybe I should combine it with the palette filter.

I'm already thinking about the post-production step. It might look weird, but the real issue is that they won’t be vectors in the end, so they also need to be smoothed. OpenToonz doesn’t seem to handle it well when the line thickness is low, and even with thicker lines, sometimes the result isn’t what we want. Well, until someone improves the vector rendering/conversion, this is what we can work with for now.

RodneyBaker:
You might want to confer with @konero as he is working on some new FX nodes to include antialiasing.  He also has been working on a FX/shader manager that loads and unloads FX as well as that is a difficult process the average user won't explore due to difficulty in placing and editing files.

FX node better than Blur?  Hmmm... Pretty much all of the FX nodes are better than blur.  
I'll give it a look and see if anything stands out.

konero:
I tried but Shun explains here why antialias node is difficult: https://github.com/opentoonz/opentoonz/issues/3863

It's not hard to do, you just call `Trop::antialias` (literally one line of code) but since it will be applied after transforms you're applying antialias to something that's been resampled, which is not good. We would need a multi-stage render framework where we can instruct certain effects to apply at certain stages: column img -> color adjust -> antialias -> transform, etc.

PR #5735, opened 2025-01-21
Increase build timeout to 120 minutes

andeon:
This will increase the build timeout to the maximum allowed by AppVeyor, which is 120 minutes, and should help avoid timeouts for longer builds.

ps. I'm not sure about the permissions; maybe need to adjust the timeout in the AppVeyor Web UI. Log in to AppVeyor and set it there.


RodneyBaker:
If login to appveyor is required to adjust that timeout then @shun-iwasawa will need to make the adjustment.

RodneyBaker:
Will press forward with an eye toward whether this PR allows the increased Appveyor build timeout.

Thanks @andeon

PR #5745, opened 2025-01-26
Revert "Increase build timeout to 120 minutes"

RodneyBaker:
Attempting to see if this will get Appveyor back online.

According to Appveyor documentation:  

> Time limitations
All plans have 60 minutes quota per build job.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -30,7 +30,6 @@ build:
   project: $(APPVEYOR_BUILD_FOLDER)\toonz\$(PLATFORM)\ALL_BUILD.vcxproj
   parallel: true
   verbosity: minimal
-  timeout: 120 # increase to maximum allowed timeout
 after_build:
 - cmd: >-
     


andeon:
So, all that's left is to try to optimize the script, maybe by caching more things. I'm not sure if parallelism would work on the (Open-source) free account.

RodneyBaker:
> So, all that's left is to try to optimize the script, maybe by caching more things. I'm not sure if parallelism would work on the (Open-source) free account.

If you search on the subject of appveyor timeouts you'll see it's a very common problem with a lot of requests for allowing a little more additional time.

andeon:
It's because you created the branch in the OT repository. The right approach is to create a branch in your fork and then send the pull request.

RodneyBaker:
Got it.  Thanks.
I just hit Revert and it took me there.
Note to self.

Regarding running the tests in Parallel with the free version of Appveyor.
Apparently we can indeed run the test 'in parallel' but only one build can occur at at time with the free account so those parallel tests will run one after the other.  Kind of defeats the purpose.

PR #5756, opened 2025-02-01
AutoSave Popup Label

Abel032:
Refactor autosave UI and disable mainwindow before saving.

https://github.com/user-attachments/assets/3a3d0eeb-0a8d-4d49-8aed-4497c2e3668a

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1739,6 +1739,18 @@ bool IoCmd::saveLevel(TXshSimpleLevel *sl) {
 bool IoCmd::saveAll(int flags) {
   // try to save as much as possible
   // if anything is wrong, return false
+
+  QMainWindow *parent = TApp::instance()->getMainWindow();
+  QLabel *Label = new QLabel("Saving...", parent);
+  Label->setStyleSheet(
+      "font-size: 20px;"
+      "background-color: black; color: white; "
+      "font-weight: bold; padding: 5px;");
+  Label->adjustSize();
+  QPoint pos = parent->rect().bottomRight();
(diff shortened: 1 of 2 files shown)


Abel032:
Disable mainWindow maybe not necessary here( Save Command would do that)

Abel032:
Consider to apply this Label for all save command.

RodneyBaker:
Appveyor timed out on this PR...

Pressing on as this is an important update for those that use Autosave.
Much appreciated @Abel032

RodneyBaker:
Unrelated to this PR but related to autosave:

Just a note to caution users that use of Autosave while rendering will (still) often lead to a crash.
Please... please... save before rendering.  Especially if you have Autosave turned on.

Abel032:
Maybe we can add check before autosave, and Save All before every render Action.

RodneyBaker:
Yes, I do think some additional measures for saving prior to rendering would be ideal.

Outside of Autosave...
Jeremy (prior to John taking over) added a Save and Render to Tahoma2D.
It's a real life saver because if a user fails to save for such a computationally expensive task... the odds are against them being happy.  Jeremy made that process optional... because everyone doesn't always want to save their scene before rendering as they might just be testing and not want to change their original scene.  As such he added a button to the Output Settings dialogue.

REF:  https://github.com/tahoma2d/tahoma2d/pull/20/files

I can't remember how we credit contributors that have ghosted themselves on github but @manongjohn probably knows.

konero:
> Maybe we can add check before autosave(if is rendering), and Save All before every render Action.

The check makes sense but saving before render (forced) would be destructive, as when someone makes a small change intended to be scrapped after rendering (rendering is their 'save'), then it's an action unintended by the user. Such a thing should only occur if autosave is enabled.

Abel032:
Just pause autosave if is rendering is good.

PR #5758, opened 2025-02-03
Fix vector thickness to eliminate some gaps

andeon:
After discussions in the last PR https://github.com/opentoonz/opentoonz/pull/5733, I adjusted most of the brushes to minimize gaps when working with Geometric Tool and improved their alignment.

inktwinside

Geometric shapes can be useful when working with the cutter tool is some cases, as they allow cropped lines to take the shape of the brush effect. Even when working with vectors, I recommend doing post-production work using FX nodes to smooth the lines, as the final result will be raster when it is rendering.


RodneyBaker:
These brushes are much improved.  Thanks @andeon 

I'm posting another version of the circle.pli file more for your inspection than for inclusion.
The thing that strikes me about the circle.pli brush is that it currently cannot have a fill (line only).  Here's where things get a odd little odd and experimental regarding my version...

In the attached I gave the circle a fill.
The line is a different style than that fill.
I set tha fill to be fully transparent but black (via the color chip that most users probably don't know is there) 
If looking at those color chips you'll see there are two more chips and they don't appear to change anything.
Those chips are assigned to a second frame of that PLI file.
If a user wan't to swap brushes in that PLI file they can just select the two frames (in the level strip) and hit Reverse.  Save Level and now the second frame is the active brush.

Hope that makes sense.
I started out just trying to see if adding the fill would be a good idea.
I do think that to be the case although what are the odds a user will know where to change that transparency?

[circleA.zip](https://github.com/user-attachments/files/18639859/circleA.zip)

At any rate I've included both circle PLI files in the attached zip.

I'm not sure how experimental we want these brushes to be but I do like the idea of providing a broad spectrum of brushes that demonstrate just how flexible (and extensible) vector brushes can be.  We have that with these brushes you have provided; demo brushes of shapes/symbols, variety in brushes, Text on a Path, etc.

(I still need to look into that raster rendering issue)

andeon:
I believe that the initial objective of creating this circle brush was to simulate writing or something like that. 

It's not necessary to have all these color chips; you could just have the two initial syles transparent, black, and one more. But the second stlye should be a color, and the third can be transparent. In the end, you’ll have two chips, with the second one being transparent.
vector

It will seem like there’s no color at all, since it will take on the background color of the theme. Maybe it’s better to have a different graphic for this, but since it's not commonly used, I’m not sure if it's necessary this change.

About your work around use of two or more frames could be explored to achieve effects somewhat different from the trail do today. However, we would need a special "settings" for this, similar to what is done with the generated vector.

RodneyBaker:
The reason I used the two additional chips is so that the assignment of the second frames colors are not the same as the first frames colors.  In this way the 'frame brushes' are entirely differrent from each other.  If the same chips are used then those colors aren't independent.

I am thinking we might want a have (as a concept) a subdirectory named 'experimental' where experimental brushes are placed therein.  Those brushes can be found at any location... even have their own github repos.  It's just a standard location where users would place experimental files.  I'd further suggest on the development side that creating another subdirectory to save bad brushes (those that break Opentoonz or brush usage) can be placed there and shared so that those issues can be resolved.  The primary files I'm thinking of here would be corrupt PLI files that actually crash Opentoonz if the file is present and simply encountered by Opentoonz.*


*It is my understanding that Tahoma2D might have implemented a method to deal with corrupted PLI files so Opentoonz doesn't crash.   I forget the specifics.

The multiframe brush is definitely food for future development.
My take on that is that we'd have to have a way to assign a pointer to a frame to use that as the active or poster frame.  This would need to be done without changing the level format so would be entirely within the scope of Opentoonz source code adding a pointer as well as UI elements that expose that pointer to the user who can point to a different frame.

The above would be a major undertaking as it should also consider other brushes and how they work as well as the trail brushes.

I'll guess that Opentoonz relying as it does on indexed color via palettes/palette files the palette code would be the most likely and appropriate place to adjust frame pointersas Settings for styles can vary a lot and the code for implementing custom settings is already used and available.

And get this... that frame pointer could be exploited not just to point to different brushes in a level but elsewhere in the code to assign (read:  animate) frames to target any frame of animation.  A style then could store the history of frame history and be easily applied to reuse animation.

I'm definitely running off topic so will try to reign it in.  ;)

This update is looking read for merger but I'd say that's your call.
If not planning any additional changes at this time we can merge.

andeon:
Yeap, I saw! That this comment is mainly to register the procedure. If you're working with two colors, it's possible with three styles. The first transparent one is reserved and not used. The second style, black, is semi-reserved only for colors. If transparency is necessary, it should be created from the third style. 

No additional changes; I think it's done!

RodneyBaker:
Outstanding.

As always, thanks for all you do.
LGTM

PR #5759, opened 2025-02-03
Avoid endless loop for some small fonts

Abel032:
[Small Fonts](https://github.com/user-attachments/files/18637936/smalle.zip)


RodneyBaker:
Looks good to me @Abel032 

Color me curious but my main fascination with this PR is with how you ran across this problem.  :)

LGTM

Abel032:
@RodneyBaker 

Because I use fmall fonts as OT's UI font, so I ran across this problem when I can't use Clapperboard as before.

PR #5760, opened 2025-02-03
Fix selection in File Browser

Abel032:
What this PR changes:

- Make single click to not change current folder except clicked the ..(parent folder)
- Do not select the parent folder when selecting All (Ctrl + A)
- Avoid switching to parent folder after right click after click "..” folder


RodneyBaker:
Proceeding with this change.

shun-iwasawa:
@Abel032 
So sorry for the delayed comment. 
I was wondering why you removed the ability to move folders with a single click, since a certain staff member at our studio find this change (the increase in required clicks) a bit inconvenient..

Do you intend to enable folder selection within the item view panel to allow some command in the future?
Or perhaps to enhance consistency with existing tools like Windows Explorer?

Abel032:
@shun-iwasawa 

Thank you for the feedback.
I made this change mainly to enhance the consistency with Windows Explorer, so that user won't feel strange when first use the file browser in OT.

The second reason is about folder deleting:
In the old version, user can't delete `emptyFolder` from right side, because it's not able to select.
图片

I personally don't have a prefer of whether to use single click to enter folders.

> * Make single click to not change current folder except clicked the ..(parent folder)

I think we can undo this, and add it back if one day we decide to add the ability of renaming, deleting and moving folders from right side view (seems not very necessary for now).
After undo this, we can deleting empty folder only from the tree view.

shun-iwasawa:
Thanks for your reply. I think I understand now. 

Yes, I agree that improving consistency is important. 
If commands like moving or deleting multiple selected folders become necessary and are implemented in the future, similar to Explorer, I think it would be better to use single-click for selecting items rather than moving folders.
However, for now, I feel the convenience of moving items with a single click still has its benefits.

As a compromise, how about reverting the left single-click to move folders, but disabling moving with a right-click and allowing folder deletion and renaming via the context menu? (I'll try making it that way.)

Abel032:
@shun-iwasawa 

I think that would be useful, and it would be a great addition if it's also possible to delete files/folders to the trash or recycle bin.

Issue #5595, opened 2024-09-22
Adapt Clip Studio Paint's Viewer Navigation Keys


konero:
### Is your feature request related to a problem? Please describe

When drawing, it's a pain when using hand, rotate and zoom tools in combination as every now and then either the tool is switched (rather than be temporary) or it's difficult to do combinations.

### Describe the solution you'd like

I propose we create two groups of shortcut keys, one for selecting the tools and another for calling them in a way that bypasses the temporary tool timer, so that we can never accidentally switch tools (this happens too often), just for the three navigating tools.

We also need to allow combinations (for lack of a better word), right now we can't hold SPACE then press SHIFT, if we do that we'll switch to the hand tool bypassing the temporary mode. If you wish to rotate (see below for a very common combination) we might hit SPACE before we hit SHIFT by just slamming our fingers down.

For example, it's very intuitive in Clip Studio Paint:
```
- SPACE for pan
- SPACE + SHIFT for rotate
- ALT + SPACE for zoom
```

If we're already holding SPACE then press either SHIFT or ALT it will flip between the modes intuitively, there's no way we can switch to these tools permanently using these shortcut keys, we'll switch back to our brush (or other tool) as soon as we release.

It might also be useful to take a cue from TVPaint in utilizing right-click:
```
- SPACE for pan
- SPACE + SHIFT + LEFT-CLICK for rotate
- SAPCE + SHIFT + RIGHT-CLICK for zoom
```

This would be easier to remember, but maybe it's tricky to bypass right-click event?

PR #5761, opened 2025-02-04
Adapt Clip Studio Paint's Style of Navigation Shortcut Keys

konero:
_Note: For anyone who tested this, I pushed a fix to handle 'legacy' entries in shortcuts.ini that already assigned reserved keys. Should work properly now by pruning them at startup._

This addresses complaints of accidental tool switching while drawing by adopting Clip Studio Paint's style to navigation.

The new system uses a state machine activated by the Space key:
- Space activates Hand tool and enters navigation mode
- Shift + Space switches to Rotate tool
- Ctrl + Space switches to Zoom tool

You can seamlessly switch between tools by holding Space and toggling Shift or Ctrl. This matches Clip Studio Paint's behaviour exactly, which should make OpenToonz feel more familiar to people who use both together. While the keys are currently hardcoded, in the future we can potentially add user customization, but as it was trouble bypassing Qt's modifier handling customization it is something I'll come back to another time.

Additional Improvements:
- Reversed zoom tool drag direction (now left/right) to match other apps behaviour
- Preventing ToolOptions panel from updating its context when in navigation mode to prevent annoying panel flicker
- Updated several default shortcuts to align with common usage:
- - Brush sizing: `[` and `]` (max), `Shift + [` and `Shift + ]` (min)
- - Pickers: `I` (style), `Shift + I` (RGB)
- - Animation: `F6` (set key)
- - Navigation: `H` (hand), `R` (rotate), `Z` (zoom)

This should reduce the learning curve a bit and make OpenToonz a little easier to integrate into pipelines (Clip Studio Paint is somewhat of a staple). But if any of these changes are problematic we can change them back. I'm guessing the brush size keys were in the middle of the keyboard for some InknPaint workflow.

Closes https://github.com/opentoonz/opentoonz/issues/5595.


RodneyBaker:
Pressing forward with this update.
Thanks @konero for bringing these shortcuts up to standard.

Mondaley:
"While the keys are currently hardcoded, in the future we can potentially add user customization"

So the viewer navigation controls can't be turned off... I spent 30 min looking for it.

This will sound strange, but I use the mouse and the keyboard more than the tablet. All the keys from the left side are dedicated to all the tools I frequently use in OpenToonz. I use the SPACE key to delete things - I dont want to press the actual DELETE key because it is further, thus slower.

I'll have to downgrade the version from 1.7.1 back to 1.7 or something like that.

I hope this will be implemented soon.

shun-iwasawa:
@konero 
I'm terribly sorry for the delayed response.
I'm planning to try to make this feature toggleable on/off in the Preferences.

In our studio, OpenToonz isn't primarily used for the drawing workflow. 
Instead, it's used to arrange drawings on an XSheet and playback or compose animations.
They have been used the Space key as a shortcut for the Preview or Playback commands.
This is because they frequently use those commands, and it maintains consistency with ones of AfterEffects.
Thus, I would like to avoid this PR reserving the Space key and restricting its use for other commands.

I think the implementation would likely be something like this:
- Add an option to Preferences to assign the Space key to Viewer navigation.
- The default value for this option would be OFF.
- When the option is turned ON, OT will unassign the Space key from any existing shortcuts if they are already assigned.

To maintain compatibility with existing shortcut settings, I expect the default value for this option would be OFF.
Please feel free to let me know if you have any suggestions. 
Thank you for your understanding.

RodneyBaker:
@shun-iwasawa @konero 
Posting into this PR reminds me of another potential issue with this PR.
I wasn't able to verify but it seems likely.
Having the option to turn this option off by default may address the problem.

Specifically, we can no longer type multiple words into the Type Tool separated by spaces if typing those words in.  As soon as we hit the Space key we are kicked out of the Type Tool text editing session.  The workaround (for me) has been to type the text elsewhere and then copy/paste it in.

To test:
- Create a toonz raster or vector level
- Type in multiple works or letters and use the Space key
- Note what happens when the Space key is pressed

konero:
@shun-iwasawa No worries. The original issue was due to panning, when drawing and panning a lot quite often the Hand Tool became selected since the key wasn’t held long enough for the temporary timer. If reducing the time further, it creates a problem where it’s difficult to select a tool by key (you have to tap it very fast).

I wonder if there’s a better solution rather than an option though, I also wasn’t a fan of reserving keys, this feature could be walked back a bit.

For example what if it was context aware? If [SPACE] == HAND ? AlwaysTempKey : NormalKeyWithTimer. We can sort of assume how the user expects Space to work in this scenario.

@RodneyBaker Indeed it was reported before, I never had time to look at it recently.

shun-iwasawa:
@konero 
Another solution could be to allow setting separate shortcut keys for tools like Temporary Hand, Temporary Rotate, and Temporary Zoom, which are used solely for temporary navigation.
These shortcut keys should always return to the original tool even if released quickly.

The advantage of this idea is that users can freely assign shortcuts to each tool, and since it integrates into OT's existing shortcut key system, implementation may be relatively straightforward.

On the other hand, the disadvantage of this idea is that it may be difficult to achieve what is possible with this PR: temporarily switching to Rotate or Zoom while holding down the Space key and pressing Shift or Ctrl.

konero:
@shun-iwasawa Ah, that’s a better solution! I’d go with your approach, much more user-friendly.

I think other apps do it this way too, so it wouldn’t be unique behavior to OT. I’m not at a desktop for a bit, but off the top of my head I think some apps like Krita/Blender do something like:

Hand Tool
Pan View
Rotate Tool
Rotate View
Zoom Tool
Zoom View

The Clip Studio state-machine was a bit of embellishment on my part, it wasn’t really necessary to resolve the panning issue, I don’t think it would be much of a drawback to remove it to be honest and something could be thought of in the future to bring it in a cleaner way.

shun-iwasawa:
Thanks for the comment. I'll implement it that way. 👍

PR #5765, opened 2025-02-06
Fix File Browser Model Refreshing

shun-iwasawa:
This PR fixes the following problem in the File Browser, probably related to #5717 :
- Folder Tree and Item View do not update properly when using `New Folder` command on a folder with no subfolders. The problem does not occur on folders that already have subfolders.

As stated in the [documentation](https://qthub.com/static/doc/qt5/qtcore/qabstractitemmodel.html#beginInsertRows), I have changed `beginInsertRows / beginRemoveRows` to be emitted just before the actual insertion/deletion of data in each node's `refreshChildren()`.


Abel032:
If this PR merged we can consider the revert of #5730

RodneyBaker:
Thank you @shun-iwasawa for the File Browser Fix!

PR #5767, opened 2025-02-06
Enhance note level LineEdit

Abel032:
Make Line Edit able to adjust width with typed words for Note Level（and words start from left）

image
image


RodneyBaker:
image

Such a seemingly simple enhancement but very welcome.
It wasn't part of this PR so I'm not sure when the use of Z key with Notelevels was fixed but... works now.  Before we couldn't type 'Opentoonz' because the z key was reserved as a shortcut.  Perhaps it was @konero's keybinging update?

Thanks @Abel032 
LGTM

Issue #1928, opened 2018-04-23
Number+Letter Filename Issues


shun-iwasawa:
After #1833 ( by @manongjohn ) was merged, the restriction on filename for sequential images was relaxed.
Corresponding to such improvement, there still be some points needed to be modified as follows:

- [ ]  The "Rename Cell Field" ( which appears by double-clicking on a Xsheet cell and is for directly inputting cells ) does not accept "number+letter" format yet.

- [ ] Load Level command fails to load whole frames if the level if the last frame has the optional letter in the frame number. `Load Subsequence Level` and `Arrangement in Xsheet` fields in the Load Level Popup should be modified in order to accept "number+letter" format.

- [ ] If the first frame of a level has the optional letter in the frame number, such level cannot be loaded in Flipbook ( by neither `View...` menubar command nor `Load / Append Images` command in the right-click menu of Flipbook ).

- [ ] A column generated as a result of `Convert to Vectors...` command fails to reproduce the cell order the source column if it contains a cell of which frame number has the optional letter.

- [ ] `Remove dot before frame number` option of `Convert...` command does not work properly if the level has the optional letter in the frame number.

- [ ] Some input fields for frame number do not accept "number+letter" format yet, such as `From frame` and `To frame` fields in `Auto Input Cell Number...` popup.

- [ ] Rendering spritesheet with "Individual" option will output sequential images with the names like `filename_number.png` . For now number has no padding and is started from zero, like `character_0.png` , `character_1.png` , `character_2.png` , and so on. Since the four-digits restriction will be abolished, these images will be recognized as one level. However due to the sequence starting from zero, the first frame cannot be loaded with warning popup. I'm not sure if the files starting from zero is necessary for game developers, but if it can be changed to starting-from-1 IMO it would be useful as they can load and preview the whole files in OT.

- [ ] After introducing this feature completely, the preference option Show "ABC" Appendix to the Frame Number in Xsheet Cell can be deprecated.

- [ ] Sequential images with non-four digits padding (such as `A.001.png` , `B_00001.jpg` . etc.) seems to fail to create thumbnail in the File Browser.

RodneyBaker:
Some of these issues (and more) have been addressed.
The identified items could use a review to determine current status.

RodneyBaker:
It appears the first line item on the list works but requires a space separating the Letter from the number.
For instance, typing "A4" will not lead to frame 4 of the level named A to be exposed but typing "A 4" will.

PR #5768, opened 2025-02-07
Add feature to rename image sequences with separator

Abel032:
Consider these file name as part of potential sequence while loading resources.
- `1.imgext` or `2.imgext` and so on (no padding and no name)
- `name1.imgext` (no padding )
- `name01.imgext` (padding = 2)
- `name001.imgext` (padding = 3)
- `name0001.imgext` (padding = 4)

- [x] Ask the user whether to rename the files if a sequence is detected.
This will not be triggered when dragging in a single file.
图片

- [x] Add Preference
图片

### DIFF
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -871,7 +871,7 @@ bool TSystem::renameImageSequence_throw(TFilePath &path) {
     ++padding;
   }
   levelBaseName         = levelBaseName.left(i);
-  QDir parentFolder    = QDir(path.getParentDir().getQString());  // No Slash
+  QDir parentFolder = QDir(path.getParentDir().getQString());  // No Slash
   QString suffix       = QString::fromStdString(path.getType());
 
   QStringList fileNames = parentFolder.entryList(QStringList(levelBaseName + "*." + suffix),
@@ -885,7 +885,6 @@ bool TSystem::renameImageSequence_throw(TFilePath &path) {
 
     }
   
-
   char separator  = '.';
   padding = 4;
   QString FrameID;
@@ -899,6 +898,7 @@ bool TSystem::renameImageSequence_throw(TFilePath &path) {
     if (FrameID.isEmpty()) continue;
       newFileName = levelBaseName + separator +
                           FrameID.rightJustified(padding, '0') + '.' + suffix;
+
 #ifdef __cpp_filesystem
     std::filesystem::path oldPath =
         parentFolder.absoluteFilePath(fileName).toStdString();
@@ -917,8 +917,7 @@ bool TSystem::renameImageSequence_throw(TFilePath &path) {
     }
 #endif
   }
-  path = TFilePath(
-      parentFolder.absoluteFilePath(levelBaseName + separator + '.' + suffix));
+  path = path.withName(levelBaseName.toStdWString()).withFrame();
   return true;
 }
 
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -7,9 +7,6 @@ using namespace std;
 #include <set>
 #include "tfilepath_io.h"
 #include "tconvert.h"
-#ifdef __cpp_filesystem
-#include <filesystem>
-#endif
 
 #ifndef TNZCORE_LIGHT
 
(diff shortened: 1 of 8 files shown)


RodneyBaker:
I like this approach:

> (no padding **_and more than 10 images_**)

That will aid in preventing image sequence false positives.

konero:
Hmm, I don't know if this is good user-experience.

If user drags in image sequence it's going to go into `extras` like `1.png`, but if they drag in another it's going to be `suffix1.png` because the user will be forced to overwrite or rename.

It's also a naming convention used by backgrounds: `1.png/01.png/010.png/bg1.png`, these won't show up in the file browser with this change individually and instead show as a sequence.

I think it's better to enforce good naming convention for sequences: `*_#{1,4}.ext` or `*.#{1,4}.ext`.

Abel032:
It's important to keep good naming convention.

Maybe we should make it as a renamer before importing or only a option in file browser.
This still needs to be considered.

RodneyBaker:
As there is a long history of filenaming problems in Opentoonz extensive testing is needed for this PR and where any shortfalls are discovered they should be detailed.

The PR itself likely doesn't touch many of those filenaming issues but it will probably fix a few as well.  These include the more obvious such as the filenaming with number and extension only (as has been mentioned above).

Not added to this PR but waiting out there in the wings...
Something we should look toward in the future but resist at present is allowing 5 digit sequences as that seems to appear more frequently these days as video editors work with more frames due to use of proxy images (I believe 5 digit may be Resolves default?).  A reason to hold off for now is that Opentoonz is not optimized for working with that amount of frames and currently it is recommended that no scene or shot should exceed 9999 frames (give or take 5 minutes in length which would constitute a very long shot).  A single shot in standard 2D animation by comparison might easily average less than 6 seconds (less than 200 frames)?

RodneyBaker:
@Abel032 
Can you check something for me.
If I have a number of images named 1.png, 2.png, 3.png, 4.png and I select all 4 and attempt to Load I get a crash with the build from this PR.  I don't get a crash in the latest nightly.

Abel032:
https://github.com/user-attachments/assets/35fcb039-c8bb-477a-80d9-4d448636c9fd

Here is my result.

RodneyBaker:
I do apologize for posting this 'bug report' now as its probably a little premature.
My next step will be to test the appveyor build and see if that also has the (potential) problem.

Here's a video that is quick (if you don't count scrolling down through the crash report).
It just demo's selecting the 4 images and an immediate crash.


https://github.com/user-attachments/assets/7fe610ed-53be-4d71-8c13-f2fbdada1cec

RodneyBaker:
Before pressing forward toward merging this update I want to mention a plus for this feature set is that it is driven by Preferences so is optional for users.  Where this doesn't meet user requirements I hope new reports will be opened to identify what isn't working optimally.

There will be some filenaming problems outside of the scope of this change.
Many of these problems are itemized in existing reports while others still need to be identified.

Abel032:
I'll try to use OT's file brower to load.

Abel032:
New commits would resolve crash when loading sequence from coded path, and deal the failure of renaming files.

Abel032:
I've pushed wrong commit, revert it later.

RodneyBaker:
I believe I ran into a problem where renaming a file and supplying a name that is already present replaces the existing file without any warning to the user so they have a chance to cancel.

I mention this here because I was moving toward merging this PR but that seems to be something we should fix before merger even though this appears to be a problem outside the scope of this PR.  It's fine for that fix to be considered in a different PR.

To test that, select an image in OT's browser, Right Click > Rename and supply a name (no need to check the Delete original option as we don't need to delete the original).  Do this same thing again with either the original again or a different image using the same name.  The first renamed image should automatically be replaced by the second renamed image.

Abel032:
> I mention this here because I was moving toward merging this PR but that seems to be something we should fix before merger even though this appears to be a problem outside the scope of this PR. It's fine for that fix to be considered in a different PR.

I can reproduce it with the nightly build, but I think we need to open another PR to solve it.

RodneyBaker:
Thanks @Abel032 Fixing that with a different PR would be best.

As for this PR... unless you have additional commits you want to add this one is ready to move forward.

Abel032:
I think it's done.

RodneyBaker:
Thanks @Abel032 for yet another useful feature enhancement.
LGTM

RodneyBaker:
Not directly related to this PR but something to consider when users do not use the recommended filenaming convention:

Sequential images with non-four digits padding (such as A.001.png , B_00001.jpg . etc.) seems to fail to create thumbnail in the File Browser.  They also appear to fail to create thumbnail/icon in Level Strip.  REF:  #1928

RodneyBaker:
@Abel032 
Is it possible that this PR ignores SVG sequences that are one letter only?

I'm seeing a case where SVG sequences with more than one letter are recognized but with only 1 letter OT only sees the first frame.

Or... perhaps the fix to the related crash:  https://github.com/opentoonz/opentoonz/pull/6084

PR #5773, opened 2025-02-10
Pick Up Line Style Option when tool mode is Area in Color Model viewer

### LINKED ISSUE #5769: Hover over Color Model, Style Picker = Mode always set to Line & Area
### Ask a question related to OpenToonz

Hi, this is all within the Palette workspace, coloring .tlv files such as props, characters, backgrounds, etc. 

I'm having this minor issue when I'm using the fill tool with area mode, and then hover over to the color model, the style picker is also in area mode, when I just want it to be both line & area. I have to press my hotkey for style picker everytime I want to color pick an inked/colored line which is kind of annoying and slows down the process. Is there a work around for this? I've tried all that I understand so far. Thanks for the help.

Abel032:
图片

Resolves #5769

### DIFF
--- a/toonz/sources/toonz/colormodelviewer.cpp
+++ b/toonz/sources/toonz/colormodelviewer.cpp
@@ -50,6 +50,7 @@
 #include <QMouseEvent>
 #include <QUrl>
 #include <QMenu>
+#include <qtoolbutton.h>
 
 #define LINES "Lines"
 #define AREAS "Areas"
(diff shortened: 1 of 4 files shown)


konero:
There's three icons for mode states:

image

Perhaps the `stylepicker_lines` icon is a better fit for this for consistency? Centerline icon is from vector preview.

Abel032:
@konero Thank you, I don't know that before.

RodneyBaker:
Initially, I was a little confused by what was happening in the color model but I chalk that up to having forgotten how to use the color model in the first place.  The underlying confusion for me was resolved when I loaded a TLV level as the color model.  In other words, when I initially was using a PNG image the icon was on and I couldn't toggle it off.  At time I couldn't understand why but of course standard raster levels don't have a distinction between line and area.

Not in the realm of this PR, the Color Model itself is a little confusing and not entirely self evident.  There should be a means to load Color Models/levels directly from the Color Model window.  It's easy to forget that we need to load the initial color model via File > Load Color Model and then can subsequently load different color models directly in the Color Model window.

Will press forward with this update as it works as advertised.

Thank you @Abel032 for implementing this new and very useful option.
LGTM

PR #5775, opened 2025-02-11
Fix Navigation Mode States

konero:
Follow up on: https://github.com/opentoonz/opentoonz/pull/5761

Fixed issues:
- T_Hand was getting called and getting stuck, it has something to do with speed of user pressing key combination
- Key buffering: T_Hand was default even if user buffered (holding a modifier) before pressing space, now we buffer by checking the modifier key directly, so user goes straight to T_Rotate if they're holding Shift before hitting Space
- Better organization of the code, more self-contained


RodneyBaker:
LGTM

PR #5778, opened 2025-02-14
docs: update installation instructions for Qt version 5.15

andeon:
Updated instructions to use Qt 5.15, adjusting references and removing mentions of earlier versions (like 5.9).


RodneyBaker:
Thanks @andeon!

PR #5786, opened 2025-02-19
Optionally hide shift and trace origin positions

konero:
Port from Tahoma2D, more info here: https://github.com/tahoma2d/tahoma2d/pull/1364

Full credit to @manongjohn


RodneyBaker:
I don't use Shift and Trace in Opentoonz enough to exhautively test this but given this code has been tested and working in T2D we'll move this forward.

I'll put it on my todo list to use Shift and Trace more.  ;)

PR #5787, opened 2025-02-19
Fix Gif Export for newer FFmpeg versions

konero:
Not exactly sure what the issue is, but it could be recent FFmpeg versions weren't happy with how we were setting up GIF export filter chain, using `-vf` for what's actually a complex filter graph (since we're doing palette generation that involves splitting/merging streams).

**Changes:**
- Switched from `-vf` to `-filter_complex` since we're doing fancy stuff with the video stream
- Added proper labelling of the filter graph inputs/outputs (`[0:v]` and `[out]`)
- Made sure FFmpeg knows we want a GIF with `-f gif`

Needs someone else to test as well. Could test on older versions, but I'm happy for users to update their FFmpeg...

### DIFF
--- a/toonz/sources/image/ffmpeg/tiio_gif.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_gif.cpp
@@ -81,7 +81,7 @@ TLevelWriterGif::~TLevelWriterGif() {
 
   double framerate = (m_frameRate < 1.0 ? 1.0 : m_frameRate);
 
-  QString filters = "fps=" + QString::number(framerate) +
+  QString filters = "[0:v]fps=" + QString::number(framerate) +
                     ",scale=" + QString::number(outLx) + ":" +
                     QString::number(outLy) + ":flags=lanczos";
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Initial result of testing:

Pros
- GIF renders (which it did not do with FFMPEG 7.0 prior to this PR)
- I am seeing a very fast render (with new FFMPEG with this PR) compared to OT prior to this PR with older FFMPEG.  

Still testing
- False alarm: I thought I saw degradation in quality but further testing doesn't confirm. I mention this here as I want to do more testing to rule out any loss of quality (need more testing)
- I get a crash when setting Output Settings to render 32bit channel widths.  No crash with 16bit width.  I suspect the crash will happen even without this PR (need more testing)

RodneyBaker:
Regarding the crash with 32bit width:

I am not able to replicate that previous crash with this PR and new FFMPEG.
It may be that I was using old FFMPEG with this PR (need to test more)

Still todo:
- Test builds of all vintages of FFMPEG with this PR (to include Tahoma2D's custom ffmpeg).
- Try to measure difference in render speed from old to PR.
- Try to identify a specific release of FFMPEG to recommend to users (this can be more than one release but a release tested and comfirmed to work well would be ideal)

RodneyBaker:
All this is related but a little off topic of this specific PR:

Initial testing with this PR **connected to Tahoma2D's custom ffmpeg** results in empty frames being rendered for 32bit channel width.  

My take is that Tahoma2D's current custom FFMPEG does not support 32bit channel width.
For those that want to test, open T2D, set Output Settings to use 32bit width and render with T2D's custom FFMPEG.  It should behave as if it renders all frames but actually render empty frames.

Recommendation for T2D users that want to use that 32bit width option would be to point T2D to newer FFMPEG (but... anticipate T2D should have same problem rendering GIF with FFMPEG 7.0 and greater).

Disclaimer:  It isn't entirely clear what we are gaining by using the 32bit option with GIF format.  I assume faster rendering.

konero:
16bit and 32bit are not supported using FFmpeg formats, 8bit PNG's are written then encoded by FFmpeg into a container. I don't know the code very well, but writing out QImage_ARGB32 > PNG is limiting it to 8bit? Might need to use EXR as the intermediate.

But that's unrelated to this PR.

RodneyBaker:
Posting intent to merge this PR as it needs to get out to more people to determine any remaining shortfalls especially related to specific versions of ffmpeg.

We will need to keep a close eye perhaps specifically on how this effects users of Opentoonz that use Tahoma2D's custom ffmpeg so @manongjohn can chart similar fixes when that custom build is next updated.  

This may result in some bug reports from users of older ffmpeg who have lost gif functionality.
However, the majority of 'power users' don't use gif format as often as some and this may be even more likely with pending update to webm output which also relies on ffmpeg support.  New users of Opentoonz will most likely download and attempt to use newer releases of ffmpeg so this update will favor movng forward and using those releases.

Where specific releases of ffmpeg are know to work particularly well with Opentoonz we should encourage users to consider that release.  Whereas prior to this PR the recommendation might have been to use ffmpeg 6.0, with this PR that recommendation shifts to preferred use of any release after ffmpeg 7.0 in order to enjoy the benefits of those newer releases.

Most importantly, thank you @konero for submitting this PR and helping us move forward.

PR #5788, opened 2025-02-19
Update Webm Export

konero:
image

**Updates FFmpeg Webm export options:**
- Upgrade codec from VP8 > VP9
- Quality handled by CRF dynamically based on resolution (no more quality slider)
- Added Encoding Speed presets (controls file compression)
- Keyframe interval option

**And import:**
- Can now import VP8/VP9 and AV1
- Changed threads to auto instead of 4 for decode to drastically speed it up for CPU's with more threads

The settings are optimized for animation as much as possible for smallest file size. Please test.

### DIFF
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -429,13 +429,29 @@ void Ffmpeg::getFramesFromMovie(int frame) {
     // frameArgs << "-accurate_seek";
     // frameArgs << "-ss";
     // frameArgs << "0" + QString::number(frameIndex / m_info->m_frameRate);
-    if (m_path.getType() == "webm") {
-      // To load in webm transparency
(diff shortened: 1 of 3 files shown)


konero:
Will come back to this.

RodneyBaker:
My apologies for the delay returning to this PR.
The update works great.

What I was distracted by the issue with Webm output being divided up into chunks/frames via Task Rendering which leads the process to continuously overwrite each set of frames based on the chunk size set in Preferences.  This is an issue with current Webm rendering as well and not just with this PR.  Setting the Chunk size in Preferences to be a value larger than the total number of frames in the scene allows the render to be processed all at the same time but I do see blank frames in some testing as output.  For that, more testing is required.  For those interested in testing be aware that after adjusting chunk size in Preferences the scene must be reloaded in Task Rendering in order for the change in chunk size to update.

For this PR however... I'm posting intent to merge.

RodneyBaker:
Thank you @konero for the update.
LGTM

PR #5790, opened 2025-02-20
Try fix lvcod64.dll crash

konero:
This is a tweak from: https://github.com/tahoma2d/tahoma2d/pull/1711 by @manongjohn to try fix the `lvcod64.dll` crash. 

I haven't been able to reproduce the issue on either OpenToonz or Tahoma2D (I've tried), but I tested the change and AVI is working fine. @RodneyBaker Were you able to reproduce crash on OpenToonz?

Rather than both Tahoma2D and OpenToonz stuck pending for someone to try it, we can take a punt on this and just merge it, and try get some people who experience the crash to install the nightly (getting people to try GitHub artifacts is far more complicated), if no good can just revert it?

The issues page got pretty spammed with this issue, would be good to solve it.

### DIFF
--- a/toonz/sources/image/avi/tiio_avi.cpp
+++ b/toonz/sources/image/avi/tiio_avi.cpp
@@ -970,7 +970,7 @@ TImageP TLevelReaderAvi::load(int frameIndex) {
 int TLevelReaderAvi::readFrameFromStream(void *bufferOut, DWORD &bufferSize,
                                          int frameIndex) const {
   assert(bufferOut && bufferSize > 0);
-  LONG bytesReaded   = 0;
+  LONG bytesReaded = 0;
   LONG samplesRead = 0;
 
   int rc = AVIStreamRead(m_videoStream, frameIndex, 1, bufferOut, bufferSize,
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Thanks @konero 

Just to be clear...
Users with this lvcod64.dll problem (in Opentoonz, Adobe and other programs) weren't necessarily having issues rendering or saving to AVI video.  The problem is/was that the mere presence of the DLL file crashes the program when the file is encountered.

It may be that a specific release of the lvcod64.dll is problematic as downloading and placing the file does not cause Opentoonz to crash for me.  As such, I cannot currently reproduce the problem on my current system.  Before all I needed to do is place the file and Opentoonz would crash at startup.  It's not clear what has changed on my end but that problem for me was so long ago that perhaps it was a different computer.

I agree that for sanities sake we can merge this fix and revert as necessary.
Then we can tell those experiencing the problem to update to the latest nightly.

I would expect reporting of this problem to continue as those who download the official Opentoonz release of v1.7 will still encounter the problem.  The fix however will (potentially) be to update to the latest nightly which contains the attempted fix from this PR.

Thanks again.
Proceeding...

Abel032:
Not related much, but we need to consider to remove srv folder or repair it for nightly build.

RodneyBaker:
@Abel032 I'm about to find out if this fix also resolves the SRV problem.
@manongjohn suggested his version of this fix might do that and this fix is quite similar.

PR #5793, opened 2025-02-22
Fix Visualize Vector As Raster

### LINKED ISSUE #4807: "Visualize Vector As Raster" mode shows low accuracy

mode
"Visualize Vector As Raster" mode shows low precision, affecting vector drawing.
I hope to use this mode to smoothly use the vector level of the software.

Add an "Visualize Vector As Full Raster" mode.

Abel032:
Resolves #4807 

- [x] Make rasterized result same as export and preview
图片
- [x] Add preference of whether to enable Anti Aliasing (in the visualization page)
图片
- [x] Enlarge the rasterized area(not only camera area)
图片

# Issue Description
This [function](https://github.com/opentoonz/opentoonz/blob/9dffdf827c7281025fd605cc85f0536753da1d05/toonz/sources/toonzlib/imagebuilders.cpp#L282C1-L380C1)(_build(int imFlags, void *extData)_) is used to rasterize Vector and return a TRasterImageP.

This function would use to get BBox of the stroke, and make it the size of the Raster Image to return.
But the BBox is not always the actual size that match the canvas, the returned Rasterized Image would be scaled again to fit the actual size.
Here shows how to reproduce it.

**1.** If we set camera DPI > 53.34, the "Visualize Vector as Raster" would scale up, the pixel is bigger.
图片

**2.** If we set camera DPI < 53.34, the "Visualize Vector as Raster" would scale down, the pixel is smaller.
图片

**3.** If we set camera DPI = 53.34, the accurancy of "Visualize Vector as Raster" would recover.
图片

I thought it might related to default *Stage::inch* that affect the way of writing or reading of vector layer.

[stage.cpp](https://github.com/opentoonz/opentoonz/blob/9dffdf827c7281025fd605cc85f0536753da1d05/toonz/sources/toonzlib/stage.cpp#L66C1-L75C39)
``` C++
/*! \var Stage::inch
                For historical reasons camera stand is defined in a coordinate
   system in which
                an inch is equal to 'Stage::inch' unit.
                Pay attention: modify this value condition apparent line
   thickness of
                images .pli.
*/
const double Stage::inch        = 53.33333;
const double Stage::standardDpi = 120;
```

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -239,6 +239,7 @@ public:
   // Visualization  tab
   bool getShow0ThickLines() const { return getBoolValue(show0ThickLines); }
   bool getRegionAntialias() const { return getBoolValue(regionAntialias); }
+  bool getRasterizeAntialias() const { return getBoolValue(rasterizeAntialias); }
 
   // Loading  tab
   int getDefaultImportPolicy() { return getIntValue(importPolicy); }
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -45,6 +45,7 @@ enum PreferencesItemId {
   // Visualization
   show0ThickLines,
   regionAntialias,
+  rasterizeAntialias,
 
   //----------
   // Loading
--- a/toonz/sources/toonz/preferencespopup.cpp
+++ b/toonz/sources/toonz/preferencespopup.cpp
@@ -1205,6 +1205,7 @@ QString PreferencesPopup::getUIString(PreferencesItemId id) {
       // Visualization
       {show0ThickLines, tr("Show Lines with Thickness 0")},
       {regionAntialias, tr("Antialiased Region Boundaries")},
+      {rasterizeAntialias, tr("Rasterize Vector with Anti Aliasing")},
 
       // Loading
       {importPolicy, tr("Default File Import Behavior:")},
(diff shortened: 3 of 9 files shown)


Abel032:
The problem that rasterized picture don't follow camera's dpi(except save all and revert scene) still exist, but I'm working on it.

Abel032:
There is something could be discussed.
A long term goal is to introduce DPI settings for .pli and make it to follow current camera dpi as for default.
But I think it's not very necessary.

Abel032:
Now rasterized picture can follow up new specified camera's dpi.

RodneyBaker:
There are many additional things to look into particularly related to Preview rendering but those are outside the scope of this PR.  The only element of that potentially related to the feature being fixed here would seem to be a shifting of pixels when moving into and out of Preview mode.

As for this PR...
Thank you @Abel032 for this vector to raster visualization fix.
Results are much improved.

LGTM

Abel032:
> a shifting of pixels when moving into and out of Preview mode.

I think it's already resolved by the last commit.
There are no shifting as I observed today.

RodneyBaker:
> I think it's already resolved by the last commit. There are no shifting as I observed today.
> 
> Does it still exist in the newest version?

Very shortly, I'll be doing clean install testing of the nightly build shortly and will report.
This is the first nightly with the tsrv install fix so that should be a nice milestone.

RodneyBaker:
Indeed, I am not seeing pixel shift in the nightly.
Thanks @Abel032

PR #5794, opened 2025-02-24
Color level icons

konero:
image

For usability.


RodneyBaker:
This will help considerably in that although positionally the level types are fairly easy to target via current icons I inevitably grab the wrong ones quite often.  Having them with color should indeed enhance useability.  The larger size should help as well.

Thanks also for the updated Notelevel icon.

RodneyBaker:
Proceeding...

Thank you @konero for the icon update.

PR #5795, opened 2025-02-24
Enhance xsheet selection

Abel032:
What this PR changes:

- [x] Enable to drag column out of the top area
- [x] Hold Ctrl to drag level
- [x] Commit new Cell ID when RenameCellField lost focus, expecting 'Escape' key press
- [x] Enable to type Frame ID directly with number on the keyboard


RodneyBaker:
Relates to the first line item identified in #1928 although more specifically just the number via  number pad entry.

RodneyBaker:
This is a subtle and yet highly useful enhancement of xsheet selection.
I believe I especially like the ability to Control + Alt copy exposures via dragging to other columns.

As these changes will not be immediately obvious to users we may need to spread the word via video demonstration.

All this to say:  LGTM

Thanks @Abel032 for the xsheet enhancement!

shun-iwasawa:
@Abel032 
Sorry for the delayed comment.
Regarding the change where RenameCellField now commits new Cell ID when it loses focus, it's causing confusion among users in our studio.

Specifically, after finishing input of timing in one column, it is common to click the adjacent column to move the RenameCellField. However, because the RenameCellField loses focus while empty, unwanted hold cells are created at the end of the column.
This is especially problematic when the `“Enable to Input Cells without Double Clicking”` preference option is enabled, as clicking anywhere in the XSheet causes hold cells to be inserted repeatedly, causing difficulty in operation.

Shouldn’t inputting a new cell when the RenameCellField loses focus be restricted to only when something has been typed into the field?

Abel032:
@shun-iwasawa 
Thanks for the feedback.

> Shouldn’t inputting a new cell when the RenameCellField loses focus be restricted to only when something has been typed into the field?
 
I think yes, it shouldn't put new cell if nothing typed into the field, Iet me fix this bug.

> `“Enable to Input Cells without Double Clicking”` preference option is enabled

Since the feature of `press keyboard numbers directly to input the cell number` is introduced I think we can do similar things without this preference.

shun-iwasawa:
@Abel032 

> I think yes, it shouldn't put new cell if nothing typed into the field, Iet me fix this bug.

Thanks! I'll review as soon as the PR will be posted.

> >“Enable to Input Cells without Double Clicking” preference option is enabled
> 
> Since the feature of press keyboard numbers directly to input the cell number is introduced I think we can do similar things without this preference.

You're right, these features aren't exactly the same, but they're quite similar.
How about modifying the preferences to create an option like `“Input To Cells”` , where users can choose via a combo box which one to enable (or disable both—for users who want to prevent accidentally editing Xsheet) ?
The default value could be “press keyboard numbers directly to input the cell number” as it seems the most intuitive 👍 .

Abel032:
@shun-iwasawa 
> How about modifying the preferences to create an option like `“Input To Cells”` , where users can choose via a combo box which one to enable (or disable both—for users who want to prevent accidentally editing Xsheet) ? The default value could be “press keyboard numbers directly to input the cell number” as it seems the most intuitive 👍 .

Good idea!
I'll implement the combo box tonight.

PR #5797, opened 2025-02-24
Revert "Fix Focus Lost when renaming new created folder"

Abel032:
Reverts #5730, as #5765 fixed the file browser model refreshing issue


RodneyBaker:
Thank you @Abel032 for the follow up revert of code no longer needed.

PR #5798, opened 2025-02-24
Chore: update aotz PNGs to match the color-level icon

andeon:
This PR depends on the changes in PR #5794, so please merge it after that one.

Now, I have changed the border to blue to indicate that the brush applies only to raster levels. The user should be aware that the MyPaint brush works best on raster levels, due to its painting characteristics, even for brushes that can work on both levels.


RodneyBaker:
Much appreciated!

RodneyBaker:
I apparently merged these out of order which was not my intent.
All should be working as expected however.

PR #5801, opened 2025-02-25
Refactor: Drop support for Qt versions before 5.15

andeon:
This change simplifies the code by removing obsolete compatibility checks, aiming to support Qt 5.15 and above.

Some people building locally may encounter issues, especially if they don't keep up with changes in the code, as the documentation wasn't always up to date. Now, it's important to keep everything in sync, as many reported issues were caused by building with older versions of Qt, especially on Linux.


RodneyBaker:
Thank you for the refactor @andeon

PR #390, opened 2016-06-02
Fix not to crash on 32bit modules because of lack of dependencies

kogaki:
This is fix for #364

PR #5802, opened 2025-02-26
Disable QucickTime support provided by srv folder and use ffmpeg instead

Abel032:
This PR would recover Nightly Build (make OT works without renaming or remove srv folder),
but we need to consider removing or repairing QuickTime format supported by srv folder afterward.

MOV format export is currently supported by ffmpeg.

### DIFF
--- a/toonz/sources/image/tiio.cpp
+++ b/toonz/sources/image/tiio.cpp
@@ -221,7 +221,10 @@ void initImageIo(bool lightVersion) {
       TFileType::declare("apng", TFileType::RASTER_LEVEL);
       Tiio::defineWriterProperties("apng", new Tiio::APngWriterProperties());
     }
-    if (!IsQuickTimeInstalled()) {
+    //if (!IsQuickTimeInstalled()) {
+    // TODO: Remove or Repair QuickTime format supported by srv folder
+    // Abel 2025/02/26
+    if (true) {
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I was looking for where the srv directory is created as a short term fix would be to just have the directory be named 'srv_off' or something of that sort.  Then those that wish to use the old  32bit mov support can just make sure the correct 32bit files are in place which apparently is the underlying problem... 64bit versions of files that need to be 32bit.

Abel032:
I think it's created when packing the setup exe.

Abel032:
[Opentoonz Installer Windows](https://github.com/opentoonz/opentoonz_installer_windows)

RodneyBaker:
That's where I thought it might be... specifically:

https://github.com/search?q=repo%3Aopentoonz%2Fopentoonz_installer_windows+program%5Csrv&type=code

What I guess I don't completely follow is how the current build system incorporates the windows installer from that repo into the build.

RodneyBaker:
Here's some of the reporting history on 'program/srv' keyword (21 or so reports):

https://github.com/search?q=repo%3Aopentoonz%2Fopentoonz+program%5Csrv&type=issues

RodneyBaker:
Here's the fix that was applied way back when this SRV problem first appeared (circa 2016 or so):

https://github.com/opentoonz/opentoonz/pull/390/files

PR #5808, opened 2025-03-01
Add SVN config combo preference

Abel032:
Provide a way to config SVN in Preference.
图片


RodneyBaker:
Proceding with this addition.
Thank you @Abel032

PR #5809, opened 2025-03-01
Adjust Notice Label Size and Color after Text Changed

Abel032:
1. Adapt to different Fonts.
2. Change font color to red if save failed.


RodneyBaker:
Thanks @Abel032 
Much appreciated.

PR #5813, opened 2025-03-02
Optimize Background Process Startup

Abel032:
If the background exe get error after started, we failed to start it.
Make sure process run without crash.

This PR would alsoo make OpenToonz to turn to use ffmpeg if t32bitsrv.exe is missing .dll.

### DIFF
--- a/toonz/sources/common/tipc/tipc.cpp
+++ b/toonz/sources/common/tipc/tipc.cpp
@@ -263,18 +263,22 @@ QString tipc::applicationSpecificServerName(QString srvName) {
 bool tipc::startBackgroundProcess(QString cmdlineProgram,
                                   QStringList cmdlineArguments) {
 #ifdef _WIN32
+  SetErrorMode(SEM_FAILCRITICALERRORS | SEM_NOGPFAULTERRORBOX);
+
   QProcess *proc = new QProcess;
 
   proc->start(cmdlineProgram, cmdlineArguments);
-  if (proc->state() == QProcess::NotRunning) {
-    delete proc;
-    return false;
-  }
-
+  
   QObject::connect(proc, SIGNAL(finished(int, QProcess::ExitStatus)), proc,
                    SLOT(deleteLater()));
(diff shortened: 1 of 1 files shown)


Abel032:
Maybe it would be better to windows api instead of QProcess.

Abel032:
I think if John's PR fixed the crash we can close this PR and if not we can merge this PR and revert it in the future.

Abel032:
Or use Windows' original API instead of QProcess, this would enable to detect crash caused by missing dll or other reason without a fixed time to wait.

Abel032:
Now the error message of t32bitserv.exe would pop up but OT would start as normal.

Abel032:
Now the error message of t32bitsrv.exe might not show because process would be deleted at once after the crash.

RodneyBaker:
Current testing of nightly release still has the tsrv startup crash so this PR is back on the table even if only for interim measures.

Abel032:
I believe this would be beneficial to ensure we can still use OT even if the srv folder doesn't work.
And with the new code added, there is no fixed delay.
Simply check if t32bitsrv.exe started normally.

RodneyBaker:
The fix for the tsrv bug appears to work.
Thanks to @manongjohn and @shun-iwasawa for closing this long standing problem down.

@Abel032 It's not yet clear how we should proceed with this PR.
It may be that we will still merge this as a hedge against similar problems occurring in the future but we do seem to have a clearer path now to such conflict resolution in the form of the setup2.bat file in the Windows Installer repo.

Let's stare at this for awhile and see if perhaps we can expand this optimization to encompass additional background related startup errors.

Abel032:
Thanks @RodneyBaker 
It's basically like the difference between hard link and soft link.

Sometimes we can't open a scene without the right plugin installed, but OT would not tell us what files are missing,and refuse to work.
This is similar to what we faced here.

RodneyBaker:
Closing as resolved by #5846

Abel032:
The build-in QuickTime would much faster than ffmpeg ver when loading from outside.

PR #5816, opened 2025-03-05
[VectorSelectionTool] Fix Rasterized Vector Image update failure

Abel032:
# Issue Description
If View->Visualize Vector as Raster is on, and we adjust line's thickness through selection tool, the rasterized Image won't update.
图片

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -476,8 +476,9 @@ return true if the method execution can have changed the current tool
   TStageObjectId getObjectId()
       const;  //!< Returns a pointer to the actual stage object.
 
-  void notifyImageChanged();  //!< Notifies changes on the actual image; used to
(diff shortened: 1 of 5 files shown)


Abel032:
> I'm not sure if this is a shortfall of the github action build and can't check with the appveyor build as that timed out.

What I downloaded works, don't know why.

> Interestingly, I believe this Fix resolves an issue with refreshing the rasterized vectors when different sampling (Resample
> Balance) is set in Preview Settings. To test, change from Standard to Nearest neighbour. That seems to be working great.

Maybe lack of refreshing when adjusting thickness, good thing anyway haha.

Abel032:
> I'm not seeing an automatic refresh of the rasterized vectors unless Preview is on.

Would vector line's thickness change if you change it's thickness from tool opinion?

Abel032:
I can't reproduce the same result (Action Build)... strange

RodneyBaker:
Proceeding with merger.

PR #5818, opened 2025-03-06
Default Palettes

Abel032:
图片

Use default palettes when creating new Toonz Raster Level and Toonz Vector Level.
These default palettes can be modified in Studio Palette Panel.

If there are no default palettes in Project Palettes, copy one from Global.
We use default palettes from Project Palettes first, and Global Palettes as second.

- [x] Add Default Clean Up Palette in Studio Palette panel
- [x] Add default Raster Drawing Palette to Global Palettes
Notice: Raster Level always share the same Palette
- [x] Add refresh button in Studio Palette Panel
图片
- [x] Add **Set As...** in Content Menu of Studio Palette Panel
图片
图片


Abel032:
Try Rebuiding.

RodneyBaker:
This is an outstanding update to palettes in so many ways.
 I will fail to capture them all here.

Firstly, thank you for continuing to make these adjustments in a way that minimally impacts users of past releases of Opentoonz.

I'm still finding more ways to appreciate this approach but the ease with which we can set a new default palette for any level type is... how shall I say this without exaggerating... inspired.

A trivial thought occurs to me that we probably should have a Minimal palette that just has the default 0 and 1 styles in order to easily revert back to the old/current default after adjusting a default to another palette.  Not so much recommendation as a random thought.   It's easy enough to create that two style palette.  If there is a 'revert' to previous default I am unaware of it.

RodneyBaker:
I see no reason to delay this addition so...
Posting intent to merge.

Abel032:
@RodneyBaker 
Thank you, I think it's completed.

RodneyBaker:
Thank you @Abel032 for this important addition.
LGTM

PR #5819, opened 2025-03-07
Update OpenCV version and remove local build commands

Abel032:
Updated the OpenCV installation to version 4.10.0.

### DIFF
--- a/.github/workflows/workflow_windows.yml
+++ b/.github/workflows/workflow_windows.yml
@@ -111,7 +111,7 @@ jobs:
           copy /Y ..\..\thirdparty\libmypaint\dist\64\libmypaint-1-4-0.dll Opentoonz
 
           IF EXIST C:\tools\opencv (
-             copy /Y "C:\tools\opencv\build\x64\vc14\bin\opencv_world*.dll" Opentoonz
+             copy /Y "C:\tools\opencv\build\x64\vc14\bin\opencv_world4100.dll" Opentoonz
           ) ELSE (
              copy /Y "C:\opencv\451\build\x64\vc14\bin\opencv_world451.dll" Opentoonz
           )
--- a/.github/workflows/workflow_windows.yml
+++ b/.github/workflows/workflow_windows.yml
@@ -32,7 +32,7 @@ jobs:
         run: |
           @echo on
           REM  install opencv
-          choco install opencv --version=4.5.1
+          choco install opencv --version=4.10.0
           REM install boost
           choco install boost-msvc-14.2 --version=1.74.0
         shell: cmd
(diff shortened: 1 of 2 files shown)


andeon:
Now, it's copying two DLLs:
```
C:\Tools\opencv\build\x64\vc16\bin\opencv_world4100.dll
C:\Tools\opencv\build\x64\vc16\bin\opencv_world4100d.dll
```
Make sure to adjust the 'else' line of code so the DLL is copied correctly, since the Windows artifact didn't include it, causing the error.
 
https://github.com/opentoonz/opentoonz/blob/7bd0730900dc7b7e85c59f9a903c8d2b44cb8aa1/.github/workflows/workflow_windows.yml#L116

.github/workflows/workflow_windows.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         run: |
           @echo on
           REM  install opencv
-          choco install opencv --version=4.5.1
+          choco install opencv --version=4.10.0
           REM install boost
           choco install boost-msvc-14.2 --version=1.74.0
         shell: cmd
```
Abel032, line 38: Are these PATHS used for Jenkins?

Abel032:
Are these PATHS used for Jenkins？
```
REM Setup for local builds
          set MSVCVERSION="Visual Studio 16 2019"
          set BOOST_ROOT=C:\boost\boost_1_74_0
          set OPENCV_DIR=C:\opencv\451\build
          set QT_PATH=C:\Qt\5.15.2_wintab\msvc2019_64

          REM These are effective when running from Actions
          IF EXIST C:\local\boost_1_74_0 set BOOST_ROOT=C:\local\boost_1_74_0
          IF EXIST C:\tools\opencv set OPENCV_DIR=C:\tools\opencv\build

          set WITH_WINTAB=Y
          IF EXIST D:\a\opentoonz\opentoonz\thirdparty\qt\5.15.2_wintab\msvc2019_64 (
              set QT_PATH=D:\a\opentoonz\opentoonz\thirdparty\qt\5.15.2_wintab\msvc2019_64       
``` 
https://github.com/Abel032/opentoonz/blob/b7d0a4b5c585947b505e9dcaced9666a55e4ae1e/.github/workflows/workflow_windows.yml#L66-L79

andeon:
I'm talking about the else block in workflow_windows.yml at line 116. It needs to be adjusted to copy opencv_world4100.dll as well.

Abel032:
If the else block is only for local machine build, I think we can remove it and L66-L79.
And create another batch to build locally on windows.

andeon:
I agree, since we're exclusively using GitHub Actions, those lines can be removed to simplify the workflow.

andeon:
@Abel032, sorry for not noticing this earlier. I realized the problem with the DLL not being copied to the Windows build artifact is that the OpenCV version changed from `vc14` to `vc16`, so the path needs to be updated to:

`copy /Y "C:\tools\opencv\build\x64\vc16\bin\opencv_world4100.dll" Opentoonz`.

Here is the [GitHub Actions log](https://github.com/opentoonz/opentoonz/actions/runs/13717351566/job/38364944011?pr=5819):
> D:\a\opentoonz\opentoonz\toonz\build>IF EXIST C:\tools\opencv (copy /Y "C:\tools\opencv\build\x64\vc14\bin\opencv_world4100.dll" Opentoonz ) 
> The system cannot find the path specified.

Abel032:
Thanks@andeon 
That was thoughtful.

RodneyBaker:
Posting intent to merge.
If anyone has concerns about the removal of the local build commands now would be a good time to discuss that.

RodneyBaker:
Thank you for the update @Abel032 
LGTM

PR #5821, opened 2025-03-08
Fixed thickness LineEdit of vector selection tool

Abel032:
image


RodneyBaker:
I may be misunderstanding this change a little.

Is this option basically the same thing as if we first set our selected vector strokes to zero and then enter our desired thickness?  The effect of those steps being to zero out the thickness of all selected srokes and then increase them to the set thickness value versus scale all vectors in the selection based on their current values and the thickness value being set.

beeheemooth:
It's not very convenient to press Ctrl+Z twice to get the previous strokes.

Abel032:
I believe there is a way to reset troke's thickness directly.
Need to refactor.

Abel032:
Refactored.

RodneyBaker:
I have some questions about this PR but I haven't been able to properly formulate them yet.

The core of this PR works as suggested but one of my questions relates to how using the Selection gizmo just below the object allows the stroke thickness to reduce the thickness to zero and increase the thickness back up again.  If we compare this to manual entry of a value to set the thickness in the toolbar setting the value to zero will 'zero out' the thickness which is a way to get rid of the scaling of different strokes selected.  If not sure what I'm talking about here:

- Draw some strokes with variable/pressure driven thickness (or draw several strokes each with a different thickness
- Enter a number other than zero in the toolbar and note the strokes scale from their current thickness and do not change to a common thickness.
Note:  It is this part I believe this PR means to address by allowing all strokes to be set to the value set.

It is slightly out of scope for the Thickness gizmo to also inherit the same change when the 'Fixed' option is checked but it is odd that that is ignored when the Thickness gizmo is used.  

I think what I'd like to see is that when 'Fixed' is toggled on scrolling the Thickness gizmo down to zero will zero out all selected stroke thickness just like manual entry will.

Moving on to another question...
'Fixed' seems to me to be a little vague of a term for placement on the toolbar.
I'm not sure what would be better but... there's a question in here somewhere.

Moving on yet again.
I find myself wondering if perhaps we can skip adding the toolbar 'Fixed' option and just have the Thickness gizmo zero out stroke width when it is scrolled to zero.

The problem of course then is that the current process of entering a value into the toolbar will continue to scale stroke thickness rather than set all strokes to the value entered.  So... we do appear to need the 'Fixed' option to effect that change.

I guess I'm trying to see the process of adjusting stroke thickness (in both ways) and achieving desired results but I'm not quite able to see the broader scope of the process.

As this PR does what it's intended I'm certainly willing to proceed with merging but that's with a strong sense that we'll need to revisit other aspects of the change.

All this to suggest:   Feedback and discussion is appreciated.  :)

Abel032:
@RodneyBaker 

Great thoughts!
Currently, we have three ways to scale stroke thickness:
- Dragging the thickness icon
- Dragging the gizmo just under the bounding box
- Inputting a value in the thickness textbox

Add an "Fixed" checkbox would make things more complex.
So I think there would be two potential options for this PR:

- just no "Fixed", user should drag on the Thickness icon or use the gizmo to scale the strokes' thickness (I think user would use the gizmo to scale strokes instead of the icon)
- no "Fixed", but add another scroll bar to scale the strokes' thickness(this seems to make the option bar too narrow)

I think it's better to simply remove the "Fixed" checkbox, and when user want to set strokes' thickness to a fixed value, user input value, otherwise user drag the gizmo to scale, which is clear enough.

RodneyBaker:
Thanks for condensing my thoughts into better form @Abel032 
And... thanks for the reminder of scrubbing the icon as I had not tested that.
I note that when using that and scrubbing to zero and then back up again all the lines do change to the same thickness.  Interesting.  The downside of that is how Undo ignores that so if we didn't mean to adjust a selection of lines to be the same width... yikes... we've just lost our variable line thickness.

You've helped me understand better what it is I need to be testing.

Abel032:
Now we have three way to adjust the strokes' thickness:

1. dragging the gizmo to scale 
2. dragging the to thickness icon to scale (scale to zero to set a fixed value)
3. type in the value to set a fixed value

RodneyBaker:
That seems like a winner to me.
The real selling point for me is that if a result doesn't meet the intent of the user they can Undo and try again with a different method.

Posting intent to merge although may not merge soon as we are still waiting on the nightlies to start building again.  In the meantime I'll continue to test and make sure we haven't missed anything.

All in all I find these options work well.

RodneyBaker:
Nightlies still aren't building but I believe this PR is otherwise ready to move forward.

RodneyBaker:
Proceeding with merger of this PR.
Thank you @Abel032 for the fix/adjustment and your patience.
We may need to adjust Opentoonz documentation so that users fully understand what they can do here but for the most part it should be self evident.

PR #5822, opened 2025-03-08
Update xsheet tool bar

Abel032:
Add these buttons as default
452432d9de4d142ca01295f43c5e3b13

### DIFF
--- a/stuff/profiles/layouts/settings/xsheettoolbar.xml
+++ b/stuff/profiles/layouts/settings/xsheettoolbar.xml
@@ -20,6 +20,7 @@
     <command>MI_Reframe1</command>
     <command>MI_Reframe2</command>
     <command>MI_Reframe3</command>
+    <command>MI_Reframe4</command>
     <separator/>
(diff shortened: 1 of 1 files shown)


beeheemooth:
Heh, icon designer ran out of imagination :)

Abel032:
Maybe we don't need to add the last one..

JParks123:
@Abel032  Can you add 4's too?

konero:
Hmm, don't need Shift Origin (very rare action), and I wouldn't include On 4's either as it's very uncommon. The -1 and +1 can handle rare exposure cases, or the user can add buttons for their own unique workflow. It's becoming a bit duplicative and we shouldn't bloat it with uncommon actions, because when X-Sheet is in vertical mode and docked to the side they won't be visible (it's why Shift and Trace was added to the Command Bar, not everything could fit for good default UX).

Abel032:
Close this as user can modify their xsheet tool bar or Command Bar for quick switch.

PR #5824, opened 2025-03-09
Simplify and organize Windows build workflow

andeon:
Simplified the setup of environment variables for paths (QT_PATH, BOOST_ROOT, OpenCV_DIR, VCINSTALLDIR) to reduce redundancy.

Consolidated and cleaned up redundant steps related to file copying and environment configuration.

Set VCINSTALLDIR correctly, allowing windeployqt.exe to locate the required Visual Studio installation directory. The warning 'Cannot find Visual Studio installation directory, VCINSTALLDIR is not set' has been fixed, which automatically generates the vc_redist.x64.exe file. Since the portable version doesn't require it, vc_redist.x64.exe is manually excluded.

The warning appeared as follows: `Warning: Cannot find Visual Studio redist directory, "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC"\redist.`

### DIFF
--- a/.github/workflows/workflow_windows.yml
+++ b/.github/workflows/workflow_windows.yml
@@ -9,6 +9,13 @@ on:
 jobs:
   Windows:
     runs-on: windows-2019
+    env:
+      QT_PATH: ${{ github.workspace }}\thirdparty\qt\5.15.2_wintab\msvc2019_64  # Path to custom Qt.
+      VCINSTALLDIR: 'C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC'  # Path to MSVC.
+      MSVCVERSION: "Visual Studio 16 2019"  # Visual Studio version.
+      BOOST_ROOT: C:\local\boost_1_74_0  # Boost library root.
+      OpenCV_DIR: C:\tools\opencv\build  # Path to OpenCV.
+
     steps:
       - uses: actions/checkout@v4
         with:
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@andeon This is outside the scope of this PR which is more to address automated builds but I've been curious so will ask here...

Can you think of specific reasons we don't supply a Visual Studio .sln (prebuilt solution) file to assist in building Opentoonz with MSVC?  Perhaps that file would be difficult to keep up to date given the nature of constant change in the code and build support files?   

I feel I should know the answer to this but have never seen it discussed.  My motivation is to see the process of custom builds simplified even more dramatically in order to encourage 'hackers' of Opentoonz code to not only resolve their own problems locally but also to faciliate forwarding of the best and most optimized of those local solutions for inclusion in Opentoonz.

andeon:
Some developers may have different requirements for building the project. For example, some might use the latest version of Visual Studio, while others might use an older version or prefer specific compiler flags or configurations. Managing these differences with a .sln file can be complicated, as it also makes handling dependency management less efficient.

However, I believe there are additional reasons, the .sln file would only be useful in restricted scenarios. I'm not sure if it would attract potential developers, and to maintain it, there would need to be someone actively using it in this way, due to the need to keep it up to date given the nature of constant change in the code and build, as you yourself concluded.

RodneyBaker:
Thanks @andeon for your response.
That definitely aligns with my thoughts.

One reason I may be asking this is that when building a fork locally it seems like a trivial thing to include the solution as well and am wondering why I don't see that as often as I would expect.

Perhaps a little further I am looking forward to the day when we actually will have some  coursework on building and modifying Opentoonz and we will want to make the onramp for the course as friction free as possible.

One of Opentoonz major strengths above and beyond almost every graphics program in its cateogory is that it is fully open source, explorable and customizable.

As is often the case I have to learn the hard way why we don't do or do not do things through trial and error but I would like to smooth out that path for those that follow after.

I do understand and appreciate that some friction... obstacles... helps us learn.
I'm trying to assess where the sweet spots are within the scope of supplying everything and building everything from scratch.

And no, Opentoonz is not the ideal code to launch into when learning how to code but there are a lot of different on ramps that can be leveraged.  Building the entire program from source is just one.

RodneyBaker:
Thanks @andeon for the updated WIndows build YML file.

Issue #5250, opened 2024-01-04
missing fclose


maddin200:
### Description

\toonz\sources\stdfx\igs_line_blur.cpp line 1251
Code:
 
  if (fprintf(fp, "# count %d\n", this->get_i32_count()) < 0) {
    pri_funct_err_bttvr("Error : fprintf(count) returns minus."); // <-- missing fclose
    return NG;
  }

  for (clp_pp_node = (pixel_point_node *)this->get_clp_first(), ii = 0L;
       (NULL != clp_pp_node) && (ii < this->get_i32_count());
       clp_pp_node = (pixel_point_node *)clp_pp_node->get_clp_next(), ++ii) {
    if (fprintf(fp, "%d %d\n", clp_pp_node->get_i32_xp(),
                clp_pp_node->get_i32_yp()) < 0) {
      pri_funct_err_bttvr("Error : fprintf(count xp yp) returns minus.");  // <-- missing fclose
      return NG;
    }
  }
  

### Steps to Reproduce

1. 
2. 

### OpenToonz Version

1.7

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

None

PR #5830, opened 2025-03-12
Fix: Add missing fclose calls for proper file handling

andeon:
- Added missing fclose(fp) in caller functions to prevent resource leaks, including in error cases.
This PR fixes: #5250

Ensures 'fclose' is called after all file operations, or in case of an error.


RodneyBaker:
Thanks for closing another loophole @andeon 
LGTM

Issue #5496, opened 2024-06-14
init variable


maddin200:
### Description

File:
toonz\sources\toonzfarm\tfarmclient\submitpage.cpp line 294
Code:
void SubmitPage::Data::browseCasmFiles() {
  static FileBrowserPopup *puCasm   = 0L;
  static FileBrowserPopup *puScript = 0L;
  FileBrowserPopup *popup; // <-- better init

### Steps to Reproduce

1. 
2. 

### OpenToonz Version

1.7

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

None

Issue #5494, opened 2024-06-14
init variables


maddin200:
### Description

File:
\toonz\sources\colorfx\flowlinestrokestyle.cpp line 86
Code:
double firstW, lastW; // <-- better init

### Steps to Reproduce

1. 
2. 

### OpenToonz Version

1.7

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

None

PR #5832, opened 2025-03-13
fix: initialize `popup`, `firstW`, and `lastW`

andeon:
- Ensure proper initialization of the `popup` pointer.
This PR fixes: #5496 
- Initialize `firstW` and `lastW` to avoid undefined behavior in `insertPoint`.
This PR fixes:- #5494

I noticed that sometimes files are encoded in Shift JIS. In VSCode, it's a good idea to enable 'Auto Guess Encoding' and, under 'Candidate Guess Encodings,' add Shift JIS to the list, so after commit it converts to UTF-8 correctly.

### DIFF
--- a/toonz/sources/colorfx/flowlinestrokestyle.cpp
+++ b/toonz/sources/colorfx/flowlinestrokestyle.cpp
@@ -50,7 +50,7 @@ bool isCuspPoint(const TPointD &p0, const TPointD &p1, const TPointD &p2) {
   p2_p1 = p2_p1 * (1.0 / n2);
 
   return (p0_p1 * p2_p1 > 0) ||
-         (fabs(cross(p0_p1, p2_p1)) > 0.09);  // more than 5‹ is yes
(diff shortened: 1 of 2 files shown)


andeon:
@RodneyBaker I checked, and it's currently set to C++11. Since Qt6 requires a minimum of C++17, it would be a good idea to switch to C++17.  I don't think there will be much code change, maybe just some tweaks to align with modern standards. I'm not sure where we should discuss this, should we open an issue?

RodneyBaker:
Yes, it would be good to open an issue where any specifics can be discussed.
If we can move entirely to C++ 17 that'd be a good thing.
As with most old code there may be some exceptions that rely on deprecated methods or functions which keep us referencing C++ 11 but if so we should identify them.

My memory suggests that one of the reasons we haven't updated some third party libraries relates to our reliance on C++ 11 (or other older sources of support).

RodneyBaker:
Thanks @andeon

Issue #5495, opened 2024-06-14
unreachable code


maddin200:
### Description

File:
toonz\sources\tnzbase\tscanner\tscannerepson.cpp line 257
Code:
    break;
    break; // <-- unreachable

line 871
 bool status1        = expectACK();
  return status1;

  return 1; // <-- unreachable

### Steps to Reproduce

1. 
2. 

### OpenToonz Version

1.7

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

None

Issue #4386, opened 2022-03-31
Unreachable code


maddin200:
/toonz/sources/toonzlib/Naa2TlvConverter.cpp line 552

Code:
  if (bs.count() == 2)
      region.type = RegionInfo::ThinInk;
    else if (bs.count() == 3) {
      continue; 
      if (bs[2] * 5 < bs[1]) region.type = RegionInfo::ThinInk; // <- unreachable
    } else {
      continue;
      int k = 1;  // <- unreachable
      int s = 0;
      while (k < bs.count() && s * 100 < region.pixelCount * 90) s += bs[k++];
      if (region.pixelCount > 100 && k <= 3)
        region.type = RegionInfo::ThinInk;  // era Ink per qualche ragione
    }

PR #5833, opened 2025-03-13
fix: unreachable code in tscannerepson.cpp and Naa2TlvConverter.cpp

andeon:
-  Removed second break at L257 and unreachable return 1; at L871 in tscannerepson.cpp.
This PR fixes: #5495

- Fixed unreachable code due to `continue` statement in `Naa2TlvConverter.cpp`.
This PR fixes: #4386

In Naa2TlvConverter.cpp, the `continue` statements may have been left unintentionally. Since the code is used to convert raster images to (TLV) format, it’s crucial to test the entire conversion pipeline to ensure that changes do not negatively affect the process **before merging**. If the `continue` statements were intentional, the code should be restructured to better clarify the intent.

### DIFF
--- a/toonz/sources/tnzbase/tscanner/tscannerepson.cpp
+++ b/toonz/sources/tnzbase/tscanner/tscannerepson.cpp
@@ -254,7 +254,6 @@ throw TException("Scan area too large, select a correct paper size");
       rasBuffer = TRasterGR8P(dimlx, dimly);
       buffer    = rasBuffer->getRawData();
       break;
-      break;
 #else
       ras       = TRasterGR8P(dimlx, dimly);
       bytes     = tceil(dimlx / 8) * dimly;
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Just to clarify...
TLV is toonz raster but you mention vector.
PLI would be vector.

andeon:
> Just to clarify... TLV is toonz raster but you mention vector. PLI would be vector.

You're right, I got confused here while writing and reading something else.

konero:
> > Just to clarify... TLV is toonz raster but you mention vector. PLI would be vector.
> 
> You're right, I got confused here while writing and reading something else. In this context, NAA clearly refer to non-anti-aliased raster images. Given that, do you know the steps to test this workflow?

**Naa2TlvConverter.cpp** should be called from convert (file > convert) or from browser, right-click an image > convert, set file format to TLV, put settings to Non-AA source.

andeon:
Thanks @konero, that's a workflow I'm not familiar with. I'm not sure how to evaluate this situation. However, the correct approach would be to simulate the conditions where the logic was not reached due to the use of the 'continue' statement, to check if it’s still working as expected. I'm not sure, maybe using Dwako's TGA samples to test.

andeon:
@RodneyBaker I believe we can move forward with merging this PR and closing the issues. If anyone notices anything strange, we can just restructure the logic in Naa2TlvConverter.cpp since it wasn't working previously anyway.

Issue #5834, opened 2025-03-13
Suggestion: Upgrade C++ Standard to C++17


andeon:
Currently, OpenToonz uses C++11 for compilation. I suggest upgrading to C++17, as it would provide access to more modern C++ features.

I also propose updating the minimum required CMake version from 2.8.11 to 3.1. This change would align with current practices, as most developers are likely using newer versions of CMake (greater than 3.1).

Please let me know if there are any concerns. My initial tests seem fully functional, but to ensure stability across different environments, I believe it’s better to submit a pull request as a WIP (Work in Progress) to test across all supported platforms, including macOS and Linux.

andeon:
Closing this now that the change has been completed.

PR #5835, opened 2025-03-13
Chore: update CMakeLists.txt to use C++17

andeon:
- Updated the CMakeLists.txt to specify the use of C++17 for the project.

~~- Adjusted the CMake minimum version requirement to 3.10 to ensure compatibility with C++17 and modern CMake features.~~

Please review these changes and let us know if there are any concerns regarding compatibility. Testing on all supported platforms is essential to ensure stability and prevent issues later.

Ref: #5834


andeon:
Adapting the CMake configuration from 2.8.11 to 3.10 will require significant effort. It might be better to approach this incrementally, as some commands or modules are deprecated, requiring a complete rewrite. Additionally, ensuring support for platforms like Haiku, BSD, and macOS will require additional time for testing.

RodneyBaker:
Support of .MOV via the Quicktime SDK can still be used (as an alternate to .MOV suport through FFMPEG) so support for it should not be removed.

andeon:
@RodneyBaker thanks for the clarification! I think this PR is done unless someone finds an issue.

RodneyBaker:
Will plan to merge this sometime this coming week.
Feedback from those that build Opentoonz locally will be appreciated.

Changes such as this help us move forward on many fronts even as we keep an eye out for breaking changes.

RodneyBaker:
The following relates primarily to CMake and not the updated requirement to C++17.
I include it here just for a perspective outside of our current scope but one relevant to many others as (mostly) captured in the vfx platform specifications; specifically where someone was asking if they should move the minimum version of CMake to 4.0 (or 3.22 as suggested by another).

The important part is later in the discussion where Larry Girtz suggests CMake has made some changes they shouldn't have made and suggests a way to get around that.

This doesn't yet apply to Opentoonz as we aren't even using CMake version 3 yet.
It should however give some context in how others have been moving forward.

Link to vfx-platform discussion on adjusting the minimum requirement for CMake:  https://groups.google.com/g/vfx-platform-discuss/c/vQ8CYMTn5co/m/NiEKSCNvBwAJ?utm_medium=email&utm_source=footer

RodneyBaker:
I'm back on track to merge this if anyone is still testing...

andeon:
Thanks for finding this about CMake, @RodneyBaker. Let's keep an eye on it when the upgrade is proposed or if it becomes inevitable for some reason. I think you can go ahead and merge it.

RodneyBaker:
Thanks @andeon for helping to keep Opentoonz moving forward.

PR #5838, opened 2025-03-15
Unify Geometric Tool thickness and adjust VectorBrush and VectorGeometric size

### LINKED ISSUE #5498: Geometry Tool Incorrect Thickness on ToonzRaster Levels
### Description

In Opentoonz v1.5, an arc made by the Geometry Tool (thickness = 1, pencil mode on) on a ToonzRaster level would be the same thickness as an arc drawn with the Brush Tool (size = 1, pencil mode on).
image

In Opentoonz v1.7.1, they are no longer the same thickness (the arc made by the Geometry Tool is 2 pixels thick):
image

This change in thickness affects all Geometry Tool modes except for Line and Polyline.  These two will still produce a 1 pixel wide line when thickness is set to 1:
image

This is the PR in which it appears to have been changed: https://github.com/opentoonz/opentoonz/pull/4333

This is the nightly in which the thickness was increased: https://github.com/opentoonz/opentoonz_nightlies/releases/tag/nightly_bd3207536e425f0c4af2d5d195708a239b3b074f

I tested the nightly before that in which the thickness is normal: https://github.com/opentoonz/opentoonz_nightlies/releases/tag/nightly_a72e35218b9b6cd91a8ef24376e7e8fcab9b2640

### Steps to Reproduce

1. On a ToonzRaster level, draw an arc with the Brush Tool (size = 1, pencil mode on)
image

2. Draw a second arc with the Geometry Tool (thickness = 1, pencil mode on)
image

3. Note the difference in size between the two

### Expected Behavior

When the thickness is set to 1, the Geometry Tool should make a shape that has a 1 pixel wide line.

The change in thickness was also noted in a reply to this report: https://github.com/opentoonz/opentoonz/issues/2887

### OpenToonz Version

Nightly (Specify below)

built March 31 2024

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Wacom

Abel032:
- Size/Thickness adjust.

compare

- Change default brush set.

图片

图片

Fixes #5498


Abel032:
> differences from Brush to Geometric Tool shapes

I think it's smaller with newest commits.

But the differences also exist between difference level type.
We can consider unifying different tools for different levels to one tool for different levels.
Like the Geometric Tool.

RodneyBaker:
That seems like a winner to me.
The real selling point for me is that if a result doesn't meet the intent of the user they can Undo and try again with a different method.

Posting intent to merge although may not merge soon as we are still waiting on the nightlies to start building again.  In the meantime I'll continue to test and make sure we haven't missed anything.

All in all I find these options work well.

Abel032:
The management of pixel size for Vector Level is not completed, we need to refactor this part in the future.

RodneyBaker:
Thank you @Abel032 for the update/fix.
LGTM

PR #5841, opened 2025-03-16
Fix file renaming logic of File Browser

Abel032:
Include the situation that target file(s) already exist.


RodneyBaker:
Thank you @Abel032 
LGTM

Issue #5839, opened 2025-03-15
Brush ctrl+alt size change shortcut doesn't work correctly


neyhden:
### Description

When using the ctrl+alt shortcut to change the brush size it doesn't work as one would expect, instead of making it bigger when moving right and smaller when moving left, it makes it bigger if your cursor is to the right of the point, and smaller to the left, regardless of where you move. It works fine with the eraser tool using alt though.

### Steps to Reproduce

1. Select the brush size
2. Hold ctrl+alt and move the mouse to change the size

### Expected Behavior

To work as it does with the eraser tool, resizing the brush according to the direction of the movement, not adding the difference to the center to the size.

### Screenshots, Video & Crash Logs

![Image](https://github.com/user-attachments/assets/74cb7185-2d9f-467d-82b7-d4851d4977ae)

### OpenToonz Version

1.7

### Operating System

Linux

### GPU

NVIDIA

### Graphics Tablet

None

RodneyBaker:
I'm not sure if this is a Linux only issue.
I'm fairly sure the effect of using Ctrl + Alt to resize the brush is refined in current nightly releases.

It's not clear what Linux build you are using but (if not a local build by you) check with the creator of that build and see if they are using the latest source code.

Also, see how https://github.com/opentoonz/opentoonz/pull/5761 might effect your experience.

Abel032:
I wonder why OT not use Alt, but chose Ctrl+Alt?

PR #5843, opened 2025-03-16
Change the Resize behavior to use Alt or Alt+Ctrl for Brush tool

Abel032:
Use Alt to adjust both max size and min size, and keep the ratio.
Use Alt+Ctrl to adjust max size.
Related to #5839 

- [x] Vector Brush Tool
- [x] Toonz Raster Brush Tool
- [x] Full Color Brush Toll


Abel032:
Close this as the changing logic is not clear enough.

PR #5845, opened 2025-03-16
Icons Refactor Part 2

konero:
Refactored so now 99% of all icons are made by a single function, instead of creating them inline all over the place. The exceptions are mimetypes and other browser icons, as the window needs more focused refactor to support high-dpi.

The icon color should now change at runtime when user switches theme, I've removed the icon theme checkbox from preferences.

The stylesheet will be parsed for this block:

```css
:TOONZCOLORS {
  -icon-base-color: @icon-base-color;
  -icon-active-color: @icon-active-color;
  etc...
}
```

You can provide custom properties here, and using the function getCustomProperty from ThemeManager you can convert it to a QColor: `getCustomProperty(QColor("icon-base-color"))`. If you open additional stylesheets, you can configure the following custom icon properties:

```
-icon-base-color
-icon-active-color
-icon-selected-color
-icon-on-color
-icon-close-color
-icon-preview-color
-icon-vcheck-color
-icon-lock-color
-icon-keyframe-color
-icon-keyframe-modified-color
```

Example:
otico4

Although I haven't adjusted the stylesheet to match this design, it's just an example. I've left the implementation in as it can become quite useful.

image

Also, the preview buttons in the viewer have increased width now, it's a super common toggle so this should make it easier to hit. Was a problem for small tablet users (16inch) with 100% scale.

And this fixes a multi-screen (that have different scale factors) issue where icons were blurry on the one with the lower scales.

<details>
  <summary>📝 Note about fractional scaling issues</summary>
It's too complicated to hide icons in menus with the way we did previously by using an empty pixmap. I've applied the same method here and it works for 1x/2x but it won't work for fractional. The real fix is to update Qt and patch the bug that prevents us using the proper method to hide icons from menus.

I'm also unsure as to how a user can activate fractional scaling without forcing it, currently it appears not supported and the sceneviewer is very broken. So, merging this technically introduces a 'bug' in that you won't be able to hide icons in menus at fractional scaling.
</details>

<details>
  <summary>📝 Notes on new usage</summary>
I've moved icons into new path structure that includes their actual size like: `path/to/16x16/icon.svg`. Since devs may and seem to submit SVG's using different size standards it's not always reliable to rely on QSvgRenderer for default size. The function in ThemeManager (which stores metadata on each icon) can give you the true size with getIconSize().

This PR swaps over from QIcon to QIconEngine - engine will load/render a pixmap at every interaction compared to QIcon, this allows us to get runtime updates much simpler than subclassing all our custom widgets to use signals.

The engine will do this:
- Creates a Cache Key based on parameters (icon_name, size, mode, state) + extra depending on cases like pane icons for DPR
- Uses the key to check QPixmapCache
- If none exists, it renders a new pixmap for the size/mode/state
- Then it adds it to the QPixmapCache using the key
- The next time it uses the same parameters - which gives the same key - it will find the pixmap in the cache this time and skip rendering a new pixmap, basically same behaviour as QIcon
- When the QPixmapCache is invalidated with clear() - the engine will re-render pixmaps as they don't exist, which means if the theme changes or other runtime actions happen the icons can self-refresh

Another point on path, if you place a icon into a folder called 'colored' it will skip colorization and maintain full opacity (off mode pixmaps are 0.8 opacity for example) like so: `path/to/16x16/colored/icon.svg`.
</details>


RodneyBaker:
Pressing forward with this refactoring.
Thanks @konero for your continuing improvements.

PR #5846, opened 2025-03-16
Fix startup crash due to incompatible DLLs (T2D)

konero:
More info here: https://github.com/tahoma2d/tahoma2d/pull/1711

We already tested a version, but no users responded... this is a direct port of the fix in Tahoma2D and reverts my version.

By: @manongjohn

### DIFF
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -1066,6 +1066,19 @@ bool TSystem::showDocument(const TFilePath &path) {
 #endif
 }
 
+bool TSystem::isDLLBlackListed(QString dllFile) {
+  QStringList dllBlackList = {"lvcod64.dll", "ff_vfw.dll", "tsccvid64.dll",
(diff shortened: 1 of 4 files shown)


RodneyBaker:
This might be something to take up with @manongjohn but I'm wondering if the blacklist could reference an .ini or .txt file that the user could modify.  This might not preclude having some hardcoded DLL files that are known to be incompatible but having a solution where we can direct a user to simply add a filename to a text file (or provide that text file) might be optimal.

Downsides to this external blacklisting?
I'm sure there are more than a few.
The most obvious would be that Opentoonz would need to be restarted at least once after the file was edited.

My primary concern is two-fold in that hardcoding files will potentially include files that should not be listed and alternatively fail to list files that should be included.

Regardless, I'm looking forward to merging this PR as it is the best bet we currently have toward resolving this frequently encountered problem.

RodneyBaker:
Thank you @manongjohn for the startup fix and @konero for the port.
LGTM

PR #5848, opened 2025-03-17
Update Commandbar Layout

konero:
Reorganizes the default Commandbar.

It seemed useful to have the Startup Popup and Manual as the first items like Clip Studio/iPad Apps do for beginners. I've added the Geometry tool types which is useful for FX compositors for making quick iris shapes and masks. The Pencil Test window has been added. And lastly adding the viewer Check modes.

image


RodneyBaker:
Thanks @konero 
Much appreciated.

PR #5851, opened 2025-03-18
Fix Linux Desktop Path Detection Failure

Abel032:
I only tested on wsl.

It's related to tahoma2d/[#1776](https://github.com/tahoma2d/tahoma2d/issues/1776)

### DIFF
--- a/toonz/sources/toonz/filebrowsermodel.cpp
+++ b/toonz/sources/toonz/filebrowsermodel.cpp
@@ -28,6 +28,9 @@
 #ifdef MACOSX
 #include <Cocoa/Cocoa.h>
 #endif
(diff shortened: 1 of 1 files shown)


RodneyBaker:
The only thing I'm not quite sure of with this PR is what 'Deal' refers to in the PR's title.
Perhaps this means 'define' or 'refine'?  Perhaps it encapsulates the idea of 'supply a default desktp path'.

I guess I'm looking out to that eventual day when we compile a changelog and hope most lines in that change log are self explanatory.

Abel032:
Sorry for the unclear description, I've corrected the titile.

RodneyBaker:

PR #5857, opened 2025-03-21
Optimize AppVeyor Build

andeon:
This PR includes a few changes to test if they improve build efficiency:

Simplified File Copying: Replaced multiple copy commands with a single xcopy for the .dll files in the libmypaint directory. This simplifies the process and reduces redundancy.

Improved OpenCV Installation: Installed OpenCV 4.10.0 with suppressed progress bars and extra output (--no-progress, --yes, and --force flags) to clean up the installation process and reduce log clutter.

Added OpenCV cache to the AppVeyor configuration. 
***Caching during PR builds is disabled by default to prevent unauthorized cache modifications, so technically, the cache will not work until it is enabled. However, enabling the cache for PRs introduces potential security concerns. I assume that after the merge, the cache becomes available for the build on the main branch and should work as expected for regular builds.

Removed the cache configuration for c:\tools\vcpkg\installed due to lack of expected functionality and build improvements.
***It could be the same situation as above, but it seems like it's not working since there's no indication of its use in the terminal log.

~~Optimized CPU Usage:  Replaced the build block with a new build_script to utilize msbuild Set maxcpucount to 2 to leverage more CPU cores, speeding up the build process.~~ It had no effect and removed the standard build logs (Messages) that are automatically included in the default build process. 

~~Added matrix configuration for Debug and Release builds to check if it influences build time and construction process.~~ However, it doesn't work as the free version only allows 1 concurrent job.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,10 +8,11 @@ configuration:
 - Release
 platform: x64
 clone_depth: 1
-cache: c:\tools\vcpkg\installed
+cache:
+  - C:\Tools\opencv
 install:
 - cmd: >-
-    choco install opencv --version=4.10.0
+    if exist "C:\Tools\opencv\build" (echo OpenCV cached, skipping install) else (choco install opencv --version=4.10.0 --no-progress --yes --force)
 
     cd thirdparty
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
This is a test.
This is (mostly) a test.

Merging to see how well it optimizes appveyor builds.
PRs that are currently building will likely need to be restarted.

Thanks @andeon

andeon:
I hope this will decrease the build time a bit. The queue has been busy lately, which is great because there are a lot of fixes and implementations coming in. I also asked the AppVeyor team to give us a bit more time (90 minutes), since it's an open-source project, and I'm waiting for a response.

PR #5858, opened 2025-03-21
[WIP]Sort vector Image with palette order

Abel032:
Add option to sort vector strokes by Vector Selection tool and prepare for vector brush with draw order.

https://github.com/user-attachments/assets/89a659b4-47f0-410d-a540-0d2331084b48


RodneyBaker:
A small but very powerful and useful addition @Abel032.
Looking very good.

beeheemooth:
Hmm... I'm curious how the vector indices change after applying this option, what happens to closed areas, and whether this option works in "Whole Level" selection modes.

Abel032:
> Hmm... I'm curious how the vector indices change after applying this option, what happens to closed areas, and whether this option works in "Whole Level" selection modes.

Indices don't change , the strokes swapped.

The whole level mode is not implemented, but would add.
And for closed area, it should keep as before, but currently I need to find a proper way to do it.

Abel032:
Update: Now the area won't be affected after sort.
But need to add support of Whole Level mode.

Abel032:
There still some problem when Vector Image contains regions.

Abel032:
The bug of losing painted region is related to moving strokes across a stroke group.
And it's a problem outside the PR.
See #5878 .

Abel032:
I'm about to create another PR to support the draw order option for vector brush.

RodneyBaker:
As the remaining issues of vector sort are outside of this PR will proceed with merger of this.
Thank you @Abel032 the new option.

Issue #669, opened 2016-07-22
Consolidate All Project Resources to Compressed Zip File


RodneyBaker:
This Feature request is primarily for research purposes as there may be several approaches to accomplishing the underlying goal of maintaining a project's resources in one location.

This feature relates to issues users face locating and maintaining all resources associated with any given project.  It roughly relates to several other issues identified and/or already addressed to inluce one currently being worked; (#373 Feature request - The ability to add a custom project root through the interface.

Issue:
Resources associated with a Project may be located in a variety of places and if/when the resources move from those locations or the links/pointers change the resource is missing from the Project.  In some cases the resource still exists (i.e. it wasn't deleted) but simply missing.  Any missing resource however may make the project unusable.

Possible Solution:  (Complex)
Using Subversion (SVN) in project workflow/management s likely an ideal solution because of the non-destructive nature of versioning files.  At any time a file that i altered can be recovered to an earlier state.  The issue with this solution is that considerable setup and maintenance is required to ensure files are properly processed.  Additionally, knowledge of versioning processes is required as well installation of supporting software (example:  Tortoise SVN, etc.)  Another negative to this approach is that the average user of OpenToonz will likely not use versioning until/unless more tightly interwoven into the UI and workflow of even the simplest of projects.

Possible Solution:  (Simple?)
Saving all associated files into a compressed file (ex:  Zip) with full paths stored will allow for all resources to be stored in one file with the capability of restoring files to original locations.  The file's extension could be .zip although naming the extension something otherwise would not change the type of file that it is.  For instance, if the file extension is .otz OpenToonz could recognize that for the compressed file that it is and handle it accordingly.

Alternatives:
There are several alternative approaches that may be superior to that of using a standard compressed file such as .zip to collect data.  This is especially true given that containers for image files/sequences are being produced to take advantage of image compression algorithms along with the use of advanced codecs.  MXF is one such container of which recently .MOX was fielded (by Brendon Bolles) for use in production.  NOX/MOX currently has plugins for Adobe After Effects and Premiere/Media Encoder.  The idea here is to store the image sequences along with other relevant files for easy access/playback.

[...]

PR #230, opened 2016-04-25
"Save All" menu which saves current scene and all the levels. (modified from pr#153)

shun-iwasawa:
This modification is for the issue #123 , and is based on walkerka's resolution https://github.com/opentoonz/opentoonz/pull/153

Major changes from walkerka's commit are as follows:
- Saving levels used in the scene is done in SceneResources::save().
- OpenToonz is now available to customize menu bar. The default menu bar will be built from template file (if it exists) when first launched. So I put "Save All" command in menu bar template files as well.

PR #5859, opened 2025-03-21
Adjust File Menu for easier understanding

Abel032:
图片 图片

- Adjust "Save Scene" to "Save Scene Only"
- Adjust Save Scene commands Order

### DIFF
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -1715,12 +1715,12 @@ void MainWindow::defineActions() {
                        "new_scene");
   createMenuFileAction(MI_LoadScene, QT_TR_NOOP("&Load Scene..."), "Ctrl+L",
                        "load_scene");
-  createMenuFileAction(MI_SaveScene, QT_TR_NOOP("&Save Scene"), "Ctrl+Shift+S",
-                       "save_scene");
-  createMenuFileAction(MI_SaveSceneAs, QT_TR_NOOP("&Save Scene As..."), "",
-                       "save_scene_as");
   createMenuFileAction(MI_SaveAll, QT_TR_NOOP("&Save All"), "Ctrl+S",
                        "saveall");
+  createMenuFileAction(MI_SaveScene, QT_TR_NOOP("&Save Scene Only"),
+                       "Ctrl+Shift+S", "save_scene");
+  createMenuFileAction(MI_SaveSceneAs, QT_TR_NOOP("&Save Scene As..."), "",
+                       "save_scene_as");
   menuAct = createMenuFileAction(MI_RevertScene, QT_TR_NOOP("&Revert Scene"),
                                  "", "revert_scene");
   menuAct->setEnabled(false);
--- a/toonz/sources/toonz/menubar.cpp
+++ b/toonz/sources/toonz/menubar.cpp
@@ -360,10 +360,10 @@ QMenuBar *StackedMenuBar::createCleanupMenuBar() {
   addMenuItem(filesMenu, MI_OpenRecentLevel);
   filesMenu->addSeparator();
   addMenuItem(filesMenu, MI_LoadScene);
-  addMenuItem(filesMenu, MI_SaveScene);
-  addMenuItem(filesMenu, MI_SaveSceneAs);
   addMenuItem(filesMenu, MI_SaveAll);
   addMenuItem(filesMenu, MI_OpenRecentScene);
+  addMenuItem(filesMenu, MI_SaveScene);
+  addMenuItem(filesMenu, MI_SaveSceneAs);
   addMenuItem(filesMenu, MI_RevertScene);
   filesMenu->addSeparator();
   addMenuItem(filesMenu, MI_ConvertFileWithInput);
@@ -496,10 +496,10 @@ QMenuBar *StackedMenuBar::createPltEditMenuBar() {
   addMenuItem(filesMenu, MI_SavePaletteAs);
   filesMenu->addSeparator();
   addMenuItem(filesMenu, MI_LoadScene);
-  addMenuItem(filesMenu, MI_SaveScene);
-  addMenuItem(filesMenu, MI_SaveSceneAs);
   addMenuItem(filesMenu, MI_SaveAll);
   addMenuItem(filesMenu, MI_OpenRecentScene);
+  addMenuItem(filesMenu, MI_SaveScene);
+  addMenuItem(filesMenu, MI_SaveSceneAs);
   addMenuItem(filesMenu, MI_RevertScene);
   filesMenu->addSeparator();
   addMenuItem(filesMenu, MI_ConvertFileWithInput);
@@ -672,10 +672,10 @@ QMenuBar *StackedMenuBar::createInknPaintMenuBar() {
   addMenuItem(filesMenu, MI_MergeColumns);
   filesMenu->addSeparator();
   addMenuItem(filesMenu, MI_LoadScene);
-  addMenuItem(filesMenu, MI_SaveScene);
-  addMenuItem(filesMenu, MI_SaveSceneAs);
   addMenuItem(filesMenu, MI_SaveAll);
   addMenuItem(filesMenu, MI_OpenRecentScene);
+  addMenuItem(filesMenu, MI_SaveScene);
+  addMenuItem(filesMenu, MI_SaveSceneAs);
   filesMenu->addSeparator();
   addMenuItem(filesMenu, MI_NewScene);
   filesMenu->addSeparator();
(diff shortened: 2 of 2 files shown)


RodneyBaker:
You are hitting all the keys areas @Abel032 
Another seemingly small but very important update.

RodneyBaker:
Pressing forward with this UI update as its sure to save some users from losing files when they only save the scene.

On a personal note, I believe this to be a better approach than that of Tahoma2D... not that the approach there is a bad one... I often use Save Scene As followed by a Save All to have more control over scene iterations and relagating some of those options to a submenu makes it more difficult to use.  

Once again, thanks @Abel032 for your work to improve Opentoonz.

pravdomil:
Word "Only" is not necessary here, "Save Scene" means "Save Scene", and should not do anything else.

If you add word "Only" then you should add "Only" to the rest of the menu for example "Revert Scene Only".

Also "Save All" is just "Save".

pravdomil:
made a https://github.com/opentoonz/opentoonz/issues/5866

RodneyBaker:
@pravdomil Thanks for your feedback.
You may likely be unaware of the problems new users have had historically losing files due to using Save Scene instead of Save All.  This change is an attempt to resolve that risk that doesn't go as far as the change (years ago) in the Tahoma2D fork that placed other save options in submenu dropdowns.

pravdomil:
thanks for context, then I think "Save Project" and "Save Scene" is the correct way of describing things.

RodneyBaker:
Ah, the joys of terminology.
Opentoonz doesn't have a 'project' file as other programs do.
A project for Opentoonz is a directory structure.

At some point we will gain the ability to zip up a directory into one file and that might then be referred to as a project file.

We don't have that capability at present.
The closest we have is the ability to manually zip up a directory (whether the original project directory or one created via Export Scene where there is a Project option)

We do not want to suggest to users that via a 'Save Project' they are saving a single project file.
An option to 'Save Project Directory and All Assets' is a bit too wide.
'Save All' is pretty self explanatory.

pravdomil:
Project can be a folder,
there is no problem with that,
and zip is not needed.

Therefore we can use Save Project.

There is no convention that it needs to be file.

beeheemooth:
Most newbie users don't understand the meaning of the "Save All" option. Intuitively, they prefer the "Save Scene" and "Save Scene As..." options, similar to other software. According to my statistics, about 90% of newbies (including me) save the scene incorrectly and get red frames. The issue of red frames is the number 1 question for all newbies. Usually, they do a lot of work in their first scene and when it is not saved, people experience great disappointment in the software, fueled by rumors like "Opentoonz is the most buggy animation software".

Here is an example where a user complained about red frames:

Screen_1452

Users also often ask: if there is a "Save All" option, then why is there no "Save All As..." option?

pravdomil:
continue in https://github.com/opentoonz/opentoonz/pull/5869

pravdomil:
continue in https://github.com/opentoonz/opentoonz/issues/5866

konero:
> The issue of red frames is the number 1 question for all newbies. Usually, they do a lot of work in their first scene and when it is not saved, people experience great disappointment in the software, fueled by rumors like "Opentoonz is the most buggy animation software".

In fairness, the software does warn you when you close it that stuff is still unsaved if you use only save scene.

andeon:
@konero @beeheemooth I think the issue is that, since the scene is intrinsically built, it should be seen as something aggregated, like levels.

When the user starts OpenToonz, they are already within a project, the sandbox. As a result, many users get lost in the concept that scenes are part of a project. This leads to overlap of files, red frames, etc. However, if, from the beginning, the user were instructed to think in terms of projects, everything would be separated, and there wouldn’t be overlap, in theory. This is because OpenToonz allows us to work with separate elements (i'ts good), unlike other programs that usually focus on a single file or make it appear that way.

Users would also understand that a scene can be imported or linked to other projects more easily. However, this would require changing the structure. I believe scenes should have their own tab, like levels, and everything related to them should be moved to that tab. There could be a "close scene" option to return to the project level, and within the project, there could be a "close project" option to return and open/create projects. OpenToonz is designed to divide the animation into scenes to maintain control over the structure and performance, but people don’t always take advantage of this. Often in "good rare scenarios," you see the user create a "scene" with 2000 frames when it should be split into a range of 4 to 5 scene files. Hope this makes sense!

pravdomil:
What about just have one "Save" command, like every other program has.

RodneyBaker:
> What about just have one "Save" command, like almost every other program has.

If one thinks of standard animation production workflow in a studio there are certainly times when you don't want a scene file to be saved but do want the levels which are now filled with color to be saved... and vice versa; there will be times when no imagery changes but the timing... or effects... changes.  In what ways could a single "Save' command account for this?  Ask the user every time to specify what they want to save?  No.  That's what we have now; the ability to save just a Scene or just a Level, etc.

This gets back to why a Save All was created in the first place... almost no one outside of studios or teams with established workflows where animation production was passed from one department to the next fell into this category of only saving the specific assets that department was working on.

 Here's a bit of historical context:  https://github.com/opentoonz/opentoonz/pull/230

RodneyBaker:
I am not opposed to having a Save and an Save All that do exactly the same thing... namely, save everything.  ;)  

Then we just have to occassionally explain to users that they both do the same thing because we don't want anyone to lose their files.   

Even then however, users will still find ways to lose files.
The primary case being that users can always move files outside of Opentoonz and thereby break the links to those files.

RodneyBaker:
I want to take a crack at your feedback @andeon although I hate to cherry pick so I'll mostly just quote the following:  

> I believe scenes should have their own tab, like levels, and everything related to them should be moved to that tab. There could be a "close scene" option to return to the project level, and within the project, there could be a "close project" option to return and open/create projects. OpenToonz is designed to divide the animation into scenes to maintain control over the structure and performance, but people don’t always take advantage of this. Often in "good rare scenarios," you see the user create a "scene" with 2000 frames when it should be split into a range of 4 to 5 scene files. Hope this makes sense!

An issue here might be that Opentoonz can only have one scene (one xsheet if you will) open at a time.  As such there won't be much reason for a Scene tab as that tab would just list external scene files.  A sub xsheet is the closest Opentoonz has to the ability to have multiple scenes open at a time although this is through a little smoke and mirror of just swapping the current scene to another while making the user believe they are still inside the same scene file.

In a way we have the Scene tab you suggest through the subxsheet navigation option where we can see the relationship of scenes relative the our current (master) scene and quickly move through them.

There is some opportunity here to expose more to the user... at many layers of aggregated objects.  This all the way from control points of vector strokes to collections of scenes (here being the general concept of "projects" that we might otherwise also understand to simply be file directories).

Placing multiple file directories into a Project File (read:  multiple scene files and their file structures all contained in one single file) would be an extrapolation of this request from yours truly:  https://github.com/opentoonz/opentoonz/issues/669

I will suggest here that what we want ultimiately is not a Project file but rather a Production file.  
I'll further suggest that consolidted production file should have a .pro  or .prod (or perhaps to really emphasize the point .production) extension.  

If we get the point where we have such a consolidated production file then we can have tabs that list specific contents of that aggregated file to include scenes.

andeon:
@RodneyBaker I said 'Tab,' but I actually meant to say 'menu.

I think the suggestion I gave could improve this; it would be interesting to prototype it to see how it will look. Just like when we need something from the level, we go to the level menu, and it could work for the scene as well. In 'File,' 'Save' would be understood as saving all the assets, since it would be in the main level. This is standard in other apps. Of course, it might confuse old users initially, but in the end, it wouldn't be a problem. 

So my best bet would be to prototype using a room-based system (Preferences > Interface > Rooms*), like the Default/Studio Ghibli approach, something called 'experimental,' for example. Everything discussed here in the 'level' menu could be implemented there. It would give time to consider the pros and cons and wouldn't negatively affect the overall experience.

RodneyBaker:
Thanks for the clarification @andeon 

Your prototyping would seem to be ideal for the portable releases although at present those are only for Windows.  My Mac won't run current releases of Opentoonz so I can't even test there.

An additional benefit of the portables is that they can remain available even after the expiry date of CI builds so users can reach back and use those 'experimentals' whenever they feel the need.

I asked @manongjohn if he'd be willing for us to use "OTX" as I would love to see that branding used for its original purpose again but he said he'd prefer not.  Unless someone can suggest another name we could just label it "ot-experiemental" with the current date and outline what is experimental about it in the description of the release.

The portables also are designed to capture snapshots and moments of time in Opentoonz development.  In this way features and explorations that might never be included in the final Opentoonz release can be available and used as needed.  

One such portable which I haven't pressed toward recently is a lite version of OT that removes pretty much everything except the very core components needed.  That and maybe a version of OT that is a standalone viewer (specifically for viewing TLV and PLI levels in scenes).  These versions being ultimately designed for teaching aspiring developers about the core components of OT.

andeon:
@RodneyBaker I was just referring to creating this 'experimental' for profiles/layouts/rooms in this context. It's very simple; perhaps I'll create this just to better demonstrate in practice.

However, regarding the OpenToonz Experimental version, it might be interesting to introduce things that are out of scope or extreme changes. For example, upgrading to QT6 would fit the purpose of the Experimental version and help prepare for the future. I used OTX for a long time, years ago, when it brought improvements that might not have fit into OpenToonz at the time. But I think we're starting to get a little out of scope here.

PR #5862, opened 2025-03-22
Update Building on macOS

pravdomil:
see diff


RodneyBaker:
I'm not seeing actual change to build instructions other than formatting (not counting the capitalization of the word 'store') so pressing forward.

Congrats on your first contribution to Opentoonz @pravdomil!

PR #5869, opened 2025-03-22
Resolve Save Scene and Save All confusion

pravdomil:
there was some confusion about people losing data because
they forgot to "Save All" and did only "Save Scene"
this PR addresses that

### DIFF
--- a/toonz/sources/toonz/menubar.cpp
+++ b/toonz/sources/toonz/menubar.cpp
@@ -1084,8 +1084,8 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   addMenuItem(fileMenu, MI_NewProject);
   addMenuItem(fileMenu, MI_LoadProject);
   addMenuItem(fileMenu, MI_OpenRecentProject);
-  addMenuItem(fileMenu, MI_ProjectSettings);
   addMenuItem(fileMenu, MI_SaveAll);
+  addMenuItem(fileMenu, MI_ProjectSettings);
   fileMenu->addSeparator();
   addMenuItem(fileMenu, MI_NewScene);
   addMenuItem(fileMenu, MI_LoadScene);
--- a/toonz/sources/toonz/menubar.cpp
+++ b/toonz/sources/toonz/menubar.cpp
@@ -1093,6 +1093,7 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   addMenuItem(fileMenu, MI_SaveScene);
   addMenuItem(fileMenu, MI_SaveSceneAs);
   addMenuItem(fileMenu, MI_RevertScene);
+  addMenuItem(fileMenu, MI_SaveDefaultSettings);
   fileMenu->addSeparator();
   addMenuItem(fileMenu, MI_LoadFolder);
   addMenuItem(fileMenu, MI_LoadSubSceneFile);
--- a/toonz/sources/toonz/menubar.cpp
+++ b/toonz/sources/toonz/menubar.cpp
@@ -1082,6 +1082,8 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   // Menu' FILE
   QMenu *fileMenu = addMenu(tr("File"), fullMenuBar);
   addMenuItem(fileMenu, MI_NewProject);
+  addMenuItem(fileMenu, MI_LoadProject);
+  addMenuItem(fileMenu, MI_OpenRecentProject);
   addMenuItem(fileMenu, MI_ProjectSettings);
   addMenuItem(fileMenu, MI_SaveAll);
   fileMenu->addSeparator();
--- a/toonz/sources/toonz/menubar.cpp
+++ b/toonz/sources/toonz/menubar.cpp
@@ -1081,9 +1081,12 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   QMenuBar *fullMenuBar = new QMenuBar(this);
   // Menu' FILE
   QMenu *fileMenu = addMenu(tr("File"), fullMenuBar);
+  addMenuItem(fileMenu, MI_NewProject);
+  addMenuItem(fileMenu, MI_ProjectSettings);
+  addMenuItem(fileMenu, MI_SaveAll);
+  fileMenu->addSeparator();
   addMenuItem(fileMenu, MI_NewScene);
   addMenuItem(fileMenu, MI_LoadScene);
-  addMenuItem(fileMenu, MI_SaveAll);
   addMenuItem(fileMenu, MI_OpenRecentScene);
   addMenuItem(fileMenu, MI_SaveScene);
   addMenuItem(fileMenu, MI_SaveSceneAs);
@@ -1095,14 +1098,6 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   addMenuItem(fileMenu, MI_ConvertFileWithInput);
   fileMenu->addSeparator();
   addMenuItem(fileMenu, MI_LoadColorModel);
-  fileMenu->addSeparator();
-  QMenu *projectManagementMenu = fileMenu->addMenu(tr("Project Management"));
-  {
-    addMenuItem(projectManagementMenu, MI_NewProject);
-    addMenuItem(projectManagementMenu, MI_ProjectSettings);
-    projectManagementMenu->addSeparator();
-    addMenuItem(projectManagementMenu, MI_SaveDefaultSettings);
-  }
   fileMenu->addSeparator();
   QMenu *importMenu = fileMenu->addMenu(tr("Import"));
   { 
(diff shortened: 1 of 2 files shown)


RodneyBaker:
I appreciate your enthusiasm.

Rather than spend a lot of time working on this PR only to be frustrated if it doesn't get merged you might want to discuss some of the issues related to this via discord?

Invite link:  https://discord.gg/mesMUaBVZY

We want to set you up for success and I don't see the 'Save All' option going away any time soon.

konero:
There's no concept of "Save Project" in Toonz, Save All refers to saving all assets loaded within the current scene, not the project, it was added because users didn't want to save levels/scene manually. To save project information you need to use the "Save Default Settings" command under Project Management.

It'd make more sense putting Save Default Settings in the Preferences group, it's about where Blender puts it.

You also removed the separator that was separating Clear Cache Folder command from other items, it's a unique destroyer command unrelated to Preferences and should be logically separate.

pravdomil:
updated the PR

Abel032:
There does have people who use OT's project as a scene(one cut) even if they are in the same of an Animation Project(such as an episode), but it's away of OT's project's original design I think.

Currently the assets of scene is stored in different folders of one project(inputs,drawings,extras,outputs), which makes it complex to transfer specific scene between different machines.
But this need is quite common.

As I know, OT's project file maintains the folder path that related to project's assets's locations and stores defualt scene properties(you can check the file names sandbox_otprj.xml).

There is a need to improve the scene management, to make one scene's assets more easier to port between different project(by copying just one scene folder).
And for those folders in project, they are global and able to be reused in every scene(for example, the color model and the character design).
We can make the feature of specifying the assets' locations also available for scene(force them to store beside the scene files) 
and then expend functions of project file and the it maintains.

Abel032:
There does have people who use OT's project as a scene(one cut) even if they are in the same of an Animation Project(such as an episode), but it's away of OT's project's original design I think.

Currently the assets of scene is stored in different folders of one project(inputs,drawings,extras,outputs), which makes it complex to transfer specific scene between different machines.
But this need is quite common.

As I know, OT's project file maintains the folder path that related to project's assets's locations and stores defualt scene properties(you can check the file names sandbox_otprj.xml).

There is a need to improve the scene management, to make one scene's assets more easier to port between different project(by copying just one scene folder).
And for those folders in project, they are global and able to be reused in every scene(for example, the color model and the character design).
We can make the feature of specifying the assets' locations also available for scene(force them to store beside the scene files) 
and then expend functions of project file and the it maintains.

Abel032:
I think OT could be very good at sorting assets for project, even all the works is done outside of OT and OT is only used for preview or pencil tests.

RodneyBaker:
With the pending addition of #5877 this PR will not be needed as users can easily customize menus to be whatever they desire that text to be.

There is of course the question of what the defaults in the deeper source code distributed via new releases should be.  While discussions continue on what is optimal the plan would be to leave menus as they are except where there is a critical need.  There are many things that can and should be changed in the default menu.

Presently we have several ways forward to change menus at the user level.  This can be via custom Room sets and (soon) custom menu labels.

Your thoughts @pravdomil?

pravdomil:
my wish is to make OpenToonz as user/beginner friendly as possible,
this PR is addresses that

solution people can resolve confusion their selfs by customizing menus,
is not a solution in terms of user friendliness

I'm getting busy with my film,
I need to focus now on that,
so I hope contribute to the discussion positively,
for more inspiration I recommend to watching:
[Achieving Excellence in Open Source Software Design](https://www.youtube.com/watch?v=12TJ-zTgiH0)
do user tests, do prototypes, do the best to serve your users and community and animation!

toonz/sources/toonz/menubar.cpp
```diff
@@ -1081,29 +1081,24 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
   QMenuBar *fullMenuBar = new QMenuBar(this);
   // Menu' FILE
   QMenu *fileMenu = addMenu(tr("File"), fullMenuBar);
+  addMenuItem(fileMenu, MI_NewProject);
```
Abel032, line 1084: This would help user understand we have a project then scenes, which is good.
But I think the most frequently used commands is about scenes, like NewScene and OpenRecentScene.

toonz/sources/toonz/menubar.cpp
```diff
@@ -1081,29 +1081,24 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
...
+  addMenuItem(fileMenu, MI_LoadProject);
+  addMenuItem(fileMenu, MI_OpenRecentProject);
+  addMenuItem(fileMenu, MI_SaveAll);
```
Abel032, line 1087: This SaveAll command doesn't relate to any project thing, only means save the current scene and the all assets(Levels/CleanUpParams...) in current scene.

toonz/sources/toonz/menubar.cpp
```diff
@@ -1081,29 +1081,24 @@ QMenuBar *StackedMenuBar::createFullMenuBar() {
...
+  addMenuItem(fileMenu, MI_ProjectSettings);
+  fileMenu->addSeparator();
   addMenuItem(fileMenu, MI_NewScene);
   addMenuItem(fileMenu, MI_LoadScene);
-  addMenuItem(fileMenu, MI_SaveAll);
   addMenuItem(fileMenu, MI_OpenRecentScene);
   addMenuItem(fileMenu, MI_SaveScene);
   addMenuItem(fileMenu, MI_SaveSceneAs);
   addMenuItem(fileMenu, MI_RevertScene);
+  addMenuItem(fileMenu, MI_SaveDefaultSettings);
```
Abel032, line 1096: This command only save scene's default settings for current project.

RodneyBaker:
> for more inspiration I recommend to watching: [Achieving Excellence in Open Source Software Design](https://www.youtube.com/watch?v=12TJ-zTgiH0) do user tests, do prototypes, do the best to serve your users and community and animation!

That's a great video and we do try to focus on that prototyping and user testing although its not as formal perhaps as it could and likely should be.

A key area of that video (around 19:30) is where he discusses how 7 of the worlds best designers will have different ideas for how to proceed with a design.  He then shifts to empahsize why its so important to prototype and user test.  Later at about 26:30 he states the '7 designers with different ideas' again to emphasize 'default behavior'.  He covers a few things I fully believe that I've struggled to convince people of over the years such as that 'no one reads tutorials' (read:  documentation) although I have softened over the years because a few important people do often read tutorials and that can help a community learn.

In the video there are some counter points that Martin addresses (question from the audience) such as Inkscape being a specialists tool.  There is a wide swath between a new user with no experience with graphics and animation versus one with experience.  Opentoonz tries to find ways to address the needs of both... and all users inbetween.

Toward the end of the video (answering the last question) Martin discusses one of his major
advantages in developing open source software... that he is 'hired by a company'.  When you have one specific owner of a process this helps tremendously in focusing on a vision for development.  Opentoonz has this in the form of Dwango (and as represented by the software's primary maintainer Shun Iwasawa) but much more loosely than Martin's approach to allow input from the Opentoonz community.

Much more can be said about this (and will be) but we can all agree that user testing is important.
And that is what leads us to where we are currently.  While user testing is extremely important its also important to understand the workings of the program itself and to address the underlying differences between Opentoonz and other programs.  

The nightlies are one of the best ways we currently have to user test and use of the nightly releases is my recommendation not only to resolve problems encountered but to sample and test the software as it move forward through recurring cycles of continual improvement.

Thanks for that video.
I had been following some of the drama related to Martin taking over Audacity and how that effected that community.  Its good to see more of the positive side of that and especially Martin's views on his approach.

Aside:  For his user testing demo of Inkscape that illustrates why I'm not a fan of throwing options at a user as they are running the program.  This is something popular these days where users have to choose between a lot of options even before they get the program started.  In the graphics world this can be seen in programs like Coreldraw and Clip Studio Pro.  If you just want to open the program and draw... not optimal.  Better to let the user open the program and then adjust.  Martin disagrees and emphasizes 'default behavior' but those popups with lots of options aren't default.  Martin is a fan of forcing a default upon the user even to the point of limiting Preferences.  Opentoonz hopes to have a great default but also allow maximum customization.
The option to turn the Startup Popup off might be a good example of this.  It can be turned back on and easily reopened but it can also be ignored entirely.  

How does this related to saving...
I am easily convinced that having a single 'Save' will help.
A problem with that is that what is being saved is variable and users require that to be variable.
In this context 'Save All' is arguably the best option.
Should a Save option be available that opens and gives the user options for what to save?  Levels?  Palettes?  Scenes?  That would likely complicate and confuse user more as they encounter those options.

A single 'Save' option will save a lot of user frustration but in my estimation that will be by accident and the merits of success by accident can be discussed.  If the user doesn't know what they are saving they can and will potentially save things not wanted.  This would be different if saving in Opentoonz was somehow fully non-destructive but currently it is not.  Read about the  measure introduced to create .bak backup files for more insight here; a step in the right direction but not a final solution.

For more about this I'll recommend reading into the history of why 'Save All' became a UI option and why the Studio Ghibli Edition Opentoonz source derived from didn't have that option.

RodneyBaker:
Closing this as there is no plan to implement the change.
If anyone has thoughts they want to share please do.

PR #5870, opened 2025-03-22
Rephrase Save Default Settings menu command

pravdomil:
see diff


RodneyBaker:
This rephrasing better captures what is happening with default settings.
Thanks @pravdomil for the change.
LGTM

PR #5873, opened 2025-03-22
Add better getters for icon properties

konero:
Slightly tweak #5845 by adding better property getters and slightly adjust disabled opacity to make it more feint. Opacity levels can now be controlled with stylesheets.


RodneyBaker:
Thanks @konero for the addition.
LGTM

Issue #5866, opened 2025-03-22
File Menu Redesign


pravdomil:
### Is your feature request related to a problem? Please describe

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/5164c5cd-1777-4356-950c-7e21bf55b71d" />

### Describe the solution you'd like

redesign of the file menu:

```
New Project
Project Settings
---
New Scene
Load Scene...
Load Recent Scene >
Save
Save Scene
Save Scene As...
Revert Scene
---
Load Folder
Load As Sub-xsheet
Load Color Model...
---
Import >
Export >
---
Print Xsheet
Print Current Frame...
---
Script >
---
Convert File...
Configure Shortcuts...
Clear Cache Folder
```

RodneyBaker:
Thanks for the suggestion @pravdomil.
There have been a number of recommended changes to the file menu over the past few years.
Unfortunately, UI changes such as these can be quite contentious and everyone has their preferences.
Fortunately, there is a remedy that should satisfy everyone.

The current suggestion for major changes to the file menu is to create a new room set / theme that can be accessed via Preferences.  Making this room set available to users in addition to the current room sets (primarily the default and the studio ghibli rooms) will allow users to try the changes which  (potentially) lead to the default room set itself being updated.

Aside:  A recent change to the File menu was a critical change because new users often would fail to use Save All and instead only save the scene file.  This minor but important exception was deemed important enough to immediately incorporate the change.

There has been some talk of creating a 'community' room set and perhaps your suggestions might be included in that?

pravdomil:
I think I understand people misunderstanding the UI
because the UI is not hierarchy structured
and Project which is the top level entity is in sub menu,
and "Save Project" command is missing

I think making a new room set/house is a good idea,
so people do not lose the brain muscle memory
and can keep using the existing UI

I would call it "House 2.0" and the default rename to "House 1.0".

pravdomil:
maybe then revert #5859
to keep people muscle memory

I would also like to have House 2.0 as the default room set,
so we don't confuse people

RodneyBaker:
That might be our first obstacle, if we can decide on a name we might have a chance of actually getting a new room set.  ;)

Note that there is nothing that prevents anyone from creating and sharing a room set.
Since room sets are optional for users it can easily be approved for inclusion.

[...]

PR #5877, opened 2025-03-23
feat: Add support for custom display labels in menu commands

andeon:
- Modified `loadMenuBar` and `readMenuRecursive` methods in `StackedMenuBar` to support an optional `label` attribute for menu commands in XML. When a `label` attribute is provided, it overrides the default text of the corresponding `QAction`.

This allows for more flexible and customizable menu bar configurations, with "user-friendly labels" for commands.

Here's a more practical and step-by-step example of how this works (Changing "Save All" to "Save").

1) Create a copy of the `Default` folder inside `stuff\profiles\layouts\rooms` and rename it (this may be Experimental).
menubar_template.xml

2) Open the **menubar_template.xml** file inside the 'Experimental' folder and edit the following.

`<command>MI_SaveAll</command>` to `<command label="Save">MI_SaveAll</command>`

3) Open OpenToonz, go to Preferences > Interface > *Rooms, and choose Experimental.

4) Close OpenToonz and reopen it to refresh the settings.

Now, instead of displaying "Save All", it will display "Save".

example menu

This process allows the user to customize their menus with greater flexibility than before, by modifying labels directly through an XML file. With this approach, users can adapt the interface to their preferences. 

Ref: #5859, #5866

**PS**: Keep in mind that custom setups can limit access to future updates. You need to be cautious because if OpenToonz comes with new configurations, like the great work @konero did with the Update Rooms Layout, those updates might not apply to users who are using their custom profile room. If we look at Studio Ghibli's room setup, it's not viable for general users, but rather just for the studio.

### DIFF
--- a/toonz/sources/toonz/menubar.cpp
+++ b/toonz/sources/toonz/menubar.cpp
@@ -269,8 +269,8 @@ QMenuBar *StackedMenuBar::loadMenuBar(const TFilePath &fp) {
   }
 
   QXmlStreamReader reader(&file);
-
   QMenuBar *menuBar = new QMenuBar(this);
+
   if (reader.readNextStartElement()) {
     if (reader.name() == "menubar") {
       while (reader.readNextStartElement()) {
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I like this approach as it further promotes one of Opentoonz most promanent features... user customization.  Few programs are as customizable as Opentoonz and that isn't even accounting for the ability to change EVERYTHING in the source code.

In short, in many cases, if a user doesn't care for the defaults... those settings can be changed.
This addition continues that trend toward trusting the user to make decisions on what works best.

> users can adapt the interface to their preferences.

Nicely done.

RodneyBaker:
Thanks @andeon  Much appreciated.

PR #5883, opened 2025-03-27
Increase drag detection threshold of Arc

Abel032:
This PR is related to #5823.
Adjust the threshold from 7.0 units to 10.0 units.
But also suggest user to avoid small dragging while clicking a point with a tablet.

CC: @Ahl76

### DIFF
--- a/toonz/sources/tnztools/geometrictool.cpp
+++ b/toonz/sources/tnztools/geometrictool.cpp
@@ -2839,12 +2839,12 @@ void MultiArcPrimitive::leftButtonDrag(const TPointD &pos,
                                        const TMouseEvent &e) {
   switch (m_clickNumber) {
   case 0:
-    if ((tdistance2(m_startPoint, pos) < sq(7.0 * m_tool->getPixelSize())))
+    if ((tdistance2(m_startPoint, pos) < sq(10.0 * m_tool->getPixelSize())))
       return;
     break;
   case 1:
     if (m_undoCount != 1 &&
-        (tdistance2(m_endPoint, pos) < sq(7.0 * m_tool->getPixelSize())))
+        (tdistance2(m_endPoint, pos) < sq(10.0 * m_tool->getPixelSize())))
       return;
     break;
   }


Ahl76:
@Abel032 Thanks for the quick fix. Unfortunately the appeveyor when it related to the stylus doesn't work properly with me. I have to wait for this PR to merge before I can test it in NB.

Abel032:
I think the Github Action Artifact would be able to work.

Ahl76:
@Abel032 Thanks for the explanation to test, it works, but unfortunately it doesn't really change the problems encountered. I always get straight lines during the tracing phase with the multi-arc tool.

Abel032:
I can increase it more or add a preference(this would make the ui more complex), but can you click slower or make sure your click don't drag too much(vertical click) on the tablet?
The threshold is 10.0 stage inch now.
(1inch = 53.333333 stage inch)

Abel032:
When you drag on the canvas using Arc tool, you would find that u need the drag a distance to see the line, that distance is the threshold.

Ahl76:
My opinion on this and sorry for my late reply. I've tested a few linearts (characters, backgrounds...) and the original multi-arc method of this PR doesn't work as well compared to its version without this PR, at least in my environment. I've also tested changing the wacom driver but it's still the same problem in my case.

 It's frustrating because a straight line can occur after several series of correct strokes. But it still happens often enough to make me feel uncomfortable and unconfident. Imo, the tracing process should be a natural one, it's boring enough not to have to focus on the details of your movements everytime, I'm not in favour of increasing the delays, it's already noticeable at the moment although acceptable. 
 
It's a shame not to have both versions working optimally at the same time, but imo, I think a preference is better as the difference is noticeable for me with the multi-arc, the arc tool bothers me less. Maybe a version with this PR or the previous one for those who don't have the problem and a version with the original version to be able to use it optimally.

andeon:
> I have to wait for this PR to merge before I can test it in NB.

@Ahl76 I’m curious about this issue. Which version did you choose to test: **Testing AppVeyor Builds** or the second one, **Downloading and Testing GitHub Action Builds**?

I remember that in the past, there were some differences between the version of OpenToonz generated by AppVeyor and the GitHub Action version. One of these problems was with the Mypaint brushes, which was fixed when I altered the workflow months ago. But now I’m not sure if that was the only issue. I noticed that in the portable version we use -DWITH_WINTAB=Y. If I’m not mistaken, the Windows installer also uses this, but the version generated by AppVeyor apparently doesn't. Maybe this isn’t related to the current issue, but it’s worth checking for these inconsistencies too.

andeon:
Thanks to report @Ahl76!

I think @RodneyBaker remembers this better. This wasn’t introduced with the fix for installing custom Qt 5.15.2 with WinTab support by shun-iwasawa, right? In Qt 5.9 versions, this wasn’t used, if I’m not mistaken. 

Change this:

`cmake ..\sources -G "Visual Studio 16 2019" -Ax64 -DQT_PATH="C:\Qt\5.15.2\msvc2019_64" -DBOOST_ROOT="C:\Libraries\boost_1_85_0" -DOpenCV_DIR="C:\Tools\opencv\build"`

To this:

`cmake ..\sources -G "Visual Studio 16 2019" -Ax64 -DQT_PATH="C:\Qt\5.15.2\msvc2019_64" -DBOOST_ROOT="C:\Libraries\boost_1_85_0" -DOpenCV_DIR="C:\Tools\opencv\build" -DWITH_WINTAB=Y`

Wouldn't necessarily fix this particular issue, but it might help avoid future problems related to tablet support when testing appveyor versions..

RodneyBaker:
Proceeding with this PR.
Thanks @Abel032 for the improvement to Geometric Arc.

Ahl76:
Why are you merging a PR that does nothing to solve the problem of accuracy and reliability and will, on the contrary, add extra latency ? Again this has to be an option if it can't be fixed because it's simply not usable as it is and forces users to use a new method, which by the way isn't any faster. I don't think imposing methods on users is the right direction. Users should be able to choose what suits them best; if some people prefer the method to the csp, that's fine, but you can't make those who used the initial method for a long time out of date. And in this case when it changes an existing workflow.

That why i suggested this as an option because it's obvious that with the current behavior we will prefer anyway to use the new implementation for realibility.

RodneyBaker:
Thanks for the feedback @Ahl76 
I merged this PR because before the Geometry Arc was almost unuseable.
Now (with this PR) it is useable.

This is certainly not to suggest it still cannot be improved.
@Abel032 is currently the most likely to continue improvement of the tool although effort from any developer will be appreciated.

Abel032:
@Ahl76 I have made another PR which introduce a preference of whether use drag-click to create arcs.

PR #5890, opened 2025-03-30
Vectorbrush Draw Order

### LINKED ISSUE #5837: Draw Order for Vector Brushes

Its just inconvenient to select a vector line and send it to back manually. Would like draw order options like raster brushes have.

Draw order options for vector brushes in the same way that raster brushes have.
Basically anything that would let me draw under lines instead of always on top.
I noticed Tohoma has this feature, but it would be nice to just use one software.

Abel032:
Resolves #5837

Add an tool option "Draw Order" for Vector Brush Tool which behave as Toonz Raster Brush Tool's Draw Order.


RodneyBaker:
@Abel032 Can you check something for me?
I'm getting a persistent and repeated crash with both the Appevyor and Github builds.
The crash always occurs with the following steps:

- Create vector level
- Create a second color style and set it to any other color than Style 1's black
- Select Style 1 and create a new Style which will be black by default
- Attempt to draw a closed circle with Style 3 (the second black).

I get a crash immediately upon releasing the pen at end of drawing the circle.

Edit:  The above is apparently not necessary to recreate the crash
Try just dupliating color Style 1 once and draw with that (Palette Order)

This appears to only be an issue when using Palette Order.
The revised steps:
- Create vector level
- Duplicate Style 1 so Style 2 is also black
- Draw a stroke with Style 2

Opentoonz should crash.

Happens with both pen and  mouse.

I'm currently restarting my system and doing due diligence to rule out anything on my end.

Abel032:
Thank you @RodneyBaker 
I can reproduce the crash.
It's caused because I lost an **else**.

RodneyBaker:
Thanks @Abel032 
I confirm that fixes the problem.

Looking very good.

RodneyBaker:
Anyone else testing this PR?
Would love to see additonal feeback.

RodneyBaker:
I am ready to merge this PR but want to give @shun-iwasawa time to restart the build system.

RodneyBaker:
It is time.
This may not yet build as a nightly but I have reason to believe we will shortly see the nightlies building again soon.

As always, thanks @Abel032 for all you do.

PR #5902, opened 2025-04-09
Add Mode and Pick Options to Finger Tool

Abel032:
image


RodneyBaker:
My initial observation is that this change helps justify having the Finger Tool on the toolbar.
The Pick option in particular is an interesting and useful addition.

I haven't been much of a user of Finger Tool so I'm having to figure out workflow but... so far... so good.

There are times when I can't figure out the logic of why a pixel is changed so that might be my focus at the moment.

The area option is potentially useful for a number of workflows to include shading effects.

konero:
Still not had time to test, what does this do exactly?

Abel032:
@konero It's similar to PaintMan's extender function, to fill small cracks between lines.

Abel032:
The algorithm could be improved.

RodneyBaker:
@Abel032 There are several of your PRs that failed appveyor tests when that wasn't working.
Can you refresh these?
Thanks!

RodneyBaker:
Thanks for the refresh @Abel032 
and the addition to the Finger Tool!

PR #5904, opened 2025-04-10
build: Update Boost to v1.86.0 in AppVeyor configuration

andeon:
This experimental PR tests whether upgrading Boost from 1.85.0 to 1.86.0 can reduce build times in AppVeyor.

Version 1.86.0 is [the latest supported version](https://www.appveyor.com/docs/windows-images-software/#boost)) in AppVeyor's environment.

In the last PR, even with optimizations, we only gained about 30 to 60 seconds.

Ref: #5857

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -26,7 +26,7 @@ install:
 
     mkdir %PLATFORM% && cd %PLATFORM%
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thanks @andeon

PR #5905, opened 2025-04-10
fix: handle MSVC warnings C4723 and C4552

andeon:
This PR addresses and resolves the C4723 and C4552 warnings encountered during the build process with MSVC.

- Fixed potential divide-by-zero issues (MSVC warning C4723) returning 0.0 as fallback to ensure safe division.
  - `igs_ifx_common.h` - (lines: 67, 70, 73, 76, 79, 92, 95, 98, 101, 111)
  - Removed unreachable break statements after return in switch blocks.

- Commented out `row + 1` to eliminate warning C4552.
  - `txsheet.cpp` - (line 396)
This line seems to be either debugging code or placeholder logic, and it currently has no effect.

### DIFF
--- a/toonz/sources/stdfx/igs_ifx_common.h
+++ b/toonz/sources/stdfx/igs_ifx_common.h
@@ -9,6 +9,8 @@
 
 #define RGBA_ORDER_OF_TOONZ6
 
+#include <cassert>
+
 namespace igs {
 namespace image {
 #if defined RGBA_ORDER_OF_TOONZ6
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Thanks again @andeon

andeon:
@RodneyBaker  The line `row + 1;` was likely intended as part of logic to adjust row/frame tracking, update a counter or test/debug. However, as it stands, it's just an unused expression and does nothing with the result. Let's not assume too much, maybe shun-iwasawa knows why it was there. That's why I commented it out with a `TODO`, so it doesn’t vanish and can be checked with more attention later.

PR #5906, opened 2025-04-11
Enhance xsheet selection

### LINKED ISSUE #5894: New logic of XsheetViewer for discussion
Image

Abel032:
## What this PR changed

- Change the behavior of dragging a single Frame Cell which has number, make it behavior as shift-drag: 
DragFrameCell

- Add a preference to always drag the Frame Cell (xsheet-图片)
DragFrameCellDirectly

- Rename/Delete the hold cells when renaming/deleting the Frame Cell:
RenameAndDeleteFrameCell

- Allow the negative drag of Smart Tab and move the SmartTab to cell's middle(little bigger): 
DragSmartTab

- Allow to assign cell by dragging the drawing from level strip directly:
DragDrawing

Resolves #5894


Abel032:
Can be refactored.

Abel032:
A better choice of Always Drag Frame Cell might be just enable to drag on the number.

andeon:
I'm seeing error messages. It seems that the project has likely exceeded the Git LFS quota. @RodneyBaker, I’m not sure if you have access to confirm if this is the case?

If the development continues at this pace, I think we should avoid working directly in PRs and only send them when they’re very close to being finished, to avoid frequent rebuilds and excessive use of the LFS quota. Maybe we could find another way to mitigate this both here and in our own branches. 

Another thing I thought about: the mechanism I created to ensure the macOS build gets generated might also be contributing a bit to the quota usage, since sometimes we see messages like 'Attempt 7 to build DMG'. I believe those attempts also count toward the quota, because the job isn’t interrupted and keeps time running, right? This isn’t really our fault, as GitHub still hasn’t resolved the issue: https://github.com/actions/runner-images/issues/7522

This month, I also exceeded the quota on my account and can't do much right now. It's the first time this has happened to me :)

Abel032:
> Maybe we could find another way to mitigate this both here and in our own branches.

Is it possible to enable the cache for Git LFS?

andeon:
@Abel032, I'm going to open an issue to discuss this separately, just to keep your PR clear of side discussions.

Abel032:
In OpenToonz, the primary (or official) way to manage timing is not by dragging frame cells, but by selecting them and using commands like **+-X** / **X's** to reframe or re-step, which is efficient when when animators tend to finalize timing in a few times.
But this may lack visual feedback and is not very intuitive.
It can also become cumbersome when animators need to adjust timing more frequently.

In most other animation software, we typically adjust timing by directly dragging frame without using cell commands — a more straightforward and beginner-friendly approach.

Abel032:
I think the rest works are refactoring, the method of identify Frame Cell should be one function.

Ahl76:
Can you leave the behaviour of the original shift when this preference is not used?

Abel032:
@Ahl76 Did you mean the shift drag?

Ahl76:
It seems to me that to use implicit cell without the preference you need to drag the cell border. What I mean is that you can also do it as before from any part of the cell by clickor holding shift. 

Unless shift is used for another command, it would help on small screens not to focus only on the edges. I use the shortcut ctl+shift to scroll drawings from my tablet but the shortcut also works to activate implicit holds. This fits in better with my workflow.

Abel032:
@Ahl76 Can you share some videos?
I can guess some behaviors you are using but I'm not sure.

Abel032:
It's possible to keep the original behavior of Shift(and Ctrl+Shift).

Ahl76:
I thought the old shortcut was shift introduced in implicit hold PR, but it looks more like shift+ctl, so the last one will be fine and preserve scrolling at the same time. Below gif using it.

Abel032:
@Ahl76 I have reintroduced the old features, have a check if you like.

Ahl76:
@Abel032 
Yes, It's fine for me, thank you for reintroducing this behavior, even though it seems that the shortcuts are different. This doesn't affect the behavior and it works as before. 

I was wondering if it would be possible to allow the cell handle that adds (held) frames  to also remove (held) frames at any time and without prior selection ? Or perhaps to allow frames to be subtracted when the 2nd opposite handle appears ?

Abel032:
@Ahl76 The smart tab also be used to repeat the selection, I added the feature to auto select the same cells in front of selected one cell.
Maybe we need something like holding alt and drag to adjust the duration of current drawing.

Ahl76:
What do you mean by autoselect ? Do you have an example of this feature ? 
Otherwise, will it have a major impact on the repeat selection function, if we're only talking about frame reduction from a cell containing a drawing and not from a selection ?
What I mean is simply that pushing the handle up (in xsheet mode) from the main drawing cell reduces frames.
Instead of doing it in 2 steps with a prior selection, (as is currently the case) , then pull up.
If this make a problem with the repeat selection function, perhaps it should only be active when using a modifier as you suggest. But if this only active from an individual drawing cell, maybe that could works, like it's the case in Tvpaint or Harmony.

Abel032:
I've implement the negative drag for the handle now, it only works when the original selection is single cell.

Ahl76:
@Abel032 Thank you to have implemented this feature. I think it's a good starting point and could be made even more intuitive in the future. Multiple selection with the option of collapsing several columns at once would also be an area for development. Thanks.

Abel032:
Here is a problem of scrolling the frame cell by default: how can we move the image sequence imported?
They are all 1's,so it's hard to select the whole level.

RodneyBaker:
@Abel032 I need to return and test this more.
Just posting to let you know it's still on the radar.

For those willing and able to test... the more feedback the better.  :)

Abel032:
Description updated

RodneyBaker:
I'm posting intent to merge this PR as it adds useful options.
A few overall concerns:

The number of options/variables is considerable when setting up workflow in the xsheet and this adds more to that.  Future work should be to refine documentation of the options and to make sure organization of Preferences is optimized and priority is given to defaults.  What those defaults are might be contentious as everyone has their own workflow preferences.  That's why it seems important to keep these variables as options.

Of note and special interest here is that the changes in this PR do not effect keyframes.
Think in terms of toggling the option to 'Show Keyframes on Cell Area' to be off as optimized testing/workflow and shifting emphasis of work with keyframes to the Timeline and Function Editor/Curve Editor (?) as that might encourage new improvements in that area (especially the Function/Curve Editor which has unique features for working with keyframes found only therein.  For those that desire working with keyframes in the xsheet the option can easily be toggled on.  My point here is simply that when working 'keyframe-less' there is a specific workflow that can be optimized.  As we add keyframes into the mix an entirely new order of complexity is layered into the mix.

TLDR:  There is a dizzying array of options and we need to identify user frustration points and work to resolve those.

As for you (@Abel032) and your continuing effort to improve and optimize Opentoonz... you have our gratitude.

RodneyBaker:
In testing this I encountered a problem with the Appveyor artifact where I couldn't grab exposures in the xsheet on some columns (related to level type and possibly column placement) by the left drag bar.  Testing with the github action builds did not have the problem.

Pressing forward with this to seat in the changes.
Feedback of the resulting nightly will be appreciated.

Ahl76:
I've retested a bit and I think there's another point that should be reviewed regarding the selection of a level when they overlap one another in same column(in the xsheet). Since the changes, it's no longer possible to use the side grip to move an entire level composed of unique drawings. 
This can be disruptive to navigation in certain scenarios. A solution using a modifier key or a quick double-click on grip could be useful to activate old if it's possible.

RodneyBaker:
@Ahl76 Thanks for documenting that shortfall.
@Abel032 is aware of this as we briefly discussed the problem via discord.
As far as I know he's still contemplating how best to approach resolving the problem.

toonz/sources/toonz/xshcellviewer.cpp
```diff
@@ -965,10 +965,23 @@ void RenameCellField::renameCell() {
...
+            xsheet->getCell(range.m_r1+1,range.m_c0);
+            ++range.m_r1);
+    cellSelection->selectCells(range.m_r0, range.m_c0, range.m_r1, range.m_c1);
+    }
     cellSelection->renameCells(cells[0]);
+  }
   else
     cellSelection->renameMultiCells(cells);
+  
+  cellSelection->fillEmptyCell();
```
shun-iwasawa: @Abel032 
Again, sorry for the late comment.
What is the purpose of calling `fillEmptyCell()` here?
Currently, typing `0` into RenameCellField to empty a cell has become impossible due to this `fillEmptyCell()` call.
Removing this line would resolve the issue, but should it be moved to a more appropriate location instead..?

shun-iwasawa: Thank you for letting me know the intention.
Is there a way to leave inbetween frames empty and input cells intermittently as before?
In fact, this isn't a rare case in Japanese animation industry. When keyframe animators do pencil testing, they often place key drawings only on keyframes, leaving the frames where inbetweens are supposed to be placed blank like as follows:

image


.. Hmm, I can only come up with the idea of separating behavior based on preferences.
It might become complicated with too many options, but perhaps you could group the other xsheet operations you've implemented or are trying to achieve (probably based on some common concept similar to implicit hold) into a single option.

Any other good ideas are very welcome. Thanks!

PR #5915, opened 2025-04-14
ci: prevent unnecessary cross-platform builds

andeon:
This PR aims to optimize how the build is triggered.

Updated GitHub Actions workflows to avoid triggering unnecessary builds on platforms where changes are not relevant. This helps save time and resources by ensuring that unaffected platforms do not run the pipeline. 

_For example, it doesn't make sense to build all platforms when only `workflow_linux.yml` is changed._

Replaced the for loop in the DMG creation process with an until loop to improve efficiency. This change may help reduce unnecessary retries and lower overall GitHub Actions quota usage.

It may be necessary to add AppVeyor to the rule to trigger only for macOS and Windows. I'm not sure where and how the release are picked up for the nightly or stable release. It's happens after a successful merge or it's manually?


RodneyBaker:
The cache does not appear to be updating to free up space.
This may be something @shun-iwasawa has to trigger in Settings.

It's happens after a successful merge or it's manually?

It's something of a mystery to me but there is a separate process for building the nightlies and the final releases.  Again, we'll have to refer to @shun-iwasawa for that.
Specifically, when the build system is either down or needs to be restarted that is something only @shun-iwasawa can do.

There are a few things @manongjohn has permissions such as ability to restart PRs but he's hesitant to use that because of his focus on Tahoma2D.

RodneyBaker:
I'm not seeing anything in this PR that will trigger the release of LTS caching but will merge to see if that helps improve the situation.

I am not optimistic that we can get past this without @shun-iwasawa assistance.

PR #5919, opened 2025-04-15
CI - macOS: Update GHA Workflow

xavier2k6:
* Bumped Runner Image to 15

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -23,7 +23,7 @@ jobs:
     timeout-minutes: 90
     # runs-on: macos-latest
     # avoid the headache of arm64 builds?
-    runs-on: macos-13
+    runs-on: macos-15
     steps:
       - uses: actions/checkout@v4
 


andeon:
I'm not sure if it's a good idea to update all the workflows to the latest tools, maybe exceptions should be made for Linux, since it tends to be less problematic or more easy to fix.

I believe they are or were aligned with the configurations of the main developers' machines to maintain consistency with their tools. For example, macOS has historically had build issues with new versions that may require code adjustments.

@RodneyBaker, what's the current criteria?

xavier2k6:
>macOS has historically had build issues with new versions that may require code adjustments.

All the more reason to make these changes due to deprecations etc.

andeon:
Between sticking with older or nearly deprecated versions and always using the latest, choosing something in the middle is a better approach. I believe it better reflects the typical machines and tools most people use.

This is the approach I’ve been applying to Linux, for example. In the past, it was only updated when support was about to be dropped. 

It’s just a suggestion and an observation, but it should be evaluated, especially by the most active developers, since they would, in theory, be the ones most impacted.

xavier2k6:
Using the latest available runners & updated tools is a better approach in "ci" as it's a more "proactive" approach than being "reactive", one of the purposes of "ci" is for testing/stressing & finding problems in the code as well as making sure it compiles correctly with newer tools.

In relation to macOS, If you look for example at the crash reports on `macOS` they're on newer OS's as shown by the kernel version aka Darwin version

It would also be more beneficial for this repo to add a `CodeQL` workflow too.

RodneyBaker:
Will move forward with this although... still much work to do.

Issue #5914, opened 2025-04-14
Strategies to mitigate LFS/Runners quota


andeon:
I'm opening this issue to discuss potential strategies for addressing our current challenges with LFS quota limits. These limitations are beginning to impact our workflow and scalability, so it's crucial that we brainstorm solutions.

I'm not convinced this is an LFS quota issue; it might be related to an Actions Runner quota instead.   

Some points we could explore:

- Avoid working directly in PRs.

- Archive or remove older assets.

- Explore alternative storage solutions.

- Closely check the macOS situation regarding DMG creation process.

- Remove LFS?

For curiosity, I used this command to list the 20 largest files ever versioned in the repository's history:

```
git rev-list --objects --all | \
  git cat-file --batch-check='%(objecttype) %(objectname) %(objectsize) %(rest)' | \
  grep '^blob' | \
  sort -k3 -n -r | \
  head -20 | \
  awk '{print $3, $4}' | numfmt --to=iec
```

**Result:**
```
11,0M  thirdparty/openblas/libopenblas_64.lib  
 8,9M  thirdparty/openblas/libopenblas_32.lib  
 7,5M  toonz/sources/translations_docs/english/fonts/ipaexm.ttf  
 5,9M  thirdparty/openblas/xianyi-OpenBLAS-e6e87a2/lapack-3.4.1.tgz  
 3,2M  thirdparty/tiff-4.0.3/lib/LibTIFF-4.0.3_64.lib  
 3,2M  thirdparty/tiff-4.0.3/lib/LibTIFF-4.0.3.lib  
 3,0M  thirdparty/superlu/SuperLU_4.1/lib/libsuperlu_4.0.lib  
 3,0M  thirdparty/superlu/SuperLU_64.lib  
 2,9M  thirdparty/superlu/SuperLU_32.lib  
 2,9M  thirdparty/LibTIFF/tiff-4.0.3/lib/LibTIFF-4.0.3_64.lib  
 2,9M  thirdparty/LibTIFF/tiff-4.0.3/lib/LibTIFF-4.0.3.lib  
 2,7M  thirdparty/LibJPEG/jpeg-9/prj/LibJPEG/ipch/libjpeg-9e36a988/libjpeg-adbec841.ipch  
 2,5M  thirdparty/LibJPEG/jpeg-9/lib/LibJPEG-9_64.lib  
 2,4M  thirdparty/LibJPEG/jpeg-9/lib/LibJPEG-9.lib  
[...]

PR #5924, opened 2025-04-19
ci: Enhance pre-build check to skip unnecessary builds

andeon:
This PR uses [`skip_commits:`](https://www.appveyor.com/docs/how-to/filtering-commits/), so there's no need to use `before_build:`. This resolves the issue in a much better and more appropriate way.

Thanks to @gep13 for pointing this out!

**References:**
PRs: #5921
Issue: https://github.com/opentoonz/opentoonz/issues/5914

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,4 +1,9 @@
 version: 1.7.1.{build}
+skip_commits:
+  files:
+    - 'doc/**'        # Skip everything inside doc/ (recursive)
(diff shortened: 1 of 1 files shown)


andeon:
Now it should work correctly the way it was intended, avoiding builds if the irrelevant files are changed.

Thanks to @gep13

gep13:
@andeon glad to hear that you got it working the way you wanted it!

PR #5931, opened 2025-04-27
ci: Remove Debug build configuration from AppVeyor

andeon:
Removed Debug from configuration based on discussion to streamline CI, prevent dual builds, and reduce LFS usage. 

The LFS quota should reset next month in about 3 days, so things should normalize a bit more after that


RodneyBaker:
Thanks @andeon

PR #5937, opened 2025-05-01
chore: remove unused MSVC 2013 libraries and unintended files

andeon:
This PR removes all `.lib` files built with Microsoft Visual Studio 2013 (MSVC 2013). These files are no longer required since change in PR https://github.com/opentoonz/opentoonz/pull/3674 when change logic in `toonz/sources/CMakeLists.txt` to use only MSVC 2015.

- Deleted unused MSVC 2013 libs (e.g.,`SuperLU_2013_64.lib`, `libpng16_2013_64.lib`) files from:
  - `thirdparty/tiff-4.0.3/lib/`
  - `thirdparty/superlu/`
  - `thirdparty/zlib-1.2.8/lib/`
  - `thirdparty/libpng-1.6.21/lib/`

We know that OpenToonz dropped 32-bit support a few years ago. It may be possible to extend this cleanup to some 32-bit libs as well, but it’s unclear whether they are still used in certain cases. Anyway, this could be addressed in a future cleanup.

Additionally, this PR removes various unintended files prefixed with `._` (e.g., ._ANNOUNCE, ._CHANGES) from `thirdparty/libpng-1.6.21/`. These are Apple-specific metadata files, likely added accidentally when using non-native file systems.

- Removal of unnecessary files from:
  - `thirdparty/libpng-1.6.21/`


RodneyBaker:
Thank you @andeon

PR #5940, opened 2025-05-05
ci: fix macOS 15 build with SuperLU in GitHub Actions

andeon:
This pull request fixes the macOS build had issues with handling SuperLU (the library was built for x86_64 only) and its dependencies on macOS 15. These changes ensure that the build now runs successfully.

- Updated CMake to support macOS 15 and correctly detect SuperLU
- Enable `-DWITH_SYSTEM_SUPERLU=ON` in GitHub Actions workflow to use system-installed SuperLU via Homebrew
- Adjusted paths for Homebrew libraries to ensure compatibility on macOS 15

I suggest not approving CI changes that are known to be broken, as this could disrupt the **few developers** who are actively writing code and can't be certain the code will run correctly in the target environment. These builds are essential for validating whether the code works properly across all platforms and also allow beta testers to download and test the latest versions.

Although the build completes successfully, I can't confirm that OpenToonz runs as expected, as I currently don't have access to a macOS system for testing. If it doesn't work for any reason, I recommend reverting the workflow again to macOS 13 until a developer with access to a macOS machine can properly address the issue.

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -20,9 +20,7 @@ on:
 
 jobs:
   macOS:
-    timeout-minutes: 90
-    # runs-on: macos-latest
-    # avoid the headache of arm64 builds?
+    timeout-minutes: 90 # Stop job if it exceeds expected build time
     runs-on: macos-15
     steps:
       - uses: actions/checkout@v4
@@ -48,16 +46,14 @@ jobs:
               fi
             done
           }
-          
           brew update
           brew cleanup
-
-          checkPkgAndInstall pkg-config glew lz4 libjpeg libpng lzo boost libusb libmypaint ccache jpeg-turbo ninja opencv qt@5
+          checkPkgAndInstall pkg-config glew lz4 libjpeg libpng lzo boost libusb libmypaint ccache jpeg-turbo ninja opencv superlu qt@5
           brew unlink qt
 
       - name: Set up cache
         run: |
-          export PATH="/usr/local/opt/ccache/libexec:$PATH"
+          export PATH="/opt/homebrew/opt/ccache/libexec:$PATH"
           mkdir -p /Users/runner/.ccache
 
       - uses: actions/cache@v4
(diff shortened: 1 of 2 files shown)


RodneyBaker:
We definitely need a Mac developer to champion Opentoonz.
We are operating largely blind where it comes to Mac PRs.

andeon:
@RodneyBaker  Do you know anyone who could run this to check if it gonna works?

RodneyBaker:
Likely suspects would include @artisteacher and @manongjohn.
@martinvanzijl as well if he is still available.

andeon:
@RodneyBaker I think it's better to approve this change now, as it could cause issues for @Abel032 or other developers when pushing new commits. The code is currently not compiling on macOS, so they can't be sure everything is working until this change is approved.

If needed, after someone tests it further with a later artifact, we can always roll back to the previous macOS 13-based workflow version. But for now, it's important to keep the flow moving.

And I have good news: the AppVeyor team has increased our build timeout to 90 minutes, which should help with builds moving forward.

RodneyBaker:
Thanks @andeon!

artisteacher:
Sorry about being late to the party, but I just tested a recent OT action build and it crashes on open - so early that the icon doesn't even appear/bounce in the dock. I'm not sure if it's related to this, but here's the crash log:
[crash-on-open-mac.txt](https://github.com/user-attachments/files/20258859/crash-on-open-mac.txt)

andeon:
@artisteacher I'm wondering if this might be related to a code signature validation issue, especially with macOS 15 being much stricter about security and app validation. Could you check if the app might have been blocked or flagged by the system, just so we can be sure?

If this artifact isn't available, try the latest PR sent by @Abel032. Because it already uses the macOS 15 workflow. If that still doesn’t work, we must need to roll back to macOS 13 workflow.

Thanks for the feedback

artisteacher:
The only way to open OT or T2D on macOS 15 is to override the security preferences - I did for both the latest PR's action build and this one. Same result - immediate crash. Not an issue with T2D builds, so you might want to check what workflow they are using.

andeon:
Tahoma2D is still using macOS 13 for the workflow build. Let me know if the [latest nightly build](https://github.com/opentoonz/opentoonz/releases/tag/nightly) of OpenToonz works for you, as it uses a similar workflow. 

If it’s positive, I’ll send a PR to revert the workflow to use macOS 13.

andeon:
@artisteacher I thought it would be better to generate the artifact before sending it to the repository. If you could test it, I would appreciate it. 

https://github.com/andeon/opentoonz/actions/runs/15081847125/artifacts/3143679873

If you confirm that it works, I’ll send the PR to fix it right after.

PR #5945, opened 2025-05-10
Enhance Assets Import

Abel032:
- [x] Add auto convert feature when loading resources (Only NAA Raster to TLV, trans to Ink layer)
- [x] AutoRename and auto Convert NAA Raster Images when importing from .xdts
- [x] Make lonely scene's New Level to be saved in $sceneFolder/relativePathSetBySandBox by default
- [x] Refactor sequence renaming
- [x] Refactor: move **LoadResourceArguments>FrameIdsSet** to **LoadResourceArguments>ResourceDatas>m_frameIds**

Note: lonely scene means the .tnz file without scenes.xml in the same folder.

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -247,6 +247,8 @@ public:
 
   // Loading  tab
   int getDefaultImportPolicy() { return getIntValue(importPolicy); }
+  int getDefaultRenamePolicy() { return getIntValue(renamePolicy); }
+  int getDefaultConvertPolicy() { return getIntValue(convertPolicy); }
   bool isAutoExposeEnabled() const { return getBoolValue(autoExposeEnabled); }
   bool isSubsceneFolderEnabled() const {
     return getBoolValue(subsceneFolderEnabled);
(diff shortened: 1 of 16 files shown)


Abel032:
> I had the option enabled, but some NAA unpainted tests failed to convert, but they also fail in the File convert

Would this still happen if you apply Binarize(level command) to these tests?

I agree to uncheck.
But I think current way is the more clear even through user can't adjust all settings of conversion ,it's a process that doesn't need user to config something, which is easier than converting popup.

The main issue of getting assets converted before importing is specify right settings for every level, it introduce more concepts, especially when we can't check the assets (blindly).

Then I would prefer:
- Add a comboBox "**Convert To TLV:**" to choose
    - **Do not Convert**
    - **Convert Every Level with Settings Popup**
    - **Convert Unpainted Aliasing Raster(Inks Only)**
- Add necessary option for the 3rd option: Append Default Palette and DPI settings
- Add some tip text to explain how they work

Abel032:
This PR originally collected multiple asset import related changes.  
For better reviewability, it will be split into several smaller PRs.

## Split plan

- [x] Refactor assets import process & image sequence renaming -> #6084
  - [x] Add NAA FullColor auto-conversion feature -> #6098
    - [x] Add rename checkbox and FullColor Image Convert combo to `xdtsImportPopup` -> #6103
---
- [x] Add fallback file path suggestion to `xdtsImportPopup` -> #6086
- [x] Allow loading `.xdts` files as scenes from Windows Explorer -> #6085

PR #5946, opened 2025-05-12
Fix order of beginInsertRows/beginRemoveRows in ExportSceneDvDirModel

Abel032:
## Issue Description

Crash when trying to export a scene.

Must emit beginRemoveRows before removing datas, and emit endRemoveRows after removing


RodneyBaker:
Thanks you @Abel032 for the fix.

PR #5948, opened 2025-05-13
Stop tracking files with Git LFS

Abel032:
This PR stops tracking certain files with Git LFS due to GitHub's LFS bandwidth limit.
These files(.dll and .lib) have been removed from LFS tracking and are now tracked normally via Git.


Abel032:
I'm not sure about how to skip the LFS in appveyor.

andeon:
Maybe between [the version](version: 1.7.1.{build}): and skip_commits:

version: 1.7.1.{build}

`init:`
 `- git lfs install --skip-smudge`

skip_commits:

I was also working on an alternative to fix this and was about to submit it. It's a little different, do you think I should send it too?

Abel032:
@andeon 
Thanks, just submit yours.

Abel032:
The re-add of dll and lib is a little complex, @andeon you can submit after this PR.

andeon:
I think it's not necessary, it just changes if it's going to merged. ;)

andeon:
To provide some context, I'm afraid that hosting the DLLs and LIBs directly in the repository will significantly increase its size over time. Binary files don’t compress well, and each time a binary changes, Git stores the entire new file, causing the repository to grow quickly. Large repositories become slower to clone or fetch, especially for new contributors or CI systems.

We should avoid this because we're also looking for ways to update some outdated libraries, meaning the repository will continue to grow due to this approach. For this reason, I believe it's better to avoid it and try the release solution.

PR #5949, opened 2025-05-13
Workaround for LFS quota

andeon:
This PR introduces a workaround to bypass the Git LFS issue, ensuring the build can continue without interruption while we implement a more permanent solution.

It does not change the repository structure or remove Git LFS support. It simply disables LFS during the checkout step for Windows and AppVeyor workflows. The LFS-tracked pointer files are then replaced by downloading the exact same precompiled third-party libraries used in the official OpenToonz repository, provided through an external [release](https://github.com/andeon/opentoonz-thirdparty-libs/releases/tag/prebuilt-libs-v1).

_This issue with Git LFS currently affects only Windows builds, where critical third-party binaries (such as .dll and .lib files) are tracked via LFS and are needed during compilation. When the LFS quota is exceeded, these files are no longer downloadable through standard Git operations, breaking the Windows CI._

ref: https://github.com/opentoonz/opentoonz/issues/5914#issuecomment-2877558184

### DIFF
--- a/.github/workflows/workflow_windows.yml
+++ b/.github/workflows/workflow_windows.yml
@@ -33,7 +33,7 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         with:
-          lfs: true
+          lfs: false
       - name: Install and Configure ccache
         run: |
           @echo on
@@ -64,6 +64,13 @@ jobs:
           7z x Qt5.15.2_wintab.zip
           move Qt5.15.2_wintab\5.15.2_wintab thirdparty\qt
         shell: cmd
+
+      - name: Install third-party prebuilt libs (LFS workaround)
+        run: |
+          @echo on
+          curl -fsSL -o thirdparty-libs.zip https://github.com/andeon/opentoonz-thirdparty-libs/releases/download/prebuilt-libs-v1/thirdparty-libs.zip
+          7z x -y thirdparty-libs.zip
+        shell: cmd    
 
       - name: Add msbuild to PATH
         uses: microsoft/setup-msbuild@v2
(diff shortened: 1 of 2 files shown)


andeon:
@Abel032 What do you think? Is this workaround acceptable? If so, feel free to close your PR. 

Otherwise, I can close this one so you can continue with yours, since these are two different ways to resolve the same issue.

Abel032:
Well I think this PR is quite good, through both solutions are not perfect.
So I think, we have better to create a third party resposity including opencv and so on? @shun-iwasawa 
Thanks @andeon , I will close my PR.

Btw, the nightly build have stopped for a month.

andeon:
@Abel032 I've already tried to figure out how this nightly build is triggered, but I still have no idea.

About the thirdparty repository, I'm doing some tests to check if it's possible to compile all the required libraries using GitHub Actions, so we won't need a dedicated developer with a Visual Studio IDE setup to do it locally. I started testing with [zlib 1.2.13](https://github.com/andeon/zlib_custom/actions/runs/15007023981/workflow), to see how it's going to work. This workflow compiles zlib using a custom .sln file, but the final idea is to have a workflow that compiles all the libraries and packages them into a single final release, working similarly to this PR.

However, it's common for one or two libraries to change while the others remain stable. So, we would need to adopt something like an LTS policy, updating the release only in case of a security risk or when a developer specifically needs a newer version for a feature or code dependency. This would be a significant improvement, as it has been 9 years since the libraries were last updated.

RodneyBaker:
Let's press forward then.
Thanks @andeon and @Abel032

PR #5953, opened 2025-05-16
Better level convert/export/visualize/process

Abel032:
### This PR would make these commands be able to get levels from level/cast/cells selection :

- Level-Convert-Convert to Vectors...
- Level-Convert-Convert to Toonz Raster... (was Vectors to Toonz Raster)
- Level-Adjust-Binarize
- Level-Export Level...

### Additional Improvements:

- Now **Convert to Toonz Raster** can convert Normal Raster with options
- Rasterize Vector with Anti Aliasing would apply to Level-Convert-Convert to Toonz Raster, when converting vector levels
- Fix the offset and thickness change when exporting ToonzVector to ToonzRaster
- Disable the Purple Block at middle when Visualize Vector as Raster is on and VectorImage is empty
- Apply show 0 Thickness line preference to Visualize Vector as Raster

### DIFF
--- a/toonz/sources/toonz/CMakeLists.txt
+++ b/toonz/sources/toonz/CMakeLists.txt
@@ -308,6 +308,7 @@ set(SOURCES
     trackerpopup.cpp
     vectorguideddrawingpane.cpp
     vectorizerpopup.cpp
+    rasterizecommand.cpp
     vectorizerswatch.cpp
     versioncontrol.cpp
     versioncontroltimeline.cpp
--- a/toonz/sources/toonz/binarizepopup.cpp
+++ b/toonz/sources/toonz/binarizepopup.cpp
@@ -7,6 +7,7 @@
 #include "menubarcommandids.h"
 #include "cellselection.h"
 #include "filmstripselection.h"
+#include "selectionutils.h"
 
 // TnzQt includes
 #include "toonzqt/intfield.h"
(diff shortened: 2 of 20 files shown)


andeon:
Very nice! I use the 'Export Level' function very frequently to convert Toonz Vector levels to .tlv (Toonz Raster Level) with the 'No Antialias' option enabled, and now the TLV offset issue is finally fixed.

Abel032:
@andeon Now you can just convert them using the level command Convert To Toonz Raster, just diable Rasterize Vector with Anti Aliasing in Preference-Visualization.

konero:
Yet to test but but question, are we able to select multiple levels on the X-sheet and convert with this? 👀

Abel032:
@konero Yes.

RodneyBaker:
I can confirm that selecting multiple levels and converting works.

An odd thing would seem to be that the order of the newly converted levels in the reverse of the source levels order.

Something to consider also is that when selecting multiple levels the preview will be of the first level selected.  I didn't see any way to toggle through selected levels.

This is really good stuff @Abel032

RodneyBaker:
Posting initial intent to merge this PR but want to do a final pass of testing to see if there is anything I've missed.  Especially as I was focused on the conversion update and not all of the other things updated as listed in Additional Improvements.

This is a really nice addition @Abel032 that will save a lot of time.
It is indeed 'better'.

PR #5954, opened 2025-05-17
Export lonely scene

Abel032:
- Support Lonely Scene Export
- Use of $sceneFolder as SandBox's root folder when lonely scene loaded

### DIFF
--- a/toonz/sources/include/toonz/toonzscene.h
+++ b/toonz/sources/include/toonz/toonzscene.h
@@ -259,6 +259,9 @@ If \b scene is in +scenes/name.tnz return name,
 
   // if the path is codable with $scenefolder alias, replace it and return true
   bool codeFilePathWithSceneFolder(TFilePath &path) const;
+  
+  // return false if the scene doesn't belong to any project
+  bool isLonelyScene() const;
 
   bool isLoading() { return m_isLoading; }
   void setIsLoading(bool isLoading) { m_isLoading = isLoading; }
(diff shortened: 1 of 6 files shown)


RodneyBaker:
For those testing or curious, @Abel032 mentions in other PR what a 'lonely scene' is:

```Note: lonely scene means the .tnz file without scenes.xml in the same folder.```

RodneyBaker:
Pressing forward with this.
Thank you @Abel032 for the additional Export option.

Ahl76:
I've traced it back to here by testing the latest NBs. It seems that the merge of this PR is what's preventing the individual saving of levels in the first place. Cf. #6005

Abel032:
thank you @Ahl76 , I'll check that.

Bracket-I-I:
Does this actually fix some problem or is it some sort of feature?

It's not breaking any safeguards, is it? This is not something that shuts down a warning or process that prevents a faulty scene/project setup, is it?

Like "Oh this scene/project is not fully complete, and OT would complain about it, so I'm shutting that complaint down and allow operations on it anyway!"

Abel032:
@Bracket-I-I 
Sometimes you need to pack one scene to other people to finish the remaining work, and you don't have to pack the whole project, then you can use this feature.
The structure of lonely scene is allowed before this PR, but this PR completes the logic and deals with some conflicts(it should be).
You can use Clip Studio Paint to export OpenToonz Scene or load a .xdts file as scene, that would be a scene doesn't belong to any project(or only belong to sandbox).

PR #5955, opened 2025-05-17
Refactor micro logic for ExportAllLevelsPopup

Abel032:
#### This PR changes the behavior of column name macros with '#' prefix when exporting levels.
To generate the suggest exported file name:
- If colname starts with #, all # characters are removed.
- If the cleaned colname contains an underscore _, return the cleaned colname as a wide string.
- Otherwise, return the original levelname unchanged.


RodneyBaker:
Posting intent to merge as this LGTM...

PR #5960, opened 2025-05-20
Revert macOS version from 15 to 13 due to issues

andeon:
Since macOS 15 runs on ARM64 by default, while macOS 13 used x86_64, this has caused build issues due to architecture-specific dependency libraries. Therefore, this change reverts the CI to use macOS 13 until a developer with access to macOS can evaluate, test, and resolve the issue before moving to the any newer version.

A DMG file on macOS is not inherently a portable app, but it can be used to distribute it in a way that is somewhat similar to a portable app. Since we are experiencing delays with the nightly releases, I’d like to suggest that @RodneyBaker also include the .dmg file along with the Windows portable version release. There's no need to create a new release for this, just remove the current attachment and update it with the latest Windows portable build and the latest .dmg file. Repeat this maybe once a month or when an important fix is available, to avoid overloading you, since manually handling this can be a bit tedious.

I was waiting for @artisteacher to validate this, but I just remember that it was working before the workflow was changed to macOS 15, so I figured I'd submit it asap.

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -21,7 +21,8 @@ on:
 jobs:
   macOS:
     timeout-minutes: 90 # Stop job if it exceeds expected build time
-    runs-on: macos-15
+    # Stick with macOS 13 to avoid arm64 headaches on macOS 15
+    runs-on: macos-13
     steps:
       - uses: actions/checkout@v4
 
@@ -46,14 +47,16 @@ jobs:
               fi
             done
           }
+          
           brew update
           brew cleanup
-          checkPkgAndInstall pkg-config glew lz4 libjpeg libpng lzo boost libusb libmypaint ccache jpeg-turbo ninja opencv superlu qt@5
+
+          checkPkgAndInstall pkg-config glew lz4 libjpeg libpng lzo boost libusb libmypaint ccache jpeg-turbo ninja opencv qt@5
           brew unlink qt
 
       - name: Set up cache
         run: |
-          export PATH="/opt/homebrew/opt/ccache/libexec:$PATH"
+          export PATH="/usr/local/opt/ccache/libexec:$PATH"
           mkdir -p /Users/runner/.ccache
 
       - uses: actions/cache@v4
@@ -70,7 +73,7 @@ jobs:
 
       - name: Build
         run: |
-          export PKG_CONFIG_PATH="$(brew --prefix jpeg-turbo)/lib/pkgconfig:$PKG_CONFIG_PATH"
+          export PKG_CONFIG_PATH="/usr/local/opt/jpeg-turbo/lib/pkgconfig:$PKG_CONFIG_PATH"
           cd toonz
           mkdir -p build
           cd build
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@andeon Mostly unrelated to this PR...

We need to look at why the github builds are referring to your third party resources and not those of the Opentoonz repository (or @shun-iwasawa's) as this creates a concern for parties requiring strict control of sources used in building Opentoonz.

It's fine (in the short term) if this sourcing is required to overcome the problems we've faced with CI builds but we need to make sure we publish that information as well as plan to resolve any conflicts.

In other words, if the third party sources are not sufficient via the main repository then we need to work to get them updated.  

If the referencing of sources is just due to how the automated builds were handled at the time of updating the build instructions and unintentionally refer to private repos then we should repair those.

Regarding this PR...
Agreed that we should the MacOS verison from 15 to 13 as you've requested.
I'll also plan to add the MacOS dmg to the portable listing although there are some concerns about the portable being listed in the first place and not generated programatically.  For the record the reason the portable is listed is due to failure of the nightlies being generated which makes it impossible to use merged code especially once the CI build storage schedule no longer allows that build to be retrieved.  In other words, I've had a request to remove the Portable listing but do not plan to remove that unless/until the nighty builds are avialable again.  Having a portable available is still a good idea but the reason for the portable's current availability purely relates to the ongoing and persistent failure of nightly builds.  I do need to update that portable (Windows) build which I will plan to do along with providing the MacOS DMG.

As always, thanks for all you do to support Opentoonz.

andeon:
@RodneyBaker I agree with everything. Just to clarify, the thirdparty resources are literally the same binaries files as in the repository. I only copied the directory structure and files and zipped them for the release. I'm not sure if you have access to create a repository under OpenToonz. If you do, you could create it the same way, and we can just adjust the script to fetch from there. Or maybe @shun-iwasawa could do the same in his repo without any problems.

Right now, this is the best way to avoid LFS issues and other problems that come from storing binaries directly in the repo during development. Technically, it doesn't affect local compilation in its current state, since it's only used for building in the GitHub workflow.

andeon:
Just to register here: After Shun adjusted the 32-bit build, the nightly builds seem to be working well again.

However, I noticed that these builds come from an external Jenkins? So, the .exe and .pkg installers for OpenToonz apparently are built there, meaning the AppVeyor and GitHub builds are mainly used for development and testing, since they don’t generate installable versions like Jenkins does.

That said, the .pkg file still seems to be dated March 19, which might be related to this https://docs.travis-ci.com/user/reference/osx/

RodneyBaker:
@andeon 
Just a note to let you know the current 'portable' listing in releases contains both Windows and Mac builds.  They are the same thing as would be downloaded via github action builds so the main plus is that people don't have to look for those.

The release date on the entry itself still says 20 Feb so that might confuse some.
Link:  https://github.com/opentoonz/opentoonz/releases/tag/OTv1.7.1.0-portable

We'll have to look at how we might plus that 'portable' release up a bit more but the builds are there in case someone feels inclined to grab them.

PR #5799, opened 2025-02-24
Cleanup: Replace deprecated endl with Qt::endl and fix operator prece…

andeon:
…dence

- This change addresses deprecation warnings introduced in Qt 5.15 and ensures compatibility with future versions of Qt.

```
/home/runner/work/opentoonz/opentoonz/toonz/sources/common/tsystem/tsystempd.cpp:614:31: warning: ‘QTextStream& QTextStreamFunctions::endl(QTextStream&)’ is deprecated: Use Qt::endl [-Wdeprecated-declarations]
  614 |   stream << "[Trash Info]" << endl;
      |                               ^~~~
```
- Fix totalFree calculation to correctly handle memory values by adjusting operator precedence.

PR #5966, opened 2025-05-25
Refactor: Update deprecated Qt functions

andeon:
This PR updates the codebase to replace deprecated Qt functions and patterns with modern alternatives, maintaining existing behavior. These changes improve compatibility with newer Qt versions and eliminate compiler warnings (such as C4996), helping ensure future maintainability.

Notable updates include:

- Replaced deprecated `QTime` with `QElapsedTimer` for timing operations
- Updated `QMatrix` to `QTransform` for transformation handling
- Changed `MidButton` to `MiddleButton` for mouse button naming
- Migrated `QString::SkipEmptyParts` to `Qt::SkipEmptyParts`
- Replaced `QDesktopWidget` with `QScreen` for screen handling
- Updated `fontMetrics().width()` to `horizontalAdvance()` for text metrics

In addition to the items listed above, many other deprecated functions and patterns were updated throughout the codebase.

ref:  #5799, #5801

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -26,16 +26,19 @@ jobs:
     steps:
       - uses: actions/checkout@v4
 
-      - name: Remove symlinks
+      - name: Clean Python symlinks
         run: |
-          # Remove existing symlinks before installing Python
-          for symlink in 2to3 idle3 pydoc3 python3 python3-config \
-                         2to3-3.11 idle3.11 pydoc3.11 python3.11 python3.11-config \
-                         2to3-3.12 idle3.12 pydoc3.12 python3.12 python3.12-config \
-                         idle3.13 pydoc3.13 python3.13 python3.13-config; do
-            rm -f /usr/local/bin/$symlink
+          # Remove potentially conflicting Python symlinks
+          find /usr/local/bin -type l \( -name "python*" -o -name "pip*" -o -name "2to3*" -o -name "idle*" -o -name "pydoc*" \) \
+            -exec sh -c 'file -b {} | grep -q "symbolic link to .*Python.framework" && echo "Removing {}" && rm {}' \;
+          # Ensure Homebrew Python is properly linked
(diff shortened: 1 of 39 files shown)


andeon:
@RodneyBaker If possible, please prioritize this update, as it updates many functions across several files that are likely in scope for future refactoring or new feature additions. Additionally, I updated the macOS workflow in this PR to avoid opening a separate one, as the Python step was failing due to very recent Homebrew updates.

RodneyBaker:
Consider it done.
Thanks @andeon

Abel032:
@andeon 
It seems that this PR introduced an unexpected behavior in the Style Editor.
Could you take a look and check what’s going on?
Here is the recorded video:

https://github.com/user-attachments/assets/7dba0616-bb03-45dc-ad2b-34f43ab6e606


I also noticed a bug that would cause unexpected scrolling when there are two xsheet panels opened:

https://github.com/user-attachments/assets/d8d7afcd-9972-4a54-a412-b912f41da25c

andeon:
@Abel032 I got the bug in the Style Editor, but I still don’t understand the part about the two Xsheet panels even after watching the video, can you explain me how can I reproduce this.

Abel032:
@andeon 
It may not a bug?
Open two xsheet panels, click cell to move the indicator (15'&13' in this instance, out of left xsheet's show range), the left xsheet should scroll but why the right also scrolling?

andeon:
Now I get what you mean, but hasn’t it always been like this? It seems like the instance not really independent.

Abel032:
I think there ayno need to scroll the right one.
Not a big problem.

Abel032:
@andeon 

Thank you for the effects to refactor old code which benefit long-term maintenance and development.

@RodneyBaker found a crash when trying to use **Adjust Current Palette to This Level** of a studio palette.
图片

I've checked the call stack, it seems related to this line in dvdialog.cpp :

`
    QRect screen = QGuiApplication::screens().at(m_currentScreen)->availableGeometry();
`

Could you investigate this?

Issue #5876, opened 2025-03-23
Fill Check Refresh failure without adjacent Assistants level


RodneyBaker:
### Description

There is apparently a refresh bug with Fill Check that relates to Assistants.
Seems narrowed down to:

https://github.com/opentoonz/opentoonz/pull/5039

### Steps to Reproduce

- Draw with Fill Check on
- Note the artifacting
- Create an Assistants level and add a gizmo (need the gizmo)
- Draw with the Assistants tool on or off but with the Assistant level exposed next to a toonz raster level
- Draw again with the Assistants level not exposed next to the toonz raster level
- Note how the presence of the Assistants level effects whether the artifacts appear or not.

See attached video.

https://github.com/user-attachments/assets/fdeba2ee-0013-4cd4-ae2d-6b50fa4d5468

cc:  @blackwarthog 

### OpenToonz Version

Nightly (Specify below)

### OpenToonz Version Information

All builds since Sep 2023

### Operating System

Windows

### GPU

Other

### Graphics Tablet

None

andeon:
I'm testing here, and it's still happening even without adding an Assistants level, just by enabling the fill check. It seems to be related to the mouse cursor release or possibly a refresh issue, as you mentioned. When the cursor moves outside, it fills automatically, or it could be triggered by pressing keys like **Space** or **S**.

RodneyBaker:
The artifacts actually STOP when the assistants level is present.
The artifacting happens by default.

I probably approached this bug report wrong.
I'm not trying to suggest adding Assistants causes the artifacting but rather that the PR introducing Assistants for Raster Levels is the first nightly where the artifacting started (as reported by @RattusaurusRex)

Interesting observation on the Locator.
I must assume the Locator sufficiently triggers a refresh of the canvas as it has to share that same image across both views.

RodneyBaker:
This should now be resolved in nightly releases.

PR #5969, opened 2025-05-29
Fix Fill Check refresh bug introduced in #5039

andeon:
This PR addresses the issue reported in [#5876: Fill Check Refresh failure without adjacent Assistants level](https://github.com/opentoonz/opentoonz/issues/5876).

It reintroduces the `eGap` (Fill Check) and `eAutoclose` checks with appropriate calls to `invalidate()` to ensure proper refresh of Fill Check rendering. 

The bug was introduced in PR #5039, where these checks were removed. It's unclear whether the original developer had a different refactoring plan in mind, but the current state results in rendering artifacts. This fix restores expected behavior for now, though the approach may need to be revisited in the future.

This commit also fixes -Wcomment warnings caused by nested block comments in the following files and lines :

- twain/ttwain_winM.c: lines 115-125
- toonzlib/tframehandle.cpp: lines 20-30

These warnings are specific to the Clang compiler due to its stricter comment parsing.

### DIFF
--- a/toonz/sources/common/twain/ttwain_winM.c
+++ b/toonz/sources/common/twain/ttwain_winM.c
@@ -106,7 +106,7 @@ void setupUI(void) {
   SetFrontProcess(&psn);
 #ifndef HAVE_DOCK_TILE
 /* We end up with the ugly console dock icon; let's override it */
-/*char *iconfile = "/tmp/image.png";
+/* char *iconfile = "/tmp/image.png";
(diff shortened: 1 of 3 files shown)


RodneyBaker:
I confirm this fixes the Fill Check refresh bug.
Much appreciated @andeon 

I will delay merging this for a little while as users can gain this fix via artifacts but won't be able to via the nightlies as they aren't yet building.

PR #5974, opened 2025-06-03
Fix for 32bit build

shun-iwasawa:
This PR makes several fixes to get the 32-bit version to build successfully. (Recent Jenkins failures were due to the 32-bit version not being able to build)

I also fixed the .clang-format configuration file to work correctly. `c++17` seems to be needed to be in lowercase.


RodneyBaker:
Thank you very much @shun-iwasawa 
Much appreciated!

PR #5976, opened 2025-06-05
Fix color wheel pointer and replace layout margins

andeon:
This PR fixes an issue in the Style Editor where the color wheel pointer was not positioned correctly. The problem was reported by _@Abel032 in https://github.com/opentoonz/opentoonz/issues/5966#issuecomment-2943138189._ **Thanks, Abel!** 

https://github.com/user-attachments/assets/7dba0616-bb03-45dc-ad2b-34f43ab6e606

- Corrects the theta angle calculation for accurate color wheel pointer positioning
- Replaces deprecated setMargin() calls with setContentsMargins() for layouts (_not directly related to the bug_)


RodneyBaker:
Thank you @andeon for the quick turnaround fix!

PR #5977, opened 2025-06-06
Fix: Vector Selection Tool doesn't apply fixed line size across multiple frames

Abel032:
Change Fixed Vector Line Size when selected more than one frame, old behavior would call changeImageThickness but not setImageThickness


RodneyBaker:
Thank you @Abel032 for the quickj fix!

PR #5978, opened 2025-06-07
Fix dialog crash in Studio Palette by initializing m_currentScreen

andeon:
This PR fixes a crash that occurred when using "Adjust Current Palette to This Level" from the StudioPaletteTreeViewer context menu. The issue was found by @RodneyBaker and reported by _@Abel032 in https://github.com/opentoonz/opentoonz/issues/5966#issuecomment-2949960568_. Thanks, folks!

图片 

- Initialize m_currentScreen to 0 in the Dialog constructor to prevent segmentation faults when accessing QGuiApplication::screens().at().
 - Replaces deprecated setMargin() calls with setContentsMargins() for layouts (not directly related to the bug)

This logic may need to be refactored in the future to better support multi-monitor setups, as suggested by comments already present in the source code


RodneyBaker:
Thanks @andeon for the fix!

PR #5979, opened 2025-06-08
Add "Fill Holes" feature for Toonz Raster Level

Abel032:
This feature would add a command to fill the holes for ToonzRaster.

## Algorithm
1. 

图片

图片

### Before:
Before
### After:
After

### DIFF
--- a/stuff/profiles/layouts/rooms/Default/menubar_template.xml
+++ b/stuff/profiles/layouts/rooms/Default/menubar_template.xml
@@ -132,6 +132,7 @@
             <command>MI_LinesFade</command>
         </menu>
         <menu title="Optimize">
+            <command>MI_FillHoles</command>
             <command>MI_RemoveEndpoints</command>
             <command>MI_ConvertVectorToVector</command>
         </menu>
--- a/toonz/sources/include/toonz/fill.h
+++ b/toonz/sources/include/toonz/fill.h
@@ -16,6 +16,14 @@ class TPalette;
 #endif
 
 #include <set>
+#include "ttilesaver.h"
+#include "timage.h"
+#include "tpalette.h"
+#include "tgeometry.h"
+#include "tpixelcm.h"
+#include "traster.h"
+#include "trastercm.h"
+#include "tpalette.h"
 
 class FillParameters {
 public:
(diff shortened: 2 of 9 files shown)


Abel032:
There are something cherry-pick problem, would solve it later.

Bracket-I-I:
How easy is this to troubleshoot if something goes wrong?

Let's say the automatic hole filler picks a wrong fill color for a few pixels, and now everything is filled, making it potentially more difficult to spot the problematic areas early 
(for example those few pixels in an aliased version of the image might bleed out in unintended way during composition with smoothing)

How easy is it to know retroactively which holes it filled? Any way to log the information? Because if I have to manually check if all paints were applied correctly to tiny holes that were automatically filled, then I'm doing the same work as I would have done before, scanning with my eyes for tiny holes.
Only at that point they would be unfilled and show up on test modes, like 'show fills' and the like.

Know what I'm getting at?
It's not guaranteed that a tiny hole is insignificant when maximum fidelity is desired in the final output. So... it would be good to have some sort of information where holes were filled automatically. 
Either in log form, or maybe as an image that is empty except with the filled spots.

In fact, maybe thats how the tool could work? Generate an image in the level, or a whole new level, that the user can then merge onto the original. This would probably be best since levels can be color tinted and all that.

Abel032:
@Bracket-I-I 
It's better if you can check by eyes, and don't have to fill every hole by hand.
I think you don't have to make sure every hole be filled correctly because these are moving pictures, just check every drawing for about 5s, and that would be enough for most situation I think.

And this function is designed for aliasing picture, I didn't test on anti-aliasing pictures.

Bracket-I-I:
I understand your viewpoint from not caring much because it's "moving pictures" but it really is a big deal...

Cleanup departments were (and often still are) very big departments because the final pictures had to be, well, clean.
Opentoonz also communicates the importance a bit by having so many pints and hints and modes to detect gaps, transparencies and all that.

It shows red lines for gaps that can be closed, it has several special viewing mode that show the inks, the paints, transparencies etc. 

Anyway, check out the credits for the Lion King from 97
The cleanup crew vastly outnumbers the animation crew lol. It's ...a little bit important. It's just not very glorified.
https://www.youtube.com/watch?v=VHWQeBrlwHY

RodneyBaker:
Pressing forward with this PR as it adds a new option without effecting any old process.

PR #5980, opened 2025-06-09
Prevent OS/theme background mismatch

andeon:
The PR Added `alternate-background-color` to `DVGui--Separator` to ensure consistent appearance across different OS and user themes. Without this, the background color is inherited from the system, which can lead to poor contrast or unreadable text.

## Preview of Light Theme on Windows 10 (Using Default System Theme)
export_level

I found this forgotten hack in an old Toonz theme, so it seems like a legacy workaround to avoid the issue. Maybe @konero could investigate this anomaly. That said, I placed the fix inside `_misc.less` because there’s a section for Unknowns + Legacy.

This issue occurs in **Export Level**, **Convert to Vectors**, and **Canvas Size**, in the text areas where this separator is present. Previously, the fix #4030 applied `alternate-background-color` to `QFrame`, and although it caused no functional issues, it didn’t feel right since it was applied globally.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -2981,6 +2981,9 @@ QStatusBar #StatusBarLabel {
   background-color: #ffffff;
   padding: 1 3;
 }
+DVGui--Separator {
+  alternate-background-color: #dfdfdf;
+}
 /* -----------------------------------------------------------------------------
    XSheet Viewer
 ----------------------------------------------------------------------------- */
--- a/stuff/config/qss/Clay/Clay.qss
+++ b/stuff/config/qss/Clay/Clay.qss
@@ -2981,6 +2981,9 @@ QStatusBar #StatusBarLabel {
   background-color: #ffffff;
   padding: 1 3;
 }
+DVGui--Separator {
+  alternate-background-color: #dfdfdf;
+}
 /* -----------------------------------------------------------------------------
    XSheet Viewer
 ----------------------------------------------------------------------------- */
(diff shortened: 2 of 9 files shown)


RodneyBaker:
Thanks @andeon 
I agree that @konero is the likely person to investigate and move this and related concerns toward broader resolution.

As for this PR, posting intent to merge but with no need to rush.
If someone is facing this mismatch routinely and would like to see it move forward immediately please let us know.  We love testing and feedback!

Abel032:
图片

Can this PR also solve this? The ruler Tool

andeon:
Nope, this only fixes the DVGui::Separator issue! 

But this issue of ruler tool apparently comes from [tooloptions.cpp](https://github.com/opentoonz/opentoonz/blob/67d0275cfcdab93490df9cbd02497911031bf974/toonz/sources/tnztools/tooloptions.cpp#L2229), where a color of background is hardcoded directly in C++. It usually doesn't match dark themes, since the text is white (or nearly white) making it difficult to read.

I was able to change the text color for e.g. black. You can quickly test it by going to `Preferences > Interface > Edit Additional Style Sheet` and adding: 

```
#ToolOptions #RulerToolOptionValues  {
  color: #000;  }
```
So maybe, to properly fix (text) this when generating a theme, we might need to update  **_viewer.less** like so:

```
#ToolOptions #RulerToolOptionValues {
  color: @viewer-ruler-tool-options-text-color;
}
```
However, the C++ code in tooloptions.cpp sets a hardcoded style using setStyleSheet, which overrides any external QSS attempting to change the background of the ruler tool's input fields. For the theme system to work correctly, this inline style would need to be removed or refactored to allow external stylesheets to apply properly.

RodneyBaker:
LGTM

PR #5981, opened 2025-06-09
Fix alpha blending in overPixT for edge cases (zero-alpha backgrounds)

Abel032:
### Issue Description
The stroke's background of exported Image do not follow the background setting.
This happens when exporting ToonzRaster Level to Raster Level using the "Export Level..." command.

Before:
图片
After:


Abel032:
I think there is somthing need to consider.

Abel032:
Wait I think this is not completed.

Abel032:
move to #5982

PR #5982, opened 2025-06-09
Fix: Premultiply background color for Export Level Commands

Abel032:
## Issue Description
The stroke's background of exported Image do not follow the background setting.
This happens when exporting ToonzRaster Level to Raster Level using the "Export Level..." command.
The problem is caused by not do premultiply for the background color is before paint it as background.

### Before:
图片
### After:
图片


RodneyBaker:

PR #5988, opened 2025-06-13
Add inverted cursor feature at stroke end point

Abel032:
Introduce an option to replace the cursor with a inverted cursor at Drawing Stroke's End Point for Vector and Toonz Raster Levels.
图片

The original behavior is to always show the system cursor instantly, which would generate a visual delay when drawing fast.

The new behavior would hide the system cursor, and always show the cursor at the brush position:
BRUSHPOS

When the brush size is bigger than 8, the cursor would be replaced by brush circles.

### DIFF
--- a/toonz/sources/common/tgl/tgl.cpp
+++ b/toonz/sources/common/tgl/tgl.cpp
@@ -160,6 +160,32 @@ void tglDrawCircle(const TPointD &center, double radius) {
   glPopMatrix();
 }
 
+void tglDrawInvertCursor(const TPointD& center,int gap, int len) {
+    TPointD up1 = center + TPointD(0.0, gap);
+    TPointD up2 = center + TPointD(0.0, gap + len);
+    TPointD down1 = center + TPointD(0.0, -gap);
+    TPointD down2 = center + TPointD(0.0, -gap - len);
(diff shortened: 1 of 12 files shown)


RodneyBaker:
LGTM
Initially I was wondering why drawing with vector levels felt very different but then I noticed the Accuracy slider was different from where I usually have it (all the way up).  Once set everything all good.

Thanks @Abel032

Bracket-I-I:
Inverted cursor is extremely non descriptive and confusing...also what does this even do? o_O
It does not fix the delay it just displaces the mouse visually, leaving it unclear where the target is for when the delay is gone.

If I start a line at some point and then draw a curve to an end point then I want my cursor there... not hidden and put where it's not ...but I suppose I can't complain too much since it's optional.

I just don't see the point of this. Especially if it's set to be the default behavior (in the future).

Abel032:
@Bracket-I-I 
Some people would be used to it.
It's like cheating, but you have choice.
Maybe you can take a look of why it's not fast when we drawing with no smooth(especially when using a mypaint style).

Abel032:
> The term "Inverted Cursor" is extremely non descriptive

I'm not very good at naming things, and I would be appreciate if you have a better name

PR #6015, opened 2025-07-17
Feat: Add build date to window title

Abel032:
Add build date to window title, it's optional and can be turn on in the about page.
图片


RodneyBaker:
Thanks @Abel032 
Much appreciated.

Aside:  We probably should turn this option on by default as that would help troubleshoot problems when screenshots are shared.

PR #6018, opened 2025-07-18
Fix undo of autocreation in empty cell

Abel032:
## Issue Description
Select an empty cell and draw with the pen on a raster or vector level column. Then undo — the newly created frame should be undone.


RodneyBaker:
It's interesting that Undo works fine with toonz raster but not with raster and vector.
Thanks for this (and the many other fixes you have implemented).

RodneyBaker:
Thanks @Abel032

PR #6019, opened 2025-07-18
fix: Duplicate Save of saved scene

Abel032:
## Issue Description
Current nightly build would ask for saving again after opening a saved scene and saving all.

---

The save path of resources should not be decided by changing scene's property, it's a little dirty.
I think we need to change this part of code in the future, to separate the resources saving and scene saving.

### DIFF
--- a/toonz/sources/include/toonz/toonzscene.h
+++ b/toonz/sources/include/toonz/toonzscene.h
@@ -56,7 +56,8 @@ public:
   std::shared_ptr<TProject> getProject() const;  //!< Returns a pointer to the project holding a scene instance.
   void setProject(std::shared_ptr<TProject>);  //!< Associates a scene to a project.
 
-  void setScenePath(const TFilePath &p);  //!< Sets the scene's file path.
+  void setScenePath(const TFilePath &p, 
+      bool changeToTitled = true);  //!< Sets the scene's file path.
   TFilePath getScenePath() const {
     return m_scenePath;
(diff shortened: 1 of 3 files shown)


RodneyBaker:
I'm not able to reproduce the original problem this PR attempts to address but am likely not following the correct steps to encounter the problem.

Abel032:
@RodneyBaker You can try following steps:

- New Scene and Save All (save the new scene)
- close OpenToonz
- open OpenToonz and open that scene
- Try to Save All

Update: 

Or more simple, just open a standalone scene, not in the project's scene folder, then try to save all.

RodneyBaker:
Following your steps (both cases) I still don't encounter the problem.
Is an extra dialogue asking the user to save suppose to appear after following those steps?

What I get is presumably just the scene getting saved again even though nothing has changed (with no dialogue appearing)

I'm using:
image

I'll try a few other options again to include making sure I'm using the Sandbox.
Edit:  No problem encountered there either.

Abel032:
> Is an extra dialogue asking the user to save suppose to appear after following those steps?

Yeah..
I think it only appear in this version:
[2025-07-18 12:31:24 UTC](https://github.com/opentoonz/opentoonz_nightlies/releases/tag/nightly_bb75bea257bfd4654904aae670e6532d72c3efb0)

RodneyBaker:
That's it.  Confirmed the problem is with that update/release.
Will merge this to fix.

Issue #6031, opened 2025-07-27
Maintenance Notice: Jenkins machines will be down for a while.


shun-iwasawa:
Dear community members, 

Thank you for your continued contributions to OpenToonz.
I will be temporarily sending the machine currently used for Jenkins CI to Dwango for security updates.
During this time, Nightly Builds updates and Pull Request builds will be disabled.
I will post progress updates on this thread. 
Sorry for the inconvenience, and thank you for your patience.

Shun Iwasawa
OpenToonz team

JParks123:
Mr Shun Iwasawa can you please look into this issue to see if this is a fixable bug?

https://github.com/opentoonz/opentoonz/issues/5408

shun-iwasawa:
Now the Jenkins machine is back to normal. Thank you for your cooperation!

PR #6034, opened 2025-07-31
Update bug_report.yml

konero:
Reduce the 'blame' tone on the user in the dialogue.


RodneyBaker:
Merging this should not be effected by the notice posted in #6031.

Thanks @konero for the update.

PR #6040, opened 2025-08-06
add naru lazy brush fx

Appule:
# OVERVIEW
This PR adds a new effect called Naru LazyBrush.
It extracts areas enclosed by black lines and generatees a mask.
This effect is useful, for example, for cutting out characters during a line test.

It is based on the paper "LazyBrush: Flexible Painting Tool for Hand-drawn Cartoons" by Daniel Sýkora et al., EUROGRAPHICS 2009.

# How to use
Naru LazyBrush Fx generates a mask from the source image.
By default, mask markers are placed automatically, as shown in the images below:

<table align="center">
  <tr>
    <td align="center">
      <br />
      <span>Source</span>
    </td>
    <td align="center">
      <br />
      <span>Result</span>
    </td>
  </tr>
</table>

The effect also fills small gaps in the line art to some extent:

<table align="center">
  <tr>
    <td align="center">
      rabbit<br />
      <span>Source</span>
    </td>
    <td align="center">
      rabbit_color<br />
      <span>Result</span>
    </td>
  </tr>
</table>

You can manually place mask markers using red, and alpha markers using blue in a reference image, like this:

<table align="center">
  <tr>
    <td align="center">
      rabbit_ref<br />
      <span>Source + Reference</span>
    </td>
    <td align="center">
      rabbit_ref_color<br />
      <span>Result</span>
    </td>
  </tr>
</table>

# Parameters
- **Mode**: 
  - **Mask + Line (default)**: Generates a mask and overlays the source image.
  - **Mask**: Generates the mask only.
  - **LoG Filter**: Shows the result of the Laplacian of Gaussian filter.
  - **Capacity Map**: Displays the node capacity map used in the graph cut algorithm.
  - **Scribble Map**: Displays the map of scribbles (mask/alpha markers).

- **Mask Color**: The color used for the generated mask.
- **Fill Hole**: Controls whether fully enclosed spaces within other enclosed areas are marked as mask or alpha.
- **Auto Mask Scribble**: Automatically generates mask scribbles.
- **Line's Weights**: Enter the approximate thickness of the line art strokes in pixels.
- **Background Direction**: Decide on the general direction of the areas you want to make transparent.

**-- Advanced Parameters --**
- **Scribble Type**: Sets the type of scribble. Options: soft, hard.
- **Outline Cutoff**: Threshold for ignoring faint outlines based on brightness.
- **LoG Filter Scale**: Controls the scale of the Laplacian of Gaussian filter.
- **Filter Smoothness**: Determines how strongly ink influences the graph cut. (Lower values make it more sensitive to faint or blurry lines.)
- **Difficulty Passing Through Ink**: Adjusts how resistant ink is to being crossed by the segmentation algorithm. 
- **Scribble Strength**: Controls how strongly the scribbles influence the mask. (Lower values will correct mistakes more aggressively.)
- **Auto Scribble Margin**: Sets the margin distance from the image border where auto-scribbles will be generated.
- **Auto Scribble Outline Threshold**: Brightness threshold that triggers auto-scribbling.

# License
This Fx is based on _The Boykov-Kolmogorov Maxflow Algorithm_ from [https://github.com/Gnimuc/BKMaxflow.jl](https://github.com/Gnimuc/BKMaxflow.jl) which is published under MIT License.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1723,5 +1723,24 @@
   <item>"STD_iwa_TiledParticlesFx.fadeout_color_fade"		"Fade-out Intensity"		</item>
   <item>"STD_iwa_TiledParticlesFx.source_gradation"		"Use Control Image Gradation"		</item>
   <item>"STD_iwa_TiledParticlesFx.pick_color_for_every_frame"	"Pick Control Image's Color for Every Frame"	</item>
+
+  <!------------------------------ Naru FXs ------------------------------------------->
+  
+  <item>"STD_naru_LazyBrushFx" 	"LazyBrush Naru" 		</item>
+  <item>"STD_naru_LazyBrushFx.mode" 		"Mode" 			</item>
+  <item>"STD_naru_LazyBrushFx.mask_color" 	"Mask Color" 	</item>
+  <item>"STD_naru_LazyBrushFx.undef_is_sink" 	"Fill Hole"	</item>
+  <item>"STD_naru_LazyBrushFx.auto_scribble" 	"Auto Mask Scribble"	</item>
+  <item>"STD_naru_LazyBrushFx.line_weight"	 	"Line's Weights"		</item>
+  <item>"STD_naru_LazyBrushFx.sink_pos" 	"Background Direction"	</item>
+  <item>"STD_naru_LazyBrushFx.lambda"	 	"Scribble Strength"		</item>
+  <item>"STD_naru_LazyBrushFx.scr_type" 	"Scribble Type"		</item>
+  <item>"STD_naru_LazyBrushFx.min_lightness" 	"Outline Cutoff"	</item>
+  <item>"STD_naru_LazyBrushFx.log_scale" 	"LoG Filter Scale"		</item>
+  <item>"STD_naru_LazyBrushFx.sigma"	 	"Filter Smoothness"			</item>
+  <item>"STD_naru_LazyBrushFx.alpha"	 	"Difficulty Passing Through Ink"			</item>
+  <item>"STD_naru_LazyBrushFx.auto_scribble_length" 	"Auto Scribble Margin"	</item>
+  <item>"STD_naru_LazyBrushFx.auto_scribble_threshold" 	"Auto Scribble Outline Threshold"	</item>
+
 </stringtable> 
 
--- a/stuff/doc/LICENSE/LICENSE_BKMaxflow.txt
+++ b/stuff/doc/LICENSE/LICENSE_BKMaxflow.txt
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2018 Yupei Qi
+
+Permission is hereby granted, free of charge, to any person obtaining a copy
+of this software and associated documentation files (the "Software"), to deal
+in the Software without restriction, including without limitation the rights
+to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+copies of the Software, and to permit persons to whom the Software is
+furnished to do so, subject to the following conditions:
+
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
--- a/stuff/profiles/layouts/fxs/STD_naru_LazyBrushFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_naru_LazyBrushFx.xml
@@ -1,0 +1,21 @@
+<fxlayout>
+  <page name="LazyBrush Config">
+    <control>mode</control>
+    <control>mask_color</control>
+    <control>undef_is_sink</control>
+    <control>auto_scribble</control>
+    <control>line_weight</control>
+    <control>sink_pos</control>
+
+   	<vbox shrink="0" label="Advanced">
+      <control>scr_type</control>
+      <control>min_lightness</control>
+      <control>log_scale</control>
+      <control>sigma</control>
+      <control>alpha</control>
+      <control>lambda</control>
+      <control>auto_scribble_length</control>
+      <control>auto_scribble_threshold</control>
+    </vbox>
+  </page>
+</fxlayout>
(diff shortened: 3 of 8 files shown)


RodneyBaker:
I'm not entirely sure how to test this as it doesn't successfully build with the automated builds.
It might be expected to build correctly via the nightlies.

konero:
Tested but trying to add the node causes crash.

image

Appule:
Sorry, I forgot to update the CMakeLists file!

shun-iwasawa:
FYI:  Naru ( @Appule  ) is my colleague. He joined our studio last April. 👍

RodneyBaker:
Thanks for the confirmation @Appule (and welcome aboard!)
It looked like a file was missing but haven't had time to delve deeper.

image

Adding Shaders/FX and such has always been difficult because it is so easy to forget something.

Appule:
I've fixed several issues. I hope it's working properly now.

Appule:
Added and renamed some parameters, and updated the Parameters section accordingly.
(specifically Auto Mask Scribble and Boundary Transparent Scribble)

RodneyBaker:
Success here.

I'm still figuring out how to work everything but I note:  It appears we need to use an image with white fills (background and fills) image as the source.  Otherwise, I get a black rectangle instead of the desired map.  

The good news is that if we fill toonz raster and vector images with white that works well.

image

shun-iwasawa:
Jenkins

Appule:
I’ve made it work with transparent background images. Thanks for the comment!

RodneyBaker:
Thanks for the update @Appule 
Much appreciated.

I believe I'm seeing an issue when drawings (on toonz raster levels) intersect with the Camera Frame.  Namely, the masked fill gets dropped.

I think @shun-iwasawa ran into that image border/camera frame issue with some other FX.

CameraFrameIssue

When the drawing doesn't intersect with the Camera Frame the fill of the mask persists:


Combination

RodneyBaker:
I've been having fun with this feature.
Chasing the boundary bug a little (vector has a history of that and here it helps to have an outside boundary to prevent clipping of vector lines at the edges).  Drawing 2 strategically placed dots (or 4 dots at the corners) outside of the main drawing will do the job.

I got distracted a little by experimenting with animated scribble controls and simulating lighting by turning parts of an image mask on/off over time.

I promise I'll try to focus on the core workflow soon though!

Here's a simple example of the Camera boundary problem that also displays the clipping issue with vector boundaries:

https://github.com/user-attachments/assets/7cb5de22-39f5-4b4b-993c-2b4a196ac4df

Workaround is as mentioned:  Add a frame or dots to extend the vector drawing.
The dots/frame can be transparent.

RodneyBaker:
The only issue I'm finding with this new feature is the boundary problem.
This is an outstanding addition to effects @Appule.

I am prepared to merge 'as is' but want to wait to see if you prefer the boundary problem fixed first.

In the meantime I'm experimenting with the many applications of the feature.

Appule:
Sorry for the late reply — I was on vacation.
I want to fix the boundary issue before merging. It may not be a perfect solution, but I think I can ease it somewhat.

RodneyBaker:
Thanks @Appule 
I'm looking forward to this getting into the hands of Opentoonz users but take whatever time you need.

(Hope you had a great vacation)

Abel032:
The boundary problem can only be solved by user input:
Require user to add the reference to ignore the unnecessary regions manually.
This is really good fx.@Appule
图片
(Boundary Transparent Scribble: Buttom and Left off)

Notice: The reference input must be normal raster level, can't be Toonz Raster.

---

BTW, I wonder if it's possible to use the colored lines from `ToonzRaster` in `Mask+Line` Mode, currently I can only do it by using the `Mask` Mode and an over fx node.
This feature would make this fx more useful in line test.

Abel032:
@RodneyBaker 
The boundary issue you noticed may related to OT's render system I think, 
the effect node might can only get one level's not complete area which limited by the camera.

RodneyBaker:
@Abel032 
@shun-iwasawa has resolved the boundary problem in a few other FX so I'll defer to his work on that.  I'll see if I can find those areas of code to reference but I'm confident Shun will know.  I do agree that is likely relates to what is processed by the camera.

Added:  I briefly explored trying to get the code from this PR to fill all areas/shapes it encounters with different colors rather than all with the same color.  I didn't get far but think it worth pursuing even if as a separate FX.  Perhaps call it Naru FX Lazy Brush plus FX.  ;)

Appule:
I fixed the boundary problem. It happened because, according to the specs of the OT, I was only processing the area within the camera instead of the entire input image. With this fix, I've changed it to process the entire input image regardless of the camera's range.

By the way, I've added some new parameters: 'Line's Weights' and 'Background Direction'.

'Line's Weights' allows you to adjust where the markers are placed based on the thickness of the lines because auto-scribble didn’t work properly if the line was too thick.

'Background Direction' is a modified version of the previous 'Boundary Transparent Scribble'. It allows you to set scribbles for the background and mask areas just by specifying a rough direction (experimental).

Abel032:
The Background Direction works good, I would consider as the possible moving direction of character or the negative direction of not moving part of character.

I tested the newest build and found two minor issues:

## Node Gragh
图片

## Original Image
图片

## After the fx
图片


## Issue
1. There are 1pix's offset of the line.
2. The line's color has turned from blue to black , I think it's better if we can keep the original image's color.

Abel032:
@RodneyBaker 
> Added: I briefly explored trying to get the code from this PR to fill all areas/shapes it encounters with different colors rather than all with the same color. I didn't get far but think it worth pursuing even if as a separate FX. Perhaps call it Naru FX Lazy Brush plus FX. ;)

It looks like a new Check Mode called `4 Colors Check`, works together with the `Gap Check`.

Appule:
I fixed two miner issues since they were easy to fix.

RodneyBaker:
> It looks like a new Check Mode called 4 Colors Check, works together with the Gap Check.

I haven't seen that but am looking forward to it.

@Appule 
Thanks for the fixes.
As the nightlies aren't currently generating we'll have a little more time to test this PR.
I haven't reached out to @shun-iwasawa but I believe he is aware. 
@andeon and @xavier2k6 have been looking into what dependencies that are changing.

RodneyBaker:
Back to having fun with lazy brush naru!

LazyBrushNaruFX

In the 'its a feature not a bug category' I am noting how rendering with a shrink value effects the results.

LazyBrushNaruFXshrink4

Thanks for the fixes!

RodneyBaker:
For those testing, it's generally best to render to 16bit.
LazyBrushNaruFX16bit

RodneyBaker:
I hope this explains what I'm seeing well enough.

There is a small issue with a halo of pixels around a drawing when Fill Holes is checked on but those same pixels are not visible when Fill Holes is checked off.  

The workaround to that is to increase the filter smoothness as this will make those external pixels go away.

Fill Holes on (pixels outside the linework):
A fillholdon

Fill Holes off (no pixels outside the linework):
A fillholdoff

Fill Holes on with Filter Smoothness slider increased (no pixels outside the linework):
A fillholdon compensate with increased Filter Smoothness

If this isn't clear I'll try to share better imagery.
I think what I might be suggesting is that having the default Filter Smoothness be higher might be appropriate.

RodneyBaker:
For those curious, this PR is still awaiting a review by @shun-iwasawa 
It will also need to be refreshed in light of changes to cmakelist.
The MacOS (Mojave) check can be dropped.

shun-iwasawa:
I believe minor issues can be improved after release. I'll merge this PR to gather more feedback.
Thank you @Appule! LGTM

PR #6043, opened 2025-08-07
Clapperboard for non-movie render

shun-iwasawa:
This PR will enable to render clapperboard as a single image file when rendering in non-movie format.

- The `Add Clapperboard` option is now available even when a non-movie file format (e.g. png, jpg, tif, etc.) is selected in the output settings.
- Clapperboard image will be saved with file names that replace the sequential number part of the output file with the specified suffix. (e.g. When the file name of the rendered image is `foo_0001.png`, `foo_0002.png`, `foo_0003.png` ..., the clapperboard image will have the file name `foo_board.png`.
- The suffix string (default is “board”) can be changed in the Edit Clapperboard dialog.
- Clapperboard image can also be exported (without running rendering) via a button in the Edit Clapperboard dialog.


RodneyBaker:
This is a very welcome update to the Clapperboard feature @shun-iwasawa.

I'd love to see an option in the UI to replace the first frame of the rendered output rather than use the suffix string although I love that you have that editble option.  That would save a lot of time loading both the .board image and the rendered sequence.  I expect in most cases that first frame would end up being 'frame.0000.png' or equivalent so could easily be discarded if not needed.

At any rate, this PR is ready to go but will delay as others may have feedback.

DAILIES board

RodneyBaker:
Thank you @shun-iwasawa for the update to Clapperboard.
Much appreciated!

PR #6048, opened 2025-08-07
Refactor(Auto Rename): Fix crash when importing sequence

### LINKED ISSUE #6035: Crash When Importing Image Sequence and Offering to Add Underscore
### Crash
When importing a image sequence with the naming convention `A0001.png` OpenToonz explains it expects `A.0001.png` or `A_0001.png` and offers to rename the sequence by adding an underscore.

If allowing it to do so, OpenToonz will crash. I can reproduce consistently, and haven't managed a successful rename.

### Feedback
The wording of the popup could also be improved, it's using technical terms which won't make sense to most people.

> Image sequence detected, but the filenames are missing a separator: OpenToonz requires a separator (such as an underscore ( _ ) or dot ( . ) between the name and the frame number to recognize sequences properly.
> 
> Example: A0001.png → A_0001.png
> 
> Would you like OpenToonz to automatically add an underscore to fix the sequence format?
> 
> File: C:\Anim\Test\A0001.png
> 
> [ Yes, add underscore ] / [ No, treat as single image ]

### Ref: https://github.com/opentoonz/opentoonz/pull/5768

Abel032:
Also optimized the description of renaming image sequence

fixes #6035

I refactored the renaming code in #5945, and I forgot that.
So I refactored it again...
Most changes made here would be undo in #5945, as I used the old code.

@RodneyBaker Maybe I should adjust #5945 after this PR?

### DIFF
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -853,11 +853,11 @@ void TSystem::copyFileOrLevel_throw(const TFilePath &dst,
 //--------------------------------------------------------------
 
 bool TSystem::renameImageSequence(const TFilePathSet &files,
-                                        const TFilePath &levelPath,
-                                        int prefixLength) {
+                                  const TFilePath &levelPath,
+                                  int prefixLength) {
   std::string levelBaseName = levelPath.withoutParentDir().getName();
-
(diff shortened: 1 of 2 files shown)


Abel032:
Close this as #5945 would provide better solution.

PR #6061, opened 2025-08-24
Fix deprecation warnings for Qt5.15 regarding QRecursiveMutex

Bracket-I-I:
I had this warning on earlier versions of Qt5, too, but Qt5.15_17_20250607_0825fcb1, the one I am on for sure.

If this is too far in the future for Opentoonz, well, then I guess this PR will be applicable when that time comes.


RodneyBaker:
This change is beyond my ability to review as I think it potentially might require plugins to be recompiled (?) Not sure.

Since we are here anyway I'll use this as an opportunity to see what a review by Co-Pilot will produce.

Bracket-I-I:
Lol, which is funny cause I don't use AI tools of that kind at all, so THAT is out MY league lol.
I guess I shall let the copilot take the wheel on the judgement.

It *seems* to be like...not spitting out anything too disconcerting.

Issue #6008, opened 2025-07-10
Build failure with GCC 15


krop:
### Description

opentoonz 1.7.1 fails to build with GCC 15:

```
[  208s] [ 41%] Building C object image/CMakeFiles/image.dir/tzp/avl.c.o
[...]

PR #6063, opened 2025-08-25
Bump CMake version up to 4, set C standard to 99

### LINKED ISSUE #5997: Current master version is not building on Linux unless old standards are passed to Cmake
### Description

I don't know how to fix it for real cause I'm not that good with cmake but
` cmake ../sources -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DCMAKE_CXX_STANDARD=17 -DCMAKE_C_STANDARD=99 -Dwarnings-as-errors=OFF` 
will at least make it compile, for now, for newer gc++ and cmakes.

### Steps to Reproduce

1. 
2. 

### Expected Behavior

_No response_

_No response_

### OpenToonz Version

1.7

_No response_

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

None

Bracket-I-I:
Should fix #5997 and #6008

Please test, especially for Mac... since I had to change the custom commands for the newer cmake syntax.

But it *should* work.

Note: This only fixes any need to add shenanigans to cmake like 
`cmake ../sources -DCMAKE_POLICY_VERSION_MINIMUM=3.5 -DCMAKE_CXX_STANDARD=17 -DCMAKE_C_STANDARD=99`

This does not fix any warnings like the Recursive Mutex that should be used in newer Qt5 sub versions as in PR:
https://github.com/opentoonz/opentoonz/pull/6061

It will only make it so (linux users and ) everyone (else) can compile it on newer libraries.

### DIFF
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8.11)
+cmake_minimum_required(VERSION 2.8.11...4.0)
 
 set(CMAKE_BUILD_TYPE_INIT Release)
 
@@ -8,7 +8,11 @@ project(OpenToonz)
 set(CMAKE_CXX_STANDARD 17)
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
 message(STATUS "C++ standard: ${CMAKE_CXX_STANDARD}")
-
+# Set C to 99 or sources/image/tzp/avl.c will not compile anymore for newer
+# compilers and libraries
+set(CMAKE_C_STANDARD 99)
+set(CMAKE_C_STANDARD_REQUIRED ON)
+message(STATUS "C standard: ${CMAKE_C_STANDARD}")
 #-----------------------------------------------------------------------------
 # Platform Specific Defaults
 
@@ -206,7 +210,7 @@ elseif(BUILD_ENV_APPLE)
     endif()
 elseif(BUILD_ENV_UNIXLIKE)
     # Needed for correct Qt detection
-    cmake_minimum_required(VERSION 2.8.12)
+    cmake_minimum_required(VERSION 2.8.12...4.0)
     set(PRELOAD_VARIABLE "LD_LIBRARY_PATH")
     if(CMAKE_SYSTEM_NAME MATCHES "Linux")
         add_definitions(-DLINUX)
--- a/toonz/sources/toonz/CMakeLists.txt
+++ b/toonz/sources/toonz/CMakeLists.txt
@@ -582,8 +582,8 @@ endif()
 
 # copy utility executables onto the directory after build
 if(LZODRIVER_FOUND)
-    add_custom_command(TARGET OpenToonz POST_BUILD COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:lzocompress> ${CMAKE_RUNTIME_OUTPUT_DIRECTORY} DEPENDS lzocompress)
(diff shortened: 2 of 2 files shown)


Bracket-I-I:
Also, this *should* be backwards compatible ...but I don't have enough experience (and willpower to downgrade my systems) to verify it...but this PR should be good to be included now, wheras the mutex one, if it doesn't happen to you (the main maintainers) then that can come later, perhaps.

Bracket-I-I:
I'll test that later. I just found out that windows fails with not finding the boost libraries...oh boy... 
The apple one might build now though.

But boy boost... that's gonna be a PITA to test without windows.

Bracket-I-I:
Again, no windows for me at this point in time, but could this be relevant? 
It seems that visual studio/vcpkg newest baseline (even at that point) provides the cmake boost finding and not cmake itself anymore?
https://github.com/microsoft/vcpkg/issues/40561

Bracket-I-I:
> Since GCC 15 enforces stricter type checking (in line with modern C standards), the current function declaration in avl.h causes a type conflict during compilation and will likely be rejected by newer compilers.
> 
> The issue is with this line:
> 
> https://github.com/opentoonz/opentoonz/blob/2ef8b0c4ee13707ef3ca28341b69b7c44d65e711/toonz/sources/image/tzp/avl.h#L226
> 
> A correct declaration would be something like this:
> 
> `TNZAPI TREE *avl__tree(int treetype, unsigned keyoffs, int (*usrcmp)(void *, void *));`
> 
> But to exactly match the definition in avl.c (which uses uint), it could be:
> 
> `TNZAPI TREE *avl__tree(int treetype, uint keyoffs, int (*usrcmp)(void *, void *));`
> 
> I don’t currently have a machine available to test this change, but since you likely have the proper environment and build setup, maybe you could verify it.

It seems to be a real dang mess where pointers are expected to be three different types but all used interchangeably.
So, there is no quick fix, a lot of casts would have to be done, all tailored to the function they're in and regarding where they are going and what is expected.

3 different permutations of int (*usrcmp) ())
one without anything, one with one void*, one with two void*s...  

Perhaps there is an equally or better performing tree implementation of that kind by now?
I mean, the one OT uses is from 1991. Not that old code must mean it's bad code, but new code sometimes IS better.

Bracket-I-I:
Dang, I don't know if I can fix this for Mac and Windows.

Perhaps! I can do it for Mac but Windows, once OT has to update cmake and co at some point...the boost not found thing, I think cannot be fixed on 'our' side since cmake does not include ways to find boost anymore, for windows at least?

Like, windows users have to use some way to configure msvc on their own to find boost. Not 'us' by 'us' asking cmake to find it.

andeon:
@Bracket-I-I Since the CMake issue now impacts development after changes in GitHub and AppVeyor, I’ve submitted an urgent PR to update the minimum required CMake version to 3.5, which is the standard minimum, and another changes to update other components to ensure proper functionality on macOS and other platforms.

While it needs to be tested, if it's approved, you should base your work on this change, as it should resolve some of the issues you're facing. Reorganizing CMake is really painful, these files often stay outdated for years in many projects.

Bracket-I-I:
Thanks for the pointer. I'll look out for it.

RodneyBaker:
Thought of you when seeing this:

https://github.com/AcademySoftwareFoundation/OpenImageIO/pull/4914/files

Bracket-I-I:
Hmm, well I'm in the process of setting up windows (but not mac << ) for something else, so I should be able to test this for two of the three big OSes soon-ish™..

So if that mutex thing is part of the cmake fix then I'll add it. But yeah, no idea right at this moment lol, thanks though ^^

Issue #6057, opened 2025-08-20
Crash drawing a straight line


YblisSp:
### What's the bug?

The program crashes after I try to use the line tool

### Steps to Reproduce

1. Have an animation project (about 28 seconds long ) with 2 toon raster layers, one with 9 frames, the other with 8 and a sound one
2. try to use line tool on the 5th frame of the 1st layer

### What did you expect to happen?

I'm not sure what I did wrong, I tried to specific to the best of my abilities, but I just wanted to draw a straight line :(

### Screenshots, Videos or Crash Logs

OpenToonz 1.7.1  (Build Jul 22 2025)

Report Date: 2025-08-20 02:30:08
Crash Reason: EXCEPTION_ACCESS_VIOLATION

Build ABI: x86_64-little_endian-llp64
Operating System: Windows 10 Version 2009
OS Kernel: 10.0.22621
CPU Threads: 12

GPU Vendor: Intel
GPU Model: Intel(R) UHD Graphics 630
GPU Version: 4.6.0 - Build 31.0.101.2111

Crash File: C:\Users\Gabriela Gil\AppData\Local\OpenToonz\OpenToonz\crash\Crash-20250820-023008.log
Mini Dump File: C:\Users\Gabriela Gil\AppData\Local\OpenToonz\OpenToonz\crash\Crash-20250820-023008.dmp

Application Dir: E:/aarrrgh/OpenToonz
Stuff Dir: E:\Novo OpenToonz

Project Name: eu sem você
Scene Name: única
Project Path: E:\Novo OpenToonz\projects\eu sem você\eu sem você_otprj.xml
Scene Path: E:\Novo OpenToonz\projects\eu sem você\scenes\única.tnz

==== Modules ====
E:\aarrrgh\OpenToonz\OpenToonz.exe
C:\Windows\SYSTEM32\ntdll.dll
C:\Windows\System32\KERNEL32.DLL
C:\Windows\System32\KERNELBASE.dll
E:\aarrrgh\OpenToonz\Qt5Xml.dll
C:\Windows\System32\USER32.dll
E:\aarrrgh\OpenToonz\Qt5PrintSupport.dll
C:\Windows\System32\win32u.dll
C:\Windows\SYSTEM32\OPENGL32.dll
C:\Windows\System32\msvcrt.dll
C:\Windows\System32\GDI32.dll
C:\Windows\System32\gdi32full.dll
C:\Windows\System32\ADVAPI32.dll
[...]

Issue #5465, opened 2024-05-22
Double clicking crashes the program


BrabaBooie:
### Description

Whenever I double click fairly fast (typically on accident when using a pen), the program crashes

### Steps to Reproduce

1. open a drawing scene (or any scene where you have the canvas)
2. select line tool
3. double tap/click
4. enjoy the crash (instructions May be flawed)

### Expected Behavior

After you click the program should pop up a crash report window.

### Screenshots, Video & Crash Logs

<img width="1372" alt="Screenshot 2024-05-22 at 11 57 42 AM" src="https://github.com/opentoonz/opentoonz/assets/112445071/f157a043-4372-4ee1-a1c9-857a1ebd6f4e">

### OpenToonz Version

1.7

### Operating System

macOS

### GPU

Intel

### Graphics Tablet

Wacom

RodneyBaker:
Likely relates to #5287

RodneyBaker:
This problem should be addressed in the latest nightly release.
Link:  https://github.com/opentoonz/opentoonz_nightlies/releases

Issue #5287, opened 2024-02-04
Crash when draw short line with Geometric Tool


n-ponto:
### Description

OpenToonz crashes when drawing a short line with raster sketch line geometric tool.
**NOTE**: this action works as expected with the default palette settings, it only seems to be when using a raster brush setting.

### Steps to Reproduce

1. Create a new toonz raster layer
2. Set the style to one of the raster brushes (I tried both the OT sketch brushes and some others)
3. Use the *Geometric Tool* set to *Shape: Line*
4. Draw a very short line tap/click in place (I'm having the issue with both mouse and drawing tablet)

### Expected Behavior

Create a dot or draw nothing (just don't crash plz)

## Crash log
OpenToonz 1.7.1  (Build Mar 14 2023)

Report Date: 2024-02-03 16:37:14
Crash Reason: EXCEPTION_ACCESS_VIOLATION

Build ABI: x86_64-little_endian-llp64
Operating System: Windows 10 Version 2009
OS Kernel: 10.0.22621
CPU Threads: 16

GPU Vendor: NVIDIA Corporation
GPU Model: NVIDIA GeForce RTX 3060/PCIe/SSE2
GPU Version: 4.6.0 NVIDIA 527.41

Crash File: C:\Users\noah\AppData\Local\OpenToonz\OpenToonz\crash\Crash-20240203-163714.log
Mini Dump File: C:\Users\noah\AppData\Local\OpenToonz\OpenToonz\crash\Crash-20240203-163714.dmp

Application Dir: C:/Program Files/OpenToonz
Stuff Dir: C:\OpenToonz stuff

Project Name: mr blue sky
Scene Name: test
Project Path: C:\OpenToonz stuff\projects\mr blue sky\mr blue sky_otprj.xml
Scene Path: C:\OpenToonz stuff\projects\mr blue sky\scenes\test.tnz

==== Modules ====
C:\Program Files\OpenToonz\OpenToonz.exe
C:\WINDOWS\SYSTEM32\ntdll.dll
C:\WINDOWS\System32\KERNEL32.DLL
C:\WINDOWS\System32\KERNELBASE.dll
C:\WINDOWS\System32\USER32.dll
C:\Program Files\OpenToonz\Qt5Xml.dll
C:\WINDOWS\System32\win32u.dll
C:\Program Files\OpenToonz\Qt5PrintSupport.dll
C:\WINDOWS\System32\GDI32.dll
[...]

PR #6064, opened 2025-08-26
Fix crashes in GeometricTool with MyPaint brush single/double click

andeon:
Fixes instant crash when using the Geometric Tool set to Line with MyPaint brushes on a ToonzRaster or Raster level. The crash occurs on single or double click.

In short, avoid crashes with MyPaint brushes in the Geometric Tool by validating strokes and adding null checks.

Ref: #5287, #5465, #6057


RodneyBaker:
Thank you @andeon for this fix!

PR #6068, opened 2025-08-28
[WIP] Standalone Scene

### LINKED ISSUE #6066: Scenes fails to decode the project folder path
### What's the bug?

After introducing #5954 , OT fails to load levels with the path with the project folder alias (like "+drawings", "+extras") when you save the scene outside of the "+scenes" project folder.

In many cases, users would like to save scenes not only under the "+scenes" project folder, but under any project folders. Therefore, when creating a new project, [OT saves "scenes.xml"s under every project folders](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonzlib/tproject.cpp#L605-L625) to know which project the scene belongs to.

However, current inplementation of `ToonzScene::isLonelyScene()` does not actually checks the existence of `scenes.xml` , but[ only checks if the scene is under the `+scenes` project folder](https://github.com/opentoonz/opentoonz/pull/5954/files#diff-0ee8edc63b37effb1882f7e0ee2497d82c3ee78a7579664cbc7f08c26bcf4369R1382-R1385). It can fails to decode the file path properly.

### Steps to Reproduce

_No response_

_No response_

_No response_

### OpenToonz Version

1.7

_No response_

### Operating System

Windows

### GPU Brand

NVIDIA

### Graphics Tablet

None

Abel032:
This PR resolves #6066

- Removed lonely scene path handling logic


shun-iwasawa:
I think you can achieve the desired behavior using the existing `Scene Folder Alias` option without needing to add a new option to `Path Alias Priority` .

Abel032:
@shun-iwasawa 

There is a slight difference between `Scene Folder Alias` Option and the new Option:
- `Scene Folder Alias` creates/loads new level under project folder `+scenes`(equal `$scenefolder` in this case) when the scene belongs to a project(scenes.xml detected in .tnz folder)
- `new option` creates/loads new level under project folders (`+scenes`, `+extra`)

I think user want `Scene Folder Alias` because they want to work without project and all fills in one directory,
and when user switch to `Project Folder Alias`, they want to work under project.

With this option, users don't need to change their preferences, no matter which workflow they are in.
It will also be more convenient when packaging a scene from a project and importing the scene back into the project.

Abel032:
This is just one idea: the production assistant would use the project to manage the whole episode, while artists work on individual cuts.

toonz/sources/toonzlib/toonzscene.cpp
```diff
@@ -318,6 +318,14 @@ void ToonzScene::clear() {
...
   assert(project);
   m_project = project;
+  m_standAlone = true;
+  for (int i = 0; i < project->getFolderCount(); i++) {
+    if (project->getFolder(project->getFolderName(i), true)
+            .isAncestorOf(m_scenePath)) {
+      m_standAlone = false;
+      break;
+    };
+  }
```
shun-iwasawa, line 328: As written in [my previous comment](https://github.com/opentoonz/opentoonz/issues/6066#issuecomment-3226673704), users may add `scenes.xml` manually in order to make any folder to belong the project.

Instead of checking `m_standalone` here, I think the detection of standalone can be obtained [this part](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonzlib/tproject.cpp#L1071-L1072) in the function `TProjectManager::loadSceneProject()` .

toonz/sources/toonz/preferencespopup.cpp
```diff
@@ -1680,6 +1688,10 @@ QWidget* PreferencesPopup::createGeneralPage() {
                            &PreferencesPopup::onWatchFileSystemClicked);
 
   bool ret = true;
+
+  ret = ret && connect(pathAliasPriorityCB, SIGNAL(currentIndexChanged(int)),
+      this, SLOT(onPathAliasPriorityChanged()));
```
shun-iwasawa, line 1693: OH, I seem to have forgotten to include this callback. Thank you!
Instead of calling onPathAliasPriorityChanged as a SLOT function, please do the following:

` m_onEditedFuncMap.insert(pathAliasPriority, &PreferencesPopup::onPathAliasPriorityChanged);`

This ensures that `onPathAliasPriorityChanged()` is called **after** the Preferences value is changed within `onChange()`.

shun-iwasawa:
> There is a slight difference between `Scene Folder Alias` Option and the new Option:
> 
> * `Scene Folder Alias` creates/loads new level under project folder `+scenes`(equal `$scenefolder` in this case) when the scene belongs to a project(scenes.xml detected in .tnz folder)
> * `new option` creates/loads new level under project folders (`+scenes`, `+extra`)
> 
> I think user want `Scene Folder Alias` because they want to work without project and all fills in one directory, and when user switch to `Project Folder Alias`, they want to work under project.
> 
> With this option, users don't need to change their preferences, no matter which workflow they are in. It will also be more convenient when packaging a scene from a project and importing the scene back into the project.

Understood. That does seem reasonable. 👍

shun-iwasawa (approved):
Thank you @Abel032 for the fix, LGTM

Abel032:
Ref: https://youtu.be/ND3cMhSif7w?t=152

PR #6069, opened 2025-08-28
Fix New Folder refresh issue in File Browser

Abel032:
Fix File Browser's **Folder** LineEdit not refreshing after creating  new folder and renaming it.
图片

It's a usable workaround, not best practice.

### DIFF
--- a/toonz/sources/toonz/dvdirtreeview.cpp
+++ b/toonz/sources/toonz/dvdirtreeview.cpp
@@ -286,9 +286,16 @@ void DvDirTreeViewDelegate::commitAndCloseEditor() {
   NodeEditor *editor = qobject_cast<NodeEditor *>(sender());
   emit commitData(editor);
   emit closeEditor(editor);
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@Abel032  If I understand what is happening here I'm still not seeing a new directory that gets renamed via Right Click > Rename update.   This aspect of the process may be a limitation of the Rename option because the directory doesn't change so it's not suprising that wouldn't update as there is nothing to update.

As for refreshing when double clicking to rename, I'm not seeing the problem prior to this PR.

Can you suggest steps that demonstrate the problem?

Abel032:
Yes @RodneyBaker 

The renaming behavior appear right after click the New Folder Button in the file browser
You can check this:
newFolder

RodneyBaker:
Thanks.  Your gif animation perfectly demonstrates the problem.
The failutre to rename a directory/folder via the Rename process is indeed a completely separate issue.

I'll proceed with merging this PR momentarily.

PR #6075, opened 2025-08-30
Refactor "Use lnverted Cursor Stroke End"

### LINKED ISSUE #6046: Issues with "Inverted Cursor"
Relates to: https://github.com/opentoonz/opentoonz/pull/5988

1. Cursor size is linked to zoom level, this is odd behavior for a cursor and makes it redundant. It's invisible if zoomed out enough, and too big to work as a crosshair when zoomed in too much.

2. When brush smoothing is cranked up high, it's impossible to steer the line if using a device that doesn't have a screen. For example, an Intuos or mouse. That's because this cursor is shown at the brush rendered position (the last stamp) rather than where the pointer device is. It also feels very odd to draw with even with a screen, usually the cursor is right under the pointer device immediately.

Abel032:
fixes #6046 

And new option name:
图片

Now the cursor would be exactly same with the old one, 
only the position of cursor would be changed to the end of stroke.

### DIFF
--- a/toonz/sources/common/tgl/tgl.cpp
+++ b/toonz/sources/common/tgl/tgl.cpp
@@ -160,32 +160,6 @@ void tglDrawCircle(const TPointD &center, double radius) {
   glPopMatrix();
 }
 
-void tglDrawInvertCursor(const TPointD& center,int gap, int len) {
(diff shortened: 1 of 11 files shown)


RodneyBaker:
Proceeding with this refactor and will await @konero's feedback.

PR #6078, opened 2025-09-01
Fix Clapperboard

shun-iwasawa:
This PR fixes the following problems regarding the clapperboard feature.
- When the item type is `Output Location`, the output file name becomes empty when it is not edited (i.e. the case when the scene name is used as the output file name).
- The file name suffix is not saved in the board settings.

@RodneyBaker 
> I'd love to see an option in the UI to replace the first frame of the rendered output rather than use the suffix string

You may have already noticed, but you can type `0000` to the `File Name Suffix` field to get the desired name. 👍


RodneyBaker:
I don't think the Mac OS (Mojave) check is legit here on github.
Perhaps that is a carry over from your local setup @shun-iwasawa?


As I recall we can safely ignore that check and press on.

RodneyBaker:
There are some issues with checks/builds that will need to be resolved so will delay merging this for now.

RodneyBaker:

Issue #5354, opened 2024-03-21
ToonzRaster Brush Smoothing Slider does not appear on Tool Options Bar when MyPaint Style is active


RattusaurusRex:
### Description

The ToonzRaster Smoothing slider works with MyPaint styles after a recent update (I believe from one of the Assistants updates), but the slider is not shown on the Tool Options Bar when a MyPaint style is selected.  You have to switch to a style with the default brush, change the Smooth setting, then switch back to the MyPaint style in order to change the amount of Smoothing.  If a user does not know this, they might be confused by the amount of smoothing applied to their brush when they can't see a way to change that amount of smoothing.

![image](https://github.com/opentoonz/opentoonz/assets/60277069/52a8ba36-1d90-4a69-acd0-c63fb5887526)

### Steps to Reproduce

1. Create a ToonzRaster level 
2. Create a style with the default brush and another with a MyPaint brush
3. With the default brush style active, use the Smooth slider to set the smoothing amount
4. Switch to the MyPaint brush style
5. Draw (note that the amount of smoothing that has been applied)
6. Switch back to the default brush style and change the smoothing amount
7. Draw with the MyPaint brush style again (note that the amount of smoothing has been changed)

### Expected Behavior

The Smooth slider affects MyPaint styles as well as the default ToonzRaster brush, so it should be shown on the Tool Options Bar when a MyPaint style is selected.

### Screenshots, Video & Crash Logs

https://github.com/opentoonz/opentoonz/assets/60277069/79db864e-0aab-47a1-9be1-b90ceffcadfb

### OpenToonz Version

Nightly (Specify below)

### OpenToonz Version Information

built 2024-03-16

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Wacom

PR #6079, opened 2025-09-01
Enable smooth option of ToonzRasterBrush when MyPaint style is activated

Abel032:
## Description:

- Enabled the Smooth option in ToonzRasterBrush when the MyPaint style is active.
- Refactored and sorted some code related to brush outline drawing.

图片

### DIFF
--- a/toonz/sources/tnztools/tooloptions.cpp
+++ b/toonz/sources/tnztools/tooloptions.cpp
@@ -1979,7 +1979,8 @@ void BrushToolOptionsBox::filterControls() {
        it != m_labels.end(); it++) {
     bool isModifier = (it.key().substr(0, 8) == "Modifier");
     bool isCommon   = (it.key() == "Lock Alpha" || it.key() == "Pressure" ||
-                     it.key() == "Assistants" || it.key() == "Preset:");
+                     it.key() == "Assistants" || it.key() == "Preset:" ||
+                     it.key() == "Smooth:");
     bool visible    = isCommon || (isModifier == showModifiers);
     it.value()->setVisible(visible);
   }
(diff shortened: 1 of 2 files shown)


RodneyBaker:
This may seem a little trivial but in most cases the size option for brushes is to the left of the smooth (or hardness) options.

image

Just an initial observation.
I'm still testing the core of this PR.

RodneyBaker:
Thanks for swapping the position of the slider.  
Much appreciated.

I'm having some trouble testing this PR because I don't see much (if any) difference in smoothness of the drawn strokes whether the smoothess is all the was at minimum or all the way at maximum.
The exception to this is when the slider is all the way up I get a slight sense of less control over placement of the stroke which I assume is the smoothness kicking in.

As I've been having some issues with smoothness settings outside of this PR I'm thinking this might be my system so will test on another computer as I get the opportunity.

I suppose a question to ask would be:  how much of a difference should we expect to see from min to max when using the slider?

The other question:  Is anyone else testing this PR that can provide feedback on what they are experiencing?

RattusaurusRex:
> Thanks for swapping the position of the slider. Much appreciated.
> 
> I'm having some trouble testing this PR because I don't see much (if any) difference in smoothness of the drawn strokes whether the smoothess is all the was at minimum or all the way at maximum. The exception to this is when the slider is all the way up I get a slight sense of less control over placement of the stroke which I assume is the smoothness kicking in.
> 
> As I've been having some issues with smoothness settings outside of this PR I'm thinking this might be my system so will test on another computer as I get the opportunity.
> 
> I suppose a question to ask would be: how much of a difference should we expect to see from min to max when using the slider?
> 
> The other question: Is anyone else testing this PR that can provide feedback on what they are experiencing?

It works like the ToonzRaster default brush Smooth slider for me, and it was already using the Smooth slider before now (even though the slider wasn't visible when you had a MyPaint style active).  https://github.com/opentoonz/opentoonz/issues/5354

Abel032:
LGTM

RodneyBaker:
I believe we will still need to look at the strength of the smooth slider across all uses but this new option for use with MyPaint styles is good to go.

PR #6080, opened 2025-09-01
Fix(OCA): correct endTime

Abel032:
## Description
Fixes `endTime` value assignment to comply with OCA standard specification. The OCA standard defines `endTime` as an exclusive boundary representing the first frame number that is **not** part of the animation.

Reference: https://oca.rxlab.guide/specs/root.html


RodneyBaker:
I don't have the setup to test (no DuIO or After Effects) but must assume the result is as intended.

PR #6081, opened 2025-09-01
Add "Delete lines and areas" when deleting styles

Abel032:
## Summary

- Add a new option when deleting style from palette:
图片

- Add dirty flag to level after deleting lines and areas (won't show though)

This feature is useful when cell painting is finished and you want to clean the auto-paint lines while keeping the styles in the palette.

### DIFF
--- a/toonz/sources/toonzlib/palettecmd.cpp
+++ b/toonz/sources/toonzlib/palettecmd.cpp
@@ -511,6 +511,7 @@ void eraseStylesInLevels(const std::set<TXshSimpleLevel *> &levels,
       else if (ti)
         TRop::eraseStyleIds(ti.getPointer(), styleIds);
     }
+    level->setDirtyFlag(true);
   }
 }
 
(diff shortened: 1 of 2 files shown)


RodneyBaker:
I'll add that this will be very useful when using the Segment Erase to tag and remove lines.
We still don't have T2D's implementation of Segment Erase for Lines so use of the Fill Tool on Line mode is still the only way we can do segment erase of lines in Opentoonz for toonz raster levels (Segment Erase for Eraser Tool only works with vectors).  

In some ways this approach to segment erasure is more versatile as we can work on tagging lines in a nondestructive way and only delete them (or make them transparent) when desired.

Abel032:
@RodneyBaker I believe that's a nice feature, I would consider to port it or refer it and implement.

The main different between the segment #0 fill and segment erase is whether to remain a #0 style line.
This line would affect fill tool filling.

RodneyBaker:
Thank you @Abel032 for the new option!

PR #6082, opened 2025-09-02
Replace Parent Directory modification

shun-iwasawa:
This PR modifies the `Replace Parent Directory` command, making the ancestor folder replaceable.

A scene exported from ClipStudio, for example, may contain folders named with each level within the scene folder, where the sequential images for that level are stored.
```
[C0001]  # scene folder
    ├─[A]
    │      A_0001.tga
    │      A_0002.tga
    │      :
    │
    ├─[B]
    │      B_0001.tga
    │      B_0002.tga
    │      :
    :
```
Conventional commands could not replace the scene folder at all levels at once (in the above example, replacing `C0001` by `C0002`).

This PR made it possible, by making the following changes:
- Now the command searches not only the parent folder of the selected levels but also their ancestor folders (currently up to three generations) and replaces any replaceable level files found.
- If this command replaces the ancestor folder instead of the parent folder, or replaces only part of the selected level, a dialog box will appear after command execution to report the results.
 
image

### DIFF
--- a/toonz/sources/toonz/filebrowserpopup.cpp
+++ b/toonz/sources/toonz/filebrowserpopup.cpp
@@ -86,11 +86,12 @@ FileBrowserPopup::FileBrowserPopup(const QString &title, Options options,
   setWindowTitle(title);
   setModal(false);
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:

PR #6083, opened 2025-09-02
Restore the function to select cell ranges by clicking the sidebar

shun-iwasawa:
This PR will restore the function to select cell ranges by clicking the sidebar, which had been disabled after introducing #5906 , as reported by @Ahl76 in [this comment](https://github.com/opentoonz/opentoonz/pull/5906#issuecomment-3140019311).

@Abel032 could you please check if this change does not break your original intension of #5906.

This PR also made the `Always Drag Frame Cell` preference option to be disabled by default. This is because users unaware of this option may accidentally drag and move cells in the xsheet without realizing it, and mess up the cells' correct order.
@Abel032 , please don't feel bad about it, I think this option is particularly useful for adjusting timing during animation.  However, unless it's clear that the vast majority wants it, I believe that when introducing an option that changes the existing user experience, it's better to set the default value to OFF at release and require explicit activation by the user.

Whether to change this option to ON by default could be discussed in a separate pull request after release.
Thank you for your kind understanding.

### DIFF
--- a/toonz/sources/toonz/xshcellviewer.cpp
+++ b/toonz/sources/toonz/xshcellviewer.cpp
@@ -699,13 +699,13 @@ void RenameCellField::showInRowCol(int row, int col, bool multiColumnSelected) {
         setAlignment(Qt::AlignLeft | Qt::AlignBottom);
         QFontMetrics fm(this->font());
         if (o->cellWidth() - 15 < fm.horizontalAdvance(text()))
-            setFixedWidth(fm.horizontalAdvance(text()) + 10);
+          setFixedWidth(fm.horizontalAdvance(text()) + 10);
         else
-            setFixedSize(o->cellWidth(), o->cellHeight() + 2);
+          setFixedSize(o->cellWidth(), o->cellHeight() + 2);
         connect(this, &QLineEdit::textChanged, this,
-                [this,o,fm](const QString &text) {
-                  if (o->cellWidth()-15 < fm.horizontalAdvance(text))
-                    setFixedWidth(fm.horizontalAdvance(text)+10);
+                [this, o, fm](const QString &text) {
+                  if (o->cellWidth() - 15 < fm.horizontalAdvance(text))
+                    setFixedWidth(fm.horizontalAdvance(text) + 10);
                 });
       }
       // other level types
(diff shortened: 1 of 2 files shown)


Abel032:
@shun-iwasawa Thank you for this change.
I originally intended to revise the changes introduced by "Always Drag Frame Cell" and set it to not enabled by default recently.
So it’s fine if this breaks the original PR, since that PR wasn’t well considered in the first place.

I also agree that it’s better to keep the default value OFF for the release, requiring explicit activation by the user.

> Whether to enable this option by default can be discussed in a separate pull request after the release.

By the way, when you say release, do you mean version 1.8, or are you referring to the nightly build?

shun-iwasawa:
> By the way, when you say release, do you mean version 1.8, or are you referring to the nightly build?

I believe it's fine to start the discussion anytime after the feature PR itself is posted.  
Of course, we could gather more diverse opinions after the V1.8 release, but even without that, I think it should be fine as long as we consider the opinions of users who utilize OT in various ways.  

My policy regarding merging PRs into OpenToonz remains unchanged from what is written [here](https://github.com/opentoonz/opentoonz/issues/2746#issue-487829469) :
  _PR can be merged if it will not interfere existing usages_.

RodneyBaker:
Thank you @shun-iwasawa for restoring the sidebar selection of cells.
It's very important to be able to select and move entire sections of exposures easily.

LGTM

Ahl76:
@shun-iwasawa thank you that works fine.

Ahl76:
@Abel032 @shun-iwasawa  It seems that the drag cell feature has a small problem when it comes to a single cell.

https://github.com/user-attachments/assets/b153ff1b-dc3d-4341-a0a6-fa483735ce9f

It's possible that this is related to this single drag cell feature from the sidebar ?

https://github.com/user-attachments/assets/5c088f1b-b643-4727-96fb-4200fb40fa31

Abel032:
@Ahl76 
One solution is to press Alt when moving the single cell.
This would temporarily disable the drag feature.

I may adjust like this:
- Only enable drag feature when both the front cell and the back cell are not equal and not empty.

I also plan to adjust the cell moving behavior again in the coming days. One possible adjustment would be:
- In Always drag frame cell mode, the drag bar would be always used to select the cell ranges.

When I pen the PR, it's welcome to test and feedback.

Abel032:
@Ahl76 
Please check if #6131 fixed the problem, thanks!

PR #6084, opened 2025-09-02
Refactor Assets Import Process & Image Sequence Renaming

### LINKED ISSUE #6035: Crash When Importing Image Sequence and Offering to Add Underscore

Abel032:
### This PR changes:

- Move LoadResourceArguments>FrameIdsSet to LoadResourceArguments>ResourceDatas>m_frameIds
- Refactor IoCmd::loadResource()
- Refactor Image Sequence Renaming

fixes #6035


RodneyBaker:

toonz/sources/toonz/iocommand.cpp
```diff
@@ -2544,33 +2464,36 @@ int IoCmd::loadResources(LoadResourceArguments &args, bool updateRecentFile,
...
 
-    if (!rd.m_path.isLevelName())
+      if (!path.isLevelName())
       path = TFilePath(path.getLevelNameW()).withParentDir(path.getParentDir());
 
-    if (std::find(paths.begin(), paths.end(), path) != paths.end()) {
+      // duplicate check
+      auto isDuplicate = [&rd](const IoCmd::LoadResourceArguments::ResourceData& existingRd) {
+          return existingRd.m_path == rd.m_path;
+          };
```
shun-iwasawa, line 2483: @Abel032 
Sorry for my delayed comment. This fix seems to have introduced the following issue:

[Steps to reproduce the issue]
- Store sequential images in a subfolder within the Project Folder
  (e.g., `+extras/foo/A.0001.png`, `A.0002.png`, `A.0003.png`...)
- Load that folder using the `Load Folder` command

Before this PR, a dialog would appear asking whether to allow duplicate levels.
After this PR, the confirmation dialog no longer opens, and the same level is placed on the xsheet as many times as the number of sequential images.

The reason for this failure is likely that the path being compared, `existingRd.m_path`, is encoded using the project folder alias, whereas `rd.m_path` is the full path before encoding.

Abel032: `the same level is placed on the xsheet as many times as the number of sequential images.`
I cannot reproduce this behavior. And, no dialog appears when I use the "Load Folder" command.
I will investigate further. Thank you for the bug report.

shun-iwasawa: Here is a movie version of the step. 
Please note that the preferences and other settings are unchanged from the default values.
load_folder

This issue can also be reproduced by dragging and dropping the folder icon for `+extras/foo` from Windows Explorer onto the OpenToonz Xsheet panel to load the levels in it.
(This is because this operation uses the same function as the `Load Folder` command.)

Abel032: @shun-iwasawa 
Yes, I can reproduce the bug by importing sequence named with underscore(_) separator now.
图片 

图片
And the behavior of loading sequence with dot(.) separator is exporsing directly to the xsheet.

shun-iwasawa: Wow, I hadn't imagined that the separator would be included in the occurrence conditions.

PR #6085, opened 2025-09-02
Enable to load .xdts as scene from OS FileBrowser

Abel032:
This allows OT to load .xdts as scene from OS's File Browser.


RodneyBaker:
Thank you @Abel032 for this addition.
It isn't always obvious how XDTS files can be ingested into Opentoonz so this will help.

PR #6086, opened 2025-09-02
Add fallback path suggestion with fuzzy name matching to XDTS Import Popup

Abel032:
This PR adds an internal fallback path suggestion feature to XDTS Import Popup.

- Triggered only when the original search finds no matching file.
- Uses fuzzy name matching. Files containing the level name can be suggested.
- Searches all first-level subfolders under the given path and the given path itself.


RodneyBaker:
I'm not a regular user of XDTS so...
Feedback from anyone with an interest in testing is always appreciated.

RodneyBaker:
Proceeding with this update.
Thank you @Abel032!

PR #6087, opened 2025-09-03
Autoclose:  Transform input angle to radians and correct angle application

Abel032:
This PR fixes angle handling by converting degrees to radians for spot samples instead of treating the input angle directly as radians.

<table>
  <tr>
    <td align="center" valign="top">
      <b>Before</b><br>
      
    </td>
    <td align="center" valign="top">
      <b>After</b><br>
      
    </td>
  </tr>
</table>


Abel032:
I'm not sure how user can test this change, but I suggest to create a image with many gaps, turn on the Gap Check, and then drag tape tool's angle from 0 to 360.

The result should be more stable than the old version.

RodneyBaker:

PR #6088, opened 2025-09-03
Improve Freehand Fill Visualization in ToonzRaster

Abel032:
Shows which areas the Fill Tool's freeHand type will fill.
(e.g., the No.1 area in the example image)

图片

### DIFF
--- a/toonz/sources/include/toonz/strokegenerator.h
+++ b/toonz/sources/include/toonz/strokegenerator.h
@@ -98,6 +98,8 @@ In caso affermativo li cancella.
   //! Visualizza tutti i frammenti
   void drawAllFragments();
 
+  void drawFilledStroke();
+
   // Only keep first and last points. Used for straight lines
   void removeMiddlePoints();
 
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Ok.  That's useful.
I have a sense that users will want to be able to adjust the color of the visualized area (perhaps via Preferences > Additional Style Sheets) but that's something for future consideration.

Posting intent to merge as this improvement will be obvious when used.

beeheemooth:
Does this show up in Frame Range mode too?

RodneyBaker:
The initial coloring of the selection does show up in Frame Range mode, yes.
The visual color only stays visible while drawing a selection. 
It does not persist in staying visible between selections.

Note the limitation of usage of Freehand and of Toonz Raster though.

RodneyBaker:
Thanks @Abel032 for the improvement!

Abel032:
@beeheemooth 
It only shows when drawing, do you think it's better if we show after drawing in Frame Range Mode?

Issue #2709, opened 2019-08-14
The whole system crashes when I run opentoonz 1.4


Gilraiser:
**Issue Summary**

The whole system crashes when I run opentoonz. This happens in this version and the Morevna project version.

**Steps to Reproduce**

The problem occurs when I run opentoonz having installed it either from appimage, flatpak or from the official Archlinux repositories.

**System Information**

OpenToonz Version: 1.4.0 Duango 
Operating System: Archlinux 64x
CPU: Amd athlon 5150 APU. 4 cores, 4 threads
Memory: 12 GB
Graphics Card: Nvidia Gforce GT 610

I hope the information is useful.
see you soon

RodneyBaker:
The official release doesn't run on Linux so I think we may be missing something about your setup.
Looks like you are using Linux across the board.

I'll recommend trying with the OTX Linux release just to see if anything changes.
It's 'experimental' but worth taking a look.
Look under the Assets dropdown here:  https://github.com/manongjohn/OTX/releases

Gilraiser:
Hi, thanks for answering. I did the test with the version you indicated and the following error appears: 

**varName:TOONZROOT TOONZROOT not set...
terminate called after throwing an instance of 'TSystemException'
Abortado (`core' generado)**

I hope the information is helpful. Thank you for releasing and maintaining this software.

manongjohn:
@Gilraiser If you used OTX, when you download the Linux version, just uncompress it. It extracts a folder called OpenToonzPortable. It sounds like you might have taken the AppImage out and ran it. You need to keep everything in that folder together.

Gilraiser:
Hi, that's exactly what I did and I gave him permission to run sudo chmod 775 -R / folder-opentoonz.

Gilraiser:
At no time did I separate the appImage from the portablestuff folder

manongjohn:
Try this 777 on the portablestuff folder and contents

Gilraiser:
Yes, this way it already worked. In the previous way 775 I have other applications run but with this appImage I didn't want to work. Thanks a lot friend. As far as I have tried it works without blocking.

Gilraiser:
Hello, I wish you are very well. I was further testing opentoonz and it happened again that the whole system was paralyzed. This time I reproduced the error in archlinux and also in lubuntu. The error tends to happen using the tools to modify nodes, for example the magnet tool.

 I hope the information is helpful.

 Thank you.

[...]

PR #6089, opened 2025-09-03
Fix(TENV): Correct the initialization of m_rootVarName in EnvGlobals

Abel032:
This PR ensures correct initialization of `m_rootVarName` in `EnvGlobals::iniit()`.

The specific issue this change fixes is unclear; I might confirm it later, or someone else can.

### DIFF
--- a/toonz/sources/common/tapptools/tenv.cpp
+++ b/toonz/sources/common/tapptools/tenv.cpp
@@ -179,7 +179,10 @@ public:
       m_applicationFullName += " " + m_version.getAppNote();
 
     m_moduleName  = m_version.getAppName();
-    m_rootVarName = toUpper(m_version.getAppName()) + "ROOT";
+    // m_rootVarName value is usually inited by SetRootVarName(), but sometimes
+    // init would be called before SetRootVarName(), so init it here one time
+    m_rootVarName = "TOONZROOT";
+    //m_rootVarName = toUpper(m_version.getAppName()) + "ROOT";
 #ifdef _WIN32
     // from v1.3, registry root is moved to SOFTWARE\\OpenToonz\\OpenToonz
     m_registryRoot =


RodneyBaker:
Interesting.
Should resolve a number of potential issues.
For example:  #2709

Random thoughts not directed at this PR:
Somewhere in the future we may want to make branding/forking easier.
In this current approach, brand derived env-vars are intentionally dropped.
If any downstream code/forks rely on toUpper(AppName)+"ROOT" at boot time those paths won’t get hit.  For OpenToonz itself that’s a good thing as it enforces the single documented TOONZROOT but for forks with different branding we might... eventually... prefer a different canonical name as this currently requires additional customization in their tree.

Thanks @Abel032 
If you run into any specific information regarding what issues this adjustment addresses it will be good to add the reference.

Abel032:
Might related to the Registry as I remember, ToonzFarm?
I'm not sure..

RodneyBaker:
This is also mostly unrelated but in trying to launch Opentoonz from batch file to test the TOONZROOT and OPENTOONZROOT environmental variables I got the missing glut32.dll (tsrv32bit) error that prevents Opentoonz from running**.  Adding a copy of glut32.dll to the root directory of where I was launching Opentoonz from (in this case F:\drive) allowed Openntoonz to launch.  This does suggest (as you mentioned) that the registry is likely involved.

(Launching from Powershell)
PS F:\> start "C:\Program Files\OpenToonz\OpenToonz.exe"
PS F:\> Attribute Qt::AA_UseDesktopOpenGL must be set before QCoreApplication is created.
[ThemeManager:::preloadIconMetadata] STARTED with path: ":/icons"
plugin search directory:C:\OpenToonz stuff\plugins
walkDirectory_: C:\OpenToonz stuff\plugins
===== PluginLoadController::finished() =====
plugin search directory:C:\OpenToonz stuff\plugins
walkDictionary: C:\OpenToonz stuff\plugins [dry]
QLayout: Attempting to add QLayout "" to ParamViewer "", which already has a layout
QLayout: Attempting to add QLayout "" to QWidget "", which already has a layout
Global QPixmapCache cleared

It's not clear to me why the glut32.dll file in the same directory as Opentoonz.exe is not recognized.

I need to to move on because I'm way beyond my competency.  ;)

**Although an instance of the program is launched in memory which must be forced to quit

PR #6091, opened 2025-09-03
Fix Fill Tool Segment for StyleId #3 by updating damInk

Abel032:
This PR fixs frequent undo history issues and fill failures in Fill Tool (Lines Mode, Segment on) for Ink lines of style #3 in the palette.

图片


RodneyBaker:
Thank you for the fix @Abel032

PR #6098, opened 2025-09-10
Add NAA FullColor Level Auto-Converting feature

Abel032:
This feature can help user convert imported NAA image sequence to ToonzRaster Level (Consider all pixels as INK), so that user can do cell painting.

图片


RodneyBaker:
It's a little odd but I thought this PR was merged already.
As always, thanks for the additional option @Abel032

PR #6099, opened 2025-09-10
Add preference: Paint under/over INK in refer fill or gap close fill

Abel032:
This PR adds a preference option of whether to do prevailing paint for reference fill.
The current implementation didn’t consider the situation where the reference level might be down/left to the fill level.
This usually happens when the level is used as both the upon and down part of one image, for example, a character’s head and his front hair and back hair.

图片

### DIFF
--- a/toonz/sources/include/toonz/fill.h
+++ b/toonz/sources/include/toonz/fill.h
@@ -26,6 +26,8 @@
 #include "preferences.h"
 #define DEF_REGION_WITH_PAINT                                                  \
   Preferences::instance()->getBoolValue(PreferencesItemId::DefRegionWithPaint)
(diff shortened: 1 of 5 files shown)


Abel032:
I noticed that the areaFiller(Rectangle / FreeHand) has two problems:
- Failed to recognize the gap close and refer lines(I've implemented this as I remember)
- In TLV SaveBox Fill mode may cause crash when refer fill enabled

I think there are something wrong with the affine calculating, would make another PR to fix them.

Update: I can't reproduce it now, weird...
Seems only happen when trying to fill in an empty ToonzRaster Level (with empty saveBox)

RodneyBaker:
Thanks for the follow up to your new features.
Much appreciated!

PR #6104, opened 2025-09-11
Fix full color eraser crash

Abel032:
## Issue Description

Crash when using Full Color Eraser with double click (FreeHand/Normal Type, FrameRange off).


RodneyBaker:
There are issues with the automated Mac OS checks but will proceed with this fix as it addresses crashing reported by several over the past few weeks.

PR #6105, opened 2025-09-12
Refactor AreaFiller of Fill tool

Abel032:
# Summary
This PR mainly fix the support of `refer fill` and `gap close` options in `area filler` type, and fixes the laggy of `Fill Check`.
图片

## Other changes are as follows:
- Mange refer image with RAII
- Add lock to the refer level's image before generating the refer Image
- Fix out of bounds crash in stagevisitor

# How to test

Here is my test cases for reference:
图片
I also tested `Fill Check`, it works with less delay now(almost the same as other check modes' speed)

### DIFF
--- a/toonz/sources/include/toonz/fill.h
+++ b/toonz/sources/include/toonz/fill.h
@@ -132,6 +132,9 @@ else fill ink and paint in rect.
   bool rectFill(const TRect &rect, int color, bool onlyUnfilled,
                 bool fillPaints, bool fillInks);
 
+  // Only for Fill Check
+  bool rectFastFill(const TRect& rect, int color);
+
   /*!
 Fill the raster region contained in spline \b s with \b color.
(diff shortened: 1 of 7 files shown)


Abel032:
@RodneyBaker 

The tests in the **How to test** section have all passed, so I believe this PR is ready for review and possible merge.
Please help confirm whether the current test coverage is sufficient.

If anyone has the time and interest, please take a look.

Abel032:
@RodneyBaker 
I've did some tests just now.
This PR looks fine to me, please review and possible merge when convenient.

RodneyBaker:
Very nice @Abel032 
I did temporarily run into an issue where I couldn't render and that apparently was due to the fill level going away (although still appearing to be present in the icon views).  I haven't seen the problem again but that's something we should look out for.

In other words, LGTM

Abel032:
The render issue might related to the image changed signal, but I'm not sure.

Issue #5754, opened 2025-01-31
Plans for Qt6?


a17r:
### Ask a question related to OpenToonz

At this point in time, Linux distributions are starting to track which packages are blocking a future Qt5 cleanup. Real Qt Company OSS support stopped back in 2020, when public availability of LTS releases started to be delayed for 1 year, stable branch repository access got closed down (with 5.15 branch repository receiving any new LTS tag one year after commercial release) and only Qt6-relevant bugs are being fixed (then backported) since then.

Qt5Script was only provided as a porting aid from Qt4, is deprecated since Qt 5.5, last commit in upstream dev repo happened in 2021-10-26. Even though it has come up already in https://github.com/opentoonz/opentoonz/issues/731 for the wrong reason: Porting away from Qt5Script would be an important first step and would let this package live longer in distribution repositories, even if it were still based on Qt5 in the mid-term.

The good news is that the porting effort to Qt6 is supposed to be much smaller than from Qt4 to Qt5. Links:
https://doc.qt.io/qt-6/portingguide.html
https://www.qt.io/blog/references-and-hints-about-porting-from-qt5-to-qt6
https://www.qt.io/blog/porting-from-qt-5-to-qt-6-using-qt5compat-library

RodneyBaker:
The two people with the most experience moving into Qt6 will likely be @shun-iwasawa and @manongjohn.

@manongjohn (fairly) recently made good strides to move into Qt6 but ultimately had to back down to the custom Qt that Shun created that Opentoonz is now using.  John could tell you a lot more about that but some of that is captured in the Tahoma2D discord.

Shun's custom Qt (I believe) can be found here:  https://github.com/shun-iwasawa/qt5/releases
The primary difference being Wintab support.

Here's the VFX Platforms tracking of movement into Qt6 (and other support software) which they are recommending:  https://vfxplatform.com/

RodneyBaker:
I don't really know much about it.
Here is Shun's original PR that implemented that custom Qt:  https://github.com/opentoonz/opentoonz/pull/3823

RodneyBaker:
Of possible interest you might want to follow Krita's current move to Qt6 where they are slowly but progressively moving forward.  One of their recent posts related they were able to open the Qt 6 enabled program and load an image.  Other progress and status is posted regularly. 

Progress toward Qt6 is being made albiet slowly.

a17r:
[...]

Issue #3593, opened 2020-11-18
(FreeBSD) error: use of undeclared identifier 'TIFFReadRGBATile_64'; did you mean 'TIFFReadRGBATile'?


yurivict:
```
/usr/ports/graphics/opentoonz/work/opentoonz-3073c05df584f91c54c112f58db8f1be4622bf22/toonz/sources/image/tif/tiio_tif.cpp:465:19: error: use of undeclared identifier 'TIFFReadRGBATile_64'; did you mean 'TIFFReadRGBATile'?
        int ret = TIFFReadRGBATile_64(m_tiff, x, y, tile.get());
                  ^~~~~~~~~~~~~~~~~~~
                  TIFFReadRGBATile
/usr/local/include/tiffio.h:433:12: note: 'TIFFReadRGBATile' declared here
extern int TIFFReadRGBATile(TIFF*, uint32, uint32, uint32 * );
           ^
```

Apparently it doesn't pick up the bundled version of the ```tiff``` project.

rev. 3073c05df584f91c54c112f58db8f1be4622bf22
OS: FreeBSD 12.2
clang-10

RodneyBaker:
I'm tagging this as bug mostly to move toward resolution of the underlying question of this report.

andeon:
@RodneyBaker I think this issue can be closed. It might be a similar situation to [opentoonz/opentoonz#6244](https://github.com/opentoonz/opentoonz/issues/6244) where the user may need to run `make install` for tiff-4.0.3, since the error is the same.

PR #6108, opened 2025-09-12
ci(build): update CMake to 3.5 and fix macOS build issues

andeon:
- Bump minimum CMake version to 3.5
- Adjust build flags and paths for compatibility in CMakeLists.txt
- Refactor post-build steps for macOS (RPATH, Info.plist, portability)
- Fix deprecated uint32/uint16 warnings in tiio_tif.cpp to support macOS build

Additional minor updates:
- Update Windows and Appveyor GitHub workflows
- Fix macOS build: add openjph and openexr (missing dylib after Homebrew update)

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -53,7 +53,7 @@ jobs:
           
           brew update
           brew cleanup
-          checkPkgAndInstall pkg-config glew lz4 libjpeg libpng lzo boost libusb libmypaint ccache jpeg-turbo ninja opencv qt@5
+          checkPkgAndInstall pkg-config glew lz4 libjpeg libpng lzo boost libusb libmypaint ccache jpeg-turbo ninja opencv openjph openexr qt@5
           brew unlink qt
       - name: Set up cache
         run: |
@@ -70,10 +70,12 @@ jobs:
           cd thirdparty/tiff-4.0.3
           CFLAGS='-fPIC' CXXFLAGS='-fPIC' ./configure --disable-lzma
           make -j $(sysctl -n hw.ncpu)
+          make install  # Ensures CMake finds the installed lib (avoid linking issues)
       - name: Build
         run: |
           export PKG_CONFIG_PATH="/usr/local/opt/jpeg-turbo/lib/pkgconfig:$PKG_CONFIG_PATH"
           cd toonz
+          rm -rf build  
           mkdir -p build
           cd build
           cmake ../sources -G Ninja \
--- a/.github/workflows/workflow_windows.yml
+++ b/.github/workflows/workflow_windows.yml
@@ -5,11 +5,11 @@ on:
     branches:
       - master  # Specify branches where the workflow should run
     paths:
-      - '**/*'                   # Include all files in the repository
-      - '!appveyor.yml'          # Ignore changes in AppVeyor
-      - '!README.md'             # Exclude README
-      - '!doc/**'                # Exclude documentation
-      - '!**/.github/**'         # Exclude everything in .github folder
+      - '**/*'                                    # Include all files in the repository
+      - '!appveyor.yml'                           # Ignore changes in AppVeyor
+      - '!README.md'                              # Exclude README
+      - '!doc/**'                                 # Exclude documentation
+      - '!**/.github/**'                          # Exclude everything in .github folder
       - '.github/workflows/workflow_windows.yml'  # Include Windows workflow file
   pull_request:
     paths:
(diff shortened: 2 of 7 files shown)


andeon:
The first PR has an issue because AppVeyor is using Visual Studio 2022, but the Qt version being searched for is built with MSVC 2019. So I had to revert to Visual Studio 2019 to test it, and it worked.

If switching to Visual Studio 2022 is necessary, then the Qt path should also be updated to the 2022 version. It seems the older CMake configuration doesn’t check this properly. Since I haven’t tested this setup, I’m leaving it as is for now. If someone could test Qt 2022 with Visual Studio 2022 to verify it works, we could update it again. 

Also, we need to check why AppVeyor isn’t using the Qt patch from @shun-iwasawa.

- [x]  **Done!** ~~I'll squash the commits to clean things up. There's no way to test AppVeyor before sending the pull request, so please don’t merge it yet.~~

Need to spot some issues related, because I think this PR too resolves them such as #3593 by fixing compilation errors (e.g., type mismatches in libtiff calls) in tiio_tif.cpp, without any loss of functionality. To validate and ensure there is no regression, it's important to check some TIFF files and compare.

xavier2k6:
AppVeyor have included newer (Boost 1.87, 1.88, 1.89) in 2019/2022 images
https://www.appveyor.com/updates/

xavier2k6:
Why didn't you update minimum CMake to `3.10+`?  xref.: #5834, also https://github.com/opentoonz/opentoonz/issues/5754#issuecomment-3288007733

EDIT:
https://cmake.org/cmake/help/latest/release/3.31.html#deprecated-and-removed-features

RodneyBaker:
For reference of what @manongjohn is doing regarding Mac OS over in Tahoma2D:  https://github.com/tahoma2d/tahoma2d/pull/1909

xavier2k6:
https://github.com/opentoonz/opentoonz/blob/3f2991eb9cd8dc062405de826bfbc2b7e0c210d8/.github/workflows/CMakeLists.txt#L1

andeon:
> Why didn't you update minimum CMake to `3.10+`? xref.: #5834, also [#5754 (comment)](https://github.com/opentoonz/opentoonz/issues/5754#issuecomment-3288007733)
> 
> EDIT: https://cmake.org/cmake/help/latest/release/3.31.html#deprecated-and-removed-features

@xavier2k6 It's a slow-moving process, and I'm taking a more conservative approach. I tested with CMake 3.5, and it might also work with 3.10. However, the main priority right now is to stabilize development so we can properly test the builds and users can start receiving the improvements. It took days of testing, tweaking, and figuring things out just to get this working. I'll need to check later if it actually works with 3.10, so it could possibly be updated in the future.

> https://github.com/opentoonz/opentoonz/blob/3f2991eb9cd8dc062405de826bfbc2b7e0c210d8/.github/workflows/CMakeLists.txt#L1

I have no idea why this CMakeLists.txt file is located in .github/workflows/. Most likely, that file is misplaced.

> For reference of what @manongjohn is doing regarding Mac OS over in Tahoma2D: [tahoma2d/tahoma2d#1909](https://github.com/tahoma2d/tahoma2d/pull/1909)

@RodneyBaker I considered this as well, but I preferred to start by trying to solve the problem more definitively. Also, in that case, I would need to check and find out if there is a way to fix some configuration on AppVeyor to ensure it works there as well.

RodneyBaker:
@andeon When in doubt I prefer the conservative approach.
That doesn't mean I always take that path but I definitely prefer it.

You all are doing great at keeping us moving ahead and where we run into roadbumps you've always come through to get us back on track.

Much appreciated.

I'm of the mind to merge this PR for the good it will do but I know we are constantly moving forward.  This is especially true as we now ramp up to the version 1.8 release.

Bottom line:  Now is the time to make our mistakes quicker so they don't appear in v1.8.

RodneyBaker:
Will proceed with merging this change with anticipation of additional changes later.

xavier2k6:
>I have no idea why this CMakeLists.txt file is located in .github/workflows/. Most likely, that file is misplaced.

I found it strange to be there too.....but might be no harm in bumping CMake `min` there until further investigation can be done.

andeon:
> I found it strange to be there too.....but might be no harm in bumping CMake `min` there until further investigation can be done.

Looking through commits, I found it was used in the old `workflow_windows.yml`, but it didn’t work correctly and is causing strange behavior with MyPaintBrushes. The current `workflow_windows.yml` does not use this file anymore. I should have removed it when updating the workflow, but I hadn’t noticed it before. This directory is specifically used for GitHub Actions workflow YAML files, so removing it is safe.

shun-iwasawa:
@andeon 
Sorry for the late comment, but after introducing this PR, it seems that 16bpc TIF images can no longer be loaded correctly. Could you please check if you can reproduce the issue on your end? Also, could you let me know if you have any idea what might be causing it?

image

andeon:
> @andeon Sorry for the late comment, but after introducing this PR, it seems that 16bpc TIF images can no longer be loaded correctly. Could you please check if you can reproduce the issue on your end? Also, could you let me know if you have any idea what might be causing it?

@shun-iwasawa  I suspected this might happen, it's related to a change in tiio_tif.cpp file. I think the issue on macOS may be due to mixing an older TIFF library with another one from Homebrew, and possibly other compatibility issues, especially since macOS 13 is becoming deprecated.That said, the build is no longer blocked.

I'm going to revert the change. Could you please share the original TIFF file here in the [PR](https://github.com/opentoonz/opentoonz/pull/6144), so we can use it for testing and validate in case the future changes?

PR #6110, opened 2025-09-13
ci: Update `checkout` action to `v5`

xavier2k6:
* https://github.com/actions/checkout/releases


RodneyBaker:
I don't see anything off with this change but it is worth noting for those that might host their own runners:  Self-hosted runners should make sure they are at v2.327.1+ otherwise the checkout step will fail under v5. 

GitHub-hosted runners are already compatible.

Thanks @xavier2k6!

PR #6111, opened 2025-09-13
Update Fill tool visualization

Abel032:
## Summary

- Draw Fragment for Free Hand type
- Fix TLV Bounds Visualization(more accurate)
  - fix the normal image bounds (you can see the gray lines at the edge of camera bounds before this PR)
  - fix the saveBox bounds (toggle the preference `Tools->Use the TLV Save` to see it, it was 0.5 bigger before this PR)
  
## How to test
StrokeFill
- Use freeHand type of fill tool, and check the shape.
- Turn on `Preference>Tools>Use the TLV Savebox to Limit Filling Operations` and switch to fill tool (or switch to select tool and toggle `Modify Savebox` on, check if the line touch the bound.
- Check the bounds of camera and current toonz raster level's bounds. (Use the animate tool if can't distinguish them)


Abel032:
LGTM, but wait for #6105 to restore the entire function of area filler.

Abel032:
Updated the newest changes from main branch.
This PR is a visual change, so it assumes not to affect the function.
I only tested in the old build, but I believe the result would be the same in the new build.

RodneyBaker:
Thank you for the update @Abel032 
LGTM

PR #6112, opened 2025-09-13
feat: fuzzy start pixel when clicking to normal fill

Abel032:
## Summary
This PR introduces a "fuzzy click fill" feature.
When starting a fill on a pure ink pixel, searche the surrounding pixels within a Manhattan distance of 1 for a non-pure ink pixel and starts the fill from there.

This feature may improve usability when clicking on thinline or isolated holes.

> NOTE: I'm not sure if this feature is good, there is such a situation: when the line is to too thin, it's easy to fill on outside the intended area.....

### Manhattan distance of 1
```
  y
  ^
  |
  |  .  X  .
  |  X  O  X
  |  .  X  .
  +------------> x
  
O: starting pixel (pure ink)
X: pixels at Manhattan distance 1
.: other pixels
```

### DIFF
--- a/toonz/sources/toonzlib/fill.cpp
+++ b/toonz/sources/toonzlib/fill.cpp
@@ -491,6 +491,39 @@ int getMostFrequentNeighborStyleId(TRasterCM32P ras,
 }
 
 //-----------------------------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


Abel032:
Maybe someday.

PR #6113, opened 2025-09-13
Update normal fill logic of Toonz Raster

Abel032:
# Summary
This PR fix and adjusts the fill logic to align with the newly added preference and improves code readability.
Mainly affect **the Normal Click Fill**, also affect area filler (ex. rectangle type) as they are inverted fill by calling the normal fill function.
The bugs related to area filler(such as freeHand and rectangle type) would be fixed in #6105

## Preference related

- **Define regions with both PAINT and INK**: This allows the fill result to stop at the edge of two different painted region
- **Do prevailing when refer fill**: This allows the fill result to paint over/behind the **PURE** INK lines

## Others
- Do not prevailing two different pure INK(tone = 0) while filling a row:
  - Old behavior(use No.4 style to fill)
  ```
  INK   0   0   1   1   1   2   2   3
  TONE  255 255 0   0   0   0   0   0
  PAINT 4   4   4   4   4   4   4   4
        ^start fill from this pixel
  ```
 
  - New behavior(use No.4 to fill)
  ```
  INK   0   0   1   1   1   2   2   3
  TONE  255 255 0   0   0   0   0   0
  PAINT 4   4   4   4   4   0   0   0
        ^start fill from this pixel
  ```
- fix crash if viewer trying to access the color of the max Index of palette (this happen when refer Image is caculating)
- extract adjustFillDepth() in fill.cpp
- Include reference image memory size in undo size calculation

# How to test
Here is my test cases for reference:
eb52bf2b-1bb0-4cb1-a990-2482fdf227df

### DIFF
--- a/toonz/sources/common/trop/quickput.cpp
+++ b/toonz/sources/common/trop/quickput.cpp
@@ -4457,8 +4457,8 @@ void doQuickPutCmapped(const TRaster32P &dn, const TRasterCM32P &up,
   int count = std::max({palette->getStyleCount(), TPixelCM32::getMaxInk(),
                         TPixelCM32::getMaxPaint()});
 
-  std::vector<TPixel32> paints(count, TPixel32::Red);
-  std::vector<TPixel32> inks(count, TPixel32::Red);
+  std::vector<TPixel32> paints(count+1, TPixel32::Red);
+  std::vector<TPixel32> inks(count+1, TPixel32::Red);
   if (globalColorScale != TPixel::Black)
     for (int i = 0; i < palette->getStyleCount(); i++)
       paints[i] = inks[i] = applyColorScaleCMapped(
@@ -4467,7 +4467,7 @@ void doQuickPutCmapped(const TRaster32P &dn, const TRasterCM32P &up,
     for (int i = 0; i < palette->getStyleCount(); i++)
       paints[i] = inks[i] =
           ::premultiply(palette->getStyle(i)->getAverageColor());
-
+  
   dn->lock();
   up->lock();
   TPixelCM32 *upBasePix = up->pixels();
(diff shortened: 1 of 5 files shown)


Abel032:
There is a minor issue with refer fill when both preferences are toggled off. It won’t cause a crash, but the surrounding inks may remain unclosed.
I don’t recommend using refer fill in this way (it may work in the future).
I believe most potential bugs have been fixed, so I won’t be as active for now.

RodneyBaker:
@Abel032 We are still having issues with the Mac OS checks but because of the various PRs that rely on each other I am planning to merge in order to get those caught up.   If there are any PRs you expressly do not want merged please add a note signifying that.

We'll catch up on the Mac OS builds but it looks like some of that will take time.

Abel032:
@RodneyBaker I've converted all pending PRs(those depends on another PR) to Draft (can't be merged), so the rest opening PRs can be merged without order.

Abel032:
@RodneyBaker 

Since the tests in the table are passed, I think this PR is able to be merged.
Please check if the test is enough for this PR.

If anyone has the time and interest, please take a look.

I will add similar **How to test** part to my opening PRs, so that we can see whether the PR is ready for merging or still in testing.

RodneyBaker:
@Abel032 
This PR looks fine to move foward.

As we move into October I will recommend that any PR actively being worked on (i.e. known to not be ready for merger) be pre-pended with "(WIP)" in the title as some of the older PRs still can be seen to have.  This will help in letting everyone know that the goal would be to have the PR included in the version 1.8 release but it is known to not be ready for that in its current state.  This will be even more important as we get to November and December.

As always, thanks for all you do to improve Opentoonz!

Abel032:
@RodneyBaker 
Thank you for your consistent support and patience in reviewing my PRs

I want to review all my open PRs (add the WIP flags first), and remove the WIP flag when I think it's okay to be merged, once I removed the flag I will let you know.
And I will be more thorough before raising a PR in the future in the future since it can improve the constant stability.

Abel032:
NOTE:
In the last test(2nd test of refer fill) there are un-perfect filling when dealing some semi trans lines + Aliased refer lines.

图片

PR #6116, opened 2025-09-14
Fix: Build gobject-introspection from source on macOS 13 (Tier 3)

andeon:
Homebrew no longer provides bottles for gobject‐introspection on macOS 13, which is now classified as Tier 3. See Support Tiers documentation. https://docs.brew.sh/Support-Tiers#tier-3

This patch ensures it is built from source, including the required dependencies.

It would be helpful if macOS users could test this change to verify that no dependencies are missing.

PR #6118, opened 2025-09-15
ci: configure AppVeyor to use custom Qt 5.15.2

andeon:
Updates the AppVeyor pipeline configuration to use the custom Qt 5.15.2 build with WinTab support.

Switch to Visual Studio 2022 while using the custom Qt 5.15.2 build.

ref: https://github.com/opentoonz/opentoonz/pull/6117
ref: https://github.com/opentoonz/opentoonz/pull/6116


xavier2k6:
You can change `boost_1_86_0` to `boost_1_89_0` now as it's on the latest AppVeyor 2019/2022 images.

andeon:
After testing, it only works with Visual Studio 2019. The issue might be related to the custom Qt 5.15.2 build, so it may need to be rebuilt in VS 2022.

VS 2022 not working: https://ci.appveyor.com/project/opentoonz/opentoonz/builds/52755542/artifacts

VS 2019 working: https://ci.appveyor.com/project/opentoonz/opentoonz/builds/52755943/artifacts

xavier2k6:
Seems strange as it works in GHA CI for Windows with MSVC 2022.....

BTW: `OpenCV` & `Boost` are installed via `choco` on windows workflow & `choco` is used via `AppVeyor` for `OpenCV`
* **choco** - `Boost` & `OpenCV` are currently stuck at `1.87.0 / 4.11.0` respectively.
* **AppVeyor** - can use `Boost 1.89.0` as it's on the image.

`OpenCV 4.12.0` has been released & it may be worthwhile to sync `Appveyor / GHA Windows CI` with latest of both `Boost 1.89.0` & `OpenCV 4.12.0` by moving from a `choco` install method to another if possible.

https://opencv.org/releases/
https://github.com/opencv/opencv/releases/tag/4.12.0
https://github.com/opencv/opencv/wiki/OpenCV-Change-Logs#version4120

andeon:
@xavier2k6 I'm checking the artifact from the Windows build, and it's not working either. Maybe after switching to CMake, the configuration became more strict.

@Abel032 Could you take a look and see if the crash in the artifact built with [VS 2022](https://ci.appveyor.com/project/opentoonz/opentoonz/builds/52755542/artifacts) might be a code issue? I’d really appreciate a second opinion, just to be sure. I suspect the problem might be related to using a Qt build for Visual Studio 2019 while the runner is using Visual Studio 2022, because in the 2019 i's just works.

Just letting everyone know that I updated the workflow to install Visual Studio 2019 on the GitHub runner, and with that, everything works as expected. However, I’d like to confirm my suspicion before submitting another pull request with the updated `workflow_windows.yml`, at least until Qt can be built with a newer version.

Abel032:
@andeon 
The artifact works well on my computer, I didn't see any crash except the Tape tool.
I can confirmed that crash can't be reproduced in the newest nightly build and my local build(with PRs merged).
But the crash can be reproduced in some PRs.

Way to reproduce:
1. new a scene
2. new a Toonz Raster Level
3. Switch to Tape Tool

Update:
The crash seem to only appear when using the portable stuff or the complete new install.
So you can reproduce it in nightly build, I guess it's caused by some preference setting.

andeon:
@Abel032 On my computer, every time I use any tool and click in the Viewer, it crashes, but only in the built version of Visual Studio 2022, in the artifacts from github action too.

xavier2k6:
@andeon _Please download & install_ [Latest Microsoft Visual C++ Redistributable Version](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#latest-microsoft-visual-c-redistributable-version) **Note:** [^1][^2]

[^1]: [The Redistributable version must be at least as recent as the MSVC build toolset used to build your app.](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#:~:text=The%20Redistributable%20version%20must%20be%20at%20least%20as%20recent%20as%20the%20MSVC%20build%20toolset%20used%20to%20build%20your%20app)

[^2]:  [However, the version of the Microsoft Visual C++ Redistributable installed on the machine must be the same or higher than the version of the Visual C++ toolset used to create your application.](https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#notes#:~:text=However,%20the%20version%20of%20the%20Microsoft%20Visual%20C++%20Redistributable%20installed%20on%20the%20machine%20must%20be%20the%20same%20or%20higher%20than%20the%20version%20of%20the%20Visual%20C++%20toolset%20used%20to%20create%20your%20application)

andeon:
@xavier2k6 When I said 'built', I meant the versions built via GitHub or AppVeyor  not built locally on my machine. Later, I'll send a PR for workflow_windows.yml so that other users can test it on different machines and we can come to a conclusion.

xavier2k6:
>When I said 'built', I meant the versions built via GitHub or AppVeyor not built locally on my machine.

When you are building locally on your machine.....it will build/run fine with the tools/redist. installed on machine, GitHub/AppVeyor are using latest releases/tools, hence why you may need the latest Redist.

Install it & see if you observe anymore crashes with GitHub/AppVeyor artifacts.

xavier2k6:
> After testing, it only works with Visual Studio 2019. The issue might be related to the custom Qt 5.15.2 build, so it may need to be rebuilt in VS 2022.
> 
> VS 2022 not working: https://ci.appveyor.com/project/opentoonz/opentoonz/builds/52755542/artifacts
> 
> VS 2019 working: https://ci.appveyor.com/project/opentoonz/opentoonz/builds/52755943/artifacts

I've tested the two of these artifacts & they both crash.....I think a `rebase` may be in order & a re-do of both 2019/2022.

You may see `Exception code: 0xc0000409` logged in `event viewer` on windows.

RodneyBaker:
This PR seems to be stuck in eternal MacOS (Mojave) check-land.
I think we can proceed regardless.

Merging to get (CI) Appveyor running again.

PR #6121, opened 2025-09-18
Fix Eye Button in Classic Column Header

shun-iwasawa:
This PR fixes the "Eye" (preview toggle) button in the "Classic" column header layout that had become unclickable after the introduction of PR #5795 .

This PR also removed the drag handle in timeline column header, since it now functions the same as the column name area.
image


RodneyBaker:

PR #6125, opened 2025-09-19
Fix Tape tool crash

Abel032:
## Issue
In a complete new install or using the portable stuff, create a toonz raster level and switch to tape tool, would cause the crash.

FIX:
- Set default value of autoClose Opacity to 255


RodneyBaker:
Thanks for the fix @Abel032

PR #6127, opened 2025-09-19
ci: downgrade to VS 2019 in current GitHub workflow

andeon:
Install minimal Visual Studio 2019 using Chocolatey in the CI workflow to build OpenToonz.

This change aims to investigate issues such as random crashes and other anomalies that may be related to the build environment.

@Abel032 Can you checking whether there are still other crashes that occur when following certain steps, and whether those can be reproduced using this current artifact.

### DIFF
--- a/.github/workflows/workflow_windows.yml
+++ b/.github/workflows/workflow_windows.yml
@@ -28,7 +28,7 @@ jobs:
       QT_PATH: ${{ github.workspace }}\thirdparty\qt\5.15.2_wintab\msvc2019_64  # Path to custom Qt
 
       # Visual Studio related variables
-      MSVCVERSION: "Visual Studio 17 2022"  # Visual Studio version
+      MSVCVERSION: "Visual Studio 16 2019"  # Visual Studio version
       
       # Boost-related variables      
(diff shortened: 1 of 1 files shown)


Abel032:
@andeon 
They all works okay.

RodneyBaker:
It seems this change may be required to get (CI) appveyor running again.

andeon:
Not quite, @RodneyBaker this concerns the Windows build artifact built with Visual Studio 2019 instead of 2022, which is the default on the GitHub VM.

 The AppVeyor configuration fix can be found in this PR: https://github.com/opentoonz/opentoonz/pull/6118. Applying this should get the AppVeyor (CI) builds running again.

RodneyBaker:
> The AppVeyor configuration fix can be found in this PR: https://github.com/opentoonz/opentoonz/pull/6118. Applying this should get the AppVeyor (CI) builds running again.

Thanks.  I was waiting for that PR to complete automated checks but I don't think it will.
Will proceed with merging that (as I assume we can).

andeon:
I’m going to close this PR since there don’t appear to be any more issues with VS 2022, so the downgrade is no longer necessary.

PR #6131, opened 2025-09-20
Adjust Cell Dragging behavior

Abel032:
## What this PR changes:

- Do not drag the single alone cell

- In "Always Drag Frame Cell" mode, enable to drag the drag bar to select the level range cells

### DIFF
--- a/toonz/sources/toonz/xshcellmover.cpp
+++ b/toonz/sources/toonz/xshcellmover.cpp
@@ -561,8 +561,9 @@ void LevelMoverTool::onClick(const QMouseEvent *e) {
   if (e->modifiers() & Qt::ShiftModifier ||
       // Or Dragging Frame Cell
       (r0 == r1 && c0 == c1) &&
-      getViewer()->getXsheet()->getCell(cellPosition)
-      != getViewer()->getXsheet()->getCell(row - 1, col))
+          getViewer()->getXsheet()->getCell(cellPosition) !=
+              getViewer()->getXsheet()->getCell(row - 1, col) &&
+          !getViewer()->getXsheet()->getCell(row + 1, col).isEmpty())
     m_qualifiers |= CellsMover::eInsertCells;
   if (e->modifiers() & Qt::AltModifier)
     m_qualifiers |= CellsMover::eOverwriteCells;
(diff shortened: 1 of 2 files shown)


Abel032:
This change need user's feedback before the merge.
And I may make more adjust.

Ahl76:
@Abel032  Thanks for tackling it again.  

What I noticed: 

With or without the ‘Always Drag Frame Cell’ preference

- the last cell in a row cannot  be dragged upwards
either from the cell itself (preference on) or using the control key on cell or the sidebar (preference off). But it works with the Shift key.


https://github.com/user-attachments/assets/662dfa96-b2f1-4167-b067-c86d8b986406


- In my opinion, when the preference is off, the first cell should behave as before and be able to be detached when you drag it up. This would allow more freedom for adjustments and splitting more easily. Even though there are workarounds for this.

- Draging a single cell works fine in my tests.

Another point not directly related to this PR:

Could this feature for deleting the entire drawing be enabled/disabled from a menu or by right-clicking? I understand that it can be useful in certain cases. But I think the default behaviour also had its advantages. Ideally, it would be best if we could switch between the two without going into the preferences. 


https://github.com/user-attachments/assets/c3830710-c5a0-466c-91ae-6b586f6c12e7




I think the more versatile OT is, the better.

Abel032:
@Ahl76 
I've updated the PR, thanks for the comment!

> Could this feature for deleting the entire drawing be enabled/disabled from a menu or by right-clicking?

It's the feature of Shift-Dragging in the video, I made the auto shift-dragging disabled now.

> * In my opinion, when the preference is off, the first cell should behave as before and be able to be detached when you drag it up. This would allow more freedom for adjustments and splitting more easily. Even though there are workarounds for this.

I think this feature can be restored as above.

> * the last cell in a row cannot  be dragged upwards
>       either from the cell itself (preference on) or using the control key on cell or the sidebar (preference off). But it works with the Shift key.

Please check if the problem is fixed now.

Ahl76:
@Abel032 

Regarding the last cell My previous feedback was not about removing its features, but rather that the last cell (in this PR) was not working as you had initially introduced it, using the Ctrl key or from the sidebar (in an another PR). As here below in a previous PR where it worked correctly.

https://github.com/user-attachments/assets/bc7df17f-723e-47bf-8207-b3f1a2236350

Otherwise personally im fine with it, thank you !

Abel032:
@RodneyBaker 
I've did some tests just now.
This PR looks fine to me, please review and possible merge when convenient.

RodneyBaker:
Posting intent to merge.
@Abel032 Cell dragging workflow is a very precious process to users so we appreciate how well you've been adjusting to user feedback and finding just the right remedy that satisfies all users.  That's never an easy thing to do.

I'm sure there will be additional tweaks in the future as users discover new workflows but so far so very good.

PR #6135, opened 2025-09-22
Fix export board image

shun-iwasawa:
Regarding the default filename for Export Board Image in the Board Settings, this PR correctly reflects the functionality to override the output settings' separate character when the suffix begins with an underscore ("_") or period (".").


RodneyBaker:

PR #6138, opened 2025-09-22
Fix cursor offset

Abel032:
What this PR changes:
- [Set Preference **Draw stroke at stroke end** off for default](https://github.com/opentoonz/opentoonz/commit/4f731a13b2d5ec7884d957fa1b913559fb65770b)

- [Fix 0.5 pixel's cursor offset of "Draw Cursor at Stroke End"](https://github.com/opentoonz/opentoonz/commit/d41934440852428ffc928a4a8777da6c9f8828ef)


RodneyBaker:

I'm not sure nightlies are building but we'll soon find out.

Issue #4469, opened 2022-06-20
Build failure with '-Werror=return-type'


krop:
### Description

Reproducible in current master branch. If tahoma2d devs are also reading this, it's also affected.

Building with -Werror=return-type is generally a good idea to avoid random behavior, and would definitely worth having to your continuous integration build.

The build started failing between april, 16 and may, 16.

```
[  363s] /home/abuild/rpmbuild/BUILD/opentoonz-1.6.40git.20220604T220422~aec886f/toonz/sources/stdfx/igs_rotate_blur.cpp: In member function '{anonymous}::Rotator& {anonymous}::Rotator::operator=(const {anonymous}::Rotator&)':
[  363s] /home/abuild/rpmbuild/BUILD/opentoonz-1.6.40git.20220604T220422~aec886f/toonz/sources/stdfx/igs_rotate_blur.cpp:234:39: error: no return statement in function returning non-void [-Werror=return-type]
[  363s]   234 |   Rotator& operator=(const Rotator&) {}
[  363s]       |                                       ^
[  363s]       |                                       return *this;
[  363s] cc1plus: some warnings being treated as errors
[  363s] make[2]: *** [stdfx/CMakeFiles/tnzstdfx.dir/build.make:1577: stdfx/CMakeFiles/tnzstdfx.dir/igs_rotate_blur.cpp.o] Error 1
```

### Steps to Reproduce

1. 
2. 

### Expected Behavior

Build succeeds with -Werror=return-type

### OpenToonz Version

GIT

### Operating System

Linux

### GPU

Other

krop:
The issue is still present in the master branch

Arbars:
Confirm. Bug still appearing.
Copmpiler: GCC10, GCC11, GCC12

Arbars:
Some sort of work-around:

`[  363s] /home/abuild/rpmbuild/BUILD/opentoonz-1.6.40git.20220604T220422~aec886f/toonz/sources/stdfx/igs_rotate_blur.cpp: In member function '{anonymous}::Rotator& {anonymous}::Rotator::operator=(const {anonymous}::Rotator&)':
[  363s] /home/abuild/rpmbuild/BUILD/opentoonz-1.6.40git.20220604T220422~aec886f/toonz/sources/stdfx/igs_rotate_blur.cpp:234:39: error: no return statement in function returning non-void [-Werror=return-type]
[  363s]   234 |   Rotator& operator=(const Rotator&) {}
[...]

PR #6140, opened 2025-09-23
Fix -Wreturn-type warnings in iocommand and igs_rotate_blur

andeon:
This PR fixes compiler warnings related to return types, as reported in issue #4469

- Replace the private declarations of default/copy constructors and assignment operator with `= delete` in `igs_rotate_blur.cpp` since we are using C++17.

- Add `return true` at the end of IoCmd::importLipSync in `iocommand.cpp` to ensure the function always returns a boolean, indicating successful import and exposure of lip-sync data.

These changes fix code and clean up `-Wreturn-type` warnings.

### DIFF
--- a/toonz/sources/stdfx/igs_rotate_blur.cpp
+++ b/toonz/sources/stdfx/igs_rotate_blur.cpp
@@ -228,9 +228,9 @@ public:
 
 private:
   // disable
-  Rotator();
-  Rotator(const Rotator&);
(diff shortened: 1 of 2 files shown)


RodneyBaker:
LGTM

andeon:
@RodneyBaker Could you please close the reported issue? The bug has been fixed. Thanks!

PR #6142, opened 2025-09-24
Fix CMake 4 configuration for plugin SDK samples

AlexanderShad:


AlexanderShad:
> What's ALT linux? A distro? I'm on Slackware, btw.
> 
> Edit: Yeah it's a distro, a Russian one. I have to check if my distro has the same complaints. I didn't see it complain about that part of the code yet, though.

Okay. Please take a look.
If you don't need these minor changes, I'll just patch them when package build, which is what I did.

RodneyBaker:
@AlexanderShad You are aware that the changes you've made only apply to plugins right?

I'm a little concerned that a fix here might be overwritten in the future when the plugins themselves are updated.

I'll approve the run of workflows as that might allow the github action builds to work but I'll tag @shun-iwasawa as a reviewer as he might have some insights into the best way forward.

andeon:
I noticed that both my recent [PR](https://github.com/opentoonz/opentoonz/pull/6140) and this one modify `igs_rotate_blur.cpp`, addressing similar issues. I'm using a C++11 (and later) approach since OpenToonz is moving to/updating to C++17 to align with the upcoming Qt6 migration.

Regarding these changes, on my distro (KDE Neon) there are no issues with cmake_minimum_required in these `CMakeLists.txt`. We recently updated the minimum required CMake version to 3.10 after recent tests.

AlexanderShad:
> Regarding these changes, on my distro (KDE Neon) there are no issues with cmake_minimum_required in these `CMakeLists.txt`. We recently updated the minimum required CMake version to 3.10 after recent tests.

ALT Linux uses gcc14 and cmake 4.0.3 for compilation.

During compilation, I got an error with these `CMakeLists.txt` files. So, I added the lines` cmake_minimum_required` there.

andeon:
If these changes are applied, it is recommended to set cmake_minimum_required(VERSION 3.10) in these subdirectories to remain consistent with the top-level CMakeLists.txt and avoid potential regressions, since CMake behavior can sometimes be inconsistent.

AlexanderShad:
Changed to 3.10

RodneyBaker:
@AlexanderShad  Without resolution of current conflicts this PR cannot move forward.

AlexanderShad:
> @AlexanderShad Without resolution of current conflicts this PR cannot move forward.

fixed the conflict

RodneyBaker:
@AlexanderShad I've changed the title of this PR as the scope has changed.
Recommend rebasing this PR to get it in sync with current source and this can be merged.
The MacOS check should also pass at that point as it is unrelated to this PR.

AlexanderShad:
Ok

RodneyBaker:
Thanks @AlexanderShad 
Especially for your patience.
LGTM

PR #6144, opened 2025-09-24
Restore tiio_tif.cpp to version from commit 67b4e98 due to TIFF loading issue

andeon:
This PR restores tiio_tif.cpp to its original version before the recent changes that introduced a regression when loading TIFF files.

After the previous PR, some TIFF images could no longer be loaded correctly, especially 16bpc files.

If possible, please share the original TIFF file used for testing, so we can confirm the fix and validate future changes.

### DIFF
--- a/toonz/sources/image/tif/tiio_tif.cpp
+++ b/toonz/sources/image/tif/tiio_tif.cpp
@@ -23,15 +23,6 @@ extern "C" {
 #include "tiffio.h"
 }
 
-// Fix for libtiff 4.0+ compatibility: Map _64 variants to standard functions (preserves 64-bit support via internal offsets)
-#if !defined(TIFFReadRGBATile_64)
-#define TIFFReadRGBATile_64 TIFFReadRGBATile
-#endif
-#if !defined(TIFFReadRGBAStrip_64)
-#define TIFFReadRGBAStrip_64 TIFFReadRGBAStrip
-#endif
-#include <cstdint>  // adde: for uint32_t/uint64_t (resolve casts)
-
 #include "tiio_tif.h"
 
 #ifdef _MSC_VER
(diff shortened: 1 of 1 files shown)


RodneyBaker:
This problem is limited to Mac OS?
I can't reproduce the problem rendering out any TIF images from Opentoonz and then bringing them back in to OT (in 8, 16 and 32 bit) on Windows.

Tested with a few different nightlies as well as github artifacts.

andeon:
@RodneyBaker I believe the issue occurs when loading the image into OpenToonz. I’ve asked @shun-iwasawa to send the image so we can properly document the issue.

RodneyBaker:
@andeon Sorry.  To clarify...
The source of my TIF's was rendering out of Opentoonz but I was both loading and importing the TIFs and didn't have any issues.  I need to test more but that was my initial result.

I agree having a TIF image that is known to cause the problem is important.
Hopefully @shun-iwasawa can provide one.

shun-iwasawa:
This is the image file I used when reporting the issue:
[16bpc_test.zip](https://github.com/user-attachments/files/22528584/16bpc_test.zip)
- 512x512 pixels
- 16bpc RGB channels (48bpp)
- Compression : None
- Pixel Order : Interleaved(RGBRGB)
- Byte Order : Macintosh

Thank you for your prompt response!

RodneyBaker:
That image does indeed fail.
Apparently it is the 48bits per pixel aspect that is key to this mystery.
TIF images rendered out of Opentoonz weren't failing as I hadn't adjusted the 8/16/32 bit TIFs to have higher bits per pixel.
Ramp that option up and TIF image coloring fails with images rendered out of Opentoonz.

Edit:  @shun-iwasawa's image (as well as others I'm testing) load correctly with this PR.
A question:  How does restoring tiio_tif.cpp effect other recent changes? 

cc:  @Abel032

Abel032:
@RodneyBaker I think there are no changes related to tiff IO recently, so it won't not be a problem since it's build 
Just make sure the build and the original function works.

RodneyBaker:
Thank you @andeon for the revert/restore.
LGTM

@shun-iwasawa Please let us know if you see anything still needed.

PR #6147, opened 2025-09-26
Revert File Browser behavior

shun-iwasawa:
This PR reverts  the left single-click to move folders in the File Browser according to the discussion which follows [this comment](https://github.com/opentoonz/opentoonz/pull/5760#issuecomment-3311011084).

This PR also enabled to delete (move to trash) and to rename folder via context menu.

Please note that if commands like moving or deleting multiple selected folders become necessary and are implemented in the future, similar to Explorer, behaviors like single-click for selecting items and double-click for moving folders may be re-introduced.


RodneyBaker:
Thank you @shun-iwasawa for the adjustment.
Much appreciated!

PR #6150, opened 2025-09-26
New Fx: Smoother Fx Iwa

shun-iwasawa:
This PR will introduce a new fx; Smoother Fx Iwa which is to apply antialiasing on the binalized image.

To apply anti-aliasing to level images, OT already has an option “Level Settings -> Add Antialiasing”.
However, when you want to make part of a level image semi-transparent (which is often the case for character shadows, for example), you need to use RGB Key Fx on the binary state image without anti-aliasing to extract the area.

This Fx is used to apply anti-aliasing to a binary level image AFTER performing necessary processing such as RGB Key.

The key feature of this effect is that it acquires the source level image at a pixel-perfect (“non-resampled”) size, performs anti-aliasing , and then resamples it as needed.

#### Source image (binarized TGA)
image

####  1. With conventional method
image
image

####  2. With Smoother Fx Iwa 
image
image

### PARAMETERS
- **Threshold** : The difference of the color channel value (0-255) recognized as the boundary for applying anti-aliasing. At the default value (10), it behaves the same as the Add Antialiasing option in Level Settings.
- **Softness** : Behaves the same as the Antialias Softness in the Level Settings.

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1628,6 +1628,10 @@
   <item>"STD_iwa_MotionFlowFx.normalizeType" "Normalize"	</item>
   <item>"STD_iwa_MotionFlowFx.normalizeRange" "Maximum Length"	</item>
 
+  <item>"STD_iwa_SmootherFx"	"Smoother Iwa"	</item>
+  <item>"STD_iwa_SmootherFx.threshold"	"Threshold"	</item>
+  <item>"STD_iwa_SmootherFx.softness"	"Softness"	</item>
+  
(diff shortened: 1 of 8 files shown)


konero:
It works perfectly, thank you! ❤️

gab3d:
That's great!
Thanks!! :)

RodneyBaker:
I see no reason to delay getting this into a nightly release so posting intent to merge.

RodneyBaker:
Thank you @shun-iwasawa for the new FX!

PR #6151, opened 2025-09-27
Fix dummy ink pixel (Tone =255) ink texture crash

Abel032:
The crash is reported by user.

I can't reproduce the crash since it disappear very quickly if you draw something on the level,
but here is the scene that caused the crash:
[one-pixel-crash.zip](https://github.com/user-attachments/files/22576246/one-pixel-crash.zip)

Click on preview to trigger the crash, and the reason seems related to this single pixel:
图片

### DIFF
--- a/toonz/sources/common/trop/tropcm.cpp
+++ b/toonz/sources/common/trop/tropcm.cpp
@@ -367,7 +367,8 @@ TRasterP putSingleInkInRasterGR8(TRasterCM32P &rasIn, int inkId) {
     if (rasOut) pixOut = rasOut->pixels(y);
 
     while (pixIn < endPix) {
-      if (pixIn->getInk() == inkId) {
+      if (pixIn->getTone() != TPixelCM32::getMaxTone() &&
+          pixIn->getInk() == inkId) {
         assert(TPixelCM32::getMaxTone() == 0xff);
         if (!rasOut) {
           rasOut = TRasterGR8P(rasIn->getLx(), rasIn->getLy());
@@ -399,7 +400,8 @@ TRasterP putSingleInkInRasterRGBM(TRasterCM32P &rasIn, int inkId) {
     if (rasOut) pixOut = rasOut->pixels(y);
 
     while (pixIn < endPix) {
-      if (pixIn->getInk() == inkId) {
+      if (pixIn->getTone() != TPixelCM32::getMaxTone() &&
+          pixIn->getInk() == inkId) {
         assert(TPixelCM32::getMaxTone() == 0xff);
         if (!rasOut) {
           rasOut = TRaster32P(rasIn->getLx(), rasIn->getLy());
--- a/toonz/sources/toonzlib/imagestyles.cpp
+++ b/toonz/sources/toonzlib/imagestyles.cpp
@@ -967,7 +967,9 @@ bool TTextureStyle::doCompute(const Params &params) const {
     if (m_params.m_type == TTextureParams::AUTOMATIC) {
       TRect bBox;
       TRop::computeBBox(r, bBox);
-
+      
+      // Just in case of empty raster
+      if (bBox.isEmpty()) bBox = TRect(0,0,1,1);
       r = r->extract(bBox);
       p += r->getCenter() - computeCentroid(r);
     }


Abel032:
@shun-iwasawa 
Could you please check what's happening while rendering here?
I can't reproduce the crash by drawing the same pixel.

RodneyBaker:
That's very strange.
I get the crash with that project also.
Drawing over the pixel fixes the problem.
Undoing that... crashes OT.

I am not getting the crash with this PR.

andeon:
Not sure if this is useful, but here's the `backtrace` from Linux, as it doesn't appear on Windows. However, the PR fixes the issue.


```
==== Backtrace ==== 
1> ./../lib/opentoonz/libtoonzlib.so(_ZNK13TTextureStyle9doComputeERKN14TRasterStyleFx6ParamsE+0x409) [0x758c3cd58e29] 
2> ./../lib/opentoonz/libtoonzlib.so(_ZThn176_NK13TTextureStyle7computeERKN14TRasterStyleFx6ParamsE+0xd) [0x758c3cd595dd] 
3> ./../lib/opentoonz/libtnzcore.so(_Z11renderRas32RK5TTileS1_14TSmartPointerTI8TPaletteE+0x1633) [0x758c3c66c173] 
4> ./../lib/opentoonz/libtnzcore.so(+0x26a0a8) [0x758c3c66a0a8] 
5> ./../lib/opentoonz/libtnzcore.so(_ZN4TRop7convertERK5TTileS2_14TSmartPointerTI8TPaletteEbb+0x119) [0x758c3c6699e9] 
6> ./../lib/opentoonz/libtoonzlib.so(_ZN14TLevelColumnFx9doComputeER5TTiledRK15TRenderSettings+0x2684) [0x758c3ce461f4] 
7> ./../lib/opentoonz/libtnzbase.so(_ZN9TRasterFx7computeER5TTiledRK15TRenderSettings+0x1bc8) [0x758c3d152b38] 
8> ./../lib/opentoonz/libtnzbase.so(_ZN11TGeometryFx9doComputeER5TTiledRK15TRenderSettings+0x5d8) [0x758c3d169fa8] 
9> ./../lib/opentoonz/libtnzbase.so(_ZN9TRasterFx7computeER5TTiledRK15TRenderSettings+0x1bc8) [0x758c3d152b38] 
10> ./../lib/opentoonz/libtnzbase.so(_ZN11TGeometryFx9doComputeER5TTiledRK15TRenderSettings+0x5d8) [0x758c3d169fa8] 
11> ./../lib/opentoonz/libtnzbase.so(_ZN9TRasterFx7computeER5TTiledRK15TRenderSettings+0x1bc8) [0x758c3d152b38] 
12> ./../lib/opentoonz/libtnzbase.so(_ZN9TRasterFx18allocateAndComputeER5TTileRK7TPointTIdERK11TDimensionTIiE14TSmartPointerTI7TRasterEdRK15TRenderSettings+0x7a7) [0x758c3d150e07] 
13> ./../lib/opentoonz/libtnzbase.so(_ZN19TImageCombinationFx9doComputeER5TTiledRK15TRenderSettings+0x445) [0x758c3d11b0c5] 
14> ./../lib/opentoonz/libtnzbase.so(_ZN9TRasterFx7computeER5TTiledRK15TRenderSettings+0x1bc8) [0x758c3d152b38] 
15> ./../lib/opentoonz/libtnzbase.so(_ZN10RenderTask3runEv+0x3e1) [0x758c3d15e231] 
16> ./../lib/opentoonz/libtnzcore.so(_ZN7TThread6Worker3runEv+0x108) [0x758c3c4c2628] 
17> /lib/x86_64-linux-gnu/libQt5Core.so.5(+0xdb674) [0x758c3d8db674] 
18> /lib/x86_64-linux-gnu/libc.so.6(+0x9caa4) [0x758c3a29caa4] 
19> /lib/x86_64-linux-gnu/libc.so.6(+0x129c6c) [0x758c3a329c6c] 
==== End ====
```

Bracket-I-I:
Yeah I can also confirm the crash on Linux.

RodneyBaker:
I'm posting intent to merge this fix although...

It would be great to have more insight into why the problem exists and the nature of this fix from @shun-iwasawa or others interested.

RodneyBaker:
LGTM

Other thoughts:
The root cause of dummy (tone=255) ink pixels were treated as coverage when generating the per-ink texture mask.  This produced an empty/invalid region that crashed the texture style’s bounding box path.

Regarding the Fix, it ignores tone=255 pixels when building masks and creates a fail safe 1×1 bounding box if/when the bounding box is empty. This prevents the problem from happening which in turn stops the crashes.

Added: We probably do want to watch to make sure this failing condition is not reintroduced somewhere down the line.  I'm not sure best how to approach that.

PR #6154, opened 2025-09-28
Add vector level support to refer fill

Abel032:
This PR adds support for vector levels to the existing refer fill feature of ToonzRaster fill tool.


Abel032:
LGTM

RodneyBaker:
Very nice @Abel032 
I ran through your list of tests and more.

I'm still not sure if I prefer Preference>Visualization>Visualize Vector with Anti Aliasing On or Off but I think On.  I need to find some good reasons to use both.

LGTM

PR #6157, opened 2025-09-30
Fix export standalone scene

Abel032:
NOTE: This PR also fixed the path lost of name field(the lineEdit at the bottom of file browser popup) after clicking a folder item.
This can tested in such a filebrowser:
图片
You can get the file browser with the title `Choose Folder` if OT ask to select a folder but not a file.
For example, click button right of the path field when exporting a scene.

## How to test
Open a scene by these method:
- new a scene (and save / without save)
- load a scene from a project
- load a scene that standalone

Then try `File>Export>Export As Standalone Scene`, and load the scene to see if every level is properly loaded.

### DIFF
--- a/toonz/sources/toonz/exportscenepopup.cpp
+++ b/toonz/sources/toonz/exportscenepopup.cpp
@@ -89,18 +89,18 @@ int collectAssets(ToonzScene &scene) {
 
 //------------------------------------------------------------------------
 
-void decodeLevelsPath(ToonzScene& scene) {
-    std::vector<TXshLevel*> levels;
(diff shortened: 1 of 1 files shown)


Abel032:
I'll test it in afternoon(UTC+8)

Abel032:
Also fixed the issue that the main window hides after exporting.

Abel032:
LGTM

RodneyBaker:
I had noticed something wrong recently with exporting standalone scene which caused me to resort to the older project export option above.  I hadn't had the chance to investigate further so I'm glad others did!

LGTM

PR #6158, opened 2025-09-30
fill holes: adjust the size to not limited(but still suggest a range of 1~25)

Abel032:
So that user can type the value more than 25:
图片

## How to test
Try to type a value that above 25, but it might be slow to fill.


RodneyBaker:
Thank you @Abel032 for the instant update and the useful way in which you implemented it.
And best of all... even with larger sizes... filling holes works great!
LGTM

PR #6159, opened 2025-09-30
[WIP] Fix scene assets importing

Abel032:
# Summery

There are 3 ways to transport scene's assets from a project/standalone scene to another project:
- Import a scene from outside
- Exporting a inside scene to an outside project
- Copy the scene file to the new project and reassign levels' path in the scene

The exporting function works: It copy the scene and assets to new project, and replace the absolute path with the alias path.
But importing function is broken for some reason: It only copy the scene file, won't copy the assets and they all leave absolute path.

This PR added back the feature of importing assets when importing the scene, but since importing and exporting use shared logic/function, but of them need to test.

# How to test
Test all feature related to scene loading and importing.
- Try to use the command `File>Load Scene` to load an outside scene and select to import it
- Try to `Load Scene` from OT's file browser
- Try to `Import Scene` from OT's file browser
- Try to `Export Scene` to other projects

# Something need to consider
After this recovery the `collectAssets()` in `exportscenepopup.cpp` is not necessary for exporting scene to project any more, because `loadSceneToProject()` will import all the assets now, but it won't affect the result as it's just doing the collecting/importing again.(collect equals import in this case)
I think it's possible to remove `collectAssets()` if export to project in export scene popup, and rename `loadSceneToProject()` to `importSceneToProject()`.
These adjustment must wait for #6157, but do not prevent the merge of this PR, since they are just refactoring.
I'll make the adjustment if #6157 is merged first.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1836,7 +1836,7 @@ bool IoCmd::loadScene(ToonzScene &scene, const TFilePath &scenePath,
   if (scene.getProject()->getProjectPath() !=
       currentProject->getProjectPath()) {
     ResourceImportDialog resourceLoader;
-    // resourceLoader.setImportEnabled(true);
+    resourceLoader.setImportEnabled(import);
     ResourceImporter importer(&scene, currentProject, resourceLoader);
     SceneResources resources(&scene, scene.getXsheet());
     resources.accept(&importer);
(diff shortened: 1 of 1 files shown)


Abel032:
- [ ] The levels can't be properly loaded when loading a scene and select to import all assets, reverting the scene can resolve
- [ ] Deal the situation that meet scene with the same name when importing(Currently OT just rename it to a new name)

Abel032:
@RodneyBaker 

Thank you for your constant support.

I may need to keep this PR as WIP for a few months, as I won't be able to contribute frequently due to personal reasons recently.
My other opening PRs are ready for review, though.

Looking forward to the official 1.8 release!

PR #6161, opened 2025-10-01
Fix(referFill): handle zero DPI for vector levels and add image DPI support

Abel032:
Also refactored for performance and accuracy.

## How to test
- Try to refer fill with `Preferences>Drawing>New Levels Default to the Current Camera Size` ON/OFF
- Try to refer fill with Animate Tool(All kinds of transformation)
- Try to refer fill with different dpi settings (adjust them in `LevelSettingsPopup`)
- Try to refer fill with `TLV Bounds Fill` ON/OFF
- Test if other old refer fill features is not broken

(The area filler can be skipped since I only edited the code related to how to generate refer image)


Abel032:
LGTM

PR #6162, opened 2025-10-02
Fix: Use sig_atomic_t for signal-safe shutdown flag

### LINKED ISSUE #44: TFarm:  Fix signal handler (Code Typo?)

andeon:
This change makes the shutdown flag safe to modify from a signal handler (SIGUSR1), preventing undefined behavior due to concurrent access.

The variable was renamed from `Sthutdown` to `shutdownRequested` for clarity and consistency.

Additionally, the usage of `using namespace std;` was removed to avoid namespace pollution and improve code clarity.

Fixes #44

### DIFF
--- a/toonz/sources/toonzfarm/tfarm/ttcpipserver.cpp
+++ b/toonz/sources/toonzfarm/tfarm/ttcpipserver.cpp
@@ -2,12 +2,12 @@
 
 #include "ttcpip.h"
 #include "tconvert.h"
+#include <csignal> // for sig_atomic_t
 
 #ifdef _WIN32
 #include <winsock2.h>
 #else
-#include <errno.h> /* obligatory includes */
-#include <signal.h>
+#include <errno.h>
 #include <stdio.h>
 #include <unistd.h>
 #include <sys/types.h>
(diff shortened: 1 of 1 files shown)


RodneyBaker:
There should be some form of celebration in closing that... the oldest report still open.
Congrats @andeon for officially kicking off the 10th anniversary celebration ramp up!
(Right on cue 1 Oct 2025)

Perhaps most importantly, this fix appears to address a nasty issue with potential lockups and shutdowns.  It may not fix all of those oddities but if even a few that's great.

LGTM

@elfring Almost sad to see this one go.  ;)

PR #6164, opened 2025-10-02
Hold Alt to toggle Region definition preference in Fill Tool

Abel032:
When left key down with `alt` pressed switch this preference from ON to OFF, or from OFF to ON, and restore it after the filling behavior.
图片


RodneyBaker:
Thanks @Abel032 
LGTM

PR #6176, opened 2025-10-08
fix: remove duplicate and unreachable 'stylepages' case

andeon:
Resolved logical error where a duplicate "stylepages" condition created an unreachable code block. The second "stylepages" condition (line 867) never executes because the first one (line 784) always catches the tag first. Additionally, the unreachable code contained a key counter that never incremented, which could cause shortcut assignment issues.

Corrected exception message to expect `<page>` tag instead of `<stylepage>` for better clarity and easier debugging.

Changes:
- Remove duplicate `stylepages` condition entirely
- Fixed exception message to expect `<page>` tag


RodneyBaker:

PR #6178, opened 2025-10-08
ci: Remove the 'brew unlink qt' step from the macOS workflow

andeon:
Since the build process already uses explicit paths via `brew --prefix qt@5`, unlinking the default qt formula is unnecessary and could lead to errors.

Homebrew installs `qt@5` as keg-only, and all references in the workflow explicitly use its path, so `brew unlink qt` is redundant.

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -3,11 +3,11 @@ name: MacOS Build
 on:
   push:
     paths:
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thanks @andeon for the update.
Glad to have you massaging the Mac OS side of Opentoonz builds.
LGTM

PR #6181, opened 2025-10-10
Update Fill Holes algorithm

Abel032:
## What This PR Changes
- Updated neighbor weighting: Cardinal Ink (tone+1), Cardinal Paint (tone+2), Diagonal Paint (tone+3)
- Pixels with tone < 64 are now treated as area boundaries

| Neighbor Type | Pixel Type | Weight Formula | Rationale |
|---------------|------------|----------------|-----------|
| Cardinal      | Ink        | tone + 1       | Direct neighbors with basic priority |
| Cardinal      | Paint      | tone + 2       | Direct neighbors with higher priority |
| Diagonal      | Paint      | tone + 3       | Corner neighbors (excluding pure ink) |

## How to Test
Test with various images.

### DIFF
--- a/toonz/sources/toonzlib/fill.cpp
+++ b/toonz/sources/toonzlib/fill.cpp
@@ -446,7 +446,7 @@ bool scanTransparentRegion(TRasterCM32P ras, int x, int y,
   if (visited[y * lx + x]) return true;
 
   TPixelCM32 *pix = ras->pixels(y) + x;
-  if (pix->getPaint() != 0 || pix->isPureInk()) return true;
+  if (pix->getPaint() != 0 || pix->getTone() < 64) return true;
   if (points.size() + 1 > maxSize) return false;
 
   visited[y * lx + x] = true;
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I was having a bit of a hard time finding good images to test with as the Fill process in current Opentoonz is working so well.  I ended up using Eraser Tool to intentionally create holes in images and see how well they filled with the 'Fill Holes' feature.

Result:  Works great.  Exceeded my expectations in many ways.

Note for future improvement:  It would be great to have Fill Holes be shortcut key assignable as that would make quick work of filling holes without resorting to use of menu.

Posting intent to merge but still testing to see if I can better understand the extend of where filling holes will reliably fail.  The ability to enter higher values than the original recommended value is a secret weapon here.

RodneyBaker:
I do notice that often after leaving the xsheet where the toonz raster level is selected and clicking through the menu to activate Fill Holes the focus is lost so that the error message appears telling the user 'no toonz raster level is selected'.  It would be nice if the process could maintain focus... or reach back to recall the last toonz raster level selected.

Abel032:
@RodneyBaker The same issue may also happen on other dialogs that requires a selection, for example the export level windows.
 I think fill holes command didn't try to Get the selection before showing the window, that caused the lost of selection.

We can address this when improving the selection logic for dialogs.
I won't be able to fix it immediately,but if others are interested, feel free to fix.

RodneyBaker:
I agree that element is likely out of the scope of this PR.
I have a few more tests to run and I don't anticipate any issues with this PR.

Posting intent to merge.

RodneyBaker:
Thank you @Abel032 for continuing to refine the various algorithms from your recent updates.
LGTM

PR #6182, opened 2025-10-10
Fix SVG import crash with non-ASCII file paths on Windows

Abel032:
Here are some non-ASCII characters as examples:
- 中文
- 日本語


RodneyBaker:
@Abel032 I didn't get a crash with the SVGs I tested but I did confirm Opentoonz gave the error refusing to load or import SVGs with non-ASCII characters.

I also confirm this PR fixes the problem.

Posting intent to merge.

PR #6183, opened 2025-10-10
Feat(next/prev Drawing): Add support of level Editing mode

Abel032:
To test this, try using `Next Drawing` and `Prev Drawing` Command in `Level Editing Mode` and `Scene Editing Mode`


RodneyBaker:
@Abel032 If I understand this change... with this PR... the user now will not be shifted out of Level Editing mode to the Scene Editing mode when using Previous/Next Drawing (or period/comma).

That's what I'm seeing which is very nice.
If I'm missing something else I'm not seeing it in the code.

Abel032:
@RodneyBaker Yes, that's right

RodneyBaker:
Thanks for the clarification and...
Thanks for this subtle yet very useful addition!
LGTM

PR #6186, opened 2025-10-11
Revise UI terms for fill region and gap close options

Abel032:
| Option/Preference | Before | After |
| ----- | ----- | ----- |
| Tape Tool Option | Ignore AutoPaint Inks | Ignore AutoPaint Lines |
| Tools Preference | Define filling region with both PAINT and INK | Define Filling Region Using both Lines and Areas |
| Tools Preference | Paint under/over INK in refer fill | Paint Under Lines in Refer Fill and Gap Close |


Abel032:
Just updated the capitalization.

RodneyBaker:
Thanks @Abel032 for the revision of terms.
I think that will go a long way in helping users get comfortable with the options.

I was going to keep this open to possibly collect additional UI term adjustments but new PRs can can be added for those and these help solidify the features you've worked hard to add.

In other words.
LGTM

PR #6187, opened 2025-10-11
Fix memory leaks by replacing 'delete' with 'delete[]

andeon:
This PR fixes memory leaks caused by incorrect use of `delete` instead of `delete[]` when deallocating dynamically allocated arrays.

Changes:
- Corrected array deallocation in `splitMatrix()` to use `delete[]`
- Fixed proper deletion of all rows in `solveSystems()` using `delete[]`

### DIFF
--- a/toonz/sources/stdfx/pins.cpp
+++ b/toonz/sources/stdfx/pins.cpp
@@ -8,7 +8,7 @@
 
 #include <cmath>
 
-//------------------------------------------------------------------------------
+//----------------------------------------------------------------------------
 
 namespace {
 
@@ -54,7 +54,7 @@ bool lineIntersection(const TPointD &P, const TPointD &R, const TPointD &Q,
   }
 #endif
 
-// ---------------------------------------- --------------------------------
+//----------------------------------------------------------------------------
 
 void subCompute(TRasterFxPort &m_input, TTile &tile, double frame,
                 const TRenderSettings &ri, TPointD p00, TPointD p01,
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I see some additional changes besides those mentioned.
These include translation of comments from Italian to English, as well as mostly small, localized fixes with cleanup.

A change of interest to me is at the end of the file where you've changed the assert to standard absolute.  Instead of only guarding against the case where den is exactly zero, it now guards against values that are very close to zero.  Nice.

This mostly effects FX pins so I'm curious where else changes like this might help to update the source.

Thanks for the fix.  All looks good from here.
Posting intent to merge

andeon:
When I'm reviewing the code, I try to translate the comments into English and clean up anything that’s not being used. I haven’t done this very often before, but now it’s easier with the tools we have available today.

RodneyBaker:
> When I'm reviewing the code, I try to translate the comments into English and clean up anything that’s not being used. I haven’t done this very often before, but now it’s easier with the tools we have available today.

I'm glad you are doing this as it helps differentiate the code being from the older Toonz years (where comments are in Italian), the middle Studio Ghibli Edition years (largely Japanese, and the current English commentary.  At some point the comments will all be English and we'll lose that particularly insightful measure so I'm enjoying the differences while we can.  Github picks up where that measure leaves off so all is still good in that respect.

As for this PR... thanks for the fix.
LGTM

PR #6193, opened 2025-10-13
Optimize gap close related algorithm

Abel032:
This PR remove the gap lines inside or outside of filling region when filling with area filler(`Rectangle`, `FreeHand`type and so on):
图片

Also fix the gap close lines when meeting some areas that have the combination of line and areas:
图片

## How to test

To test the first one, try to fill a region that contains multiple regions, using the `FreeHand` or `Rectangle` type of `Fill Tool`, with `Paint Under Lines in Refer Fill and Gap Close` ON or OFF.
Here is an example:
图片
The result should not have any residue:
图片

And for the second one please test it with various images, especially those already painted images.

### DIFF
--- a/toonz/sources/toonzlib/autoclose.cpp
+++ b/toonz/sources/toonzlib/autoclose.cpp
@@ -250,6 +250,28 @@ const bool isSmallEnclosedRegion(const TRasterCM32P &ras, int x, int y,
 
   return visited.size() <= maxSize;
 }
+
+bool doSegmentsIntersect(int x1, int y1, int x2, int y2, int x3, int y3, int x4,
+                         int y4) {
+  auto cross = [](int x1, int y1, int x2, int y2) -> int {
+    return x1 * y2 - y1 * x2;
+  };
+
+  int dx1 = x2 - x1;
+  int dy1 = y2 - y1;
+  int dx2 = x4 - x3;
+  int dy2 = y4 - y3;
+
+  int d1 = cross(dx2, dy2, x1 - x3, y1 - y3);
+  int d2 = cross(dx2, dy2, x2 - x3, y2 - y3);
+  int d3 = cross(dx1, dy1, x3 - x1, y3 - y1);
+  int d4 = cross(dx1, dy1, x4 - x1, y4 - y1);
+
+  if (d1 == 0 || d2 == 0 || d3 == 0 || d4 == 0) return false;
+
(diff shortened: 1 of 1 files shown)


Abel032:
@RodneyBaker 
Due to personal reasons, I wasn't able to test this PR as thoroughly as I would have liked. I'd appreciate it if you could give it a extra review before merging. Thanks!
If any issues arise after it's merged, please feel free to contact me directly, and I'll revert the change.

Abel032:
Pushed to fix a boundary problem.

RodneyBaker:
I haven't done exhaustive testing with this PR but I do think it important the two fixes here be merged to the nightlies which appear to resolve those specific problems.

Posting intent to merge but still doing a few more tests.

RodneyBaker:
@Abel032 
The fixes appear to work as expected.  Thanks!

The part that does not appear to work quite as well as before is that the gaps seem to need to be smaller now than before (with Max value set).  I need to test older release versus this PR but my sense is that shapes that would fill before now do not when the gap is more than a few pixels wide.

As for this PR...
LGTM

Abel032:
> The part that does not appear to work quite as well as before is that the gaps seem to need to be smaller now than before (with Max value set).  I need to test older release versus this PR but my sense is that shapes that would fill before now do not when the gap is more than a few pixels wide.

Just remind me if you find out the difference.
I will do some comparisons.

RodneyBaker:
@Abel032 
I am seeing something in the nightly now that this PR is included that I didn't see when testing the PR.  I will do a clean install and see if it still persists.

Namely, when using the Rectangle fill with Gap Close the initial rectangle drawn is freezing without filling the shape.  A second drawing of the rectangle then completely the task.

Very early result but wanted to let you know what I'm seeing of this PR now in the nightly.
Hopefully it will go away upon a clean install.

Update:  Good news.  After a clean install I'm not seeing that freeze upon first selecting the shape.  

It is still hard to fill drawings that I'm pretty sure we could fill before. THAT might relate to antialiased lines.

Abel032:
For the freezing, I guess it might related to refer fill since the refer fill and gap close shares the same logic(all considered as refer plus image onto filing image).
As for the difficulty of filling shape..Not sure but investigating

Abel032:
图片


I do observed some image like this might showing possibility to fill but can't be filled with `normal fill`/`rectangle fill`.

Clue: those pixels that has paint under line

PR #6197, opened 2025-10-15
Fix(AreaFiller): Do not paint autoPaint lines when No.0 Style activated

Abel032:


RodneyBaker:
This one is self evident.
LGTM

Issue #5471, opened 2024-05-27
Brush size does not work well in vector levels


Yatori:
### Description

Hello!
The size slider is not accurate, and when I type the values it does not work with values with “.” like 0.3, 2.5, etc.
with pressure off or on the same thing happens.

When I try to manually type in a value, the size always stays with the last value that was set in the slider, and with the slider it is not possible to use precise values,

### Steps to Reproduce

1. On a vector level, without using the slider, type in the brush size, with a number like 0.3, 4.5, etc.
2. To notice it do it with the pressure off.
3. Now draw and you will see that the size does not change.

### Expected Behavior

I hope to be able to enter values with “.”, such as 0.2, 0.3, 2.5, etc, to have more control over the stroke size with the brush.

### Screenshots, Video & Crash Logs

Just in case I tried the same thing in the latest version of Tahoma2d 1.4.2 and it works fine.

Error in opentoonz

https://github.com/opentoonz/opentoonz/assets/10714612/f21c869b-fc1d-45d6-b8bd-38004974b3ce

Works in tahoma

https://github.com/opentoonz/opentoonz/assets/10714612/563f60b1-41b2-4011-97f7-7de2355bb23e

### OpenToonz Version

1.7

### OpenToonz Version Information

Opentoonz 1.7.1

### Operating System

Windows

### GPU

Intel

### Graphics Tablet

Huion

RattusaurusRex:
I can't replicate this issue.  Entering decimals works fine -- the slider updates and the line width is accurate when drawn.  

Maybe you can try a nightly version to see if this is fixed for you as well?

Yatori:
@RattusaurusRex Hi, thanks for your reply,  The last nightly build ([nightly build 2024-05-01](https://github.com/opentoonz/opentoonz/releases/tag/nightly)) gives error and fails to start,
![Imagen de WhatsApp 2024-06-04 a las 17 34 02_47a4671e](https://github.com/opentoonz/opentoonz/assets/10714612/d8f03e36-4f82-4717-83f5-44390536a0eb)

The problem with decimal numbers is not only in my windows machine, I also tested the same stable version 1.7.1 in linux and the same thing happens.

Are you sure they work for you? I check to see if the thickness changes with the selection tool or by changing the size to distant values such as 1 to 9.5, the number changes in the input text box, but not on the stroke.

papio-pl:
[...]

PR #6198, opened 2025-10-15
Fix brush size input for non-English locales

andeon:
This PR fixes the issue reported in #5471 where users with non-English system locales could not input decimal values in brush size fields. Neither dot (.) nor comma (,) worked to set decimal values, and the brush size didn’t update when typing these values.

English users are encouraged to test and verify that no unintended side effects occur due to this change.

Now, decimal values like 0.3 or 0,3 can be entered regardless of system language or locale. This is especially helpful in locales where the numeric keypad uses a comma as the decimal separator (e.g., pt-BR, fr-FR, es-ES), improving usability and input speed.

I replaced `m_slider->repaint();` with `m_slider->update();`. Apparently, `update()` had issues in Qt versions prior to 5.15, but these have since been fixed. It’s now the preferred approach in recent Qt versions to ensure the UI updates correctly without impacting performance.

Changes:

- Allow both comma and dot as decimal separators
- Ensure the brush size field updates correctly when typing decimal values
- Improve code quality using modern C++ features
- Standardize UI layout margins for consistency

Ref: #5471

**Ps:** I noticed that the Smooth field in the `vector level` rounds values when typing, which is unrelated to this PR. This may also happen with other fields and will need to be reviewed in the future.

### DIFF
--- a/toonz/sources/include/toonzqt/doublefield.h
+++ b/toonz/sources/include/toonzqt/doublefield.h
@@ -60,6 +60,8 @@ protected:
   void mousePressEvent(QMouseEvent *) override;
   void mouseMoveEvent(QMouseEvent *) override;
   void mouseReleaseEvent(QMouseEvent *) override;
+  // Override the keyPressEvent to allow users to type a comma as a decimal separator.
+  void keyPressEvent(QKeyEvent *event) override;
 
 signals:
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Much appreciated.

PR #6199, opened 2025-10-16
Viewer Navigation Tools

shun-iwasawa:
This PR was implemented as an alternative to #5761 and #5775 , based on [this](https://github.com/opentoonz/opentoonz/pull/5761#issuecomment-3387918172) discussion.

This PR introduces three new tools; Pan View, Rotate View, and Zoom View.
These tools perform the same functions as the Hand, Rotate, and Zoom tools, but are used solely for temporary navigation while the shortcut key is pressed.
Unlike the original tools, it will return to the original tool even if the shortcut key is released quickly.

The default shortcut keys are assigned as follows:
- Pan View : `Space`
- Rotate View: `Shift` + `Space`
- Zoom View: `Ctrl` + `Space`

These can be customized in the Configure Shortcuts Dialog, just like other tools.

### DIFF
--- a/toonz/sources/include/tools/toolcommandids.h
+++ b/toonz/sources/include/tools/toolcommandids.h
@@ -32,4 +32,9 @@
 #define T_Finger "T_Finger"
 #define T_EditAssistants "T_EditAssistants"
 
+// Viewer Navigation tools (available only during the shortcut key is pressed)
+#define T_ZoomView "T_ZoomView"
+#define T_RotateView "T_RotateView"
+#define T_HandView "T_HandView"
+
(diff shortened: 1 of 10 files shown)


shun-iwasawa:
@konero
Could you please check if the usability of this PR is acceptable?

konero:
@shun-iwasawa Works great, no issues. 👍

RodneyBaker:
Outstanding.

Thanks for the verification @konero 

This change will resolve many users issues with use of Space key.

Thank you @shun-iwasawa for the new Viewer Navigation Tools!
Posting intent to merge.

PR #6202, opened 2025-10-16
Add commands: rotate left and rotate right

Abel032:
This PR introduces left/right view rotation and fixed few related issues:

## Changes

- Add left/right rotate commands: Keyboard shortcuts, context menu options, and panelbar buttons, the rotate angle is controlled by rotate tool's option

图片

- Add steeping rotate to `rotate tool`, use it by toggle `rotate by step` on
- Add rotate right/left angle to tool option, only affects the rotate commands
- Fix camera-center rotation: Make it available as tool option
- Fix ruler accuracy: Maintain correct ruler measurements during and after rotation
- Fix flip button states: Keep flip button checked status synchronized with canvas state

# How to test

1.  Rotate the view using different methods:

    - Config Press the keyboard shortcuts.
    图片

    - Right-click and use the options in the context menu.
    图片

    - Click the rotation buttons on the panel bar.
    图片

2.  Check ruler spacing during rotation:

    - Ensure the ruler is visible.
    - Rotate the view.
    - Confirm that the ruler's markings remain evenly spaced, without becoming more crowded or spread out.
    图片

3.  Verify flip button states:

    - Flip the view.
    - Check that the `flipY` and `flipX` buttons update to reflect the new orientation.
    图片

### DIFF
--- a/toonz/sources/tnztools/tooloptions.cpp
+++ b/toonz/sources/tnztools/tooloptions.cpp
@@ -2826,7 +2826,7 @@ RotateToolOptionsBox::RotateToolOptionsBox(QWidget *parent, TTool *tool,
   setFrameStyle(QFrame::StyledPanel);
   setFixedHeight(26);
 
-  TPropertyGroup* props = tool->getProperties(0);
+  TPropertyGroup *props = tool->getProperties(0);
   assert(props->getPropertyCount() > 0);
 
   ToolOptionControlBuilder builder(this, tool, pltHandle, toolHandle);
@@ -2842,7 +2842,7 @@ RotateToolOptionsBox::RotateToolOptionsBox(QWidget *parent, TTool *tool,
   button->addAction(resetRotationAction);
 
   connect(button, SIGNAL(clicked()), resetRotationAction, SLOT(trigger()));
-
+  
   m_layout->addStretch(1);
   m_layout->addWidget(button, 0);
   m_layout->addSpacing(5);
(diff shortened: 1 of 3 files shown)


Abel032:
Looking for opinions on the default shortcuts for rotating 90° left/right.

Abel032:
Refresh the build

DarrenTAnims:
> Looking for opinions on the default shortcuts for rotating 90° left/right.

This looks really useful. My only comment would be that some users might prefer to rotate 45 degrees at a time or even 15, so perhaps it might be useful to put the rotation amount in the preferences, defaulting to 90 degrees.
Apart from that, great work! 😀

Abel032:
@DarrenTAnims 
I added a scroll bar to rotate tool's option bar to adjust the stepping angle.
It affects the `Alt-rotate` and `rotate left/right` commands.

RodneyBaker:
@Abel032 I'm posting intent to merge this PR.
You likely will continue to tweak the underlying feature set but the core of what you have here is solid.

There is no rush however so please let me know if you prefer to delay merger.

The one thing I observed and communicated via discord was that it may be best to have the Rotate Tools rotation value rely on the sliders value for normal movement.  This would then tie the slider value to normal rotation and the additional Alt keying option can allow for the specialized control.

Abel032:
@RodneyBaker 
Thank you for the feedback.
I think it's kind of user preference, I wish to add a new checkbox to help decide whether rotate by stepping and an scroll bar to determine the rotate angle of rotate left/right commands, in this PR.

update: implemented

Abel032:
@RodneyBaker I think it's okay to merge this?
I tested it again just now and it works good on my computer.

RodneyBaker:
Yes, this PR is ready.
Thanks @Abel032

PR #6204, opened 2025-10-18
Enhance input validation and clamping in DoubleValueField

andeon:
This PR update improves a little the input handling in DoubleValueLineEdit by adding validation and value clamping.

- Input values are now clamped within the allowed range to prevent out-of-range errors that could lead to crashes or unexpected behavior.

- Additionally, the maximum character length is limited to prevent excessively long or invalid inputs, including those introduced by copy-paste actions.

Previously, infinite values were allowed, which could lead to instability. I think that in the future, these fields should have their own individual limits set to better handle value constraints.


RodneyBaker:
I cannot think of any case where this change will be adverse or overly restrictive.
In other words:  LGTM

Thanks @andeon

Issue #5114, opened 2023-09-28
Corrupt Vector Level causes crash when viewed in Browser or loaded into Scene


RattusaurusRex:
### Description

If a Vector Level becomes corrupted (which seems to happen if Opentoonz crashes during Auto-save), opening the scene containing the corrupt level will cause Opentoonz to crash.  If the corrupt level is loaded into a new scene by dragging the file into the program, Opentoonz will crash:
  
![corrupt_crash](https://github.com/opentoonz/opentoonz/assets/60277069/5b1d5e0f-75d7-4ddd-82a6-c625f47a5e85)

If the Browser is used to navigate to the folder where the corrupt level is stored, Opentoonz will crash:

![browser_crash](https://github.com/opentoonz/opentoonz/assets/60277069/47875737-201d-44a0-8133-a07860334aee)

The corrupt level CAN be viewed and opened without crashing in Opentoonz 1.5, allowing any undamaged drawings to be rescued by copying the drawings to a new level:

<img width="265" alt="corrupt_level" src="https://github.com/opentoonz/opentoonz/assets/60277069/4d8daef8-7db6-4d92-97c6-782c9b2d4714">

<img width="452" alt="corrupt_level_browser" src="https://github.com/opentoonz/opentoonz/assets/60277069/ee8febfe-fd7f-4148-b80c-ea6e8b9353c6">

The corrupt level cannot be saved, however:
<img width="494" alt="save_level_fail" src="https://github.com/opentoonz/opentoonz/assets/60277069/7698e14a-b3e0-4956-9bc8-b915501e37ae">

### Steps to Reproduce

1. Load the corrupt Vector Level into Opentoonz by dragging it into the program --> crash
2. Use the Browser or Level -> Load Level to navigate to the corrupt Vector Level's location --> crash

### Expected Behavior

Opentoonz should be able to handle loading or viewing the corrupt level without crashing.  There should be something to tell the user that the level is corrupt so they don't try to continue using it, but are able to access the undamaged drawings in the level so that all of their work is not lost forever.

### Screenshots, Video & Crash Logs

This contains a corrupt Vector Level file that I received from another Opentoonz user:

[crash.zip](https://github.com/opentoonz/opentoonz/files/12754552/crash.zip)

### OpenToonz Version

Nightly (Specify below)

### OpenToonz Version Information

2023-09-11 15:55:50 UTC

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Wacom

[...]

PR #6205, opened 2025-10-18
Fix crash opening PLI files

andeon:
This PR addresses an issue where partially corrupted PLI files caused loading failures. The fix prevents crashes by ensuring safer file handling. The fix can be tested using the corrupted level file shared in the referenced issue: https://github.com/opentoonz/opentoonz/issues/5114

Ported from: https://github.com/tahoma2d/tahoma2d/pull/1530

**Ref:** https://github.com/opentoonz/opentoonz/issues/5114

### DIFF
--- a/toonz/sources/common/timage/tlevel.cpp
+++ b/toonz/sources/common/timage/tlevel.cpp
@@ -11,7 +11,8 @@ TLevel::TLevel()
     : TSmartObject(m_classCode)
     , m_name("")
     , m_table(new Table())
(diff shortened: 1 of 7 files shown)


RodneyBaker:
Thanks @andeon for the port and @manongjohn for the fix.
Much appreciated.

RodneyBaker:
This has been a particular nasty bug as the user might not even know they are encountering the corrupted PLI file as all they are doing is navigating to that location.

PR #6090, opened 2025-09-03
New features for ToonzRaster Tape Tool and Fill Tool

Abel032:
This PR introduces several improvements for ToonzRaster Tape Tool and Fill tool:

## Autoclose improvements

- Refactor configuration with new settings structure.
- Add cache support to improve performance.

## Raster Tape Tool

- Enlarge processing area and use cache for Area Type.
- Ignore AutoPaint Inks option for TAutocloser and raster Tape tool.
- Use different fill color for closed areas during gap check.

## Raster Fill Tool

- Add Refer Fill and Gap Close options.
- Add preference to define region using both paint and inks.(Effect how fill tool work)
- Refactor FreeHand & Rectangle Fill
  - Allow filling inside the saveBox or image bounds Ink.
  - Allow to paint auto-paint lines in FreeHand type (fixes #5094)

PR #6213, opened 2025-10-20
Fix loading wrong env file

shun-iwasawa:

After introducing PR #6090 , the current project at OpenToonz launch now loads from the wrong ENV file.
(It has become `[TOONZROOT]/profiles/env/username.env` , not `[TOONZPROFILES]/env/username.env`.)

The direct cause of this issue was that the RasterTapeTool constructor referenced the TEnv values.
Since RasterTapeTool instance is created before [TEnv parameters are initialized](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/toonz/main.cpp#L151-L152), thus [VariableSet::load()](https://github.com/opentoonz/opentoonz/blob/master/toonz/sources/common/tapptools/tenv.cpp#L387) was being performed on an incorrect ENV file.
To fix this issue, I moved the functions referencing TEnv values into the `onActivate()` function, following the approach used in other tools.

On the other hand, another fundamental cause is that the initial values for `EnvGlobals::m_rootVarName` and `EnvGlobals::m_systemVarPrefix` are set to incorrect values in `EnvGlobals::init()`.
To make the source code more robust for future development, I set the initial values to the correct values using the newly introduced constant `ToonzVersion::systemVarPrefix`.

### DIFF
--- a/toonz/sources/common/tapptools/tenv.cpp
+++ b/toonz/sources/common/tapptools/tenv.cpp
@@ -82,7 +82,9 @@ public:
                    QString("/Contents/Resources/SystemVar.ini");
 #else
 #ifdef HAIKU
-    settingsPath = QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation) + "/SystemVar.ini";
+    settingsPath =
+        QStandardPaths::writableLocation(QStandardPaths::AppConfigLocation) +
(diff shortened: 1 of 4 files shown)


RodneyBaker:
Much appreciated!

PR #6214, opened 2025-10-20
Fix auto commit cell number when nothing typed

Abel032:
This PR makes some changes about cell number input field:

- Fix the issue that cell rename field commits hold cells when lost focus even nothing typed.
- Add preference： `Cell Input Method`.

图片

- Fix the `Auto Fill Empty Cell` feature: Only work when commit by focusing out by mouse even or `Ctrl+Enter`.

Ref: https://github.com/opentoonz/opentoonz/pull/5795#issuecomment-3420739560

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -244,7 +244,9 @@ public:
   // Visualization  tab
   bool getShow0ThickLines() const { return getBoolValue(show0ThickLines); }
   bool getRegionAntialias() const { return getBoolValue(regionAntialias); }
-  bool getRasterizeAntialias() const { return getBoolValue(rasterizeAntialias); }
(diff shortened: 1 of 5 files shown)


Abel032:
The black table only appear on my local build, but not happen on Github Action's build

shun-iwasawa:
It works very nice! 

- Ctrl+Enter for the Fill Empty Cell feature would be fine on our end. Users who would like to use this feature frequently may find pressing the Ctrl key each time a bit cumbersome, but for now, let's wait for feedback after release. Thank you for your consideration.
- Regarding the condition for executing `renameCell()` when focusing out, “the field text has changed from when it was displayed” might be more appropriate than “the field is not empty.” I’ll try it after merging this PR.

Thank you for the quick fix. LGTM

PR #6217, opened 2025-10-21
Fix(XshToolBar): Invisible after startup

Abel032:
Fixed: The XSheet Toolbar is not visible after startup.
This issue usually does not occur on the Timeline.


RodneyBaker:
I know someone that will be very happy about this fix.

Thanks @Abel032!
LGTM

PR #6219, opened 2025-10-22
Fix rename cell field focus out behavior

shun-iwasawa:
As [commented here](https://github.com/opentoonz/opentoonz/pull/6214#issuecomment-3424476296), this PR will modify the rename cell field behavior when it losing focus.
The condition for executing `renameCell()` when focusing out is changed from “the field is not empty” to “the field text has changed from when it was displayed”.

This modification will take care the following situation:
1. For example, assume you have a xsheet like this:
    image

1. You want to change the number in the first three cells to `5`. 
    So you select the first cell, type `5`, and press `Enter`.
    image

1. The change has been completed successfully. In order to finish renaming, you will click another cell to hide the Rename Cell Field.
    However, this operation changes the numbers starting from the fourth frame to hold cells with number `2`. In my opinion, this is an unintended result.
    image


Abel032:
Looks good to me.

RodneyBaker:
LGTM

PR #2143, opened 2018-07-06
Set System Paths with Command Line Qualifiers

shun-iwasawa:
This PR is requested by some Japanese animation studio.

For now, working configuration (such as UI, preference settings, current projects, etc.) is separated for each user. And one user is usually associated with one current configuration.

In Japanese animation production, there is a case that a machine for pencil testing is shared by multiple users, as it is connected to a camera stand, which is quite a large instrument.
Using OT for pencil testing just like other de facto softwares needs some special configuration, and such configuration is different from the one for other "normal" workflow like ink&painting or compositing. Therefore, there was a need to have multiple working configuration for single user. One is for pencil testing, and another one is for the other workflow.
To achieve such demand, they had been prepared two stuff folders, and separated registry values between the pencil test machine and other machines in order to use the different configurations. However, registry values are not easy to maintain as it needs administrative right to modify. 

This PR enables OT to input command line qualifies, which will specify the system paths (like `TOONZROOT`, `TOONZPROFILES`, etc.) overriding the registry values.

Here are the usage (you can see this by typing `OpenToonz_1.2.exe -help` in command line as well) :
```
usage: 
       OpenToonz_1.2.exe -help
       OpenToonz_1.2.exe -release
       OpenToonz_1.2.exe -version
       OpenToonz_1.2.exe [ -layout filename ] [ -TOONZROOT folderpath ] [ -TOONZCACHEROOT folderpath ] [ -TOONZCONFIG folderpath ] [ -TOONZFXPRESETS folderpath ] [ -TOONZLIBRARY folderpath ] [ -TOONZPROFILES folderpath ] [ -TOONZPROJECTS folderpath ] [ -TOONZSTUDIOPALETTE folderpath ]
       OpenToonz_1.2.exe [ -layout filename ] [ -TOONZROOT folderpath ] [ -TOONZCACHEROOT folderpath ] [ -TOONZCONFIG folderpath ] [ -TOONZFXPRESETS folderpath ] [ -TOONZLIBRARY folderpath ] [ -TOONZPROFILES folderpath ] [ -TOONZPROJECTS folderpath ] [ -TOONZSTUDIOPALETTE folderpath ] scenePath


  -help
       Print this help page
  -release
       Print the current Toonz version
  -version
       Print the current Toonz version
  -layout filename
       Custom layout file to be used, it should be saved in $TOONZPROFILES\layouts\personal\[CurrentLayoutName].[UserName]\. layouts.txt is used by default.
  -TOONZROOT folderpath
[...]

PR #6220, opened 2025-10-22
Fix registering multiple system var paths with command line qualifiers

shun-iwasawa:
This PR fixes a long-standing bug that has remained since the introduction of #2143 (since 2018!).

- When specifying `-TOONZROOT` and other variables (such as `-TOONZPROFILES`, `-TOONZCONFIG`, etc.) together as command-line arguments, the values of `-TOONZPROFILES` or `-TOONZCONFIG` may not override the value specified by `-TOONZROOT`.

This was because I mistakenly believed that `std::map::insert` and `std::map::emplace` would overwrite values when the key already existed, just like `QMap::insert` does. My apologies!

### DIFF
--- a/toonz/sources/common/tapptools/tenv.cpp
+++ b/toonz/sources/common/tapptools/tenv.cpp
@@ -274,7 +274,7 @@ public:
   }
 
   void setArgPathValue(std::string key, std::string value) {
-    m_argPathValues.emplace(key, value);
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thank you for the fix @shun-iwasawa.
You remind me that I need to explore that multiple stuff directory approach!

PR #6221, opened 2025-10-22
FillTool: Add gap close distance option

Abel032:
The gap close distance option is synced to the one of `raster tape tool`.
This PR also changes:

- The distance slider of `raster tape tool` is int type now.
- The default value of close distance changed from 10 to 30

### DIFF
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -521,6 +521,7 @@ class FillToolOptionsBox final : public ToolOptionsBox {
   ToolOptionCheckbox *m_selectiveMode, *m_segmentMode, *m_onionMode,
       *m_multiFrameMode, *m_autopaintMode, *m_closeGap, *m_referFill;
   ToolOptionPairSlider *m_fillDepthField;
+  ToolOptionIntSlider* m_gapCloseDistance;
 
 public:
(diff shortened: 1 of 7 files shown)


Abel032:
Fix issue: The autoclose settings not initialized properly if gap check is on when startup, add `invalidate()` to clear the old segments cache.

Abel032:
`Invalidate()` sometimes not work, clear all cache manually when auto-close settings changed.

Abel032:
After introduecing #6213 (It's correct) the initialization of auto close settings moved to TapeTool::onActivate().
And the gap close distance option implemented here would be initialize when fill tool activated.
So expect switching to tape tool first, the autoclose settings won't be initialized properly.

I need to make them share the same environment variables....

Abel032:
By adding extern declaration to `autoclose.h` and define/use them in different tools' cpp, managed to share and use the same variables to initialize the autoclose settings.

RodneyBaker:
This appears to work as advertised.
Initially I thought you were making the Gap Close Distance be the same in all places so we would see if we change the value in Fill Tool it will also change in Tape Tool etc. but now I see that is not the case.

The Gap Close Distance slider in the Fill Tool is very useful.
(and the increase of default from 10 to 30 is very welcome)
Thanks!

PR #6222, opened 2025-10-22
Resolve preferences compatibility

shun-iwasawa:
Related to #6214 , this PR will resolve compatibility of the `Cell Input Method` preference option.
If the `Input Cells Without Double Clicking` option is activated, the default value of  `Cell Input Method`  will become `Input by Single Click` to maintain usability and avoid confusion in case the number keys are already assigned as shortcuts for certain commands.


RodneyBaker:
Thank you @shun-iwasawa for this addition.
LGTM

PR #6224, opened 2025-10-23
Rename Tool Option Selective to Empty Only

Abel032:
- Geometric Tool : `Selective`->`Empty Only`
- Finger Tool: `Selective`->`Empty Only`
- Fill Tool: `Selective`->`Empty Only`
Store `Empty Only` option as environment variable so that user can restore it in next startup:
- Add new `Empty Only` shortcut and assign F7 as default
图片

- Change old `Selective` shortcut to F9
图片

---

Keep the `Selective` option of eraser tool and paint brush tool.

## Fix
- Fix repaint (Shift-Fill) failure of fill tool when `Empty Only` enabled and disable auto-Paint Lines feature for `#0` style.
- FillTool: Use `Shift` to disable `Empty Only` temporarily

## Feat
- Allows to use `Shift Key` to disable `Empty Only` temporarily


RodneyBaker:
Thanks for the fixes, adjustments and feature @Abel032 
(Pressing forward as Mac OS Mojave check is still eternally frozen)

PR #6226, opened 2025-10-24
Fix Toonz Raster brush offset with odd canvas dimensions

Abel032:
图片

Fix this by rounding input position before the inputmanager, affecting the cursor drawing and actually drawing.


RodneyBaker:
We'll keep an eye on this in case there is something we've missed.
All my test thus far suggest LGTM.

Thanks @Abel032 for fixing the case of pixels in the offset (and odd!) canvas dimension.

PR #6228, opened 2025-10-25
Fix(XSheetDepth): Invisible after startup

Abel032:
Fixed: The XSheet Depth bar is not visible after startup.
This issue usually does not occur on the Timeline.

Maybe we can replace all `visibleRegion().boundingRect()` to `geometry()`, but let's not replace them just in case.


RodneyBaker:
Thank you for the fix @Abel032 
LGTM

PR #6230, opened 2025-10-25
Style Picker: OrganizePalette: move picked style to current style page

Abel032:
Old: Move the picked to the end of 1st page in palette
New: Move picked style to the end of current style's page in palette

## How to test

This feature only available in color model viewer, try to import color model so that there would be at least 2 pages in palette.
And then select one style in one page and pick style in another page from the color model viewer.

### DIFF
--- a/toonz/sources/include/toonz/palettecmd.h
+++ b/toonz/sources/include/toonz/palettecmd.h
@@ -104,9 +104,10 @@ DVAPI void renamePalettePage(TPaletteHandle *paletteHandle, int pageIndex,
 DVAPI void renamePaletteStyle(TPaletteHandle *paletteHandle,
                               const std::wstring &newName);
 
-/* called in ColorModelViewer::pick() - move selected style to the first page */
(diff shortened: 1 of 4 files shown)


Abel032:
Fixed a small undo/redo issue.

RodneyBaker:
I believe this PR just needs to be rebased onto current master to resolve conflict.

Abel032:
@RodneyBaker 
I just merged master on Github

Abel032:
Missing a param..
Had better merge/rebase on local.

RodneyBaker:
There may be some potential downstream effects of the difference between the old 3 argument wrapper and the change to accept 4 arguments introduced in this change.  I don't know enough about this so will likely need to stare at it a little more.

Be be clear, I don't see a problem here... I'm trying to find one (if it exists).  ;)

RodneyBaker:
Thank you @Abel032 for the adjustment.
We'll want to make sure we publicize the change but it should also be obvious to those that use the process.

PR #6231, opened 2025-10-25
Feat(StylePicker): Add Option Replace Style

Abel032:
Add new option `Replace Style` to `Style Picker` tool:

图片

With this option, user can repaint Line and Area of current style with the new picked style form color model viewer.
Only support Toonz Raster and Vector level, conflict with `organize palette` option.

This PR also refactored some code of `organize palette` related to palette switch detection, the `Replace Style` will also be disabled automatically when current palette switched.

## How to test
Please test this with both `Toonz Raster Level` and `Toonz Vector Level` and also test if `Redo` and `Undo` works as expected.

### DIFF
--- a/toonz/sources/tnztools/stylepickertool.cpp
+++ b/toonz/sources/tnztools/stylepickertool.cpp
@@ -10,6 +10,7 @@
 #include "toonzqt/tselectionhandle.h"
 #include "toonzqt/styleselection.h"
 #include "toonzqt/gutil.h"
+#include "toonzqt/icongenerator.h"
 
 // TnzLib includes
 #include "toonz/txshsimplelevel.h"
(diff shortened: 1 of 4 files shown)


Abel032:
Looks okay

RodneyBaker:
@Abel032 In testing I've really been enjoying this option.
It's been awhile but when I first tried working with Color Models this is how I believe I expected the process to work.

Something I did note that should be looked at is how when the Line and Passive options are on how that effects the use of this new Replace Style.  That would appear to be something for future development however.  I don't see it as problematic as the process appears to simply ignore those options.  It might be confusing to the user however.

Posting intent to merge as this is a very useful addition to not only the Style Picker but for use with Color Models.

RodneyBaker:
Thank you @Abel032 for yet another useful feature.
LGTM

PR #6236, opened 2025-10-28
Feature: Report missing/unknown XML tags during scene loading

Abel032:
This PR make it possible to know which tag in scene file(`.tnz`) is lost/unknown/broken while loading, so that user can choose to delete/fix them using text editor.

图片

## How to test
1. Corrupt a .tnz file (e.g., with a text editor).
2. Load it and check if the broken tags are reported.


RodneyBaker:
I am quite sure I've never had this much fun breaking an Opentoonz project file before.
This will help considerably with troubleshooting.
LGTM

PR #6238, opened 2025-10-28
[Need Discussion] Feat: Layout Guide (improved Safe Area)

Abel032:
## Summary
This PR introduces new features to existing `Safe Area` feature.

### 1. Allows to choose a layout preset or edit layout presets for the scene when creating new scene at startup popup window
图片
(The layout preset name would be stored in .tnz file.)

### 2. The default level resolution now can be set to `camera size`/`layout size`/`preference value`.
图片

When `Current Layout Template Size` activated, the default size of new level woule be set the same as the `Layout Template` in current `Layout Preset`.
If there are no availiable layout templates, new level would be created with the `Camera Size`
(This preference replaced the old option:`New Levels Default to the Current Camera Size` )

### 3. Show BG card of layout template in Scene Viewer 
Now the background is not only limited to camera area, but also the area of layout template:
图片

The color can be set in `Scene Settings`:
图片

### 4. The `Safe Area` panel button has been renamed to `Layout Guide` and got new behavior
图片
With button activated, viewer would not only draw the safe area, but also draw the layout template.
图片

This feature works on `Scene Viewer`(Scene Editing and Level Editing mode), and `Flip Book Panel`:
图片

### 5. Allows to open an context menu with new options by right click panel button
图片

- `Clear Current`: Reset current layout preset to empty: 图片

-  `Edit Layout Presets...`: Edit Layout Presets to open an editor to adjust params of layout preset
图片

> The defualt folder of layout templates is `stuff/library/layouts`
(Use areas to manage the safe areas)

> Current layout template would be added as `Scene Cast`, and put into `Layout` Folder, won't be cleared by command `Remove Unused Levels`

> The preset would be applied to current camera when layout activated 

## Compatibility 
- Load `stuff/config/layoutpresets.ini` for default, if not exist then load `stuff/config/safeareas.ini`, if both not exist, create `layoutpresets.ini`
- Use a default preset of `80% area + 90% area` for new created scene / old scene, only changing it by `Left clicking on context menu item` or `Edit Layout Presets...` would create new tag for current scene, making it incompatible with old OT version.
- The output option `Put Layout Image` is toggled off for default, the tag/property would be written only when it toggled on, so that the scene can be opened with old version
- The scene with one Layout Preset Name preset might be missing preset on another computer. With right clicking the panel botton, missing preset would be shown with red, `clicking context menu red item` or `Edit Layout Presets and click red item listed in presets` to create the missing preset.

## Fix
- Fix the 1-pixel clipping of flip book's camera frame (also other frames)

## Presets
Added 4 in total:
图片

Source for the new layout presets:
https://aja.gr.jp/jigyou/chousa/aja_layout

## How to test
Please confirm if the features and compatibility listed above working as expected.


toonz/sources/toonz/framesPresetsEditorPopup.cpp
```diff
@@ -0,0 +1,691 @@
+#include "framesPresetsEditorPopup.h"
+
+// ���� Qt �ؼ�
```
Abel032: delete this

konero:
No time for code feedback but here's what I got from testing:

1. There is excessive I/O access on each tick, like adjusting a Color slider, selecting an item, typing in the name field. I don't know the current pattern but I would defer persistence and keep the .ini contents in memory to handle the GL draws. Add an 'Apply' button to write to disk without closing the window, and make the OK button as 'Apply + Close' and the Cancel button 'Revert + Close'.

2. Add Area/Remove Area doesn't redraw until a value has been changed.

Abel032:
@konero 
Thanks for the feedback.

- I agree we need less IO acess, I will try to add a setter to set a static pointer variable(QSetting), and call it when previewing clear it when finishing preview (maybe passing a custom structure would be better?🤔)
- Good suggestion for better button names 
- I will fix the bug of not updating preview

RodneyBaker:
Thanks for adding this feature @Abel032 
I did a quick test and will test more later as I got a crash when attempting to render with a layout.
There might be a little confusion in naming here but I don't see that as too much of an issue.
Reference Frames, Safe Area, Layout... it will be good to lock that terminology down.

The 'Put Layout Image' I would call 'Add Layout Image' to balance with the 'Add Clapperboard' over on the other side.  I was thinking 'Add Layout Overlay' but I don't think it's an overlay.

There is a lot to like about this feature and I'm still finding new elements of it.
I really like that we can use PLI/vector layout images.

Something I need to test:  When using the Hand Tool (middle mouse button) the layout image moves.  This seems useful but I'm not sure it is intended.

*The crash appears to relate to Wintab so perhaps a build issue.
I haven't been able to successfully render with a layout yet.

RodneyBaker:
Update to my last.

It appears the crash I experienced relates to selecting a PLI/vector image for the layout.
If I use a raster/PNG no issue and I can successfully render.

Also, the odd movement with the layout image only seems to happen if a vector image is selected.

Abel032:
@RodneyBaker 
I can reproduce the odd movement and render crash here, thanks for the feedback!
I'll try to resolve this issue.

Abel032:
And about the terminology..I think Layout Image contains reference frames, and reference frames contain safe areas...
Maybe `Add Layout Overlay` would be better name than `Put Layout Image`.

It's always welcome for more suggestions from users.

Abel032:
TODO: 

- [x] Support vector layout image put on rendered frames.

Maybe only do the over after rendering instead of adding new nodes would be better.

Abel032:
The image viewer(flip book) generates 1 pixel offset after rendering.
Not sure what caused it, check it tomorrow, other parts seem to work fine.

konero:
`MI_Frames` feels too vague, just skimming the code it's not clear what 'frames' refers to. It could mean X-Sheet frames, FlipBook/rendered frames, drawings (the 'Add Frames' command). Using a more specific name would make the intent clearer.

I'd suggest using/keeping the term `SafeArea`, which is correct terminology in video editing, you could use `LayoutTemplate` for the image overlay part. These would make the code more readable and self-explanatory.

The filenames should follow the same convention for consistency, ex: `framesPresetsEditorPopup.h` > `safeAreaEditorPopup.h`.

You don't need `QUrl` include since you commented out the part that used it (if that's everything): https://github.com/opentoonz/opentoonz/pull/6238/commits/5da56c74802285a3e552d38d2c9484ec5156e6ee#diff-1ad56b3f25743de5ddec50a66bfa6e660d5c591eb9e0a91dc14cecd46ac380d3R457

I'm struggling to find time for any real testing lately (in middle of moving) so I haven't managed to try everything. From looking at the code it seems possible to render out the `SafeArea` borders (I'm not referring to image layout template) onto the final result? In `rendercommand.cpp` you added a lot of headers for one small feature, could it not instead be encapsulated into a helper somewhere else?

Thinking back to the drawing board for a sec - I would forbid the use of vector, that's overcomplicating things I think (who is making templates in PLI?) and you won't need to do specific scaling for it. if you make the logic to overlay a `LayoutTemplate` raster image then rasterize the `SafeArea` border and treat it the same way.

But I might be wrong, I could only skim the code so far.

Abel032:
About the initialize thought of renaming safearea to frames, the layout paper usually not only contain the safearea, but also the camera and the coloring/scan area.
I agree that the name **Frames** feels vague, but safe area also not so clear.
Maybe there is a proper name, not `safe area` and not `frames`.

I will remove the includes and commented code together later.

About the logic of generating new Fx Node for putting layout paper(in `rendercommand.cpp`), I'm not sure where can I put it currently. It can't render the `SafeArea` boards.

For the support of PLI, I think user won't use vector level as layout template...
And currently it just behavior the same like raster level.
But it's still acceptable now I think.

Bracket-I-I:
A (title) safe area is actual filmmaking stuff. (Including, but not limited to animation).

Getting rid of that for something called 'a reference frame' is breaking age old convention.

I'm not saying that conventions should be kept forever, especially with modern displays where bezels and the like get smaller and more standardized, and how cameras are 'digital' but..it's still not some throwaway term.

Abel032:
What about just call it LayoutTemplate, it contains the safe area, camera area and scan area.

Abel032:
This feature not just provide pegbar but also allows you to see the margin out of camera area, but safe area can't because it's inside the camera area.
These margin would also be used to write some information like the cut number and the duration.

Abel032:
As I know, Ghibli do not use OT to draw, but only for color modelling, cel painting and composition.

Abel032:
I think we use layout templates a lot in production, and it is a good tool.

Bracket-I-I:
The safe area being inside the camera makes sense because it's what ends up being put to the screen, and that's where placement of elements inside the safe area matters.

Imagine panning the camera to move it to another part of a large background and centering on text.
If the safe area is not glued to the camera anymore, how do you know the text is in the safe area, which was designed to ...well, make it likely that something is actually visible where it matters.

If you don't have safe area glued to the camera then you might end up with something like this
(courtesy of wikipedia images)
image

Again, with modern displays, it's not that important, but it's also not completely unimportant.

Abel032:
I know how safe area works, it also make sense in animation, but different with film industry, we often draw outside of camera to ensure we have more space to adjust (for example, to make the camera shaking) and we can write messages for animator or other stuff in these spaces.

Bracket-I-I:
The level is always visible when moving the camera, unless one chooses 'camera view', so any frame information is not lost...
Even if you scale the camera tiny, you will still see the level outside of it.

But that aside, frame information like ( 1b ) or something doesn't make it in the final output, that's just information for inbetweens and keys and all that, it can be drawn anywhere.

But maybe I"m misunderstanding you.
My viewpoint is: 
On the topic of safe areas, they are the only thing that matters, nothing else.
I realize that there are 'layout templates' with time info, peg bars (optional), a virtual camera frame with some space in the margins, yeah.

But that's already in opentoonz in the form of just levels. 

Let's say though, the layout gets put in the safe area with this new feature though. 
So you have all those nice peglines and other layout bits.

Like a field to put in which second + frame you have like
image

Where it says time (  +  )
So you draw in, with a brush, the time in seconds and extra frame count there.
Let's say (2 + 1)
...and then you turn off safe area...

and now your entire layout is gone o_O? 
2 and 1 would float on the page , now that (  +  ), which is part of the layout template is gone. So instead of ( 2 + 1 ) it's 2   1

But again, maybe I don't understand you. To me, if I wanted a layout template, I'd do it in some way like:

If I only want to work on one level, and not on a stack of layered columns, I would reserve one frame of the level, usually the last frame.
There I would put  the layout template there, and if I wanted to draw something on top of it, I would use a pinned onion skin so I'd see a ghost of the template while I draw a rough from scratch.

This is basically the same as your feature, and if I were to turn off the onion skin, then the template would vanish too.
Just like the template would vanish if you tied it to the safe area display.

The other approach would be to make two columns, one has the template the other the drawing, and then using 'x sheet as animation' thing...draw like it's photoshop with layers.
That way the layout would be visible all the time.

This approach makes the most sense, because it's separate..and that's what I'm arguing.
On paper, you can't have it separate, you get a template on paper and draw on that, and then maybe scan or photograph it.
But by drawing on the physical paper, you will be 'in the good areas' automatically, if the areas are on the paper too.

And once it's digital what matters is the safe area only. 

In any case though, in both of these approaches, you can move and scale and rotate the camera freely and you'd still see the margins.
Like here

image
Only if I enabled camera view, then I wouldn't see the rest of the template.

And the template in this case is just a copy pasted raster file. 
So yeah...

Abel032:
I personally think it's okay to separate the information (duration number for example)and the layout template.

Adding template as level is possible but not very convenience, since you have to import it and adjust its duration everytime since OT do not support static layer or the scene template（it don't pack assets).
But it is necessary in some cases.

Another related issue is, the default canva resolution is binding to camera resolution or the size that defined in preference.
So you might see people do like this: 
- Import the layout file and then set the camera resolution to the level settings.
- Change the resolution of camera or switch camera when exporting or just previewing the final result.

Not saying this feature is a good final solution, but I think it's a good chance to consider the relationship of default canvas size and the camera size.

It looks like there are more inconvenience things when using layout templates in OT than other software.

Abel032:
So I may add more functions related to the layout template, like allowing user to use the resolution of layout to new the levels and be able to specify the layout preset when newing a scene.

Abel032:
I converted this PR to draft because I wish to add some layout related features(Connect layout size to level creation and the background color card, their function just like the canvas margin, provide a default size of scene canvas and keep the information of output region at same time) in this PR, and sort the code further(maybe a standalone file for layout related functions) so that they can be removed or replaced by other feature in the future easier.

Update:
https://aja.gr.jp/jigyou/chousa/aja_layout
A possible layout template.

amotarzi:
@Abel032 The most important rule of OT development is to never break existing functionality! 

Large studios incl. Studio Ghibli rely on these existing functionalities and we can’t just change or tweak them at will. 

I believe it would be best if you create your own forks based on OpenToonz similar to other forks e.g. Morevna Project.

We should not continue turning the main branch of OpenToonz into something it’s not designed to and not need! It’s already perfect for processing high-end 2D animation. In contrast OpenToonz should be developed further towards stability and what it already excels at!

Abel032:
@amotarzi 
Sorry, I don't know what is good or bad now.
But there is a difficulty when creating template scene with layout, and able to export in camera size while editing in layout size easily.

Abel032:
You can consider this as something like proposal.

RodneyBaker:
@Abel032 You are proceeding just fine with this feature.
Having said that, it is good to be reminded that we want to proceed with care so I'm glad to see the feedback this PR is getting.  

@amotarzi 
Able has been very responsive to realigning his features to make sure they follow the guidelines put forth for developing Opentoonz.  Consider that forking is a good idea for everyone and it would be good for you also to fork Opentoonz.  That is the absolute best way we can maintain a copy of Opentoonz guaranteed not to change unless we specifically allow the change to our fork.  

I encourage all to review the PRs where @shun-iwasawa has worked with Abel to keep changes optional.  There are some 200+ requests open at the moment for additonal changes to Opentoonz and that doesn't include the most obvious ones related to stability (bug and crash reports specifically).  Abel has been very diligent in tackling those stability related issues too.

For general reference, here's a link to Shun's development guidelines:  #2746

We really do need to place those guidelines front and center.
Here is the core of that guide:

> A basic policy for development of OpenToonz so members can review and merge PRs without hesitating.
> 
> The policy consists of one principle and two supplementary notes:
> 
> PR can be merged if it will not interfere existing usages.
> Note, however, that:
> 
> If there is the possibility that introducing a PR may conflict with Opentoonz existing license, please wait for owner's review.
> From several month before releasing stable release date (to be announced), please do not merge "feature PR"s.

> Opentoonz has many features which enables various way of producing animation.
> A modification for some workflow may become disadvantage for others. In such case please consider to make it optional (via user profiles) so that users can select whether to use it.
> 
> Here are examples:
> 
> Bugfix : Fine to be merged.
> 
> Refactoring : Fine to be merged.
> 
> Enhancement / Speed-up : Fine to be merged.
> 
> Translation : Fine to be merged.
> 
> New Fx : Fine to be merged.
> 
> New command : Fine to be merged.
> 
> New panel : Fine to be merged.
> 
> New preset (fx, camera settings, shortcut, etc.) : Fine to be merged.
> 
> Altering default user profiles (preferences, panel layout and shortcut) : Fine to be merged.
> 
> Altering UI which cannot be controlled by user profiles (like adding buttons to the panel or adding commands to the context menu) : Basically it's fine to be merged, but please be careful to make sure it will not disturb the conventional usage.
> 
> Altering existing fx or rendering behavior : Basically it's fine to be merged, but please ensure it will not change a rendering result of a scene which is made with the previous version.

As you have emphasized @amotarzi, we want to proceed very cautiously where changes alter existing usage.  In production an extra keypress or mouse click can be problematic as those keypresses will significantly add up over time.

We are currently waiting for feedback from studios such as Ghibli and now is the time to press forward with features as we are fast approaching feature freeze at end of December.  That freeze being designed to specifically allow time to refine changes based on user feedback.

If and when changes effect existing usage it is essential for users to identify the problem.
Once identified we can then resolve that problem.

RodneyBaker:
I haven't gone far into testing this PR but I'm liking what I see thus far.
While testing I've even been using this feature as a 'anime style' vertical storyboarder as that was one of the first custom templates I created and tested.

image

I really like how you have the option to 'Add Layout to Xsheet'.
That's very useful and perhaps particularly for resizing the camera to specific templates as we can go into Camera Settings select that layout in the xsheet and set the camera to the currently selected level.
Nice.

image

I have had a few crashes which I must assume relate to the conflicts still needing to be resolved.
Keep up the good work.  (Added:  Later a complete lock up where tools are not operational and input from keyboard not accepted)

(In the above you can see areas where I failed to have the background of the template be transparent.  It's useful to know that we can make areas both transparent and opaque.  I can see this being useful for masking information we want to see while working but not see in final render)

Added:  The crash appears to occur when I start drawing in to add some details to the shot after animating the camera (making an animatic based on the storyboard I'm playing with).  It may be at the point of creating a new level (by drawing) where I'm trying to add details to the current shot.  Sorry, as this won't be clear... just trying to get it sorted in my brain.  ;)

Adding a no frills animatic based on that storyboard just moving the camera over the storyboard with layout template turned off:


https://github.com/user-attachments/assets/b03a23af-ac96-475c-ae11-a125d09ee592

RodneyBaker:
I have to share this because... I'm having fun.
Here I'm using the template... added to the xsheet... to view a storyboard.
I have a vector level as a child of a Second Camera (Camera 2) which how us what Camera 2 will see.

(Extra Naru Lazy Brush FX autofill thrown in because... remember this is fun!)


https://github.com/user-attachments/assets/be504755-ced6-45ab-b6a5-8d1c599ae4e8

If we then activate Camera 2 and render we aren't surprised to see what we expected:


https://github.com/user-attachments/assets/563e2495-2fb1-4865-88bc-1d55366398df

Abel032:
I observed some strange behaviors and crash when creating new level by drawing on the canvas, I guess that's what @RodneyBaker met.

Abel032:
There is a bug need to be fixed, but I don't have time recently.
Have converted this PR to draft.

Ahl76:
I've done some testing with this PR. 

1. Bugs and Glitches Encountered

* **Random Crash in Autocreation Mode:**
    A random crash occurs when drawing directly with autocreation enabled. The crash does not seem to happen if the Level is created manually before drawing.

* **Stroke Display Glitch (Undo/Redo):**
    Display issues with the stroke occur intermittently, specifically after using Undo/Redo commands. The glitch becomes noticeably worse as drawing continues or when hovering over the canvas with the stylus or mouse. (it's same with appveyor or windows build)

issues artifacts

* **Tool Selection at Startup:**
    It is impossible to manually select a tool (with stylus) from the UI immediately after launching OpenToonz. Only the keyboard shortcut works to activate the tool.

~~* **Raster Bounding Box Offset:**  
    Maybe it's intentional, but the raster bounding box is offset from the default zone of a template, without using the 'Add to Xsheet' feature.~~

Added : 
      My bad, I was mistaken.  
The feature works correctly with the proper preference and when the layout guide is set before creating the level.

* **Template Line Visibility:**
    Lines in image-based templates tend to disappear and reappear inconsistently depending on the zoom level.

display template lines

### 2. Potential suggestions for the Future

* **Guide Color Selection:**
    Double-clicking on the color square to change color of the guides should open the Style Editor as is the case in certain OT functionalities.

* **Slide-to-Edit on Canvas:**
    For better visual comfort and workflow speed, allow value editing by sliding the cursor directly on the canvas (similar to how certain FX or the Cleanup Camera currently work).

https://github.com/user-attachments/assets/ca7604f9-75c2-49c3-b659-b3ddbb64999a

* **Safe area mask:**
    Possibility of masking the safe area with colors would be appreciated and would quicly add clarity (for guides done in OT)

safe area mask


* **Overlays:**
    Add the ability to use Overlays (similar to those found in the Clapperboard tool) to display it on the layout

https://github.com/user-attachments/assets/c06d459b-7dcb-414a-8635-c0c284c298a0

* **PDF Export & Storyboarding:**
    Possibility of PDF exportation, it would be great. As for the clapperboard, the pre-production board panel, this tool would be a way to do storyboarding and exporting to PDF would be optimal. Unfortunately, currently none really have all the aspects to carry it out end-to-end. But ofc theses optional suggestions arent core to the main purpose of this PR.

RodneyBaker:
Looking forward to any updates you have on this PR @Abel032

Abel032:
These bugs appear to be fixed in the latest code.
I think the issue was caused by frequent layout level checks and incorrect DPI initialization for new levels.

- Random Crash in Autocreation Mode
- Stroke Display Glitch (Undo/Redo)
- Tool Selection at Startup

RodneyBaker:
image

Looking good.

There are a few areas which are not optimal in my estimation but they don't relate to the code and feature itself but more to the flow/workflow of using the layouts.

- The Output Settings text 'Put Layout Template' is reads to me as awkward. 'Put' being the awkward term.

image

If it said 'Add Layout Template' it would balance with the Add Clapperboard on the reverse side.

- While on one hand useful, it is awkward that when moving the cursor to select the option to 'Add Layout to Xsheet' that the templates all update the canvas while moving to select the option.  I'm not sure how we would resolve this.  Mostly a minor inconvenience.

- When a layout is added to the xsheet this creates a sometimes useful although potentially confusing situation where the layout guide is overlaid onto the canvas but the template is also exposed on the frame.  This can cause confusion until the user realizes they need to turn one or the other off.  This again is something the user can resolve on their own and not quite sure how we would remedy this short of temporarily turning the guide off which also wouldn't be optimal.

- An additional area of confusion is where the 'Put Laytout Template' option is on in Output Settings and the guide is also on.  This can lead to a case where we see two instances of the template.  Turning off the guide option in the toolbar resolves the problem but in order to see the layout overlay with Preview on the guide opton must be turned on.  This leads to an awkward situation where the what is seen in Preview and in the Final Render can be different.

None of these are showstoppers in any way and don't prevent this PR from moving forward.
I add the information as general feedback.

Abel032:
I think removing the ‘put layout template’ and ‘hover to preview on stage’ features would solve the problem.
We can still preview layout in layouts editor popup and apply the layout template by adding it to xsheet.

update： please don't merge before removing

RodneyBaker:
Something I've been seeing when testing this PR but only now have started to lock down:

image


Using the Camera Settings Option to 'Use Current Level Settings' to resize the camera to the current level's size is something I use often. (Underused I expect because the button is a little hidden)

I note that when transferring a layout guide to the xsheet using this PR's 'Add Layout to Xsheet' option and then clicking the button to 'Use Current Level Settings' more often than not it seems to leave room at the top for the Camera title overlay area and adds that same amount of space to the bottom of the resulting camera assignment.

Added:  Importantly, it appears that when rendering the correct camera sizing is used despite what is being shown on canvas.

RodneyBaker:
Actually, I may have spoken too soon about the rendering part:

rendertheguide

Abel032:
@RodneyBaker 
You can clear current layout guide or edit it and reset the offset to 0.

So this layout feature make it easier to export with camera size but harder to export in full size....
Maybe we need to cancel this feature and try making camera editing more easier..
图片
图片

Or just introduce canvas and margin to OT..
This feature still need consideration.

The main issue is we need two kind of size, one with bleed and one without bleed (trim size).
If we use central symmetry layout, the solution would be very simple, just create two cameras.
If the layout is not central symmetry, create two cameras and animate one of them.

RodneyBaker:
Oh.  Clear Current Layout Guide did the trick.
Not sure how I missed that.
Thanks for that!

This is your PR to cancel but I definitely will want to store a good copy away for use if you do decide to close it down.  I'd leave it open and mark it as WIP... especially if you think it won't likely move forward in time for inclusion in version 1.8.  The ability to add these custom layouts is very useful as is but I do understand some of the challenges you are facing.

Given some of your stated goals, it regrettably reads this like we should plan to introduce this feature set after the release of v1.8

The bottom line:  This PR is great but apparently it can be even greater with additional time.

andeon:
Perhaps it could be released with reduced options. I remember when the Assistants were introduced, it was done that way. Even today, we don't have menus to add 'Level > New', for example. Sometimes it feels like a hidden feature.

The only problem I see is breaking previous edits, but with a newly introduced feature, that is to be expected. If it doesn't change the workflow for those who don't use it or use in old way(Safe Area)? Since layout guides can be toggled on or off. Then it shouldn't be an issue, right?

I don't see why it shouldn't be included, unless it doesn't meet the developer's own criteria or actually breaks something in the pipeline. If it's treated as an 'Advanced' or 'Experimental' feature, it allows power users to benefit from it. To reinforce: if it doesn't break the existing flow, I don't see why it shouldn't be included in the v1.8 cycle.

RodneyBaker:
Thanks for the feedback and encouragement @andeon
I certainly prefer this new feature to be in v1.8 and I am confident it can happen.
We still have time.

Opentoonz v1.8 will be a lesser release if it isn't included.
And... this PR is still labeled with 1.8  ;)

RodneyBaker:
New release schedule has been posted.
REF:  https://github.com/opentoonz/opentoonz/discussions/6357#discussioncomment-15966129

shun-iwasawa:
First of all, thank you @Abel032  for proposing this large-scale new feature and for the repeated revisions to the PR.

I think it’s a great idea to provide presets for setting the size of new Raster/ToonzRaster levels.
As you mentioned, it’s common practice to make the drawing area slightly larger than the camera size.
I believe having predefined sizes as presets will be effective for project management.

[Question]
You made it possible to offset the layout center and used the AJA layout paper template — was this to maintain consistency with Clip Studio (or traditional paper animation)?
Personally, I feel that having different top and bottom margins for the levels might make the subsequent workflow somewhat cumbersome..

[Suggestion]
AJA-Fr01 is a standard where sizes are specified in actual dimensions (not pixels).
OTOH OpenToonz can handle actual dimensions within the software.
Currently, you prepare multiple AJA-Fr01 images with different DPI values to match the camera DPI,
but would it be difficult to provide a single actual-dimension template so that
it displays correctly even if the camera DPI and the template DPI differ?


I sincerely apologize for my delayed response.
Concerning this feature, I would like to further explore it after the release of V1.8, with the goal of implementing it in V2.0.
Thank you for your understanding.

RodneyBaker:
To add to the discussion I'm started a port of T2D's Scene Overlay here:  https://github.com/OpenAnimationLibrary/opentoonz-dev/pull/42

That is just a working PR with no intent to merge that PR before considering carefully how that feature set and this one can be combined and refined.

PR #6240, opened 2025-10-30
FFmpeg path auto-detection

Abel032:
This PR helps auto-detecting the ffmpeg path when starting OpenToonz, it use the command `where` for windows and `which` for Linux/MacOS to get the path of ffmpeg that stored in system's env.

This PR also switched the order of third party initialization and toonz environment initialization:
Initialize thirdparty path first.

### DIFF
--- a/toonz/sources/toonz/main.cpp
+++ b/toonz/sources/toonz/main.cpp
@@ -488,11 +488,11 @@ int main(int argc, char *argv[]) {
   TBigMemoryManager::instance()->setRunOutOfContiguousMemoryHandler(
       &toonzRunOutOfContMemHandler);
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I'm really looking forward to this @Abel032 as it will personally save me a lot of time reconnecting Opentoonz to ffmpeg when testing every release.

I believe the appveyor checks are just timing out.

Abel032:
Refresh the builds

RodneyBaker:
The nightlies (for windows) still aren't building but when they do this will be part of them.

Thank you @Abel032 for this important addition.
It will save so much time constantly reconnecting to ffmpeg.

PR #6241, opened 2025-10-30
Feat: Lazy load for rooms

Abel032:
This PR makes the panels in rooms to be created when current room switched, instead of initialize all of them when startup.

It can be toggled off in `Preferences>General`

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -190,6 +190,9 @@ public:
   PathAliasPriority getPathAliasPriority() const {
     return PathAliasPriority(getIntValue(pathAliasPriority));
   }
+  bool isLazyLoadRoomsEnabled() {
+      return getBoolValue(lazyLoadRooms);
+  }
 
   // Interface  tab
   QStringList getStyleSheetList() const { return m_styleSheetList; }
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -13,6 +13,7 @@ enum PreferencesItemId {
   projectRoot,
   customProjectRoot,
   pathAliasPriority,
+  lazyLoadRooms,
 
   //----------
   // Interface
--- a/toonz/sources/toonz/cleanuppaletteviewer.cpp
+++ b/toonz/sources/toonz/cleanuppaletteviewer.cpp
@@ -40,7 +40,7 @@ int search(TPalette *plt, TCleanupStyle *style) {
   assert(false);
   return -1;
 }
-}
+}  // namespace
 
 //********************************************************************************
 //    CleanupPaletteViewer implementation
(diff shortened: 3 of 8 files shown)


konero:
Was this resolving an issue? There's a small amount of stutter introduced when switching room for the first time (I don't personally like it, I change rooms frequently).

There's a long lag spike when switching to a room with a Tasks panel in it. I didn't test all panels, potentially others?

Abel032:
@konero 
It may avoid some potential crash, since the `Cleanup Settings Panel` may generate dangling palette pointer. (I think it was fixed though).

The Initialization of Task Panel need more time since it need to detect existing network port on system and open a new port, also initialize every task param input field.

The original thought of this PR is come for speed up the startup process, by lazy loading some panels...

Here is a list that the most time cost startup processes:

1. t32bitserv (Even QuickTime not installed on my computer)
2. Ceate Menubars and Rooms (Panels)

I may add two preferences for them, and toggle them to `Disable QuickTime encoder/decoder` and `Lazy load rooms` for default.

RodneyBaker:
I don't see the one time delay in loading rooms (the first time) as a negative over all necessarily.
Still need to test that out more though.

The one element I have questions about in this PR relates to the Task room which appears to have that more significant delay due to reasons mentioned by @Abel032 and network/server support.  

For my part, I use Task Rendering almost exclusively without network/server setup.
My curiosity is that how much of the change to focus the Task room on use of a network is at the cost of using Tasks with just the same unnetworked computer Opentoonz normally renders on.

This isn't necessarily a matter for this PR to address but I guess the point I'd like to make is:

Every user can leverage/exploit Task Rendering even if they aren't using a network.
They can just run it without setting up a network.

I want to emphasize this because the vast majority of users don't even visit the Task room because they believe it requires use of a network.

Abel032:
I didn't dig deeper, but maybe the render task need to be sent by local network(only one computer), and it might also need to open a port to connect the potential server even we didn't setup the farm server.

RodneyBaker:
My take is that we likely do not want to automatically connect to a server (for a number of reasons) to include:

1) The time it takes to connect to a server
2) The resource used in connecting to a server
3) Antivirus checks that flag/intercept the connection to a server 

If this auto-connectivity is a Preference similar to that of the check for Opentoonz updates which is off by default that would potentially make good sense as long as with the ability to render with the option both on would not adversely be effected.

For my part, I should lean more into server use of Task Rendering but for the most part I haven't had that need.  For me, the use of a server with Task Rendering does complicate the process as well as increase the need for troubleshooting when things don't work as expected.

I guess what I'm saying is that at last check, Task Rendering via server is an area that still requires some development and bug fixing.  Task Rendering without a server mostly just works.

RodneyBaker:
I like this change but am wondering if we can make it an option in Preferences.  For those that don't change rooms often this will speed things up but for those that dive in and change rooms a lot it may not be optimal.

I'm looking to see if there is some middle ground between but that's what immediately comes to mind.
I think there is another option but I'm not sure what it is.

My point though is that this is a nice change.
We just need to account for @konero's concern.

Abel032:
Currently it's added as an option in `Preference>General`.

RodneyBaker:
Currently it's added as an option in Preference>General.

That is perfect for moving this PR forward.
As always @Abel032, thanks for all you do.

PR #6242, opened 2025-10-30
Fix(t32birserv): Terminate the process when quiting or failing to init quickTime

Abel032:
This PR adjusted the logic of launching `t32bitserv.exe` for Windows.
- Ensure that there are only one t32bitserv.exe running in background, close it when the application closed (By the QPointer)
- Terminate `t32bitsrv.exe` if it didn't find installed QuickTime.

To use the QuickTime encoder/decoder controlled by `t32bitserv.exe`, you need to download and install it from this [link](https://support.apple.com/en-us/106375).


RodneyBaker:
I'm glad we are doing this PR now and not near the official v1.8 release.
The issues with t32bitserv.exe were quite problematic and cause a lot of headaches.
As such it might make people a little nervous if merging was at that release time.

Thanks @Abel032 for the fix.

Issue #4552, opened 2022-09-27
Failed to save Plastic Level after applying Autorenumber option to frames in plastic column.


beeheemooth:
### Description

I managed to reproduce one of the annoying glitches that cause frames to disappear in Plastic Levels.

**A Plastic Mesh Level becomes unsaveable if its first frame is moved and autorenumbered.**

_Note: I wrote a similar report first in Tahoma2D, but it turned out that it is present in Opentoonz_

### Steps to Reproduce

1. Create a new level (the type doesn't matter) and draw some shape. (You can save the scene, by the way)
2. Apply the Plastic Tool to the drawn shape, as a result a new Plastic Level will be formed.
3. Move the first frame in the plastic mesh column down (or up if you didn't create the level in the first frame)
4. Apply Autorenumber to the moved plastic mesh frame. Note that in the Level Strip, the frame has also changed its number (as it should), and an asterisk has appeared next to the level name, which indicates that the level has changed.
5. Try to apply "Save All" or "Save Level" and you will get a warning window about an error in saving the level. If you ignore it and close the scene, then after restarting the scene, the mesh frame will be red.
If you return the frame to its original position and apply Autorenumber again, the level can be saved.

### OpenToonz Version

Nightly (Specify below)

### OpenToonz Version Information

built Aug 25 2022 17;37;33

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Huion

beeheemooth:
@RodneyBaker Here I mean a closed and filled shape so that the Plastic Tool can be applied. It doesn't matter how you draw the shape.
I tried to reproduce the problem on another computer and I succeeded.
If I can reproduce the problem in both Tahoma and Opentoonz on different computers, then it is logical to assume that the problem still exists. The fact that you could not reproduce it is probably due to different versions of Windows.

RodneyBaker:
Yes, I agree.  The diffferent versions of Windows is the most likely candidate.

I do have a little suspicion that something like the older logitec webcam that was causing problem with Opentoonz starting... drivers like that... can produce odd things.  

I do also note that you have a Huion tablet which is mentioned often in reports with bugs/crashes.
It would be odd indeed if your reported problem relates to the Huion tablet though.

If someone else on WIndows 8.1 doesn't have the issue then... I'd look for a driver issue.

beeheemooth:
[...]

PR #6243, opened 2025-10-31
Fix renumbered mesh save issue

andeon:
This PR resolves a bug that occurred when saving renumbered meshes and, in some cases, could cause file or image data to be lost. Verify the fix by following the steps in [#4552](https://github.com/opentoonz/opentoonz/issues/4552)

Ported from https://github.com/tahoma2d/tahoma2d/pull/1317
Co-Authored-By: manongjohn 19245851+manongjohn@users.noreply.github.com

Ref: https://github.com/opentoonz/opentoonz/issues/4552

Note: Comments were translated into English, and indentation now follows the default `.clang-format`

### DIFF
--- a/toonz/sources/toonzlib/txshsimplelevel.cpp
+++ b/toonz/sources/toonzlib/txshsimplelevel.cpp
@@ -97,7 +97,7 @@ QString getCreatorString() {
 
 //-----------------------------------------------------------------------------
 
-bool checkCreatorString(const QString &creator) {
+bool checkCreatorString(const QString& creator) {
   int mask = 0;
   if (creator != "") {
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@andeon I was never able to reproduce the problem using either beehemooth or Rattus's steps.
I attempted to again for quite awhile just now and still could not.

I am going to proceed with merger and those that can reproduce the problem should be able to confirm the result.

Thanks to @manongjohn for the fix and to you @andeon for the port from Tahoma2D.

Added:  Nightlies may not be getting generated so confirmation may take awhile.

andeon:
I was able to reproduce it by saving each step, and then got a warning window saying: `Failed to save the following resources:` message. Could you please close the issue? Thanks!

PR #6245, opened 2025-11-01
Preference: Use QuickTime coder

Abel032:
On some computers it takes long time to start `t32bitserv.exe` and detect/load installed QuickTime coder, this PR adds a new preference to `Import/Export` tab, the default value is off.

图片


RodneyBaker:
Thanks for the Preferences option @Abel032 

Aside:  While looking through the code some time back I saw that at one time the MOV format was preferred because it allowed targeted removal of single frames rather than adjusting the whole video file.  It looked like that feature was not enabled and I've never followed up to find out more.  My assumption is however that if the code allows for it it might only be the native Quicktime support that allows for that data swap and not the ffmpeg process.

PR #6246, opened 2025-11-02
Prevent crash by checking m_oldStyle before getPointer()

andeon:
This PR fixes a potential crash by adding a null check before calling getPointer() on m_oldStyle. 

- A crash has been observed on Linux when using styles from the palette in a raster level. Although not yet reported, it may also occur on macOS. _(No issues observed on Windows.)_
- Comments have been translated into English, and indentation has been updated to follow the default .clang-format rules.

### DIFF
--- a/toonz/sources/toonzqt/styleeditor.cpp
+++ b/toonz/sources/toonzqt/styleeditor.cpp
@@ -15,8 +15,8 @@
 #include "toonz/cleanupcolorstyles.h"
 #include "toonz/palettecontroller.h"
 #include "toonz/imagestyles.h"
-#include "toonz/txshsimplelevel.h"  //iwsw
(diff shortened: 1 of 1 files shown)


Abel032:
Not the scope of this PR, but I get a crash when initializing the the cleanup setting panel recently, the function name is buildGui().
It try to get the cleanup palette from the palette controller, but the result is a deleted palette pointer.

andeon:
I think cleanuppaletteviewer.cpp needs to be updated to always check whether the palette handle and the palette itself are valid before using them. There are still many parts of the program that need similar checks to prevent these kinds of issues.

RodneyBaker:
I trust that Linux (and potentially Mac) users will let us know if they still have related issues.

PR #6251, opened 2025-11-03
Fix stroke fill redo in normal

Abel032:
This PR fixes the redo of normal stroke fill（freehand fill) not working (without refer fill and gap close options).

This bug is caused by redo with re-caculation instead of tileSet since it use less memory, but when I did this change, I forget to store the transformed stroke but stored the original stroke.


RodneyBaker:

PR #6252, opened 2025-11-03
Remove recolor painted lines feature of areafiller

Abel032:
This PR removes a feature of repaint painted lines beside enclosed regions(the enclosed region has the same style as the painted lines), this can avoid unintended repainting, especially in AA Image case.

| Old | New |
| ----- | ----- |
| before | after |

But user can still paint painted lines by `Lines Mode`, or use `Ctrl` to only paint autoPaint lines.


RodneyBaker:
I would press forward with this PR but there are conflicts that must be addressed first.

RodneyBaker:
LGTM

PR #6253, opened 2025-11-03
ci: Update Python version in macOS workflow

andeon:
This PR updates the Python version used in the macOS workflow to ensure the build can continue successfully.


RodneyBaker:
Will merge this as the Mac build is unlikely to complete until it is merged.
Thanks @andeon for the update.

PR #6260, opened 2025-11-09
fix(fillTool): avoid unintended savebox update for bound fill

Abel032:
This PR fixes unintended savebox update only happen when this preference is toggled on:
图片

- Also fix another issue: when saveBox is empty(erased by earser), fill tool can't work.

## Fill Tool:
| Before | After |
| ----- | ----- |
|  |  |


Abel032:
Looks like the `MacOS(Mojave)` has been removed and appveyor returned to normal.

RodneyBaker:
> Looks like the MacOS(Mojave) has been removed and appveyor returned to normal.

Which means we are back on track.
Thank you for the fix!

PR #6261, opened 2025-11-10
Darker Theme

JParks123:
Screenshot (77)
Darker Theme for OpenToonz application.
Screenshot (562)

Screenshot (565)

This theme is intentionally created to be very Dark for those of us who have sensitive eye issues with the bright Themes currently available.


RodneyBaker:
Just adding a note that we do need to chart a way forward for themes that are not generated using LESS.  There is a middleground to be had in that but the ideal way forward is certainly to use LESS.

I've experimented with generating the LESS from the QSS and while that process works there is quite a bit of manual adjustment in that process to get it where it needs to be.

All this to suggest that not using LESS will not stop this PR from moving forward but it might signal that policy will be to only accept LESS generated QSS from this point forward.  Of this I am not entirely convinced as we want to make the process of generating themes as easy as possible for users.  Not so much so they can add themes to the Opentoonz distro but so they can create and share themes with each other.

A few things we can do moving forward:
- Thoroughly document the process of creating Opentoonz themes with LESS.  Most of the information is out there but not optimally collected and presented.
- Consider the generation of themes to be one on ramp into Opentoonz development.  New users can spend time getting comfortable with QSS / LESS while learning how to customize and improve other aspects of the program.
- Consider other approaches to using QSS. Example:  Copy/Pasting an entire theme into Prefernces > Additional Stylesheets is one one to have Opentoonz change themes.  Perhaps we can add Save and Load options at that location.
- Etc.

All of this (Except the specifics of best practices using LESS to generate QSS) is outside the scope of thie particular PR.

andeon:
The problem with not having LESS is that the theme can easily break depending on updates, since it’s built from that group of files. I think unofficial themes should ideally be shared externally and maintained by their original creator, because maintaining them is complicated, especially when there are many style, images and color variations.

This is one of the reasons I’ve never submitted themes here. Doing so can create a maintenance burden for other contributors if the original creator used different methods to create it and then becomes unavailable. In other communities, like Blender, [Krita](https://krita-artists.org/c/resources/brushes-and-bundles/32), GIMP, etc., this is how it works. You can find themes on forums, DeviantArt, and other external sites. Often, you’ll also find nice themes broken because they are no longer updated, this happens a lot in Blender, for example, since its development has accelerated in recent years.

The real issue is that the "OpenToonz community" has never really formed around creating and sharing extras like themes, brushes, or FX.

JParks123:
@andeon Thanks a lot, I guess I will close this abomination.

RodneyBaker:
@JParks123 Not sure why you closed this PR as it likely would have been merged.  It signals to others in the future why directly adding QSS is not the preferred method and you helped get that method published and highlighted.

Every PR is different and each one is considered on a case by case basis.

One good thing that has come out of this is that we need to refine the build instructions as it can be hard to find the part relating to Developing Stylesheets (Themes).  Link:  https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_stylesheet.md

For the record, the main thing that needs to be PR'd for your Darker theme to work is the custom icons.  A user can copy/paste a QSS theme   (entirely or just the parts changed) into Preferences > Additional Stylesheets.

Or...

This Additional Stylesheets approach is a valid way forward for users that just want to tweak and share themes.  Place the custom icons directory inside the new theme directory.  Then those icons can travel with a zip file and be found by Opentoonz easily when the theme is unzipped and put in place.

JParks123:
> @JParks123 Not sure why you closed this PR as it likely would have been merged. It signals to others in the future why directly adding QSS is not the preferred method and you helped get that method published and highlighted.
> 
> Every PR is different and each one is considered on a case by case basis.
> 
> One good thing that has come out of this is that we need to refine the build instructions as it can be hard to find the part relating to Developing Stylesheets (Themes). Link: https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_stylesheet.md
> 
> For the record, the main thing that needs to be PR'd for your Darker theme to work is the custom icons. A user can copy/paste a QSS theme (entirely or just the parts changed) into Preferences > Additional Stylesheets.
> 
> Or...
> 
> This Additional Stylesheets approach is a valid way forward for users that just want to tweak and share themes. Place the custom icons directory inside the new theme directory. Then those icons can travel with a zip file and be found by Opentoonz easily when the theme is unzipped and put in place.


Many people have been asking for this Darker UI theme so I decided to move forward with this project.

konero:
You haven't ported this to LESS, your AI just copy-pasted the QSS into a `.less` file.

JParks123:
> You haven't ported this to LESS, your AI just copy-pasted the QSS into a `.less` file.

Thanks for pointing that out — you're right. I mistakenly included raw QSS instead of porting the colors to variables. I'll update the PR with a proper LESS conversion.

andeon:
I recommend creating a repository for this, like I did with https://github.com/andeon/opentoonz-themes
, so updates and maintenance are managed by yourself.

Now If you want to integrate it with OpenToonz, check the tutorial @RodneyBaker shared about using LESS. It needs to be integrated with other base LESS files, meaning you should use the existing bases as references and create a separate LESS file containing only the differences. Each file should link to the others, so maintenance becomes easy: if you change something in the base, the change is automatically applied to all others. Just take a look at examples like blue.less, synthwave-theme.less, etc.

When adding a theme, you’re passing responsibility to others for keeping it updated. Following this approach is crucial. If the standard isn’t followed, maintaining it becomes unmanageable.

JParks123:
> I recommend creating a repository for this, like I did with https://github.com/andeon/opentoonz-themes , so updates and maintenance are managed by yourself.
> 
> Now If you want to integrate it with OpenToonz, check the tutorial @RodneyBaker shared about using LESS. It needs to be integrated with other base LESS files, meaning you should use the existing bases as references and create a separate LESS file containing only the differences. Each file should link to the others, so maintenance becomes easy: if you change something in the base, the change is automatically applied to all others. Just take a look at examples like blue.less, synthwave-theme.less, etc.
> 
> When adding a theme, you’re passing responsibility to others for keeping it updated. Following this approach is crucial. If the standard isn’t followed, maintaining it becomes unmanageable.

Thanks for the feedback. I reworked the theme to follow the LESS architecture and now only override variables, similar to the existing themes.

konero:
I think we're going round in circles a bit here.

1. This will not compile to your theme design, the provided QSS file and LESS do not match.
2. You've hard-coded QSS classes into the LESS theme, that's fine for hacks but not the majority; it defeats the purpose of a future-proofed theme structure.
3. It's missing the 'out' compile path for the EasyLess extension.

@andeon Is it possible (and if there's a consensus on it being a good idea) to prevent action builds from triggering when modifying objects in `stuff/config/qss`? Currently, for the most part we can copy-paste the QSS into the preferences box for testing. And the rest is easy to checkout via CLI. Might add a little friction in some scenarios perhaps, need more input, assuming it doesn't affect nightly builds.

andeon:
@konero I’ve even considered completely blocking the `stuff` folder, in theory (since I haven’t analyzed the entire structure), because if it’s just being copied, there’s no real reason to generate builds when it is changed. Of course, in that case, any changes inside would only be testable by downloading the folder directly. But I think some folders, as you mentioned, could work this way, but we’d need to verify.

For example, generating a build just because we added brushes or FXs, which don’t actually change any lines of code, doesn’t make sense. In general, builds should occur only due to compilation steps. However, there are still external checks without any filters that are triggered automatically by any change.

For example, I blocked builds so that when a single workflow is changed, only the one for the relevant platform is triggered. However, the external checks are still triggered automatically, as explained above.

I already explained that PRs should be submitted only when they are ready to be evaluated, to avoid this excess of builds, since some quotas may be exceeded depending on the number of activations, and maybe there’s some immunity for open-source projects. My biggest concern is with the external checks, as I’m not sure if there’s a quota or any blocking mechanism in place if the limit is exceeded.

JParks123:
> @konero I’ve even considered completely blocking the `stuff` folder, in theory (since I haven’t analyzed the entire structure), because if it’s just being copied, there’s no real reason to generate builds when it is changed. Of course, in that case, any changes inside would only be testable by downloading the folder directly. But I think some folders, as you mentioned, could work this way, but we’d need to verify.
> 
> For example, generating a build just because we added brushes or FXs, which don’t actually change any lines of code, doesn’t make sense. In general, builds should occur only due to compilation steps. However, there are still external checks without any filters that are triggered automatically by any change.
> 
> For example, I blocked builds so that when a single workflow is changed, only the one for the relevant platform is triggered. However, the external checks are still triggered automatically, as explained above.
> 
> I already explained that PRs should be submitted only when they are ready to be evaluated, to avoid this excess of builds, since some quotas may be exceeded depending on the number of activations, and maybe there’s some immunity for open-source projects. My biggest concern is with the external checks, as I’m not sure if there’s a quota or any blocking mechanism in place if the limit is exceeded.

Understood. I'll make sure to test changes locally and only push when the update is ready to be evaluated so we avoid triggering unnecessary builds.

PR #6262, opened 2025-11-12
Update README badges with correct GitHub Actions workflows

andeon:
This PR updates the build status badges in the `README.md` to match the current GitHub Actions workflows:

- Windows (`workflow_windows.yml`)
- MacOS (`workflow_macos.yml`)
- Linux (`workflow_linux.yml`)

This also removes the Travis CI badge. Maybe it is still used internally for some tasks related to official releases,  but since it is not used in public/community development and does not reflect the current branch or build status, the reference has been removed.

Fixes [opentoonz/opentoonz#5962](https://github.com/opentoonz/opentoonz/issues/5962)

PR #6263, opened 2025-11-13
Update README badges with correct GitHub Actions workflows 2

andeon:
This PR saves and converts the README files to UTF-8 encoding.

Thanks to @Abel032 for spotting it. I forgot to save it before converting, but now it should be in UTF-8 format. I hope it’s correct now.

Ref: [opentoonz/opentoonz#6262](https://github.com/opentoonz/opentoonz/pull/6262)

### DIFF
--- a/doc/README_chs.md
+++ b/doc/README_chs.md
@@ -10,21 +10,21 @@
 
 ## 关于OpenToonz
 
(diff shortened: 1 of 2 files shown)


Abel032:
@andeon Yes it is correct now, thanks 👍

Issue #4442, opened 2022-05-09
Assertion '__n < this->size()'  failed, when browsing in "Choose folder" dialog : 


fenugrec:
### Description

When browsing folders and clicking "Up one level" a few times then selecting a folder, opentoonz crashed .

### Steps to Reproduce

1. Start Opentoonz
2. In the "Create new scene" area, click button with three dots to change `Save In:`location
3. Press "Up one level" a few times
4. In the list on the right, double-click any folder
5. crash

### Expected Behavior

Not crash : )

### Screenshots, Video & Crash Logs

Build is the one from the Arch linux repos.

Starting from the console, I get the following output after the crash:
```
/usr/include/c++/11.2.0/bits/stl_vector.h:1045: std::vector<_Tp, _Alloc>::reference std::vector<_Tp, _Alloc>::operator[](std::vector<_Tp, _Alloc>::size_type) [with _Tp = FileBrowser::Item; _Alloc = std::allocator<FileBrowser::Item>; std::vector<_Tp, _Alloc>::reference = FileBrowser::Item&; std::vector<_Tp, _Alloc>::size_type = long unsigned int]: Assertion '__n < this->size()' failed.
Aborted (core dumped)
```

### OpenToonz Version

1.6

### OpenToonz Version Information

Build date May 4, 2022

### Operating System

Linux

### GPU

AMD

TomatDividedBy0:
Can confirm, have run into the _exact_ same problem.

flurick:
@fenugrec @TomatDividedBy0 looked like theres been some change to the startup window, can you still reproduce this issue?

flydiscohuebr:
Same problem using flatpak version https://github.com/flathub/io.github.OpenToonz/issues/22

hfiguiere:
Accessing any element in an empty container is UB. 

Flatpak build with assertions on, so libstdc++ do catch it.

andeon:
@RodneyBaker I tried to reproduce this with the AppImage and a local build, but I couldn’t get the crash to show up. Since this issue is from 2022, it seems like it was fixed at some point, so we can probably close it.

If it happens again, they can comment here or open a new issue with the crash report.

hfiguiere:
You need to build with `-D_GLIBCXX_ASSERTIONS` to reproduce. It's default on Flatpak, but elsewhere it usually is not.

andeon:
[...]

PR #6268, opened 2025-11-16
Fix FileBrowser crash when selecting folder

andeon:
This PR fixes a crash that occurred when selecting a folder after navigating "Up one level" multiple times in the FileBrowser.

The crash was caused by accessing `m_items` with invalid indices, leading to out-of-bounds vector access. Invalid selection indexes are now safely ignored.

This issue is noticeable in builds with `-D_GLIBCXX_ASSERTIONS` enabled (e.g., Flatpak builds), which perform extra STL bounds checking.

Fixes: #4442


RodneyBaker:

Issue #2297, opened 2018-09-16
Pasting doesn't paste as a selection, ready to edit


DarrenTAnims:
### Issue Summary
When you paste a selection onto a different cell, I'd expect the pasted lines to still be selected so that I can move or resize them, but it isn't unless you first draw a selection on the page before pasting.

### Screenshot or Video Reference
![screencapture](https://user-images.githubusercontent.com/28575907/45598927-0bb22200-b9db-11e8-938c-846c8ed532a6.gif)

### Steps to Reproduce
What I did in the video is:
- Created a toonz raster level
- On frame 1 draw a circle
- On frame 2 draw a triangle
- Go to frame 1, draw a selection box around the circle and copy (I pressed Ctrl+C)
- Go to frame 2 and paste (I pressed Ctrl+V), the circle appears on top, not selected
- I then pressed undo, then paste, then undo again, to show multiple clicks doesn't help
- Then click and drag to create a selection
- Paste - and the circle is pasted as a selection and I can move it to where I need it

### Expected Results
As above

### Actual Results
As above

### System Information
I'm using the latest OTX version, but have also tried it with the nightly build from 2018-09-12 02:00:12 UTC:
https://github.com/opentoonz/opentoonz_nightlies/releases/tag/nightly_98926de9b60802499ab899617149d6ffe527b3d8

RodneyBaker:
I observed a workaround for this problem where the pasted image doesn't have the selection box around it.
Just prior to pasting select an area of any size (this to trigger the selector).
Then paste.
The resulting pasted image should be inside the selection box ready for movement/manipulation prior to making it permanent.

PR #6273, opened 2025-11-19
Fix selection on paste for Vector / Toonz Raster

andeon:
This PR addresses the issue reported in **#2297**, where pasting a selection into a different cell does not keep the pasted strokes selected.

When pasting into a Toonz Raster or Vector level, the pasted lines **would not be selected**, so the user couldn’t immediately move or resize them. This was particularly inconvenient because the user had to manually draw a selection before pasting or rely on undo and paste again to manipulate the pasted strokes.

- **Vector**: works even when pasting into a non-existent level (column) or into an uncreated filmstrip, because vector strokes are independent objects and don’t require a pre-existing image to retain selection.

- **Toonz Raster**: pasting with a selection box works **only if the level (column) already exists**. If the column doesn’t exist, pasted strokes are incorporated directly into the image in that cell, **losing the selection box**, so the pasted content cannot be immediately manipulated. For individual cells, pasting works independently as long as they belong to an existing column.

ref: opentoonz#2297

### DIFF
--- a/toonz/sources/tnztools/vectorselectiontool.cpp
+++ b/toonz/sources/tnztools/vectorselectiontool.cpp
@@ -50,6 +50,7 @@ const int l_dragThreshold = 10;  //!< Distance, in pixels, the user has to
 
 namespace {
 
+// Get four points from vector image based on style IDs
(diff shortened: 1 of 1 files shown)


RodneyBaker:
LGTM

PR #6276, opened 2025-11-19
Fix stroke selection and crash after multi-frame paste in filmstrip

andeon:
This PR ensures that only the strokes of the current frame are selected after a paste, preventing out-of-bounds access when pasting into multiple frames.

Previously, if the user already had strokes selected in the image and pasted into multiple frames in the filmstrip, all strokes from all affected frames would be selected. This was an uncommon but possible scenario, which could result in invalid stroke indices if the user tried to manipulate them, potentially leading to crashes.

Now, even in this case, only the newly pasted strokes of the current frame are selected, preventing crashes.

ref: https://github.com/opentoonz/opentoonz/pull/6273

### DIFF
--- a/toonz/sources/toonz/filmstripselection.cpp
+++ b/toonz/sources/toonz/filmstripselection.cpp
@@ -253,40 +253,52 @@ void TFilmstripSelection::cutFrames() {
 }
 
 //-----------------------------------------------------------------------------
-
 void TFilmstripSelection::pasteFrames() {
   TXshSimpleLevel *sl = TApp::instance()->getCurrentLevel()->getSimpleLevel();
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Mojave check is still frozen but will press on. 
Do we even need that check?

andeon:
This check is for the external macOS release, but it’s still pending even though nightly builds are being generated. Not sure why it’s stuck, possibly due to a timeout or deadlock in the build. Maybe @shun-iwasawa needs to restart something there.

RodneyBaker:
Pressing forward with this.
Thanks @andeon for the fix!

PR #6277, opened 2025-11-19
Fix(fillTool): stroke fill one click crash with savebox limit

Abel032:
This PR fix a crash related to zero size raster image to be filled:
1. Enable `limit filling with Savebox`
2. Use fill tool, `freehand`  type on toonz raster level
3. single click outside of savebox


RodneyBaker:
At a guess I'd say Appveyor check just timed out.
The code is simple and straightforward enough.

RodneyBaker:

Issue #6269, opened 2025-11-17
Crash when deleting Node on Stage Schematic


RattusaurusRex:
### What's the bug?

Same issue as here: https://github.com/opentoonz/opentoonz/issues/3659

### Steps to Reproduce

1. Click on a node in the Stage Schematic and press Delete

### Screenshots, Videos or Crash Logs

![Image](https://github.com/user-attachments/assets/8628f146-3838-44d7-91e3-6e084947119d)

### OpenToonz Version

Nightly (specify below)

### Exact Version Info (optional)

built Nov 11, 2025

### Operating System

Windows

### GPU Brand

NVIDIA

### Graphics Tablet

None

RodneyBaker:
Confirmed.
That is a very reliable and consistent way to close down Opentoonz.
I hope you didn't lose any files in the process.

Tagging as 'Known Issue' because everyone should know to avoid this.

RodneyBaker:
Should be resolved by #6278

PR #6278, opened 2025-11-19
Fix crash when deleting a node in the Stage Schematic

andeon:
This PR fixes a crash (#6269) that occurred when deleting nodes in the Stage Schematic.

The fix safely copies the selected objects before deletion, preventing crashes even after repeated selections.

Ref: #6269


RodneyBaker:
LGTM

Issue #2021, opened 2018-05-29
UI:  Glitches


gab3d:
When translating the latest strings, in preparation for v1.2.1 I've seen some small glitches in the UI.

- Bad location of "auto stretch frame" option text in the Preferences (it goes over drop down menu easily).
- FX Browser: buttons' text space does not adjust properly to other lenghts.
- Merge TLV levels: buttons' text space does not adjust properly to other lenghts.

RodneyBaker:
@gab3d  This report could use a status check.
Perhaps if the stated issues are still valid a separate issue could be opened for each?

These do appear to relate to a few other similar UI spacing issues.
Once Identified perhaps we can cross reference them in this report as well.

For now, adding this to the UI Scrub project.

gab3d:
They are all still valid, yes...

As a sample...
In the following image the buttons should display: **Insertar** / **Agregar** / **Reemplazar** ...only the lenght of the last one is correct.
![fx_browser](https://user-images.githubusercontent.com/18104091/67627622-f04c5400-f835-11e9-9971-8933a632f419.png)

RodneyBaker:
Adding the translation tag back as that difference is button size requirement was possibly overlooked.

andeon:
@RodneyBaker I sent a PR to fix the button sizes. I'm not entirely sure about other issues, but if the user points out the places or shares screenshots here, we can address them later. For now, I think this issue could be closed.

PR #6284, opened 2025-11-21
fix increase minimum size of buttons in insertfxpopup

andeon:
This Pr fixed UI issue from #2021 where FX Browser buttons (Insert, Add, Replace) were too narrow for longer translated strings. 

Minimum width increased from 65px to 90px to prevent text clipping and ensure consistent layout across languages.

fixes: #2021


RodneyBaker:
LGTM

Issue #6283, opened 2025-11-21
Crash using the Polygon Tool


GARTHENNIS:
### What's the bug?

OpenToonz 1.7.1  (Build Mar 14 2023)

Report Date: 2025-11-21 15:02:56
Crash Reason: EXCEPTION_ACCESS_VIOLATION

Build ABI: x86_64-little_endian-llp64
Operating System: Windows 10 Version 2009
OS Kernel: 10.0.19045
CPU Threads: 8

GPU Vendor: Intel
GPU Model: Intel(R) HD Graphics 4600
GPU Version: 4.3.0 - Build 20.19.15.4531

Crash File: C:\Users\Tony\AppData\Local\OpenToonz\OpenToonz\crash\Crash-20251121-150256.log
Mini Dump File: C:\Users\Tony\AppData\Local\OpenToonz\OpenToonz\crash\Crash-20251121-150256.dmp

Application Dir: C:/Program Files/OpenToonz
Stuff Dir: C:\OpenToonz stuff

Project Name: sandbox
Scene Name: Key scene
Project Path: C:\OpenToonz stuff\sandbox\sandbox_otprj.xml
Scene Path: C:\OpenToonz stuff\sandbox\scenes\Key scene.tnz

==== Modules ====
C:\Program Files\OpenToonz\OpenToonz.exe
C:\Windows\SYSTEM32\ntdll.dll
C:\Windows\System32\KERNEL32.DLL
C:\Windows\System32\KERNELBASE.dll
C:\Windows\System32\USER32.dll
C:\Windows\System32\win32u.dll
C:\Windows\System32\GDI32.dll
C:\Windows\SYSTEM32\OPENGL32.dll
C:\Program Files\OpenToonz\Qt5PrintSupport.dll
C:\Program Files\OpenToonz\Qt5Xml.dll
C:\Windows\System32\gdi32full.dll
C:\Windows\System32\msvcrt.dll
C:\Windows\System32\msvcp_win.dll
C:\Windows\System32\ADVAPI32.dll
C:\Windows\System32\ucrtbase.dll
C:\Windows\System32\sechost.dll
C:\Windows\System32\RPCRT4.dll
C:\Windows\System32\SHELL32.dll
C:\Windows\System32\bcrypt.dll
C:\Windows\System32\combase.dll
C:\Windows\System32\ole32.dll
C:\Windows\System32\OLEAUT32.dll
C:\Program Files\OpenToonz\glut64.dll
[...]

PR #6286, opened 2025-11-23
Fix crash when drawing outside canvas with Geometric Tool

andeon:
This PR fixes crashes that occur when using MyPaint brushes with geometric tools (shapes) on Toonz Raster or Raster levels in OpenToonz.

- Prevents crashes when using MyPaint brushes on Toonz Raster or Raster levels **outside the canvas**.
- Fixes typo: `reasterTool` → `rasterTool`.
- Resolves crashes that occur when drawing with **left-click followed by right-click**.

**Fixes:** [Issue #6283](https://github.com/opentoonz/opentoonz/issues/6283#issuecomment-3564884624)

**PS:** In the future, this code will need refactoring due to repetition. Consider creating a helper function, unifying some undo operations, and improving shapes that lack certain functions.

### DIFF
--- a/toonz/sources/tnztools/geometrictool.cpp
+++ b/toonz/sources/tnztools/geometrictool.cpp
@@ -1,4 +1,5 @@
 
+
 #include "geometrictool.h"
 #include "toonz/tpalettehandle.h"
 #include "tools/toolhandle.h"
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I was under the impression this was already fixed but I guess the main difference is the 'outside the canvas' element of the equation.

andeon:
If the drawing stays even just a millimeter within the internal canvas, it doesn’t crash. It only crashes when drawing completely outside. I think that’s why this bug was never noticed, since it’s an old issue. I also discovered the mouse left-and-right click bug.

Many people may have experienced crashes because of this without realizing it, as it can happen from an accidental right click.

RodneyBaker:
With a bit of luck perhaps this will catch some of the odd crashes that are hard to pin down.

Issue #3265, opened 2020-04-30
Generate QSS files on compile time


morevnaproject:
Currently QSS files are generated ["manually"](https://github.com/opentoonz/opentoonz/blob/master/doc/how_to_stylesheet.md#how-to-compile) and stored in Git history.

This creates additional conflicts for PRs (when a lot of frequent changes happen to UI) - the PRs are harder to merge.

Since QSS files are generated from LESS files, I think that this process should be automatic and should take place at compile time (handled by CMake build system).

Benefits:
This will allow to remove QSS files from Git history and have less conflicts in PRs -> easier merges -> less headache for contributors.

To solve this we need to find a commandline tool to convert LESS files to QSS on WIndows. 
Anyone knows something like this? ^__^

ave4:
Also, the locale is duplicated in two places: **toonz/sources/translations** and **stuff/config/loc** Which is bad for me too

RodneyBaker:
As it has been 5 years, any additional thoughts on this request?

cc:  @andeon @konero @able @shun-iwasawa

andeon:
@RodneyBaker I think the translations could be generated during the build.

I noticed that Shun Iwasawa has to manually commit the `.qm` files. Since `.qm` files are binary and generate artifacts, they should be built, not versioned or manually created by the translators. Only the `.ts` files need to stay in the source.

I’m going to check how to make CMake generate the `.qm` files automatically during the build and copy them to `stuff\config\loc` in the process, so we could remove all `.qm` files from the repository.

As for the `QSS`, I don’t see much of a problem like with the `.qm` file case. It looks like a `.css` file in the end, but it adds extra work when changes are needed because it has to be generated separately and sent together with the LESS files. 

However, the workflow could be adjusted to handle this using Node.js and resolve it as well.

andeon:
@RodneyBaker After reflecting and performing some tests, even with automatic compilation, at least the `Default.qss` file must be present in the repository for safety, because without it the UI would be completely broken if anything failed. Therefore, keeping one or ten QSS files wouldn’t make much difference, since the effort required to regenerate them for updates is essentially the same. As a result, compiling them during the build process doesn’t provide a meaningful benefit, and it would require additional dependencies (e.g., Node.js and lessc on Windows) to be installed on all systems, whether in CI environments or locally, with no apparent gain.

[...]

PR #4828, opened 2023-04-14
Translation files for v1.7

shun-iwasawa:
This PR updates translation files (.ts) for all languages.
Also added Japanese translation.

PR #6290, opened 2025-11-25
Enable automatic translation generation and folder renaming via CMake

andeon:
This PR replaces the old workflow where `.qm` translation files were committed directly to the repository. All translation output is now generated during the build using CMake, ensuring a cleaner repository and reducing manual maintenance. This resolves part of https://github.com/opentoonz/opentoonz/issues/3265.

Changes:
- Generates `.qm` files from `.ts` files using `lrelease` at build time.  
- Automatically renames translation folders from language codes to native display names.

The translation system can still be disabled with `-DWITH_TRANSLATION=OFF`. When disabled, no `.qm` files are generated and no language folders are created. In the internal development workflow for Linux and macOS, translations are typically disabled to reduce build overhead, as translation files are generally not needed for development or testing. 

In CI pipelines for macOS and Windows (Nightly or Stable), translations should remain **enabled**, either by omitting the `-DWITH_TRANSLATION=OFF` flag or explicitly setting it to `ON`, to ensure that translation files are generated for builds and releases. This is usually handled by Jenkins CI on external machines, so @shun-iwasawa should ensure that this setup is correctly in place.

Benefits:

Now, @shun-iwasawa no longer needs to commit binary `.qm` files for new translation fixes, keeping the repository clean. Translators only need to edit the `.ts` files inside `toonz/sources/translations`; the `.qm` files will be automatically generated during new builds.

TODO: In the future, I plan to make the CMakeLists.txt more modular to leverage new CMake functions, further simplifying maintenance and improving flexibility.

Ref: https://github.com/opentoonz/opentoonz/issues/3265#issuecomment-621807656


RodneyBaker:
This looks very useful.
I do have one question:
If I follow what you are saying each person either has to submit their translation as a Pull Request or they have to supply the .ts file in a PR/report where someone else will create the PR on their behalf.

While I'm sure he will see this PR anyway, I'm tagging @shun-iwasawa as Reviewer as he is always in the thick of generating translations manually.

andeon:
@RodneyBaker  I’m not sure how the translations were sent at all in your current lifecycle. This PR: https://github.com/opentoonz/opentoonz/pull/4828  shows how chaotic the situation can get under the current system, some people even upload files directly there. But it depends on each translator’s knowledge, which may be why some files were attached directly in those PRs.

With this update, anyone can simply edit the `.ts` translation files, create a PR, and submit it. There’s no need to manually generate the `.qm` file and send it together.

RodneyBaker:
Another big plus to this approach is that translations can be updated at any time whereas before they were only accomplished at the time of 'annual' release.

Added:  The reason translation files were uploaded to the PR was that some didn't know how to generate the required files or navigate the process.  In those cases @shun-iwasawa had those translators supply the files and did the work for them.

This PR can move forward but I do want to make sure this new process makes everything easier for @shun-iwasawa and not harder.  This appears to easily make the process work but Shun is the one that does that work.

andeon:
Close/reopen to trigger the build that previously failed on AppVeyor

RodneyBaker:
In order to keep translations moving forward for release of v1.8 will merge this PR.
Of interest, the old way of adding translations can still be used.
This PR automates that process.

Thanks @andeon for that capability.

RodneyBaker:
I'm guessing there are 200+ untranslated source texts in Opentoonz.

@gab3d I know you may not be keen about translating both T2D and OT** but something I would like to request is that if possible you look at this automated translation generation and see how that effects contributors ability to update translations.

The assumptions is that this new approach should be easier and more accessible than the prior method but we really need to hear from those that do the work of translation.


**Assuming this new approach works well, we likely can share translations between programs where that code is in sync; write once use universally.

shun-iwasawa:
@andeon
So sorry for my delayed response. Thank you for making the repository more lightweight!

One point: since `qt5_create_translation` has been omitted in this PR, it is no longer possible to call `lupdate` and update the .ts files by building the conventional `translation_**` projects.

Although it will become somewhat cumbersome, I think I'll bring back the project for performing `lupdate` in a separate PR.

On the other hand, regarding the operation of calling `lrelease` during the build as implemented in this PR, would it be difficult to integrate it into the source code module itself (i.e., `toonzqt`, rather than `translate_toonzqt`)?

andeon:
@shun-iwasawa  No, it's not difficult, you can do it easily. The same logic would need to be repeated in each module (tnzcore, toonzlib, etc.). To avoid duplication, you can create a shared CMake function in a .cmake file (e.g.,` toonz/make/build_translations.cmake`) and call it in each module. This way, the logic is centralized and can be reused  for example, with `build_module_translations(toonzqt)` in the `CMakeLists.txt`.

Something like this in each module:

```
# toonzqt/CMakeLists.txt
add_library(toonzqt ...)

if(WITH_TRANSLATION)
    include(${CMAKE_SOURCE_DIR}/../cmake/build_translations.cmake)
    build_module_translations(toonzqt)
endif()
```
The old translation block in the main CMakeLists.txt (with `add_translation` and `qt5_create_translation`) would then be removed, keeping the root file cleaner.

shun-iwasawa:
@andeon 
I apologize for the confusing wording; what I proposed was not to move the translation-related functions to each module’s `CMakeLists.txt`.

Currently, the process of `lrelease`-ing ts files to qm files is performed within the `translate_toonzqt` target added by `add_custom_target`.
When you run cmake with the `add_custom_target` command, in a Windows + Visual Studio environment, it generates an additional project named `translate_toonzqt`.

As I mentioned in my previous post, I’m considering bringing back a project specifically for running `lupdate` (let's say, for example, naming it `translate_update_toonzqt`).

However, I was a bit concerned that this would result in a large number of projects named `translate_**` appearing in Visual Studio’s Solution Explorer, which could become cumbersome..

image


That’s why I asked if it would be possible to integrate the process of `lrelease`-ing ts files to qm files into the build process for `toonzqt` itself, rather than creating a new `translate_toonzqt` project to handle it.

However, this is merely a visual issue in Visual Studio. I’m more concerned about potential problems arising from combining everything into one project, so please don’t worry too much about this!

andeon:
Ah, now I understand. You want to avoid creating extra projects in Visual Studio. I think you can use add_custom_command(TARGET toonzqt PRE_LINK ...) inside the CMakeLists.txt file for toonzqt. This way, lrelease runs before linking without creating a separate target.

Something like this inside `toonzqt/CMakeLists.txt`:

```
if(WITH_TRANSLATION)
    set(LANG_CODES japanese italian french spanish chinese german russian korean czech)
    set(GENERATED_QMS "")
    foreach(lang ${LANG_CODES})
        set(ts_file "${CMAKE_SOURCE_DIR}/translations/${lang}/toonzqt.ts")
        set(qm_file "${CMAKE_BINARY_DIR}/../../stuff/config/loc/${lang}/toonzqt.qm")
        if(EXISTS "${ts_file}")
            add_custom_command(
                TARGET toonzqt PRE_LINK
                COMMAND ${CMAKE_COMMAND} -E make_directory "${CMAKE_BINARY_DIR}/../../stuff/config/loc/${lang}"
                COMMAND ${Qt5_LRELEASE_EXECUTABLE} "${ts_file}" -qm "${qm_file}"
                BYPRODUCTS "${qm_file}"
                COMMENT "lrelease: ${lang}/toonzqt.qm"
                VERBATIM
            )
        endif()
    endforeach()
endif()
```
No separate target will appear in Visual Studio and lrelease runs automatically before linking the modules. You can try centralize the logic in a CMake function to avoid repeating code.

PR #6292, opened 2025-11-27
ci: Update AppVeyor configuration to use VS2022 image

andeon:
This PR updates the AppVeyor build configuration to use the Visual Studio 2022 image.

Additionally, some new `environment` variables have been added to facilitate maintainability.

We had been monitoring several intermittent crashes and other anomalies that might have been related to the previous build environment (see ref: [opentoonz/opentoonz#6127](https://github.com/opentoonz/opentoonz/pull/6127)). Had these issues been confirmed, the configuration might have needed to be downgraded.

So far this week, I haven't observed any issues in the current Windows workflow using the custom Qt 5.15.2_wintab build with VS2022. It’s unclear whether the improvement is due to recent code changes, local Windows updates, or updates to the image itself, but the workflow appears stable at the moment.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,7 +1,11 @@
 version: 1.7.1.{build}
 
 environment:
-  QT_PATH: '%APPVEYOR_BUILD_FOLDER%\thirdparty\qt\5.15.2_wintab\msvc2019_64'
+  QT_PATH: "%APPVEYOR_BUILD_FOLDER%\\thirdparty\\qt\\5.15.2_wintab\\msvc2019_64"
+  OPENCV_VERSION: "4.11.0"
+  OPENCV_DLL: "C:\\Tools\\opencv\\build\\x64\\vc16\\bin\\opencv_world4110.dll"
+  BOOST_ROOT: "C:\\Libraries\\boost_1_89_0"
+  MSVC_VERSION: "Visual Studio 17 2022"
 
 init:
   - git lfs install --skip-smudge  # Avoid issue with GitHub LFS quota
@@ -14,7 +18,7 @@ skip_commits:
 pull_requests:
   do_not_increment_build_number: true
 skip_tags: true
-image: Visual Studio 2019
+image: Visual Studio 2022
 configuration: Release
 platform: x64
 clone_depth: 1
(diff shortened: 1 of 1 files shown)


andeon:
~~I’m putting this as a draft because~~ I’m going to try using the `-T v142` parameter. Since we are using third-party libraries built with MSVC 2019 (OpenCV vc16, Qt 5 MSVC2019, Boost MSVC2019), using the v142 toolset should avoid ABI mismatches and is expected to result in faster, more stable builds while still using Visual Studio 2022.  

If no build performance gain is observed, we could continue using Visual Studio 2019 for now, and close this pr.

I changed the paths in the `environment:` section to use double backslashes (`\\`) instead of slashes (`\`) or (`/`) to avoid parsing issues.

After the build, we need to test the [artifact](https://ci.appveyor.com/project/opentoonz/opentoonz/builds/53151833/artifacts) to check if it works as expected.

RodneyBaker:
I am leaning toward merger of this PR as now is a good time to make this adjustment as we approach end of year and start to approach the ramp up into version 1.8.

While we want to remain flexible regarding final changes to the next full release, the more we can lock down, optimize and ensure maximum stability of Opentoonz for users the better.

I mention this as the current release schedule states End of December as Feature Freeze but this date will likely move to accommodate developers.  Where it comes to refactoring, builds and such the feature freeze won't affect that but we do want to proceed with caution as we will need to eventually halt such efforts.

andeon:
@RodneyBaker I hope version 1.8 doesn’t take too long. I believe that this year OpenToonz had a large number of crashes and bugs fixed, but we will have the real proof when this version reaches the general public. That way, we can fix whatever is still missing. Based on the tests and reported bugs, I think stability has improved significantly.

RodneyBaker:
Yes, I've noticed major stability gains in a number of places.
Locking version 1.8 down will allow us to move on toward a full version 2.0 release that moves completely to Qt6 and/or Apple Silicon.  Either of these two would be sufficient to move the version marker to that milestone.

I do think it important to take a moment in 2026 to stop, reflect and celebrate how far Opentoonz has come in ten years.  That doesn't mean we have to delay moving forward... quite the opposite.  Once version 1.8 is released, users can lock in and use that release which will open the path to where we can move boldly forward.

As always, thanks for your work that has not only brought us to this moment but is preparing us for  the future.

RodneyBaker:
> I hope version 1.8 doesn’t take too long.

@andeon 
We are 11 days away from a 'soft feature freeze' so that should help us move toward version 1.8 release.  I am adding the word 'soft' to the generally harder 'Feature Freeze' because I know there are features that do not yet have Pull Requests opened against them.  I do think its important that we hit that milestone however so everyone has a clear understanding that we are running out of time.  Please note that I do not mean this suggests version 1.8 will be out before the actual 10th Anniversary but only that the focus from 1 January 2026 forward will shift.

RodneyBaker:
Pressing forward with this with expectation that VS2022 will build correctly with CI Appveyor.
Depending on success, we will need to work with @shun-iwasawa to do a follow up for the nightly builds.

PR #6293, opened 2025-11-27
Add Windows Toonz Thumbnail Provider

Abel032:
This PR integrates the `toonzpreview.dll` component to enable Windows Explorer thumbnail previews for Toonz files.

## Features:

- Supported Files: `.tnz` (Scene), `.pli` (Vector), `.tlv` (Toonz Raster) and `.tga` (Tagged Graphics/Truevision, this type is optional when install).
- A new Install/Uninstall button is added under `Preferences>Addons`.
- Includes registration for the default `.plt` (Palette) file icon.


RodneyBaker:
This is one of those features I didn't realize we needed but will be so very useful.
Initial testing show this working well.

A few observations:
- The thumbnail often shows the first column and first frame of the level which in many cases (for me at least) is a background image that isn't particularly useful in visually identifying the resource.  I have been considering having column 1 be a poster frame/informational column and this feature might push me more in that direction.
- A few toonz files (such as custom vector brushes) appear perfectly.  This is likely because there is only one image in the file and it's designed to be big, bold and exactly what that file represents.
- If would be really nice if the thumbnail provided could process .bak files but it would actually have to process the toon prefix plus the .bak.  For instance, the preview would need to recognize  '.pli.bak' as a .pli file, a '.tlv.bak' as a .tlv file, etc.  This enhancement to the feature would allow users to see inside backup files in a way that is currently not possible.

Looking very good @Abel032

Abel032:
@RodneyBaker 
Windows can't set the default icon or thumbnail for files like `*.ext.bak`.
For the default index of thumbnail, I think first frame of level and first frame of scene is would be common.But maybe we can provide multiple frames in one pircture?

konero:
In the interface, how about moving the install button to a separate category called 'Addons'? Putting it in General (for those that don't want to use it) might get in the way.

How about changing just a single button to a QGroubBox with a QLabel:
- QGroupBox: "Windows Explorer Thumbnails (Shell Extension)":
- - QLabel: "Enable thumbnails for OpenToonz files (.tnz, .pli, .tlv) in Windows Explorer. Administrator privileges are required; you may need to restart Explorer for changes to take effect."
- - QPushButton: "Install" / "Uninstall"

Abel032:
@konero Okay, let's move it to a new tab page.

Ahl76:
@Abel032 

Krita_issue


I tried the appveyor for this PR, which seems to work as expected, but I received an error message after launching Krita following the installation of the DDL. After reinstalling Krita, I was able to launch it again randomly later, but the error returned after retesting this PR.

Abel032:
@Ahl76 
Thanks for the feedback. I've updated the code. Could you please check if the issue still persists?

Ahl76:
Thanks, the issue seems to be fixed.

Just FYI when Krita is installed with its own addon, uninstalling the OpenToonz addon by clicking “uninstall” doesn’t actually disable it—the thumbnail seems still active even after uninstall it.

It looks like uninstalling Krita itself is needed to make the thumbnail inactive again, even if I don’t think it’s a big issue.

Abel032:
@Ahl76 
It sounds like a refreshing issue, you may need to restart the system or all the explorer.exe to make the uninstallation to take effect.

Abel032:
I think with this addon on we can diable the feature of generating `sceneIcons` folder.

Update: I'm not free these days, so maybe another PR after this ><

Ahl76:
In my case, the problem persists even after restarting. However, the issue does not seem to apply to newly created pli tlv and tnz files

Abel032:
Maybe for windows you have to clean all cached thumbnails after uninstalling the addon.
I guess editing the file and save can clean the thumbnail cache, or use some tools to clean all thumbnail caches of system.

RodneyBaker:
@Abel032 I have some concerns that perhaps you can look into when you get some time.

I am curious what will happen when this Windows Toonz Thumbnail Previewer encounters files created with forks such as Tahoma2D.  Specifically when it encounters scenes that contain features outside the scope of normal toonz scene data.

My thinking is that the previewer is going to crash Windows Explorer if/when it encounters a 'malformed' .tnz file.

I'm trying to test and confirm but issues with incompatibilities between OT and T2D scene files makes that a little difficult.

Added:  In other words, if I attempt to preview an OT .tnz scene out of Opentoonz I can access it without problem.  If I import that same scene into T2D and export it out, then attempt to preview it... Windows Explorer crashes/closes.  At least that is what I seem to be seeing.

RodneyBaker:
As this is an optional install (and can be uninstalled easily) I am leaning toward merging this PR.
Feedback from usage in the field can then dictate what (if any) changes are required.
As such, posting intent to merge.

Abel032:
@RodneyBaker I think I would add some adjustment this month.
But it's okay to merge I think.

- Ensure thumbnail provider won't crash even met wrong type files (I didn't get the crash currently, need to confirm)
- Register the file type rightly so that they can be grouped by explorer

RodneyBaker:
I'll delay merging for now as there is no rush.
Those that want to install the thumbnail previewer/provider can download the artifact and use that to install.

Thanks for the update on your plans.

Ahl76:
Regarding thumbnails, I saw other issues or potential shell extension conflicts, and recently encountered persistent symptoms on Windows 11 that i think linked to this feature.

Initially, some standard folders (e.g., Music) and newly created folders appeared invisible in the Explorer. This issue was resolved by a simple PC restart.

I also experienced a persistent issue with the miniaturization of a system folder (e.g., 'Compressed'). This folder displayed an abnormally tiny icon size. This miniaturization was impossible to correct using standard Windows display methods or by purging the icon cache (iconcache_*.db) via the Windows Terminal.

In my case, attempts to resolve the issue using the DLL uninstall/deactivation button in OpenToonz Preferences or by uninstalling Krita were ineffective.

The problem was resolved by manually deleting the entire OpenToonz build folder that contained the thumbnail component. The affected folder's icon immediately returned to its normal size, which strongly suggests the OpenToonz shell extension DLL was the direct source of the persistent interference.

Abel032:
Thanks @Ahl76  I'll check that.

Abel032:
@Bracket-I-I Of course only for Windows and when addon installed...

Abel032:
> I also experienced a persistent issue with the miniaturization of a system folder (e.g., 'Compressed'). This folder displayed an abnormally tiny icon size. This miniaturization was impossible to correct using standard Windows display methods or by purging the icon cache (iconcache_*.db) via the Windows Terminal.

I couldn't reproduce this on my computer, maybe others get similar result.

Ahl76:
I will try it again to see if i get same result

Ahl76:
Strangely enough, I haven't encountered the issues from my previous tests, at least not until now. I don't know if it was related to specific steps or processes like krita, but so far I haven't had any particular problems. No issues with installation or uninstallation either.

RodneyBaker:
Thanks @Ahl76 for the confirmation.
Pressing forward with this addition.
For what it's worth I've been using this PRs thumbnail provider since initial posting of this PR and have really enjoyed being able to preview OT files without openting them!

Thanks @Abel032 for another useful addition.


Aside:  Does anyone know if MacOS and Linux have an equivalent to this process?

RodneyBaker:
@Abel032 There is a reported problem suggesting this PR may be the cause of a Stage Schematic crash.  I have some thoughts on where the problem could be but as that is pure guesswork will look deeper and share the actual problem.

Steps to replicate crash:
- Create two columns with toonz raster levels
- In the Stage Schematic select both toonz raster column nodes
- Right Click and choose Collapse
This should result in a crash.

A naive question might be whether the m_context might be local only and therefore being lost as it is not passed back.  But surely this shouldn't relate to the actual thumbnail provider extension itself.

PR #6295, opened 2025-11-29
Reorganize Preferences Pages

Abel032:
This PR refactors the existing Preferences Popup Window UI. No new options or features have been added.
The goal is to improve the logical grouping of settings.

## Changes
- Tabs: Renamed several Tab Pages and adjusted the order
- Options: Renamed, moved, reordered, and (re)grouped existing options

**Feedback Welcome**: This is a significant UI change, any feedback or suggestions regarding the new tab names, option groupings and other aspects before merging is welcomed.

| **General** | **Interface** | **Preview/Render** |
| :---: | :---: | :---: |
| 图片 | 图片 | 图片 |
| **Load/Import** | **Saving** | **Decoder/Encoder** |
| 图片 | 图片 | 图片 |
| **Drawing** | **Tools** | **Xsheet** |
| 图片 | 图片 | 图片 |
| **Onion Skin** | **Animation** | **Auto Lip-Sync** |
| 图片 | 图片 | 图片 |
| **Colors** | **Vector Visualize** | **Version Control** |
| 图片 | 图片 | 图片 |
| **Touch/Tablet Settings** | | |
| 图片 | | |

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -472,7 +472,7 @@ public:
   bool previewAlwaysOpenNewFlipEnabled() const {
     return getBoolValue(previewAlwaysOpenNewFlip);
   }
-  bool fitToFlipbookEnabled() const { return getBoolValue(fitToFlipbook); }
+  bool fitToFlipbookEnabled() const { return getBoolValue(fitToFlipbookWhenPreview); }
   bool isGeneratedMovieViewEnabled() const {
     return getBoolValue(generatedMovieViewEnabled);
   }
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -166,7 +166,7 @@ enum PreferencesItemId {
   blankColor,
   rewindAfterPlayback,
   previewAlwaysOpenNewFlip,
-  fitToFlipbook,
+  fitToFlipbookWhenPreview,
   generatedMovieViewEnabled,
   shortPlayFrameCount,
 
--- a/toonz/sources/toonz/preferencespopup.cpp
+++ b/toonz/sources/toonz/preferencespopup.cpp
@@ -1201,7 +1201,6 @@ QString PreferencesPopup::getUIString(PreferencesItemId id) {
 
   const static QMap<PreferencesItemId, QString> uiStringTable = {
       // General
-      {defaultViewerEnabled, tr("Use Default Viewer for Movie Formats")},
       {rasterOptimizedMemory, tr("Minimize Raster Memory Fragmentation*")},
       {autosaveEnabled, tr("Save Automatically")},
       {autosavePeriod, tr("Interval (Minutes):")},
@@ -1214,7 +1213,7 @@ QString PreferencesPopup::getUIString(PreferencesItemId id) {
        tr("Replace Toonz Level after SaveLevelAs command")},
       {backupEnabled, tr("Backup Scene and Animation Levels when Saving")},
       {backupKeepCount, tr("# of backups to keep:")},
-      {sceneNumberingEnabled, tr("Show Info in Rendered Frames")},
+      {sceneNumberingEnabled, tr("Add Info water mark in Rendered Frames")},
       {watchFileSystemEnabled,
        tr("Watch File System and Update File Browser Automatically")},
       //{ projectRoot,               tr("") },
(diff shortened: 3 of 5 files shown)


RodneyBaker:
@Abel032 Now is indeed a good time to consider how best to organize Preferences.
Thanks for taking on the task of getting this UI update started.

DarrenTAnims:
I really like most of the changes made here. I think it groups them together better and in a better order and has been long overdue. As with any UI change, there are always personal opinions involved, so feel free to ignore any/all of these suggestions :-) I know you will have thought about this for a long time already, so I've got to hand it to you. Great job!

Codec section:
This used to be called Import/Export. I'm not cure codec means much to most users, whereas import/export does, but it also involves rendering, so it might be better as: Import/Render Setup.
Also, I don't use it, but the "Fast Render Path" seems to not be here.

Tools section:
I like that in previous sections you've put the groupings of options together, but in this one, you've got the "Minimize savebox after editing" in between the Fill too & Brush cursor options. I feel like it would be better, as a lonely option, to sit with the other tickbox options at the very bottom of the page, before the temp keypress length time.
Also, why is the Vector Snapping option in the middle of the screen and not to the left? I can see that the left side of the dropdown aligns with the one at the top, but it seems to just be sitting in the middle for a reason, like it's a different section.

Xsheet section:
There's a lot of options here. I'm pleased you've grouped them more 😁

Version Control:
I feel like it might be worth mentioning the version control systems you can connect to and maybe add a link to show the page in the help guide for setting it up?

Finally, the order of the sections. I know it's often a personal preference, but while I'm adding feedback...
I'd be tempted to try to have the sections follow more logically how you'd use them. Put version control with the Loading & Saving (as that's what it's about). Then follow that with Preview/Render, and then the Codec section, renamed, as you'll need that for rendering. Onion Skin after Drawing. And the other options at the bottom.
So if I was changing it to make more sense for me, personally, I'd basically choose this:

Basic Options (renamed from General).
Interface
Loading
Saving
Version Control
Preview/Render
Import/Render Setup (renamed from codec)
Tools
XSheet
Drawing
Onion Skin
Visualizing Vectors (renamed)
Animation
Auto Lip-Sync
Colors
Touch/Tablet Settings

Abel032:
> This used to be called Import/Export. I'm not cure codec means much to most users, whereas import/export does, but it also involves rendering, so it might be better as: Import/Render Setup.

I think `Import/Export` overlaps with `Loading`/`Saving`, and `Codec` and it's a advanced page so I think it's okay.
User can read the guide text in this page to understand how it works.

> Also, I don't use it, but the "Fast Render Path" seems to not be here.

This option has been moved to `Preview/Render`.

> you've got the "Minimize savebox after editing" in between the Fill too & Brush cursor options. I feel like it would be better, as a lonely option, to sit with the other tickbox options at the very bottom of the page, before the temp keypress length time.

The main consideration is to comparsie with the option "Use the TLV SaveBox to Limit Filling Operations" on the top of it, it's also a feature that only belongs to Toonz Raster Level.

> Also, why is the Vector Snapping option in the middle of the screen and not to the left? I can see that the left side of the dropdown aligns with the one at the top, but it seems to just be sitting in the middle for a reason, like it's a different section.

I also wondering, the code in this section is not different.

> I feel like it might be worth mentioning the version control systems you can connect to and maybe add a link to show the page in the help guide for setting it up?

Yes, it would be better, but maybe future TvT

> Put version control with the Loading & Saving (as that's what it's about).

I put it at the buttom because it's an advanced feature and not often to be used by most users.

RodneyBaker:
My thought on the Codec tab:

FFMPEG is of such importance (and likely will become moreso in the future) that I'd be in favor of naming that tab 'ffmpeg'.  This also because in the future there will likely be more entries related to ffmpeg.  Having that dedicated space may encourage developers to lean in that direction as well.  While outside the scope of this PR there should be a link or button to take the user to ffmpeg.org

All this to suggest that while Opentoonz does not include ffmpeg in its distribution we gladly acknowledge how useful that software is to our development efforts.

We do however want to assist users in finding and placing the ffmpeg distro that meets their needs.  Your recent addition that allows Opentoonz to find ffmpeg on a users system is an excellent example of that.

Aside:  Darrent offered 'Visualizing Vectors' and I'll suggest a similar 'Vector Visualization' as we appear to have enough room for that.   ;)

Abel032:
Remove WIP tag, suggestions still welcomed!

Abel032:
- import / export
- codec
- midea processing
- thirdparty
- ffmpeg
- import / render setup

Abel032:
> Maybe you are not interested in using opentoonz and are just developing for it, for experience, credit and clout...but the users have to see this thing for the rest of their opentoonz 'usage carreer'

I'm also an OT user 😞

Abel032:
Updated the tab names.
`Load` -> `Load/Import`
`Import/Export` -> `Decoder/Encoder`

RodneyBaker:
Posting intent to merge.
We want to give this change plenty of time for users to explore and for translators to tweak respective languages as they see fit.

Thanks @Abel032 for the reorganization of Preferences

Issue #4011, opened 2021-07-13
Random Wave FX not working properly


Arsyudio:
**Describe the bug**
Random wave fx not working properly or even not working at all.

**To Reproduce**
List steps to reproduce the behavior.
1. Create a sub x-sheet column that contain animated character/object that using plastic tool
2. Add Random wave fx on the FX schematic and set the values on the first frame and the last frame of the sub x-sheet
3. Preview the animation

**Expected behavior**
Random wave FX should have been work by setting up values on the first and last frame of the sub x-sheet. Before version 1.5 (and the nightly builds), random wave just works perfectly fine.

**Screenshots**
Here's a video about it :

https://user-images.githubusercontent.com/87380672/125476609-83265a4f-1144-42b9-a9c0-26adab1801ea.mp4

**Device Information:**
 - OS: Windows 10 21H1 OS Build 19043.1083
 - OpenToonz Version: Nightly 2021-06-30
 - GPU: Nvidia MX130
 - GPU Driver Version: 457.63

RodneyBaker:
Most FX requires use of Preview and it isn't clear from the video that you are using Preview.
Can you verify that please.

Note that I assume you've accounted for this but because it isn't documented specifically I ask the question...

Arsyudio:
I did. I was using render>preview to seethe final result. Unfortunately, It
didn't work properly. It only works on sub x-sheet columns that don't
include any plastic tool in it.
Regards,

*Ardy Arsyudio*
2D Animator, Illustrator
& Cartoonist @ardy.arsyudio

RodneyBaker:
Sorry for the delay in returning to this issue.
I am also having issues with plastic mesh and Random Wave FX but even more than that I'm just having issues with Plastic Meshes...

Will have to investigate more.

RodneyBaker:
I believe the best we can do in the short term is recommend workarounds until the problem can get fixed.
The two workarounds I can think of:

1) Use an earlier release of Opentoonz that doesn't have the problem for that specific part of the production.

2) Render the image sequence out with mesh animation but without FX then set up a secondary scene that uses the rendered sequence piped through the FX.

If you have any more information on what what the last release was that works as it should that will be appreciated.
It appears you have identified the last nightly release of v1.4 as working but perhaps it isn't the nightly but an earlier release.

RodneyBaker:
Interestingly, as soon as I posted the previous comments I tried again and got my setup to work.
[...]

PR #6299, opened 2025-11-30
Fix Random Wave FX on sub-xsheets with frame rounding

andeon:
This PR fixes the Random Wave FX on sub-xsheets. The FX was using (int)frame, which truncated fractional frame values and caused incorrect row lookups. It replaces (int)frame with tround(frame) to ensure proper frame rounding, restoring correct behavior when keyframing on sub-xsheets.

Previously, users had to duplicate the Random Wave FX node in the sub-xsheet or use separate nodes inside and outside as a workaround.

Only fractional frames are affected. No collateral issues are expected, but feedback is welcome.

Fixes: https://github.com/opentoonz/opentoonz/issues/4011


RodneyBaker:
Nice catch @andeon.
Thanks for the fix!

PR #6312, opened 2025-12-14
Fix onion skin not updating for area filler

Abel032:
Fix the issue that onion skin option not updating for area filler(Rectangle, FreeHand...).

How to reproduce:
Turn on/off `onion skin` dirctly and do not change any other options after that, then try to fill.


RodneyBaker:

PR #6313, opened 2025-12-14
Fix csv export stop marker

Abel032:
Fix corrupted encoding of “×”.


RodneyBaker:

PR #6317, opened 2025-12-17
Refactor: Modernize Qt signals/slots and improve AutoLipSyncPopup & LipSyncPopup

andeon:
This PR continues the modernization of OpenToonz’s Qt codebase, focusing on eliminating legacy `QSignalMapper`, old `SIGNAL/SLOT` macros, and potential memory leaks. See: https://github.com/opentoonz/opentoonz/issues/6311

While modernizing the code, significant slowdowns in AutoLipSyncPopup and LipSyncPopup on Linux (and possibly on macOS) were discovered and subsequently fixed.

Changes include:

- **Completely removed** `QSignalMapper` **from the codebase** and replaced old `SIGNAL/SLOT` macros with type-safe lambdas and function pointers in the following files:
```
toonz/sources/toonz/tooloptionsshortcutinvoker.cpp,
toonz/sources/toonz/sceneviewercontextmenu.cpp,
toonz/sources/toonz/autolipsyncpopup.cpp, and
toonz/sources/toonz/lipsyncpopup.cpp
```
- Refactored AutoLipSyncPopup and LipSyncPopup, improving Linux performance for both popups
- Removed unnecessary sender() usage, unnecessary returns/asserts, and added clearer, comprehensive English comments
- Replaced deprecated QLayout::setMargin() with setContentsMargins() for proper layout handling
- Updated thumbnail generation using IconGenerator asynchronously; vector thumbnails that were previously cropped now display correctly

**Before / After:**
lipsync

ref: https://github.com/opentoonz/opentoonz/pull/6310

PR #6319, opened 2025-12-18
Fix AutoLipSync hang and cancellation with asynchronous Rhubarb

andeon:
This PR complements PR #6317, which refactored LipSyncPopup and AutoLipSyncPopup. It addresses long-standing issues in the Auto Lip Sync tool and improves user experience.

- Executes Rhubarb asynchronously, preventing UI freezes on large or slow audio files

- Enables safe user cancellation and proper cleanup of temporary files. In the legacy implementation, closing the Auto Lip Sync popup did not cancel the Rhubarb process, which continued running in the background, potentially confusing users and consuming system resources

- Adds a modal progress dialog with a Cancel button, disabling the close (“X”) button to avoid accidental closure

autolipsync


RodneyBaker:
Proceeding...

Thank you @andeon for the additional fix in Rhubarb AutoLipSync!

Issue #6311, opened 2025-12-13
Modernize Qt/C++ usage in OpenToonz (Post C++17 Migration)


andeon:
OpenToonz is already compiled using C++17, which is great progress. However, a significant portion of the codebase still relies on legacy Qt APIs and pre-modern C++ patterns.

This plan proposes a gradual cleanup and modernization, leveraging C++17 and modern Qt best practices without changing behavior or introducing breaking changes. These improvements also simplify a potential future migration to **Qt 6**.

The goal is to:
- Silence deprecation warnings
- Improve readability and maintainability
- Prepare the codebase for a smoother future transition

## Scope of Proposed Changes

Some changes have already started gradually, but this topic serves as a central record for the effort.

### ~~1) QSignalMapper -> Lambdas in connect()~~ ✅
> 
> ~~QSignalMapper is obsolete and no longer needed with modern signal/slot syntax.~~
> 
> ~~Example file: tooloptionsshortcutinvoker.cpp~~
> 
> ~~**Before:** `QSignalMapper *mapper = new QSignalMapper(this); connect(action, SIGNAL(triggered()), mapper, SLOT(map())); mapper->setMapping(action, id); connect(mapper, SIGNAL(mapped(int)), this, SLOT(onTriggered(int)));`~~
> 
> ~~**After (lambda-based):** `connect(action, &QAction::triggered, this, [this, id]() { onTriggered(id); });`~~
> 
> ~~Additional notes: `#include <QSignalMapper>` can be removed. Lambdas are type-safe, simpler, and more readable.~~

### ~~2) `setMargin()` -> `setContentsMargins()`~~  ✅

> ~~`QLayout::setMargin()` is deprecated.~~
> 
> ~~**Before:** `layout->setMargin(0);`~~
> 
> ~~**After:** `layout->setContentsMargins(0, 0, 0, 0);`~~

### ~~3) `Q_FOREACH / foreach` -> C++ range-based `for`~~ ✅

> ~~Qt’s `Q_FOREACH` and `foreach` macros are obsolete and unnecessary in a C++17 codebase, since C++ has supported range-based `for` loops since C++11.~~
> 
> ~~**Before:** `foreach (const XdtsFrameDataItem &data, m_data) { ... }`~~
> 
> ~~**After:** `for (const XdtsFrameDataItem &data : m_data) { ... }`~~

### ~~4) `QRegExp` -> `QRegularExpression`~~ ✅

[...]

PR #6320, opened 2025-12-18
Replace deprecated QLayout::setMargin() with setContentsMargins()

andeon:
This PR continues the modernization of OpenToonz’s Qt codebase, focusing on replacing all occurrences of the deprecated `QLayout::setMargin()` calls with `setContentsMargins()`.  See: [#6311](https://github.com/opentoonz/opentoonz/issues/6311)

- Replaced `setMargin(0)` with `setContentsMargins(0, 0, 0, 0)`.
- No functional changes, purely modernization.

**Note:** The `CMakeLists.txt` was used to find deprecated (legacy 32-bit) files from the active code. Legacy files, such as `cameracapturelevelcontrol_qt.cpp` and `penciltestpopup_qt.cpp`, are only used in 32-bit builds, which are no longer officially supported. The `setMargin` was not updated there. In the future, these files should be moved from, or commented out in, the `CMakeLists.txt` for later exclusion.

For example, `setMargin` was not updated in the file `toonz/studiopaletteviewer.cpp` because it is dead code and no longer used. The active file, `toonzqt/studiopaletteviewer.cpp`, has already been updated. To better organize the repository and preserve legacy or orphaned (dead code) files for reference, it may be helpful to create a `/deprecated` folder inside each relevant source directory and move these files there first. Later, they can be clearly excluded from the repository once it is decided how these files will be handled.

### DIFF
--- a/toonz/sources/stopmotion/stopmotioncontroller.cpp
+++ b/toonz/sources/stopmotion/stopmotioncontroller.cpp
@@ -338,12 +338,12 @@ StopMotionController::StopMotionController(QWidget *parent) : QWidget(parent) {
 
   QVBoxLayout *controlLayout = new QVBoxLayout();
   controlLayout->setSpacing(0);
-  controlLayout->setMargin(5);
+  controlLayout->setContentsMargins(5, 5, 5, 5);
 
   {
     {
       QGridLayout *camLay = new QGridLayout();
-      camLay->setMargin(0);
+      camLay->setContentsMargins(0, 0, 0, 0);
       camLay->setSpacing(3);
       {
         camLay->addWidget(new QLabel(tr("Camera:"), this), 0, 0,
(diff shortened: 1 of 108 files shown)


andeon:
Pull request close/reopened to restart the macOS CI build that previously failed in Formulae.

**Note**: This manual intervention (closing/reopening) could be avoided by using “**Re-run jobs**” in GitHub Actions for the workflow that failed. This option restarts the workflow from scratch and usually resolves issues like this, especially if it was just an “apparent hang” during installation. However, “**Re-run jobs**” is limited to maintainers or users with admin permissions.

RodneyBaker:
> This manual intervention (closing/reopening) could be avoided by using “Re-run jobs” in GitHub Actions for the workflow that failed. 

If only that were ALWAYS true.
In the case of the Mac OS Mojave checks they have been getting stuck in such a way that doesn't allow them to be restarted.  Further, if we look in the github actions listing that build isn't even listed.  This is why at one point I asked if that Mojave check was even needed.  For me at least, the option just isn't there to restart that build specific build.

You are correct though that in most cases (all other cases?) those with premissions can restart the failed builds (either one at a time or all together).

Added:  I should also state that I personally cannot restart Appveyor checks.  As far as I know only two people (and the author of the specific PR) can do that Appveyor refresh (@shun-iwasawa and @manongjohn).

andeon:
Yeap, `macOS (Mojave)` and `Windows (new)` are managed by an external environment. I believe this is a DWANGO-maintained machine, managed exclusively by @shun-iwasawa. 

I’m adding this mental note as a reminder of the current setup, since I sometimes forget these details, and I also forgot that AppVeyor can be refreshed if someone has the necessary permissions.

RodneyBaker:
Thanks for the additional confirmation @andeon 
That makes sense.

As for this PR... LGTM.  Thanks!

PR #6321, opened 2025-12-19
ci: optimize macOS build workflow

andeon:
This PR updates the macOS GitHub Actions workflow to avoid upgrading all installed Homebrew packages on every run. Upgrading all packages can sometimes break the build because Homebrew formulae change frequently, and updates might introduce incompatibilities.

- Increase job timeout to 180 minutes for better reliability
- Only install missing packages, skip unnecessary upgrades
- Remove premature 'brew cleanup' to preserve caches and speed up repeated runs

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -20,7 +20,7 @@ on:
 
 jobs:
   macOS:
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Hopefully this will help with the MacOS Mojave CI builds.
Thanks @andeon 
LGTM

PR #6326, opened 2025-12-20
Fix scene thumbnail backgrounds

JParks123:
Added white Backdrop for Transparent images.

image
     
 Co Authored by: @justburner @manongjohn


RodneyBaker:
Thanks for the port @JParks123 
That looks much better.
LGTM

PR #6336, opened 2025-12-25
Increase 3D viewer scale.

JParks123:
This PR sets the 3D viewer scale closer by default.
Resetting the 3D view also scales it back to default size.


RodneyBaker:
(Note:  This PR is ported from T2D)

Thank you for this very subtle yet useful update @manongjohn and @JParks123
Much appreciated.

PR #6338, opened 2025-12-26
Fix pasting external clipboard image

JParks123:
Resize pasted external clipboard image on new level creation


JParks123:
I just read Justburners initial PR for this and realized it was not made to fit into the scene on purpose to maintain clear resolution.
so I will abandon this PR.

PR #6379, opened 2025-12-27
 Display file paths in Scene Cast

JParks123:
This PR adds the path column to the Scene Cast to make it easy to see where all your assets are, especially if you Load assets instead of Import and/or if the level name differs from the actual file name. Path will also be in the Export File List.

 this is a Tahoma port: [Thanks MJ](https://github.com/tahoma2d/tahoma2d/pull/639)


RodneyBaker:
This may not appear in the nightlies for awhile but pressing forward.

Thanks for the useful addition @manongjohn and @JParks123.

PR #6380, opened 2025-12-27
Reorganizing Command bar Checks Layout and inclusion of Vector guided controls

JParks123:
this PR reorganizes the Command bar checks layout, promoting the ones that work together as features side to side while adding in the Black BG check.

image
image
image

### DIFF
--- a/stuff/profiles/layouts/settings/commandbar.xml
+++ b/stuff/profiles/layouts/settings/commandbar.xml
@@ -44,13 +44,11 @@
     <separator/>
     <command>MI_ACheck</command>
     <command>MI_GCheck</command>
-    <separator/>
-    <command>MI_BCheck</command>
     <command>MI_TCheck</command>
-    <separator/>
-    <command>MI_PCheck</command>
-    <separator/>
+    <command>MI_BCheck</command>
     <command>MI_IOnly</command>
+    <separator/>
     <command>MI_ICheck</command>
+    <command>MI_PCheck</command>
     <separator/>
 </commandbar>


RodneyBaker:
I see the additions to and reordering of the commandbar.xml‎ file.
Are there actually fixes you still plan to add to this PR?

JParks123:
> I see the additions to and reordering of the commandbar.xml‎ file. Are there actually fixes you still plan to add to this PR?

I can retitle it to reorganizing checks layout, as to promote hidden features. I also added the Vector guided Drawing controls.

konero:
Can we please not keep reordering the interface? It's better to just place the Black BG check between the current Transparency + Fill check position.

JParks123:
> Can we please not keep reordering the interface? It's better to just place the Black BG check between the current Transparency + Fill check position.

the Transparency check and the Black BG check when turned on at the same time changes the initial black outline white, this helps to fill empty uncolored areas that are black from the check, I will reorder for you @konero 

image

konero:
It's still illogical, based on what you're showing. UX-wise you group by type, not workflow. So it should be:
```
Gap Check
Fill Check
Transparency Check
Black BG Check
Inks Only
==========
Ink Check
Paint Check
```

JParks123:
> It's still illogical, based on what you're showing. UX-wise you group by type, not workflow. So it should be:
> 
> ```
> Gap Check
> Fill Check
> Transparency Check
> Black BG Check
> Inks Only
> ==========
> Ink Check
> Paint Check
> ```

Ok I can set it to that list you posted.

stuff/profiles/layouts/settings/commandbar.xml
```diff
@@ -25,6 +25,9 @@
     <command>MI_EnterGroup</command>
     <command>MI_ExitGroup</command>
     <separator/>
+    <command>MI_VectorGuidedDrawing</command>
```
konero: In my opinion we don't need this as the default, these are very advanced features and sadly they are undocumented. I've tried to create documentation myself, but there are many shortcomings outside of that, that I feel it's too advanced and cumbersome to present to general users, the feature is unfinished. There's many prerequisites the user has to do before learning how to use these. I suggest leaving these out for now. This goes for the Drawing Controls panel as well (the dockables don't need to be commands on the panel).

RodneyBaker:
Proceeding with this update.  
We can tweak further as needed.

Thank you @JParks123 for adding the missing controls.

PR #6381, opened 2025-12-28
Refactor: Modernize Qt/C++ (foreach, QRegularExpression, signals/slots)

andeon:
This PR continues the modernization of OpenToonz’s Qt/C++ codebase, with a primary focus on replacing all deprecated `foreach` macros. Although replacing `QRegExp` with `QRegularExpression` and modernizing legacy `SIGNAL/SLOT` macros were initially planned as separate tasks, they have been included here to avoid revisiting the same files multiple times. See: #6311

- Replace deprecated `foreach` macros with C++ range-based for loops.
- Replace `QRegExp` with `QRegularExpression`.
- Modernize legacy `SIGNAL`/`SLOT` macros to the Qt5 connection syntax.
- Fix `_tick2Id` being set twice and `_referenceFrameId` not initialized in `xdtsio.cpp`, which could lead to incorrect reference frame marks during XDTS export.

### DIFF
--- a/toonz/sources/toonz/custompaneleditorpopup.cpp
+++ b/toonz/sources/toonz/custompaneleditorpopup.cpp
@@ -1,3 +1,5 @@
+
+
 #include "custompaneleditorpopup.h"
 
 // Tnz includes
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Proceeding with this although I don't think nightlies are building at the moment.
As such it may be awhile before the changes appear in nightlies.

Thank you @andeon for the refactoring/modernization.

PR #6386, opened 2025-12-31
Fix: TCubic::getSpeed used an incorrect formula

ideasman42:
The Bézier derivative formula itself is standard calculus (Bernstein basis polynomial of degree 3).

Reference wikipedia: https://en.wikipedia.org/wiki/B%C3%A9zier_curve

---

Note: this reads better when wrapping is forced: e.g.

Before:

```.cc
  return 3.0 * ((m_p1 - m_p0) * s * s + 2 *
                (m_p2 - m_p0) * s * t + /* <- m_p0 is used a second time! */
                (m_p3 - m_p2) * t * t);
```

After:
```.cc
  return 3.0 * ((m_p1 - m_p0) * s * s + 2 *
                (m_p2 - m_p1) * s * t +
                (m_p3 - m_p2) * t * t);
```

---

The curve fitting will be higher quality, from a test in: https://github.com/ideasman42/curve-fit-nd

This shows a before and after (where after can fit the curve within the error margin using one less "vertex"):

Before:
test_curve_freehand_01_before

After:
test_curve_freehand_01_after

### DIFF
--- a/toonz/sources/common/tgeometry/tcurves.cpp
+++ b/toonz/sources/common/tgeometry/tcurves.cpp
@@ -338,7 +338,7 @@ TPointD TCubic::getPoint(double t) const {
 //-----------------------------------------------------------------------------
 TPointD TCubic::getSpeed(double t) const {
   double s = 1 - t;
-  return 3.0 * ((m_p1 - m_p0) * s * s + 2 * (m_p2 - m_p0) * s * t +
+  return 3.0 * ((m_p1 - m_p0) * s * s + 2 * (m_p2 - m_p1) * s * t +
                 (m_p3 - m_p2) * t * t);
 }
 //=============================================================================


Abel032:
Curious about what would this fix.
For example, would it affect some tool and fix its behaviour?

andeon:
I think this fix corrects tangent calculations, improving vector stroke quality. It may affect tools like Brush, Eraser, and others that rely on thick or variable-thickness styles, resulting in more accurate outlines, smoother curves, and better stroke joins, I guess.

ideasman42:
Updated the PR with before/after comparisons.

RodneyBaker:
Thank you @ideasman42 for the fix.

It's nice to have you contributing to Opentoonz again.
LGTM

Issue #1885, opened 2018-03-22
Schematic:  Renamed nodes still display original names


gab3d:
Refresh problems of 3 different intensities are currently present, related to the renaming of nodes in the Stage Schematic:

1. Cosmetic:
Renamed objects are not displayed by their new name in Xsheet, until:
- another column node is selected in the schematic.
- the mouse pointer goes over the Xsheet or Palette areas.
- you click on any other UI area (on the outside of the currently active stage schematic).

2. Somewhat problematic:
The Animate Tool's selected object dropdown list does not refresh upon a node name change until:
- another tool is activated and then back to Animate tool.
- any parent relationship is altered in the schematic.
- the scene is reloaded.

3. Problematic:
The Xsheet's Parent slot text always shows the parent object's original name (no matter what type of object it is: pegbar, column or camera).

And , finally, while looking at the above issues, I've noticed another "name related" problem:
Whereas the Table is correctly displayed by its translated name in the Stage Schematic, it's always displayed by its English name in the following places:
- Viewer: right click menu "Select Table" option.
- Animate tool: currently selected object dropdown list.
- Xsheet: Parent slot text.

As always, thanks for listening!

artisteacher:
I don’t usually rename the nodes, so I hadn’t noticed this. Interestingly, renaming columns in the Xsheet/Timeline immediately updates the name in the Animate Tool’s tool option and in the Stage Schematic.

RodneyBaker:
Is this still an issue in current nightly releases?

RodneyBaker:
Confirmed that issues identified here are still present in current releases of of this date.

RodneyBaker:
There are some aspects of this report that appear to be resolved.
For instance, No 3... if the user has the Preferences option Xsheet > Link Column Name with Level (as well as other updates) appears to have resolved this.

Not sure aobut the other parts... but I do think in the case of No 1 they refresh more easily.

No 2 seems to be resolved.

RodneyBaker:
Closing as at least partially resolved.
@gab3d Recommend opening a new report for areas still problematic (if any).

PR #6406, opened 2026-01-04
Fix node rename not updating instantly in Animate Tool

andeon:
This PR partially addresses #1885: Stage Schematic node renaming now updates immediately on the node itself and in the Animate Tool dropdown. There is no need to switch tabs, create nodes, or reload the scene to force a refresh.

- As part of the modern Qt/C++ refactor [#6311](https://github.com/opentoonz/opentoonz/issues/6311), all `signal/slot `connections were updated to the Qt5 connection syntax.
- Renaming is now asynchronous.
- Clean up dead code, remove unnecessary asserts, and add defensive checks where needed.
- Fixed a Clang analyzer memory leak warning.

Note: `StageSchematicGroupNode` still has a pre-existing bug and could potentially be refactored in the future to follow the SSOT principle, which seems like the best solution I can think of, although someone else might have a better idea. Currently, expanding or collapsing a group behaves as if it has two separate states. If two Stage Schematic windows are open, one with the group expanded and the other collapsed, changing the group name in one can crash the other because their states are not synchronized. This scenario is unusual, as it’s unlikely someone would open two Stage Schematic windows and edit the same group simultaneously in different states, but it’s still a bug anyway.

### DIFF
--- a/toonz/sources/include/toonzqt/stageschematicnode.h
+++ b/toonz/sources/include/toonzqt/stageschematicnode.h
@@ -404,6 +404,10 @@ signals:
   void currentColumnChanged(int index);
   void editObject();
 
+public slots:
+  // Public wrapper slots for modern Qt5 connection syntax
+  void onHandleReleasedPublic() { onHandleReleased(); }
+
 protected slots:
   void onHandleReleased();
 };
@@ -567,7 +571,7 @@ protected slots:
 
 //========================================================
 //
-// class StageSchematicSplineNode
+// class StageSchematicGroupNode
 //
 //========================================================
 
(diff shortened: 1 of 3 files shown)


RodneyBaker:
@andeon When you get a chance can you close and reopen this PR to refresh the Appveyor check?
Thanks!

Interesting bug you've detailed and one we definitely should look into.
It's great that we can have multiple windows of the same thing open at once but the potential for conflict is real.

RodneyBaker:
Apparently something else going on with Appveyor here.

andeon:
I’m going to run `git commit --amend` to try force it

RodneyBaker:
Looks like that did the job.
Thanks @andeon for the fix.
As always, much appreciated!

PR #6409, opened 2026-01-08
Show Column Parent's Color context menu option

JParks123:
Shows Column Parent's Color context menu option

### DIFF
--- a/toonz/sources/toonz/xshcolumnviewer.cpp
+++ b/toonz/sources/toonz/xshcolumnviewer.cpp
@@ -3080,23 +3080,25 @@ void ColumnArea::contextMenuEvent(QContextMenuEvent *event) {
     if (!xsh->isColumnEmpty(col)) {
       menu.addAction(cmdManager->getAction(MI_ReplaceLevel));
       menu.addAction(cmdManager->getAction(MI_ReplaceParentDirectory));
+    }
 
-      // if (containsVectorLevel(col)) {
-      //  menu.addSeparator();
-      //  QAction *setMask =
-      //      new QAction(tr("Temporary Mask (Not in final render)"), this);
-      //  setMask->setCheckable(true);
-      //  setMask->setChecked(xsh->getColumn(col)->isMask());
-      //  setMask->setToolTip(
-      //      tr("Only Toonz Vector levels can be used as masks. \n Masks don't
-      //      "
-      //         "show up in final renders."));
-      //  bool ret = true;
-      //  ret      = ret &&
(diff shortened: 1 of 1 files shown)


RodneyBaker:
For more information on this fix see:  https://github.com/tahoma2d/tahoma2d/pull/1924

konero:
I'm concerned about adding a checkbox directly to the top-level context menu, which adds a stateful decision into a list that's usually made of one-off actions. Menus work best when they contain the most commonly used, single-action items. A checkbox adds a persistent state and adds visual/mental overhead to what should be a quick interaction.

How about putting it into a sub-menu: [View]?

JParks123:
> I'm concerned about adding a checkbox directly to the top-level context menu, which adds a stateful decision into a list that's usually made of one-off actions. Menus work best when they contain the most commonly used, single-action items. A checkbox adds a persistent state and adds visual/mental overhead to what should be a quick interaction.
> 
> How about putting it into a sub-menu: [View]?

I think it works best in the top menu like in Tahoma, since it changes the color of Xsheet column
 dropdowns.
the Navigation tags toggle is there too.   

https://github.com/user-attachments/assets/9f5fdb11-5c38-48aa-a1c5-69ae22fbd7b3

RodneyBaker:
@konero  Yours is a valid concern and one we need to address.
In short, I agree with you.

However,
This PR lies outside of your concern on several levels although I'll try to be brief and just hit the highlights.

This PR is a fix of code created by @manongjohn for Tahoma2D.
As a direct port with no changes adjusting the menu item to a submenu is a little out of scope.

The author of the port doesn't have the skill and experience to implement your suggestion.
Not without assistance.

The suggested change really needs to get buy in from @manongjohn and other developers so original code doesn't have to be refactored every time a port is created.  Developers need to understand the importance of your concern, namely:

> Do not add stateful decisions into the context menu usually reserved for one-off actions. Menus work best when they contain the most commonly used, single-action items. For example, a checkbox adds a persistent state and adds visual/mental overhead to what should be a quick interaction.

There may be such guidance already in place but if so it hasn't been highlighted well enough.
This PR should serve in part to emphasize how best to proceed.

In short, this PR will likely proceed not due to lack of addressing your concerns but because it is a fix ported directly from Tahoma2D.  Moving inappropriate menu items to submenus can be affected via further PR to alleviate any concern. 

I hope you find this acceptable.

RodneyBaker:
Moving forward with this PR.
Thanks @JParks123 for the Tahoma2D port.
LGTM

PR #6410, opened 2026-01-08
Fix Control Point Editor Pen Lag

JParks123:
This fixes pen lag in the Control Point Editor tool when there is a large number of strokes in the frame.
this is a T2d fix.


RodneyBaker:
REF:  https://github.com/tahoma2d/tahoma2d/pull/1869

RodneyBaker:
Thank you for the port @JParks123 
LGTM

PR #6413, opened 2026-01-09
Fix Flow Paint Brush Iwa Fx Swatch Viewer Crash

JParks123:
This fixes a crash that happens when you turn on the Preview or Camera Preview in the Flow Paint Brush Iwa Fx's Swatch Viewer.

Crash was caused when rendering to an offline GL Context but no surface created for it to use. A surface is now created.

Tahoma2D port: https://github.com/tahoma2d/tahoma2d/pull/1501


RodneyBaker:
Thank you @JParks123 for another successful port.

PR #6416, opened 2026-01-09
Fix enabling tool options on Vector Select All

JParks123:
When you select all on a vector level it doesn't activate the attribute boxes in the tool option bar.

https://github.com/tahoma2d/tahoma2d/issues/1596

T2d port:  https://github.com/tahoma2d/tahoma2d/pull/1626


RodneyBaker:
Another excellent fix from @manongjohn ported by @JParks123 
Much appreciated.
LGTM

PR #6417, opened 2026-01-10
Hide global set keys when no skeleton created

JParks123:
Hide global set keys when there's no skeleton created to prevent crashes. 
Tahoma port: https://github.com/tahoma2d/tahoma2d/pull/574


RodneyBaker:
This is an important port as it surely will prevent users from losing work when using Plastic Tool.
Thanks @manongjohn for the change and @JParks123 for the port.
Much appreciated.

PR #6420, opened 2026-01-11
Fix preview undo collapse crash

JParks123:
This fixes a crash that occurs when you enter preview mode, collapse columns and then undo.

Tahoma2D port:  https://github.com/tahoma2d/tahoma2d/pull/1650


RodneyBaker:
I confirm this is still a problem in Opentoonz and this fixes the crash.
Thanks for the port @JParks123!

PR #6422, opened 2026-01-11
Fix build on gcc 15.1: prototypes must match

hfiguiere:


RodneyBaker:
An interesting update to some very old (even elegant) foundational code optimized for Toonz (vintage 1992 code even).
This code appears to still be included in Opentoonz mostly for backwards compatibility with the older palette file format (?)

Thanks for looking into this @hfiguiere

PR #6444, opened 2026-01-19
[test] Fix hexagonal shapes when drawing quick curves

konero:
Assumption is when drawing quickly, input events report consecutive points at same pos, caused calcTangent() to compute 0-len tangents resulting in sharp corners instead of smooth Hermite splines. This is just a test, needs user trials...

### DIFF
--- a/toonz/sources/tnztools/inputmanager.cpp
+++ b/toonz/sources/tnztools/inputmanager.cpp
@@ -364,6 +364,27 @@ TInputManager::addTrackPoint(
   double time,
   bool final )
 {
+  // Filter out points that are too close to the previous point.
+  // This prevents issues with some tablet drivers that send duplicate
+  // or near-duplicate position events, which can cause sharp corners
+  // in curve interpolation instead of smooth curves.
+  // The final point is always added to ensure proper stroke termination.
+  if (!final && track->size() > 0) {
+    const TTrackPoint &last = track->back();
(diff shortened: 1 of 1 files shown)


RodneyBaker:
You had me at 'Hermite Splines'.

There is a lot of good stuff going on in this code.
A general observation (mostly for my own understanding) regarding arc-length dependence:

If length accumulation ever breaks, everything breaks.  
Fortunately OpenToonz tracks are strict about this so we should not see that break.

Zero tangents at ends:
This slightly flattens stroke ends and although usually desirable can be noticeable with very short strokes.

Curvature clamping:  (There is none?)
Despite this, and apparently by design, high-frequency zig zagging lines can still produce strong curvature.

In summary:  What we see appears to be a Hermite spline where time is fake, distance is king, and bad tablet data is quietly ignored.

In other words, its an artist friendly approach where a stroke/spline is local and controllable. Each segment depends only on two points and two tangents.  If we change one point or tangent it doesn’t wreck the whole stroke. 

Aside:  I intentially use stroke (singularly) here as Opentoonz is doing something I can't quite pin down with strokes keeping them independent of other strokes.  This is both good and bad.  Good in the sense that strokes are isolated and therefore more manageable.  Bad in that multiple strokes cannot be processed at the control point level at the same time by the user.

This doesn't really address the fix of this PR which others who are experiencing the problem will need to verify.  I am currently not experiencing the issue others have reported.

PR #6445, opened 2026-01-19
Fix heap corruption in Iwa_TiledParticlesFx

andeon:
This PR addresses a crash that occurred on Linux, and possibly on macOS as well.

- Use const references for TSpectrum to avoid deep copy and off-by-one write
- Pass m_parent to Iwa_Particle constructor for safe access to Fx parameters
- Updated comments to English (translated from Japanese and Italian)

I’m not certain why the original deep copy was used or if it was necessary; any insights would be appreciated. The deep copy of TSpectrum was replaced with a const reference to avoid unnecessary copies per particle, prevent off-by-one writes and heap corruption, while keeping the original read-only behavior intact. This may not improve performance, the main goal is to fix the Linux issue.

### DIFF
--- a/toonz/sources/stdfx/iwa_particles.cpp
+++ b/toonz/sources/stdfx/iwa_particles.cpp
@@ -1,3 +1,5 @@
+
+
 //------------------------------------------------------------------
 // Iwa_Particle for Marnie
 // based on ParticlesFx by Digital Video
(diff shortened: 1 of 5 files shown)


RodneyBaker:
I converted @gab3d's report to a discussion as he was curious about what TiledParticlesFx was used for but...

Edit:  This PR has addressed the primary crash experienced therein.

REF: #6392

RodneyBaker:

PR #6508, opened 2026-01-21
Fix empty source switching crash

JParks123:
Crashes when trying to swap links between multiple input ports on some fx but there is actually nothing attached to any of the input ports. This is definitely user error, but some of the FX get a crash when trying to connect the sources of a single FX together.


RodneyBaker:
Thanks @JParks123 for the port.

For those curious, the ported PR:  https://github.com/tahoma2d/tahoma2d/pull/1241
The original bug report:  https://github.com/tahoma2d/tahoma2d/issues/1232

PR #6509, opened 2026-01-21
Fix crash when adding to task list from browser

JParks123:
A crash occurs when using Add as Render Task or Add as Cleanup Task from File Browser, but never once opened a Tasks panel. Adding tasks like this does not properly initialize the Task manager causing a crash when you try to add something to it. The only way to properly initialize it is to open a Tasks panel.

This updates the commands to check if the Task manager was initialized properly and if not, it will open a Tasks Panel first.

Ported from Tahoma2d: https://github.com/tahoma2d/tahoma2d/pull/970

### DIFF
--- a/toonz/sources/toonz/fileselection.cpp
+++ b/toonz/sources/toonz/fileselection.cpp
@@ -239,6 +239,10 @@ void FileSelection::enableCommands() {
 //------------------------------------------------------------------------
 
 void FileSelection::addToBatchRenderList() {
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I confirm the original crash (although I thought there were some open reports here in Opentoonz that detail the problem but couldn't find any)... and I confirm this fixes the problem.

Thank you @JParks123 for porting the fix!

PR #5117, opened 2023-09-30
Fix SVG sequence file loading

ChengduLittleA:
This PR fixes an issue when trying to load an SVG file sequence.

Modified it so SVG files are treated as Vector Images instead of Vector Levels, which allows the SVG files to be detected as a single image file or a series of files (filename.####.svg or filename_####.svg format).

Ported from https://github.com/tahoma2d/tahoma2d/pull/1212

Co-authored-by  @manongjohn .

PR #6521, opened 2026-01-23
Refactor: Modernize Qt/C++ code (QRegularExpression, atomic, handle leaks, validations)

andeon:
This PR continues the modernization of OpenToonz’s Qt/C++ codebase, focusing on replacing `QRegExp` with `QRegularExpression`. While doing so, I also applied additional fixes and refactors to improve correctness, stability, and modern Qt usage. See: https://github.com/opentoonz/opentoonz/issues/6311

- `tsystem.cpp` – Fix handle leaks, improve filesystem/path validation, modernize showDocument, enhance exception handling, and minor refactors for safety and clarity.

- `tfilepath.cpp` – Sanitized paths, removed legacy wide-string parsing. 

- `tiio_ffmpeg.cpp`–  Attempt to improve FP32 raster handling by clamping pixel values. Memory issues for heavy FX remain and will require future changes to caching, streaming, proxy handling, or other solutions to further improve performance.

- `schematicnode.cpp, camerasettingswidget.cpp` – Modernized signal-slot connections, improved null safety, used nullptr and range-based loops.

- `hexcolornames.cpp` – Migrated color tables to QHash, improved XML loading and error handling.

Now supports typing and autocompletion of hex codes. Numbers 0–7 correspond to `default main color` list entries for quick access (for clarification). See demo below ↓↓↓

https://github.com/user-attachments/assets/887dc001-9f08-475c-bf4f-002e6338e62c

Note: I decided to split this work to keep progress moving forward,  Remaining files needing `QRegExp` replacement  (28 files) will be submitted in separate PR or PRs to keep the work manageable.

### DIFF
--- a/toonz/sources/include/toonzqt/camerasettingswidget.h
+++ b/toonz/sources/include/toonzqt/camerasettingswidget.h
@@ -46,8 +46,8 @@ class CheckBox;
 
 class SimpleExpValidator final : public QValidator {
 public:
-  SimpleExpValidator(QObject *parent) : QValidator(parent){};
-  State validate(QString &input, int &pos) const override;
+  explicit SimpleExpValidator(QObject *parent = nullptr) : QValidator(parent) {}
+  QValidator::State validate(QString &input, int &pos) const override;
 };
 
 //---------------------------------------------------------------
@@ -57,10 +57,10 @@ class SimpleExpField final : public QLineEdit {
   QString m_previousValue;
 
 public:
-  SimpleExpField(QWidget *parent);
+  explicit SimpleExpField(QWidget *parent = nullptr);
 
-  void setValue(double);
-  void setValue(double, int, int);
+  void setValue(double value);
+  void setValue(double value, int w, int h);
   double getValue();
 
 protected:
@@ -104,6 +104,8 @@ class DVAPI CameraSettingsWidget final : public QFrame {
 
   void savePresetList();
   void loadPresetList();
+
+  // Use the version with double parameters instead
   bool parsePresetString(const QString &str, QString &name, int &xres,
                          int &yres, QString &ar);
 
@@ -112,14 +114,21 @@ class DVAPI CameraSettingsWidget final : public QFrame {
                          QString &yoffset, double &ar, bool forCleanup = false);
 
 public:
-  CameraSettingsWidget(bool forCleanup = false);
+  explicit CameraSettingsWidget(bool forCleanup = false);
   ~CameraSettingsWidget();
+
+  CameraSettingsWidget(const CameraSettingsWidget &) = delete;
+  CameraSettingsWidget &operator=(const CameraSettingsWidget &) = delete;
+
+  // Allow move semantics if needed
+  CameraSettingsWidget(CameraSettingsWidget &&) = delete;
+  CameraSettingsWidget &operator=(CameraSettingsWidget &&) = delete;
 
   void setPresetListFile(const TFilePath &fp);
 
   // Defines the level referred by the button "Use level settings".
-  // Calling setCurrentLevel(0) disables the button
-  void setCurrentLevel(TXshLevel *);
+  // Calling setCurrentLevel(nullptr) disables the button
+  void setCurrentLevel(TXshLevel *xshLevel);
 
   // camera => widget fields (i.e. initialize widget)
   void setFields(const TCamera *camera);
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Lots of good stuff in this one.
All very impressive but this catches my attention perhaps more than the others:

> Now supports typing and autocompletion of hex codes. Numbers 0–7 correspond to default main color list entries for quick access 

Very useful.

> tiio_ffmpeg.cpp– Attempt to improve FP32 raster handling by clamping pixel values. Memory issues for heavy FX remain and will require future changes to caching, streaming, proxy handling, or other solutions to further improve performance.

Yes, ffmpeg is quite a beast.  Thanks for whittling away at this.

Will spend some time testing but initial pass looks good.

> Note: I decided to split this work to keep progress moving forward, Remaining files needing QRegExp replacement (28 files) will be submitted in separate PR or PRs to keep the work manageable.

Good call.

RodneyBaker:
Proceeding with this refactoring although I do apologize to those that would prefer to have more time to investigate the code.  It is important we get the code into the hands of users so they can test it thoroughly via the nightly releases as we approach v1.8 release.

TLDR:  Everything behaves appropriately here.

Thank you @andeon for the continuing efforts to maintain and modernize Opentoonz code.

JParks123:
@andeon this PR breaks the SVG sequence importing function. 
https://github.com/opentoonz/opentoonz/pull/5117

image

andeon:
@JParks123 Could you explain how `import` is used and outline the steps required to reproduce this issue? The failure may be related to QRegularExpression parsing, possibly in `tfilepath.cpp`, and it might require some adjustment.

JParks123:
> @JParks123 Could you explain how `import` is used and outline the steps required to reproduce this issue? The failure may be related to QRegularExpression parsing, possibly in `tfilepath.cpp`, and it might require some adjustment.

To create an SVG sequence file you can even use an OpenToonz vector sequence. Export it out with the OCA export option in the File menu, don't forget to uncheck the Rasterize vector, so it will export as an SVG folder with the sequence files contained.
Import an SVG sequence by finding where you stored it in the Browser room in drag dropping it into the Scene cast, it should contain all the images as a single file SVG icon with the sequence amount number shown. This PR breaks the Sequence making it only a single SVG icon image Vector file, try the nightly prior to this PR and it works well.

andeon:
@JParks123 Please, Could you please verify if it works correctly after the build in this PR https://github.com/opentoonz/opentoonz/pull/6568?

JParks123:
> @JParks123 Please, Could you please verify if it works correctly after the build in this PR #6568?

@andeon Ok, I will try.

JParks123:
> @JParks123 Please, Could you please verify if it works correctly after the build in this PR #6568?


@andeon The Build you linked does not fix the problem, unfortunately.

this is how it should look in the image below.
image

RodneyBaker:
@Andeon  We have another mystery to solve.
It appears that as of this nightly release rendering to 18bit channel width MP4s and GIFs flips at least one frame upside down.

Odd for sure.
As near as I can tell the problem starts to occur on or about this PR.

toonz/sources/common/tsystem/tfilepath.cpp
```diff
@@ -827,12 +820,16 @@ bool TFilePath::isUneditable() const {
 
 //-----------------------------------------------------------------------------
 TFilePath TFilePath::withType(const std::string &type) const {
-  assert(type.length() < 2 || type.substr(0, 2) != "..");
-  int i            = getLastSlash(m_path);  // cerco l'ultimo slash
-  std::wstring str = m_path.substr(i + 1);  // str e' il path senza parentdir
+  if (type.length() >= 2 && type.substr(0, 2) == "..") {
```
RodneyBaker: Possible area for source of MP4, GIF vertical flip bug?

andeon:
> @andeon We have another mystery to solve. It appears that as of this nightly release rendering to 18bit channel width MP4s and GIFs flips at least one frame upside down.
> 
> Odd for sure. As near as I can tell the problem starts to occur on or about this PR.

I’m going to take a look. Do you have a workflow to make my test more precise, like specifying options, the number of frames, etc.?

RodneyBaker:
The steps I've been using to reproduce the flipped frame are:

- Create multiple frames in xsheet (content shouldn't matter but my test scene does have keyframes)
- Render with Output Settings set to MP4 and channel width set to 8bit (16 and 32bit shouldn't have the problem and is the current workaround)
- Note that a frame (or frames) will be flipped upside down
In my current test of some 20 nighltlies I used the same scene and the flipped frame was always the second frame.

I'll look to see if there is something about my test scene that causes a specific frame to be flipped.

It is reported that other formats such as GIF have the same problem but my main test was restricted to MP4 to reduce variation and make sure each test was exactly the same.  Each nightly was installed over the top of the last and I moved forward and back through the nightlies many times as I was having problems narrowing down which nightly had the problem versus not.

This is the nightly where the problem appears to start appearing:  https://github.com/opentoonz/opentoonz_nightlies/releases/tag/nightly_c0d22d56da79b3f2c20db1a16d1a5f3d3474f8ce

If I just draw an x in the corner of a frame and extend that to be multiple frames, then render as 8bit channel width to MP4 every other frame appears to flip vertically (not horizontally).  No keyframes.

PR #6523, opened 2026-01-23
Fix SVG stroke-width at lower resolutions

JParks123:
This fixes SVG stroke's importing too wide at lower resolutions.

Tahoma2d Port: https://github.com/tahoma2d/tahoma2d/pull/2004


RodneyBaker:
Thank you for the port @JParks123 
Much appreciated.

And thanks @manongjohn for the work you've been doing to update handling of the SVG format.

PR #6524, opened 2026-01-23
Fix to stop writing unfilled regions to SVG

JParks123:
When vector drawings are converted to SVG, not only are strokes written to the SVG file, but all fillable regions internally tracked are included. Fillable regions are written to the SVG as stroke-less paths which may or may not have a fill color.

This fix stops writing unfilled regions to the SVG file. These are not necessary and serve no purpose. In other vector drawing programs, they are useless invisible shapes. Only regions with a fill color will be written to the file.

Tahoma port: https://github.com/tahoma2d/tahoma2d/pull/2007


RodneyBaker:
Thank you @JParks123 for the port.
Much appreciated.

PR #6526, opened 2026-01-24
SVG load fixes

JParks123:
Fixed lines loading with border incorrectly when stroke data is not present (fill shapes).
Fixed double application of scale data causing some line thickness to be incorrect
Fixed application of transform:rotation data correcting position of shapes
Fixed lines from multi-line shapes drawn in the wrong position
Fixed processing of path:transform data found after path data instead of before it.
Added support to load stroke-linecap, stroke-linejoin and stroke-miterlimit data
Remove unnecessary variable.
Autoclose open 0 thickness lines (fill shapes) instead of adding and grouping additional closing line
Replace transparent lines, used for closing shapes, with 0 thickness lines so they are visible in Viewer when Show Lines with Thickness 0 is enabled

Tahoma port: https://github.com/tahoma2d/tahoma2d/pull/1995

### DIFF
--- a/toonz/sources/image/svg/tiio_svg.cpp
+++ b/toonz/sources/image/svg/tiio_svg.cpp
@@ -27,6 +27,7 @@ namespace  // svg_parser
 {
 
 struct NSVGpath {
+  float *svgPts;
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thank you @JParks123 for the port.
One more those interested in SVG improvements will be very happy to see.

PR #6567, opened 2026-01-25
Fix Extend Fill

Abel032:
This PR optimized the extend fill algorithm to avoid some unexpected fills.

- Only fill closed areas when filling from a line with auto-Paint style to another line with normal style.
图片

- Stricter condition for filling normal gaps.
图片

Simple Test File:
[Desktop.zip](https://github.com/user-attachments/files/24848467/Desktop.zip)


RodneyBaker:
This works great @Abel032 
Thank you for the fix.

PR #6568, opened 2026-01-26
Fix: Image sequence display

andeon:
This PR fixes the issue mentioned by @JParks123 in #6521,  where some SVG not show sequential frame files correctly in browser.

### DIFF
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -501,7 +501,6 @@ public:
           case_compare = (a_char < b_char) ? 1 : 0;
         }
       }
-      i++;
     }
     return (case_compare == 1);
   }


RodneyBaker:
In my test this PR does NOT fix the problem being seen.
To test:

Create a SVG sequence (it can be copies of the same SVG file just named appropriately as a sequence; svgfile.0001.svg, svgfile.0002.svg, etc.)

Open OT's Browser (or other location) and note that the frame number will show 1 frame.
If loaded only  the first SVG file in the sequence gets loaded.

The desired result should allow multiple frames to be seen and loaded into OT.

I'll test with the appveyor build once that passes checks to make sure what I'm seeing isn't the result of just the github build.
Update:  I confirm the Appveyor build also does not resolve the problem.

RodneyBaker:
@andeon Here's the original change that allowed eligible SVGs to be recognized as a sequence:

https://github.com/opentoonz/opentoonz/pull/5117/changes

From what I can tell the problem should live in `toonz/sources/image/tiio.cpp`
Or more specifically, one of the changes in other files that overrides settings in that file,

andeon:
I'm checking this, and `tsystem.cpp` affects the display part, since the images are generated correctly but are not being recognized properly. I'll revert the changes and leave this for later correction to simplify this logic. Now that we're aware of the rules and the other files involved, it's better to test everything more carefully. It's probably something simple and obvious, but regex can be tricky to fix sometimes.

andeon:
Since this manual PR didn’t fully solve the problem, I’m not sure if I should close it and instead revert the original commits directly. I did the commits separately to make handling this type of situation easier. How is this usually handled here, @RodneyBaker ?

RodneyBaker:
There isn't a particular rush to fix the problem as in this case I don't think it's a complex problem but rather just tricky to locate the specific change.

The workaround for the moment... for those requiring loading of SVG sequences... is to revert back a few nightlies or rename and load the SVGs and single images.

If the problem cannot be fixed... particularly by the time of v1.8 release.... then reverting is certainly the preferred option.

It is, to me, reasonable that major updates such as this refactoring will take time to fully discover unintended consequences and in this case we've discovered an outlier.  The timing is good however as we still have time to determine what (if any) other problems may be introduced due to the change.

Importantly, it's vital that we modernize the source code to the greatest extent possible.
As such I'd say we are moving forward appropriately.

Let's look at the original to see if a single revert (or two) might be the best way to proceeed but my sense is that we just need to indentify why SVGs are not being treated the same as PNGs where it comes to ingesting sequential files.

andeon:
Ok, since I have some time, I’ll try to isolate the problem, but tsystem.cpp seems to be one of the main files involved, so I’ll play around with it.

andeon:
@JParks123 and @RodneyBaker, could you test regenerate again and check if it fixes the issue? As I suspected, it was something very simple and trivial, but sometimes not so easy to spot.

@RodneyBaker, when you do the final merge, could you select ‘squash and merge’ to clean up the history? Since there are review comments in the code, I’m not sure if using amend or rebase/squash on my side would be safe.

JParks123:
> @JParks123 and @RodneyBaker, could you test regenerate again and check if it fixes the issue? As I suspected, it was something very simple and trivial, but sometimes not so easy to spot.
> 
> @RodneyBaker, when you do the final merge, could you select ‘squash and merge’ to clean up the history? Since there are review comments in the code, I’m not sure if using amend or rebase/squash on my side would be safe.

@andeon this works good, Thanks for the fix.  
image

image

RodneyBaker:
Thanks @andeon I confirm this fixes the problem.

> ... it was something very simple and trivial, but sometimes not so easy to spot.

Exactly what I thought although looking at your fix I definitely wasn't expecting that.
Nice work.

> ...when you do the final merge, could you select ‘squash and merge’ to clean up the history?

Will do and thanks again.

PR #6310, opened 2025-12-12
refactor: C++ improvements, RAII, Qt modernization

andeon:
This pull request addresses critical issues reported by Clang Static Analyzer and consolidates changes across multiple files to improve code safety and maintainability.

- Replaced manual memory management with smart pointers
- Added RAII guards for OpenGL state, improved exception safety, and eliminated memory leaks
- Updated old-style `connect()` calls to the new Qt syntax, ensuring forward compatibility with Qt6
- Replaced raw `0/NULL` with `nullptr`
- Improved context menus: conditional creation
- General code cleanup and consistent modern C++ patterns
- Added comprehensive English comments

ref: https://github.com/opentoonz/opentoonz/issues/5834, https://github.com/opentoonz/opentoonz/pull/5966

**PS:** It may be beneficial to create a dedicated issue for the Qt modernization migration, so future PRs that remove Qt deprecations (`setMargin`, `QRegExp`, old-style signals/slots, etc.) can reference it like the C++17 migration issue. This will help track progress and context for forward compatibility with Qt6. For now, the main PR initiating these changes is linked for reference: [#5966](https://github.com/opentoonz/opentoonz/pull/5966). I plan to make these changes incrementally to allow proper testing of behavior and prevent regressions, rather than submitting everything at once.

PR #6590, opened 2026-01-29
Fix compiler warnings and potential memory leaks in TIPC module

andeon:
This pull request addresses critical issues reported by Clang Static Analyzer and consolidates changes across multiple files.  See: https://github.com/opentoonz/opentoonz/issues/6311

- Fixed potential memory leak in SocketController lifecycle management
- Updated deprecated Qt signal/slot syntax to modern Qt5 style
- Fixed abstract class deletion warning by adding virtual destructor to MessageParser

The memory leak was in the SocketController objects not being properly cleaned up when sockets disconnected. Using Qt's parent-child relationship ensures automatic cleanup when the parent socket is deleted.


RodneyBaker:
Pressing forward with thanks @andeon 
Keep up the great work.

PR #6324, opened 2025-12-19
Fix reading corrupted PNG.

JParks123:
Fix reading corrupted PNG.

PR #6595, opened 2026-02-01
Fix potential memory leaks and stream handling

andeon:
This pull request addresses issues reported by Clang Static Analyzer and consolidates changes across multiple files focus in leaks corrections. 

- `tiio_png.cpp` - Fix Linux crash when opening corrupted PNG images in local builds. Damaged rows are now filled with a checkerboard pattern (apparently the industry default), allowing users to visually identify corrupted areas instead of silent failures (for example, when receiving png files without knowing they are damaged). Support for interlaced (Adam7) images remains incomplete and fails; however, this is not critical for applications like OpenToonz. 
checkoard

**References / Origin:** [bad.zip](https://github.com/user-attachments/files/24175890/bad.zip) containing corrupted PNG to reproduce the issue. 
The original fix ported from tahoma2d/tahoma2d#1982 via PR #6324

**Other improvements across the codebase:**

- Added null checks and proper cleanup to prevent memory leaks.

- Replaced manual memory management with safer patterns.

- Modernized smart pointer implementation: added move semantics, nullptr, and implicit bool conversion.

- Remove `using namespace std;`  See: [#6311](https://github.com/opentoonz/opentoonz/issues/6311)

### DIFF
--- a/toonz/sources/include/tsmartpointer.h
+++ b/toonz/sources/include/tsmartpointer.h
@@ -35,14 +35,10 @@ class DVAPI TSmartObject {
 public:
   typedef short ClassCode;
 
-  TSmartObject(ClassCode
-#ifdef INSTANCE_COUNT_ENABLED
-                   classCode
-#endif
-               )
+  TSmartObject()
       : m_refCount()
 #ifdef INSTANCE_COUNT_ENABLED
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thank you @andeon for yet another fix
LGTM

PR #6597, opened 2026-02-01
feat(viewer): add Clear Viewer Content command

Ahl76:
This command allows users to clear the content of a frame without removing the cell from the Xsheet or Timeline. The result is similar to the Delete command used in the Level Strip. It simplifies the process by avoiding the need to select content and click delete within the viewer panels.

Clear viewer content

Key points :

- Clears the artwork within the frame faster.
- Maintains the animation timing.
- Supports keyboard shortcut assignment.
- Works in the Viewer and the Level Strip.

### DIFF
--- a/toonz/sources/toonz/CMakeLists.txt
+++ b/toonz/sources/toonz/CMakeLists.txt
@@ -323,6 +323,7 @@ set(SOURCES
     filmstripcommand.cpp
     iocommand.cpp
     levelcommand.cpp
+    viewercommand.cpp
     print.cpp
     subscenecommand.cpp
     svncleanupdialog.cpp
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -1901,6 +1901,7 @@ void MainWindow::defineActions() {
   createMenuEditAction(MI_PasteInto, QT_TR_NOOP("&Paste Into"), "",
                        "paste_into");
   createMenuEditAction(MI_Clear, QT_TR_NOOP("&Delete"), "Del", "delete");
+  createMenuEditAction(MI_ClearViewerContent, QT_TR_NOOP("&Clear Viewer Content"), "", "clear_viewer");
   createMenuEditAction(MI_Insert, QT_TR_NOOP("&Insert"), "Ins", "insert");
   createMenuEditAction(MI_InsertAbove, QT_TR_NOOP("&Insert Above/After"),
                        "Shift+Ins", "insert_above_after");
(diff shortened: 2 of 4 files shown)


RodneyBaker:
Works exactly as advertised.
Posting intent to merge although there is no rush.

I note that while your screenshot uses Backspace (which is more than appropriate for clearing the current selection/canvas as that is common in many graphics programs) you didn't actually assign any shortcut key to the command by default.  This is safe and will certainly keep users from accidentally encountering the command.

I do wonder... a little... how users will discover the feature short of seeing it listed in v1.8 release notes, word of mouth, exploring the shortcuts listing, etc. 

All this to say:  this will be a very useful addition.

RodneyBaker:
Thanks @Ahl76 for the option to clear the viewer's content from xsheet/timeline.
And congratulations on your first Pull Request improving Opentoonz!
Much appreciated.

Ahl76:
Thanks @RodneyBaker. I'm glad this is merged :)

RodneyBaker:
@Ahl76 A follow up note...

It has been reported that having the Clear Viewer Content command listed under Menu > Edit is confusing given that the command is not actually listed in the menu at all.  It does seem appropriate to be categorized under Menu > Edit however.  If not designed for use in the menu maybe it could fall under a different category.

Perhaps this could be a follow up addition to the feature as you find time.  

This shouldn't slow anyone down but there's a bit of feedback for you.
Perhaps more importantly, I'm hearing that users are enjoying having the ability to clear the viewer's content so easily.

Ahl76:
@RodneyBaker ​Thanks for the feedback. Indeed, the command is currently only under the "Edit" category in the "Configure Shortcuts" window. Do you have any suggestions for a better location? Otherwise, I could add it to the main application's "Edit" menu later to reduce confusion.

RodneyBaker:
I personally don't have a strong preference but it seemed very important to the one reporting it to me that the command located in the Edit category was not to be found in the Edit menu.  

Note that this is a dangerous command in that we don't want it placed where the user will accidentally select it.  Should it therefore be placed in the Misc category?

If in the Edit menu, would placing it after 'Delete Cells' be appropriate?

I'd say placing it in the Edit menu for now will be the best approach and less of a chore for you to make the adjustment.  If problematic for you I'd say assign it to Misc as a shortcut and the user can be told to find it there.

I anticipate that given we are expecting to see a v2.0 rather than v1.8 release, the following v2.1 release will very likely see some major UI changes to take advantage of Qt6 and Apple Silicone.  Refinement of where things are optimally can be fought over during that development cycle.  ;)

All this to suggest it's your call where to place the command but we should place the command in the Edit menu if it's in the Edit category for shortcuts.

Ahl76:
Imo, ​given its similarity to the Delete function in the Level Strip, I agree it makes sense to place it in the "Edit" menu directly after the "Delete" command, and we can always revisit it later if needed.

PR #6600, opened 2026-02-02
Refactor: Modernize Qt/C++ code (QRegularExpression, signals/slots)

andeon:
This PR continues the modernization of OpenToonz’s Qt/C++ codebase, with a primary focus on replacing all QRegExp with QRegularExpression. While doing so, I also applied additional fixes and refactors. See: https://github.com/opentoonz/opentoonz/issues/6311

##### Key Changes:

   - Replaced all QRegExp instances with QRegularExpression
   - Modernized signal/slot connections to use Qt5 syntax
   - Replaced `0`/`NULL` with `nullptr` for null pointer safety
   - Added `explicit` to single-argument constructors
   - Used `= delete` to explicitly disable copy operations
   - Added `const` qualifiers where appropriate
   - Improved type safety with `static_cast` and `dynamic_cast`
   - Used range-based for loops instead of traditional C++98 loops for readability and safety
   - Applied `auto` keyword where appropriate to reduce verbosity
   - Fixed inconsistent parameter names
   - Standardized code formatting and indentation

### DIFF
--- a/toonz/sources/toonz/shortcutpopup.cpp
+++ b/toonz/sources/toonz/shortcutpopup.cpp
@@ -51,8 +51,8 @@ ShortcutPopup *ShortcutPopup::s_instance = nullptr;
 //=============================================================================
 // ShortcutItem
 // ------------
-// Lo ShortcutTree visualizza ShortcutItem (organizzati in folder)
-// ogni ShortcutItem rappresenta una QAction (e relativo Shortcut)
+// The ShortcutTree displays ShortcutItem (organized in folders)
+// each ShortcutItem represents a QAction (and its Shortcut)
 //-----------------------------------------------------------------------------
 
 class ShortcutItem final : public QTreeWidgetItem {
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@andeon I think @Ahl76's recent PR adjusted shortcuts which will require a refresh of this PR.
Other than that I'd say it is good to go.

RodneyBaker:
Thanks for refreshing the PR @andeon 
While I am ready to move forward with this PR let's give it a little more time to simmer as my current rate of merging is raising people's blood pressure.

A question about the individual commits versus squashing...
I follow your desire to keep the commits separate and organized and like that for running through reviews but for final merger squashing will save a lot of time for those that come afterward.  

It is important to understand how that final squashing effect the contributor (you in this case) personally although we recognize that squashing during merger does not effect the breakdown of separate commits you have on your branch.

In the future we'll likely codify this into policy so developers know what to expect.

andeon:
I think the default should be squash-and-merge. The only case where extra care is needed is when doing a port with multiple contributors, since squashing can overwrite the original description and obscure or remove proper credit for all co-authors. This also helps keep the project history clean.

RodneyBaker:
Moving forward with this @andeon 
Thank ou for the refactoring modernization.

PR #6601, opened 2026-02-03
Fix memory leaks and improve resource management across audio loaders and OpenGL

andeon:
This pull request addresses issues reported by Clang Static Analyzer and consolidates multiple fixes focused on memory leak prevention, uninitialized memory issues.

`tsio_mp3.cpp, tsio_mp3.h` - Fixed memory leak in TSoundTrackReaderMp3::load() by replacing raw pointers with std::unique_ptr, enhanced cache file handling in getRawAudio() by validating cache existence.

`tsio_wav.cpp, tsio_wav.h` - Fixed a potential memory leak by replacing raw pointers with std::unique_ptr for TFMTChunk and TDATAChunk.

`iwa_soapbubblefx.cpp` - Fixed a memory leak in do_distance_transform() by replacing a raw pointer with std::unique_ptr<float[]>, Fixed uninitialized memory in dt() by initializing the z array and uninitialized memory in processNoise() by initializing the noise_phases array.

`shadingcontext.cpp, shadingcontext.h` - Fixed a memory leak in TQOpenGLWidget::initializeGL() by removing an unused QOffscreenSurface, Fixed makeCurrent() to avoid creating a new OpenGL context on every call; the existing context is now reused when valid. **_Todo: If TQOpenGLWidget is used only internally, the DVAPI macro can be safely removed._**

`filesgi.cpp` - Removed `using namespace std;` to improve code safety and consistency. See: https://github.com/opentoonz/opentoonz/issues/6311

### DIFF
--- a/toonz/sources/image/sgi/filesgi.cpp
+++ b/toonz/sources/image/sgi/filesgi.cpp
@@ -42,8 +42,6 @@
 #endif
 #include <stdarg.h>
 
-using namespace std;
-
 // IMAGERGB: SGI image header structure
 struct IMAGERGB {
   unsigned short imagic;  // Magic number
@@ -230,7 +228,7 @@ static IMAGERGB *iopen(int fd, OpenMode openMode, unsigned int type,
     image->wastebytes = 0;
     image->dorev      = dorev;
     if (do_rgb_write_header(image, f) != IMAGERGB_HEADER_SIZE) {
-      cout << "iopen: error on write of image header\n" << endl;
+      std::cout << "iopen: error on write of image header\n" << std::endl;
       free(image);
       return nullptr;
     }
@@ -238,7 +236,7 @@ static IMAGERGB *iopen(int fd, OpenMode openMode, unsigned int type,
   } else {
     // READ mode
     if (do_rgb_read_header(image, f) != IMAGERGB_HEADER_SIZE) {
-      cout << "iopen: error on read of image header" << endl;
+      std::cout << "iopen: error on read of image header" << std::endl;
       free(image);
       return nullptr;
     }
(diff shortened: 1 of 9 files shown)


konero:
There's random comments in the code: `// FIXED: Removed unused variable 'get'`, or `// Optional: add explicit constructor and destructor` that could be removed. 😅

andeon:
Sometimes I write comments in Portuguese and forget to remove them when using automated scripts. While working with SourceGit, I sent them for auto-translation into English along with the Italian and Japanese translations.

RodneyBaker:
I haven't investigated this deeply yet but while testing ran into this error while Exporting a Scene via File > Export Scene (all options):

image

I'm not seeing anything in this PR's code that looks suspectbut perhaps an update here creates a conflict with code in the Scene exporter code.

This doesn't crash OT, just prevents export, which suggests we probably need to look at the Export code.

andeon:
> I'm not seeing anything in this PR's code that looks suspectbut perhaps an update here creates a conflict with code in the Scene exporter code.

Sorry, I don’t fully understand the workflow. When you say Scene via File → “Export Scene”, do you mean File → “Export Current Scene” and the available export options there, such as [ ] export as Standalone Scene? I’m testing with OpenToonz_sample and I’m not able to reproduce the error, so it may be related to something specific in the scene, Maybe?

RodneyBaker:
Yes, File > Export Current Scene.

At this exact moment it does appear to be something related to the scene itself as I can't even open the scene.

Disregard this for now and I'll add more information if/as I discover anything new.
The problem DOES NOT appear to be isolated to this PR's build (although perhaps the scene was corrupted while using it).

Uploading the scene file (no levels) for anyone that wants to look into that.
[ReferFill.zip](https://github.com/user-attachments/files/25085483/ReferFill.zip)

RodneyBaker:
Shortly after posting my last I downloaded the github build again and since then have not run into the problem.  Perhaps a bad build.

Will press forward with this as I see no reason to delay.
I'll be on the lookout for the problem I outlined in the nightly.

Thanks @andeon!

PR #6602, opened 2026-02-03
Allow Custom Splash Screen

JParks123:
This PR enables users to change the splash screen or use the OpenToonz default.

To display the custom splash place an SVG file named splash.svg in the Opentoonz program directory.  This is the same directory as Opentoonz.exe.

### DIFF
--- a/toonz/sources/toonz/main.cpp
+++ b/toonz/sources/toonz/main.cpp
@@ -442,9 +442,21 @@ int main(int argc, char *argv[]) {
 
   TEnv::setApplicationFileName(argv[0]);
 
-  // splash screen
-  QPixmap splashPixmap = QIcon(":Resources/splash.svg").pixmap(QSize(610, 344));
+// splash screen (override with local file if present)
+QString exeDir = QCoreApplication::applicationDirPath();
+QString localSplashPath = QDir(exeDir).filePath("splash.svg");
 
+QPixmap splashPixmap;
+
+if (QFileInfo(localSplashPath).exists() && QFileInfo(localSplashPath).isFile()) {
+  splashPixmap = QIcon(localSplashPath).pixmap(QSize(610, 344));
+  if (splashPixmap.isNull()) {
+    // fallback if loading fails
+    splashPixmap = QIcon(":Resources/splash.svg").pixmap(QSize(610, 344));
+  }
+} else {
+  splashPixmap = QIcon(":Resources/splash.svg").pixmap(QSize(610, 344));
+}
 #ifdef _WIN32
   QFont font("Segoe UI", -1);
 #else


DarrenTAnims:
It's a shame you can't just use a png image.
But I'm sure some will find this useful still :-)

RodneyBaker:
splash

Use this (or similar) as the splash.svg locally and it will point to a PNG named splash.png

The entire text of that svg file:  

```
<?xml version="1.0" encoding="UTF-8"?>
<svg xmlns="http://www.w3.org/2000/svg"
     xmlns:xlink="http://www.w3.org/1999/xlink"
     width="610" height="344" viewBox="0 0 610 344">

  <defs>
    <linearGradient id="greenGrad" x1="0" y1="0" x2="0" y2="1">
      <stop offset="0%" stop-color="#bbffbb"/>
      <stop offset="100%" stop-color="#0a5a0a"/>
    </linearGradient>

    <clipPath id="roundClip">
      <rect x="0" y="0" width="610" height="344" rx="24" ry="24"/>
    </clipPath>
  </defs>

  <!-- Background gradient -->
  <rect x="0" y="0" width="610" height="344" rx="24" ry="24" fill="url(#greenGrad)"/>

  <!-- Overlay image (clipped to rounded corners) -->
  <image x="0" y="0"
         width="610" height="344"
         xlink:href="splash.png"
         clip-path="url(#roundClip)"/>
</svg>


```
I include in this SVG examples of creating a gradient fill and rounding corners to demonstrate how the SVG file allows adjustment of the splash screen.

If we don't like the name splash.png for the splash image?  Simply change the pointer's name to something more appropriate.

RodneyBaker:
I post the above to suggest that a very good reason (one of them at least) to use SVG is that it allows the user-developer to adjust what is displayed in the splash.

There are restrictions and constraints of course.
Opentoonz is not set up to handle complex SVGs.
If we reference a gif animation for instance, OT will currently display only the first frame of that animation.

RodneyBaker:
Pressing forward with this so users can explore for themselves.
Thank you @JParks123 for adding the option to customize the splash screen.

konero:
Please allow time for feedback, not even a day passed. 😂 

You don't need the QFile objects (there's two of them), just load the QPixmap (it already handles existence) and returns null object if not found. Otherwise this is just more I/O when booting the app.

Also maybe a Mac pathing issue, but I don't have much experience, potentially not outside the bundle with that path...?

RodneyBaker:
@konero I agree QPixmap is a better approach.
I assisted @JParks123 on this PR and it's completely my fault we didn't pursue the QPixmap approach.  I wasn't thinking broadly enough.  I've played with some alternative code and there are a number of things we could do to speed up the process and I fully expect we will in the future.  This PR simply introduces the concept. 

In discord you mentioned this custom splash should be a Preference.
There are a number of reasons why this particular implementation does not pursue that approach.
Firstly, those forking OT may not want their splash screen to be so easily replaced.
Users that know to place the splash.svg in the required location however can bypass the default.
Is that the correct way forward?  Perhaps not.
It does however attempt to address some concerns raised about altering the splash.

Regarding @DarrenTAnims's desire for direct PNG access that also would speed up the process.
Having said that there are some specific advantages to using SVG however so I'd say the code should first look for locall PNG and then SVG.  This check for both however just adds more processing.  Perhaps naively, I also see such uses of SVG as an onramp for user-developers of Opentoonz as HTML and SVG are excellent ways to lean into programming.

Lots more to say here but I'll just mention the MacOS side of this equation.
We definitely need more MacOS users and testers in our community.
Until that time, the vast majority of code will not get scrubbed for MacOS until after it hits the nightlies.  

Thanks for your feedback.

PR #6607, opened 2026-02-03
CI - Windows: Bump OpenCV to 4.13.0

xavier2k6:


RodneyBaker:
As far as I can tell this update doesn't effect use of plugins created with prior versions of OpenCV.
If anyone runs into problems please let us know.

RodneyBaker:
Pressing forward with this OpenCV update.
Thank you @xavier2k6

PR #6609, opened 2026-02-04
Fix memory leaks and add null pointer checks

andeon:
This pull request addresses issues reported by Clang Static Analyzer and consolidates multiple fixes focused on memory leak prevention and uninitialized memory issues.

- Replaced raw pointers with smart pointers to prevent memory leaks and use-after-free issues.  
- Added null pointer checks, initialized all pointers and variables, and ensured proper resource cleanup in all code paths.  
- Modernized C++ code: used `auto`, `size_t` loops, explicit `std::` qualifiers, and constructor initializer lists; removed `using namespace std;`, modernized signal/slot connections to use Qt5 syntax See: [6311](https://github.com/opentoonz/opentoonz/issues/6311).  
- Removed unused variables, dead stores, and translated all comments to English.

For better portability, existing `UINT` types should be refactored in the future. `size_t` is recommended for vector or container indices, as it matches the platform's addressable size and avoids truncation on 64-bit systems, while `unsigned int` can be used for general-purpose counters where platform-dependent size is not critical. Some `UINT` instances remain and should be replaced according to context to ensure portable and safe code.

Note: The analyzer reports over 30 files with potential memory leaks that require further investigation. Some warnings may be false positives, and the results may not reflect all possible memory leaks, as the tool is limited in scope. After finishing this phase, I will use additional tools to perform a more comprehensive code analysis.  Reference: https://github.com/opentoonz/opentoonz/pull/6310


RodneyBaker:
I hope in the future others will look back at some of these code modernizations as they can be quite educational given the changes to standards over the years.  It's both fascinating from the perspective of the new code but also from the historical perspective of how some code had to deal with processes before modern usage was put into place.

Thanks again for updating the code and particularly for fixing memory leaks.

(Added)  I should have mentioned this part of your description:  

> After finishing this phase, I will use additional tools to perform a more comprehensive code analysis. Reference: https://github.com/opentoonz/opentoonz/pull/6310

This will be an important milestone to hit particularly as it will help validate all of the other changes.  There is always a risk of an update that looks reasonable and works not being optimal, particularly in areas that reference older code that cannot be updated yet due to current constraints.

RodneyBaker:
@andeon I think we've found another process that didn't make it through through the modernization update.

Perhaps somewhere around:   (Sorry, bad link.  Will update)

How to test:
- Using current nightlies after this PR was merged
- Create a vector line
- Open Control Point Editor
- Select stroke/spline area between two control points (zoom in if necessary)
- Note how the spline reacts, deforming 
- Compare this to previous releases where doing the same steps allows finely controlled adjustment of the spline.

I haven't determined where exactly the problem is.

andeon:
Maybe something in `controlpointselection.cpp or controlpointselection.h` I’m going to take a look.

PR #6614, opened 2026-02-09
Fix controlling Toonz Raster brush resize

JParks123:
This fixes an issue where tracking of mouse was incorrect when resizing the Raster which made it difficult to control the size of both the min and max sizes.


gab3d:
This is a port from Tahoma2D, isn't it?

JParks123:
> This is a port from Tahoma2D, isn't it?

yes, I will fix soon.

PR #6616, opened 2026-02-12
Fix memory leaks, dead stores and improve resource management

andeon:
This pull request addresses issues reported by Clang Static Analyzer and consolidates multiple fixes focused on memory leak prevention and uninitialized memory issues.

* Replaced raw pointers with smart pointers (`std::unique_ptr`) in `FileBrowser`, `FillHolesPopup`, and `FilmstripCommand` to prevent memory leaks and ensure exception safety during undo operations.

*  Added null pointer checks before accessing levels (e.g., in `dRenumberCells`), initialized pointers with `nullptr`, and used `TXshSimpleLevelP` (smart pointer) to prevent use-after-free issues.

* Modernized signal/slot connections to use Qt5 syntax where applicable, and introduced `std::unique_ptr` for UI elements like `QMenu` and `QActionGroup`, using `.release()` only when transferring ownership to Qt's parent-child system. See: [6311](https://github.com/opentoonz/opentoonz/issues/6311)
 
* Removed unused variables ("dead stores") in multiple selection files, deleted redundant layout allocations, and translated all remaining Italian and Japanese comments to English for better maintainability.

Note: The analyzer reports over 30 files with potential memory leaks that require further investigation. This PR addresses a significant portion of these in the "Selection" and "Command" modules.  Reference: #6310

### DIFF
--- a/toonz/sources/toonz/cellkeyframeselection.cpp
+++ b/toonz/sources/toonz/cellkeyframeselection.cpp
@@ -18,6 +18,9 @@
 #include <QApplication>
 #include <QClipboard>
 
+// C++ includes
+#include <memory>
+
 //=============================================================================
 //  TCellKeyframeSelection
 //-----------------------------------------------------------------------------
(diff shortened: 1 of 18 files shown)


RodneyBaker:
Proceeding as it is only the Mac Mojave check that is offline.
Thanks @andeon

PR #6618, opened 2026-02-13
Fix sub-xsheet icon generation crash

Abel032:
This behavior would cause OT to crash in current Nightly Build, this bug was introduced by this PR: #6293

20260214025040_rec_
Need to avoid bug and test related functions when introducing new features.
This bug also caused because I didn't review my code, should use the normal "getCurrentCamera()" to get scene's default camera and make it only work for  component `toonzpreview.dll`.


RodneyBaker:
I confirm this fixes the crash.
Thanks @Abel032 Much appreciated!

PR #6622, opened 2026-02-14
Fix memory leaks and dead stores

andeon:
This pull request addresses the final set of issues reported by the Clang Static Analyzer and modernizes the codebase across `toonzqt` and `toonzlib`, focusing on modernization and memory safety

- Replaced raw pointers with smart pointers and added null pointer guards to fix crashes in playback and drag-and-drop.

-  Migrated signals/slots to Qt5 syntax and used `std::unique_ptr` for UI components like `QActionGroup`. See: [6311](https://github.com/opentoonz/opentoonz/issues/6311)

-  Fixed division-by-zero in PlaybackExecutor, removed dead stores/unused variables in `flipconsole.cpp` and `dvdialog.cpp`, and translated legacy Italian/Japanese comments to English.

Note: This PR completes the planned fixes for memory-related warnings identified by the Clang Static Analyzer. However, this is only the tip of the iceberg, further analysis with dynamic tools and deeper architectural refactoring will be necessary to address more complex resource management issues that static analysis cannot detect. Reference: https://github.com/opentoonz/opentoonz/pull/6310

### DIFF
--- a/toonz/sources/include/toonz/scriptbinding_centerline_vectorizer.h
+++ b/toonz/sources/include/toonz/scriptbinding_centerline_vectorizer.h
@@ -1,9 +1,9 @@
 #pragma once
-
 #ifndef SCRIPTBINDING_CENTERLINE_VECTORIZER_H
 #define SCRIPTBINDING_CENTERLINE_VECTORIZER_H
 
 #include "toonz/scriptbinding.h"
+#include <memory>
 
 class ToonzScene;
 class TXshSimpleLevel;
@@ -12,11 +12,10 @@ namespace TScriptBinding {
 
 class DVAPI CenterlineVectorizer final : public Wrapper {
   Q_OBJECT
-  CenterlineConfiguration *m_parameters;
+  std::unique_ptr<CenterlineConfiguration> m_parameters;
 
 public:
   CenterlineVectorizer();
-  ~CenterlineVectorizer();
 
   Q_INVOKABLE QScriptValue toString();
   WRAPPER_STD_METHODS(CenterlineVectorizer)
@@ -58,7 +57,7 @@ public:
   void setEir(bool v);
 
 private:
-  QScriptValue vectorizeImage(const TImageP &src1, TPalette *palette);
+  QScriptValue vectorizeImage(const TImageP &src, TPalette *palette);
 };
 
 }  // namespace TScriptBinding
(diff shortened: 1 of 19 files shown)


RodneyBaker:
@andeon I am seeing a problem that is not related to this PR specifically but likely is related to one of the previous modernization updates

We can currently render out properly to MP4 but when loading these MP4s are coming into OT's xsheet blank.

I believe the process was working fine prior to 11 Feb but this is just to get the problem on your radar so when found, assuming its related to one of your PRs, you might include it here or in other PR (likely preferred).

Aside:  I've been using a variety of different releases and only now am experiencing the problem which is odd.

RodneyBaker:
As this is the final update in your current modernization plan I'm not opposed to merging this to get all of those updates seated so we have a solid foundation to test on and root out any problems.

andeon:
I suspect a bug in the directory concatenation in tiio_ffmpeg.cpp from my PR in late January. Gonna look into fixing it now.

RodneyBaker:
Thanks @andeon
I'm looking forward to locking the current slate of moderization updates down so we can focus fully on getting version 1.8 out.  We are right at the point where we have adequate time to test and secure an initial Release Candidate.  The more time we focus on and devote to that testing phase the better.  

At the point where we call it 'finished' on our side we can turn it over to @shun-iwasawa and he can dictate the final phase of release.  It's important that we get to that point where we can say, 'to the best of our ability as a community we are done'.  Then we can shift to addressing any issues Shun requires for us to address.  The remainder will then depend upon his availability and schedule.

We are getting very close now.

cc:  @abel  @Ahl76 @JParks123 @konero @JParks123 @manongjohn @shun-iwasawa @xavier2k6
(Thanks to all for getting us so close to our approach to the finish line)

andeon:
@RodneyBaker, I've stopped submitting new feature changes to focus solely on bug fixes. I might consider exceptions for algorithmic optimizations only if there is a significant performance gain. Technically, we are currently in Alpha, but we need to move toward a Beta or RC (Release Candidate) version for widespread user testing before the final merge. My goal is to catch any remaining crashes, even though many have already been resolved since late last year by collaborators in general.

RodneyBaker:
Pressing forward with this to close out your collection of modernization updates and fixes.
Thanks @andeon!

PR #6625, opened 2026-02-15
Fix blank frames on import

andeon:
This PR fixes an issue introduced by #6521, where frames were appearing blank in the xsheet when importing MP4s, as reported by @RodneyBaker.

This patch should resolve the issue by ensuring a more robust frame extraction and retrieval process, especially for files with large frame counts.

* Changed frame naming from `%04d` to `%06d` to prevent indexing collisions and ensure compatibility with long video sequences.
* Implemented persistent metadata caching (LX, LY, FPS, Frame Count) via `.txt` files in the cache folder and `mutable` class members. This prevents redundant `ffprobe` calls that were causing UI lag and potential race conditions during frame loading.
* Improved the `getImage` logic to ensure the `QImage` is fully loaded and correctly converted to ARGB32 before being copied to the Toonz raster, preventing the "blank frame" syndrome.

### DIFF
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -163,11 +163,9 @@ void Ffmpeg::createIntermediateImage(const TImageP &img, int frameIndex) {
         QObject::tr("Failed to create QImage for intermediate file"));
     return;
   }
-
   // qi = qi.copy();
   // Currently redundant,  removed for performance & clarity.
   // If raster-to-QImage sharing is needed in the future, we can revisit.
-
   QByteArray formatBa = m_intermediateFormat.toUpper().toLatin1();
   const char *format  = formatBa.constData();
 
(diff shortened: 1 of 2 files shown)


RodneyBaker:
This looks good @andeon 
I was expecting just a fix to the problem but you've plussed everything up even more.
Much appreciated.
LGTM

Aside:  I was a little concerned about `Changed frame naming from %04d to %06d` but I see why you did that and how it was implemented.

andeon:
@RodneyBaker during testing, I used an an example [MP4 file](https://file-examples.com/index.php/sample-video-files/sample-mp4-files/ )  (like this [one](https://file-examples.com/wp-content/storage/2017/04/file_example_MP4_480_1_5MG.mp4)) with about 900 frames, despite it being a small file.

This made me think of users who convert video clips to use as background elements, such as a TV screen or a backdrop panel. Since the quality is low, users might extract very long sequences that push the frame count into 5 digits. While this might not be a common workflow, addressing it now could avoid potential issues down the road."

RodneyBaker:
This was really great thinking on your part and because it's MP4 we are dealing with we aren't running up against the standard 4 digit output of regular image sequencing.  That is definitely waiting in the wings however.

More digits will be needed as we move forward and some likely want that already but the trick will be not to break the classic 4 digit standard.  BUT... here's the kicker:

One of the reasons for 4 digit limitation is that more frames than that will slow most systems to a crawl unless/until we can optimize for more memory use (or use of dynamic proxies).  I also reason that is why the zoom slider for the xsheet only allows 72 frames to be shown at a time on most screens.  It intentionally limits the number of frames the user is interacting with at any given time.  Not optimal but smart.

As Opentoonz gains more 'video editing' capability... which seems inevitable... some approach to using proxies like Davinci Resolve will likely be needed and yes, more digits in sequence filenames.

PR #6596, opened 2026-02-01
Custom panels : add 22 new UI templates for the Custom Panel Editor

Ahl76:
This PR adds 22 .ui templates to the custom panel templates folder:

* Grids
* Radials
* Bars (vertical / horizontal)
* Minis

Based on Shun Iwasawa's work in #4495.
Related to #6594 and #4373.

Some exemples 

Custom panels exemples

Note: Given the number of templates, I'm considering future improvements to their organization (folders or collapsible sections).The icons in these templates are slightly larger than usual I also wanted to add a GUI option to show/hide the background (transparency), but I haven't been able to work as expected.  Feedback is welcome.

PR #6594, opened 2026-02-01
UI: Bind-to-Room system, Panel Mirroring, and Dynamic Shortcuts

Ahl76:
*I am not a developer, but I wanted to contribute to OpenToonz using AI. I've tried my best to follow the project's coding style. Please let me know if there are any bugs, logic issues, or things to improve. If the work isn't acceptable (stability, performance etc.) for the project, feel free to decline it.*

This Pull Request introduces several interconnected features intended to enhance the flexibility and efficiency of the OpenToonz interface: related to #5035 #5033 #4495 #4373

### 1. Bind to Room (BTR) Toggle Feature - Dynamic Panel Visibility

Bind to room toggle feature

Initially developed to improve custom panels and subsequently generalized across the software.

Allows users to bind or pin/link panels to specific workspaces (rooms) via a toggle button in the title bar and/or a right-click context menu option (Bind to Room). Bound panels automatically show or hide when switching rooms. When inactive, the behavior remains identical to the current OpenToonz standard ( No changes ). It's just like local vs global.
##### Preferences

The button can also be hidden in the preferences (Interface / Show room bind buttons). In some cases, a restart is required for the update to take effect across all panels.

##### Benefits

This reinforces the existing room system in OpenToonz:

- More modular workspaces
- Less visual clutter
- Accelerated workflows

### 2. Dynamic Shortcuts - Dynamic Presets and Size Shortcuts

Automatically registers all brush preset, custom panels and 23 universal sizes (0.5-1000px) as assignable keyboard shortcut commands. These are also available for use with custom panels in custom panel editor, using clear prefixes like [Panel] or [Size] to keep them organized. 

I struggled a bit with the original behavior of this section; I've tried to restore the initial behavior, but please double-check if it looks okay to you

📁 Menu Commands
  ├─ 📄 File
  ├─ 📄 Edit
  ├─ 📄 Scan & Cleanup
  ├─ 📄 Level
  ├─ 📄 Xsheet
  ├─ 📄 Cells
  ├─ 📄 Play
  ├─ 📄 Render
  ├─ 📄 View
  ├─ 📁 Windows
  │   └─ 📁 Custom Panels           ← 3ème niveau !
  │       ├─ [Panel] My Custom Panel 1
  │       ├─ [Panel] My Custom Panel 2
  │       └─ ...
  └─ 📄 Help

📁 Tool Modifiers
  ├─ 📁 Brush Presets
  │   ├─ [Raster Brush] MyPreset1
[...]

PR #6627, opened 2026-02-16
Preset management panel, custom panel removal, dynamics icons/shortcuts 

Ahl76:
This PR is a continuation of the implementation started here #6594 / #6596  and brings several improvements:

### **Brush Preset Panel Management**

Improvement of the brush preset panel by allowing page and preset organization features similar to the level palette.

Tab and page management: Organization now allows sorting brushes across multiple pages with naming freedom. Some features and UI available:

- Right-click on a tab to: add or remove a page.
- "Physical" button to add a page.
- Double-click on a tab to: rename the page.
- Ctrl + Drag on a tab to: reorder the pages.
- The last selected preset is remembered per page during navigation.
- A palette icon is displayed to the left of each tab.
- A confirmation/warning message indicates the exact number of presets deleted when removing a page.

https://github.com/user-attachments/assets/f34ede18-e7a9-4d32-81d2-1a88007e74cf

**Actions on presets:**

Activation of the [+] and [-] buttons integrated in the previous PR.

- [+] Button: Adds a new preset.
- [-] Button: Deletes the selected preset.
- Right-click on a preset to:
    - Cut / Copy (to internal clipboard) and rename.
    - Paste (to the current page).
    - Delete (permanent removal).
    - Move to Page (direct transfer to another page via a sub-menu).

https://github.com/user-attachments/assets/15e38a12-d915-4596-a1d4-0cd04ffdaec9

**Multiple selection and Batch mode:**

- "Show Checkboxes" option in the hamburger menu to display checkboxes.
- Clicking a checkbox selects the item without loading the brush on the canvas.
- Right-click in multiple selection mode allows applying actions (Copy, Cut, Delete, Move to Page) to the entire checked group simultaneously.

https://github.com/user-attachments/assets/dcf921b3-70f9-4ce0-8f84-764b48764979

Preset information improvements (snapshot). When adding a preset, an instantaneous capture is taken. It records the complete state of the brush, including almost all parameters of the tool options bar as well as detailed style editor settings (styles, textures, mypaints  and specific parameters etc. with the exception of color (except for textures)).

**Dynamic nomenclature:**

Improvement of the displayed label detection in real-time according to the active style:

- Vector: Vector Brush, Custom Vector Brush, Generated, Trail, Texture.
- Toonz Raster: Toonz Raster Brush, Texture, MyPaint Brush Tnz.
- Raster: MyPaint Brush, Raster Brush, Raster texture.

I added Raster Texture even if not truly supported by textures because it allows a workflow that can be useful with color.

**2. Dynamic Icon Rendering**

Visual improvement allowing icons (sample strokes) to reflect the real tool settings when saving the preset.

Operating logic and Fallback:

- Activation via the "Use Sample Strokes" option in the hamburger menu (saved preference).
- Priority to custom icons if they exist on disk as previously.
- If disabled: uses generic icons according to the level type as previously.
- If enabled: generates a dynamic preview based on the brush settings.

Universal Sizes Rendering ([Size_]):

https://github.com/user-attachments/assets/25323f97-d02d-4168-b3b4-38f8f2572242

- Draws a circle with the size number below to facilitate the use of universal sizes directly without custom icons actions.
- Visual evolution: full circle for small sizes, prominent semi-circle for large sizes.

Sample Strokes Rendering:

- Draws a sine wave reflecting:
    - Thickness, Opacity and Hardness.
    - Tapering driven by the min/max thickness ratio.
- Visual adaptation according to the brush type:
    - Vector: Perfectly smooth line.
    - Raster: More pronounced grain effect.
    - Toonz Raster: Subtle grain with reduced alpha.

Note: Icons and sample strokes are also available with custom panels.

**3. Custom Panel Management & UI Visibility**

Panel Removal (Remove Button):

- Click on "Remove": Displays the list of existing panels.
- Confirmation: Physical deletion of the interface file (.ui).
- Field for the selected or available custom panel and collapsible menu to select multiple panels at once.
- Automatic cleanup: The command disappears from menus and keyboard shortcuts (in "Configure Shortcuts" and in the "Custom Panel Editor").
-
remove custom panel

4. Customizable Room Shortcuts

Follows a feature request related to shortcuts, under the same principle as the Panels prefixes etc. a room category has been created with a room registration system similar to custom panels.

- Dynamic Registration: Automatic creation of a command for each workspace (Room).
- Identification: Use of the [Room] prefix in shortcut settings.
- Navigation: Instant switching between layouts via the keyboard.


RodneyBaker:
@Ahl76 There are lots of great additions here.
I've been exploring and enjoying what you've added.

RodneyBaker:
The following isn't a problem with this PR as much as it is an issue with how Opentoonz styles and indexed color works, especially with vector levels:

Let's assume we have a number of styles defined in our palette and have drawn a number of frames in the xsheet with those various styles.

Next we select the default (No. 1) style to continue to draw

Next we open the Tool Brush Panel and select a preset.

Now we note that everything drawn with style 1 will update its style to our newly selected preset.

The problem:  We don't likely want those previously drawn images to update to the current preset.

It isn't clear how we can get past this autoupdating of linework.
The concern is that the user might not immediately realize they are reassigning styles to linework and once they do realize it returning to that previous style may be very hard if not impossible if we haven't previously saved that specific preset to return to.

A thought in the direction of a solution might be to supply a default preset that captures the default brush.  At least then the user would be able to return to that default.  This doesn't resolve the potential problem but would allow users who use the default vector brush to have a safe-haven.

Hope the outline of the problem is clear.

Added:  Locking the palette does not prevent the autoupdating of vector styles via presets.

andeon:
@Ahl76, it might be interesting to have an option for the active style with a toggle to disable automatic property updates. Sometimes, a user wants to use configurations of a preset (like size or other brush options) without actually changing the underlying style properties to like how it functioned in previous version. I believe this would partially resolve the issue @RodneyBaker pointed out, as it gives us the choice to avoid altering the style accidentally when we only want to swap the brush dynamics.

This feature is very usefull essentially allows you to draw directly in OpenToonz, not that it wasn't possible before, but the use of presets enables seamless swapping. This is vital because managing endless palettes in "Raster levels" is incredibly tedious, especially since the final artwork ends up integrated anyway. It's a massive time-saver for both the general public and independent artists. I believe some of these workflows eventually find their way for studios pipelines too.

RodneyBaker:
Ok.  The 'Show Checkboxes' option prevents auto-updating the style.
I missed that somehow.
What I THINK I see is that when the 'Show Checkboxes' option is on styles don't auto-update.
When 'Show Checkboxes' is off the styles do auto-update.

At this point I'm not sure I actually know what the individual checkboxes are for so I'm reading back through your PRs to see what I've missed.

Here's the pertinent part:

>Multiple selection and Batch mode:
> "Show Checkboxes" option in the hamburger menu to display checkboxes.
> Clicking a checkbox selects the item without loading the brush on the canvas.
> Right-click in multiple selection mode allows applying actions (Copy, Cut, Delete, Move to Page) to the entire checked group simultaneously.

andeon:
I think it is necessary to lock color 0 (style 0) from being edited. It seems to cause anomalies when saving and reloading the scene if style 0 is changed. This style should only be modified through what the `Tool Options Bar` allows.

Ahl76:
@RodneyBaker @andeon

Thanks for your feedback and remarks. Here is a proposal that addresses the points raised with two toggle modes and a "Plain Color" option (available aswell in right click menu). 

1. "Non-Destructive" 

In order to protect existing content/strokes.
The system scans the palette to check if an identical style already exists to reuse. If no match is found, it automatically creates a new index instead of modifying the current style.

Point of attention: Be aware that with this mode enabled, the level palette is modified without direct user intervention. This is an automation of what would be done manually to preserve existing work.

2. "Selective Preset" 

This mode allows changing the brush without touching the current style.
Principle: It only applies the Tool Options Bar settings (size, pressure, etc.). Style Editor data is ignored, but the brush type (Native, Mypaint, Texture) is still preserved.
Note: For now, MyPaint brushes, dynamics settings are only compatible with other MyPaint presets due to their single-slider parameter structure.

3. "Plain Color / No Style" Option

A quick action icon allows clearing the selected style's overlays to return to the native style.

If the names are ambiguous, we can change them. Some possible tracks (even if they might also be ambiguous):

For Non-Destructive mode:

Protect Existing Strokes, Auto-Match Style

For Selective Preset mode:

Tool Parameters Only,  Selective Settings, Brush Settings Only, Dynamics Only

Additional Notes

Undo/Redo Management: An initial implementation for Undo/Redo of icon-related actions (notably when Non-Destructive mode is disabled) has been added, but this part is currently incomplete. I can remove it if it's confusing without the rest.

Known Issues:

On Raster Levels: With non-destruction mode enabled, the automation and scan can fail if the native style has been manually deleted (redstyles), requiring manual intervention to recreate it.

Ahl76:
> I think it is necessary to lock color 0 (style 0) from being edited. It seems to cause anomalies when saving and reloading the scene if style 0 is changed. This style should only be modified through what the Tool Options Bar allows.

@andeon 

Is this a problem you encountered specifically while testing this PR? Because it doesn't seem to me that Style 0 can even be modified here ?

andeon:
For test:

1. Create a new Toonz Raster level.

2. Select a MyPaint brush and configure its settings.

3. Save this brush as a Preset.

4. Apply this preset to Color #0 (Style #0) (Brush Preset window). 

This forces Color #0 to adopt the current brush properties, replacing its default plain style. (On Vector levels, can be replicated using Generated, Trail, or Vector brush too)

In OpenToonz, Color #0 is strictly intended as the primitive transparency style (the 'invisible ink'). Under normal circumstances, you should not be allowed to change its plain style or delete it. 

Technically, this logic should also be extended to Color#1. Because Color#1 is the default system style used for Line Processing and Cleanup, experienced users generally avoid change it. However, Color#1 ends up being used more often because it doesn't have the same strict restrictions as Color #0; while you are not allowed to delete it, the software doesn't strictly block you from changing its style.

Ahl76:
@andeon

Thanks for the steps, I can reproduce and see the problem. I will provide a fix to restore Style #0 to the way it used to function. Have you noticed any other regressions so far?

JParks123:
@Ahl76 I am getting crashes using this Windows build every time I attempt to use the Fill tool. I also get a lot of crashes using the room shortcuts unfortunately I only tried them so far mapped to the num keys, I still love the feature hopefully you can still get it to work without all the crashes.

RodneyBaker:
New release schedule has been posted.
REF:  https://github.com/opentoonz/opentoonz/discussions/6357#discussioncomment-15966129

Ahl76:
@JParks123 

​Sorry for the late reply.

​On my end, I don't experience any crashes in my various environments on Windows 10 or Windows 11, including the Windows build and the artifact from this PR. Whether it's with the room shortcuts or the toolbar icons, I can't get it to crash.

​Can you provide the precise steps that lead to your crashes? I suggest you go to the OpenToonz stuff/ config folder to delete the text files related to the brushes, in case they might be corrupted, or try a clean installation from A to Z. If this hasn't been done yet, it might help to resolve the issue.

Ahl76:
@RodneyBaker could you re run ?

RodneyBaker:
@Ahl76 No, only the original author (you) and @shun-iwasawa (and @manongjohn) can trigger Appveyor.

If you close and the reopen the PR it should re-trigger the checks.

RodneyBaker:
This will likely be controversial but...
I'm looking to merge this PR as it addresses issues noted in prior PRs related to custom panels and shortcuts.

I note that @JParks123 did not supply specifics related to crashes experienced so cannot specifically test for that.

JParks123:
@Ahl76 @RodneyBaker sorry for the late response to create the crash with the Rooms shortcuts you would have to first set your shortcuts and then switch Rooms from Two very opposite ends like the first Room and then the last Room especially if you mix the Task room in there that one seems to be the buggy Room to mix in with back in forth shortcuts.

Ahl76:
Could you provide a video showing what you’re doing? Unfortunately I can’t reproduce the crash by following your instructions, although I have noticed some conflicts with keyboard shortcuts.

RodneyBaker:
@Ahl76 As a draft PR this cannot move forward.

Disclaimer:  Current PRs may represent proposals which allow inclusion in nightly release for the purpose of thorough testing and feedback from the Openntoonz community.  This is with an understanding source code can and will change as the software approaches the next full release.

RodneyBaker:
Merging with the understanding this update falls into the Experimental category of v1.8 features.  This suggests elements of the experimental feature set are likely to change before the next full release.

Issue #4544, opened 2022-09-24
Let User Choose Color of Paint Check


konero:
### Is your feature request related to a problem? Please describe

Working on high brightness, vibrant color displays the current color for Paint Check (pure red 🟥) is eye blistering.

Not sure if the photo shows the problem or not, but it's physically painful to stare at.

![image](https://user-images.githubusercontent.com/19820721/192115250-a29141a3-391b-414c-ab35-04e8a295508b.png)

This feature is super useful.

### Describe the solution you'd like

Letting the user choose the color in preferences. 👌

RodneyBaker:
Agree that Paint Check is VERY useful.   I hand't realized just how useful.
I also think the paint check color should e customizable.

This request represents one of several that would improve the various check processes.

At any rate, I believe this is where the Paint Check color is set:  https://github.com/opentoonz/opentoonz/blob/5aa5de6505d2612661466a707fdfd8392082cae5/toonz/sources/common/tvrender/tglregions.cpp#L263

RodneyBaker:
Recent update relate to Transparency Check should help us hone in on how best to approach this and other similar updates to the check features.  REF:

PR #6633, opened 2026-02-22
feat: Add preference to customize Ink and Paint Check colors

andeon:
This PR allows users to select custom colors for the **Ink**, **Ink#1**, and **Paint Color Check**. Previously, these were hardcoded to pure red, which was reported to cause eye strain on high-brightness displays and could be difficult to see depending on the color palette choose.

ink_paint_check

- Added a new color setting in the Preferences menu for **Ink**, **Ink#1**, and **Paint** colors.

- Modernized signal/slot connections to use Qt5 syntax where applicable, removed using `namespace std; `See: [6311 ](https://github.com/opentoonz/opentoonz/issues/6311 )

-  Translated all remaining Italian and Japanese comments to English to improve maintainability

- Resolved an issue where the application would crash when using the Select Tool to delete a drawing area selection while `TLV Save Box` (preferences) and certain check options (like Fill Check) were active.

This was actually done a while ago, but it slipped through the cracks while I was focusing on more critical fixes. I'm adding it now, since there's still some time.

Fixes: https://github.com/opentoonz/opentoonz/issues/4544

### DIFF
--- a/toonz/sources/toonzlib/stagevisitor.cpp
+++ b/toonz/sources/toonzlib/stagevisitor.cpp
@@ -129,7 +129,7 @@ QImage rasterToQImage(const TRasterP &ras) {
 
 //! Draw orthogonal projection of \b bbox onto x-axis and y-axis.
 void draw3DShadow(const TRectD &bbox, double z, double phi) {
-  // bruttino assai, ammetto
+  // Not the most elegant solution
 
   double a = bigBoxSize[0];
   double b = bigBoxSize[1];
@@ -454,13 +454,10 @@ TRasterP RasterPainter::getRaster(int index, QTransform &matrix) {
 //-----------------------------------------------------------------------------
 
 /*! Make frame visualization.
-\n	If onion-skin is active, create a new raster with dimension containing
-all
-                frame with onion-skin; recall \b TRop::quickPut with argument
-each frame
-                with onion-skin and new raster. If onion-skin is not active
-recall
-                \b TRop::quickPut with argument current frame and new raster.
+\n If onion-skin is active, create a new raster with dimension containing
+all frame with onion-skin; recall \b TRop::quickPut with argument
+each frame with onion-skin and new raster. If onion-skin is not active
+recall \b TRop::quickPut with argument current frame and new raster.
 */
 
 namespace {
@@ -486,9 +483,8 @@ void RasterPainter::flushRasterImages() {
   int lx = rect.getLx(), ly = rect.getLy();
   TDimension dim(lx, ly);
 
-  // this is needed since a stop motion live view
-  // doesn't register as a node correctly
-  // there is probably a better way to do this.
+  // this is needed since a stop motion live view doesn't register
+  // as a node correctly there is probably a better way to do this.
   if (rect.getLx() == 0 && lx == 0) {
     rect = m_clipRect;
     dim  = m_dim;
@@ -516,8 +512,7 @@ void RasterPainter::flushRasterImages() {
   // Clear the buffer - it will hold all the stacked nodes content to be overed
   // on top of the OpenGL buffer through a glDrawPixel()
   ras->lock();
-
-  ras->clear();  // ras is typically reused - and we need it transparent first
+  ras->clear();  // ras is typically reused and we need it transparent first
 
   TRect r                 = rect - rect.getP00();
   TRaster32P viewedRaster = ras->extract(r);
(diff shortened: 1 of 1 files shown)


DarrenTAnims:
This is a great addition.
While on the subject of colours, one common complaint I see is that the Animation tool widget is often not easily visible. It's shown in pink, which isn't always clear. Would it be possible to add an option on this page to change that? Or to add a colour that you can change in a stylesheet (or both)?

andeon:
@DarrenTAnims The color for the Animate Tool widget is hardcoded in the source code at this specific line:

https://github.com/opentoonz/opentoonz/blob/ed4aa37c8954655b3aaac04cb50b5ef4e1de9d9d/toonz/sources/tnztools/edittool.cpp#L1302

I believe the best approach to make it customizable would be in Preferences, as the Tool Options bar is already quite cluttered (especially in 'All' mode, which uses a lot of options). Additionally, it would be great to have an option to increase the widget size. On high-resolution monitors, the handles are very difficult to see and grab. I think if it were larger, most users would prefer to just use the 'All' mode, as it's much more practical, especially when using a tablet and pen. I'll check the best way to implement this for a new, separate PR.

RodneyBaker:
For those looking into the code, note that the majority of changes relate to commentary and translation.

Aside:   I couldn't help but be reminded that while not changed in this PR, the autosave code in the files touched appears to be best practice which was a major win in earlier updates.  It kept autosave from triggering mid stroke which had been a big problem. Those with autosave problems should definitely be using the nightlies versus older releases.

There seems to be a lot of raw pointers in these files.
If they are part of QObjects or singletons then perhaps not problematic but if not... potential sources of leak.

There is room for additional optimization where there are calls that trigger redundant refreshes.  This in the slot calls:  `m_currentTool->onImageChanged(...)`
Not sure what the best course for additional optimization would be short of just finding ways to bypass those calls. This of course assumes the user wouldn't notice the lack of those refreshes.

Another aside, not really related to this PR, this:
```
if (xsh->getColumn(col) && xsh->getColumn(col)->getSoundColumn())
  return TImage::VECTOR;
```

Reads as odd to me.
This may be an Opentoonz specific return to avoid some potential conflict.
To my untrained eye, it would be good to have a comment here to explain this particular usage as it appears to be a Toonz specific requirement.

Might this work just as good or better?:
```
TXshColumn *column = xsh ? xsh->getColumn(col) : nullptr;
if (column && column->getSoundColumn()) {
  // NOTE:
  // Sound columns are not image columns but many tools/UI paths still expect
  // valid TImage::Type and can behave poorly with TImage::NONE.
  // We intentionally return VECTOR as a compatibility fallback so the tool
  // framework remains stable when an audio column is selected.
  return TImage::VECTOR;
}
```
All this to say, this PR appears to deliver as advertised.
I do think as we start moving toward a version 2.0 we should review the various checks with a goal to view them from the broad scope of usage particularly when more than one check is being used.  That and... I'd dearly love to see users be able to render with the checks included in the output.  This will be even more important as checks continue to be improved.

Something I am still testing:
I think I'm seeing an issue with the checks (particularly Fill Check) and the TLV Save Box.
There is much I don't know about that Save Box option and it also could relate to the github build.

Update:  We may want to check with @Abel032 on this but the issue appears to relate to this newer option in Preferences:  
image

Added:  In looking deeper I believe this may relate to my tablet (driver) somehow not refreshing parts of the screen.  I'll restart my computer and test more.

andeon:
I have identified a significant Signal Bottleneck in TApp. The current architecture triggers redundant refreshes during high-frequency user actions.Maybe in a  future refactor is needed to implement Batching/Coalescing and to modernize the pointer management (RAII) to improve both performance and maintainability. The problem is that there are a lot of things that need to be refactored or reviewed. When you change one file, you discover other things that could be implemented or revised.

andeon:
@RodneyBaker, To be honest, I don't usually use "TLV Save Box". I've noticed that when this and Fill Check are both enabled, the application crashes when I try to select "select tool" a drawing and delete it. I'm test some old versions and still there not sure when this bug was introduced. Maybe related the save box implementations.

RodneyBaker:
> I have identified a significant Signal Bottleneck in TApp.

Tapp.h touches a lot of processes.
In many ways that's a good thing and it establishes context for many tools and processes but presents some significant risk as well if/when changed.  My sense is that refactoring Tapp can be a source of significant optimization but we surely would be targeting version 2.0 timeframe as changes to Tapp will have wide effects across Opentoonz.

I'm intrigued by the possibilities but don't know enough to suggest how best to move forward with Tapp short of 'move forward very carefully'.

RodneyBaker:
>  I don't usually use "TLV Save Box". 

I have never used 'TLV Save Box' short of in testing as I've never quite understood or needed to adjust it.  Obviously it has its uses but I find much of the result of using it confusing.

@Abel032 might be able to shine some light on this for us as he's recently been working with it programmatically.

andeon:
@RodneyBaker, I decided to include this fix directly in this PR because it’s closely related to the current work. Opening a separate PR would be a waste of time and would require redundant testing, as this validation check solves a crash specifically triggered when TLV Save Box and some Checks are active together.

Better to address it here while the context is fresh. I suggest avoiding a squash merge so this specific fix remains clearly documented in the history.

RodneyBaker:
Thanks @andeon 
My only concern in cases of adding fixes directly into feature PRs, and it's a very minor concern here, is that the fix can get lost.  Your suggestion not to squash merge should address that and make the fix easier to see.

Will test.

RodneyBaker:
While there is no rush... posting intent to merge.

image

RodneyBaker:
Pressing forward with this with the anticipation that we'll update documentation of the Ink and Paint check options in light of this addition.

Thanks @andeon

RodneyBaker:
@andeon here's another bug hunt challenge.
It has been reported that nightlies form this PR forward have a broken Close Gap feature for the Fill Tool.

I have not yet confirmed this report but by the time I can you might be able to spot the problem and have it fixed.

To test:
- Draw shapes with mostly closed lines or open small gaps in drawn shapes
- Change to Fill Tool and make sure Close Gap checkbox is toggled on (Refer Fill etc off)
- Use Freehand mode to circle the shapes and see if it will fill the shapes with color
- Test this same process in previous nightlies

That's all I have at the moment.

Thanks!

JParks123:
@andeon  by changing the color of all the line checks Red by default you pretty much break all checks  from working the way they are designed to work.  
For example: the red color and the color of the Tapes look identical. 
image
 it makes the lines for all checks red now.
it needs to be default Black for the ink lines to work correctly with the different checks. 
image

andeon:
@JParks123 Sorry, I don’t quite understand what you mean. These colors were already **red** by default before, so I kept them the same to avoid breaking the user experience. If users don’t like them, they can adjust the colors to their preference.

Of course, this could be adjusted through a UX review, but I’d prefer to leave that to @konero  or @shun-iwasawa , since they have access to the studio where people can better evaluate it. That said, I do agree that this default red is harder to see against the pink gap for example.

JParks123:
> @JParks123 Sorry, I don’t quite understand what you mean. These colors were already **red** by default before, so I kept them the same to avoid breaking the user experience. If users don’t like them, they can adjust the colors to their preference.
> 
> Of course, this could be adjusted through a UX review, but I’d prefer to leave that to @konero or @shun-iwasawa , since they have access to the studio where people can better evaluate it. That said, I do agree that this default red is harder to see against the pink gap for example.

it default makes all the different checks lines red, the original line check is already red yes, but this change makes every different check have red lines including the ones that were originally Black, this includes gap and fill checks, rendering them very hard to use, the bottom two were default red but the top ink check was default black and you turned it red which breaks all the checks now.

andeon:
I'm testing an older build and it appears red for me. I created three new styles and the selected color updates the ink to red. I may have done something wrong, but I’m not seeing this black.

andeon:
Now I understand what you mean. If only the “Gap” option is enabled, the inks turn red it need be black. I gonna check this.

RodneyBaker:
While I'm not seeing what @JParks123 is suggesting, what I think I'm seeing is that Ink#1 Check Color is being ignored and/or overridden by Ink Check Color.

To test:
Turn on Ink No1 Check (but not In Check Color)
Set the color of Ink No1 Check Color in Preferences to Green or some other color that should not appear with checks, etc.  
Draw anything with Style No 1 
Note that all lines drawn with Style No 1 get Ink Check Color's color and not Ink No1 Check's Color.

In other words, I'm not seeing any time in which Ink No 1 Check Color is appearing.

Edit:  Ok, now I'm confused but perhaps onto something.  Ink No 1 Check Color is appearing when Style 2 is being used.  Perhaps this is an off by one error?

andeon:
@RodneyBaker I need to check this. I think something is not properly disabling the inks. Right now the inks seem to be active together. For example, if Ink#1 is active, the other should be inactive, and vice versa. Right now it looks like they are working together.

Issue #6626, opened 2026-02-16
Exit full screen window mode crash


truecuckoo:
### What's the bug?

Accidentally went full screen on multi viewer. Exit full screen. Then crash with the following report

### Steps to Reproduce

I can reproduce on demand. I'm on Mac OS 15.7.3
1. Drawing on Opentoonz raster layer with Wacom Cintiq 24 stylus.
2. Use stylus button mapped to Right Click to bring up menu.
3. Upon entering full screen I noticed that another menu was displayed instantaneously, that says "GUI Show / Hide" suggesting that there was a lingering button press in the system.
4. I click outside this menu, to avoid it.
5. Then I clicked the Stylus menu button again to bring up the menu, and select exit full screen.
7. Crash
8. As I exited the crash report window by pressing quit, I noticed that a line was being drawn on the drawing before it was shut down. Perhaps this whole operation was taking place at the same time as a brush stroke was being made. Maybe the Wacom both draws and opens a menu at the same time, and causes a conflict?
9. It happens 100% every time with the Cintiq stylus. It never happens with the trackpad.

<img width="323" height="252" alt="Image" src="https://github.com/user-attachments/assets/2e57db85-d2b3-450a-a1b5-56365b33b289" />
Wacom Cintiq driver version 6.4.9-2

### What did you expect to happen?

Exit full screen, no crash.

### Screenshots, Videos or Crash Logs

[Crash-20260216-173504.log](https://github.com/user-attachments/files/25346053/Crash-20260216-173504.log)

### OpenToonz Version

1.7 nightly build

### Exact Version Info (optional)

1.7.1 nightly build 31. january

### Operating System

Mac OS 15.7.3

### GPU Brand

Apple M1 Pro 10 core

### Graphics Tablet

Wacom Cintiq 24"

andeon:
Hi @DarrenTAnims, sorry to bother you, but could you please test this specific crash scenario on the latest Windows Nightly Build to see if it's cross-platform? On Mac, the logs aren't always clear about the exact pointer failure. If it crashes on your end in Windows, could you share the crash report? 

 It would help determine if this is a core Qt event handling issue or something specific to the macOS mapping.

As we are approaching the v1.8 release, We trying to help eliminate as many crashes as possible before the code freeze. Thanks for the help!

DarrenTAnims:
I've just downloaded it and it works fine for me.
I can record a video and share it on one of the Discord servers, if that'll help.

[...]

PR #6639, opened 2026-02-28
Prevent crash when toggling Full Screen with Wacom Stylus

andeon:
This PR fixes a crash  on Mac where the program would close unexpectedly if you used your stylus button to open a menu or exit full-screen mode.

This fix is a port of some logic from Tahoma2D, combined with some trial-and-error adjustments. If this initial change doesn't fully resolve the issue, I may need to port more related code from Tahoma2D to ensure a complete fix. I don't have a macOS device to run tests myself, so it's a bit of a shot in the dark.

I would really appreciate it if @truecuckoo could test this to see if it fixes the crash in this first phase, If it's fix the crash related, I'm also asking other tablet users to check and ensure it hasn't introduced any unintended side effects or changed tablet behavior.

Fixes: https://github.com/opentoonz/opentoonz/issues/6626

### DIFF
--- a/toonz/sources/toonz/sceneviewerevents.cpp
+++ b/toonz/sources/toonz/sceneviewerevents.cpp
@@ -67,7 +67,7 @@
 #include <QGestureEvent>
 #include <QPainter>
 
-// definito - per ora - in tapp.cpp
+// defined currently in tapp.cpp
 extern QString updateToolEnableStatus(TTool *tool);
 
 //-----------------------------------------------------------------------------
@@ -79,7 +79,7 @@ void initToonzEvent(TMouseEvent &toonzEvent, QMouseEvent *event,
   toonzEvent.m_pos      = TPointD(event->pos().x() * devPixRatio,
                                   widgetHeight - 1 - event->pos().y() * devPixRatio);
   toonzEvent.m_mousePos = event->pos();
-  toonzEvent.m_pressure = 1.0;
+  toonzEvent.m_pressure = pressure;
 
   toonzEvent.setModifiers(event->modifiers() & Qt::ShiftModifier,
                           event->modifiers() & Qt::AltModifier,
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thanks @andeon 
We also need to look at the nightly build for MacOS as well as that has been quite unreliable.
I hesitate to merge this as doing so can make finding the MacOS build more difficult.
We can certainly point users to the github action artifacts however.

On the positive side, merging will allow Windows users to confirm no adverse effects on that end.

andeon:
@RodneyBaker I hadn't noticed this issue, it was introduced with PR https://github.com/opentoonz/opentoonz/pull/6594

When toolpresetcommandmanager.cpp was added, it started calling MyPaint functions directly from the main executable. If the file (CMakeLists.txt) isn't updated to tell the Linker. The build breaks at the final stage. I'll provide a fix for this.

truecuckoo:
I’m currently travelling. Will try when I get back, the following week. I’m not very GitHub savvy. So if there’s no clear downloadable installer I have no experience compiling or replacing only specific files etc, unless there’s a clear explanation.

andeon:
@truecuckoo, You can download the latest automated build (Artifact) directly from this PR here: https://github.com/opentoonz/opentoonz/actions/runs/22523965957/artifacts/5704285237

**Note:** This specific artifact is based on macOS-15-Intel and comes as a DMG file, which acts as a portable version for Mac. You must be logged into GitHub for the link to work.

In Pull Requests like this, you can usually find pre-built files for Windows, macOS, and Linux. Testing these builds is a great way for users to help developers confirm if a bug fix actually works on different hardware before the code is officially merged!

RodneyBaker:
The only reason I can see for this PR not moving forward is failure to pass checks.
Even though this is a Mac specific fix it should still pass Appveyor check.

andeon:
@truecuckoo, this is still pending. If you can let us know whether this fixes the crash you were experiencing, we can evaluate whether it should go into the new release or be postponed. Since this is specific to macOS and your Cintiq hardware, we’re not able to validate it on our side. Thanks!

RodneyBaker:
@andeon 
Would the reason this is failing on Appveyor check be that the code is limited to MacOS?
From what I can tell, the Appveyor check should still successfully pass.

andeon:
This is just a server connection issue that didn’t complete the build.

However, we need to confirm on macOS whether this fix works, since it’s specific to that platform. If it doesn’t work there, there’s no need to implement it, as it will require further adjustments or a different approach.

PR #6640, opened 2026-02-28
Fix MyPaint linker error on macOS build

andeon:
This PR resolves a regression introduced in PR #6594, reported by @RodneyBaker, which caused "undefined symbols" errors during the final linking stage on macOS. It ensures the MyPaint library is correctly propagated through the build system.

- Root CMakeLists.txt: Updated to export `MYPAINT_LIB_LIBRARIES` to the internal cache

- toonz/CMakeLists.txt: Added `${MYPAINT_LIB_LIBRARIES}` to the target_link_libraries for the Apple target


RodneyBaker:

Issue #2852, opened 2019-10-21
Change zise Animate tool in opcion "All"


zarco1794:
the image size to Animate tool in the opción "All" it s sou small, take control is very dificulty. maybe it would be better to increase the size of the "animate tool All" icon
![1](https://user-images.githubusercontent.com/39812147/67245083-898ffa80-f431-11e9-9829-5f324c28cfe8.JPG)
It is very difficult to use the "Mass" option
https://media.giphy.com/media/Wt0sKUnXqFl5GNTqTZ/giphy.gif 

![2](https://user-images.githubusercontent.com/39812147/67246467-e80aa800-f434-11e9-9fc3-10ef47d70ec8.JPG)

beeheemooth:
I agree with you: the manipulator is really very small. In addition, despite the fact that it is bright in color, it is sometimes difficult to find.

gab3d:
Other software usually have a way for the user to adjust the manipulator size, perhaps that could be implemented in OT.
as the + and - are already used for the viewport zoom, perhaps a modifier key could be used, such as Ctrl, Shift, whichever... together with the + and - keys.

artisteacher:
Personally the size doesn't bother me. I do think that sometimes the pink color blends in too much - especially if there is a background at a similar value. The widget might benefit from being two-toned, like how the selection box handles were changed in #2804.

zarco1794:
The "shift" command could be added for greater configuration accuracy: if I select "All" and want to rotate the image, pressing "shift" at the same time to make the rotation slower, and have more precision

shun-iwasawa:
@zarco1794 

> The "shift" command could be added for greater configuration accuracy: if I select "All" and want to rotate the image, pressing "shift" at the same time to make the rotation slower, and have more precision

This feature is planned to be introduced. However I'm thinking to use the "Alt" modifier for it.
When using the tool with the Position mode, "Shift" is already occupied for moving along vertical / horizontal axis and "Ctrl" is for moving Z value.

RodneyBaker:
Reference:  #1886

zarco1794:
maybe this size
![1](https://user-images.githubusercontent.com/39812147/67607758-fe2da680-f75b-11e9-8285-96f120b3ec5b.JPG)

size actuallity
![arrgnzo](https://user-images.githubusercontent.com/39812147/67607862-56fd3f00-f75c-11e9-8146-e8ecbe8e36e0.jpg)

[...]

PR #6641, opened 2026-03-01
feat: Add scale and color settings for Animate Tool handles

andeon:
This PR adds new customization options for the Animate Tool handles in the Preferences.

https://github.com/user-attachments/assets/d81799ac-c7a3-4400-8ac7-bd60c852f821

- Handle Color: Allows changing the color of the handles for better visibility against different background colors.
- Handle Scale (%): Allows users to scale the transformation handles from 100% to 600% (useful for high-DPI screens or tablet users).

Hey @DarrenTAnims, could you check if the performance feels okay? Technically, scaling the handles is just a basic coordinate multiplication (pos×scale) and changing their color shouldn't impact FPS at all. I just want to be 100% sure there's no lag when dragging or interacting with the viewer on different hardware.

Closes: https://github.com/opentoonz/opentoonz/issues/2852

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -363,6 +363,12 @@ public:
   int getTempToolSwitchTimer() const {
     return getIntValue(tempToolSwitchTimer);
   }
+  double getAnimateToolHandleSize() const {
+    return getDoubleValue(animateToolHandleSize);
+  }
+  TPixel32 getAnimateToolColor() const {
+    return getColorValue(animateToolColor);
+  }
 
   // Xsheet  tab
   QString getXsheetLayoutPreference() const {
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -125,6 +125,8 @@ enum PreferencesItemId {
   useStrokeEndCursor,
   clickTwiceToCreateArcs,
   tempToolSwitchTimer,
+  animateToolHandleSize,
+  animateToolColor,
 
   //----------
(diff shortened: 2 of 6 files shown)


RodneyBaker:
This addition works exactly as advertised.
Thanks @andeon.
As we are now two weeks out of feature freeze this addition is good timing.

DarrenTAnims:
Hi @andeon ,

Sorry for the delay. I've been so busy lately.
I've just downloaded the artifact and it works brilliantly and exactly as I'd expect. Thank you so much for adding this. I've wanted these changes for a very long time. I think they'll be very useful for everyone.

However, since playing with it for a short while, I've a couple of suggestions (for you to ignore if you wish, obviously :-) )

1. The highlight colour isn't configurable. So if you have a green object, it can be invisible when you hover over one of the nodes in the All mode. It highlights the node and text in green.

2. (I know it's beyond the fix you've added) The lines are all drawn 1 pixel wide, no matter how large the percentage. It might look better to increase the thickness slightly as it is increased in size.

Thanks again for adding those changes, buddy.

andeon:
@DarrenTAnims Due to hardware-level limitations in legacy OpenGL, consistent thick lines cannot be reliably achieved across all systems. A robust solution would require rewriting the rendering pipeline using modern shaders or texture-based approaches, which is beyond the scope of these modifications. I attempted to simulate outlines and backgrounds, but the results were inconsistent and visually unappealing. This should be re-evaluated once the migration to Qt6 begins to find a better approach.

@RodneyBaker, if no one finds any further problems, the PR is done!

RodneyBaker:
LGTM

PR #6642, opened 2026-03-01
Adds the Assistant tool icon to the Column and Layer

JParks123:
Adds the Assistant tool icon to the Column and Layer.


RodneyBaker:
@JParks123 
That little icon does not appear ideal for the purpose of identifying Assistant Layers.
Recommend exploring other options.

There are a number of reasons for this approach being sub-optimal.
One of which is that if we add Icons such as this to one level type then the logical direction is to add icons for all level types.  While that is certainly one way forward I believe we can do better.

If identifying Assistants columns better is the goal my suggestion would be to adjust the color to more appropriately suggest what it contains.  I agree the current color is a little too similar to that of toonz raster.

PR #6652, opened 2026-03-07
Special Modifier Key: Viewer Scrub

manongjohn:
Requested port from Tahoma2D, this PR adds a new viewer shortcut key: `Viewer Scrub`

image

While this shortcut is held, clicking and dragging backwards/forwards on the viewer will perform a `Previous Frame`/`Next Frame` action.


RodneyBaker:
Posting intent to merge but do want to conduct a few more tests.
During initial testing I wasn't able to use some shortcuts as the modifier key but that might be due to some override on my system.

Once I used a generic shortcut key assignment, the modifier key worked as expected.

RodneyBaker:
Pressing forward with this.
Thanks @manongjohn for the new option.
LGTM

PR #6654, opened 2026-03-08
Fix Selection Tool line color

JParks123:
This PR changes the colors of the selection line to Red instead of Black for the Freehand and Polyline modes, the Rectangle mode will be Black with Light bg and white with Dark bg, all modes will become white when Black BG check is active.

### DIFF
--- a/toonz/sources/tnztools/selectiontool.cpp
+++ b/toonz/sources/tnztools/selectiontool.cpp
@@ -1219,27 +1219,9 @@ void SelectionTool::drawFreehandSelection() {
 
 void SelectionTool::drawRectSelection(const TImage *image) {
 
-  TPixel color;
-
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@JParks123 
Initial testing shows everything working as advertised.
My sense is that additional changes will be required in the future but this is a simple and useful change.

Things like the white on black background check can be odd particularly if the linework has been adjusted to be white but that's a trivial observation on my part.

There is a slight chance this PR could be reverted based on it changing defaults that others may rely on in production so just be aware of that.

Posting intent to merge but there is no rush.

RodneyBaker:
Proceeding with this change to the Selection Tool's line color.
Thank you @JParks123 for the very useful update.
LGTM

PR #6655, opened 2026-03-09
Fix crash on pasting external image into camera column

### LINKED ISSUE #5370: Crash when Pasting Image into Opentoonz if Camera Cell is Selected
### Description

If a cell in the Camera column is selected when pasting an image into Opentoonz, the program will crash.

### Steps to Reproduce

1. Copy an image to the clipboard
2. Select a cell in the Camera column
3. Paste

### Expected Behavior

Opentoonz should paste the image into a new Raster level in a new column instead of crashing if a Camera cell is selected.

https://github.com/opentoonz/opentoonz/assets/60277069/1914f747-5d86-408e-83b2-dabb632f27cc

### OpenToonz Version

Nightly (Specify below)

built 2024-03-16

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Wacom

andeon:
This PR prevents a segmentation fault that occurred when pasting an external image (e.g., from a web browser) into the camera column by blocking paste operations in that column.

I considered implementing the same confirmation dialog used for other columns (which asks whether to paste into a **New Raster Level** or **Cancel**), but this would require additional testing since other paste scenarios would also need to be handled. Because it could also be triggered unintentionally, I opted for this simpler approach to provide a more immediate fix and prevent the crash, which could otherwise cause users to lose their work.

Closes #5370

### DIFF
--- a/toonz/sources/toonz/cellselection.cpp
+++ b/toonz/sources/toonz/cellselection.cpp
@@ -2076,8 +2076,12 @@ void TCellSelection::pasteCells() {
   const RasterImageData *rasterImageData =
       dynamic_cast<const RasterImageData *>(mimeData);
   if (rasterImageData || clipImage.height() > 0) {
-    if (isEmpty())  // Nothing selected.
+    if (isEmpty()) return;
+    // Prevent pasting raster images into the camera column (c0 < 0)
+    if (c0 < 0) {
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Simple fix executed perfectly.
Thanks @andeon!
LGTM

PR #6657, opened 2026-03-10
Fix Close Gap behavior in quickput.cpp

andeon:
This PR fixes a regression from PR #6633, reported by @RodneyBaker, that broke the Fill Tool’s Close Gap feature. Gaps in shapes should now be fillable normally.


RodneyBaker:
I confirm that resolves the problem.
Thank you @andeon for the quick turn around.
LGTM

Issue #3958, opened 2021-05-24
Imported jpg image displayed in wrong position in standard camera view, but renders correctly.


MJA66:
Hi,

I want to install a jpg file as a background for my animation.
When I drag my file onto the canvas it properly creates a new column with the drawing showing in the level menu. When I double click on the drawing however, it is placed outside its binding-box. 

Screenshot 2021-05-21 at 19.54.00.png

When I select the binding box. It properly shows the picture inside it.
Screenshot 2021-05-21 at 19.57.21.png

When I use the camera stand view, the bounding box is put properly in place, but the drawing is completely outside of the camera view. 
Screenshot 2021-05-21 at 20.00.20.png

In 3D view it shows as I want and expect it. 
Screenshot 2021-05-21 at 20.02.06.png

And it also renders as expected.
Screenshot 2021-05-21 at 20.03.04.png

The problem that the jpg shows wrong in drawing mode and in camera stand view which makes it impossible to use. 
I have no clue what is going wrong, or what I am doing wrong. I have tried different jpg files, and png, and ...

This makes the tool unusable for me, because I want to use the background as a reference in animating mathematical formulas. 

Any help appreciated,
MArk.
![Screenshot 2021-05-21 at 19 54 00](https://user-images.githubusercontent.com/6573741/119390161-f124c080-bccc-11eb-8ede-e1608c84138a.png)
![Screenshot 2021-05-21 at 19 57 21](https://user-images.githubusercontent.com/6573741/119390167-f3871a80-bccc-11eb-8897-6150932e7b4d.png)
![Screenshot 2021-05-21 at 20 00 20](https://user-images.githubusercontent.com/6573741/119390170-f4b84780-bccc-11eb-94cd-e7e5c569bc2d.png)
![Screenshot 2021-05-21 at 20 02 06](https://user-images.githubusercontent.com/6573741/119390175-f5e97480-bccc-11eb-87be-4efd0ce6bf76.png)
<img width="827" alt="Screenshot 2021-05-21 at 20 03 04" src="https://user-images.githubusercontent.com/6573741/119390176-f6820b00-bccc-11eb-92de-4a749a8353a8.png">

RodneyBaker:
Very odd.
Can you supply the project file that exhibits this problem?

Also, if you start a new project/scene from scratch and use a different image do you get the same result?

[...]

PR #6661, opened 2026-03-12
Fix crash when reading empty and corrupted JPEG files

andeon:
This PR updates `tiio_jpg.cpp` to improve handling of empty or corrupted JPEG files by generating a checkerboard pattern for missing image data.

- If a JPEG file is empty (`nullptr`) or completely unreadable (e.g., missing or corrupted header), default dimensions (1920x1080) are applied. Actual display size may vary depending on DPI or other internal metrics.
- Generates a checkerboard pattern for missing image data so OpenToonz can safely display a placeholder.
- Ensures that opening, converting, and browsing empty or corrupted JPEGs does not crash the application.

Two sample files are included to reproduce the issue in the current stable version or to check the fix: [corrupted_jpeg_samples.zip](https://github.com/user-attachments/files/25923911/corrupted_jpeg_samples.zip)

Note: The `#ifdef CICCIO` sections appear to be leftover platform-specific/legacy handlers from the original Toonz development. Consider removing them in a future cleanup.

### DIFF
--- a/toonz/sources/common/tiio/tiio_jpg.cpp
+++ b/toonz/sources/common/tiio/tiio_jpg.cpp
@@ -3,10 +3,10 @@
 #ifdef _MSC_VER
 #pragma warning(disable : 4996)
 #endif
-//#include "texception.h"
-//#include "tfilepath.h"
-//#include "tiio_jpg.h"
-//#include "../compatibility/tnz4.h"
+// #include "texception.h"
+// #include "tfilepath.h"
+// #include "tiio_jpg.h"
+// #include "../compatibility/tnz4.h"
 
 #include "tiio_jpg.h"
 #include "tiio_jpg_exif.h"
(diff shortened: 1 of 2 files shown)


RodneyBaker:
I'm rerunning the MacOS check although I don't think it will pass.
This mostly just to satisify my curiousity.

RodneyBaker:
The only reason this fix has not moved forward at this time is due to checks not passing.

andeon:
The macOS (Mojave) sometimes it works and sometimes it doesn’t start. 

The other macOS workflow is outdated. It only run on merge or if I send an amend commit, because it’s using a deprecated workflow from before the [fix](https://github.com/opentoonz/opentoonz/pull/6662) so the re-run not work.

RodneyBaker:
Will press forward with this as an essential fix.
Let's look for follow up fixes related to JPG as this isn't guaranteed to address all issues reported regarding JPEG failure from the past.

RodneyBaker:
The only report I saw open related to JPEG issues was #3958
Other reports may have already been addressed or were mentioned in comments of other reports.

There does appear to be more reporting of JPEG issues with MacOS although I've really no basis for believing the problems are isolated to Mac systems other than Mac users have reported most of those problems.

JPEG has had some interesting updates of late and we should investigate how we can best move forward to support modern JPEG standards.  This of course without losing legacy support.  

Should I mention the JPEG AI International Standard?
Probably not.  ;)

All this to say, we could use a developer whose primary focus is just on JPEG ecosystem and how it can best interface with Opentoonz.  

By the time we get to v3.0 it will be a very different world for animation software.
Establishing a trajectory for JPEG support is a reasonable goal in the near term.
Implementation, not so much.  Especially as the JPEG format remains a moving target.

REF:  https://jpeg.org/jpeg2000/

RodneyBaker:
Something slipped past us @andeon 

REF:  Crash when rendering to and loading JPG and WEBP files reported in #6679

PR #6662, opened 2026-03-12
ci: Fix macOS Qt6 symlink conflicts from OpenCV/VTK

andeon:
This PR fixes macOS CI failures caused by Homebrew’s opencv update (4.13.0), which now depends on vtk. That automatically installs Qt 6.10 modules, creating symlink conflicts when installing qt@5.

- Unlink all Qt6 formulas (`brew list | grep '^qt' | grep -v '@5' | xargs -n1 brew unlink`)
- Install qt@5 cleanly and force link

These changes restore the macOS build and future-proof it against new Qt6 modules.

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -44,7 +44,7 @@ jobs:
             fi
           done
           brew link --overwrite --force python@3.14 || true
-          
+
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thanks @andeon

PR #6670, opened 2026-03-15
Fix PNG Colormap serialization breaking scene loading

andeon:
This PR fixes a regression from PR #6595, reported by @Gaspard--, where the `Colormap` property in PNG format settings was serialized into scene XML files. This could lead to invalid pointer values, preventing scenes from loading correctly.

Specifically, saving a scene with PNG output would write the following line to the scene XML:

`<property name="Colormap" type="pointer" value="0000000000000000"/>`

OpenToonz cannot interpret this line, causing the scene to fail to load.

After this fix, new scenes will save correctly without writing the Colormap property, and scenes will load normally. Existing `.tnz` files containing the line still require manual cleanup.

### DIFF
--- a/toonz/sources/image/png/tiio_png.cpp
+++ b/toonz/sources/image/png/tiio_png.cpp
@@ -569,14 +569,13 @@ private:
 //=========================================================
 
 Tiio::PngWriterProperties::PngWriterProperties()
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Pressing forward with this fix to get it into a nightly to prevent corruption of scene files.
LGTM

PR #6671, opened 2026-03-15
 Add TLV Savebox checkbox to Fill tool options bar

JParks123:
This PR will add the TLV Savebox checkbox to the Fill tool options bar, allowing users to turn it off and on as they please, instead of having to find it hidden in preferences.

### DIFF
--- a/toonz/sources/tnztools/filltool.cpp
+++ b/toonz/sources/tnztools/filltool.cpp
@@ -1986,7 +1986,7 @@ FillTool::FillTool(int targetType)
     , m_autopaintLines("Autopaint Lines", true)
     , m_referFill("Refer Fill", false)
     , m_extendFill("Extend Fill", true)
-    , m_savebox("Savebox", Preferences::instance()->getFillOnlySavebox())
+    , m_savebox("Savebox", false)
 {
   m_areaFillTool       = new AreaFillTool(this);
   m_normalLineFillTool = new NormalLineFillTool(this);
@@ -2436,7 +2436,9 @@ void FillTool::resetMulti(bool resetAreaFiller) {
 bool FillTool::onPropertyChanged(std::string propertyName, bool addToUndo) {
 
   bool rectPropChangedflag = false;
+
   if (propertyName == m_savebox.getName()) {
+  Preferences::instance()->setValue(FillOnlysavebox, m_savebox.getValue());
   notifyImageChanged();
 }
 


RodneyBaker:
I don't see anything in the code that should cause this but I get an error with both Appveyor and Github Action builds with this PR:

image

JParks123:
> I don't see anything in the code that should cause this but I get an error with both Appveyor and Github Action builds with this PR:
> 
> image

I tracked it down. The issue was caused by accessing Preferences in the FillTool constructor when initializing the Savebox property. That triggered Preferences initialization too early during startup. I moved the preference read into FillTool::onActivate() instead so the tool loads the saved value when it becomes active. The constructor now just uses a default value. After this change the menubar template warning no longer appears. It works good now. : D

RodneyBaker:
That has fixed the program loading problem.  Thanks!

This is a useful addition but I think I'm seeing a problem with the logic of the Checkbox (in Fill Tool) versus the checkbox in Preferences.

Namely, there are times when having the option on in Preferences doesn't work and there are times when the Savebox option is off in the Fill Tool where the Savebox still appears to be used.  In one of those cases it's just a matter of filling an area which just happens to have been previously filled before with the Savebox option on... all good there.  That makes it seem like the Savebox is still being used when it actually isn't.  A bit awkward but reasonable.

I guess a question I now have is:  When is the Preferences option to use the Savebox for Filling still being used when the Savebox option in Fill Tool isn't in sync with toggling the option on/off in Preferences?  This area of logic may need @Abel032's eyes to make sure this option isn't turning off something that otherwise should be working because the Savebox option is on in Preferences but not in the Fill Tool.

Hope that makes some sense.
I do see the usefulness of having the Savebox option in Fill Tool.  I guess it's just not clear to me why toggling one doesn't automatically toggle the other.

Abel032:
Suggest not to keep both options existing 🤔
I think if we add savebox as a fill tool option,we need to remove it from preference, and use it as fill tool param in the code but not a preference value.

konero:
In which scenarios does this need to be toggled on and off so frequently that it should be added to the tool options bar?

JParks123:
> In which scenarios does this need to be toggled on and off so frequently that it should be added to the tool options bar?

It has many ways to be used but here is one scenario.  
image

RodneyBaker:
Having the option to turn on the Savebox is certainly nice but having that option but not being able to adjust the Savebox without stopping and going to the Selection Tool is awkward.

Are there also plans to have the Modify Savebox option included in Fill Tool?

JParks123:
> Having the option to turn on the Savebox is certainly nice but having that option but not being able to adjust the Savebox without stopping and going to the Selection Tool is awkward.
> 
> Are there also plans to have the Modify Savebox option included in Fill Tool?

No plans to do that yet, for now it will remain in the Selection tool options.

• Adds a Savebox checkbox to the Fill Tool options bar
• Mirrors existing "Fill Only Savebox" preference
• Keeps preference and tool option synchronized
• No behavioral change to existing fill logic

konero:
@JParks123 Can you please build changes locally on your machine, and properly stage commits when they're ready for review? This PR is triggering a lot of CI jobs.

Aside from that, the proposed changes make no sense, seem to remove unrelated code and duplicated other parts. I can't review due to constant changes, but one of the changes made I believe would have broken custom menu templates.

RodneyBaker:
You are abusing the process @JParks123 
Please, do not do that.

andeon:
@JParks123 We’ve already discussed the issue of flooding the CI with changes. PRs are reviewed by others, and this is not a sandbox environment. 

Please only submit a PR to the main repository when you believe it is ready. Continuing this behavior may result in you being blocked from the repository, as it appears to be intentional despite multiple explanations and warnings.

JParks123:
> @JParks123 We’ve already discussed the issue of flooding the CI with changes. PRs are reviewed by others, and this is not a sandbox environment.
> 
> Please only submit a PR to the main repository when you believe it is ready. Continuing this behavior may result in you being blocked from the repository, as it appears to be intentional despite multiple explanations and warnings.

Sorry, Its done now  everything works well, it doesn’t break anything and my approach is the correct minimal integration.

JParks123:
> > @JParks123 We’ve already discussed the issue of flooding the CI with changes. PRs are reviewed by others, and this is not a sandbox environment.
> > Please only submit a PR to the main repository when you believe it is ready. Continuing this behavior may result in you being blocked from the repository, as it appears to be intentional despite multiple explanations and warnings.
> 
> Sorry, Its done now everything works well, it doesn’t break anything and my approach is the correct minimal integration.

I decided to close this PR maybe someone will take it on as an FR in the future.

PR #6672, opened 2026-03-16
Feat assistant level auto switch

Ahl76:
This PR adds the following features :

Auto-Switch & keep : Automatically switches to the Assistant Tool when creating an Assistant Level, and returns to the previous tool when switching back to the previous level.

image

When this mode is enabled, the last chosen assistant will be kept in memory as a base each time you return to the assistant level. This will speed up the workflow for adding the same assistant consecutively. Users can always return to the old behavior by disabling it, if they prefer to avoid adding assistants inadvertently.

Assistant Shortcuts: All assistant types are now available in Configure Shortcuts and a shortcut to cycle through them has also been added

image

and compatible with the Custom Panel.

image

New Assistant Level: Available like other level commands, with reuse of its icon. Also available in xsheet tool bar.

New assistant level


RodneyBaker:
Pressing forward with this PR as I don't see anything in the code that suggests adverse impact on current processes.  If something is found that cannot be remedied easily we'll revert.

Thank you @Ahl76 for you contribution and your patience.

PR #6673, opened 2026-03-17
fix: Resolve Ink Check conflicts

andeon:
This PR addresses color conflicts and rendering glitches between Ink Check and Ink1 Check that surfaced following PR #6633 (reported by @JParks123 and @RodneyBaker). 

- Unified Ink Logic: Pre-calculates a single inkColor per pixel, ensuring solid lines and anti-aliased borders match perfectly, eliminating "glow" artifacts and satisfying Gap Fill requirements.

- Color Neutralization: Added a pre-loop settings cleanup to restore original palette colors when diagnostic checks are disabled. This prevents accidental "color bleeding".

- Style#1 Priority: Implemented hybrid logic where inkCheck takes priority if Style 1 is the active selection, otherwise, the dedicated ink1Check is preserved independently.

### DIFF
--- a/toonz/sources/common/trop/quickput.cpp
+++ b/toonz/sources/common/trop/quickput.cpp
@@ -4027,20 +4027,18 @@ void doQuickPutCmapped(const TRaster32P &dn, const TRasterCM32P &up,
       (aff * TRectD(-0.5, -0.5, up->getLx() - 0.5, up->getLy() - 0.5));
   if (boundingBoxD.x0 >= boundingBoxD.x1 || boundingBoxD.y0 >= boundingBoxD.y1)
     return;
-  int yMin = std::max(tfloor(boundingBoxD.y0), 0);
-  int yMax = std::min(tceil(boundingBoxD.y1), dn->getLy() - 1);
-  int xMin = std::max(tfloor(boundingBoxD.x0), 0);
-  int xMax = std::min(tceil(boundingBoxD.x1), dn->getLx() - 1);
-
+  int yMin       = std::max(tfloor(boundingBoxD.y0), 0);
+  int yMax       = std::min(tceil(boundingBoxD.y1), dn->getLy() - 1);
+  int xMin       = std::max(tfloor(boundingBoxD.x0), 0);
+  int xMax       = std::min(tceil(boundingBoxD.x1), dn->getLx() - 1);
   TAffine invAff = inv(aff);
   int deltaXL    = tround(invAff.a11 * (1 << PADN));
   int deltaYL    = tround(invAff.a21 * (1 << PADN));
   if ((deltaXL == 0) && (deltaYL == 0)) return;
-  int lxPred = up->getLx() * (1 << PADN) - 1;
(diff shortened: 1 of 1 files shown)


andeon:
@RodneyBaker I'm a bit confused about the schedule. I see that features were frozen on March 16, but I think the fixes should come in 1.8, because it doesn't make sense to deliver them only in the next release (2.0) since they are already available.

I also think they should be included before the first RC so we can take advantage of the wave of downloads and usage to test more thoroughly and eliminate any remaining bugs.

RodneyBaker:
@andeon Thanks for the opportunity to clarify (assuming I can).

@shun-iwasawa may be adjusting the release number and that might be the source of confusion.

The tags for 2.0 are for the next release although it isn't quite clear if Shun has given up on having the next release be v1.8 or v2.0.  He has stated the goal would be to move to Qt6 and Apple Silicone which would make it a v2.0 release but in latest posting mentioned a v1.8.  A typo?  An update to the release numbering?  Not sure.  My understanding was/is that if we can't get to Qt6/Apple Silicone the next release will be v1.8.

All this to say, FIXES are destined for the next release.
This doesn't guarantee they will be in the next release as not all fixes are created equal.

I am largely stepping back from merging PRs because it is important that whatever goes forward is actually a fix and not something that breaks things at the last minute.  Having said this, I plan to merge any obvious fix as soon as it can be merged.

All this to say that if the confusion is due to PR labels I can drop those.
Both v1.8 and v2.0 currently refers to the next release until I'm advised otherwise.
If you see a v2.1 please know that it definitely means it will not be in the next release.

Clear as mud?  
Hopefully that clears things up a little pending more word from @shun-iwasawa .

RodneyBaker:
I should have added:

Some PRs have not been merged yet... such as this one... because I really do not like to merge PRs that have not passed all checks.  The Mac Mojave check we have no control over so I reserve the right to merge those but have been holding off in hope that check will be updated and PRs will pass all checks before merging.

So... we are fully focused on bugfixes, translations and testing with the current schedule:

> March16: Feature Freeze
> Through mid-April: Bug fixes, translation files updates
> April 15: Release of V1.8 RC
> Early May: Release of V1.8 (or V1.8 RC2, if necessary)

RodneyBaker:
Will press forward with this fix as soon as possible although noting the Mac Mojave check is a continuing failure that needs to be addressed.

konero:
> The tags for 2.0 are for the next release although it isn't quite clear if Shun has given up on having the next release be v1.8 or v2.0. He has stated the goal would be to move to Qt6 and Apple Silicone which would make it a v2.0 release but in latest posting mentioned a v1.8. A typo? An update to the release numbering? Not sure. My understanding was/is that if we can't get to Qt6/Apple Silicone the next release will be v1.8.

From his statement he clearly seems to suggest he'll be releasing 1.8 next, then 2.0 in the future in the hope the Qt6 work goes smoothly. If not, likely we'll get a 1.9 release. Realistically, I believe we're going to get a 1.9 and 2.0 branch, with both available to download, so 2.0 can be stress tested for awhile. Krita is 2 years in and still unstable for example.

PR #6677, opened 2026-03-19
[fix] Restore TPanel Border Pixmaps

konero:
Restores the ability to use border images for titlebars in stylesheets, useful for seeing what panel is active.


RodneyBaker:
LGTM

Issue #6679, opened 2026-03-19
[bug] Cannot Import Images (.webp, .jpg, etc.)


konero:
### What's the bug?

In the most recent build, we're unable to import images like `.webp`, `.jpg`. At the moment, only `.png` seems to work. We tried random images, and images downloaded from a browser.

`.webp` will give us a popup message saying "Failed to delete file".
`.jpg` will hard-crash.

If anyone else can test/confirm.

### Steps to Reproduce

1. Drag a `.webp` or `.jpg` file into the Viewer

### Screenshots, Videos or Crash Logs

`.webp`:
<img width="749" height="139" alt="Image" src="https://github.com/user-attachments/assets/c98282f2-3f86-4853-a079-a25151aea752" />

### OpenToonz Version

Other (specify below)

### Exact Version Info (optional)

Most recent build as of 19th Mar 2026

### Operating System

Windows

### GPU Brand

NVIDIA

### Graphics Tablet

Wacom

RodneyBaker:
I confirm the problem.  Thanks @konero

andeon:
@konero I have submitted the fix for the JPG module (I finally started to understand why there is so much 'rage' in the legacy code Italian comments). Regarding the WebP issue, I've observed that it works as expected on Linux (and likely macOS) but fails on Windows. I suspect a regression in path handling, specifically regarding how FFmpeg interprets absolute paths. It seems to have broken around mid-January or February. Since my primary development environment is Linux, it may take me some time to fix, as I need build for a Windows versions to test.

andeon:
@konero Was Webp image static or animated? I noticed that animated images don't seem to work. I even tried testing this in the stable version of OpenToonz and it didn't work there either.

Could you confirm this or point me toward the specific images so I can run some tests myself?

RodneyBaker:
As a side note to this report, webp hasn't been a file format Opentoonz has been able to load or render to without conversion.   Perhaps you are mistaking the .webp format for .webm?

konero:
@andeon Just a static image, I tried with ones downloaded from a browser, seems to happen with any. The image appears to load into the viewer and I see no issues with the content, but we just get that popup dialog for some reason.

andeon:
Thanks @konero, I was able to reproduce it. I tested by renaming the .webp file to something with a lot of digits and letters (e.g., 16r4feererwr4c14e.webp) and got the same message. The image is imported and displayed, but the popup still appears.

[...]

PR #2322, opened 2018-10-10
Save and Load DPI Information of JPEG Format

shun-iwasawa:
This PR enables save & load dpi information of JPEG format.

For loading, dpi will be loaded either from JFIF or EXIF header.
For saving, OT will write dpi information always in JFIF header for now.

Regarding the reading function of EXIF header, I referred to source code of a public domain software "[Exif Jpeg header manipulation tool (jhead)](http://www.sentex.net/~mwandel/jhead/)" by Matthias Wandel.

PR #6680, opened 2026-03-19
Fix JPG Reader crash on Windows

andeon:
This PR refactors the `JpgReader` to use the native `setjmp/longjmp` mechanism for error handling. Previously, the JPEG implementation used C++ exceptions (`throw`) inside `libjpeg` callbacks. While this approach may work on Unix-based systems due to different ABI handling. On Windows, throwing exceptions across C library boundaries (where the C library does not know how to handle C++ stack unwinding) leads to EXCEPTION_ACCESS_VIOLATION. 

This is the root cause of the crash reported by @konero and confirmed by @RodneyBaker in https://github.com/opentoonz/opentoonz/issues/6679. Thanks!

### DIFF
--- a/toonz/sources/common/tiio/tiio_jpg.cpp
+++ b/toonz/sources/common/tiio/tiio_jpg.cpp
@@ -29,6 +29,20 @@
 #include <stdexcept>  // for std::runtime_error
 
 //=========================================================
+// Jpg Error Management
+//=========================================================
+
+extern "C" {
+static void tnz_error_exit(j_common_ptr cinfo) {
+  // Cast the error manager to our custom struct to access the jump buffer.
+  // This must match the struct definition used in the class.
+  Tiio::JpgReader::tnz_error_mgr *myerr =
+      (Tiio::JpgReader::tnz_error_mgr *)cinfo->err;
+  longjmp(myerr->setjmp_buffer, 1);
+}
+}
+
+//=========================================================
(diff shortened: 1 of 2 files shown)


konero:
Confirmed this fixes the issue with `.jpg`, cheers. 🍺

RodneyBaker:
Something unrelated to this fix that stands out as potentially needing to be addressed is the following: 

`exifReader.process_EXIF(mark->data - 2, mark->data_length);`

This suggests to me that because libjpeg is already supplying the ```mark->data``` there is risk of reading before the buffer's start point.

Should this accept what is passed by Libjpeg?

If there is a special reason to move backward it would be good to comment there in the code why that is preferred.

But... all this to say that this fix does appear to resolve the JPG reader crash.

Will delay mergering for a moment but do want to get the fix into nightlies to keep those who download the last build from running into the problem.

RodneyBaker:

andeon:
@RodneyBaker, this logic was transplanted from the jhead project (as seen in PR #2322) to handle EXIF metadata. The `mark->data - 2` offset exists because jhead expects the buffer to include the 2-byte JPEG marker size, while libjpeg provides a pointer that starts immediately after those bytes. It's a legacy behavior, and we could probably consider refactoring this to accept the pointer directly from libjpeg and avoid pointer arithmetic that goes behind the buffer's start point

PR #6682, opened 2026-03-21
Keep link to Studio Palette when creating new default palettes

Abel032:
Allows to keep the link to Studio Palette when creating new default palettes.

图片


Abel032:
Keep the link to Studio Palette would benefit if you copy & paste between toonz raster levels share styles from one same studio palette -- won't generate new styles.

RodneyBaker:
Pressing forward with this.

PR #6683, opened 2026-03-21
Fix selection lost of export level popup

Abel032:
This PR fixes:

1. Can't export selected levels(columns/cells/cast) when `Export Level..` excuted first time.
2. Selected levels of `Export Level Popup` not update when xsheet column selection changed.

But I didn't fully understand how selection system of OT works, 
especially this function: `void DvDirTreeView::currentChanged()` in `dvdirtreeview.cpp`.


RodneyBaker:

PR #6685, opened 2026-03-21
Fix long path issue in Ffmpeg by using hash

andeon:
This PR addresses an issue related to long file paths on Windows, especially considering the duplicated cache path and how temporary files were previously named. The issue was reported by @konero in #6679, thanks!

This change introduces a 6-character MD5 hash of the normalized file path as a prefix for all temporary PNG files. It also unifies the naming pattern to _in%06d.png across import, preview, and export flows.

This approach keeps file paths shorter, avoids the Windows 260-character limit, and ensures proper cleanup of temporary files. It is more robust than the previous approach, as the hash provides a short, consistent, and sufficiently unique identifier for each file.

@konero, could you please check if this fixes the issue on your side, or if there’s anything I might have missed?

Closes:#6679

### DIFF
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -17,6 +17,7 @@
 #include <QDir>
 #include <QFile>
 #include <QDateTime>
+#include <QCryptographicHash>  // For MD5 hash
 #include "tmsgcore.h"
 #include "thirdparty.h"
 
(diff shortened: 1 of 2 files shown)


konero:
I gave it a test, it seems I no longer see the popup message about failing to delete file. 👍

RodneyBaker:
Given what I'm seeing and @konero's blessing... pressing forward with this fix.
It's a very interesting and from what I can tell very useful fix.
Let's see how it works in the wild.
Translation:  LGTM

RodneyBaker:
I am crashing when rendering in latest nightly (based on this PR) but need to double check to make sure proper ffmpeg is being used etc.

Just want to give a heads up in case others experience that.

andeon:
@RodneyBaker I tested it several times here and it didn’t crash. Could you tell me if it’s crashing with the specific exported format? I usually test with the OpenToonz sample, but I also tried with other projects. I haven’t updated FFmpeg on Windows yet. I’m going to download the latest version just to check, since it’s probably the version the user will have.

RodneyBaker:
I changed FFMPEG versions and haven't had an issue since.
Not quite sure which version this working FFMPEG is as I have several I tested some time ago.
Will test some more.

The crashes were happening with MP4 and gif. (a hint that ffmpeg was at play)
No issues with PNG.

andeon:
@RodneyBaker I tested it with the[ latest auto-build](https://github.com/BtbN/FFmpeg-Builds/releases) (2026-03-25 13:12), `ffmpeg-master-latest-win64-lgpl.zip`, and also tested the latest build `ffmpeg-8.1-full_build.7z`  from the [Gyan CodexFFmpeg releases (mirror)](https://github.com/GyanD/codexffmpeg/releases), and it works. But I didn’t remember there being so many Windows builds like this which one are we recommending to users?

RodneyBaker:
There is no official recommendation short of suggesting to use the latest release.
The Essentials being the one with the licensing most would prefer.
The Full release for those more likely to use FFMPEG via the command line (in addition to using it with Opentoonz.

This would be the main place I might send those who ask:  https://www.gyan.dev/ffmpeg/builds/#release-builds

The Essentials in .zip format being the most accessible as .7z (7zip) isn't recognized on as many systems as .zip.

I had not realized that just 10 days ago they had a new release of ffmpeg.
That release appears to work fine and as such should be the recommended version as far as I'm concerned.

Aside:  I have an ffmpeg script to improve the size of gif animation (with and without transparency that will reduce the size of a gif by half or more.  After the next release of Opentoonz we should consider implementing that if we don't go all the way and expose ffmpeg scripting so users can override the hardcoded settings.

konero:
No issues on my end with crashes so far. I use ffmpeg full.

Issue #6690, opened 2026-03-23
[bug] Can't resize Style Editor/Palette when docked together vertically (intermittent)


konero:
### What's the bug?

From the Drawing room, I'm unable to adjust the width of these two panels by clicking and dragging on the dock border line.

_**TBD**: Is it only the Drawing room, or always when docked in this formation?_

### Steps to Reproduce

1. Go to the Drawing room
2. Try to resize the docked Style Editor/Palette sidebar by clicking and dragging on the dark dock border line

### OpenToonz Version

Other (specify below)

### Exact Version Info (optional)

Latest build as of 23rd March 2026

### Operating System

Windows

### GPU Brand

NVIDIA

### Graphics Tablet

Wacom

konero:
I will guess Auto Adjust Panel Width option found in the Palette window menu might be the source of the bug.

<img width="273" height="182" alt="Image" src="https://github.com/user-attachments/assets/4d929264-66ae-4b75-97e2-b0653d32d1aa" />

andeon:
I'm confirming this issue, mainly in the **Style Editor**. I'll investigate it and try to fix it. When the panel **Pallete** is undocked, it can’t be resized, but when the **Style Editor** is undocked it works correctly.

andeon:
@konero it has a strange behavior. I undocked the **Style Editor** and noticed I couldn’t resize it in floating mode, and double-clicking the title to auto-resize didn’t work either. Then I opened another one and saw that it worked. I re-docked the one that was resizable and it started working again, so I saved it. After testing for a while, closing OpenToonz and reopening it with different scenes. I noticed the same behavior again, so something is changing its state.

I tried another approach: when we close OpenToonz, it reopens in the last room used. In this case, I was using the Timeline Room. It didn’t work there (same issue), but in the Drawing Room it was working. If I switch back and forth, the behavior is consistent. So apparently the state changes when it’s opened, or possibly when it’s closed.

In the last test, I closed it in the Browser Room. When reopening, I switched to the Drawing and Timeline rooms, and everything was working.

konero:
@andeon Thanks for looking into it.

I wanted to note I've stumbled on another related bug, I couldn't resize the Level Strip. This was in the Timeline room, and I had docked an X-Sheet next to it. The app had been restarted a few times since then. It got shrunk to a point where it was too small to see, but it couldn't be adjusted. At the same time, the Style Editor/Palette issue happens.

[...]

PR #6694, opened 2026-03-26
fix: Intermittent UI lock and docking resize in Style Editor and Palette

andeon:
This PR attempts to address a issue where certain UI panels (specifically the Style Editor and Filmstrip) could become permanently locked and unresizable on Windows. Reported by @konero in #6690

The DockLayout::redistribute() logic temporarily sets widgets to a Fixed Size to protect their proportions during layout recalculations. Apparently, the "unlock" phase (which restores the original min/max sizes) was being skipped in some cases, leaving the widgets stuck. 

Closes: #6690

### DIFF
--- a/toonz/sources/toonzqt/docklayout.cpp
+++ b/toonz/sources/toonzqt/docklayout.cpp
@@ -11,59 +11,48 @@
 #include <QDesktopWidget>
 
 //========================================================
-
 // TO DO:
-//  * Usa la macro QWIDGETSIZE_MAX per max grandezza settabile per una widget
-//      => Dopodiche', basterebbe troncare tutte le somme che eccedono quel
-//      valore... Comunque...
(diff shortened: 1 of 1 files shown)


konero:
So far so good, I've been able to resize all the panels and not ran into the bug. 👍 

Not sure if this will take more time to be sure, if someone else who experienced the issue can also confirm, since it was an intermittent issue, but I can't reproduce it while trying at least.

**Update [31.03]**: Still not ran into any issues on my end, I've been able to adjust the size anytime.

andeon:
@konero, thanks for testing! I'm also no longer able to reproduce the issue with the fix applied. 

I believe this is ready to be merged to resolve this annoying behavior.

RodneyBaker:
Thank you for the fix @andeon 
LGTM

PR #6703, opened 2026-03-31
fix TZL icon buffer overflow and savebox exception

andeon:
This PR addresses a critical crash occurring when loading certain (TZL) files. The issue was caused by a typographical error in the icon buffer validation logic and a fatal exception triggered by mismatched savebox dimensions.

- Corrected the icon buffer size check in` tiio_tzl.cpp`. The code was incorrectly using `iconLx * iconLx` (width squared) instead of `iconLx * iconLy` (width * height). This led to false-positive corruption errors and crashes when handling non-square icons.

- Improved stability by replacing a fatal TException with a non-breaking error log when the image savebox exceeds raster boundaries. This ensures the level can still be loaded even if the thumbnail data is malformed.

- Translated legacy Italian and Japanese comments to English to maintain consistency with the official codebase.

Test: The crash was identified using legacy Toonz tutorial materials. While I'm uncertain if these specific files were corrupted locally or were originally malformed, a copy from the original site is attached below (Since the original hosting site is no longer available):

[Toonz_TW_Tutorial_Material.zip](https://github.com/user-attachments/files/26387162/Toonz_TW_Tutorial_Material.zip)

### DIFF
--- a/toonz/sources/common/tcache/timagecache.cpp
+++ b/toonz/sources/common/tcache/timagecache.cpp
@@ -525,7 +525,12 @@ CompressedOnDiskCacheItem::CompressedOnDiskCacheItem(
 
 CompressedOnDiskCacheItem::~CompressedOnDiskCacheItem() {
   delete m_imageInfo;
-  TSystem::deleteFile(m_fp);
+  if (m_fp.isEmpty()) return;
+  try {
+    TSystem::deleteFile(m_fp);
+  } catch (...) {
+    // Ignore any error do not let it escape the destructor.
+  }
 }
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


JParks123:
maybe you can add the Savebox on/off checkbox to the Fill tool options bar or even a shortcut for it, since I closed my PR.

RodneyBaker:
@andeon 
The following is submitted for your consideration:

[AI_Assisted_Review_PR6703_TZL_Icon_Loading_Fix.pdf](https://github.com/user-attachments/files/30672696/AI_Assisted_Review_PR6703_TZL_Icon_Loading_Fix.pdf)

andeon:
@RodneyBaker  I've been pretty tied up lately. I still need to set up my new machines before I can contribute again. I'm not sure if I'll get to it this month, maybe next month. But I think this code just needs a small change.

I thought we'd see more bugs in this release, but it looks like there are a lot fewer than I expected.

RodneyBaker:
No rush @andeon 
It's good to know you are still out there.

RodneyBaker:
@andeon This isn't to try to rush this but just to keep it moving forward.
Given that this touch (slightly) on use of Savebox, which had some unrelated changes in recent past and will have more in the future) I created https://github.com/OpenAnimationLibrary/opentoonz-dev/pull/113 in order to do just that.

If those prove to be satisfactory we can pivot to that.

PR #6714, opened 2026-04-04
[git] Move feature request to discussions

konero:
Remove the feature request template and swap to a link that points to open a new post in the ideas section of discussions tab instead. Also removes the questions template, probably a better idea to use discussions for these as well (or just use blank issue).


RodneyBaker:
I believe this is an important change that helps keep the focus of issues reported on bugs, crashes and other critical priorities.  Feature Requests should have adequate discussion and buy in from Opentoonz users and developers and the Discussions area is ideal for that.

Developers are welcome to open issues for their in progress PRs to point to as they see fit although the PR itself will more often than not be adequate for that purpose.  This to allow the tag lines such as 'implements', etc. which helps document the activity.
When in doubt we can also adjust labels as needed.

Importantly, we want developers to be able to look through the issues to quickly and easily see items that need immediate attention.

Thanks @konero for helping us move in the right direction.

PR #6716, opened 2026-04-05
[fix] Add fallback for pane off button color to prevent black buttons

### LINKED ISSUE #6619: Mac: Icons in top right corner (field guide, camera stand, 3d view etc.) are black on black - can't see them
### What's the bug?

Macbook pro M1 pro 32GB. Icons in the upper right corner of the view screen are black icons on black background and impossible to distinguish. On mouseover, and when clicked and activated, they are visible.
This is true in all rooms that includes the viewer. Brighter colour themes do not solve this.

### Steps to Reproduce

1. Run software

Visible icons

Image

### OpenToonz Version

1.7 (nightly build jan 31 2026)

### Operating System

MacOS 15.7.3

### GPU Brand

Apple M1 Pro 32 GB 10 cores

### Graphics Tablet

None

konero:
Fixes #6619.

Since offColor (qproperty) was newly added to the pane buttons in the stylesheets unless the user updates their stylesheets the color comes back as invalid, which is shown as black. This adds a fallback to transparent.

### DIFF
--- a/toonz/sources/toonz/pane.cpp
+++ b/toonz/sources/toonz/pane.cpp
@@ -390,7 +390,9 @@ void TPanelTitleBarButton::setOffColor(const QColor &color) {
   }
 }
 
-QColor TPanelTitleBarButton::getOffColor() const { return m_offColor; }
+QColor TPanelTitleBarButton::getOffColor() const {
+  return m_offColor.isValid() ? m_offColor : Qt::transparent;
+}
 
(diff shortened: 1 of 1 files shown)


konero:
I've tested as many scenarios as I can think users would run into this issue, I believe this resolves it and is good to go.

RodneyBaker:
We'll have to see if anyone runs into outliers that aren't covered.

Thanks @konero for the fix.

PR #6725, opened 2026-04-07
[fix] Fix neutral theme x-sheet colors

konero:
Primarily this fixes Neutral theme's X-Sheet colors. Somewhere they became very dull and veered away from the original colors. I can't find what commit this started from, but I remember we should try maintain the original colors and this should correct it. It probably was due to when I added a 'ui-contrast' global variable, and it was too close to the Default theme's value. Overall for Neutral's changes, I've referenced the original V1 theme, I hear a lot of complaints about the current one, it's a tough theme to work with being middle-grey!

**Changes:**
- drastically simplified the logic by removing the 'adjust' variables that were scattered throughout the base, unfortunately I introduced much more complexity than was necessary with that approach and things are now much more simpler
- now we mostly have derived colors based on the bg-color (at the top of the theme file), contrast adjustments are made to these, about 5 colors each for light and dark tones, use when appropriate as base tones
- removed the Synthwave theme, this was actually intended to be a high contrast accessibility theme but I had no time to really delve into WCAG and there are still some limitations, so it would more or less be aesthetic preference which defeats the purpose, it's a good bet to remove this for now
- unwind the complex mix() logic used on x-sheet cell highlights, now using a simpler saturate(lighten()) and removes a lot of visual noise as it was introducing technically 'new' colors

**Fixes:**
- restore the original level colors [NEUTRAL]
- restore the lost visual details from 1.7.1R [DARK] (bordered textfield, menu, etc.)
- bold the text in the curve editor [ALL] (improve legibility of frame numbers)
- increase the contrast of border lines on x-sheet [ALL] to compensate for users with poor monitor calibration
- lighten topbar as it was in V1 [NEUTRAL] to match menu bg color (improve legibility)
- lighten x-sheet frame number column as it was in V1 [NEUTRAL] (improve legibility)
- lighten empty x-sheet cell range as it was in V1 [NEUTRAL]
- add disabled state for palette panel's lock button [ALL]
- lighten DvItemViewerPanel's sort column header to make it seem decoupled from the item list [ALL]

**Contentious (potentially things to revert if it's a problem):**
- the time indicator on the x-sheet/function editor is now magenta [NEUTRAL ONLY] in-sync with the curve editor, so as to catch the user's eye much faster, let me know if this is a problem

Mostly nothing much has changed, it's a lot of backend work to simplify things, except for the glaring usability fixes.


RodneyBaker:
Haven't scrubbed through all the changes of this PR yet.
The main area of concern for me is what gets changed in the defaults.

As for the Neutral Theme change...
Those that find the pink xsheet current frame color (potentially) contentious can add the following to Preferences > Edit Additional Style Sheet and alter the color value to their liking:

`XsheetViewer {qproperty-CurrentRowBgColor: #da7fd2;}`

For instance, although perhaps a bit too subtle:  

`XsheetViewer {qproperty-CurrentRowBgColor: #cdcdcd;}`

konero:
Added an update to Neutral, this time to try make it closer to V1's style (the classic blue).

image

Very specially, the issue was reading filenames under anything but perfect lighting conditions, the faded highlight just hasn't been working well for quite some time, at least from our end.

This makes the theme feel a bit less 'cohesive' and pretty, but it does improve the usability, which is what matters in the end.

I reverted the time marker color to a bright grey (as it was in V1), and removed the hue from the highlights on toolbuttons etc. in general as well.

I took inspiration from old Blender, Krita, Vegas, Avid and other interfaces to try find a balance, hopefully this works.

The blue is from Windows XP era.

KaseyFarron:
Really like the classic look of Neutral, reminds me of the first release all those years ago with the green rooms tab! No longer looks dull or faded 👍

RodneyBaker:
Anything that gets @KaseyFarron to decloak and comment has got to be legit.  ;)

Posting intent to merge.
We'll keep an eye on the areas that are potentially contentious.

RodneyBaker:
Thank you @konero for the update/fix.
Apologies for the delay.

Issue #6717, opened 2026-04-05
Fail to find glew 2.3


marillat:
### What's the bug?

Debian unstable amd64
GCC 15.2.0
git master up to date but same with 1.7.1.1

```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
GLEW_LIBRARY
    linked by target "toonzlib" in directory /build/opentoonz-dmo-1.7.1.1/toonz/sources/toonzlib
    linked by target "tnzstdfx" in directory /build/opentoonz-dmo-1.7.1.1/toonz/sources/stdfx
```

andeon:
Since you are on Debian Unstable, library paths change frequently. Even if you install the package now, your current CMake configuration is likely "stuck" on the previous error.

Install the missing dependency:

```
sudo apt update
sudo apt install libglew-dev
```
Clear the cache to wipe your build folder to force CMake to look for the libraries again:

`rm -rf build/*`

Now, retry the build and check if the error is gone!

marillat:
For the record I'm a Debian maintainer.

The glew 2.3 package  is not well packaged for now.

https://bugs.debian.org/glew

This bug has been 'added' I'll try again once this issue is fixed.
One more bug need to be fixed https://bugs.debian.org/1132709

marillat:
@andeon did you try to build opentoonz with GLEW 2.3 ?

andeon:
I've been recently testing with 2.3.1 to investigate issues with Wayland, but my stable version is still 2.2.0:

```
$ pkg-config --modversion glew
2.2.0
```
The problem seems to be that the OpenToonz script isn't "smart" enough to find libraries in modern Debian Multiarch locations or handle GCC 15 strictness. I imagine it's looking in `/usr/lib` and missing the `/usr/lib/x86_64-linux-gnu` path (which is the standard for Debian, right?), while also confusing variable names across sub-modules.

So, will clearing the cache and pointing to the library manually resolve the build?

```
rm CMakeCache.txt
cmake .. \
  -DGLEW_INCLUDE_DIR=/usr/include/GL \
  -DGLEW_LIBRARY=/usr/lib/x86_64-linux-gnu/libGLEW.so \
  -DCMAKE_CXX_FLAGS="-fpermissive" \
```

marillat:
> So, will clearing the cache and pointing to the library manually resolve the build?

[...]

PR #6836, opened 2026-04-13
Fix GLEW detection issues on newer systems

andeon:
This PR updates the GLEW detection logic in CMake to improve compatibility with newer Linux distributions, addressing the issue reported by @marillat in #6717.

The updated logic prefers the imported target GLEW::GLEW when available, while maintaining fallback mechanisms for legacy variables and pkg-config. This change ensures a more robust build configuration and resolves detection issues recently observed on Debian-based systems.


RodneyBaker:
Proceeding with this fix.
Thanks @andeon!

PR #5891, opened 2025-03-31
Refactor palette(Cleanup&Level) handling in scene management

Abel032:
I think this PR would resolve most crashes related to a dangling pointer of the palette (ACCESS VIOLATION of palette or style).
Here I list some of them out:
#5728 #5892 #5514
But I get the crash when I click on xsheet while OT loading a scene, which is different.

PR #5746, opened 2025-01-26
Fix Clean Up Palette Related Crash and cleanup parameters related adjustment

Abel032:
### How to reproduce the crash

1. Open a scene
2. Save All
3. switch cleanup related opinions(like change mono mode to color mode)
4. a palette related crash

This PR would resolve the crash that related to a dangling palette pointer and a nullptr of palette.

To not store current cleanup parameters to scene's parameters' cache(already write to scene file) if autosave.

Fixes #5257

PR #6871, opened 2026-05-13
Revert References of Global Cleanup Parameters

shun-iwasawa:
This PR is related to #5746 and #5891. This attempt to restore the functionality of the Global Parameter while preventing the crash.
As I mentioned in [this comment](https://github.com/opentoonz/opentoonz/pull/5746#issuecomment-4418174095), the Global Parameter in Cleanup Settings is intended to hold scene-wide settings, and I have reverted some code necessary for that functionality.

As far as I've tested it in my environment, the crash issue doesn't seem to be occurring.

@Abel032 Could you please let me know if you have any concerns about this?

### DIFF
--- a/toonz/sources/toonz/cleanupsettingsmodel.cpp
+++ b/toonz/sources/toonz/cleanupsettingsmodel.cpp
@@ -127,7 +127,7 @@ public:
       return false;
     }
 
-    CleanupSettingsModel::instance()->loadSettings(loadPath);
+    CleanupSettingsModel::instance()->loadSettings(loadPath, true);
     return true;
   }
 };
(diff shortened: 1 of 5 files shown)


Abel032:
@shun-iwasawa Sorry for the delayed reply because of some personal reasons.

I can't recall the details of the crash, but it was caused by a dangling pointer of palette stored in cleanup param.

I think we can do more test to find out after the merging, and I may review related code before the 1.8 release.

shun-iwasawa:
I’m going to merge this PR for further testing. @Abel032  Please let me know if you notice anything.

Abel032:
Sorry for reporting this after the 1.8 release, but I found a bug that may related to this change.

Current palette would switch to cleanup palette after saving the scene (inlcude the autosave).
20260625235658_rec_


I've tested last build before this PR and it doesn't have the issue.
https://github.com/opentoonz/opentoonz_nightlies/releases?page=3
image

The builds after this PR have the issue.
59b66133-83c3-4bbf-8b5d-19dd4148d292

PR #6874, opened 2026-05-16
Enable "Ignore AutoPaint Inks" function for RasterTape Tool

Abel032:
This PR enables "Ignore AutoPaint Inks" function for raster tape tool.

图片


shun-iwasawa:
LGTM

PR #6879, opened 2026-05-17
Fix: Converted level has no palette

### LINKED ISSUE #6875: Convert of aliased raster to Toonz raster fails to save palette
### What's the bug?

When opening an aliased raster image Opentoonz now recognizes it as a candiate for conversion to toonz raster which works fine.

There is a problem however when the conversion completes in that no palette is present unless/until the toonz raster level is saved.

If automatically saving is not preferred perhaps a message could prompt the user to save so the palette will be present?

@Abel have you seen this when opening and converting aliased raster images?
The concern is that users won't know what to do with the lack of a palette connected to the new toonz raster level.

Once saved/reloaded the TLV level behaves as expected.

### Steps to Reproduce

1. 
2. 
3.

_No response_

_No response_

### OpenToonz Version

1.7

_No response_

### Operating System

Windows

### GPU Brand

NVIDIA

### Graphics Tablet

None

Abel032:
### How to reproduce

1. enable “Convert Imported NAA Image Sequences to TLV“
图片

2. Load sequence images(Anliased) from OT's browser
图片

3. Do auto-conversion.

fixes #6875

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -2846,7 +2846,7 @@ void IoCmd::convertNAARaster2TLV(
   TApp *app                           = TApp::instance();
   ToonzScene *scene                   = app->getCurrentScene()->getScene();
   for (auto &rd : rds) {
-    TFilePath &path = rd.m_path;
+    TFilePath path = rd.m_path;
     if (path.getDots() == ".." &&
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thanks @Abel032 
I confirm this appears to fix the problem.

We'll wait for @shun-iwasawa confirmation as Appveyor appears to be down.

RodneyBaker:
Thanks @Abel032 for the fix!

PR #6882, opened 2026-05-20
Modernize macOS builds for Apple Silicon

bgyss:
## **Summary**

- Switch the macOS path to a native arm64 Apple Silicon workflow using Nix + mise
- Add bundle packaging, arm64 validation, and optional signing/notarization hooks
- Modernize app metadata so build dates and version strings are generated consistently
- Document the new setup plus the current feature gaps versus the legacy Intel macOS build

## **Testing**

- `mise run configure`
- `mise run build`
- `mise run package-macos`
- `mise run check-macos-arm64`
- `codesign --verify --deep --strict --verbose=2` on the packaged app
- `git diff --check`


bgyss:
looking into the appveyor failure

bgyss:
I looked through the AppVeyor log. This does not appear to be a compiler/linker failure from the PR. The build reached normal MSBuild compilation, then AppVeyor’s MSBuild logger failed while uploading an existing `C4251` warning:

`Error sending AppVeyor compilation message: The remote server returned an error: (500) Internal Server Error.`

MSBuild then stopped with `MSB4166` / `MSB4017` because the AppVeyor logger crashed. The warning it was processing comes from longstanding code in `tspectrumparam.h`, not from this PR.

Recommendation: rerun the AppVeyor job. If this repeats, the durable CI fix is probably to avoid AppVeyor’s automatic MSBuild logger path and run the build via an explicit `build_script` / `cmake --build` command instead.

PR #6883, opened 2026-05-20
Fix Output Settings > Linear Color Space Checkbox

shun-iwasawa:

[STEP TO REPRODUCE]
- Prepare two scenes: one with `Output Settings > Linear Color Space` set to `ON` (Scene A), and another with `Output Settings > Linear Color Space` set to `OFF` and `Channel Width` set to `16bit` (Scene B).
- Load Scene A and open Output Settings dialog.
- While Output Settings is displayed, load Scene B.

This causes the value of `Output Settings > Channel Width` to be incorrectly displayed as `8bit` instead of `16bit`.

The cause of the issue is that updating the `Linear Color Space` UI triggers a change to the `Channel Width`. This action should only occur when users click the checkbox.


RodneyBaker:
I confirm the problem and that this fixes the issue.

PR #6885, opened 2026-05-20
Japanese translation for v1.8

shun-iwasawa:
This PR will update Japanese translation for V1.8


RodneyBaker:
Will press forward with this although I suspect there could be some additional translations before v1.8 goes final.

Thanks @shun-iwasawa!

PR #6887, opened 2026-05-20
Fix tooloption assertion failure

shun-iwasawa:
This fix prevents an assertion failure that occurs when attempting to display the Fill Tool options in the Tool Option Bar.
Since the “Close Gap” option is bound only to the Toonz Raster Fill tool, `m_closeGap` will become null for the Vector Fill tool.


RodneyBaker:
LGTM

PR #6890, opened 2026-05-20
feat: auto detect painted Aliased Image and convert to TLV

Abel032:
This PR adds a new auto-conversion feature that detects painted NAA images and automatically converts them to TLV using 'Painted NAA Image' mode after importing.


RodneyBaker:
Thank you @Abel032 for the addition and for your patience.
LGTM

PR #6894, opened 2026-05-21
Fix duplicate levels of folder command

Abel032:
This PR fixes a bug discovered by @shun-iwasawa in #6084

### How to reproduce
1. Use `Load Folder` command to load a project folder(e.g., +extras)
图片
图片

2.1 Sequence with dot(.) separator would be loaded correctly
图片
2.2 Sequence with underscore(_) would wrongly duplicate and expose without notice dialog.
图片


shun-iwasawa:
LGTM Thank you @Abel032 for the fix !

PR #6902, opened 2026-05-23
Validate TLV/TZL header tables

bgyss:
## Summary

- Validates TLV/TZL frame counts, offset-table positions, table byte sizes, suffix lengths, and checked read/seek behavior.
- Fails malformed TLV headers cleanly before decoded values drive unchecked table logic.
- Adds `tzl_header_regression`.

## Validation

- Original split source commit: `92066d5e9`
- Regression target: `tzl_header_regression`
- Previous full-stack validation included the regression test and `image` target build.


RodneyBaker:
@bgyss Please let us know how we can help in your exploration.

PR #6903, opened 2026-05-23
Run external FX without shell parsing

bgyss:
## Summary

- Replaces POSIX external-FX `system()` command execution with `QProcess`.
- Passes the selected executable and options as explicit arguments so shell metacharacters remain data.
- Adds `externfx_process_regression` for a payload that would create a marker file if interpreted by a shell.

## Validation

- Original split source commit: `d5f0531c1`
- Regression target: `externfx_process_regression`
- Previous full-stack validation included the regression test and `tnzbase` target build.


RodneyBaker:
This PR won't be able to move forward in current form.
It is too broad and needs to be broken down into reviewable components.

A good start might to rebase/recreate the PR from current master, containing only the External FX fix and its tests.

Keep the QProcess conversion; it is a good improvement.

Other considerations moving forward:
- Address the ability of serialized scenes to control m_executablePath and m_args, or explicitly document why that is considered trusted.
- Replace the /usr/bin/true regression with a cross-platform helper executable test.
- Make the test verify that the helper actually launched and received the expected literal arguments.
- Add a security regression where a serialized ExternalProgramFx attempts to select a shell executable. The expected behavior should be that untrusted scene data cannot cause arbitrary execution.
- Decide deliberately whether %ENVIRONMENT_VARIABLE% expansion should remain supported on Windows.
- Propagate and handle runExternalProgram() failure instead of silently ignoring its boolean result.

PR #6904, opened 2026-05-23
Run crash symbolizer without shell parsing

bgyss:
## Summary

- Runs `atos`/`addr2line` via `QProcess` instead of composing a shell command for `popen`.
- Uses explicit argument lists, merged output, and a bounded wait while preserving symbolization behavior.

## Validation

- Original split source commit: `4e8122e2b`
- Previous full-stack validation built the `OpenToonz` target.
- This path is build-validated because a reliable unit test would require a crash harness.


RodneyBaker:
@bgyss 
The crash handler needs improvement but this PR arguably makes signal safety 
worse and not better.

The issue:
`popen()` is not POSIX async-signal-safe, but it calls `fork()`, isolating 
the child. The parent's signal handler context is brief.

Your QProcess changes add more library code to the signal handler; mutexes, 
malloc, Qt internals. This is unsafe in signal context (POSIX.1-2017 §2.4.3).

The trade off:
- Current: Light violation (sprintf safe, popen forks quickly)
- Proposed: Heavier violation (Qt code in signal handler)

Better path:
Defer symbolization outside the signal handler. Capture raw addresses in-signal (safe), process them post-crash in normal context where QProcess is perfect.

References:
https://pubs.opengroup.org/onlinepubs/9699919799/ (POSIX.1-2017 §2.4.3)
https://man7.org/linux/man-pages/man7/signal-safety.7.html

Recommend closing this PR and exploring that redesign instead.

PR #6911, opened 2026-05-27
Unify XDTS / SXF inbetween symbols

shun-iwasawa:
This PR unifies the symbols used to represent Inbetween Symbols in XDTS with those used in SXF.
This is to match the actual XDTS file format specification.

Reference: #4398


RodneyBaker:
Thank you @shun-iwasawa 
Much appreciated!

Issue #6913, opened 2026-05-27
Xsheet Selection Missing Behavior


RattusaurusRex:
### What's the bug?

*Missing Behavior*
In OT 1.7.1, to select frames and keys, you hold Ctrl and drag to select both.  This no longer works and I haven't gotten any answers on there being a new command to do this.

https://github.com/user-attachments/assets/58167214-d968-49da-b8ad-75f84f058ad1

*Key Deletion and Undo Bug* 
While trying to figure out if some other command had replaced the selection behavior, I found this bug where holding Alt while making a selection and moving the selection deletes all keys.  This action is not undoable, and the keys can only be brought back by reverting the scene.  Possibly only affects Mesh columns.

https://github.com/user-attachments/assets/9c38c32c-594f-43b4-8955-c0cdb62a7ff2

*Undo Crash* 
( In 1.7.1 and in #6916 ) There is a crash that can result from using Ctrl + drag if you also select the camera column.  This causes the columns to shift over.  Pressing Undo results in a crash.

https://github.com/user-attachments/assets/af62d3a3-c2cf-49eb-b7df-f5542ef012dc

### OpenToonz Version

Nightly (specify below)

### Exact Version Info (optional)

Mar 31 2026

### Operating System

Windows

### GPU Brand

NVIDIA

### Graphics Tablet

None

RattusaurusRex:
Updated the report to include a crash I found while testing the PR to bring back Ctrl + drag.

shun-iwasawa:
It seems that the second problem—where Mesh Column keys disappear when you Alt+drag the cells—was also present in V1.7.1.

PR #6916, opened 2026-05-30
Remove feature: Use Ctrl to drag cells

Abel032:
This PR removes the feature of holding Ctrl to drag cells without clicking the side bar.
To restore the feature of holding Ctrl to select both cells and keyframes.

Reference: #6913


RodneyBaker:
I tried testing this.
I'm still not able to properly drag exposures and keyframes together with Ctrl key.
Not that I've ever been able to do that reliably.

@shun-iwasawa I belive this is an important fix to get into the final v1.8 release.
Having said that, I'm not sure if this particular fix is what is needed.

RattusaurusRex:
It works for me.  I did find a crash (I updated the original report) that can happen if you accidentally select the camera column as well.  This isn't a new crash -- I could replicate it in 1.7.1 as well.

shun-iwasawa:
I confirmed it works fine in my environment as well.
@Abel032, thank you for fixing the regression. LGTM

@RattusaurusRex 
Thank you for reporting the problem. 
Sorry to trouble you, but could you please repost the remaining two problems in #6913 as separate issues?

PR #6934, opened 2026-06-16
Fix Crash on Dragging Camera Column Cells + Undo

### LINKED ISSUE #6924: Camera Column Selection causes Undo Crash
### What's the bug?

There is a crash that can result from using Ctrl + drag if you also select the camera column. This causes the columns to shift over. Pressing Undo results in a crash.

https://github.com/user-attachments/assets/d2f4124e-fd8a-4eee-b1f9-752689296e96

_No response_

_No response_

### OpenToonz Version

1.7

_No response_

### Operating System

Windows

### GPU Brand

NVIDIA

### Graphics Tablet

None

shun-iwasawa:
This PR will fix #6924 

- Modified `TCellSelection::selectCells` so that the cell selection won't contain the camera column cells.
- Modified`CellKeyframeMoverTool::onDrag` to move selected ketframes only when the selected cells are actually moved, which will prevent the selected keyframes from being dragged separately from the selected cells.


RodneyBaker:

PR #6946, opened 2026-06-24
update naru lazybrush

Appule:
## Overview

This PR updates `Naru LazyBrush` to support multi-color painting. It introduces a new feature that allows users to segment and colorize images into multiple distinct colors using a Reference image.

### 1. Basic Features

By simply connecting the line art to the Source input, the algorithm automatically fills the character area with a default gray color.

|  |  |
| :---: | :---: |
| Source | Result |

* `Color`: Changes the base fill color.
* `Threshold`: Adjusts the line darkness sensitivity. Lowering the threshold allows the algorithm to react to fainter lines.

<table width="100%">
  <tr>
    <td width="25%"></td>
    <td width="25%"></td>
    <td width="25%"></td>
    <td width="25%"></td>
  </tr>
  <tr align="center">
    <td>Source</td>
    <td>Threshold = 0.5</td>
    <td>Threshold = 0.3</td>
    <td>Threshold = 0.1</td>
  </tr>
</table>

### 2. Reference Image Feature (Manual Scribbles)

By feeding a marker image into the Reference input, you can manually place color scribbles at specific locations.

####  How to Create a Marker Image

1. Create a new raster or vector image with a **transparent background**.
2. Select the color you want to fill with.
* **Important:** Ensure you use a **completely opaque color (100% Alpha)**. Any transparency will cause the marker to be ignored.

3. Draw/mark the areas you want to color.

This workflow makes multi-color coloring seamless. It can even handle ambiguous boundaries that look like dotted lines (e.g., ears or hands).

<table width="100%">
  <tr>
    <td width="50%"></td>
    <td width="50%"></td>
  </tr>
  <tr align="center">
    <td>Source + Reference</td>
    <td>Result</td>
  </tr>
</table>

<table width="100%">
  <tr>
    <td width="50%">mouse_A</td>
    <td width="50%">mouse_col</td>
  </tr>
  <tr align="center">
    <td>Source + Reference</td>
    <td>Result</td>
  </tr>
</table>

#### Background Handling (BG Reference)
In the internal processing of multicolor painting, colors spread out from each color marker, defining the boundaries between them. To prevent the colors from spreading indefinitely, a **stopper (end point) that indicates “beyond this point is the background (outside)”** is necessary.

The color specified in the BG Reference functions as this “background stopper.”

Difference from transparent pixels:
“Transparent areas” within the reference image are treated simply as “areas where no color has been applied.” In contrast, the color specified in the BG Reference serves as a clear instruction (sink node) to “cut out this area as the background,” enabling automatic background transparency processing.

<table width="100%">
  <tr>
    <td width="50%">frog_t_A</td>
    <td width="50%">frog_t_col</td>
  </tr>
  <tr align="center">
    <td>Source + Reference</td>
    <td>Result</td>
  </tr>
</table>

---

### Mode

Switches the output visualization.

* `Mask + Line`: Blends the original image with the calculated mask using the selected Blend Mode.
* `Mask`: Renders the mask only.
* `LoG Filter`: Displays the image after applying the Laplacian of Gaussian (LoG) filter with minor value adjustments. Dark areas indicate regions likely to be segmented (Useful for debugging).
* `Scribble`: Visualizes the auto-generated and reference-based scribbles (Useful for debugging).

### Blend Mode

Changes the blending method between the mask and the original image.

### Auto Scribble Settings

Configuration for automatic marker placement.

* `Enable Auto Scribble`: Enables auto-placement of character markers.
* `Color`: Sets the color for auto-placed markers.
* `Enable Auto BG Scribble`: Enables auto-placement of background markers.
* `BG Reference`: Defines the color treated as the background. This color is also used when manually specifying background scribbles.
* `Threshold`: Determines the line darkness threshold for placing the two markers mentioned above.

### Advanced Parameters

Parameters related to the **Boykov-Kolmogorov Maxflow Algorithm** used in the reference paper.
Redundant parameters from the previous implementation have been removed for a cleaner, more concise setup. The parameters are tuned for general-purpose use based on the paper.

* `LoG Scale`: Controls the scale of the Laplacian of Gaussian filter. Higher values make it easier for the mask to split, even along faint lines.
* `Scribble Type`: Defines the scribble constraints.
  * `soft`: Scale the terminal edge capacity using `Scribble Softness`. Depending on the softness value, bleeding scribbles are more easily corrected by neighboring colors.
  * `hard`: Set the terminal edge capacity to a sufficiently large value. Bleeding scribbles will not be corrected and remain fixed to that color.
* `Scribble Softness`: Determines how aggressively bleeding scribbles are corrected (Higher values correct mistakes more aggressively).

### DIFF
--- a/stuff/config/current.txt
+++ b/stuff/config/current.txt
@@ -1728,19 +1728,15 @@
   
   <item>"STD_naru_LazyBrushFx" 	"LazyBrush Naru" 		</item>
   <item>"STD_naru_LazyBrushFx.mode" 		"Mode" 			</item>
-  <item>"STD_naru_LazyBrushFx.mask_color" 	"Mask Color" 	</item>
-  <item>"STD_naru_LazyBrushFx.undef_is_sink" 	"Fill Hole"	</item>
-  <item>"STD_naru_LazyBrushFx.auto_scribble" 	"Auto Mask Scribble"	</item>
-  <item>"STD_naru_LazyBrushFx.line_weight"	 	"Line's Weights"		</item>
-  <item>"STD_naru_LazyBrushFx.sink_pos" 	"Background Direction"	</item>
-  <item>"STD_naru_LazyBrushFx.lambda"	 	"Scribble Strength"		</item>
+  <item>"STD_naru_LazyBrushFx.blend_mode"	 	"Blend Mode"		</item>
+  <item>"STD_naru_LazyBrushFx.enable_auto_scribble" 	"Enable Auto Scribble"	</item>
+  <item>"STD_naru_LazyBrushFx.auto_scribble_color" 	"Color"	</item>
+  <item>"STD_naru_LazyBrushFx.enable_auto_bg_scribble" 	"Enable Auto BG Scribble" 	</item>
+  <item>"STD_naru_LazyBrushFx.ref_bgcolor" 	"BG Reference" 	</item>
+  <item>"STD_naru_LazyBrushFx.auto_scribble_threshold" 	"Threshold"	</item>
+  <item>"STD_naru_LazyBrushFx.log_scale" 	"LoG Scale"		</item>
   <item>"STD_naru_LazyBrushFx.scr_type" 	"Scribble Type"		</item>
-  <item>"STD_naru_LazyBrushFx.min_lightness" 	"Outline Cutoff"	</item>
-  <item>"STD_naru_LazyBrushFx.log_scale" 	"LoG Filter Scale"		</item>
-  <item>"STD_naru_LazyBrushFx.sigma"	 	"Filter Smoothness"			</item>
-  <item>"STD_naru_LazyBrushFx.alpha"	 	"Difficulty Passing Through Ink"			</item>
-  <item>"STD_naru_LazyBrushFx.auto_scribble_length" 	"Auto Scribble Margin"	</item>
-  <item>"STD_naru_LazyBrushFx.auto_scribble_threshold" 	"Auto Scribble Outline Threshold"	</item>
+  <item>"STD_naru_LazyBrushFx.lambda"	 	"Scribble Softness"		</item>
 
 </stringtable> 
 
--- a/stuff/profiles/layouts/fxs/STD_naru_LazyBrushFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_naru_LazyBrushFx.xml
@@ -1,21 +1,18 @@
 <fxlayout>
   <page name="LazyBrush Config">
     <control>mode</control>
-    <control>mask_color</control>
-    <control>undef_is_sink</control>
-    <control>auto_scribble</control>
-    <control>line_weight</control>
-    <control>sink_pos</control>
+    <control>blend_mode</control>
 
-   	<vbox shrink="0" label="Advanced">
-      <control>scr_type</control>
-      <control>min_lightness</control>
-      <control>log_scale</control>
-      <control>sigma</control>
-      <control>alpha</control>
-      <control>lambda</control>
-      <control>auto_scribble_length</control>
-      <control>auto_scribble_threshold</control>
-    </vbox>
+    <separator label="Auto Scribble Settings"/>
+    <control>enable_auto_scribble</control>
+    <control>auto_scribble_color</control>
+    <control>enable_auto_bg_scribble</control>
+    <control>ref_bgcolor</control>
+    <control>auto_scribble_threshold</control>
+
+   	<separator label="Advanced"/>
+    <control>log_scale</control>
+    <control>scr_type</control>
+    <control>lambda</control>
   </page>
 </fxlayout>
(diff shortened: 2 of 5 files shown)


Appule:
Since we need to ensure compatibility with older versions, please wait until then before merging.

RodneyBaker:
Thank you @Appule for this highly anticipated update.

As this and other FX requires Preview to be on to see the results on canvas, can you look into why Setting the Shrink value in Preview Settings (try something like 2 or 4) to accelerate that viewing of live coloring via Naru Lazybrush causes no fill to appear?
My sense is that this refreshing issue requiring Preview to be onfor FX to appear is the main issue that will make using Naru Lazybrush FX difficult.  Allowing users to set the Shrink value (with option to apply to main viewer set on) in Preview Settings will help with this.

There is also an issue with seeing the results in the LazyBrush Swatch viewer regardless of how the shrink value is set.  In the swatch viewer I can't see any fill being driven by the FX.

I was not expecting to see the Blend mode.
Very... very nice.

I've just started testing but that's first look.

image


Looking good!

RodneyBaker:
I'm not sure I fully understand the checkbox for Auto Scribble as I seem to get the same result regardless of if that checkbox is on or off.

Investigating...

(The refresh of the canavas view is working much faster now than at first and I'm not sure what I changed)

Appule:
This effect generates a graph from the image and determines where to place masks by solving the maxflow/mincut algorithm. In other words, rather than applying individual calculations to each pixel, information from adjacent pixels—and even pixels farther away—affects the overall result. Therefore, I believe the preview display shows results that differ from the actual output because the processing is applied to the image that has been scaled down for preview purposes.

Next, Auto Scribble is a feature that automatically generates masks within a character’s outline.
Therefore, when nothing is connected to the Reference, or when a transparent image is connected, it automatically creates a mask within the character’s area.
I’m not sure under what situations you’re toggling this feature on or off, but if you’re doing so while the character is already colored, it’s possible that the automatically placed scribbles will be overridden by the manually placed ones, resulting in no change even after toggling the feature.

shun-iwasawa:
There are two possible reasons why the rendering results of this Fx vary according to the “Shrink” value in the Preview settings.

1. Resizing the source image caused the contour lines to become fainter, so they were no longer recognized as boundaries of the filled areas.
2. If the reference image was drawn with thin lines, those lines were resampled along with the background during resizing, causing all the reference pixels for the fill color to become semi-transparent and be ignored. (The reference pixels in the reference image must be 100% opaque.)

If cause 2 is the issue, redrawing the lines in the reference image with a thicker stroke should resolve the problem.

In any case, the preview result with shrink is not guaranteed to be identical to the full-size result. Please understand that this is a trade-off for rendering speed.

Regarding the Swatch Viewer, when you zoom in on a portion of the image, it renders at a higher resolution based on that magnification. For most fxs, this isn’t a problem because processing is limited to the area cropped in the Swatch Viewer. However, some fxs like this expands the bounding box of the calculation area to cover the entire scene’s camera frame, so the increase in resolution significantly extends the calculation time. This is likely why the calculation results may take a long time to appear.

shun-iwasawa:
LGTM Thank you @Appule for upgrading the fx!

RattusaurusRex:
>1. Basic Features
By simply connecting the line art to the Source input, the algorithm automatically fills the character area with a default gray color.

I cannot get the basic mask to work, even with the simplest of drawings.  It only seems to create a rectangle around the drawings.  

image

Appule:
Thanks for the feedback!

Actually, this effect does not generate the background scribble if the lines touch the top, bottom, left, or right edges of the image.

As you can see in the image, since the circular line art is touching the edges of the frame, the background fill is failing to generate correctly. Please try adding some margin around your drawing—even just on one side will do—to allow the algorithm to detect the background area.

Appule:
Actually, I rethinked that—you'll need a margin on at least two sides. Applying it to either the top/bottom or left/right should be enough, but just to be safe, could you try adding a margin on all sides and see if that works?

RattusaurusRex:
It's not touching any edges.  The canvas is 1920x1080, same as the camera.  It's very far from any edge. It just seems like it's filling the savebox no matter what I do.

More lines to show that there are many shapes within the margins of the canvas: 
image


Here is the old version of the effect working on the same shape: 
image

Appule:
In older versions, setting 'Background Direction' to 'All' would force a background scribble at the four corners of the image, which allowed it to crop correctly.

Even if it looks like there is plenty of margin in your camera view, the lines in the rat image—specifically the circles near the top and right edges, along with the lines touching the other boundaries—are actually extending to all four edges of the image file's own bounding box (not the 1920x1080 canvas). 

In your current setup, because these lines reach the very edges of the image data itself, it prevents the auto background scribble from being generated. 

Could you please try adding some transparent or white padding directly inside the actual image file? This will ensure that the areas intended to be the background extend all the way to the edges of the image data instead of the linework.

RattusaurusRex:
> Could you please try adding some transparent or white padding directly inside the actual image file? This will ensure that the areas intended to be the background extend all the way to the edges of the image data instead of the linework.

So what you're saying is that I need to add a bunch of transparent marks to every frame of an animation to make this effect work now?

Appule:
Thank you for testing that and sharing the details.

My bad for not catching this sooner and making you double-check! After digging into it, I found that Auto Scribble has a compatibility issue with ToonzRasterLevel.

I've just opened a new PR to fix this. #6952
Let me know if it works for you now!

PR #6949, opened 2026-06-26
Fix stuff folder initialization

### LINKED ISSUE #6948: The preferences are not loaded correctly when the stuff folder is specified as a command-line argument
### What's the bug?

After the introduction of #6240, even when the stuff folder is specified via command-line arguments as below, the settings still refer to the registry-specified `TOONZPROFILES` directory.

`> OpenToonz.exe -TOONZROOT “path/to/stuff/folder”`

Furthermore, if the registry is not set on each client machine and the stuff folder is specified ONLY via command-line arguments, multiple warning pop-ups stating `“Cannot open menubar settings template file. Re-installing Toonz will solve this problem.”` appear at startup, and the Rooms template fails to load.

#6240 changed the order so that `ThirdParty::initialize()` is now called before `initToonzEnv()`, where the location of the stuff folder is set.
Since `ThirdParty::initialize()` accesses Preferences, it may access an incorrect Preferences file.

_No response_

_No response_

### OpenToonz Version

1.8

_No response_

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

None

shun-iwasawa:
This will fix #6948 . 

Move the calling of `ThirdParty::initialize()` in the function `initToonzEnv()` , just after configuring the stuff folders.

@Abel032
Could you please verify whether this change fulfills the original intent of #6240?

### DIFF
--- a/toonz/sources/toonz/main.cpp
+++ b/toonz/sources/toonz/main.cpp
@@ -181,6 +181,9 @@ static void initToonzEnv(QHash<QString, QString> &argPathValues) {
     fatalError("Folder \"" + toQString(stuffDir) +
                "\" not found or not readable");
 
+  // Setup third party
+  ThirdParty::initialize();
+
   Tiio::defineStd();
   initImageIo();
   initSoundIo();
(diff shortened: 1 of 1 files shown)


Abel032:
@shun-iwasawa 
Thanks for the fix! I've verified that #6240 is still fulfilled with this change. 
LGTM.

RodneyBaker:
Based on your response @Abel032 pressing forward with this.

PR #6952, opened 2026-06-30
Fix naru lazybrush CulcArea

Appule:
## Overview
This pull request fixes a compatibility issue where the "Auto Scribble" feature failed to correctly generate background scribbles when using `ToonzRasterLevel`.

## Background / Cause of the Bug
Previously, if a drawn line within a `ToonzRasterLevel` reached the edge of the image data's bounding box, the Auto Scribble algorithm could not properly detect or calculate the background area, resulting in a failure to generate the fill mask.

## Changes
- Adjusted the internal calculations of the "Auto Scribble" feature to correctly handle the boundaries of `ToonzRasterLevel`.

## Verification
Fixed the bug reported in #6946, where using `ToonzRasterLevel` resulted in the fill simply creating a rectangle around the drawing.

### DIFF
--- a/toonz/sources/stdfx/naru_lazybrush.cpp
+++ b/toonz/sources/stdfx/naru_lazybrush.cpp
@@ -211,7 +211,7 @@ private:
 
   //-------------------------------------------------------------------
 
-  void initRasters(TRasterP& fullRas, TRasterP& refRas, TTile& fullTile,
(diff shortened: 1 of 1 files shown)


RattusaurusRex:
Works as expected -- thanks for fixing this!  

image

RodneyBaker:
Thanks @RattusaurusRex for the confirmation.
Thanks @Appule for the fix!

PR #6960, opened 2026-07-14
Add Brazilian Portuguese (pt_BR) translation

draxxptb:
### Description
This PR introduces the official Brazilian Portuguese (pt_BR) localization for OpenToonz.

As OpenToonz has a growing community of animators in Brazil, having the software localized to `pt_BR` is a highly requested feature to improve accessibility for local artists and students.

### What was done
- Added `portuguese_brazil` to `LANG_CODES` and `LANG_DISPLAY_NAMES` in `toonz/sources/CMakeLists.txt` so the language is properly compiled by Qt and displayed in the user's Preferences menu.
- Translated `.ts` files for the core modules: `colorfx.ts`, `image.ts`, `tnzcore.ts`, `tnztools.ts`, `toonz.ts`, `toonzlib.ts`, `toonzqt.ts` (totaling over 6,200 strings).
- Applied industry-standard 2D animation terminology used by Brazilian artists instead of literal translations (e.g., *Frame* -> Quadro, *Pegbar* -> Barra de pinos, *Onion Skin* -> Casca de Cebola).

### Note on Accuracy
This is the initial version (V1) of the localization. While the translation was extensively reviewed to match standard animation terminology and UI patterns, there might be a few context-specific strings that will need minor adjustments as the community starts using it in their daily workflow. Feedback and future minor translation fixes from the community are expected and welcome!

### DIFF
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -660,11 +660,11 @@ if(WITH_TRANSLATION)
     # Language data and display names
     # ---------------------------------------------------------------------
     set(LANG_CODES
-    japanese italian french spanish chinese german russian korean czech
+    japanese italian french spanish chinese german russian korean czech portuguese_brazil
       )
 
     set(LANG_DISPLAY_NAMES
-    "日本語" "Italiano" "Français" "Español" "中文" "Deutsch" "Русский" "한국어" "Čeština"
+    "日本語" "Italiano" "Français" "Español" "中文" "Deutsch" "Русский" "한국어" "Čeština" "Português (Brasil)"
       )
 
     list(LENGTH LANG_CODES CODE_COUNT)
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -664,7 +664,7 @@ if(WITH_TRANSLATION)
       )
 
     set(LANG_DISPLAY_NAMES
-    "日本語" "Italiano" "Français" "Español" "中文" "Deutsch" "Русский" "한국어" "Čeština" "Portuguese (Brazil)"
+    "日本語" "Italiano" "Français" "Español" "中文" "Deutsch" "Русский" "한국어" "Čeština" "Portuguese"
       )
 
     list(LENGTH LANG_CODES CODE_COUNT)
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -664,7 +664,7 @@ if(WITH_TRANSLATION)
       )
 
     set(LANG_DISPLAY_NAMES
-    "日本語" "Italiano" "Français" "Español" "中文" "Deutsch" "Русский" "한국어" "Čeština" "Português (Brasil)"
+    "日本語" "Italiano" "Français" "Español" "中文" "Deutsch" "Русский" "한국어" "Čeština" "Portuguese (Brazil)"
       )
 
     list(LENGTH LANG_CODES CODE_COUNT)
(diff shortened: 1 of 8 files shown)


RodneyBaker:
@draxxptb If you close and reopen this PR it should restart and pass CI checks.
Thanks

RodneyBaker:
Posting intent to merge but the CI checks are struggling.

RodneyBaker:
I'm looking into why this PR fails checks.
As no program code changes we may move forward despite those failures but want to also use this as a trial text of current translation capability via direct use of PRs.
.ts files are now included in the build and that is no doubt why we are seeing the friction.

@andeon This might be a question for you as author of the current approach to adding/editing translations.

We could intentionally target the addition of .ts files to not trigger a build although I'm not certain that is desired.

Assuming all CI checks pass (without building) the translation update would just be included in subsequent updates/nightly releases.

I'm actively looking for ways to decrease building where it isn't necessary and this is just the latest to appear on that radar.

Bottom line:  There is almost no value in compiling OpenToonz three or more times via the CI (github action) builds especially when the nightly will build it all again.  There are some outliers to account for in this but as a general rule that seems to fit.

konero:
It says in the error logs: `EXEC : lrelease error : Unexpected characters '%n p�ginas'` on line 3443 (toonz.ts).

I don't handle translation files or file encoding at all so I got little experience, but it might be something to do with the file being saved at some point other than UTF-8. Open it in VSCode and click UTF-8 at the bottom right, click Save with Encoding and choose UTF-8, maybe that'll fix it.

image

RodneyBaker:
Thanks @konero that's a great lead.
I'll investigate and report back.

From a quick look the failure occurs specifically inside the MSBuild step so if encoding is involved it would have to be during actual translation compilation (or other build target) and not while CMake reads the UTF-8 related registration change.

Update 1:
The p�ginas in the log is almost certainly a (secondary) Windows console display issue and not a build issue. The Windows runner is apparently displaying the UTF-8 á incorrectly when printing the log.  That suggests a false positive for a build problem with UTF-8.  This isn't conclusive but seems very likely.  It was well worth checking out and educational as we see those kinds of symbols in logs quite often.

RodneyBaker:
Update 2 (a probable cause)

The problem appears to be related to the use of:

numerus="yes"

More specifically we need to change this:

```
<message numerus="yes">
    <source>%n page(s)</source>
    <translation>%n páginas</translation>
</message>
```

to this:

```
<message numerus="yes">
    <source>%n page(s)</source>
    <translation>
        <numerusform>%n página</numerusform>
        <numerusform>%n páginas</numerusform>
    </translation>
</message>
```
Note the use of `<numerusform>`

We have to do this for each numerus="yes" entry and not just the one reported at line 3443. If not lrelease may simply stop at the next malformed plural entry.

Added:  **The good news is that I only see that one instance that needs to be fixed.**
Update:  I found one additional place where the change was required.

I can repair on my end but always best to have the original author's consent.
@draxxptb are you still here?

RodneyBaker:
Informationally, I'm testing a validation check over at OT-Dev and will see how that works while exploring.  While not particularly useful now it could become so if/when we move toward auto translation. 

REF:  https://github.com/OpenAnimationLibrary/opentoonz-dev/blob/main/.github/workflows/translation_validation.yml

Anyone interested in the translation side of Opentoonz development is encouraged to add their thoughts into the mix.

RodneyBaker:
Apologies for the delay.
Thank you @draxxptb for the translation!

draxxptb:
Hi @RodneyBaker  @konero, sorry I couldn't be present in this discussion to help resolve the issue. I've been quite busy with college work. But thank you for fixing it!

PR #6961, opened 2026-07-15
Fix cleanup palette switch after scene save

### LINKED ISSUE #6959: Style Editor switches to Cleanup Palette on Save
### What's the bug?

The Style Editor switches from the current level palette to the Cleanup Palette after saving.  If using a Raster level, continuing to draw will result in the wrong color being drawn.  

Before save:
Image

After save:
Image

https://github.com/user-attachments/assets/842787c0-a4bc-4838-8d61-ffd00e2957a6

### OpenToonz Version

1.8

_No response_

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

_No response_

RodneyBaker:
Restore cleanup settings in place after a normal scene save so palette handles keep a stable palette pointer. Avoid the unnecessary restore and cleanup-palette rebind during autosave.

Fixes opentoonz/opentoonz#6959

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1524,12 +1524,12 @@ bool IoCmd::saveScene(const TFilePath &path, int flags) {
   }
   TApp::instance()->setSaveInProgress(false);
 
-  cp->assign(&keepCP);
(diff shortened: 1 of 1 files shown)


shun-iwasawa:
LGTM Thank you @RodneyBaker for the fix!

PR #6962, opened 2026-07-15
Fix macOS build with Homebrew OpenCV 4

RodneyBaker:
Homebrew’s unversioned `opencv` formula now installs OpenCV 5, which is incompatible with OpenToonz’s OpenCV 4 requirement. This updates the macOS workflow to install `opencv@4` and explicitly directs CMake to its configuration directory.

This should bring macOS github action builds back online.
I don't think this will address similar issue with nightly builds.

### DIFF
--- a/.github/workflows/workflow_macos.yml
+++ b/.github/workflows/workflow_macos.yml
@@ -59,7 +59,7 @@ jobs:
 
           brew update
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Proceeding with this fix to get MacOS builds running again.

PR #6965, opened 2026-07-24
Fix heap buffer overflow in BMP colormap loading

Chantology:
NOTE: This change was created by Claude Opus 4.8 and requires proper code review by a maintainer before merging.

BmpReader::readHeader() computes the colormap length from unvalidated header fields: biClrUsed is read straight from the file, and when it is zero the fallback (1 << biBitCount) can be as large as 1 << 15 because the surrounding guard only checks biBitCount < 16. The colormap is a fixed 256-entry buffer (new TPixel[256]), and the read loop writes `cmaplen` entries into it, so a crafted BMP overflows the heap. The only prior guard was assert(cmaplen <= 256), which is compiled out in release builds (NDEBUG).

A malformed 8-bit BMP with biClrUsed = 100000 reliably segfaults tconverter in a release build; with this fix it converts cleanly, and well-formed BMPs (which never exceed 256 palette entries) are unaffected.

Clamp the colormap length to the buffer capacity before the read loop.

### DIFF
--- a/toonz/sources/common/tiio/tiio_bmp.cpp
+++ b/toonz/sources/common/tiio/tiio_bmp.cpp
@@ -233,6 +233,18 @@ hd->biPad = 0;
   if (m_header.biBitCount < 16) {
     int cmaplen =
         (m_header.biClrUsed) ? m_header.biClrUsed : 1 << m_header.biBitCount;
+    // Proposed by Claude Opus 4.8 — REQUIRES MAINTAINER CODE REVIEW.
+    // Security fix (heap buffer overflow): cmaplen is derived from unvalidated
+    // header fields — biClrUsed is read straight from the file, and when it is
+    // zero the fallback 1<<biBitCount can be as large as 1<<15 (the guard above
+    // is only biBitCount < 16). The colormap is a fixed 256-entry buffer
+    // (new TPixel[256]) and the loop below writes cmaplen entries into it, so a
+    // crafted BMP overflows the heap. The previous sole guard, assert(cmaplen
+    // <= 256), is compiled out in release builds (NDEBUG). Clamp to the buffer
+    // capacity; valid indexed BMPs never exceed 256 palette entries, so this
+    // does not change behaviour for well-formed input. Malformed files may
+    // decode to a garbled image, but no longer corrupt memory.
+    if (cmaplen < 0 || cmaplen > 256) cmaplen = 256;
     assert(cmaplen <= 256);
     m_cmap.reset(new TPixel[256]);
     TPixel32 *pix = m_cmap.get();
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Hi Chan,

Thank you for your contribution to OpenToonz.

A few observations:

The good news is the proposed change is moving in the right direction and addresses the immediate heap buffer overflow.

There is still a concern with invalid palette sizes being unconditionally set to 256. 
A valid maximum palette size depends on the BMP bit depth:

* 1 bit BMP: 2 colors
* 4 bit BMP: 16 colors
* 8 bit BMP: 256 colors

Invalid palette sizes should therefore be rejected or validated against the image’s bit depth rather than indiscriminately clamped to 256. Clamping a malformed 1 bit or 4 bit BMP to 256 entries could cause the reader to consume data beyond the expected palette and leave the file stream incorrectly positioned.

Negative values will be changed to 256 as well in the current implementation.

As the detailed eleven line explanation is already captured in the PR description it does not need to be repeated in the source code. Any comment retained in the code should be limited to explaining behavior not otherwise clear from the code itself.

A possible direction the code could take:
```
// correctly compute BMP colormap length

if (m_header.biBitCount == 1 || m_header.biBitCount == 4 ||
    m_header.biBitCount == 8) {
  const UINT maxColorCount = 1u << m_header.biBitCount;
  const UINT cmaplen =
      m_header.biClrUsed ? m_header.biClrUsed : maxColorCount;

  if (cmaplen > maxColorCount)
    throw TImageException(m_path, "Invalid BMP color map size");

  m_cmap.reset(new TPixel[256]);
  TPixel32 *pix = m_cmap.get();

  for (UINT i = 0; i < cmaplen; ++i) {
    pix->b = getc(m_chan);
    pix->g = getc(m_chan);
    pix->r = getc(m_chan);
    pix->m = 255;
    getc(m_chan);
    ++pix;
  }
}```

Note to self:  Still need a regression test covering excessive biClrUsed values.

Chantology:
Thank you very much! I was doing some tests and let Claude check if there is anything it can find to reduce crashes, because there are so many reports about it. 

Will submit a new version soon! :)

RodneyBaker:
With your latest update, my technical objections appear to be resolved, subject to updating the PR description and successful CI builds.

I'll run some tests once those builds finish to see if everything works as expected.
It's always possible we've missed something.

Thanks @Chantology

RodneyBaker:
@Chantology 
Testing was going smoothly until I tested an 8 bit grayscale image (one rendered out of Opentoonz itself).  This crashes both the build from this PR and regular OT builds.

The good news:  The 8bit grayscale problem does not appear to be related to this PR.
As such we can proceed with merging this fix.

Investigation still pending regarding the 8bit grayscale BMP crash.

RodneyBaker:
LGTM

Chantology:
thank you @RodneyBaker 
I can maybe try to investigate about #6972 on Friday :)

PR #6966, opened 2026-07-24
Make Linux installs work without the launcher script

Chantology:

## What this does

Makes a plain `make install` / `ninja install` on Linux (and other Unix) produce a self-sufficient, relocatable application that runs **without** the generated `bin/opentoonz` launcher script. This removes the main obstacle to packaging OpenToonz as a Flatpak, RPM or deb, and keeps the AppImage working.

Previously the installed binary only worked when started through a generated wrapper whose only jobs were to (a) export `LD_LIBRARY_PATH` and (b) copy `stuff` / write `SystemVar.ini` into the user's config dir on first run. Both responsibilities now live in the build and the application itself, so the wrapper is gone.

## Changes

1. **`$ORIGIN`-relative RPATH** — Linux binaries were built with an absolute build-tree `RUNPATH`, and the CLI tools/libraries are installed via `install(FILES|PROGRAMS $<TARGET_FILE:…>)`, which copies files verbatim without rewriting RPATH — so installed binaries pointed back at the build directory. Now:
   ```cmake
   set(CMAKE_INSTALL_RPATH "$ORIGIN/../lib/opentoonz:$ORIGIN")
   set(CMAKE_BUILD_WITH_INSTALL_RPATH TRUE)
   ```
2. **First-run config seeding in the app** — new `TEnv::initUserStuffDir()` copies the installed read-only `stuff` into the per-user config dir and writes a minimal `SystemVar.ini` setting only the root variable (all other paths already fall back to `<stuff>/<subdir>`). No-op on Windows/macOS, in portable mode, with a `-TOONZROOT` override, or when already seeded.
3. **Same seeding in the CLI tools** — `tconverter`, `tcomposer`, `tcleanup`.
4. **Retire the launcher** — install the real binary plus a lowercase `opentoonz` → `OpenToonz` symlink, so the `.desktop` file's `Exec=opentoonz` keeps working.

## Review feedback addressed

All three points from @RodneyBaker's review:

- **Unix platform guards aligned.** The install rules drop the launcher for every `BUILD_ENV_UNIXLIKE AND NOT BUILD_TARGET_WIN` target, but seeding was guarded to `defined(LINUX) || defined(FREEBSD)` — so on Haiku/NetBSD/OpenBSD the launcher was gone with nothing replacing it, and the app would abort at startup. Widened to `!defined(_WIN32) && !defined(MACOSX)`.
  Widening alone would have written the file to the *wrong place*, because the seeding hardcoded `~/.config/<AppName>` while `getSystemVarPath()` reads it from `QStandardPaths::AppConfigLocation` on Haiku. The per-platform location is now split out into `getSystemVarFile()` and used by both reader and writer, so they cannot disagree.
- **First-run copying detects partial failure.** `TSystem::copyDir()` ignores the result of every `mkdir`/`QFile::copy` and never throws, so the surrounding `try`/`catch` could not observe a failure — and because seeding was skipped whenever `stuff` merely *existed*, an interrupted copy left a partial tree that every later run treated as complete, silently and permanently. The copy is now checked, staged under `<stuff>.incomplete`, and renamed into place only on full success; on failure the staging dir is removed and `SystemVar.ini` is not written, so the next launch retries. `TSystem::copyDir()` itself is left alone — it has two other callers, out of scope here.
- **AppImage claim completed.** The `Create Artifact` step moved `share/opentoonz/stuff` out of the appdir before packaging, so a standalone AppImage had no data to seed from. It now *copies* instead of moving: the AppImage retains `usr/share/opentoonz/stuff` (exactly where `initUserStuffDir()` looks, relative to the running executable), while the tarball still ships `portablestuff` beside it so portable-mode PR testing per `doc/how_to_test_prs.md` is unchanged. Deleting the lines outright would have satisfied the AppImage case but silently removed portable mode from the CI artifact, so the data is duplicated instead (~29 MB, the size of `stuff/`).

Also updated from current master, as requested.

Release publishing is deliberately **not** included; that was a separate concern in review and belongs in its own PR.

## Testing (Fedora 44 · GCC 16 · CMake 4.3 · Qt 5.15.18)

- `ninja install` to a prefix produces `bin/opentoonz -> OpenToonz` and no wrapper script.
- Launched via the `opentoonz` symlink with a **pristine `HOME`** and **`LD_LIBRARY_PATH` unset**: the binary starts (all libraries resolved through RPATH, confirmed with `ldd`), seeds the config, and gets past the stuff-dir check.
- Installing to a custom `-DCMAKE_INSTALL_PREFIX` works, and copying the whole install tree to an unrelated path still runs (true relocation).
- `tconverter` from an installed layout with a pristine `HOME` self-seeds and converts an image.
- **Failure path tested explicitly**: making a file inside the installed `stuff` tree unreadable to force a mid-copy failure → the error is reported, no partial `stuff/` or staging dir remains, `SystemVar.ini` is not written, and repairing the source then re-running seeds successfully.

## Notes for reviewers

- The `/opt/opentoonz` default prefix is unchanged; it only applies when `-DCMAKE_INSTALL_PREFIX` is not supplied and is fully overridable (verified).
- The config location is unchanged (whatever `getSystemVarFile()` reports per platform) rather than switching to XDG data dirs, to avoid relocating existing users' data.
- The pre-existing limitation that user `stuff` is copied once and so does not pick up new packaged resources after an upgrade is untouched, and per review is left for a separate PR.
- An earlier revision of this PR broke the Windows and macOS builds by calling `QCoreApplication::applicationDirPath()` in `tconverter.cpp`, where the `<QGuiApplication>` include is inside a `LINUX || FREEBSD` guard. That is fixed — the path is now derived inside `initUserStuffDir()` and no call site touches Qt.

- **Approach / intent given to the model:** make an installed OpenToonz work on Linux without the launcher wrapper, so the project can be packaged normally; then work through this PR's review feedback point by point. Findings were verified empirically on Fedora 44 (installing to temp prefixes, relocating the install tree, running with a pristine `HOME`, and forcing a copy failure) rather than asserted.


RodneyBaker:
This PR relates to #9667 but critically does not fully address the removal of portablestuff from the published AppImage.  


The current proposal works for a normal RPM, DEB, Flatpak, or ninja install layout where the installed tree contains:

bin/OpenToonz
share/opentoonz/stuff
lib/opentoonz/...

This PR actually makes the packaging mismatch more explicit as it establishes a proper installed location for stuff, but #6967 ’s AppImage process removes that location.

So we appear to get a situation where stuff directory content is resolved for normal installations; unresolved for #6967 ’s standalone AppImage.

There are many useful improvements in these PRs.
They need to be refined a little more.

Importantly, there is an issue of write permission where permission is granted for both GCC and Clang builds and for ordinary workflow executions.  Only the GCC tag publication step requires write access.

RodneyBaker:
Chan,
My recommendation:  Fork Opentoonz to where you can fully implement and test these changes.  Once changes are validated they can be ported into Opentoonz-master.

Keep the changes short and easily manageable and they will be easy to port and review.

If there is a need to keep higher level of attention on the effort maintain a document in Opentoonz Discussions or as appropriate.  In short, invite activity in your fork where full attention can be placed upon the changes.

This is perhaps the best direction particularly due to use of ai-authoriing via Claude.
For additional information please see:  https://github.com/opentoonz/opentoonz/discussions/6937

RodneyBaker:
Here's kind of a bottom line from my POV:

PR #6966 provides the mechanism needed for a self contained AppImage to initialize itself correctly, but PR #6967 currently removes the required installed stuff directory from the AppImage before publication. 

The two PRs will work together only after the AppImage workflow retains appdir/usr/share/opentoonz/stuff, or publishes the existing AppImage plus portablestuff archive rather than the AppImage alone.

Chantology:
thank you very much! I will do another round, do tests and submit smaller PRs - does it make sense to reject this PR and I create a new one or shall we keep it here?

RodneyBaker:
> thank you very much! I will do another round, do tests and submit smaller PRs - does it make sense to reject this PR and I create a new one or shall we keep it here?

My take is if unnecessary comments that can be captured in the PR itself can be reduced or eliminated this PR can proceed.  Thanks

RodneyBaker:
> thank you very much! I will do another round, do tests and submit smaller PRs - does it make sense to reject this PR and I create a new one or shall we keep it here?

Smaller and tightly scoped PRs are preferred but there are times when larger and broader PRs make more sense.  The one things we can say is that... IF... the PR can reasonably be broken down into smaller parts that makes the process of review a lot easier.

@shun-iwasawa lays out our intent for AI assisted Pull Requests quite well here:  https://github.com/opentoonz/opentoonz/discussions/6937

Specifically:  ```pull requests generated by AI will be reviewed through the same process as those written by humans.
Therefore, when creating pull requests, please break them down into small, separate requests—one per topic of change or improvement—so that humans can track and understand them.```

The part that will be a challenge for most of us:  ```Additionally, when using AI, if possible, please include or attach:

The name and version of the AI model that generated the changes
The prompt (or design document) entered into the AI (please apply a CC-BY license if necessary)
If another contributor writes a prompt based on a previously posted prompt, it is recommended that the name of the original prompt author be listed under “Co-authored by” in that PR.```

From what I have seen most ai-assisted development consists of many prompts so it seems reasonable that we might focus more on the desgin document.  

Given that AI is very likely generating the initial PR, feeding the AI Assistant a copy of Shun's message should help a lot.   We can probably refine that into a more specific ai readable directive.

All this to suggest that you are doing just fine thus far but we are still smoothing out the rough spots.  You are helping to move us forward in this regard.  Where you have specific concerns I'll recommend adding those to Shun's discussion.  I do think we want to keep things as open as possible because learning is also a vital part of this process.  Opentoonz is uniquely positioned in many ways to take advantage of ai assisted development.

For those with less need to explore ai driven development without directly adjusting Opentoonz source code Shun has even provide a path for that.  REF:  https://opentoonz.github.io/ai_tools/index.html

We want to be particularly careful to maintain the intent of these general guidelines and avoiding practices that will inevitably stifle development.  We see this in the form of 'No AI' policy for other software which while well intended misses the mark.  Keeping PRs simple and direct helps us accomplish that.

RodneyBaker:
@Chantology 
A quick assessment (and very possibly not up to date):

Additional changes are still required.
The normal Linux installation design is largely sound but the AppImage claim must either be completed or removed.

The Unix platform guards need to be aligned.
First run copying must detect partial failure. 

Recommend:  After these changes are made update from current master and verify via Linux CI checks.

The older “user stuff is copied only once and therefore does not automatically receive new packaged resources after an upgrade” limitation remains. That behavior predates this PR and can reasonably be documented and handled in separate PR.

Thanks

Standard disclaimers apply:  I can't yet speak Linux so it's good to have you here.

Chantology:
Thank you for the feedback! After reading some Issues, I would complete the AppImage for now. 
Maybe in the near future there can be a link to its Flathub on the homepage since this should be able to work on many distributions: 
https://flathub.org/en/apps/io.github.OpenToonz 

I had to ask someone to update to 1.8 which is not optimal. We need to find a better workflow for that :)

Chantology:
new PR with updated title and description

Chantology:
Built the AppImage from this branch and tested it end to end:

- Built in an `ubuntu:22.04` container, then run on **Fedora 44** — so cross-distro
  portability is confirmed in practice, not just in theory.
- Launched from a **pristine `HOME`**, from a directory with no `portablestuff`
  nearby (so portable mode is not involved).
- Result: it seeds `~/.config/OpenToonz/` from the `stuff` bundled inside the
  AppImage and writes `SystemVar.ini` with
  `TOONZROOT=<home>/.config/OpenToonz/stuff`, then starts normally.

That was the one thing still unverified — the AppImage claim is now demonstrated
rather than asserted.

Screenshot From 2026-08-08 13-56-58

RodneyBaker:
@Chantology 
My first inclination is to move this PR toward merger BUT... I hesitated because the code touches more than just Linux.

That's a good thing because we don't want Linux code to be treated any different that for other platforrms and I need a reset to steer clear of a mentality of 'Oh, that's just for Linux and Opentoonz doesn't officially support Linux so it doesn't matter'.  I'm learning.

I love comments, I really do, BUT... others do not love comments and the prevailing concensus is not just to reduce comments but to eliminate them entirely.  And now we finally get to the point.

Please run through this code again with a focus on transferring the comments into the code itself.  There are some comments for instance that just address why old code was removed.  

Another example is where comments are numbered:
```
// 1. Seed the writable stuff tree from the installed read-only copy. Stage it
// under a temporary name and rename into place only once the whole tree
// copied, so an interrupted or failed copy (out of space, permissions)
// leaves nothing behind that a later run would mistake for a complete
// "stuff" and skip. On failure the staging dir is removed and we bail out
// without writing SystemVar.ini, so the next launch retries from scratch.
if (!TFileStatus(userStuffDir).doesExist()) { ... }

// 2. Write a minimal SystemVar.ini. Only the root variable is required;
// every other path falls back to <stuff>/<subdir> in TEnv/ToonzFolder.
if (!TFileStatus(systemVarFile).doesExist()) { ... }
```

Numbering steps in comments is a tell tale sign that the function is doing two (or more) distinct jobs glued together. Pulling each into its own function would convert the comment into the function name/doc, and shrink initUserStuffDir() to a readable sequence.  For instance:

```
void TEnv::initUserStuffDir() {
#if !defined(_WIN32) && !defined(MACOSX)
  EnvGlobals *eg = EnvGlobals::instance();
  if (eg->getIsPortable()) return;
  if (eg->getArgPathValue(eg->getRootVarName()) != "") return;

  TFilePath userStuffDir = getUserStuffDir(eg);
  if (!seedStuffTreeIfMissing(userStuffDir)) return;
  writeInitialSystemVarIfMissing(eg, userStuffDir);
#endif
```
}

It may be appropriate to have ai assistants run a pass toward the end of refinement specifically targeting transfer of comments/instructions into code.  This perhaps even better if the code is using constructs used commonly throughout Opentoonz source.  (I say this with the knowledge and reservation that OT source is often considered outdated and in need of modernization)

Now to the intesting part.
All of this is written in large part beause I saw a lot of code I thought was very appropriate to be present in the code.  Examples include: 

few explain cross-file coupling or non-obvious rationale that no amount of renaming would surface:

The getSystemVarFile() split-out comment (tenv.cpp)
This explains why the function exists (so the writer and reader of SystemVar.ini can't disagree), which is exactly the kind of thing a reader can't infer from the diff alone. 
Keep this.

The copyDirOrFail() comment about TSystem::copyDir() swallowing errors
This is essential context for why a near duplicate copy function exists instead of reusing the existing one. 
Keep, although perhaps trim to one sentence?

The symlink-not-followed comment inside the copy loop
This protects against someone trying to simplify it later into infinite recursion.
Keep this.

The $ORIGIN-relative RPATH comment in CMakeLists.txt
CMake has no place to hang a docstring so this is the right home for that reasoning.
Keep this.

The tenv.h doc comment above initUserStuffDir()'s declaration
This is a proper public API doc comment in exactly the right place (header, above declaration). No change needed.

All this, for what its worth.
I enjoyed reading ALL of the comments. :)

Additional thought:
Worth considering for historical purposes if nothing else is how we might leverage github versioning to first include entirely too excessive commentary and even ai assistant instruction but then on subsequent pass remove all of that.  That missing/versioned commentary then can be recovered as needed.

Thanks for your patience @Chantology

Chantology:
thank you very much for the detailed feedback @RodneyBaker ! I will go through it again and I agree with the amount of comments. It's quite a lot and I am sure it's hard to review. Will go through your feedback in a couple of days. 

In general I would be happy if Open Toonz could have proper Linux support in the future. No idea how an experienced programmer could join the team since this would be super helpful for reviewing and providing comments that help this project in the long run.

Chantology:
@RodneyBaker this one is also good to go now?

Chantology:
Updated from current master (4c0e5848e): merges cleanly, builds clean on Fedora 44, and the branch now inherits the `ubuntu-22.04` AppImage runner from #7010.

Recap of what changed since your last review, since it wasn't spelled out:

- **Comments cut** (d1f3f5169): the numbered `// 1. … // 2. …` blocks you quoted are gone, along with the comments that only explained why old code was removed.
- **AppImage claim verified end to end** ([comment above](https://github.com/opentoonz/opentoonz/pull/6966#issuecomment-5226008663)): built in an `ubuntu:22.04` container, run on Fedora 44 from a pristine `HOME`, seeds `~/.config/OpenToonz` from the `stuff` inside the AppImage and starts normally.

Worth noting: #7001 just picked up a report that is exactly the bug this PR fixes, on a fresh system the AppImage dies with `Undefined or empty: ""`, because first-run seeding currently lives only in the `bin/opentoonz` script that the AppImage never runs.

Chantology:
clang-format check failed. Will have a look at it soon!

Chantology:
@RodneyBaker this one should be good to go finally. Let me know if there are any questions!

RodneyBaker:
@Chantology 
All three of the previous review points are addressed.

Merge readiness: close, with two things worth fixing yet:

There is a spurious 29 MB copy for users with a custom root. initUserStuffDir() bails on portable mode and on a -TOONZROOT argument, but not on a SystemVar.ini that already points TOONZROOT somewhere else. A user whose root is, say, /mnt/data/OpenToonz stuff but who has no ~/.config/OpenToonz/stuff gets the whole tree copied into ~/.config on next launch, sits through it, and never uses it. The ini itself is untouched, so it's waste rather than damage. One extra early return — if getSystemVarPath(getRootVarName()) is non-empty and is a directory, return — closes it.

The guard alignment is complete in tenv.cpp but not at the tconverter call site. initUserStuffDir() is now compiled for !_WIN32 && !MACOSX, but in tconverter.cpp the QGuiApplication instance is still inside #if defined(LINUX) || defined(FREEBSD). On Haiku/NetBSD/OpenBSD the function is compiled in and reaches QCoreApplication::applicationDirPath() with no application object, which returns an empty string and a Qt warning, therefore seeding silently no-ops on exactly the platforms the widening was meant to cover. tcomposer and tcleanupper construct QApplication unconditionally, so only tconverter is affected, and only when it runs before the GUI has ever seeded. Narrow, but it's the same class of mismatch that prompted the point. Deriving the exe directory from getApplicationFileName() when no instance exists would fix it without creating a QGuiApplication on a headless BSD.

Additional Notes (not blockers)
CMAKE_BUILD_WITH_INSTALL_RPATH TRUE replaces the default build RPATH, which previously included linker search paths. CI uses distro Qt so it stays green, but anyone building against Qt from a custom prefix (aqtinstall and similar) will now need LD_LIBRARY_PATH to run build/bin/OpenToonz. Worth a line in the PR description at least.
The real root cause is install(PROGRAMS "$<TARGET_FILE:…>") for the CLI tools; install(TARGETS), which the GUI binary already uses, would let CMake rewrite RPATH properly and make the BUILD_WITH_INSTALL_RPATH workaround unnecessary. A reasonable follow-up no needed for this PR.

Files created by install(CODE) don't land in install_manifest.txt, so the symlink is invisible to manifest-based uninstall. Most packagers glob the tree, so this is cosmetic.
Two instances launched simultaneously on a first run: the loser's QDir::rename fails, it removes its staging dir, prints the error, and skips writing the ini for that run. Re-checking whether the target now exists before treating the rename as a failure would make it graceful.
doc/how_to_build_linux.md still says copying stuff into ~/.config/OpenToonz is "currently required to run OpenToonz." Still true for build-tree runs, now false for installs.
Nothing in CI executes the binary, so the entire new startup path rests on the author's manual Fedora testing. A cheap smoke step in the Ubuntu job would change that: HOME=$(mktemp -d) QT_QPA_PLATFORM=offscreen appdir/usr/bin/tconverter …, then assert SystemVar.ini and stuff/ exist.  Given that no reviewer on the project can easily test Linux, that's probably the highest-value addition here.

I'll suggest one more pass through the comments to make sure they are as concise as possible.  They looked useful and relevant but I've been on a 'brief, concise and generally terse (BCgT) kick for code comments of late. Obviously not so much BCgT on the above wall o' text but for that I blame ChatGPT.

Chantology:
Thanks @RodneyBaker   checked both against current head, both reproduce.

**tconverter guard:** confirmed. `QGuiApplication` sits inside `#if defined(LINUX) || defined(FREEBSD)` while `initUserStuffDir()` compiles for `!_WIN32 && !MACOSX`. `tcomposer` and `tcleanupper` build `QApplication` unconditionally, so tconverter is the only one. The `getApplicationFileName()` approach works cleanly — `setApplicationFileName(argv[0])` already runs three lines before the `initUserStuffDir()` call, so nothing needs reordering.

**Spurious copy:** confirmed, and it's 29 MB exactly. One question on the guard before I write it. `if (root non-empty && isDirectory) return` leaves a case open: if `TOONZROOT` points somewhere the user has since moved or deleted, `isDirectory()` is false, so we copy 29 MB that still never gets used — the ini isn't rewritten either way. A comparison closes that:

```cpp
TFilePath configuredRoot = eg->getRootVarPath();
if (!configuredRoot.isEmpty() && configuredRoot != userStuffDir) return;
```

That skips any custom root whether or not it currently exists, while still repairing a missing *default* root. Is that your intent, or would you rather the copy proceed as a repair attempt when the configured root is gone?

**CI smoke test:** agreed, highest value here: nothing in the Linux job executes a binary today. Your invocation works as written: `LINUX` is defined in CI so tconverter does get its `QGuiApplication`, and seeding runs before argument parsing. I'll add it.

Also noted: the `LD_LIBRARY_PATH` caveat goes in the description, and `install(TARGETS)` stays a follow-up.

Comment pass last, once the code stops moving. That's the third time comments have come up on this PR, so I'll treat BCgT as the standing default rather than a one-off.

The `how_to_build_linux.md` line I'll send separately. This is a different topic.

RodneyBaker:
The 29 MB lands in a directory nothing ever reads.
Reading the branch head: `writeInitialSystemVarIfMissing() ` returns as soon as SystemVar.ini exists, so in the missing-custom-root case TOONZROOT keeps pointing at the vanished path no matter what the copy does. 

"Proceed as a repair attempt" would only mean anything if the seeding also rewrote the user's root which is the wrong behaviour: a root that isn't there is usually an unmounted external or network volume rather than a deleted one. 

Silently repointing it at ~/.config/OpenToonz hands the user an empty environment, hides the real cause, and leaves their config pointing somewhere else once the volume comes back. Skipping lets the existing startup error say the root is missing, which is the accurate diagnosis.

There's also a determinism argument for their formulation: `isDirectory()` makes the behaviour depend on whether the drive happens to be mounted at that instant, with same config, two outcomes, one of them a 29 MB copy. 

Comparing against the default location is config-driven and stable. 

So yes, skip any root the user has configured elsewhere, existing or not; only ever seed or repair the default location.

Two things worth passing back with it:
Since `configuredRoot` is computed anyway, key the ini write off that rather than off the file existing.  That is: if (configuredRoot.isEmpty()) writeInitialSystemVarIfMissing(...), or drop the doesExist() check inside the helper. 

As written, an ini that exists without a TOONZROOT key (empty file left by an interrupted sync(), a hand-edited one) still gets the full copy and still never gets the key, which is exactly the Undefined or empty: "" from #7001. `QSettings::setValue` merges, so any other keys in the file survive.

And the comparison is a plain TFilePath compare, not path identity.
Trailing slashes normalise, but a symlinked or differently spelled $HOME in the ini won't match, so the repair-the-default case is skipped. That's the harmless direction (no copy, clear startup error), so it isn't worth canonicalising; just worth knowing it's a string comparison.

That's the story for now at least...

Chantology:
@RodneyBaker  All five items are in and this is ready for another look.

**tconverter** needed a different mechanism than we discussed. `getApplicationFileName()` will not work here: `TEnv::setApplicationFileName()` stores `fp.getName()`, which tfilepath.h documents as "no extension, no dots, no slash", so it holds the bare string `tconverter` with no directory to derive from. I said that approach worked in my last comment. That was wrong, sorry.

Your underlying constraint still holds, so I kept it. `QGuiApplication` needs a display and would abort on a headless box. `QCoreApplication` does not, and `applicationDirPath()` works with any instance:

```cpp
#if defined(LINUX) || defined(FREEBSD)
  QGuiApplication app(argc, argv);
#elif !defined(_WIN32) && !defined(MACOSX)
  QCoreApplication app(argc, argv);
#endif
```

Checked with DISPLAY and WAYLAND_DISPLAY unset: the new branch returns a correct applicationDirPath, the old code returns an empty string.

**Root guard** as you described it. `configuredRoot` drives the ini write too, so an ini that exists without a root gains one and keeps its other keys. Confirmed against Qt 5.15 that setValue merges rather than truncating.

**Smoke step** in the Ubuntu job runs the installed tconverter under a throwaway HOME and asserts three things: a first run seeds and writes TOONZROOT, a root configured elsewhere is left alone with no copy, and an ini without a root gains one while keeping TOONZPROFILES. Plus an ldd check for the $ORIGIN RPATH. It targets /opt/opentoonz rather than the AppImage staging dir so it exercises the real installed layout. It fails against the pre-fix logic at the custom-root case, so it is a real regression test rather than one that passes either way.

**Comments** trimmed, 40 lines down to 21, no code changes. The tenv.h summary also gained the custom-root no-op case, which the old wording predated.

Ubuntu, Fedora, Windows, Formatting and AppVeyor are all green. macOS failed on `Failed to CreateArtifact: Unable to make request: ETIMEDOUT`, which is the artifact upload after the build itself succeeded. It passed on the previous head and the only change since is comment text, so it needs a re-run rather than a fix. Could you kick it off when you get a moment?

PR #6967, opened 2026-07-24
ci(linux): add Fedora build job and publish AppImage to Releases

Chantology:
NOTE: This change was created by Claude Opus 4.8 and requires proper code review by a maintainer before merging.

Two Linux CI gaps addressed:

1. Fedora coverage. The Linux workflow only built on Ubuntu, so Fedora-specific breakage (package renames, newer GCC/Qt) went unnoticed even though OpenToonz builds and runs there. Add a compile-only regression job running in a fedora:latest container with the equivalent dnf dependency set. It intentionally does not build an AppImage; its purpose is to catch build regressions on a modern distro.

2. Release artifact. The Ubuntu job already builds an AppImage but only ever uploaded it as a workflow artifact, so there was no official Linux binary for users or downstream packagers (e.g. Flathub). Add a tag-triggered step that attaches the AppImage to the GitHub Release (gcc build only, to avoid duplicate assets).

The Fedora dependency list was verified to install in the fedora:latest container and to build OpenToonz on Fedora 44.

### DIFF
--- a/.github/workflows/workflow_linux.yml
+++ b/.github/workflows/workflow_linux.yml
@@ -25,6 +25,9 @@ on:
 jobs:
   Ubuntu:
     runs-on: ubuntu-24.04
+    # Needed so the tag-triggered step can attach the AppImage to the Release.
+    permissions:
+      contents: write
     strategy:
       matrix:
         compiler: [gcc, clang]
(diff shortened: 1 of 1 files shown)


RodneyBaker:
We need some additional Linux minds involved in this but a first look suggests this PR might work best served by splitting into two PRs:

Fedora compile coverage
Linux release packaging and publication

The Fedora compile coverage appears mostly complete but I have some concerns about the release packaging and publication.  

Perhaps some of this is covered by #6966 so will investige.

Aside:  I need to set up some Linux systems...

cc: @andeon

Chantology:
I will submit smaller PRs

ci(linux): add a Fedora compile check - #6986
ci(linux): don't fail the AppImage build on a transient AppStream check- #6987

Issue #3563, opened 2020-10-24
"Selected strokes matched mode" for the selection tool (edit accros levels)


openanim:
**Is your feature request related to a problem? Please describe**

This request like this one #3562 is in order to help and save time in multiples operations. It's based on the principle matched strokes found in some OT tools and very stably on the control point editor see below.

![selected matched strokes](https://user-images.githubusercontent.com/29790512/97064687-140cbc00-15a8-11eb-8c89-2ea94dc7ecba.gif)

I suggest  a new "selected strokes matched" mode for the selection tool which would also work as the point control tool.

![request selected matched stokes](https://user-images.githubusercontent.com/29790512/97064843-774b1e00-15a9-11eb-90a6-d552b13b368a.gif)

This would make it possible to directly select all the strokes which correspond to each other in the level and to be able to edit them as with the selection tool in standard mode.

- remove
- copy / paste / cut
- colorize by selecting a style

This would have several advantages like

- no need to first, create a style then to select strokes of the same style in order to color them or edit them in any way

- this always select the strokes that match, regardless of the complexity of the animation

- very useful in the auto inbetween correction process in the case of matching errors of different keyframes in pose to pose method

Overall imo that could be potentially help for everything if we could extend the "match strokes features" in different area of OT  to editing/colorize matched strokes, move/remove matcheds points etc. whitout limitations as frame range feature has since that ill follow automatically the animation.

PR #6973, opened 2026-07-28
Hide line tool

Ahl76:
**Summary**

This pull request adds the Hide Line Tool (more information can be found here #6464), which lets artists hide parts of vector strokes without deleting them. Hidden/invisible sections stay in the drawing and can be shown again later.

**Inspiration**

The tool follows the interaction model of OpenToonz’s existing vector Eraser, especially the Segment mode (draw a guide line; affect strokes where it touches), along with the familiar Normal and Freehand types.

Unlike the Eraser, Hide Line does not remove geometry: it only hides selected portions. That non-destructive approach is inspired by the workflow philosophy found in tools such as Cacani.

**What it does**

Hide selected portions of vector lines instead of erasing them
Reveal those portions again when needed (unhide)
Keep the rest of the drawing and workflow intact (fill, inbetweening, and display take hidden parts into account)

**How to use it**

Select the Hide Line tool in the toolbar
Choose a type of selection:
Normal : brush-like hiding along the stroke
Segment : hide where a guide line touches existing strokes
Freehand : hide inside a freehand selection

**Choose a mode (difference for paint fill) :**

Invisible > the stroke portion is not drawn and still acts as a barrier for fill. In the viewer overlay it appears as a solid orange line.
Hidden > the stroke portion is not drawn, but fill can pass through it. In the viewer overlay it appears as a green dashed line.
Optionally use Unhide to restore previously hidden sections

**View option**

A View menu command (Hide Line: Show Invisible/Hidden Strokes, shortcut also available) shows or hides an overlay of the masked stroke parts in the viewer, so we can check what is hidden while editing:

orange solid = Invisible (blocks fill)
green dashed = Hidden (allows fill through)

Shortcuts are available to switch tool type, mode, and unhide

Notes : This feature was developed with Cursor’s AI models.


RodneyBaker:
@Ahl76 I've barely spent time with your Hide Line Tool but for what I have seen and tested it's working quite well.  Nice work.  

I had some ideas of what this feature would be going in and I am happy to report this is simpler than what I imagined.  Very intuitive.* 

Your description has answered almost all of the questions I had while exploring.
The code itself will likely answer the remainder.

I like the non-destructive aspect of this feature set.
My brain started asking:  Why not add a delete option along with the invisible and hidden options but... then we adding destructive processes into the mix likely best suited for the Eraser Tool.  I believe you are on to something better here.

I do have some follow up questions (no real concerns) relating to feature requests in the past that this PR almost addresses.  I'll try to formulate those more fully before posting about them.  In my estimation they are all outside the scope of this PR.

Aside:  This feature set also reminds me a lot of current Fill Tool Capability with both vector and toonz raster that is largely undocumented so it makes me think this tool can be further extended to support toonz raster levels in the future.  Perhaps not but I mention it here as a potential follow up for when those mostly hidden processes are updated.  Those updates could perhaps be best added to the Hide Line Tool. 

At any rate, just a note to say thanks for leaning forward with this.

*The one part that wasn't immediately intuitive was how to display the invisible/hidden lines and your description answered that.

RodneyBaker:
Render appears to work fine but I believe I see an issue with copy/paste and conversion.

To test:
Create a vector level with invisible and hidden vector strokes and filled areas.
Render to confirm all is working well.

In xsheet select vector frame > Right Click > Convert to Toonzraster
Hidden and invisible lines likely will be visible and fills in those areas defined by those lines may change.

Still investigating... 

Quick assessment (don't take this assessment too seriously):

When `rd.m_showHideLineStrokes` is true, it additionally draws the diagnostic orange and green centerline overlays.

That is appropriate in the Scene Viewer but probably inappropriate for:

Conversion to Toonz Raster
Conversion to Raster
render to file
thumbnails
icons
clipboard rasterization
any off screen RasterPainter operation

I think what I'm trying to say is that some processes don't know the vector segments/strokes should be considered hidden/invisible.  As such they will likely treat them as visible.

Added:  A (partial) workaround for conversion and rendering:  Treat all hidden and invisible lines as being assigned to a transparent style.  This will allow the conversion/rendering process to complete albeit without those lines.  Not a fix but worth investigating.

Ahl76:
@RodneyBaker  Thanks for testing and for your feedback on the issues you encountered. I’ll look into this over the next few days.

vpaintanimation:
Does it support setting a frame range?

RodneyBaker:
> Does it support setting a frame range?

No.  Not yet.

Update:  Yes, it does.

RodneyBaker:
@vpaintanimation 
My recommendation is the addition of Frame Range support is out of scope for this PR.
As such it should be pursued as an enhancement to the feature set after this PR is merged.
Entirely your call however @Ahl76 

[Frame Range addition to PR6973.pdf](https://github.com/user-attachments/files/30667668/Frame.Range.addition.to.PR6973.pdf)

RodneyBaker:
It is best to merge this before #6974

RodneyBaker:
@Ahl76 Adding a Frame Range option is outside of the scope of this PR in my estimation but I did some exploratory work on that here:  https://github.com/OpenAnimationLibrary/opentoonz-dev/pull/85

RodneyBaker:
@Ahl76 
I believe this is where I test your true patience with me as there is a lot I want to say and I don't think I can say it correctly.  This part I can though and hopefully it can express where I personally want to go with this PR.   I love this feature and want to merge it now.

Having said that, I've attached a document that supplies some background on the Tape Tool's hidden line tagging simularities (for toonz raster) and much much more regarding this PR.  None of what is contained in the PDF should stop merger but it should suggest there is much more work to be done.  How much needs to be done?  Not much and yet lots more.  So please, read on!

[Hide_Line_Tool_Obstacles_and_Opportunities.pdf](https://github.com/user-attachments/files/31253389/Hide_Line_Tool_Obstacles_and_Opportunities.pdf)
 
ChatGPT condensed a lot of my notes/thoughts into this PDF and if it doesn't make sense feel free to blame ai.  It added a few thoughts and opinions I didn't suggest.  The last paragraph reads to me as nonsense.  Frame Range opton in the future... not even mentioned.  ;)

RodneyBaker:
I would be remiss not to comment on the most interesting part of this PR.
How the code is dealing with the segments of a stroke without breaking that stroke into parts.
There's some high level stuff going on here.

Interestingly, I think I've discovered a problem while admiring this code.
If we hide parts of lines and then save the scene (via Save All), then reopen the scene the hidden parts are no longer hidden.  In other words, the segments hidden by the Hide Line Tool do not appear to stay hidden after a save and reload.

Likely worse, when saving the level via Save Level I can ocassionally get a crash.
I need to investigate that farther.

Maybe both me and this computer are tired so I'll sleep on this.

RodneyBaker:
Possible fix:

```
TUINT32 ParsedPliImp::writeHideLineSegmentsTag(HideLineSegmentsTag *tag) {
  assert(m_oChan);

  const double scale = 100000.0;
  const int count    = (int)tag->m_segments.size();

  TINT32 maxValue = count;

  for (const THideLineSegment &seg : tag->m_segments) {
    const TINT32 w0 = (TINT32)(scale * seg.m_w0);
    const TINT32 w1 = (TINT32)(scale * seg.m_w1);

    maxValue = std::max(maxValue, w0);
    maxValue = std::max(maxValue, w1);
  }

  // count, w0 and w1 are all written using the current dynamic width.
  setDynamicTypeBytesNum(0, maxValue);

  const UINT dataBytes = m_currDynamicTypeBytesNum;

  // Payload:
  //   count                  = 1 dynamic integer
  //   each segment:
  //     w0                   = 1 dynamic integer
  //     w1                   = 1 dynamic integer
  //     mode                 = 1 byte
  const UINT tagLength =
      dataBytes + count * (2 * dataBytes + sizeof(UCHAR));

  const TUINT32 offset =
      writeTagHeader((UCHAR)PliTag::HIDE_LINE_SEGMENTS_GOBJ, tagLength);

  writeDynamicData((TINT32)count);

  for (const THideLineSegment &seg : tag->m_segments) {
    writeDynamicData((TINT32)(scale * seg.m_w0));
    writeDynamicData((TINT32)(scale * seg.m_w1));
    *m_oChan << (UCHAR)(seg.m_mode == THideLineMode::Hidden ? 1 : 0);
  }

  return offset;
}
```

The significant change:

```
TINT32 maxValue = count;

for (...) {
  ...
  maxValue = std::max(maxValue, w0);
  maxValue = std::max(maxValue, w1);
}

setDynamicTypeBytesNum(0, maxValue);
```

Whereas the current code says:

`setDynamicTypeBytesNum(0, count);`

konero:
I tested this but there's a few issues.

1. Freehand doesn't seem to do anything.

2. Erasing a line that's has a segment hidden causes it to be unhidden.

hl1</br>


3. Undoing a hide line action doesn't update the viewer, so the cursor needs to be moved to force an update.

4. Titling the view mode `Hide Line: Show Invisible/Hidden Strokes` is verbose, what about just `Show Hidden Strokes`? The terminology is mixed, this is a "Hide Line" tool, either use Line or Stroke, "Hide Stroke Tool", "Show Hidden Strokes", etc.

Ahl76:
@RodneyBaker

A few thoughts:

On conversion: I went through Level > Convert > Convert to Toonz Raster (I did not find the path via right-click). Toonz Raster has no Hide Line, so those segments become regular ink. I don’t think we should deal with that in this PR.

Treating Invisible as zero-alpha could be a later workaround. Matching Invisible / Hidden on Toonz Raster for full compatibility would be a real addition, and it seems to me out of scope here.

The same applies to the fill-related invisibility you mentioned: it seems to me like a different concept. Linking or unifying the two can be discussed later. This tool was meant to stay non-destructive for auto-inbetween. Toonz Raster has no auto-inbetween, so the Eraser already gives the destructive equivalent. Extending similar behavior there in another PR would seem preferable to me.

I would rather measure the real impact of this vector tool first, then see whether Toonz Raster / Raster could benefit from the same architecture. A later pass seems a better place for the opportunities in your PDF.

On the save/reload issue : confirmed, and thank you for the proposed fix. I appliedyours suggestions. Hidden / Invisible segments now seem to persist after Save All and reopen. If you can confirm from your side.

I still cannot reproduce the Save Level crash. If you have a file that triggers it, that would help.

I did add Frame Range in the end. Thank you for the start of your proposal. I used it as a starting point, added the missing interpolations and features, and a few new features :

Shift continues from the last guides; without Shift, the tool returns to the first cell
Ctrl adds another zone on the same range (1-1, 2-2)
numbered pending preview; Normal follows the ink
Lasso (Normal only) and Follow Ink are experimental, both in where they appear and how they work
Stroke correspondence is planned as well. It is related to that request  #3563, but it will come later since it would affect more several tools.

@konero

Freehand seems to work in most cases on my side. Could you try again and make sure the whole target stroke is enclosed by the selection? To be clear, it is meant to behave like the vector Eraser Freehand, not the Toonz Raster one.
I also added Lasso. I’m not sure yet whether it is better as an option or as its own type.

2 and 3. Confirmed and fixed : erasing a stroke should no longer clear its hidden segments, and undo/redo should now refresh the viewer without moving the cursor.

On Hide Line : Show Invisible/Hidden Strokes, agreed it is long. I had the same question, but I ended up with that name so both Invisible and Hidden overlays would be covered. Show Hidden Strokes is shorter, but it may sound as if Invisible is excluded. Vector levels can also hide ink with zero alpha or zero thickness, so I wanted to avoid mixing those cases as well. I am unsure what is best. A more generic name such as Show Hide Line Overlay might work, even if it is a bit vague as well. Hide I find Stroke Tool is accurate, though it may sound a bit more technical. Hide Line Tool feels more familiar, a bit more general-audience. Further opinions on the wording would help.

Thank you for the feedback.

RodneyBaker:
@Ahl76 Thanks much for this latest update.
I scrolled through the code but haven't dived into running it (but will yet tonight)

Disregard that. @vpaintanimation (aka HaibinWu) is such a fan of your feature he asked about it and so...

Looking very good.
I'll try to respond to your last post where appropriate (later) but from a quick skim I'm in full agreement.  Expanding the feature is future work and doesn't even have to be by you.  I'm just adding thoughts as they occur and trying to anticipate user-developer questions before they are asked.

Just reopening a scene and seeing those hidden/invisible line hints:  exciting stuff.
Bug fixed!  (and I haven't been able to reproduce the Save Level Crash... yet)

More later... back to testing. (and this PR back to v1.9 label)

RodneyBaker:
I did get the crash again but it was after first running into an issue where I couldn't save over a previously saved PLI (saved from this same PR's build).

At first I was saving fine (or I thought so given that I hadn't tried to reload the same files yet).

Try 'Save Level As' and 'Save Level' and loading/reloading those same files several times and see if you don't get some strangeness.  I'll try to get some specific steps.

I'm also seeing those same 1kb PLI files refuse to load which suggest that might be why there's a problem if they aren't properly saving.

Added:  I'm able to load those 1kb PLI files in other builds but the PLI file comes in empty.

Ahl76:
Rodney could you try again ?

vpaintanimation:
I suddenly thought of something. Since hidden lines can be marked and matched one-to-one with the corresponding lines in the next keyframe, could OpenToonz’s Fill Tool with frame range enabled work in the same way?

What I mean is: could the Fill Tool in frame range mode also support marking correspondences, such as 1-1, 2-2, 3-3?
IMG_20260825_072730

RodneyBaker:
@Ahl76 A new pass suggests you've fixed the problems I identified. Working great.

I ran into some interesting challenges while drawing (workflow related) and each time I started to think 'ah, this could be problematic... you had a feature that covered the problem.  An example was where lines had been segmented so I couldn't easily tag them differently than they were... oh... there the Lasso.  Done.

Yes, originally I did think the Hide Line Tool might not rise to be worthy of being a full fledge Tool on the Toolbar .  I was wrong.

Something I haven't seen anyone comment on is the 'Follow Ink' option.
Okay... that's nice.

The multi tags for Frame Range (which @vpaintanimation has commented on) is something other the other Tools with Frame Range will be jealous of.

I have questions.  I have thoughts.
At this exact moment I'm mostly just enjoying the process.

Nice work.

vpaintanimation:
Regarding the order of marking, I think there should be a checkbox option. When this option is enabled, each time you select a color style and click the insertion point, it will automatically create a marker.

For example, if you select the red fill style and click on a filled area, it will create Marker 1. Then, if you select the blue style and click another area, it will create Marker 2, and so on.

Ahl76:
Thank you for confirming the crash is fixed.

@RodneyBaker @vpaintanimation Asi said earleir, I plan to cover other tools later, with real stroke correspondences across several of them. The same about extending pairing to other Frame Range tools. On that note, the fill tool frame range option feels broken in the current nightly, but I have not looked into it in depth.

Some Tape Tool issues have been addressed, though there may still be cases that are not fully cleared up.

Thank you for the feedback.

Ahl76:
View toggle is now Hide Line Indicators. Tooltip: Show Invisible/Hidden Strokes.

Invisible / Hidden Mode tooltips: Fill-block and Fill-through. Wording can change if this is still unclear.

Branch is up to date with current master.

PR #6975, opened 2026-07-29
Persist viewer view and room panel layout across sessions

Ahl76:
Goal : Keep viewer zoom, pan, and room panel layout between sessions

**Problem**

When reopening OpenToonz, the viewer often returned to a default zoom and pan, and the size of panels in the room could change on startup. Users lost the view and workspace layout they had set up in the previous session.

**Proposal**

Keep the last viewer view and room panel layout between sessions, and restore them only once the window is ready, so startup no longer resets what the user had arranged.

The viewer comes back with the same zoom, pan, and reference mode as before. Panel sizes in the room are kept as you left them after closing and reopening OpenToonz.

**Notes**

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -238,6 +238,9 @@ public:
   bool isViewerIndicatorEnabled() const {
     return getBoolValue(viewerIndicatorEnabled);
   }
+  bool isRestoreViewerViewFromLastSessionEnabled() const {
+    return getBoolValue(restoreViewerViewFromLastSession);
+  }
 
   // Visualization  tab
   bool getShow0ThickLines() const { return getBoolValue(show0ThickLines); }
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -43,6 +43,7 @@ enum PreferencesItemId {
   showRoomBindButtons,
   displayIn30bit,
   viewerIndicatorEnabled,
+  restoreViewerViewFromLastSession,
 
   //----------
   // Visualization
--- a/toonz/sources/toonz/mainwindow.cpp
+++ b/toonz/sources/toonz/mainwindow.cpp
@@ -58,6 +58,7 @@
 #include <QPushButton>
 #include <QLabel>
 #include <QMessageBox>
+#include <QTimer>
 #ifdef _WIN32
 #include <QtPlatformHeaders/QWindowsWindowFunctions>
 #endif
@@ -418,7 +419,28 @@ void Room::load(const TFilePath &fp, RoomLoadParams &params) {
 
   layout->restoreState(state);
 
+  // Store layout state for deferred re-apply after the first show.
+  // Without this, TMainWindow::resizeEvent → redistribute() recalculates
+  // panel sizes before the window has reached its final geometry.
+  m_pendingLayoutState        = state;
+  m_hasPendingLayoutRestore   = true;
+
   m_initialized = true;
+}
+
+//-----------------------------------------------------------------------------
+
+void Room::showEvent(QShowEvent *event) {
+  TMainWindow::showEvent(event);
+
+  if (m_hasPendingLayoutRestore) {
+    m_hasPendingLayoutRestore = false;
+    DockLayout::State savedState = m_pendingLayoutState;
+    DockLayout *layout           = dockLayout();
+    QTimer::singleShot(0, this, [layout, savedState]() {
+      layout->restoreState(savedState);
+    });
+  }
 }
 
 //=============================================================================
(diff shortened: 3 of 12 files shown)


RodneyBaker:
Overview:  
There are really two features in this PR.

The Room layout element doesn't create layout persistence from scratch.
OpenToonz already saves DockLayout::State into the room .ini. 

What this PR (#6975) does is add a second deferred restore after the Room is first shown here  TMainWindow::resizeEvent() calls redistribute() which can then overwrite the saved panel proportions during startup.

The Viewer element is genuinely a new persistence. 
It writes both 2D m_viewAff matrices and referenceMode into the panel's QSettings.
It then reapplies them with QTimer::singleShot(0) after the Viewer is shown.

Thel strategy is reasonable.
The problems are in defining exactly what constitutes the Viewer state.

This appears to be a bug:  In the Viewer State the restored Reference Mode and its UI can disagree.

The title bar creates three radio style buttons: Camera Stand, 3D View and Camera View.
The proces explicitly initializes Camera Stand as pressed. 
The button group drives SceneViewer::setReferenceMode() when the user selects one.
BUT the new restoration code does this directly:

`if (refMode >= 0) sv->setReferenceMode(refMode);`

It does not update m_referenceModeBs or the corresponding title bar button.

As such in the following scenario:

- Select Camera View
- Close OpenToonz
- Restart OpenToonz
- Viewer internally restores Camera View
- Title bar still visually says Camera Stand View

The button set already has a mechanism for selecting one of its buttons, but the restoration bypasses it.

Potentially:  Create one BaseViewerPanel level operation that synchronizes both:

applyReferenceMode(mode)
    ├─ SceneViewer state
    └─ title-bar button state
 
Then use that operation for persistence restoration.

I'll shift to a different problem next...

RodneyBaker:
This PR saves all six components of each TAffine.
Example:
a11 a12 a13
a21 a22 a23

The process later calls setViewMatrix() to restore it.
That almost reads like zoom/pan persistence but m_viewAff contains more than just zoom and pan.

Rather than explore code here I'll suggest the following as tests:

1. Rotate Viewer 30°
2. Close OpenToonz
3. Reopen
4. Confirm Viewer is still rotated
5. Choose Reset Rotation

Expected:
Viewer returns to zero rotation.

With the current implementation  step 5 likely will not behave correctly because the matrix says 30 degrees while m_rotationAngle says zero degrees.

Recommend making a deliberate choice:  Either persist only actual zoom/pan, stripping rotation/reflection from the saved transform or (preferably) persist a coherent Viewer state including the logical rotation and flip fields along with the affine.

A small ViewState structure would be preferrable over exposing more independent getters.

Next up...

RodneyBaker:
Camera3D doesn't actually preserve its zoom/pan.
This directly contradicts part of the PR's stated behavior.
We'll need to investigate that where a possible test is:

- Switch to 3D View
- Zoom in
- Pan the 3D view
- Rotate the 3D view
- Close OpenToonz
- Restart

I believe the PR can restore `Reference Mode = Camera3D` but not the 3D view the user left behind.

The paths forward would appear to be to explicitly exclude the 3D Camera from this feature set (for now) or persist the complete Camera3D state,

There are some additonal things and some code comments are suspect but I believe that covers the most important.

Ahl76:
Thanks. The follow-up is in.

Reference mode now updates both the Viewer and the title bar buttons when restored. View restore keeps zoom and pan only, no rotation (for now). The 3D view state is saved too. I also added a preference, off by default, so people who want the usual startup view can leave it as is.

RodneyBaker:
@Ahl76 
Assessment and movement to merge:

Future cleanup rather than reasons to hold this PR:
- The persistence implementation is distributed across BaseViewerPanel and SceneViewer.
- A more formal ViewState abstraction can make this cleaner (as noted in original review)
- The blank addition in onSceneSwitched() is harmless noise: neither affects correctness sufficiently to justify another revision.

A test worthwhile before merger:
- [x] Enable the preference: independently exercise Camera Stand/Camera View and 3D View
- [x] Close OpenToonz, reopen: confirm that restored view plus reference mode button agree.
- [x] Rotate/flip a 2D viewer before shutdown: confirm these do not return and zoom/pan does return after restart. 

Passing this test should directly validate the intended boundary of this implementation.

Thanks!

Ahl76:
@RodneyBaker 

I retested with the preference on and had no problems with these cases.

RodneyBaker:
Thank you @Ahl76 for this update persisting viewer and room panels across sessions.
These smaller updates to UI aide us considerably on our way to v2.0
LGTM

JParks123:
this PR has added a delayed restore of saved panel sizes. It can put an older, shorter layout back into a taller window, leaving blank space underneath.

RodneyBaker:
Added a fix to catch the specific case identified by @JParks123 at https://github.com/opentoonz/opentoonz/pull/7124
Thanks!

Ahl76:
@JParks123 @RodneyBaker Thanks for tackling the issues from this PR, and for following through with #7124 and #7130.

PR #6976, opened 2026-07-29
Turn docked panels into tabs

Ahl76:
**_This is an experimental UI/workspace feature intended for nightly testing. Please report tabbed panel docking, layout restore, undocking, and multi monitor issues with platform and workspace details._**

**Description**

Problem

OpenToonz’s interface can fill up quickly. With many panels open, especially on smaller screens, the workspace becomes crowded and could be harder to organize.

**Proposal**

Allow docked panels to be grouped into tabs, so users can save space and arrange their workspace more flexibly.

Drag a panel onto another panel’s title bar to merge them into a tab group. A visual cue appears to show that the panels can be joined. Hold and drag a tab’s title area to pull that panel out of the group again. While reordering tabs, an insertion marker shows where the tab will land, and the new order is applied when you release the mouse. 
**Notes**

This feature was developed with Cursor’s AI models. If this PR suits you, the session layout persistence PR #6975 will need to be merged before this one.

### DIFF
--- a/stuff/config/qss/Blue/Blue.qss
+++ b/stuff/config/qss/Blue/Blue.qss
@@ -12,6 +12,7 @@ QToolBar * {
 }
 :TOONZCOLORS {
   -bg-color: #43464d;
+  -hl-color: #4f7bc7;
   -icon-base-opacity: 0.8;
   -icon-disabled-opacity: 0.3;
   -icon-base-color: #dfdfdf;
--- a/stuff/config/qss/Clay/Clay.qss
+++ b/stuff/config/qss/Clay/Clay.qss
@@ -12,6 +12,7 @@ QToolBar * {
 }
 :TOONZCOLORS {
   -bg-color: #514d4c;
+  -hl-color: #c26350;
   -icon-base-opacity: 0.8;
   -icon-disabled-opacity: 0.3;
   -icon-base-color: #dfdfdf;
--- a/stuff/config/qss/Dark/Dark.qss
+++ b/stuff/config/qss/Dark/Dark.qss
@@ -12,6 +12,7 @@ QToolBar * {
 }
 :TOONZCOLORS {
   -bg-color: #303030;
+  -hl-color: #4f7bc7;
   -icon-base-opacity: 0.8;
   -icon-disabled-opacity: 0.3;
   -icon-base-color: #dfdfdf;
--- a/stuff/config/qss/Default-Green/Default-Green.qss
+++ b/stuff/config/qss/Default-Green/Default-Green.qss
@@ -12,6 +12,7 @@ QToolBar * {
 }
 :TOONZCOLORS {
   -bg-color: #484848;
+  -hl-color: #278d44;
   -icon-base-opacity: 0.8;
   -icon-disabled-opacity: 0.3;
   -icon-base-color: #dfdfdf;
--- a/stuff/config/qss/Default/Default.qss
+++ b/stuff/config/qss/Default/Default.qss
@@ -12,6 +12,7 @@ QToolBar * {
 }
 :TOONZCOLORS {
   -bg-color: #484848;
+  -hl-color: #4f7bc7;
   -icon-base-opacity: 0.8;
   -icon-disabled-opacity: 0.3;
   -icon-base-color: #dfdfdf;
--- a/stuff/config/qss/Default/less/layouts/_mainwindow.less
+++ b/stuff/config/qss/Default/less/layouts/_mainwindow.less
@@ -16,6 +16,7 @@ QToolBar * {
 
 :TOONZCOLORS {
   -bg-color: @bg-color;
+  -hl-color: @hl-color;
 
   // Opacity
   -icon-base-opacity: @icon-base-opacity;
--- a/stuff/config/qss/Light/Light.qss
+++ b/stuff/config/qss/Light/Light.qss
@@ -12,6 +12,7 @@ QToolBar * {
 }
 :TOONZCOLORS {
   -bg-color: #d1d1d1;
+  -hl-color: #7098f0;
   -icon-base-opacity: 0.8;
   -icon-disabled-opacity: 0.3;
   -icon-base-color: black;
--- a/stuff/config/qss/Neutral/Neutral.qss
+++ b/stuff/config/qss/Neutral/Neutral.qss
@@ -12,6 +12,7 @@ QToolBar * {
 }
 :TOONZCOLORS {
   -bg-color: #808080;
+  -hl-color: #dddddd;
   -icon-base-opacity: 0.8;
   -icon-disabled-opacity: 0.3;
   -icon-base-color: black;
--- a/toonz/sources/toonz/pane.cpp
+++ b/toonz/sources/toonz/pane.cpp
@@ -798,11 +798,11 @@ void TPanelTitleBar::paintEvent(QPaintEvent *) {
         dw->windowTitle(), Qt::ElideRight, rect.width() - 50);
 
     painter.setBrush(Qt::NoBrush);
-    // Title text attenuation is reserved for inactive tabs within a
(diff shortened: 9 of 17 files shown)


RodneyBaker:
@Ahl76  I like the concept you have here.
It's moving in the right direction.

I have some surface level concerns about the UI itself but most of that will work itself out.
Instead I'll share this review of the code that gets more deeply into concerns of the code itself.

In the review document, there is a reference to 'Deleting Comments' that in rereading doesn't seem to have any context.  That refers to an earlier pass on the code where I was going to focus on how we could transfer the content of the comments into the code itself (thereby removing all the comments).  I struggled with that a little and started again with more focus on the code itself.  Still, the number one rule of commenting code remains:  don't add comments.  Let the code speak for itself.**

[Review of 6976 Turn Docked Panels into Tabs.pdf](https://github.com/user-attachments/files/30646846/Review.of.6976.Turn.Docked.Panels.into.Tabs.pdf)

Ahl76:
Thanks for the review. I went through the PDF and applied the changes in the latest push. Let me know if anything still needs work.

Note: full layout persistence across sessions still needs #6975.

RodneyBaker:
The architecture is now largely sound.

More info to come but here are the areas of interest:

- Harden the state parser
- Clarify floating target behavior
- Make the remaining tab group capability API typed rather than property string based.

I had some thoughts on these before but don't have any notes to refer to presently.

Regarding tests.  The main one I have to do personally:  After adding tabbed panels, do all the docking behaviors that worked before still work?  I believe they do but just have to confirm.  The following tests are in progress:

| Test | Action | Expected result |
| --- | --- | --- |
| Normal split docking | Drag a panel to left/right/top/bottom docking zones | It docks exactly as before #6976 |
| Create tab group | Drag one normal panel onto another panel's title | Both become tabs; intended panel becomes active |
| Switch tabs | Click each tab repeatedly | Correct panel appears; no geometry jump |
| Reorder tabs | Drag tabs left/right | Order changes only on drop; no flicker or lost panel |
| Pull tab out | Drag a tab vertically/outside the tab bar | Panel becomes floating and follows cursor correctly |
| Redock detached tab | Drag that floating panel back into a normal docking position | Normal split docking still works |
| Collapse group | Make a two-tab group, then remove one tab | Remaining panel becomes a normal panel with its title bar restored |
| Three or more tabs | Add 3–5 panels | Switching/reordering/removing arbitrary tabs remains correct |
| Remove active tab | Make three tabs and pull out the currently active one | A sensible remaining tab becomes active; no blank region |
| Split beside tab group | Create tab group, then dock another panel beside it | Entire tab group remains intact and split layout is correct |
| Resize | Resize separators surrounding a tab group | Tab group and neighboring panels resize normally |
| Maximize | Double-click tab strip / maximize active panel | Panel maximizes correctly |
| Restore maximize | Restore the maximized tabbed panel | Tab bar returns; inactive tabs remain hidden |
| Floating panel | Float a normal panel and move it | Existing floating-panel behavior remains unchanged |
| Toolbar/Command Bar | Try to tab ToolBar, Command Bar, Tool Options | They refuse tab grouping as intended |
| Theme switch | Switch Default/Dark/Light/etc. | Tab highlight/merge indicator follows the theme |
| Room change | Make tab groups, switch rooms, return | Current room state doesn't corrupt or lose panels |
| Save/restore state | Save a layout containing reordered tabs, restore it | Same grouping, order, active tab and geometry return |
| Shutdown | Close OpenToonz with several tab groups | Clean shutdown, no crash |

Something I find useful about this addition is that should a user decide not to use this feature they can continue to operate exactly as they did before.  This strictly meets the definition of optional Shun set forth for adding new features.

RodneyBaker:
When mentioning 'Harden the state parser' that needs clarification so I'll attempt that here.

You've already improved this substantially but perhaps we can go farther.
Specifically we want to make the parser match the grammar it now claims to implement.

`DockLayout::restoreState()` can be more strict about rejecting malformed or internally inconsistent saved layout data before it modifies the live docking layout.


`const int orientation = !tokens[1].toInt();`

This doesn't check whether tokens[1] was a valid integer.

As such, corrupted data like `-1 banana ...` can be interpreted as an orientation when it shoujld be rejected.

Perhaps:

```
bool ok = false;
int savedOrientation = tokens[1].toInt(&ok);

if (!ok || (savedOrientation != 0 && savedOrientation != 1))
  return false;
```

Moving on...
The documented grammar says there is one root:

`state := maximizedIndex rootOrientation region`

What we don't currently have is exactly one complete root region and nothing afterward.
Ask your ai about this and it should clarify.

Next...
The writer creates tab groups thusly:
{ 3 7 9 @1 }

But parseTabGroup() initializes:  int activeIndex = 0;

If a corrupted state contains: { 3 7 9 }
The parser can simply treat panel 3 as active.
That's forgiving, but it contradicts the documented grammar requiring @activeIndex.
A bit further this case doesn't make sense:  { 3 7 9 @0 @2 }
But the current parser can encounter multiple active index tokens and simply let the last one win.

Here's what ChatGPT suggests:  

> Harden restoreState() validation: the new parser is substantially safer, but it should strictly validate the documented hierarchy grammar before replacing the current layout. In particular, validate root orientation, require exactly one complete root region, require exactly one valid @activeIndex per tab group, reject an out-of-range active index, and reject trailing/unconsumed tokens. This should be a relatively small defensive refinement rather than a redesign.

In other words, don't rewrite the parser.
It just needs to be hardened a little more to catch malformed data.

And here's where this actually leads:
This becomes more important because #6975 intends to make room layout state persist between application sessions. A malformed or old layout string could potentially survive restarts instead of existing only briefly in memory.

Hope that make sense.

Ahl76:
Thanks. Requested changes are in the latest push.

FYI, on the test list : I already ran the same checks during development. No major issue on my side. Thanks for the double-check.

RodneyBaker:
For the past scub throughs my main focus has been on the code and as such my views more technical.  

For this (current) pass stepped back from the technical with a view more to the artist/animator and am more interested in workflow and operation.   
Verdict:  I really like what this panel tabs capability brings to the table.

I am eager to see what others think of it.
This particularly from the originally stated problem.

> Problem
> 
> OpenToonz’s interface can fill up quickly. With many panels open, especially on smaller screens, the workspace becomes crowded and could be harder to organize.
> 
> Proposal
> 
> Allow docked panels to be grouped into tabs, so users can save space and arrange their workspace more flexibly.

Verdict:  This PR implements the goal of effectively utilizing workspace while allows the user to organize windows/panels through use of tabs.

Added:  The process of 'tabbing' panels echos that of docking which reenforces things users already work with regularly when organizing the interface.  Nice touch.

Something that has stuck in my mind since first looking at this PR was how the 'tabs' do not look like standard tabs as we might think of them.  This is not necessarily a bad thing as visual tabs might consume unnecessary space and detract from a simplified interface.

I mention this partially in anticipation of users asking this very question; why don't these tabs look like tabs?

This doesn't need an answer for this PR to move forward but I expect the question will be asked. As such it will be good to consider an answer and consider how the tab area might be styled via QSS to resemble classic UI tabs.

I don't see much else operationally to review.  Turning 'Docked panels into Tabs' works exactly 
as we might expect.

Nice work.

Next up for me will be to step back even further to determine what (if anything) we've missed.

RodneyBaker:
@Ahl76 
This PR will presumably need to be updated based on the fix introduced by https://github.com/opentoonz/opentoonz/pull/7106

Ahl76:
@RodneyBaker The branch now includes the #7106 fix. Thanks !

RodneyBaker:
@Ahl76 
Moving forward with this addition to get into into nightlies for more thorough field testing.
I am classifiying this with the other experimental features that are updating the UI which will continutally be defined and redefined all the way into v2.0.

As always, many thanks for your patience.
LGTM

I am classifiying this with the other experimental features that are updating the UI which will continutally be defined and redefined all the way into v2.0.  See added disclaimer in PR description.

PR #6977, opened 2026-07-29
Remember the Style Editor tab between sessions

Ahl76:
**Problem**

After closing OpenToonz or switching rooms, the Style Editor always came back on the Color tab. If you were working on Texture, Vector, or Raster, you had to open that tab again each time.

**Proposal**

Remember the last Style Editor tab and reopen on it after a restart or a room switch. The Style Editor returns to the tab you were using (Color, Texture, Vector, Raster, or Settings).

When Remember Style Editor Tab is on, the Vector page filters (Generated, Trail, and Vector Brush) are remembered as well. When it is off, that persistence is not saved or restored.

**Notes**

### DIFF
--- a/toonz/sources/include/toonz/palettecontroller.h
+++ b/toonz/sources/include/toonz/palettecontroller.h
@@ -25,6 +25,7 @@
 //    Forward declarations
 
 class TPaletteHandle;
+class TPalette;
 
 //=====================================================
 
@@ -63,6 +64,9 @@ public:
   TPaletteHandle *getCurrentPalette() const { return m_currentPalette; }
 
   void setCurrentPalette(TPaletteHandle *paletteHandle);
+
+  // Load cleanup without selecting it.
+  void assignCleanupPalette(TPalette *palette);
 
   // centralized handling of color auto-apply. see StyleEditor.
   // when ColorAutoApply is disabled then changes should made on the ColorSample
--- a/toonz/sources/include/toonz/palettecontroller.h
+++ b/toonz/sources/include/toonz/palettecontroller.h
@@ -50,6 +50,7 @@ class DVAPI PaletteController final : public QObject {
   //!  the last setCurrentPalette() invocation.
   TPixel32 m_colorSample;
   bool m_colorAutoApplyEnabled;
+  bool m_skipCleanupEdit;
 
 public:
   PaletteController();
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -241,6 +241,9 @@ public:
   bool isRestoreViewerViewFromLastSessionEnabled() const {
     return getBoolValue(restoreViewerViewFromLastSession);
   }
+  bool isRestoreStyleEditorTabEnabled() const {
+    return getBoolValue(RestoreStyleEditorTab);
+  }
 
   // Visualization  tab
   bool getShow0ThickLines() const { return getBoolValue(show0ThickLines); }
--- a/toonz/sources/include/toonz/preferencesitemids.h
+++ b/toonz/sources/include/toonz/preferencesitemids.h
@@ -44,6 +44,7 @@ enum PreferencesItemId {
   displayIn30bit,
   viewerIndicatorEnabled,
   restoreViewerViewFromLastSession,
+  RestoreStyleEditorTab,
 
   //----------
   // Visualization
--- a/toonz/sources/include/toonzqt/styleeditor.h
+++ b/toonz/sources/include/toonzqt/styleeditor.h
@@ -920,6 +920,9 @@ class DVAPI StyleEditor final : public QWidget, public SaveLoadQSettings {
   QPushButton *m_textureSearchClear;
   QPushButton *m_vectorsSearchClear;
   QPushButton *m_mypaintSearchClear;
+  QPushButton *m_specialButton     = nullptr;
+  QPushButton *m_customButton      = nullptr;
+  QPushButton *m_vectorBrushButton = nullptr;
 
   TColorStyleP
       m_oldStyle;  //!< A copy of current style \a before the last change.
@@ -931,6 +934,7 @@ class DVAPI StyleEditor final : public QWidget, public SaveLoadQSettings {
   bool m_enabledOnlyFirstTab;
   bool m_enabledFirstAndLastTab;
   bool m_colorPageIsVertical = true;
+  int m_currentPageIndex     = 0;
 
 public:
   StyleEditor(PaletteController *, QWidget *parent = 0);
@@ -1051,6 +1055,9 @@ private:
   QFrame *createVectorPage();
   QFrame *createMyPaintPage();
   void updateTabBar();
+  int tabIndexForPage(int page) const;
+  void rememberPage(int page);
+  void applySavedPage();
 
   void copyEditedStyleToPalette(bool isDragging);
 };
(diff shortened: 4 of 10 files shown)


RodneyBaker:
@Ahl76 There are some issues with this change and I'll try to elaborate more as I can.
There are additional (minor?) issues in the comments where the comments suggest one thing while the code appears to demonstrate something other.

Tests should help identify the most obvious problems.
The core of the proposal to retain Style Editor values between sessions is of course useful and valid.

Added:  I believe the main area of concern is with the restriction override from onStyleSwitched().

Ahl76:
@RodneyBaker 

Thanks for your feedback.

To clarify, what I initially wanted was that when I reopen OpenToonz, the Style Editor would already be on the same tab as in the previous session, for example the MyPaint brushes tab or Settings.

Following the structural issue you pointed out with how tabs are shown at startup, I think a compromise could be to restore the last tab only once a level is created. In that case, at startup with no level, the Style Editor would show cleanup as usual, with the Color and Settings tabs. But I still find that less than ideal.

Otherwise, would a preference be acceptable, or is this a structural problem?

RodneyBaker:
Yes, I do think this is a structural problem but not one that is difficult to address.

Most of my concerns are minor to trivial and explore the potential for users to experience one thing or another.  Importantly for you to know is that on the surface this change to the Style Sheet works as expected.

There are some outliers that need to be locked down so that we don't have to address them later and luckily we have plenty of time to explore those.  

I started taking notes to capture my thoughts on potential areas of concern and lost those notes so will likely start again from scratch.  The direction of those thoughts were for where specific processes dictate what appears and that can conflict with the Style Editor remembering things between sessions.  A downside of publishing many PRs at the same time is that the details can all flow together.  Even as I'm typing this my brain is thinking about another of your PRs related to the Style Editor.

I am not overly concerned about proceeding with this PR because I know you'll address any issues we find.

RodneyBaker:
Recommend folding code commentary into code wherever possible.

Ahl76:
@RodneyBaker

The current changes should address the points you raised. The Style Editor should remember its last tab after a restart or a room switch, without overriding the usual tab restrictions.

There is a preference Remember Style Editor Tab (Interface, on by default). Turning it off should restore the previous behavior when there is no level.

Comments that contradicted the code have been removed.

Could you confirm whether this matches what you had in mind ?

RodneyBaker:
Key findings (which has a preference focus):

**The preference does not actually disable tab persistence.**
`rememberPage()`, `save()`, `load()`, and `applySavedPage()` operate regardless of “Remember Style Editor Tab.” 

With the preference off, reopening a normal level still restores the saved tab; the preference currently affects only no palette and cleanup assignment behavior. 
Recommend:  Either gate persistence with the preference or rename/redefine the option.

**Cleanup palette notifications are suppressed too broadly.**
assignCleanupPalette() blocks every signal from m_currentCleanupPalette while replacing it. CleanupPaletteViewer depends on paletteSwitched()/paletteChanged() to rebuild, so a scene switch can leave that viewer representing the previous cleanup palette. The code should suppress only the automatic palette-selection behavior, not every observer notification.

**The new no-palette mode can expose stale editor contents.**
When no palette exists and the preference is enabled, onStyleSwitched() enables the full tab set. It clears the internal style pointers but does not clear SettingsPage or PlainColorPage, and only some controls are disabled. After switching from an existing level, “No Style Selected” can therefore be accompanied by controls displaying the previous style.

**The branch needs updating.**
Windows, Linux, macOS, and AppVeyor builds pass but the branch is 59 commits behind current master. 

Windows, Linux, macOS, and AppVeyor builds pass.
The clang-format failure is currently unrelated to this PR.  
It stops in fullcolorbrushtool.cpp, which this PR (#6977) does not change.
The workflow compares current base directly against the stale head. 
Updating the branch should remove that false scope and permit a meaningful formatting run.

The lasted fixes correctly addressed the important mapping behavior: forced Color/Settings restrictions no longer overwrite an otherwise valid saved full-tab choice. 

The preference and cleanup signal issues are what need to be resolved.
After updating the branch with corrections we should then test;
- Room switching
- Cleanup mode
- No level state
- Cleanup Palette window

Thanks @Ahl76

Ahl76:
@RodneyBaker

Thank you for the review.

The current changes should address the points you raised hopefully. The branch is also updated with master.

Could you take another look ?

I am not entirely sure about the preference. It is mainly there to keep the previous behavior, in case a workflow needs it, but I struggle to see many such cases.

RodneyBaker:
Interestingly, I think you've fixed everything noted but now the (new) clang-format check is genuinely failing with this PR's code.**

The clang-format check reports violations in `StyleEditor::enable()` at styleeditor.cpp:3793–3796  around the formatting of `flagsChanged` and `rememberTabs`.

Recommend running beautification (aka clang-format) on tht block and push the result.


**I cannot yet confirm if the strictness of the current clang-format check is set appropriately but that appears to be so.  Running clang-format on your end will help confirm this.

For the past few passes at this PR I've mostly been focused on the code.
We... that is to say I or someone testing... need to re-verify the intent of this PR/title simply works.  (Easy enough to do)

Thanks @Ahl76

RodneyBaker:
While waiting for the cmake-format here are the things testers can confirm independently:

- Open a level and select each non Color tab.
- Close the scene or create a new empty scene.
- Browse and click entries inside the restored Texture, Vector and Raster/MyPaint pages.
- Switch rooms while still having no level.
- Create or load a level and confirm the chosen tab restores correctly.

Repeat the above with `Remember Style Editor Tab` option disabled in Preferences.
Report results.

RodneyBaker:
I'm seeing two things worth asking about (not merge blocking).

The three Vector-page filters (Generated, Trail and Vector Brush) are persisted through `vectorParts`. As that goes beyond remembering the Style Editor tab between sessions, the PR description should either mention that 'feature' or that persistence should be removed from the PR. As this appears intentional and the user would surely prefer that persistence... recommend the feature be added to the description.

`StyleChooserPage::paintEvent()` now returns when it has no parent. 
It's not clear if this is intentionally defensive or meant to address a specific problem.

Ahl76:
@RodneyBaker

I re-ran the tester checklist with Remember Style Editor Tab on and off. Room switching, empty scene / no level, creating or loading a level, and Cleanup all looked correct on my side. Each room keeps its own Style Editor tab.

Yes, the vector page filters (Generated, Trail, and Vector Brush) are intentional. The PR description now mentions them. They are saved and restored only when Remember Style Editor Tab is on.

-----------------------------------
StyleChooserPage::paintEvent() returns early when there is no parent, so it does not call parentWidget()->width() on a missing parent. That can happen while a page is rebuilt during a room switch.

RodneyBaker:
@Ahl76 I believe you'll have to rebase this prior to moving forward.

Ahl76:
@RodneyBaker The branch is rebased on current master.

RodneyBaker:
Current assessment:

There appears to be one remaining merge conflict against current master:

toonz/sources/toonz/preferencespopup.cpp
Location: PreferencesPopup::createInterfacePage()
This PR adds insertUI(RestoreStyleEditorTab, lay);
ot-master currently adds the customHelpLink field and tooltip at the same insertion point.

The resolution should be straightforward: retain both additions, likely in this order:

insertUI(showRoomBindButtons, lay);
insertUI(RestoreStyleEditorTab, lay);
insertUI(customHelpLink, lay);
getUI<FileField*>(customHelpLink)
    ->setToolTip(
        tr("Leave blank to use the local OpenToonz documentation index. "
           "To open a PDF at a specific page, append #page=12."));

The PR is currently 28 commits behind master. Seven PR touched files also changed upstream, but the other six merge automatically. No competing behavior or architectural decision is involved.

Ahl76:
@RodneyBaker The branch is updated with current master. Both preferences are kept, in the order you suggested.

PR #6980, opened 2026-07-29
Fix image cache handling for unsupported image types

RodneyBaker:
Prevent unsupported image classes from entering the raster compression path and initialize reference counting safely.

Motivation
A crash with Assistants levels was encountered while testing a potential SVG level type.
Assistants reveal the problem because they are not valid for processing with raster compression.  

A byproduct is faster creation and maintenance of levels whose images no longer enter the compression path.

Co-authored with ChatGPT (GPT-5.6).


RodneyBaker:
Proceeding with this fix as it is largely self evident.

PR #6981, opened 2026-07-30
Style Editor color wheels

Ahl76:
**Proposal**

This PR adds new chromatic wheel styles to the Style Editor , Round and Hexagonal , alongside the classic layout. It gives more visual choice when picking colors.

You can switch the shape from the hamburger menu (Wheel Shape) or by right-clicking the wheel. The chosen style is remembered between sessions. A HiDPI fix is included so the wheel scales correctly when switching shapes. The right-click menu also exposes Bottom Bar and Bind to Current Room.

**Notes**

Depends on the collapsible Style Editor bottom bar PR #6979 (should be merged first).  

### DIFF
--- a/toonz/sources/include/toonzqt/styleeditor.h
+++ b/toonz/sources/include/toonzqt/styleeditor.h
@@ -872,6 +872,8 @@ class DVAPI StyleEditor final : public QWidget, public SaveLoadQSettings {
 
   DVGui::TabBar *m_styleBar;
   QStackedWidget *m_styleChooser;
+  QSplitter *m_mainSplitter;
+  QFrame *m_bottomWidget;
 
   DVGui::StyleSample
       *m_newColor;  //!< New style viewer (lower-right panel side).
@@ -908,6 +910,7 @@ class DVAPI StyleEditor final : public QWidget, public SaveLoadQSettings {
   QAction *m_rgbAction;
   QAction *m_hexAction;
   QAction *m_searchAction;
+  QAction *m_bottomBarAction;
   QActionGroup *m_sliderAppearanceAG;
   QAction *m_hexEditorAction;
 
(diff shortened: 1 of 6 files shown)


RodneyBaker:
@Ahl76 I believe the main obstacle for this moving forward is that the new triangular selector’s barycentric conversion is mathematically incorrect.

For triangle weights representing full-hue color, white, and black, the correct relationship is:

```
wHue   = Saturation × Value
wWhite = (1 − Saturation) × Value
wBlack = 1 − Value

Value      = wHue + wWhite
Saturation = wHue / Value       when Value > 0
```

The current branch instead treats wHue directly as saturation and builds the marker with wHue = saturation.  That produces substantial errors. For example:

```
Requested color: Saturation 100%, Value 50%

Correct triangle position:
    halfway between full hue and black

Current branch position:
    full-hue vertex
```

Conversely, the midpoint between full hue and black is read as 50% saturation and 50% value rather than 100% saturation and 50% value.

This will need to be addressed before merger as it isn't just a visual refinement.

I need to review my assessment for accuracy but that's what I'm currently assessing.

RodneyBaker:
Since we are here, I'll recommend splitting #6981 into:

- HiDPI correction, preferably a small bug-fix PR.
- Additional Style Editor wheel modes, including geometry tests.
- Shared panel/context menu extension (if still desired).

Removing Bottom Bar and room binding dependencies would also allow the wheel work to be rebased directly onto master rather than waiting for #6979

Ahl76:
Following the review feedback, this work is split into separate PRs:
- Round/Hex wheel modes + triangle fix: #6989
- HiDPI resize fix (depends on the modes PR): #6990

A follow-up for the color wheel context menu / Bind to Room can come later if needed.

If anything is missing from this split, please tell me and I will adjust.

Closing this PR in favor of those. Thank you for the review.

PR #6982, opened 2026-07-30
Compact Settings page in the Style Editor

Ahl76:
**Problem**

Some style parameters are hard to read in the Settings tab, especially MyPaint brush options, which take a lot of space. On smaller screens this layout is not very comfortable.

**Proposal**

Add a Compact Settings option in the Style Editor hamburger menu. It switches the Settings page to a denser two-line presentation (label above, controls below) so more parameters fit and stay easier to read. The standard single-row layout remains available, and the choice is remembered between sessions.

**Notes**

### DIFF
--- a/toonz/sources/include/toonzqt/styleeditor.h
+++ b/toonz/sources/include/toonzqt/styleeditor.h
@@ -825,6 +825,13 @@ class SettingsPage final : public QScrollArea {
   bool
       m_updating;  //!< Whether the page is copying style content to its widget,
                    //!  to be displayed.
+  bool m_compactMode;
+
+  void applyCompactMetrics();
+  void rebuildParamLayout();
+  void applyCompactEditorLayout();
+  int paramEditorRow(int p) const { return m_compactMode ? (p * 2 + 1) : p; }
+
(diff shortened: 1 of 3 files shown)


RodneyBaker:
This PR can likely move forward once confirmed it can be implmented independently of other PRs associated with Style Editor update.  That does appear to be the case.

Ahl76:
Yes, it's ready for review when convenient.

RodneyBaker:
@Ahl76 
I think this PR is very close.

Problem:
`configureCompactEditorRow()` can hide both controls of `DVGui::FileField` in compact mode. This affects real FILEPATH parameters in styles such as `TTextureStyle` and `OutlineViewerStyle`.

The correction should be straightforward: apply that numeric field layout helper only to INT and DOUBLE parameters, or explicitly preserve FileField controls.

Ahl76:
I limited the compact row helper to INT and DOUBLE parameters, as requested.

On Texture (Load From File) with Compact Settings on, the path field and browse button look the same before and after this change, so I cannot tell from the UI that something was broken.

PR #6983, opened 2026-07-30
Organize Custom Panel templates in folders

Ahl76:
**Problem**

Recent additions #6596 have brought many Custom Panel templates into OpenToonz. They are still shown as a single flat list, with no way to arrange them.

**Proposal**

This PR adds folder-based organization for Custom Panel templates, in the same spirit as folder layouts already used elsewhere in OpenToonz from disk. Nested folders are supported up to 3 levels. Built-in templates ship in default category folders (Grids, Horizontal Bars, Vertical Bars, Radials); users can also create and arrange their own folders. The Custom Panel editor and the Windows --> Custom Panels menu follow that layout.

A search field is also available to filter templates in the selector.

**Notes**

Depends on #6627 (should be merged first). This branch is based on that PR to avoid conflicts in the custom panel area.


RodneyBaker:
Proceeding with this as part of the experimental feature set introduced in v1.8

PR #6984, opened 2026-07-30
Shift & Trace: extra reference box modes for raster drawings

Ahl76:
**Problem**

On raster drawings, Shift & Trace is less comfortable to use than on vector: the reference box is often too large and does not follow the drawing closely, so moving, rotating, or scaling the previous or following reference is harder.

**Proposal**

This PR brings Shift & Trace on raster closer to the vector workflow, and adds extra reference box modes so the box can follow the drawing more closely (full image, savebox, or drawing content). You can also select the previous or following reference more easily.

**Notes**

### DIFF
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -738,6 +738,7 @@ class ShiftTraceToolOptionBox final : public ToolOptionsBox {
   QFrame *m_prevFrame, *m_afterFrame;
   QRadioButton *m_prevRadioBtn, *m_afterRadioBtn;
   QPushButton *m_resetPrevGhostBtn, *m_resetAfterGhostBtn;
+  QComboBox *m_ghostBBoxCombo;
   TTool *m_tool;
   void resetGhost(int index);
 
@@ -753,6 +754,7 @@ protected slots:
   void onResetAfterGhostBtnPressed();
   void onPrevRadioBtnClicked();
   void onAfterRadioBtnClicked();
+  void onGhostBBoxComboChanged(int index);
   void updateColors();
 };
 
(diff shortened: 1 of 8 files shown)


RodneyBaker:
I haven't yet looked into this as I don't feel I know enough about Shift & Trace to properly assess the change.  I won't let that stop me from digging in but...

Feedback from those that use Shift & Trace is greatly appreciated.

RodneyBaker:
@Ahl76 There are a few findings that should be fixed before merging:

Ctrl-drag can perform the wrong operation when starting on an unselected ghost.
I selected Following Drawing, then Ctrl-dragged from the previous drawing’s center toward the following drawing. Instead of creating the action path, it moved the previous ghost’s pivot. Selecting Previous Drawing first and repeating the identical drag created the path correctly.

The new ghost selection branch returns before the existing Ctrl handling and can retain a pivot, scale, or edge operation. Preserve the Ctrl action path behavior when selecting another ghost. Priority: P2, functional regression. [Relevant code](https://github.com/opentoonz/opentoonz/blob/ecd3424fd8a9d327391243465c4a7a3999818128/toonz/sources/tnztools/shifttracetool.cpp#L623-L679)

Content boxes become stale after undo/redo changes a reference drawing.
I added a stroke outside the previous drawing, returned to the middle drawing, and activated Content mode. Undo removed the stroke but left the enlarged box. Switching modes refreshed it; subsequently redoing the stroke restored artwork outside the cached smaller box.

The cache tracks the level, drawing number, and reference row, but does not account for changed pixels. It needs invalidation when reference images change, including undo/redo. Priority: P2, functional regression. [Cache implementation](https://github.com/opentoonz/opentoonz/blob/ecd3424fd8a9d327391243465c4a7a3999818128/toonz/sources/tnztools/shifttracetool.cpp#L113-L148)

There is also a smaller consistency regression: the new boxes use hard coded red/green, while the toolbar still follows the user’s onion skin colors. I'd say those configured colors should be restored. [Drawing code](https://github.com/opentoonz/opentoonz/blob/ecd3424fd8a9d327391243465c4a7a3999818128/toonz/sources/tnztools/shifttracetool.cpp#L370-L383)

PR #6989, opened 2026-07-31
Round and Hexagonal color wheels in the Style Editor

Ahl76:
Add Round and Hexagonal wheel display modes, selectable from the Style Editor menu under Wheel Shape. Also fix the saturation/value triangle so the marker matches HSV correctly.

Replaces part of #6981 (split after review). 

**Notes**

### DIFF
--- a/toonz/sources/include/toonzqt/styleeditor.h
+++ b/toonz/sources/include/toonzqt/styleeditor.h
@@ -135,10 +135,10 @@ public:
 
 enum CurrentWheel { none, leftWheel, rightTriangle };
 
-enum ColorWheelDisplayMode {
-  ClassicWheelDisplay,
-  CompactNestedWheelDisplay,
-  ExtendedCorrectedWheelDisplay
+enum class ColorWheelDisplayMode {
+  Classic = 0,
+  Round   = 1,
+  Hexagonal = 2
 };
 
 class DVAPI HexagonalColorWheel final : public GLWidgetForHighDpi {
@@ -161,7 +161,7 @@ class DVAPI HexagonalColorWheel final : public GLWidgetForHighDpi {
   float m_innerRadius;
   float m_ringInnerScale = 0.72f;
 
-  ColorWheelDisplayMode m_displayMode = ClassicWheelDisplay;
+  ColorWheelDisplayMode m_displayMode = ColorWheelDisplayMode::Classic;
   CurrentWheel m_currentWheel;
 
   // used for color calibration with 3DLUT
@@ -175,8 +175,8 @@ private:
   void computeHexVertices();
   void computeInnerHexVertices();
   void computeClassicLayout(int w, int h);
-  void computeCompactLayout(int w, int h);
-  void computeExtendedLayout(int w, int h);
+  void computeRoundLayout(int w, int h);
+  void computeHexagonalLayout(int w, int h);
   void updateLayout(int w, int h);
   void drawClassicHexWheel(float v);
   void drawHueRing();
@@ -186,7 +186,7 @@ private:
   void drawCurrentColorMark();
   void drawHueRingBaton(int hue, float innerDist, float outerDist,
                         const QPointF &center);
-  void computeCompactSVTriangle();
+  void computeRoundSVTriangle();
   QPointF svTriangleMarkerPos(float s, float v) const;
   static bool svTriangleBarycentric(const QPointF &p, const QPointF &hueV,
                                     const QPointF &blackV,
--- a/toonz/sources/include/toonzqt/styleeditor.h
+++ b/toonz/sources/include/toonzqt/styleeditor.h
@@ -541,6 +541,7 @@ public:
 
   void updateColorCalibration();
   void setColorWheelDisplayMode(ColorWheelDisplayMode mode);
+  ColorWheelDisplayMode colorWheelDisplayMode() const;
 
 protected:
   void resizeEvent(QResizeEvent *) override;
--- a/toonz/sources/include/toonzqt/styleeditor.h
+++ b/toonz/sources/include/toonzqt/styleeditor.h
@@ -135,6 +135,12 @@ public:
 
 enum CurrentWheel { none, leftWheel, rightTriangle };
 
+enum ColorWheelDisplayMode {
+  ClassicWheelDisplay,
+  CompactNestedWheelDisplay,
+  ExtendedCorrectedWheelDisplay
+};
+
 class DVAPI HexagonalColorWheel final : public GLWidgetForHighDpi {
   Q_OBJECT
 
@@ -147,7 +153,15 @@ class DVAPI HexagonalColorWheel final : public GLWidgetForHighDpi {
   float m_triEdgeLen;
   float m_triHeight;
   QPointF m_wp[7], m_leftp[3];
+  QPointF m_innerWp[7];
+  QPointF m_circleCenter;
+  float m_hexEdgeLen;
+  float m_hexTriHeight;
+  float m_outerRadius;
+  float m_innerRadius;
+  float m_ringInnerScale = 0.72f;
 
+  ColorWheelDisplayMode m_displayMode = ClassicWheelDisplay;
   CurrentWheel m_currentWheel;
 
   // used for color calibration with 3DLUT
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Ok.  Let's see where we are now.

The mathematical defect identified in the original #6981 review is fixed.
(The HSV triangle formula is now correct)

Forward marker conversion:  
wHue   = Saturation × Value
wWhite = (1 − Saturation) × Value
wBlack = 1 − Value

Inverse conversion:
Value      = wHue + wWhite
Saturation = wHue / Value

The marker to HSV round trip appears to be exact.

The refactoring preserves the existing filled hexagonal wheel as Classic and introduces the other modes as options. 

The selected mode is stored in TEnv, so it is a user level preference rather than scene or palette data. 

The menu defaults safely to Classic for users who have no stored value.

PROBLEM 1:
Dragging outside the triangle changes the color to black.

`svTriangleBarycentric()` returns false whenever any barycentric weight falls below -0.02. The caller then does this:

```
if (!svTriangleBarycentric(...)) {
  wHue = wBlack = wWhite = 0.0f;
}
```
Zeroing all three weights produces:

```
Value = wHue + wWhite = 0
Saturation = 0
```

So the result is black.

To test this:
Press inside the saturation value triangle.
Drag slightly beyond an edge.
Within the small -0.02 tolerance, the values are clamped reasonably.
Move a little farther.
The color abruptly jumps to black.

A potential fix would likely:
Calculate barycentric weights for every nondegenerate point.
Clamp negative weights to zero.
Renormalize the three weights.
Convert the normalized weights to HSV.

Conceptually:
```
if (!computeBarycentricWeights(...)) {
  // Degenerate triangle: preserve the current color.
  s = m_color.getValue(eSaturation);
  v = m_color.getValue(eValue);
  return;
}

wHue   = std::max(0.0f, wHue);
wBlack = std::max(0.0f, wBlack);
wWhite = std::max(0.0f, wWhite);

const float sum = wHue + wBlack + wWhite;
wHue   /= sum;
wBlack /= sum;
wWhite /= sum;
```

Note:  There is a problem with HiDPI displays this PR doesn't cover BUT... that is covered in one of the other split PRs.  As such if no other problem is found the plan would be to merge #6989 and #6990 consecutively.  No significant delay between merges.

PROBLEM 2: 
The global preference and individual Style Editor panes can potentially disagree

Probabable solution:  Always apply the selected mode to the current wheel, even when the global value is already equal.

Example:

```
void StyleEditor::onColorWheelDisplayModeSelected(QAction *action) {
  bool ok       = false;
  int displayId = action->data().toInt(&ok);
  if (!ok) return;

  const auto mode = normalizedColorWheelMode(displayId);
  StyleEditorColorWheelDisplayMode = static_cast<int>(mode);
  m_plainColorPage->setColorWheelDisplayMode(mode);
```
}

Also synchronize each pane from `TEnv` in `StyleEditor::showEvent()`:

```
m_plainColorPage->setColorWheelDisplayMode(
    normalizedColorWheelMode(
        static_cast<int>(StyleEditorColorWheelDisplayMode)));
```
That should cover normal room switching but may not cover multiple Style Editors open at the same time.

Thanks for your patience.
This is considerably more complicated than I thought it would be.
If you have any quesiton please let us know.

Ahl76:
@RodneyBaker Thanks for the detailed review.

I pushed two follow-up commits on this PR :

1. Dragging outside the saturation/value triangle should no longer jump to black. Outside points are clamped and renormalized; the current color is kept only if the triangle is degenerate.

2. Wheel Shape should always be applied to the current Style Editor pane, even when TEnv already stores that mode, and the pane resyncs from TEnv on show.

The Wheel Shape menu now also checks the shape this pane actually shows, so the label matches the wheel when several Style Editors disagree.

Please tell me if I missed something.

RodneyBaker:
@Ahl76 Thanks for your quick turnaround.

I have been considering how we might best move forward with this process of ai assisted development especially given the amount of code ai can quickly generate.  Your current PRs are helping in this regard.  

As it takes considerable time to review any change to Opentoonz source code,
As ai assisted code can produce a lot of code in a short amount of time,
In light of direction from @shun-iwasawa regarding pull requests using assistance from ai (REF:  #6937), I submit the following:

Given my role in reviewing PRs:
A first step of any PR submitted moving forward will likely be an ai-assisted review.
This to catch any obvious issues before actually accepting the PR and allowing the change to move forward.

Initial review will likely be in the form of a PDF document capturing the ai assisted review of the PR.  This document will be directive in nature outlining required changes for the PR to be accepted.  The document will also serve as a record of review as we need to make sure the review itself is as accurate as possible and doesn't direct inappropriate change.

Aside:  I do realize the PDF format may not be the ideal format for ai to ai communication but human review and understanding of the code is a stated goal of this process.  We aren't just throwing code at the wall here.  As W. Edwards Deming reported said, "Can we not learn?"

A standard way around this particular process (which has its own share of challenges) is to refine the code change in the dedicated Opentoonz fork while conducting through tests and internal review followed by an invitation to consider the adoption of that code into Opentoonz master.  No code is submitted to Opentoonz master throughout this process and changes occur directly in the fork.  This approach also presumptively reduces cost.

Stand by for a test of these thoughts.
@Ahl76 As we want to help the ai assisted developer in this process get to their goal and have the best possible experience, your feedback on this is appreciated.

Atttached is an example of what an AI assisted review as first pass in this process might look like:
[AI_Assisted_Review_PR6991_Right_Click_Wheel_Shape.pdf](https://github.com/user-attachments/files/30588729/AI_Assisted_Review_PR6991_Right_Click_Wheel_Shape.pdf)

(no human involved in that initial review outside of the 'review this' style prompt)

The attached PDF is for example only.  I will upload this ai review to the appropriate PR.

Ahl76:
@RodneyBaker Following up on the review / naming addendum:

Just pushed

- Renamed the wheel mode enum and helpers to match the UI: Classic / Round / Hexagonal (replacing the CompactNested / ExtendedCorrected development names).

Already on this PR from the earlier follow-ups

- SV triangle math fix (marker matches HSV).
- Dragging outside the triangle no longer jumps to black.
- Wheel Shape always applies to the current Style Editor pane (even when TEnv already matches), resyncs from TEnv on show, and the menu checkmark reflects this pane’s wheel when several Style Editors disagree.

RodneyBaker:
@Ahl76 Thanks for the quick response to suggested corrections.

There is a lot to juggle in these changes so I appreciate your patience.
Waiting in the wings is very like other users who have specific ideas in mind for what the Style Editor Color Wheels should look like and this starts to look like a formidable mountain to climb.

The good news is that is that the timing is right to take on this challenge.

The primary feedback I've read hasn't suggested any specific problems but does suggest concern for the change.  I know you are open to adjusting so I don't see that as an obstacle.

The standard concerns of ai assisted development remain valid and I believe we are handling that as best as we can.  Opentoonz is uniquely positioned to take advantage of that.

I did invite you over to Opentoonz-Dev but not sure if you've seen that.

Regarding these current PRs... we'll keep pluggin away at them as best we can.
Thanks for leaning forward.

A personal request once the dust settles would be to add a square 'color wheel' as an option.
That and we might want to look at the long closed Advanced Color Selector PR which was a separate window and determine what (if anything) from that feature set might be brought over into the Style Editor.  Also, anyone know a Color Scientist that might champion this cause and draft a plan to take this color management to the next level?  ;)

Ahl76:
@RodneyBaker Thanks ! Much appreciated. Of course, further refinements remain welcome once the current stack lands. Noted regarding the directions you mentioned, those look like solid avenues to explore, and the Advanced Color Selector was a good reference. And yes, I accepted the Opentoonz-Dev invite earlier today, thank you.

JParks123:
> Add Round and Hexagonal wheel display modes, selectable from the Style Editor menu under Wheel Shape. Also fix the saturation/value triangle so the marker matches HSV correctly.
> 
> Replaces part of #6981 (split after review).
> 
>
> 
>
Hello @Ahl76  I am a big fan of the stuff your doing but I have to be honest I think these color wheels aren't as good as the one's already in the software, have you seen the one in the original Toonz software it might be in the OpenToonz code somewhere that one to me is much better as well. I really like what you are doing with the Brush presets and the Tool properties panels, maybe focus on getting those to the finish line with other stuff you are working on but these color wheels are not very good and I hate to admit it, this is just my opinion.
image

RodneyBaker:
An important consideration for changes to the Style Editor is going to be getting input from @shun-iwasawa and the studios that use this on a daily basis.

My take is that while adding new options is useful and even necessary it probably should not be at the expense of removing the old workflow.

How might this change?
v2.0 will be surely be a breaking change in many ways and a completely new Style Editor with different look and feel workflows would signal that release as a major change.  Those not wishing to move forward at the time would stay on v1.9 or whatever the previous version to v2.0 is while others would update and settle in to a new normal.

I'm tagging this PR as a v2.0 Candidate with the thought that parts of the current Style Editor update can be cherrypicked and incorporated into the Style Editor now versus later.  Changes that remove current functionality will need more discussion.

Ahl76:
@JParks123 Thanks for the feedback. Classic is still the existing OpenToonz wheel. Hexagonal is close to Classic in practice, and Round is a fairly common layout elsewhere, I think.

This approach may not be the best fit as it stands.

@RodneyBaker Thank you for the additional context. I aim to keep existing workflows intact as much as possible, though unintended side effects can still occur. Further refinement may be needed before another review.

PR #6990, opened 2026-07-31
Fix color wheel size when switching wheel shapes

Ahl76:
**Problem**

When switching the color wheel shape, the wheel could appear oversized and cropped on HiDPI displays.

**Proposal**

Resize the wheel with the correct logical size so the device pixel ratio is applied only once.

Replaces part of https://github.com/opentoonz/opentoonz/pull/6981 (split after review).

Depends on #6989

**Notes**

This fix was developed with Cursor’s AI models.


Ahl76:
Rebased onto current #6989 . HiDPI-only commit on top.

PR #6999, opened 2026-08-04
Ci/fedora werror return type

Chantology:
The Fedora job added in #6986 builds with the project's own flags, which set no `-Werror`. A function that falls off the end without returning a value is therefore only a *warning* in CI, and can reach a release: #6940 is exactly that, found by an RPM packager rather than by us, because distro `%optflags` turn `-Wreturn-type` into a hard error.

This passes `-Werror=return-type` in the Fedora job so the same class of bug fails in a PR instead.

## Scope

Deliberately limited to the Fedora job and to this one warning. A missing return is a genuine bug in every case, never a style preference, so it can't produce argued-about failures. Keeping it in CI rather than in `CMakeLists.txt` avoids imposing `-Werror` on contributors' local builds and on older or newer compilers that may diagnose differently.

## Verification

- The flag survives the `set(CMAKE_CXX_FLAGS "${CXX_WARNINGS} ${CMAKE_CXX_FLAGS}")` assignment in `toonz/sources/CMakeLists.txt` and reaches every compile rule — confirmed by configuring locally with the exact CI command line and checking the generated `build.ninja`: 1288 compile rules carry `-Werror=return-type`, appended after the existing `CXX_WARNINGS`.
- Workflow YAML parses and the `Build` step's line continuation survives the block scalar.

## Note on the base

This branch is based on #6998, so it currently shows that commit too. Without the #6940 fix, the guard would stop the build at `stylepickertool.cpp:324` and reveal nothing about the rest of the tree; with it, a green run here is a full-tree confirmation that no other missing return exists. Happy to rebase onto master once #6998 is merged so this is a single commit.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Chantology:
@RodneyBaker is here anything missing?

RodneyBaker:
This is such a little change it shouldn't be necessary but to close the loop do a rebase and then after final check... good to merge.

The only thing holding this PR back was preferred order of PRs to be merged.
All prerequisites are now merged.
Will double check to make sure all is in order.

RodneyBaker:
Since this branch is 40 or so commits behind master, I believe rebase is still needed for this PR.

Todo:
- Rebase onto current master
- Confirm the diff remains the single Fedora-workflow change
- Let fresh CI pass (especially Fedora)
- Merge (before any other change on master)

(The “Note on the base” in the PR description is stale and can be removed)

PR #7002, opened 2026-08-06
Add load and save options to Additional Style Sheet

RodneyBaker:
Adds Load... and Save... controls to the Additional Style Sheet editor.

The new controls import and export text based style sheet files (CSS, QSS, TXT and .theme extensions) without changing the existing OK, Apply, or Close behavior.

This allows users to load/reload, edit and save Additional Style Sheet data directly from the dialogue.  Recommend using the .theme extension to suggest QSS not generated with LESS.

Ported from OpenAnimationLibrary/opentoonz-dev#60.


RodneyBaker:
Pressing forward with this as it is self explanatory.
@shun-iwasawa If changes are required you know I'm good for it. :)

PR #7003, opened 2026-08-06
Modify Transparency Check + Inks Only Behavior

shun-iwasawa:
This PR is related to #6329.

Regarding the behavior of Transparency Check, when it is simultaneously enabled with “Inks Only,” the pixel correction implemented in #6329 will not be performed. Instead, just like in previous versions, all semi-transparent areas of the inks will be visually emphasized.

This feature is necessary for detecting and removing faint ink noise that occurs when scanning and cleaning up hand-drawn animations on paper.

Since the change in #6329 was made to make it easier to find areas where paint was missed, I believe there will be no negative impact from the behavior changing when used in conjunction with the “Inks Only” option.

### DIFF
--- a/toonz/sources/toonzlib/stagevisitor.cpp
+++ b/toonz/sources/toonzlib/stagevisitor.cpp
@@ -595,7 +595,8 @@ void RasterPainter::flushRasterImages() {
           std::set<int> autoPaints;
           if (autocloseSettings.m_ignoreAPInks) {
             for (int styleIdx = 0; styleIdx < plt->getStyleCount(); ++styleIdx)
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Thanks @shun-iwasawa for the modification and the description of how it improves Transparency Checks.  The specific requirement this addresses is particularly useful from a workflow perspective.

Posting intent to merge although the MacOS check's reporting failed.

RodneyBaker:
Pressing forward with this PR while noting Mac Mojave check is problematic.
Thanks @shun-iwasawa

PR #7006, opened 2026-08-08
Harden autosave save path

RodneyBaker:
## Goals

Harden the existing synchronous autosave path against crashes, lockups, and false save success without changing autosave scheduling or redesigning persistence.  Address numerous complaints about loss of data during use of autosave.

## Highest value changes

* Skip scene icon rendering during autosave, removing render/OpenGL work from the timer driven save path.
* Replace GUI thread save and Undo/Redo spin waits with non blocking guards.
* Treat scene write exceptions as failures so unsaved work remains dirty.
* Propagate level, palette, sound, and resource failures through Save All; partial saves remain dirty and retryable.
* Replace a completed scene temporary without first deleting the current scene file.
* Keep autosave non modal; untitled scenes are skipped rather than opening Save Scene As.

This change utilized ChatGPT 5.6 for research and processing and was first tested through [[OpenAnimationLibrary/opentoonz-dev#65](https://github.com/OpenAnimationLibrary/opentoonz-dev/pull/65)], where it was successfully built, tested, and merged before being proposed upstream.

## For Future Follow up

* Transactional writes for levels, palettes, sound, and other resources.
* Complete state restoration after exceptions and stale temporary file recovery.
* Application wide safe save points for playback, selections, schematic edits, and scene switching.
* Immutable recovery snapshots plus fault injection and stress tests.
* Refinement of Backup system

### DIFF
--- a/toonz/sources/common/tsystem/tsystem.cpp
+++ b/toonz/sources/common/tsystem/tsystem.cpp
@@ -4,6 +4,7 @@
 
 #include <set>
 #include <atomic>
+#include <cstdio>
 #include "tfilepath_io.h"
 #include "tconvert.h"
 
@@ -28,6 +29,7 @@
 #include <QHostInfo>
 
 #ifdef _WIN32
+#include <windows.h>
 #include <shlobj.h>
 #include <shellapi.h>
 #include <winnt.h>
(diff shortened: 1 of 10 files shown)


RodneyBaker:
The current risk of autosave related data loss is the primary driver of this update.
Future work should continue to refine and extend the autosave process.

Note to users:  Use of autosave is not considered a reliable means of safeguarding data and should be used in addition to regular saving with emphasis on using Save All unless there are specific things that should not be saved or updated.

Chatgpt 5.6 was used in research, processing and testing.

PR #7007, opened 2026-08-08
Add ZIP option to new project and standalone scene export

RodneyBaker:
Summary
Adds an optional, (intentionally) non destructive ZIP packaging step to File > Export > Export Scene for both Create New Project and Export as Standalone Scene.

- Places the Create ZIP archive of exported folder option in the shared export dialog
- Enables it for Create New Project and Export as Standalone Scene
- Disables it for Choose Existing Project so unrelated existing project contents are not archived
- Keeps every exported directory after ZIP creation
- Includes the exported project or scene folder itself as the archive root
- Reports ZIP creation errors without undoing or deleting the completed export

After all selected scenes and resources export successfully the complete newly created project directory is written to an adjacent <project-name>.zip.

Export as Standalone Scene
Each successfully exported standalone scene folder is written to an adjacent <scene-folder>.zip.

Choose Existing Project
The ZIP option is disabled when Chosing Existing Project. Existing projects may contain unrelated files that should not be included automatically.

Implementation
The ZIP writer creates standards compatible archives using stored entries and streams files in chunks, avoiding an additional archive dependency. ZIP64 is not included, so individual files and archives remain subject to the classic ZIP 4 GB range.

Background
This advances the request to consolidate exported resources into a portable ZIP package:
https://github.com/opentoonz/opentoonz/discussions/6456

Testing
Build on Windows, macOS, and Linux
Create a new project with ZIP disabled and verify existing behavior is unchanged
Create a new project with ZIP enabled and verify:
the complete project folder remains
an adjacent <project-name>.zip is created
the archive contains the project folder as its root
all selected scenes and project resource folders are present
Export a standalone scene with ZIP enabled and verify existing adjacent scene folder ZIP behavior
Switch to Choose Existing Project and verify the ZIP option is disabled
Extract both archive types and open the exported scene/project
Verify nested folders, Unicode filenames, empty directories, and missing/unreadable file error handling

Implements #373

### DIFF
--- a/toonz/sources/toonz/exportscenepopup.cpp
+++ b/toonz/sources/toonz/exportscenepopup.cpp
@@ -32,6 +32,11 @@
 #include <QApplication>
 #include <QMainWindow>
 #include <QStandardPaths>
+#include <QCheckBox>
+#include <QDirIterator>
+#include <QFileInfo>
+
+#include "simplezipwriter.h"
 
 #include <vector>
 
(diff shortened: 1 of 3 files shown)


RodneyBaker:
Merging this as not effecting anything outside the export scene process.
Follow up post morten should confirm everything works as desired cross platform.

Future work:  While simple zip utlity (without compression) can be useful, investigate how large productions can benefit from use of zstd (or similar compression technology).  REF:  https://facebook.github.io/zstd/

(Review how EXR 3.5 and other forward leaning efforts are adding support for zstd)

PR #7008, opened 2026-08-08
fix(stopmotion): handle JPEG load failures safely

Chantology:

## The bug

`JpgConverter::loadJpg()` dereferences a null `FILE*` whenever `fopen()` fails. The guard

```cpp
if (success && fseek(jpegFile, 0, SEEK_END) < 0 ||
    ((size = ftell(jpegFile)) < 0) || fseek(jpegFile, 0, SEEK_SET) < 0)
```

parses as

```cpp
(success && fseek(...) < 0) || (ftell(...) < 0) || (fseek(...) < 0)
```

so `success` short-circuits only the **first** operand. With `jpegFile == NULL`, the `ftell()` and second `fseek()` still execute and the process dies.

Reachable from the stop motion live view (`stopmotion.cpp` calls `loadJpg` on captured frame paths) whenever the file is missing or unreadable.

## Why the fix is more than the parentheses

Correcting only the parenthesisation would **move** the crash, because the rest of the function assumes the happy path:

| Code | Problem on a failure path |
|---|---|
| `fclose(jpegFile)` | unconditional → `fclose(NULL)` after a failed `fopen` |
| `tjFree(jpegBuf)` | `jpegBuf` is uninitialised unless its allocation ran |
| `tjDestroy(tjInstance)` | `tjInstance` is uninitialised unless init ran |
| `image = TRaster32P(width, height)` | `width`/`height` uninitialised, and this runs even when `success == false` |
| `memcpy(rawData, imgBuf, …)` | `imgBuf` may be NULL |

So this initialises the pointers and sizes, guards the cleanup calls, and only builds the raster once decoding actually succeeded.

## How it was found

Building with `-Wparentheses`, which reports exactly this condition:

```
warning: suggest parentheses around ‘&&’ within ‘||’ [-Wparentheses]
```

The build currently enables only `-Wmissing-declarations`, `-Wundef` and `-Wwrite-strings`, so it was never surfaced. (A follow-up enabling `-Wparentheses` is worth considering — I checked the whole tree and the only other 14 hits are benign unparenthesised bit-field writes and intentional `&&`/`||` nesting, so the flag is cheap to turn on. This same class of mistake also caused the crash fixed in #7000.)

## Verification

A standalone program reproducing the original control flow:

```
fopen failed, success=0, f=(nil)
about to evaluate the original condition...
Segmentation fault (core dumped)   -> exit 139
```

and the corrected form:

```
handled cleanly, success=0        -> exit 0
```

The tree also builds clean with the change.

**Limitation, stated plainly:** I could not unit test `loadJpg` itself — it is compiled into the `OpenToonz` executable and the tree has no wired-up test harness (`ttest.cpp` exists but is not in CI). The verification above demonstrates the mechanism, not the function. A manual check via stop motion live view with a missing frame file would confirm it end to end.

- **Approach / intent given to the model:** rather than auditing code by hand, enable a compiler warning across the whole tree and investigate what it reports. All 15 hits were classified; 14 were confirmed benign by checking operator precedence empirically, and this one was traced to a null dereference and reproduced.

### DIFF
--- a/toonz/sources/stopmotion/jpgconverter.cpp
+++ b/toonz/sources/stopmotion/jpgconverter.cpp
@@ -120,29 +120,32 @@ void JpgConverter::saveJpg(TRaster32P image, TFilePath path) {
 //-----------------------------------------------------------------------------
 
 bool JpgConverter::loadJpg(TFilePath path, TRaster32P& image) {
-  long size;
-  int inSubsamp, inColorspace, width, height;
+  long size = 0;
+  int inSubsamp, inColorspace, width = 0, height = 0;
   unsigned long jpegSize;
-  unsigned char* jpegBuf;
-  FILE* jpegFile;
-  QString qPath      = path.getQString();
-  QByteArray ba      = qPath.toLocal8Bit();
-  const char* c_path = ba.data();
-  bool success       = true;
-  tjhandle tjInstance;
+  unsigned char* jpegBuf = NULL;
+  FILE* jpegFile         = NULL;
+  QString qPath          = path.getQString();
+  QByteArray ba          = qPath.toLocal8Bit();
+  const char* c_path     = ba.data();
+  bool success           = true;
+  tjhandle tjInstance    = NULL;
 
   /* Read the JPEG file into memory. */
   if ((jpegFile = fopen(c_path, "rb")) == NULL) success = false;
-  if (success && fseek(jpegFile, 0, SEEK_END) < 0 ||
-      ((size = ftell(jpegFile)) < 0) || fseek(jpegFile, 0, SEEK_SET) < 0)
+  if (success &&
+      (fseek(jpegFile, 0, SEEK_END) < 0 || (size = ftell(jpegFile)) < 0 ||
+       fseek(jpegFile, 0, SEEK_SET) < 0))
     success = false;
   if (success && size == 0) success = false;
   jpegSize = (unsigned long)size;
   if (success && (jpegBuf = (unsigned char*)tjAlloc(jpegSize)) == NULL)
     success = false;
   if (success && fread(jpegBuf, jpegSize, 1, jpegFile) < 1) success = false;
-  fclose(jpegFile);
-  jpegFile = NULL;
+  if (jpegFile) {
+    fclose(jpegFile);
+    jpegFile = NULL;
+  }
 
   if (success && (tjInstance = tjInitDecompress()) == NULL) success = false;
 
@@ -161,19 +164,28 @@ bool JpgConverter::loadJpg(TFilePath path, TRaster32P& image) {
   if (success && tjDecompress2(tjInstance, jpegBuf, jpegSize, imgBuf, width, 0,
                                height, pixelFormat, flags) < 0)
     success = false;
-  tjFree(jpegBuf);
-  jpegBuf = NULL;
-  tjDestroy(tjInstance);
-  tjInstance = NULL;
+  if (jpegBuf) {
+    tjFree(jpegBuf);
+    jpegBuf = NULL;
+  }
+  if (tjInstance) {
+    tjDestroy(tjInstance);
+    tjInstance = NULL;
+  }
 
-  image = TRaster32P(width, height);
-  image->lock();
-  uchar* rawData = image->getRawData();
-  memcpy(rawData, imgBuf, width * height * tjPixelSize[pixelFormat]);
-  image->unlock();
+  // width/height and imgBuf are only meaningful once decoding succeeded
+  if (success) {
+    image = TRaster32P(width, height);
+    image->lock();
+    uchar* rawData = image->getRawData();
+    memcpy(rawData, imgBuf, width * height * tjPixelSize[pixelFormat]);
+    image->unlock();
+  }
 
-  tjFree(imgBuf);
-  imgBuf = NULL;
+  if (imgBuf) {
+    tjFree(imgBuf);
+    imgBuf = NULL;
+  }
 
   return success;
 }
--- a/toonz/sources/stopmotion/stopmotion.cpp
+++ b/toonz/sources/stopmotion/stopmotion.cpp
@@ -366,8 +366,12 @@ bool StopMotion::buildLiveViewMap(TXshSimpleLevel *sl) {
     int frameNumber = id.getNumber();
     if (TSystem::doesExistFileOrLevel(liveViewFp.withFrame(frameNumber))) {
       TRaster32P image;
-      JpgConverter::loadJpg(liveViewFp.withFrame(frameNumber), image);
-      m_liveViewImageMap.insert(std::pair<int, TRaster32P>(frameNumber, image));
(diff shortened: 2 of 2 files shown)


RodneyBaker:
Interesting.  I had thought this addressed in previous change addressing shortfalls with the JPG format.

All looks good here short of additional testing/confirmation.

Outside this the scope of this PR:
`inSubsamp/inColorspace` appears to be left uninitialized but that's preexisting and not touched by this PR.  It's not a regression.  Just not fully addressed.

I am reminded by the code spacing of this PR that we need to find a way to run beautification on PRs... automatically (?) and pariticularly at the ai level to prevent unnecessary changes back and forth for spacing (?). Actually, I persontally need to understand this process of beautifcation relative to Opentoonz source code better.  All of this with the understanding we should run such only on the PR/changed content to avoid expanding the diff to unrelated code and potentially the entire source code base.

Stop motion animators!  We are in need of a few testers.

RodneyBaker:
In interest of seeing this PR move forward...

The good news:  My earlier concern about `inSubsamp` and `inColorspace` in no longer significant.  (I can't remember if I shared that concern)

New concerns:
`loadJpg() `itself is now failure safe but `StopMotion::buildLiveViewMap()` ignores its returned bool and inserts the output raster into the cache anyway. 

On a failed read/decode that raster is null; later `loadLiveViewImage() ` treats any cached entry as success, and `SceneViewer::onNewStopMotionImageReady() ` dereferences `m_lineUpImage` to clone it.  What's that all mean?  That process creates a remaining route to a null raster crash.

```
if (success && fseek(jpegFile, 0, SEEK_END) < 0 ||
    ((size = ftell(jpegFile)) < 0) || fseek(jpegFile, 0, SEEK_SET) < 0)
```
Here a failed `fopen()` can still lead to `ftell(NULL)/fseek(NULL, ...)`.



There are two callers of loadJpg() in stopmotion.cpp, and they handle failure differently.

`loadLiveViewImage() ` does it correctly.

But `buildLiveViewMap()` currently does this:

```
TRaster32P image;
JpgConverter::loadJpg(liveViewFp.withFrame(frameNumber), image);
m_liveViewImageMap.insert(
    std::pair<int, TRaster32P>(frameNumber, image));
```
Which ignores the failure return value.

With this PR (#7008), a bad/unreadable JPEG now cleanly returns false and leaves that local image null. However, `buildLiveViewMap() ` still caches that null image.

Later,  `loadLiveViewImage() `considers the presence of ANY cache entry a success.
A cached null raster can therefore still be dereferenced.

Perhaps a tiny companion fix might help:

```
for (TFrameId id : fids) {
  int frameNumber = id.getNumber();
  if (TSystem::doesExistFileOrLevel(liveViewFp.withFrame(frameNumber))) {
    TRaster32P image;
    if (JpgConverter::loadJpg(liveViewFp.withFrame(frameNumber), image)) {
      m_liveViewImageMap.insert(
          std::pair<int, TRaster32P>(frameNumber, image));
    }
  }
}
```

Or a slight more defensive:  i`f (JpgConverter::loadJpg(liveViewFp.withFrame(frameNumber), image) && image)`

The above reads like a lot but is actually a small change to `buildLiveViewMap()`.

By way of moving way beyond my compentency I had ChatGPT 5.6 draft what success in the code for the entire file might look like.  The PDF also includes the important secondary stopmotion fix in stopmotion.cpp that accompanies the draft fix suggestion in jpgconverter.cpp.

[PR7008_suggested_final_code.pdf](https://github.com/user-attachments/files/31204182/PR7008_suggested_final_code.pdf)

Chantology:
Thank you!  I traced the whole chain in the code and it holds up:

- `buildLiveViewMap()` caches the raster regardless of the return value (stopmotion.cpp:369)
- `loadLiveViewImage()` returns `true` on any cache hit, without checking the entry (stopmotion.cpp:1101-1104)
- that sets `m_hasLineUpImage = true` with a null `m_lineUpImage` (stopmotion.cpp:1024)
- `onNewStopMotionImageReady()` then calls `->clone()` on it (sceneviewer.cpp:1249), and `TSmartPointerT::operator->` returns the raw pointer with no null check

So it's a real route, and `buildLiveViewMap()` was the only one of the two callers not checking. Companion fix applied.

**One addition:** the same state reaches a second deref you didn't mention. In `saveXSheetFrame()`, `if (m_hasLineUpImage) JpgConverter::saveJpg(m_lineUpImage, liveViewFile)` (stopmotion.cpp:1643) and `saveJpg()` starts with `image->getLx()`. So this fix closes two paths, not one: the draw path and the capture path.

**On the two variants:** I went with the plain `if (loadJpg(...))` rather than `&& image`. `TRasterPT::create()` is `new TRasterT<T>(lx, ly)` (traster.h:398), which is non-null or throws, and `loadJpg()` only assigns `image` inside its `if (success)` block, so a `true` return already implies non-null and `&& image` would be unreachable. Happy to add it anyway if you'd rather have the belt-and-braces version at a crash site.

**On the PDF:** I diffed its `jpgconverter.cpp` listing against the PR head identical, so its own conclusion that no further change is needed there is right, and I've left that file untouched. I deliberately didn't paste the listing in wholesale: it ends with a trailing newline the current file lacks, and re-applying a whole file to change nothing invites accidental edits.

I also left the pre-existing `clang-format` drift in stopmotion.cpp alone (lines ~1366, ~1408, ~2221 want reflowing) — that's the diff expansion into unrelated code you flagged last time. My changed lines are format-clean against `toonz/sources/.clang-format`.

Out of scope but noting it: `m_lineUpImage = m_liveViewImage; m_hasLineUpImage = true;` at lines 1820/1836/1857 has no null check either. That's the live-view capture path, unrelated to JPEG loading — separate PR if it's worth hardening.

RodneyBaker:
> Out of scope but noting it: m_lineUpImage = m_liveViewImage; m_hasLineUpImage = true; at lines 1820/1836/1857 has no null check either. That's the live-view capture path, unrelated to JPEG loading — separate PR if it's worth hardening.

Interestingly, a null check isn't always needed.
A lack of such might actually hint that the programmer knows full well what the code is doing.

Take for intance (and just as a random example) if a programmer where to say:
"m_liveViewImage cannot be null here because reaching this function implies Live View has already supplied an image."

The area where a null check is important are those where the value can be null.
Hope that makes sense.

Should that be hardened?  Not sure.
I expect most ai will think it should be but best practices of the past suggest it is case by case.
A code focused question we can ask:  What guarantees this value cannot be null and is that still true today?

To address the actual question of hardening here or in other PRs:
JPEG has had enough problems of late that I recommend separate PRs for anything not dealing directly with JPEG.

Regarding this PR... let's see if we can't get it over the finish line.
I believe we are very close.

RodneyBaker:
@Chantology 
I've adjusted the title slightly to capture the narrower scope.
I'm posting intent to merge this with the goal of validating the fix works as expected.

RodneyBaker:
Thank you @Chantology for the fix and your continued patience.
LGTM

PR #7009, opened 2026-08-08
style: parenthesise expressions flagged by -Wparentheses

Chantology:

## What and why

Building the tree with `-Wparentheses` reports 15 sites. **One was a real null-pointer dereference** — fixed separately in #7008. The other **14 are correct as written**, but rely on the reader knowing the relative precedence of `<<`, `&`, `|` and `&&`/`||`.

This makes the grouping explicit so the flag can be enabled — the flag being what surfaced that crash in the first place.

| File | Count | Pattern |
|---|---|---|
| `image/tzp/toonzrle.cpp` | 11 | `outval = outval & mask \| value << offs;` → `(outval & mask) \| (value << offs)` |
| `toonzqt/imageutils.cpp` | 1 | `key \| (modifiers & ~0xf0000000)` |
| `toonzlib/fill.cpp` | 1 | `&&` nested in `\|\|` |
| `toonz/xshcellmover.cpp` | 1 | `&&` nested in `\|\|` (matches its own "Or Dragging the Frame Cell" comment) |

## No behaviour change — verified, not assumed

The parentheses were chosen to match how the compiler already parses these expressions. To confirm rather than trust that:

- Operator precedence was checked with a standalone program (`outval & mask | val << offs` does equal `(outval & mask) | (val << offs)`).
- `toonzrle.cpp` was compiled before and after, and the disassembly of the resulting object files compared: **byte-identical machine code**.

## Result

With this PR and #7008, the tree builds clean under `-Wparentheses` (verified by a full build — the only remaining warning is the one #7008 fixes). That makes it possible to enable the flag afterwards.

## Notes for reviewers

- This is deliberately *only* the mechanical part. Enabling the flag is a separate step, and would naturally fold into #6999, which already adds `-Werror=return-type` to the Fedora job — putting both flags on one line avoids two PRs touching the same `cmake` invocation.
- Suggested order: #7008 and this one, then the flag in #6999.
- If you'd rather not carry cosmetic churn in the RLE codec, the alternative is to leave `-Wparentheses` off and accept that this class of bug stays invisible — I'd argue against that given it just produced a crash, but it's your call.

- **Approach / intent given to the model:** enable a compiler warning across the tree, classify every hit rather than assuming, fix the real bug separately, and make the remaining benign ones explicit. Each classification was checked against the code's intent, and the no-op claim was verified by comparing compiled output.

### DIFF
--- a/toonz/sources/image/tzp/toonzrle.cpp
+++ b/toonz/sources/image/tzp/toonzrle.cpp
@@ -202,25 +202,25 @@ start_from_in0:
       break;
 
     case 0xA:
-      outval = outval & not_colmask | GET_IN0 << col_offs;
+      outval = (outval & not_colmask) | (GET_IN0 << col_offs);
       goto start_from_in1;
 
       break;
 
     case 0xB:
-      outval = outval & not_penmask | GET_IN0 << pen_offs;
+      outval = (outval & not_penmask) | (GET_IN0 << pen_offs);
       goto start_from_in1;
 
       break;
 
     case 0xC:
-      outval = outval & not_colmask | *in++ << col_offs;
+      outval = (outval & not_colmask) | (*in++ << col_offs);
       goto start_from_in0;
 
       break;
 
     case 0xD:
-      outval = outval & not_penmask | *in++ << pen_offs;
+      outval = (outval & not_penmask) | (*in++ << pen_offs);
       goto start_from_in0;
 
       break;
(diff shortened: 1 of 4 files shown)


Chantology:
@RodneyBaker this one should be good to go? :)

RodneyBaker:
Per your request this one is waiting on https://github.com/opentoonz/opentoonz/pull/6999.

Whoops.  I read that out of order.
Proceeding.

RodneyBaker:
As always, thanks for the improvements and your patience.
While this PR is 11 commits behind none of those touch the changes here.
LGTM

PR #7011, opened 2026-08-08
fix(appstream): correct the AppStream metadata

Chantology:

## What this fixes

`appstreamcli` reports three issues with `io.github.OpenToonz.appdata.xml`. This file is what **GNOME Software, KDE Discover and Flathub render**, so these are user-visible rather than merely cosmetic.

| Finding | Fix |
|---|---|
| `developer-name-tag-deprecated` | `<developer_name>` → `<developer id="…"><name>…</name></developer>` |
| `developer-info-missing` | added a developer id |
| `description-first-para-too-short` | the first paragraph is what software centres show first; *"This is software for producing a 2D animation."* replaced with a sentence that actually describes the application |

The existing paragraph about the project's history is kept unchanged as the second paragraph.

Also: component type `desktop` → `desktop-application` (current spec), release marked `type="stable"`, and a caption added to the screenshot.

## Validation

```
master:   ✔ Validation was successful: infos: 3, pedantic: 2
this PR:  ✔ Validation was successful: pedantic: 1
```

The one remaining pedantic item is `cid-contains-uppercase-letter`, **deliberately not addressed** — the component id must keep matching the `.desktop` filename and the Flathub app id, so renaming it would break both.

## Notes for reviewers

- **The new description text is my wording — please check it reads the way you want**, since it is what users see in software centres. Current text:
  > OpenToonz is a complete 2D animation package for producing hand-drawn and cut-out animation. It combines drawing and painting tools, an xsheet and timeline for timing, a node-based compositing system with a large library of effects, and scanning and cleanup tools for traditional pencil-on-paper workflows.
- **The developer id `jp.co.dwango` is my choice** (reverse DNS of dwango.co.jp) — happy to change it to whatever you consider correct, since it is an identity claim rather than a technical detail.
- The screenshot caption was written after actually viewing the image; it shows the main window with the scene viewer, level strip, xsheet and palette.
- **This does not let us re-enable the AppStream check disabled in #6987.** That failure was the network reachability probe on the bug tracker URL, which stays flaky regardless of this file's contents. The justification here is software-centre presentation only.

## Two things spotted but left alone

- The screenshot is hosted in the **Flathub** repo (`raw.githubusercontent.com/flathub/io.github.OpenToonz/…`), so this project's metadata depends on a third-party repository's layout. Moving it in-tree would be more robust, but that's a separate decision.
- That screenshot shows **OpenToonz 1.2.1** — it is several versions out of date. Replacing it needs a new image, which is out of scope here (and the file lives in the Flathub repo, not this one).

- **Approach / intent given to the model:** fix the concrete findings `appstreamcli` reports for this file, verifying with the tool before and after rather than assuming. The screenshot caption was written after downloading and viewing the image, not guessed from the filename.


RodneyBaker:
The description can easily scare readers away but the actual changes/updates are straightforward. To points of interest from me:

The product description.

> OpenToonz is a complete 2D animation package for producing hand drawn and cut out animation. It combines drawing and painting tools, an xsheet and timeline for timing, a node based compositing system with a large library of effects, and scanning and cleanup tools for traditional pencil on paper workflows.

(I've removed ai inflected dashes)

This a reasonable update and can be refined as needed.
No opposition to the change.  Just want it to be as correct as possible.


The introduction of the term 'stable'.

   <release version="1.8.0" date="2026-06-15" type="stable" />

To my knowledge this is the first use of the term 'stable' referenced to Opentoonz releases.
This is not an insignificant.  There is (at least) one reference to 'unstable' for the Nightly Release on the Opentoonz website.

“Stable” can mean (at least) four different things, perhaps even simultaneously:
- Release maturity: a normal public release as opposed to alpha, beta, Release Candidate or nightly.
- Runtime reliability: this version is unlikely to crash, corrupt projects or behave unpredictably.
- Compatibility stability: APIs, file formats, scripting interfaces, or workflows won't unexpectedly change.
- Support status: this is the version maintainers recommend and are committed to actively maintaining or patching.

Those views represent very different 'promises'.
The primary question this causes me to ask is:  who is certifying a given release as stable?
Or perhaps more importantly, who is defining and certifying that contract of stability?

None of this is particularly problematic but is more for general consideration.
For legal contract concerns interested parties should refer to Opentoonz licensing.


Aside:  The developer ID jp.co.dwango reverse DNS of dwango.co.jp is rather insightful.  I dare say more than I fully realize.

Chantology:
thanks a lot for the comments. What would you recommend instead of "stable"? Just marked as "release"? My bad regarding the short dashes and thanks for cleaning, I only remove long dashes that are very typical for AI "—"

RodneyBaker:
Ironically, the term 'stable' actually fits well.  It just carries those typical meanings which is why I mention it.  Other options might be 'final' or 'release' as you suggest.

With regard to the dashes... settle in for more verbosity from me.
The em dash is one of the most common ai generated additions that people recognize.  What we see however, perhaps especially with code related text, is an abundance of standard dashes where many are technically correct but we can easily see a paragraph where expect to see no dashes but instead see ten.  I almost prefer to leave the dashes just to preserve the ai generated text. We aren't trying to hide the fact text is ai generated.  Knowing that is a good thing.  I remove the excess because it's unnecessary.
I wouldn't sweat that too much.  That wasn't a request for you to remove the dashes as much as it was me letting everyone know the quote isn't actually a true quote because... I've changed something.  In this case, I removed dashes.

Chantology:
'final' sounds very definitive, 'stable' or 'release' feel more right to me but I don't have a preference. 

Regarding the the product description it looks good without dashes, I can do another round written like you suggested!

RodneyBaker:
As near as I can tell that release was final.
It won't be revisited.
Any change thereafter will have a new number.

I just thought it worth mentioning about use of the term 'stable'.

As this is your PR you get to make the call.

Chantology:
just poking @Chantology 
should be fine now. Decided for sticking with 'stable' and removed the dashes just like you proposed :)

RodneyBaker:
Interestingly, from what I can tell, if no type is supplied the type defaults to 'stable'.
The other options being development and snapshot.

So rather than imply a stable release implicitly you are just stating the type as stable explictly.

All this to say, "Posting intent to merge".

Chantology:
@RodneyBaker do you think it's fine or would you rather have something else as default f.e. dev?

RodneyBaker:
It's fine.
If I had to choose I'd just leave the type off and it would autmatically be set to 'stable'.
Using a custom type was a bad idea on my part.

Chantology:
thank you :)

RodneyBaker:
Pressing forward with this.
Thanks @Chantology 
LGTM

PR #7013, opened 2026-08-08
Fixed foldername casing for versioncontrol

Sotem:
When trying to use "Get Revision..." from the version control menu on a scene opentoonz would give an error wherein is stated that '/tmp/sceneicons' is a non existing folder.
This is because throughout the code 'sceneIcons' is used to create the tmp directory.
I've changed the '/tmp/sceneicons' to now use the uppercase variant.

Afterwards the functionality has been tested and it is working correctly now.


RodneyBaker:
I see no reason to delay this.
Thanks @Sotem for type typo fix.

Sotem:
You're welcome!
If time allows I wish to look at some more at the version control implementation.
It seems that it allows for a lot of improvement, I don't think it is working as intended though that may be my misunderstand.

PR #7015, opened 2026-08-09
[Idea #6370] Fix FX group visibility toggle

doangelshavewings:
## Summary
- Make a group visibility control reflect every grouped child.
- Toggle preview visibility correctly for Level, Zerary, and Palette columns.
- Preserve unified Level-column preview and camera-stand behavior.

## Testing
- Compiled `fxschematicnode.cpp` with the project compile command.
- `git diff --check` passed.
- The full target cannot link locally because the bundled SuperLU archive has no arm64 slice.


RodneyBaker:
Prioritizing review of PRs with those that contain fixes proposed fixes.

doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6370.

PR #7016, opened 2026-08-09
[Idea #6533] Open Level Settings from schematic nodes

doangelshavewings:
## Summary
- open Level Settings when an eligible level-column node is double-clicked in the FX or Stage Schematic
- keep node renaming available with Ctrl+double-click on its name
- leave empty, unsupported, and non-level columns unchanged

## Testing
- compiled fxschematicnode.cpp and stageschematicnode.cpp with the configured CMake compile commands


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6533.

PR #7017, opened 2026-08-09
[Idea #6457] Focus Schematic on selected nodes

doangelshavewings:
## Summary
- add Show Selected Nodes Only and Show All Nodes to FX and Stage Schematic context menus
- keep the filter view-only, preserving scene data and render visibility
- restore the prior view with Show All Nodes, Escape, or when the filtered selection disappears

## Testing
- compiled schematicviewer.cpp, fxschematicscene.cpp, stageschematicscene.cpp, and generated Qt metadata with the configured CMake compile commands


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6457.

PR #7018, opened 2026-08-09
[Idea #6473] Identify FX types in menu entries

doangelshavewings:
## Summary
- label generator and schematic FX consistently in Add, Insert, and Replace menus
- label plugin, preset, and macro entries without changing their action data
- use each built-in FX's existing zerary behavior to identify level-generating entries

## Testing
- compiled addfxcontextmenu.cpp with the configured CMake compile command


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6473.

PR #7019, opened 2026-08-09
[Idea #6707] Unify Schematic selection and navigation

doangelshavewings:
## Summary
- pan the schematic by dragging empty background in Select mode
- preserve Shift-drag rubber-band selection and middle-button panning
- remove the redundant Hand mode after its navigation behavior is reachable from Select mode

## Testing
- compiled schematicviewer.cpp and generated Qt metadata with the configured CMake compile commands


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6707.

PR #7020, opened 2026-08-09
[Idea #6971] Add Schematic reroute insertion

doangelshavewings:
## Summary
- insert the existing FX Pass Through node by double-clicking an FX Schematic link
- insert a persistent identity Pegbar routing point by double-clicking a Stage Schematic link
- preserve Stage parent handles when deleting or undoing intermediate Stage objects

## Testing
- compiled schematicnode.cpp, fxschematicscene.cpp, stageschematicscene.cpp, stageschematicnode.cpp, tstageobjectcmd.cpp, and generated Qt metadata with the configured CMake compile commands


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6971.

PR #7021, opened 2026-08-09
[Idea #6460] Add Text Iwa box anchors

doangelshavewings:
## Summary

- add a nine-position anchor parameter to Text Iwa
- retain centered placement as the default for existing scenes
- keep horizontal text alignment independent from the box anchor

## Testing

- compiled iwa_textfx.cpp from the configured OpenToonz build


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6460.

PR #7022, opened 2026-08-09
[Idea #6549] Add Texture FX preserve-source mode

doangelshavewings:
## Summary

- add an appended Over Source mode to Texture FX
- retain the source beneath transparent texture pixels
- preserve source alpha at antialiased edges

## Testing

- compiled texturefx.cpp from the configured OpenToonz build


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6549.

PR #7023, opened 2026-08-09
[Idea #6551] Add undo for FX schematic node moves

doangelshavewings:
## Summary

- record FX Schematic node drags as a single undo item
- restore every changed DAG position for undo and redo
- retain the existing Alt-link operation undo path

## Testing

- compiled fxschematicscene.cpp from the configured OpenToonz build


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6551.

PR #7024, opened 2026-08-09
[Idea #6470] Add simple Corner Pin controls

doangelshavewings:
## Summary

- add a default-off Simple Controls mode to Pinned Texture
- derive the texture source quad from the connected texture bounds
- retain the existing source and destination cage controls as the default

## Testing

- compiled cornerpinfx.cpp from the configured OpenToonz build


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6470.

PR #7025, opened 2026-08-09
[Idea #6740] Add FX Schematic node display options

doangelshavewings:
Adds persisted FX Schematic display controls for compact wide nodes, larger node text, and optional FX IDs.\n\nThe compact layout widens normal, Zerary, and group nodes while reducing their title-box height and keeping their controls on the outer edge. Display changes refresh the schematic immediately without changing the stored graph.


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6740.

PR #7026, opened 2026-08-09
[Idea #6813] Add Bilateral Blur FX

doangelshavewings:
Adds a native Bilateral Blur raster effect with animated Radius and Color Threshold parameters. It uses spatial and color similarity weights to smooth within regions while preserving color and alpha edges.\n\nThe effect works in float precision, expands input requests by its filter radius, and passes the source through unchanged at zero radius.


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6813.

PR #7027, opened 2026-08-09
[Idea #6814] Add contiguous reference blur FX

doangelshavewings:
Adds a two-input Contiguous Blur effect for overlay shading workflows. Source is blurred normally, then source pixels in the contiguous band touching opaque Reference pixels are restored to prevent a gap from appearing beside line art.\n\nWith no Reference connected, the effect behaves as a normal blur. Radius is animatable and zero passes Source through unchanged.


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6814.

PR #7028, opened 2026-08-09
[Idea #6560] Add Roughen Edges FX

doangelshavewings:
Adds a native Roughen Edges raster effect for hand-drawn-style outlines. The effect applies coherent Perlin displacement at alpha edges while retaining fully opaque interiors.\n\nAmplitude, Scale, and Evolution are animatable. The effect accepts any source that renders to the FX raster pipeline and passes Source through unchanged at zero amplitude.


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6560.

PR #7029, opened 2026-08-09
[Idea #6455] Add Afterimage Trail FX

doangelshavewings:
Adds a native Afterimage Trail effect that renders discrete past source frames as fading echoes behind the current frame.\n\nCount controls the number of echoes, Rate controls their frame spacing, and Start/End Opacity control the trail fade. The current source remains independently on top of the trail.

### DIFF
--- a/toonz/sources/stdfx/CMakeLists.txt
+++ b/toonz/sources/stdfx/CMakeLists.txt
@@ -100,6 +100,7 @@ endif()
 
 set(SOURCES
     adjustlevelsfx.cpp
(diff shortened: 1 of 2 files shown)


RodneyBaker:
This idea addresses requests for an after image trail effect.

This PR in currently state appears to have no mechanism for actually adding the proposed FX into the intereface where it can be accessed by the user.

doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6455.

PR #7030, opened 2026-08-09
Add exposed Macro FX properties

doangelshavewings:
## Summary\n- add per-property Macro FX exposure controls alongside the existing full nested FX controls\n- add an Exposed tab that edits only selected nested properties\n- persist exposure selections in Macro FX data and preserve them through cloning and undo\n\n## Testing\n- compile tmacrofx.cpp and fxsettings.cpp with the project commands and -Werror\n- ninja -C /private/tmp/opentoonz-6370-build toonzqt (blocked at the pre-existing arm64-incompatible bundled SuperLU archive before toonzqt links)

### DIFF
--- a/toonz/sources/common/tfx/tmacrofx.cpp
+++ b/toonz/sources/common/tfx/tmacrofx.cpp
@@ -88,6 +88,13 @@ void collectParams(TMacroFx *macroFx) {
   }
 }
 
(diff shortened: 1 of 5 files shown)


doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6815.

PR #7031, opened 2026-08-09
Add Bezier gradient mask FX

doangelshavewings:
## Summary\n- add a keyframeable closed cubic Bezier mask generator with four anchors and paired tangent handles\n- add continuous asymmetric inner/outer feathering, opacity, and invert controls\n- add FX settings and Viewer point controls for anchors and handles\n\n## Testing\n- compile beziergradientmaskfx.cpp with the project compile command and -Werror\n- git diff --check

### DIFF
--- a/stuff/profiles/layouts/fxs/STD_bezierGradientMaskFx.xml
+++ b/stuff/profiles/layouts/fxs/STD_bezierGradientMaskFx.xml
@@ -1,0 +1,23 @@
+<fxlayout>
+  <page name="Bezier Mask">
+    <vbox label="Path">
+      <control>anchor_1</control>
+      <control>outgoing_1</control>
+      <control>incoming_1</control>
+      <control>anchor_2</control>
+      <control>outgoing_2</control>
+      <control>incoming_2</control>
+      <control>anchor_3</control>
+      <control>outgoing_3</control>
+      <control>incoming_3</control>
+      <control>anchor_4</control>
+      <control>outgoing_4</control>
+      <control>incoming_4</control>
+    </vbox>
+    <separator label="Mask"/>
+    <control>inner_feather</control>
+    <control>outer_feather</control>
+    <control>opacity</control>
+    <control>invert</control>
+  </page>
+</fxlayout>
--- a/toonz/sources/stdfx/CMakeLists.txt
+++ b/toonz/sources/stdfx/CMakeLists.txt
@@ -102,6 +102,7 @@ set(SOURCES
     adjustlevelsfx.cpp
     artcontourfx.cpp
     backlitfx.cpp
+    beziergradientmaskfx.cpp
     blendtzfx.cpp
     blurfx.cpp
     bodyhighlightfx.cpp
--- a/toonz/sources/stdfx/beziergradientmaskfx.cpp
+++ b/toonz/sources/stdfx/beziergradientmaskfx.cpp
@@ -1,0 +1,262 @@
+#include "tfxparam.h"
+#include "tparamset.h"
+#include "tparamuiconcept.h"
+#include "trop.h"
+#include "trasterfx.h"
+#include "stdfx.h"
+
+#include <algorithm>
+#include <array>
+#include <cmath>
+#include <limits>
+#include <vector>
+
+namespace {
+
+constexpr int kPointCount = 4;
+constexpr int kSamplesPerSegment = 24;
+
+struct CubicPath {
+  std::array<TPointD, kPointCount> anchors;
+  std::array<TPointD, kPointCount> outgoing;
+  std::array<TPointD, kPointCount> incoming;
+};
+
+TPointD cubicPoint(const TPointD &p0, const TPointD &p1, const TPointD &p2,
+                   const TPointD &p3, double t) {
+  const double u  = 1.0 - t;
+  const double u2 = u * u;
+  const double t2 = t * t;
+  return p0 * (u2 * u) + p1 * (3.0 * u2 * t) + p2 * (3.0 * u * t2) +
+         p3 * (t2 * t);
+}
+
+std::vector<TPointD> flatten(const CubicPath &path) {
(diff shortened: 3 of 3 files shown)


RodneyBaker:
First pass of this PR includes:

A new effect:

A closed four segment cubic Bézier path
Four anchors
Incoming and outgoing handles for every anchor
Keyframeable point parameters
Independent inner feather and outer feather
Opacity
Invert
Viewer controls for all 12 points
8-bit, 16-bit, and floating-point raster output
A dedicated FX settings layout
Registration in stdfx/CMakeLists.txt

A likely coordinate origin bug:

The new renderer establishes its starting coordinate as:

```
const TPointD tileOrigin =
    tile.m_pos + tile.getRaster()->getCenterD();
```

and then evaluates each pixel as:

`renderToWorld * (tileOrigin + TPointD(x, y))`

That does not appear consistent with how OpenToonz normally treats `tile.m_pos` in zerary/gradient FX.

For example, SquareGradientFx takes `tile.m_pos` itself as the raster space starting location and transforms that through the inverse affine.  The newer Iwa gradient code makes the convention even clearer.  Iwa_CorridorGradientFx converts control points to raster coordinates using:

```
aff * point
    - (tile.m_pos + tile.getRaster()->getCenterD())
    + dimOffset;
```

where `dimOffset` is itself approximately the raster center. Thus the center terms essentially cancel and `tile.m_pos` serves as the tile origin.

I'll add a potentially more serious problem in the next post.

RodneyBaker:
With regard to the above post, the new FX instead adds the raster center before beginning its x/y traversal.

That could result in:
- The mask being displaced by roughly half the rendered tile dimensions
- Different positioning depending on tile dimensions
- Potentially visible seams or discontinuities if rendering is divided into multiple tiles
- Viewer controls no longer lining up with the rendered mask

This PR should be merged until these concerns are addressed or remedied.

The area of code to consider is somewhere near:  `renderToWorld * (tile.m_pos + TPointD(x, y))`

This should be changed to whatever half pixel convention OpenToonz expects, or alternatively,  transform the Bezier geometry into tile local coordinates as Iwa gradient FX does.

More concerns to consider in the next post...

RodneyBaker:
Rendering cost is potentially very high in this PR.

Each cubic is flattened to 24 segments × 4 cubics = 96 polygon edges.  For every output pixel, the code then does:

distanceToPath() // checks all 96 edges.
isInside() // walks all 96 edges again.

So that's roughly 192 segment examinations per pixel.

At 1920×1080, that is on the order of:

2,073,600 × 192 ≈398 million edge tests per frame.

**This is complicated considerably as the effect is presumably something users will animate.**

Redesign of the entire algorithm isn't required before initial inclusion but at minimum the distance test and inside outside crossing test could be performed in one traversal of the 96 segments, or nearly halving that work.

Potential remedies or longer term optimizations could include:
- a quick path bounding box
- adaptive Bezier subdivision
- scanline inside/outside evaluation
- cached flattened geometry while parameters don't change
- spatial subdivision
- special handling of pixels clearly beyond the outer-feather region

This concern is amplified because `doGetBBox()` currently returns `infiniteRectD`, so OpenToonz doesn't get an opportunity to limit normal (non inverted) calculations for the mask.

More to come...

RodneyBaker:
Some other concerns include:

- Fixed 24-sample Bézier approximation may become visible
- The feather parameters are correctly declared as `fxLength`, but are used directly when calculating distance

The feathering isn't necessarily wrong but to confirm correctness we need to test for:

- full-resolution render
- 1/2 preview
- 1/4 preview
- transformed/scaled FX
- (possibly) non uniform transforms (not quite sure how to test for those)

The result of shape and feather should remain visually consistent.

RodneyBaker:
The following is out of scope for this initial introduciton to the Bezier Gradient Mask FX but should be considered for follow up PR:

Bezier interaction is functional but primitive.
The Viewer exposes 12 independent POINT controls: Anchor 1, Outgoing Handle 1, Incoming Handle 1, etc.

There is no actual Bezier relationship imposed between incoming and outgoing handles.
As such moving an anchor does not appear to automatically move its handles.
Rotating one tangent does not affect the opposite tangent.
There is no smooth/corner tangent mode.
The handles appear to be ordinary independent points.

That's perfectly valid as an initial implementation but we should describe thee current implementation as independent Bezier handles rather than implying a traditional linked Bezier editing interface.

The advantage of the proposed (current) implementation is that it is extremely simple and fully keyframeable.

RodneyBaker:
Let's continue with some good things about this PR.

The overall code is unusually self-contained for a new FX. 
There is no invasive infrastructure work.
CMake gets one source entry, the UI gets one layout XML, and everything else lives in one implementation file.

Continuing...

`fxLength` is correctly assigned to spatial parameters.
Opacity is constrained 0–100.
Feather values cannot become negative.
Alpha is clamped.
Degenerate individual flattened segments are handled.
32-bit, 64-bit and float output are supported.
`enableComputeInFloat(true) `is present.
Inner/outer feathering is (genuinely) asymmetric.
Inverting occurs after feather calculation (the order appears to make good sense)
The code is readable and relatively easy to modify.

In short, it provides a good foundation to build upon.

RodneyBaker:
Current assessment:  This PR is promising but not ready to merge without testing and adjustment (if needed) after considering the identified concerns.

Suggested priorities:
- Verify/fix `tile.m_pos` + `getCenterD()` positioning.
- Test full-frame versus tiled rendering for seams.
- Test Preview shrink levels and transformed FX.
- Benchmark 1080p and (preferably) 4K masks.
- Optimize the doubled 96 edge traversal if performance appears to be poor.
- Consider adaptive curve flattening as a refinement.

Review of this PR was assisted by ChatGPT 5.6

doangelshavewings:
Consolidated into draft PR #7056, which includes the implementation for Discussion #6800.

PR #7032, opened 2026-08-09
Allow numeric entry for Type Tool size

doangelshavewings:
## Summary
- make the Type Tool Size field directly editable
- validate and retain entered positive numeric sizes in the existing size list
- apply entered sizes and synchronize other Type Tool option bars

## Testing
- compiled `tooloptions.cpp` with the project CMake command
- ran the same compile with warnings as errors after suppressing pre-existing macOS OpenGL deprecation warnings from `tgl.h`
- regenerated Qt MOC metadata and verified the new slot


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6809.

PR #7033, opened 2026-08-09
Add Alt temporary picker preference

doangelshavewings:
## Summary
- add an Off, Style Picker, or RGB Picker choice under Tools preferences
- temporarily switch to the selected picker while Alt is held in an idle Scene Viewer
- restore the prior tool immediately on Alt release without changing drag or combined-modifier behavior

## Testing
- compiled preferences, tool-handle, Scene Viewer, event, and preferences UI source units with their project CMake commands
- ran those units with warnings as errors after suppressing pre-existing macOS OpenGL deprecation warnings from shared headers


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6516.

PR #7034, opened 2026-08-09
Add Animate Tool arrow-key nudging

doangelshavewings:
## Summary
- move the selected Animate Tool object one logical pixel with each arrow key
- use Shift plus an arrow for a ten-pixel nudge
- reuse existing position locks, keyframe creation, undo handling, and object refresh behavior

## Testing
- compiled `edittool.cpp` with its project CMake command and warnings as errors after suppressing pre-existing macOS OpenGL deprecation warnings from shared headers


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6550.

PR #7035, opened 2026-08-09
Expand column reorder drag target

doangelshavewings:
## Summary
- allow column reordering drags from the existing header handle, column number, or column name
- retain the existing column selection, multi-column move, and undo behavior
- update header tooltips to describe the expanded drag target

## Testing
- compiled `xshcolumnviewer.cpp` with its project CMake command and warnings as errors after suppressing pre-existing macOS OpenGL deprecation warnings from shared headers


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6458.

PR #7036, opened 2026-08-09
Add Level Strip drawing substitution

doangelshavewings:
Adds a Level Strip context-menu action to replace the current Xsheet cell with the selected drawing.

The action is available only for one selected drawing in scene mode and for a level column. It records the replacement in Xsheet undo history.


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6500.

PR #7038, opened 2026-08-09
[Idea #6552] Add Fast Render Current Frame to PNG

doangelshavewings:
## Summary
- Add a `Fast Render Current Frame to PNG` command that renders only the current frame to the existing Fast Render path.
- It saves and restores both the output path and the frame range, so Output Settings is left exactly as the user had it.
- Because it goes through the normal render path rather than grabbing the viewport, camera box and other overlays are absent and the result is full render resolution.
- No default shortcut. Clear messages when PNG is unavailable or the current frame lies outside the scene.

This mirrors the existing `MI_FastRender`, which already establishes the pattern of a one-click render that bypasses Output Settings. The gap it closes is workflow rather than capability: exporting one frame previously meant opening Output Settings, switching format to PNG, narrowing the range, rendering, then putting all three back.

Discussion: https://github.com/opentoonz/opentoonz/discussions/6552 (closed on the grounds that the result is already achievable; this adds the one-click path, not a new capability)

## Testing
- Compiled `rendercommand.cpp` and `mainwindow.cpp` with the project compile command; no new warnings.
- `git diff --check` passed.

### DIFF
--- a/stuff/profiles/layouts/rooms/Default/menubar_template.xml
+++ b/stuff/profiles/layouts/rooms/Default/menubar_template.xml
@@ -261,6 +261,7 @@
         <command>MI_Render</command>
         <separator/>
         <command>MI_FastRender</command>
+        <command>MI_FastRenderFrame</command>
     </menu>
     <menu title="View">
         <command>MI_ViewTable</command>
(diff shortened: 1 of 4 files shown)


doangelshavewings:

PR #7039, opened 2026-08-09
[Idea #6573] Allow extra FFmpeg encoding arguments

doangelshavewings:
## Summary
- Add an `Extra FFmpeg Args` format property to the MP4, WebM, GIF and APNG writers.
- Stored in the existing format property group, so it saves with the scene and with output presets.
- Arguments are appended after the arguments the writer builds and before the output path.
- An empty field produces exactly the command line built today.

Safety:
- Arguments are split on whitespace, honouring double quotes, and passed to `QProcess` as a list, so no shell is involved. This is consistent with the hardening in #6903 and #6904.
- The tokens `-i`, `-y` and `-n` are rejected with a warning: they would retarget the process rather than configure the encode, and the writer owns the input pattern, the overwrite flag and the output path.

This is deliberately the smaller of the two designs discussed. The external-instruction-file prototype from the thread is not implemented; a plain per-format field keeps the failure mode legible.

Discussion: https://github.com/opentoonz/opentoonz/discussions/6573

## Testing
- Compiled all five changed files with the project compile command; no new warnings.
- Ran a harness over the tokenizer and deny list: empty and whitespace-only input yield no arguments; flag/value pairs, codec selectors and hardware-acceleration flags pass through; quoted values survive as one token including colons; extra whitespace collapses; `-i`, `-y` and `-n` are rejected anywhere in the list; and `-i` inside a quoted metadata value is correctly not rejected.
- `git diff --check` passed.

### DIFF
--- a/toonz/sources/image/ffmpeg/tiio_apng.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_apng.cpp
@@ -55,6 +55,12 @@ TLevelWriterAPng::TLevelWriterAPng(const TFilePath &path, TPropertyGroup *winfo)
 
   ffmpegWriter = new Ffmpeg();
   ffmpegWriter->setPath(m_path);
+  {
+    TProperty *extraProp = m_properties->getProperty("Extra FFmpeg Args");
+    if (extraProp)
+      ffmpegWriter->setExtraArgs(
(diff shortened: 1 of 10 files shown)


doangelshavewings:

PR #7040, opened 2026-08-09
[Idea #6773] Add tcomposer output width and height options

doangelshavewings:
## Summary
- Add `-width n` and `-height n` to `tcomposer`, setting the render resolution directly via `TCamera::setRes`.
- Either may be given alone, in which case the other follows the camera aspect ratio.
- They take precedence over `-shrink`, which would otherwise scale the result a second time; the substitution is reported in the render log.
- Non-positive values are rejected with a usage error. Behaviour is unchanged when neither option is given.

`-shrink` was the only way to change render resolution from the command line. It divides the camera resolution by an integer and is marked obsolete, so arbitrary output sizes required editing the scene. This matters for automated pipelines that render at reduced resolution during production and full resolution for finals.

Discussion: https://github.com/opentoonz/opentoonz/discussions/6773

## Testing
- Built `tcomposer` and the full `OpenToonz.app` target on macOS arm64.
- Confirmed `[ -width n ] [ -height n ]` now appear in the usage line, in the expected position, and are absent from a build of unmodified `master`.
- `git diff --check` passed.
- End-to-end render at a forced resolution still needs a real scene file; none ship in `stuff/`.

### DIFF
--- a/toonz/sources/tcomposer/tcomposer.cpp
+++ b/toonz/sources/tcomposer/tcomposer.cpp
@@ -604,6 +604,8 @@ int main(int argc, char *argv[]) {
   RangeQualifier range;
   IntQualifier stepOpt("-step n", "Step");
   IntQualifier shrinkOpt("-shrink n", "Shrink");
+  IntQualifier widthOpt("-width n", "Output width in pixels");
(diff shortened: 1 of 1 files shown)


doangelshavewings:

PR #7041, opened 2026-08-09
Add cell actions to Xsheet context menu

doangelshavewings:
Adds Clone Cells to the Xsheet cell context menu.

Adds Merge for selections whose non-empty cells all belong to one writable level, reusing the existing Filmstrip merge operation and keeping the action disabled for incompatible selections.


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6373.

PR #7042, opened 2026-08-09
Use Up and Down to switch scene columns

doangelshavewings:
Makes unmodified Up and Down change the active column in scene mode after the current tool declines the key.

Tool-specific arrow keys, Shift+Up/Down hold skipping, and Level Strip navigation are unchanged. At the first or last column, the key is consumed instead of seeking frames.


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6731.

PR #7043, opened 2026-08-09
[Idea #6535] Render non-contiguous frame ranges

doangelshavewings:
## Summary
- Add a `Frames` field to Output Settings accepting a printer-style list such as `1-3, 5, 8-10`.
- Frame numbers are one-based, matching the existing Frame Start and End fields.
- Overlapping and unordered input is normalised into one ascending selection and echoed back, so merges are visible.
- Malformed input is rejected with a message and nothing is applied.
- An empty field leaves Frame Start and End authoritative.

Compatibility is the main design constraint:
- The list is written to the scene only when non-empty, so scenes and output presets using a single range serialize exactly as before.
- A stale or malformed stored list falls back to the single range rather than failing the load.
- The selection is honoured only at 1:1 time stretch. Stretching resamples the timeline, so an arbitrary frame list has no well-defined meaning there and the contiguous range stays authoritative.

The render loop now iterates an explicit frame vector. When no list is set that vector is populated with the previous arithmetic, so the contiguous path is unchanged.

Discussion: https://github.com/opentoonz/opentoonz/discussions/6535

## Testing
- Full `OpenToonz.app` target builds and links on macOS arm64 (1595/1595, no errors), so the new slot's moc is regenerated and the connection resolves.
- Ran a harness over the parser, formatter and frame expansion: the discussion's own example round-trips; adjacent singles merge; overlapping and unordered ranges merge; whitespace is tolerated; empty text, frame 0, reversed ranges, dangling dashes and non-numeric input are all rejected; expansion yields exactly the listed frames; step strides across the whole selection; frames past the scene end are clipped.
- `git diff --check` passed.

## Note on overlap
opentoonz-dev #34 edits `toutputproperties.h`, `outputproperties.cpp`, `sceneproperties.cpp` and `outputsettingspopup.cpp`. If that lands first this branch should be rebased onto it.

### DIFF
--- a/toonz/sources/include/toutputproperties.h
+++ b/toonz/sources/include/toutputproperties.h
@@ -5,6 +5,10 @@
 
 #include "tfilepath.h"
 
+#include <string>
+#include <utility>
+#include <vector>
+
 #undef DVAPI
 #undef DVVAR
 #ifdef TOONZLIB_EXPORTS
@@ -82,6 +86,9 @@ private:
   double m_frameRate;
 
   int m_from, m_to;
+  /*-- Optional non-contiguous selection. Empty means "use m_from/m_to",
+   * which keeps existing scenes and presets byte-identical. --*/
+  std::vector<std::pair<int, int>> m_frameRanges;
   int m_whichLevels;
   int m_offset, m_step;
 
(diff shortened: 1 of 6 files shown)


doangelshavewings:

PR #7044, opened 2026-08-09
[Idea #6764] Write scene-origin anchor into sprite sheet text file

doangelshavewings:
## Depends on #7037
This branch is stacked on `idea/6471-individual-frame-padding` and currently shows that commit too. Merge #7037 first; this branch will then reduce to its own commit. The dependency is real rather than cosmetic: the anchor includes the padding offset, which is only applied to individual frames once #7037 lands.

## Summary
- Append `Anchor X` and `Anchor Y` to the sprite sheet text file, giving the scene origin in pixels from the top-left of one frame cell.
- Emit a text file for the `Individual` format, which previously wrote none, describing its cell geometry, padding and anchor.
- Every existing line keeps its wording and position, so current parsers are unaffected; the change is append-only.

The discussion asks for "pivot point data". A true level pivot is not reachable here: `TLevelWriter` is handed images with no scene, column, stage object or hooks, and reaching one would mean widening a shared I/O interface used by every format. The writer also crops every frame with one global union trim box, so a per-frame pivot would emit the same value repeated.

What the writer *can* derive is the scene origin. Each frame is a rendered camera raster whose centre is that origin, and the trim box, scale and padding place it inside the cell. That is the value needed to put a trimmed sprite back where it sat in the camera, which is what the requester is currently recovering by hand in another application.

Discussion: https://github.com/opentoonz/opentoonz/discussions/6764

## Testing
- Compiled `tiio_sprite.cpp` with the project compile command; no new warnings.
- Ran a harness over the anchor formula: it is the raster centre when untrimmed, unscaled and unpadded; it shifts by the trim origin; it tracks Scale at 25%, 50% and 200%; padding adds absolute pixels on top of the scaled position; and the origin round-trips for several combinations of trim, scale and padding.
- `git diff --check` passed.

### DIFF
--- a/toonz/sources/image/sprite/tiio_sprite.cpp
+++ b/toonz/sources/image/sprite/tiio_sprite.cpp
@@ -99,6 +99,15 @@ TLevelWriterSprite::~TLevelWriterSprite() {
   int totalHorizPadding = 0;
   int spriteSheetWidth;
   int spriteSheetHeight;
+  // Scene origin inside one frame cell, in pixels from the cell's top-left
+  // corner. A level pivot is not reachable here: the writer is handed only
+  // images, with no scene, column or stage object. What it does know is that
+  // every frame is a rendered camera raster whose centre is the scene origin,
+  // so the trim box, scale and padding place that origin inside the cell.
+  QString anchorX = QString::number(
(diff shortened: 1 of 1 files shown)


doangelshavewings:

PR #7045, opened 2026-08-09
Add automatic current frame selection

doangelshavewings:
## Summary\n- add an opt-in Xsheet preference to select the current cell while editing a scene\n- leave playback, the Level Strip, and the camera column unchanged\n- refresh the Xsheet cell and row areas when the selection changes\n\n## Testing\n- compiled xsheetviewer.cpp with warnings treated as errors\n- compiled preferencespopup.cpp with warnings treated as errors\n- compiled preferences.cpp with warnings treated as errors


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6744.

PR #7046, opened 2026-08-09
Add hold cell creation toggle command

doangelshavewings:
## Summary\n- add a configurable Xsheet command to toggle creation in hold cells\n- assign Caps Lock as the default shortcut while allowing users to change it\n- reuse the existing preference without changing the creation behavior\n\n## Testing\n- compiled xsheetcmd.cpp with warnings treated as errors\n- compiled mainwindow.cpp with warnings treated as errors


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6753.

PR #7047, opened 2026-08-09
Collapse Timeline column selection on click

doangelshavewings:
## Summary\n- collapse a selected Timeline column group when it is clicked without modifiers\n- preserve multi-column movement when the mouse drag actually moves columns\n- leave classic Xsheet selection behavior unchanged\n\n## Testing\n- compiled xsheetdragtool.cpp with warnings treated as errors


RodneyBaker:
It isn't quite clear what this proposal suggests or what problem it addresses.

doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6766.

PR #7048, opened 2026-08-09
Stop drawing substitution at level bounds

doangelshavewings:
## Summary\n- stop normal and similar drawing substitution at the first and last drawing\n- record only changed cells in substitution undo entries\n- omit undo entries when a substitution reaches a level boundary\n\n## Testing\n- compiled xsheetcmd.cpp with warnings treated as errors


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6767.

PR #7049, opened 2026-08-09
Select keys in selected cells

doangelshavewings:
## Summary\n- add a configurable command to select keyframes within a selected cell range\n- replace the cell selection with the resulting keyframe selection\n- expose the command in the cell and keyframe context menus\n\n## Testing\n- compiled xsheetcmd.cpp with warnings treated as errors\n- compiled xshcellviewer.cpp with warnings treated as errors\n- compiled mainwindow.cpp with warnings treated as errors


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6808.

PR #7050, opened 2026-08-09
Add select parent column command

doangelshavewings:
## Summary\n- add a configurable command to select the current column's parent column\n- update the active column and Timeline selection together\n- expose the command in the column-header context menu\n\n## Testing\n- compiled xsheetcmd.cpp with warnings treated as errors\n- compiled xshcolumnviewer.cpp with warnings treated as errors\n- compiled mainwindow.cpp with warnings treated as errors


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6819.

PR #7051, opened 2026-08-09
Make Shift Origin toggle hide references

doangelshavewings:
## Summary\n- make Show Shift Origin control unshifted Shift and Trace references\n- preserve original Edit Shift behavior by enabling the control when Edit Shift starts\n- apply the same visibility behavior in scene and Level Strip rendering\n\n## Testing\n- compiled stage.cpp with warnings treated as errors\n- compiled sceneviewer.cpp with warnings treated as errors


RodneyBaker:
At a glance I don't follow what problem(s) this PR specifically addresses.

doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6474.

PR #7052, opened 2026-08-09
Add Pencil Test interval countdown sound

doangelshavewings:
Adds an optional countdown sound for Camera Capture interval captures.

- Adds a disabled-by-default Play countdown sound option.
- Plays one sound per displayed countdown second and when capture begins.
- Keeps the control unavailable while an interval capture is active.

Tested by compiling penciltestpopup.cpp with warnings treated as errors.


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6480.

PR #7053, opened 2026-08-09
Add Palette Window style filtering

doangelshavewings:
Adds a filter field to the Palette Window toolbar.

- Matches style names case-insensitively; plain text matches anywhere and * / ? enable wildcard matching.
- Keeps palette selection, renaming, tooltips, and drag/drop mapped to the original styles.
- Provides an inline clear control and an Options > Visible Toolbar Buttons toggle.

Tested by compiling paletteviewer.cpp and paletteviewergui.cpp with warnings treated as errors.


doangelshavewings:
Consolidated into draft PR #7057, which includes the implementation for Discussion #6534.

PR #7055, opened 2026-08-09
Rendering, export and interchange improvements

doangelshavewings:
Six independent improvements to the rendering and export path, each kept as its own commit so they can be reviewed, and if necessary reverted, separately. Each addresses a specific feature request; the discussion is linked per item below.

Supersedes #7037, #7038, #7039, #7040, #7043 and #7044, which are closed in favour of this branch.

---

### 1. Apply sprite sheet padding to individual frames
Discussion: https://github.com/opentoonz/opentoonz/discussions/6471

Top, Bottom, Left and Right Padding were honoured only by the Grid, Vertical and Horizontal layouts. The `Individual` format saved each trimmed frame directly, so padding was silently dropped — which is the option users need when sprites are too large for one sheet. Disabling Trim Empty Space is not a workaround; the reported case requires it enabled.

Each frame is now drawn onto a transparent canvas expanded by the configured padding. When every padding value is zero the original image is saved unchanged, so existing output is byte-identical. The change is confined to the `Individual` branch; the three sheet layouts are untouched.

### 2. Write scene-origin anchor into the sprite sheet text file
Discussion: https://github.com/opentoonz/opentoonz/discussions/6764

`Anchor X` and `Anchor Y` are appended to the sprite sheet text file, giving the scene origin in pixels from the top-left of one frame cell. The `Individual` format, which wrote no text file at all, now writes one describing its cell geometry, padding and anchor. Every existing line keeps its wording and position, so current parsers are unaffected.

The request asks for "pivot point data". A true level pivot is not reachable here: `TLevelWriter` is handed images with no scene, column, stage object or hooks, and reaching one would mean widening a shared I/O interface used by every format. The writer also crops every frame with one global union trim box, so a per-frame pivot would emit the same value repeated. The scene origin *is* derivable — each frame is a rendered camera raster whose centre is that origin — and it is the value needed to place a trimmed sprite back where it sat in the camera.

Depends on item 1: the anchor includes the padding offset.

### 3. Add tcomposer output width and height options
Discussion: https://github.com/opentoonz/opentoonz/discussions/6773

`-shrink` was the only way to change render resolution from the command line. It divides the camera resolution by an integer and is marked obsolete, so arbitrary output sizes required editing the scene. This matters for automated pipelines that render reduced during production and full for finals.

`-width n` and `-height n` set the resolution directly via `TCamera::setRes`. Either may be given alone, in which case the other follows the camera aspect ratio. They take precedence over `-shrink`, which would otherwise scale a second time, and the substitution is reported in the render log. Non-positive values are rejected. Behaviour is unchanged when neither is given.

### 4. Add Fast Render Current Frame to PNG
Discussion: https://github.com/opentoonz/opentoonz/discussions/6552

Exporting one frame at render resolution meant opening Output Settings, switching format to PNG, narrowing the range, rendering, then putting all three back. `MI_FastRender` already establishes the pattern of a one-click render that bypasses Output Settings; this is its single-frame counterpart.

It saves and restores both the output path and the frame range, so Output Settings is left exactly as the user had it. Because it uses the normal render path rather than grabbing the viewport, camera box and other overlays are absent and the result is full render resolution. No default shortcut.

Note: the discussion was closed on the grounds that the result is already achievable. That is true — this closes the workflow gap, not a capability gap.

### 5. Allow extra FFmpeg encoding arguments
Discussion: https://github.com/opentoonz/opentoonz/discussions/6573

The FFmpeg-backed writers built a fixed argument list, so options the UI does not expose — hardware acceleration being the motivating example — were unreachable without editing the source.

MP4, WebM, GIF and APNG each gain an `Extra FFmpeg Args` format property, stored in the existing property group so it saves with the scene and with output presets. Arguments are appended after the writer's own and before the output path. They are split on whitespace honouring double quotes and passed to `QProcess` as a list, so no shell is involved, consistent with the hardening in #6903 and #6904. The tokens `-i`, `-y` and `-n` are rejected with a warning: they would retarget the process rather than configure the encode. An empty field produces exactly the command line built today.

This is deliberately the smaller of the two designs discussed; the external-instruction-file prototype is not implemented.

### 6. Render non-contiguous frame ranges
Discussion: https://github.com/opentoonz/opentoonz/discussions/6535

Output Settings could describe only one contiguous range, so rendering frames 1-3 and 8-10 meant two renders and reconciling the output numbering by hand. This matters most for sprite sheets, where duplicate frames are skipped but the rest must keep their relative order.

A `Frames` field accepts a printer-style list such as `1-3, 5, 8-10`, one-based to match the existing Frame Start and End fields. Overlapping and unordered input is normalised into one ascending selection and echoed back so merges are visible; malformed input is rejected and nothing is applied.

Compatibility: the list is written to the scene only when non-empty, so scenes and presets using a single range serialize exactly as before, and a stale or malformed stored list falls back to the single range rather than failing the load. The render loop now iterates an explicit frame vector which, when no list is set, is populated with the previous arithmetic. The selection is honoured only at 1:1 time stretch, since stretching resamples the timeline and an arbitrary frame list has no well-defined meaning there.

Overlap note: opentoonz-dev #34 edits the same four files; if it lands first this needs a rebase.

---

## Testing

- Full `OpenToonz.app` target builds and links on macOS arm64 with all six commits applied: **1595/1595, zero errors**. Also built and linked at each individual commit.
- `clang-format` reports no violations on any changed line. The files carry pre-existing violations under newer clang-format versions (pointer alignment); those were left alone rather than reformatted.
- `git diff --check` passes.
- Focused harnesses were run against the logic that can be exercised in isolation:
  - **Sprite padding** — zero padding is byte-identical to current output; padded canvas size, content offset, transparent margins on all four sides, asymmetric values not symmetrised, absolute under Scale, and a fully transparent frame all verified.
  - **Anchor** — raster centre when untrimmed/unscaled/unpadded; shifts by the trim origin; tracks Scale at 25/50/200%; padding adds absolute pixels; origin round-trips across several trim/scale/padding combinations.
  - **FFmpeg arguments** — flag/value pairs, codec selectors and hardware-acceleration flags pass; quoted values survive as one token; `-i`, `-y`, `-n` rejected anywhere in the list; `-i` inside a quoted metadata value correctly not rejected.
  - **Frame ranges** — the discussion's own example round-trips; adjacent singles and overlapping/unordered ranges merge; empty text, frame 0, reversed ranges, dangling dashes and non-numeric input rejected; expansion yields exactly the listed frames; step strides across the whole selection; frames past the scene end are clipped.
- `tcomposer` was built and `-width`/`-height` confirmed present in the usage line and absent from an unmodified `master` build.
- Not covered locally: end-to-end renders needing a real scene file, since none ship in `stuff/`.

## Build note

Building natively on Apple Silicon currently fails to link because `thirdparty/superlu/libsuperlu_4.1.a` is a fat i386/x86_64 archive with no arm64 slice. The SuperLU 4.1 source already ships in the repo and builds clean for arm64 with `-Wno-implicit-function-declaration` (clang 15+ rejects the C89 style). Regenerating that one archive is enough to make the full target link — relevant to #6882.

### DIFF
--- a/stuff/profiles/layouts/rooms/Default/menubar_template.xml
+++ b/stuff/profiles/layouts/rooms/Default/menubar_template.xml
@@ -261,6 +261,7 @@
         <command>MI_Render</command>
         <separator/>
         <command>MI_FastRender</command>
+        <command>MI_FastRenderFrame</command>
     </menu>
     <menu title="View">
         <command>MI_ViewTable</command>
--- a/toonz/sources/image/ffmpeg/tiio_apng.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_apng.cpp
@@ -55,6 +55,12 @@ TLevelWriterAPng::TLevelWriterAPng(const TFilePath &path, TPropertyGroup *winfo)
 
   ffmpegWriter = new Ffmpeg();
   ffmpegWriter->setPath(m_path);
+  {
+    TProperty *extraProp = m_properties->getProperty("Extra FFmpeg Args");
+    if (extraProp)
+      ffmpegWriter->setExtraArgs(
+          QString::fromStdString(extraProp->getValueAsString()));
+  }
   if (TSystem::doesExistFileOrLevel(m_path)) TSystem::deleteFile(m_path);
 }
 
@@ -215,14 +221,17 @@ TImageP TLevelReaderAPng::load(int frameIndex) {
 Tiio::APngWriterProperties::APngWriterProperties()
     : m_scale("Scale", 1, 100, 100)
     , m_looping("Looping", true)
-    , m_extPng("ExtPng", false) {
+    , m_extPng("ExtPng", false)
+    , m_extraArgs("Extra FFmpeg Args", L"") {
   bind(m_scale);
   bind(m_looping);
   bind(m_extPng);
+  bind(m_extraArgs);
 }
 
 void Tiio::APngWriterProperties::updateTranslation() {
   m_scale.setQStringName(tr("Scale"));
   m_looping.setQStringName(tr("Looping"));
   m_extPng.setQStringName(tr("Write as .png"));
+  m_extraArgs.setQStringName(tr("Extra FFmpeg Args"));
 }
--- a/toonz/sources/image/ffmpeg/tiio_apng.h
+++ b/toonz/sources/image/ffmpeg/tiio_apng.h
@@ -75,6 +75,7 @@ public:
   TIntProperty m_scale;
   TBoolProperty m_looping;
   TBoolProperty m_extPng;
+  TStringProperty m_extraArgs;
   APngWriterProperties();
   void updateTranslation() override;
 };
--- a/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
+++ b/toonz/sources/image/ffmpeg/tiio_ffmpeg.cpp
@@ -189,6 +189,54 @@ void Ffmpeg::createIntermediateImage(const TImageP &img, int frameIndex) {
   m_cleanUpList.push_back(tempPath);
 }
 
+namespace {
+
+bool tokenizeFfmpegArgs(const QString &args, QStringList &tokens) {
+  QString current;
+  bool inQuotes = false;
+  bool started  = false;
+  for (const QChar &c : args) {
+    if (c == '"') {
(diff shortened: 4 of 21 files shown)


doangelshavewings:
Consolidated into draft PR #7057, which now carries the focused commits and linked discussions.

PR #7058, opened 2026-08-09
(Planning) Scripting APIs, libtiff cleanup and a first TLV test

doangelshavewings:
Four changes to scripting, developer architecture and build hygiene, each a separate commit. Nothing here is user-visible except the two new script APIs, so the acceptance criteria are mostly "the same bytes come out" rather than "the user can now do X".

---

### 1. Script APIs for project and scene folders
Discussion: https://github.com/opentoonz/opentoonz/discussions/6849

ToonzScript could only work with absolute paths, so scripts were not shareable without every user hand-editing hardcoded variables. Adds project and scene folder accessors so a script can compute paths relative to the scene.

### 2. ToonzScript plain-text and JSON file access
Discussion: https://github.com/opentoonz/opentoonz/discussions/6375

ToonzScript could locate and manipulate file paths but not read or write their contents, so data had to be smuggled through scene fields or produced by tools outside OpenToonz. The reporter's case was importing rigging data to set up scenes.

`FilePath` gains `readText`, `writeText`, `readJson` and `writeJson`. Text is UTF-8 and `writeText` creates the parent directory, matching the rest of the file API.

JSON is converted explicitly between `QJsonValue` and `QScriptValue` rather than evaluating the document as script source, which would execute whatever the file happened to contain. Parse errors, unreadable files and unwritable destinations raise script errors carrying the underlying reason. Only objects and arrays may be written.

This does not attempt the ToonzScript rewrite the thread also discusses; it removes the specific blocker the reporter hit.

### 3. Read TZP byte order through the public libtiff API
Discussion: https://github.com/opentoonz/opentoonz/discussions/6345

`tiio_tzp.cpp` reached into libtiff's private `TIFF` struct to detect big-endian TZP files written on Irix, which forced it to include `tiffiop.h`. `TIFFIsByteSwapped` reports whether file order differs from host order, so combining it with the host order recovers the file's own endianness. Identical on every currently supported platform, all little-endian, and still correct on a big-endian host.

`tiio_tzp.cpp` no longer includes `tiffiop.h`. This is a partial answer to that discussion; see the note at the end.

### 4. A TLV round-trip test harness
Discussion: https://github.com/opentoonz/opentoonz/discussions/6746

That discussion proposes rewriting `tiio_tzl.cpp`, 2,500 lines carrying branches for every TLV version ever written, with no tests. A rewrite of the format that holds users' artwork is not reviewable without evidence, and the repository ships no TLV files to test against. Since OpenToonz can write TLV, the harness generates its own corpus.

It pins down a narrow, total contract: whatever ink, paint and tone go in must come back out, and writing the same level twice must produce the same bytes. Cases cover multiple frames, extreme aspect ratios, odd dimensions and a wide palette.

Building tests is opt-in through `WITH_TESTS`, so a normal build is unchanged. **This is the first test target in the tree**; `enable_testing()` lives beside it so later work has somewhere to put tests.

#### It found a defect on its first run

Some levels cannot be read back at all, failing with `Loading tlv: buffer size error`. Two independent triggers:

- **any level shorter than six rows**, regardless of width, frame count or style count. The boundary is sharp and reproducible.
- **levels whose data compresses poorly.** A 32x32 level round-trips with 16 styles but fails with 256, identical in shape otherwise.

Both are consistent with the writer under-allocating the LZO output buffer: LZO can expand incompressible input, and small inputs are dominated by its fixed overhead. This is recorded as a known issue and reported by the harness rather than asserted, so the suite stays green while the defect stands. Fixing it looks like the natural first task of the rewrite, at which point the probe becomes an assertion.

---

## Testing

- Full `OpenToonz.app` target builds and links on macOS arm64, 1595/1595, zero errors, with `WITH_TESTS` both off and on.
- `ctest` passes: 1/1.
- `clang-format` reports no violations on the changed lines. These files carry pre-existing violations under newer clang-format versions, which were left alone rather than reformatted.
- `git diff --check` passes.

## Scope notes

Three further items from these discussions were considered and deliberately left out, because they cannot be verified from a clean checkout:

- **The rest of #6345.** The remaining `tiffiop.h` user, `toonzrle.cpp`, installs a custom TIFF codec by assigning `tif_decoderow` and driving `tif_rawcp` by hand. Extracting it behind `TIFFReadRawStrip` is the right fix, but OpenToonz has no TZP writer and no `.tzp` file exists here, so there is nothing to test a read against. Separately, replacing `TIFFReadRGBAStrip`/`TIFFReadRGBATile` to keep 16 bits means reimplementing `tif_getimage_64.c`, a 2,381-line fork of libtiff's photometric conversion covering palette, YCbCr, CIELAB, LOGLUV and every bit depth; getting that subtly wrong silently corrupts every TIFF import.
- **The libtiff version bump** to 4.2+, also requested in #6345. Worth doing on its own merits, since 4.0.3 dates from 2012.
- **#6358, replacing GLEW with libepoxy.** The code change is small and libepoxy is the right choice given EGL support is the stated motivation. But Windows GL dependencies ship from an external prebuilt bundle, so such a PR would break Windows CI until that bundle carries libepoxy.

## Build note

Building natively on Apple Silicon currently fails to link because `thirdparty/superlu/libsuperlu_4.1.a` is a fat i386/x86_64 archive with no arm64 slice. The SuperLU 4.1 source already ships in the repo and builds clean for arm64 with `-Wno-implicit-function-declaration`. Regenerating that one archive is enough to make the full target link, which is how everything above was verified. Relevant to #6882.

### DIFF
--- a/toonz/sources/CMakeLists.txt
+++ b/toonz/sources/CMakeLists.txt
@@ -746,6 +746,14 @@ add_subdirectory(tnztools)
 
 add_subdirectory(${SDKROOT}/lzo/driver lzodriver)
 
+# Tests are opt-in so a normal build is unaffected. This is the first test
+# target in the tree; see discussion #6746.
+option(WITH_TESTS "Build the test targets" OFF)
+if(WITH_TESTS)
+    enable_testing()
+    add_subdirectory(tests)
+endif()
+
 add_subdirectory(toonz)
 add_subdirectory(tcleanupper)
 add_subdirectory(tcomposer)
--- a/toonz/sources/image/tzl/tiio_tzl.cpp
+++ b/toonz/sources/image/tzl/tiio_tzl.cpp
@@ -2121,7 +2121,7 @@ TImageP TImageReaderTzl::load14() {
     fread(&actualBuffSize, sizeof(TINT32), 1, chan);
 
     if (actualBuffSize <= 0 ||
-        actualBuffSize > (int)(iconLx * iconLx * sizeof(TPixelCM32)))
+        actualBuffSize > (int)(iconLx * iconLy * sizeof(TPixelCM32)))
       throw TException("Loading tlv: icon buffer size error.");
 
     TRasterCM32P raux = TRasterCM32P(iconLx, iconLy);
--- a/toonz/sources/image/tzl/tiio_tzl.cpp
+++ b/toonz/sources/image/tzl/tiio_tzl.cpp
@@ -1,5 +1,6 @@
 
 
+#include <vector>
 #include "tiio_tzl.h"
 #include "tmachine.h"
 #include "tsystem.h"
@@ -2120,15 +2121,23 @@ TImageP TImageReaderTzl::load14() {
       throw TException("Loading tlv: bad icon size.");
     fread(&actualBuffSize, sizeof(TINT32), 1, chan);
 
-    if (actualBuffSize <= 0 ||
-        actualBuffSize > (int)(iconLx * iconLy * sizeof(TPixelCM32)))
+    // actualBuffSize is a compressed length. LZO can expand incompressible
+    // input and carries fixed per-block overhead, so a small or noisy icon
+    // legitimately stores more bytes than its raw size. Bounding it by the raw
+    // size therefore rejected valid files; bounding it by LZO's documented
+    // worst-case expansion accepts them while still capping what a malformed
+    // file can make us allocate.
+    TINT32 rawIconSize   = (TINT32)(iconLx * iconLy * sizeof(TPixelCM32));
+    TINT32 maxBuffSize   = rawIconSize + rawIconSize / 16 + 64 + 3;
+    if (actualBuffSize <= 0 || actualBuffSize > maxBuffSize)
       throw TException("Loading tlv: icon buffer size error.");
 
-    TRasterCM32P raux = TRasterCM32P(iconLx, iconLy);
-    if (!raux) return TImageP();
-    raux->lock();
-    imgBuff = (UCHAR *)raux->getRawData();  // new UCHAR[imgBuffSize];
-    fread(imgBuff, actualBuffSize, 1, chan);
+    // Read the compressed payload into its own buffer. It used to be read into
+    // the icon raster, which is what forced the bound above to be a raw size.
+    std::vector<UCHAR> iconBuff(actualBuffSize);
+    imgBuff = iconBuff.data();
+    if (fread(imgBuff, actualBuffSize, 1, chan) != 1)
+      throw TException("Loading tlv: icon read error.");
 
 #if !TNZ_LITTLE_ENDIAN
     Header *header    = (Header *)imgBuff;
@@ -2142,8 +2151,6 @@ TImageP TImageReaderTzl::load14() {
     if (!codec.decompress(imgBuff, actualBuffSize, ras, m_safeMode))
       return TImageP();
     assert((TRasterCM32P)ras);
-    raux->unlock();
-    raux = TRasterCM32P();
 
 #if !TNZ_LITTLE_ENDIAN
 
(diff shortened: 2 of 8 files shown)


RodneyBaker:
The goal here appears to be to remove old architectural obstacles so future development becomes easier and testable.

As a whole this  collection is too broad to consider together but is fine for discussion and documentation.  The code (commits) do not represent changes that will necessarily move forward but rather explore how the present code could best move forward given the outcome of discussion and consensus. (See more about the broad scope of this PR below)

Items 1 and 2 relate to ToonzScript work.
Item 3 is also independent with a focus on maintenance work.
Item 4 directly relates to workng with and specifically testing the TLV format.

More deeply, this PR's commits attempt to address the referenced discussions.

1.  Portable ToonzScript paths. Adds getProjectFolder() and getSceneFolder(), addressing the problem that reusable scripts otherwise need hard coded absolute paths. That directly answers Discussion #6849.
2. Script data I/O. Adds FilePath.readText(), writeText(), readJson() and writeJson(). This is a substantial improvement to ToonzScript because scripts can exchange external data rather than encoding information indirectly into scenes. The motivating example was importing rigging information such as JSON.
3. Small TZP/libtiff cleanup. Removes one dependency on libtiff's private tiffiop.h by replacing direct access to TIFF internals with TIFFIsByteSwapped(). This is sensible maintenance, but it is neither scripting nor TLV. It only partially addresses the larger libtiff problem because other private API dependency remains elsewhere.
4. Establish a TLV safety net and fix a TLV defect. This is much larger and more consequential than its position in the PR suggests. It creates OpenToonz's first opt in CTest target, generates TLV levels, writes them, reads them back, compares Toonz Raster ink/paint/tone data, checks repeatable output, and tests unusual dimensions and palette sizes. In doing so it exposed a real failure in the current TLV reader, which the PR then proposes a fix.

It remains to be seen how this planning and discussion of this PR will move forward but it does provide paths to consider as we move the code forward.

The PR itself cannot be merged given how overly broad it is.

RodneyBaker:
I want to focus on Item 4 and since it's not broken out on its own I'll do that here.

This path appears thoughtful about Toonz Raster format itself. 
The generated raster explicitly exercises the three pieces of TPixelCM32: 12-bit ink, 12-bit paint and 8-bit tone, including intermediate tone values rather than merely fully painted/fully inked pixels. 

That is exactly the sort of data we would want protected before touching TLV encoding.

The original TLV discussion (see reference) documented those three components and why tone is significant.

The proposed test cases include small images, multiple frames, tall/narrow and wide/short rasters, odd dimensions, many styles, one-pixel and very-short images. 

That is good targeted coverage rather than merely writing one ordinary TLV.

RodneyBaker:
Perhaps most importantly with Item 4... and potentially the most important part of this PR...

A bug is addressed.

The existing TLV 1.4 reader treated actualBuffSize (the compressed data size) as though it could never exceed the uncompressed raster size. But LZO compression can expand small or poorly compressible data.

Here we have exposed failures for extremely short levels and high-entropy/style-heavy levels.

The PR consequently changes production code in TImageReaderTzl::load14(). Instead of reading the compressed stream into a raster sized buffer it allocates a separate std::vector<UCHAR> based upon the stored compressed size and decompresses from there.  

I admit that I don't follow all of that but do recognize what the change is proposing to do.
It is the consequences of addressing the problem that I have some concern with.

Lets examine.

The PR proposes a calculation:

`rawSize + rawSize / 16 + 64 + 3`

This being the maximum acceptable compressed buffer size. 

But `TRasterCodecLZO::compress()` doesn't store only the LZO compressed payload. 
It prepends its own Header, and outDataSize explicitly becomes:

`outSize + headerSize`

Before that buffer is returned.
`TLevelWriterTzl::saveImage() ` subsequently writes that complete rCompressed size into the TLV as buffSize.

So the reader's theoretical maximum seemingly should account for the codec header too.

There is a related boundary issue also and we'll need to look into that.

@doangelshavewings Can you clarify the following:

minimum: actualBuffSize >= sizeof(Header)
maximum: LZO worst case payload plus the codec header, unless there is another reason that header is already included in the formula.

A little into the good stuff:
The semantic round trip is direct and correct:  ink + paint + tone written → identical ink + paint + tone read.

We'll need to be convinced about making byte for byte identical TLV output a fundamental requirement for a future rewrite.  This might be overly restrictive?

For current TLV levels that seems reasonable but for defining what a future tiio_tzl.cpp rewrite is permitted to do, we should define:
Semantic TLV equivalence = mandatory
Byte identical encoding = useful diagnostic / current writer regression

In other words, the original #6746 proposal called for keeping the TLV architecture and compatibility, not necessarily preserving every byte emitted by today's implementation.

RodneyBaker:
Recommended breakout of PRs:

PR A:   Expand ToonzScript filesystem/data APIs
Items 1 and 2 belong together. Relative project/scene paths plus text/JSON I/O collectively make scripts much more portable and capable.**

PR B:  Remove private libtiff access from tiio_tzp.cpp
Item 3. Tiny, focused architectural cleanup with an easily understood purpose.

PR C:  Add TLV round trip testing and fix compressed buffer handling
Item 4, including WITH_TESTS, the CTest target, tlv_roundtrip_test.cpp, and the load14() fix. Normally I would break out the fix as a separate PR but the regression test and bug fix are sufficiently connected they can be kept together rather than artificially splitting them apart.  The main reason to split the fix into its own PR would be to put additional spotlight on that fix.

I believe these PRs may already exist and are linked as discussions but are now closed.

**So much of Toonzscript development depends on moving to Qt6 and modern ECMA.  As this upate will be a major breaking change that update logically will target Opentoonz 2.0.  Any update to current Toonzscript should be to shore up current scripting support without introducing breaking change.

PR #7060, opened 2026-08-11
Fix layer blending alpha

### LINKED ISSUE #5408: Layer Blending Effects don't work as expected with partially transparent pixels
### Description

Layer Blending Effects like Add, Add Ino, Multiply, and Multiply Ino do not work as expected with antialiasing/partially transparent pixels (I used these specific effects for this report, but it applies to many of the other blending effects).  

For example, the partially transparent pixels on this circle become a dark halo when the Add or Add Ino effect is used to combine it with the dark yellow shape:

add2

This is the expected behavior of an Add blending mode (screenshot from Fire Alpaca):

add4

Another example -- using Multiply Ino with the Clipping Mask results in a visible line on the belly where there are partially transparent pixels in the "front" column:

multiply5

Turning off the Clipping Mask and using a Matte In hides the line: 

multiply6

However, using the regular Multiply with a Matte In also results in a visible line:

multiply7

This is the expected behavior of a Multiply blending mode (screenshot from Fire Alpaca): 

multiply8

I changed the settings and the color space options on these effects and did not find any combination of settings that solved these issues.

Here are the image files used to test the effects: 

[images.zip](https://github.com/opentoonz/opentoonz/files/15017701/images.zip)

### OpenToonz Version

Nightly (Specify below)

2024-03-26

### Operating System

Windows

### GPU

NVIDIA

### Graphics Tablet

Wacom

konero:
Fixes #5408 

See the issue for info. This should fix the haloing seen when using **clipping mask** modes (on layer blending FX). I __think__ the issue is it was multiplying `Fore` opacity by `Back` alpha, still blending against partially transparent `Back` pixels. With soft edges the `Back RGB` was already dimmed, then blend dimmed it again, so the result was a darker RGB along the edge.

Now the fix treats clipping as a source-atop op:
- Saves `Back` alpha
- Recover the `Back's` unpremultiplied color
- Blend once against the opaque working `Back` pixel
- Multiply result RGB by the saved `Back` alpha
- And finally restore the appropriate output alpha

With all that said, the math of this is still really difficult and I've been looking at this issue for a long time. @RattusaurusRex if you can run your tests again against this fix to confirm if there are any other FX with any issues. I also need someone else to test 8/16/32 bit preview + output and comparing against the official release build, to make sure nothing changed.

### DIFF
--- a/toonz/sources/stdfx/ino_blend_cross_dissolve.cpp
+++ b/toonz/sources/stdfx/ino_blend_cross_dissolve.cpp
@@ -11,9 +11,6 @@ class ino_blend_cross_dissolve final : public TBlendForeBackRasterFx {
 public:
   ino_blend_cross_dissolve() : TBlendForeBackRasterFx(false) {}
   ~ino_blend_cross_dissolve() {}
-
-  bool propagatesClippingMaskAlpha() const override { return true; }
-
   void brendKernel(double& dnr, double& dng, double& dnb, double& dna,
                    const double upr, double upg, double upb, double upa,
                    const double up_opacity,
--- a/toonz/sources/stdfx/ino_blend_cross_dissolve.cpp
+++ b/toonz/sources/stdfx/ino_blend_cross_dissolve.cpp
@@ -11,6 +11,9 @@ class ino_blend_cross_dissolve final : public TBlendForeBackRasterFx {
 public:
   ino_blend_cross_dissolve() : TBlendForeBackRasterFx(false) {}
   ~ino_blend_cross_dissolve() {}
+
+  bool propagatesClippingMaskAlpha() const override { return true; }
+
   void brendKernel(double& dnr, double& dng, double& dnb, double& dna,
                    const double upr, double upg, double upb, double upa,
                    const double up_opacity,
--- a/toonz/sources/stdfx/ino_blend_subtract.cpp
+++ b/toonz/sources/stdfx/ino_blend_subtract.cpp
@@ -12,6 +12,11 @@ public:
   ino_blend_subtract()
       : TBlendForeBackRasterFx(true, true) {}  // with alpha_rendering switch
   ~ino_blend_subtract() {}
+
+  bool propagatesClippingMaskAlpha() const override {
+    return m_alpha_rendering->getValue();
+  }
+
   void brendKernel(double& dnr, double& dng, double& dnb, double& dna,
                    const double upr, double upg, double upb, double upa,
                    const double up_opacity,
--- a/toonz/sources/stdfx/ino_blend_subtract.cpp
+++ b/toonz/sources/stdfx/ino_blend_subtract.cpp
@@ -12,11 +12,6 @@ public:
   ino_blend_subtract()
       : TBlendForeBackRasterFx(true, true) {}  // with alpha_rendering switch
   ~ino_blend_subtract() {}
-
-  bool propagatesClippingMaskAlpha() const override {
-    return m_alpha_rendering->getValue();
-  }
-
   void brendKernel(double& dnr, double& dng, double& dnb, double& dna,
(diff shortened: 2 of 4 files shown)


RattusaurusRex:
Cross Dissolve turns black in this PR:
image
Current behavior: 
image

I'm not immediately seeing any issues with the other blending mode effects.  Could this fix also be applied to the Matte effects?

I'm having a crash when trying to preview at 32 bit, but that seems unrelated to this PR since I also get a crash in the OT nightly (no crash in Tahoma).

konero:
Thanks for testing! I fixed Cross Dissolve, it just needed to declare that its calculated alpha must be preserved like Subtract.

For Matte effects, is there some examples I can see for the issues so I got something to compare with? I can't remember what it was with Matte effects.

RattusaurusRex:
I mentioned the haloing issue with Matte In + Multiply in my original bug report:
image
image

konero:
So far from testing Matte appears to be working as intended, I think this might be a compositing choice. Even though it produces that seam, I can only speak technically the nodes are doing what they should.

Since it's at least seemingly unrelated to clipping mask, a new issue can track this one for a way around it. 👍

RodneyBaker:
Added @shun-iwasawa as requested reviewer as this fix will (intentionally) change pixel output.

Situation | Previous render expected to change?
-- | --
Clipping Mask off | No, essentially same path
Clipping on, Back alpha = 1.0 | No, essentially same math
Clipping on, Back alpha between 0 and 1 | Yes, intentionally
Clipping on, Back alpha = 0 | Normally still visually transparent, but hidden RGB handling changes
Soft/antialiased Back edges | Yes, this is the primary target of the fix

konero:
Refactored the fix a bit and now the fix is hopefully much more simplified. The exact fix now...

OT stores these pixels in premultiplied form, their RGB values are already multiplied by their alpha.

`effective Fore opacity = Fore opacity * Back alpha`

But the Back RGB was already reduced by Back alpha, and the blend kernel also performed alpha aware compositing. So around soft Back edges the Back alpha therefore affected the calculation more than once. Which produced the darker fringe/multiplied-looking pixels (see #5408).

The process now is:
- Save Back alpha `Ab`
- If `Ab` is zero, output transparent
- Divide Back RGB by `Ab` to recover its straight color
- Temporarily treat the Back as opaque (alpha 1)
- Run the normal blend kernel using the original Fore opacity
- Multiply the resulting premultiplied RGBA by `Ab`

Conceptually like:
```
working Back = (Back RGB / Ab, alpha 1)
working result = blend(working Back, Fore)
final result = working result * Ab
```

Which makes Back alpha act exactly once, as the clipping boundary.

So the propagation overrides approach I tested before was unnecessary thankfully which has been refactored out, now the fix is much simpler. There is no real exceptions now, each kernel already describes its intended alpha behavior.

RattusaurusRex:
The halos are back in the update (Add): 
image

Previous version: 
image

konero:
I just tested the GitHub build (https://github.com/opentoonz/opentoonz/actions/runs/34466836550), seems to be fine on my end, could you maybe have accidentally run a different build? I tested with default settings. 🤔 

image

image Screenshot 2026-09-11 010208

RattusaurusRex:
I've figured out how I was getting the halos -- I had turned on 32 bit in the Output/Preview Settings.  That seems to still have halos (in current and previous versions).

konero:
Ah that was it, I could reproduce it, it should be fixed now. For some reason OT doesn't always swap color space, I have to force a few preview re-renders or close the app for reliability.

Since the math is really tough on this one, can't say for 100% sure if this is a great fix, it's a very small fix though so if a problem does occur when a bigger 32 bit floating point project is tested it'll be easy to revert (without affecting 8/16 bit). But from what I've been able to throw it at, it seems OK so far.

Thanks for testing!

PR #7061, opened 2026-08-11
Optimize brush engine and pointer events

konero:
Intermittently, fast pen strokes such as arcs, curves, and circles could appear angular or develop polyline-like cusps. I have not ruled out hardware/driver differences (for example AMD vs. Intel), but the working theory is that the brush engine sometimes received sparse tablet input and reconstructed the stroke too literally.

This update uses smoother freehand reconstruction while keeping high-frequency pen input limited to just brush tools, trying not to interfere with selection/editing tools and retain throttled behavior.

**Changes:**

- Uses centripetal Catmull-Rom freehand interpolation to reconstruct smoother natural curves while reducing loops and overshoot on sparse of fast input (usually very reliable)
- Uses monotone pressure interpolation so pressure changes remain stable without artificial peaks or dips
- Preserves intentional sharp corners, handles duplicate points and stroke endings and now carries stylus tilt through the stroke path (even if we don't utilize it yet)
- All brush tools receive every distinct stylus sample now
- Keeps throttling for non-brush tools to avoid unnecessary responsiveness changes in selection, transform tools, etc.
- Adds a freehand regression test for dev builds (we had no way to get feedback otherwise)

**Must Test:**

Please test with a stylus/pen where possible the following:

- Fast strokes in Raster, Toonz Raster and Toonz Vector levels, draw curves, circles, arcs, scribbles and long fast lines
- Slow strokes and deliberate sharp shapes like zigzags, hair strands, triangle tips, boxes and corners
- Pressure, try going from light > hard > light and try rapid changes
- Many short strokes drawn very fast in succession, we want to confirm drawing doesn't become unresponsive
- Adjusting control points, transforming raster selections (using the stylus/pen)
- Check against a release or night to check if the feel is different, look for any lag, floaty or delay
- Mac drawing, I don't have one so I can't test

**Benchmark Note:**

In my scribble tests, the current release build used about 6-7% of CPU, while this branch used about 0.7-1.7%. The exact reason for this I still don't know, I don't want to make a performance claim here yet, and I checked in-case we're now under-processing something, but didn't find anything.

**Example of Issue:**
image

**Facts:**
- AI was used (Codex Sol 5.6, Terra 5.6, Luna 5.6) to:
- - Search the codebase
- - Create a unit test
- - Port much of my Javascript/Python code to C++
- - All code has been checked by me; anything ported by Codex that is too complex for me to understand was double-checked using Gemini Pro

@RattusaurusRex 🙏


konero:
Latest change: Optimize `calcTangent()` by reusing cumulative lengths already stored in `TTrackPoint` and precomputing the 80 degree cusp cosine, avoids redundant distance and trigonometry calculations without changing modifier pipeline at all, etc.

Adds a small regression check for uneven spacing/repeated samples/angles around the cusp threshold, for the test.

```
Tangent calculation:             10.18ms -> 8.72ms  (14% faster)
Full freehand modifier workload: 73.53ms -> 69.05ms (6% faster)
```

RodneyBaker:
You've got some very interesting things going on here @konero 

A point of interest that I don't yet understand related to `calcTangent()`:
For a tangent stored for one point that can be/is also used as the ending tangent of a segment:
For uniformly spaced points those scales are nearly interchangeable.
For irregular tablet spacing they are not interchangeable.

The irony is that uneven sample spacing is precisely the situation this PR is trying to improve.

I need a way to communicate this more effectively so might resort to recording a video.

konero:
Will put this on hold, the deeper I go into the code the more I see that contributes to all the jank, just layers of hotfixes and workarounds, there's an invisible bug where if the brush gets sudden focus on the canvas it malfunctions (unrelated to this) but will be interesting to look into.

Will see if I can break it out into tiny incremental steps over time.

RodneyBaker:
As much progress as this PR has I you are making the right choice.

The 'jank' you mention I see more as exploration into an interim fix.  Probing at that fix.
Your instinct is on track as will be better if we can get to the full fix.

Issue #4632, opened 2022-12-03
Allow User to Show/Hide Macro FX Properties


konero:
### Is your feature request related to a problem? Please describe

When creating macro, often is the case that most of the properties are not required to be adjusted by the end-user, but instead maybe one or two values. Right now, it can be frustrating figuring out which value was intended to be adjusted, especially if the macro is shared with other users without a note.

It's something that can be done with Toon Boom Harmony, and is very useful.

### Describe the solution you'd like

The idea I've come up with so far is improving the macro like so...

This is the current settings panel for a Blur node:

![Default Settings Panel](https://user-images.githubusercontent.com/19820721/205416011-7772c4ec-a973-4958-a071-a6d411560e8c.jpg)

If we collapse it into a macro and we bring that node back up again in the settings panel, a checkbox will be added next to each value that can be shown on the front page of the macro:

![Checking Box Brings Value to Front Page](https://user-images.githubusercontent.com/19820721/205416105-8bede7e3-dfe5-4873-a839-da9bf2acfcf0.jpg)

The macro front page will add the node as a tab, but the name of the tab will use the custom name set by the user for the node, not the generic node type name to make it more descriptive:

![Adding Other Properties Creates Tabs](https://user-images.githubusercontent.com/19820721/205416153-9254a5c9-c1cb-49d6-9846-96fecfdda22d.jpg)

PR #7062, opened 2026-08-11
Pin parameters to front-page of macros

konero:
This is a continuation of @doangelshavewings's pull request, based on my request in #4632.

I tested it and the theory worked, but there was issues: params using type name instead of user name, and placing a duplicate of all the params in a box at the bottom of each tab wasn't very intuitive.

I've modified it to fix the interface issues, to no longer auto-switch to the pin page, and a checkbox now appears beside each parameter instead of grouped at the bottom (same layout as Resolve's). Each param now is listed under a category instead of being truncated with everything randomly, and listed like: `Custom Name (Type Name)`. I removed this feature's actions off the scene's undo stack, since there's no point being able to undo an interface option.

A macro that has pinned params will not be compatible with older versions of OpenToonz because there will be a `<exposedparams>` block, the scenes containing them will still load, but the macro will likely be removed from it.

**Example:**

https://github.com/user-attachments/assets/5fb2b2f4-6ae7-44e9-873a-1c425de5b0c2

### DIFF
--- a/toonz/sources/common/tfx/tmacrofx.cpp
+++ b/toonz/sources/common/tfx/tmacrofx.cpp
@@ -79,13 +79,20 @@ std::vector<TFxP> sortFxs(const std::vector<TFxP> &fxs) {
 
 // raccoglie tutti i parametri dai vari TFx e li assegna anche alla macro
 void collectParams(TMacroFx *macroFx) {
-  int k;
-  for (k = 0; k < (int)macroFx->m_fxs.size(); k++) {
-    TFxP fx = macroFx->m_fxs[k];
+  const std::vector<TFxP>& fxs = macroFx->getFxs();
+  for (int k = 0; k < (int)fxs.size(); k++) {
+    TFxP fx = fxs[k];
     int j;
     for (j = 0; j < fx->getParams()->getParamCount(); j++)
       macroFx->getParams()->add(fx->getParams()->getParamVar(j)->clone());
   }
+}
+
+bool isMemberFx(const TMacroFx* macroFx, const TFx* fx) {
+  const std::vector<TFxP>& fxs = macroFx->getFxs();
+  return std::find_if(fxs.begin(), fxs.end(), [fx](const TFxP& candidate) {
+           return candidate.getPointer() == fx;
+         }) != fxs.end();
 }
 
 }  // anonymous namespace
@@ -254,6 +261,7 @@ TFx *TMacroFx::clone(bool recursive) const {
   TMacroFx *clone = TMacroFx::create(clones);
   clone->setName(getName());
   clone->setFxId(getFxId());
+  clone->m_exposedParams = m_exposedParams;
 
   // Copy the index of the passive cache manager.
   clone->getAttributes()->passiveCacheDataIdx() =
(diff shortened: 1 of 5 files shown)


RodneyBaker:
Thanks @konero 
I'm tagging this with v2.0 Candidate due to the breaking change but that doesn't suggest it can't move forward earlier.  The anticipation is that v2.0 will include many breaking changes so that's just an easy target to assign.

konero:
@RodneyBaker Semver's "major = breaking" convention (if that's what you're describing) is really about API/interface stability, for anyone building plugins or tools against an app who need a signal that things they rely on might have changed.

Since a macro is content and not API a newer format not displaying in older app versions is a forward compatibility limitation, it just needs a note in the changelog for whichever the next release is (1.9 or otherwise), it doesn't break any contract that anyone depends on for tooling. If that makes sense?

Krita 5 = Qt5
Krita 6 = Qt6 (scripts/plugins for Qt5 may not work without adapting)

RPG Maker MV
RPG Maker MZ (different core, MV plugins won't work here without adapting)

DarrenTAnims:
This looks really useful.
However, to me, from using other software, the terminology isn't right.
Pinning is something that a user chooses to do in the interface. There's usually a pin button and the user requests to pin something somewhere.
What this feature is, is showing a tab with adjusted/set values, so a better name for the tab could be: Changed or Edited or similar.
I think that'd make it more obvious what it's showing.
But great feature. Very useful.

RodneyBaker:
Agreed.
We need to discuss how we can improve this. 

Since users can expect significant change with v2.0 my labeling is largely just to suggest 'pay attention, with this change something significant may be changing'.  Your mention that any  release containing the change from this PR will cause scenes not to open in older releases is significant.  The degree of that significance... identifiable.  We've had several of this type of 'breaking change' that didn't bump the release number before and they didn't cause a bump in version numbering.  A key there does appear to be how that change was managed.  Being optional, reversible, documented; the change can be avoided and undone leading to compatibility again.  This touches upon what you mention about API, etc.  

All this just to suggest my use of the term 'breaking change' is mostly meant to say, this change will break something; the layman's reading versus the technicians.  In this case that break being the change to compatibility with older releases. 

My current labeling of PRs is mostly just to keep the PRs moving forward.

I will suggest where this PR is more broadly associated with other significant changes in the Schematic that might be additional reason to target a v2.0 release.  For that we'd have to look more deeply into what is being targeted for change.

Bottom line:  In the end, @shun-iwasawa will decide the version number.  Our role is to supply the change.

konero:
> @DarrenTAnims [...] from using other software, the terminology isn't right. Pinning is something that a user chooses to do in the interface. There's usually a pin button and the user requests to pin something somewhere. What this feature is, is showing a tab with adjusted/set values, so a better name for the tab could be: Changed or Edited or similar. I think that'd make it more obvious what it's showing.

That's not right, you might have misunderstood. The feature is pinning the parameter to the front page, from there you can adjust it. It's a feature from Toon Boom Harmony/DaVinci Resolve. The idea is instead of looking through 20 tabs for one value to adjust, you pin it to the front of the macro. If you name the node, the context is clear for what the value is for.

```
Halation Softness (Blur)
- Value: 20px
```

> @RodneyBaker Your mention that any release containing the change from this PR will cause scenes not to open in older releases is significant.

I didn't say that anywhere. The scene will open, the macro just won't appear. It's a basic forward compatibility limitation, we have them all the time, you don't reserve that for contract changing builds otherwise we'd have no new features.

> Bottom line: In the end, @shun-iwasawa will decide the version number. Our role is to supply the change.

Which will be as I explained it.

RodneyBaker:
I stand corrected. 
You said:

> A macro that has pinned params **will not be compatible with older versions of OpenToonz** because there will be a <exposedparams> block, the scenes containing them will still load but the macro will likely be removed from it.

Emphasis added.

> ...there will be a <exposedparams> block, the scenes containing them will still load but the macro will likely be removed from it.

Alll good.  We can confirm how likely through testing.

DarrenTAnims:
> That's not right, you might have misunderstood. The feature is pinning the parameter to the front page, from there you can adjust it. It's a feature from Toon Boom Harmony/DaVinci Resolve. The idea is instead of looking through 20 tabs for one value to adjust, you pin it to the front of the macro. If you name the node, the context is clear for what the value is for.

Ah - after rewatching the video, I can now see that the tickbox is the pinning mechanism.
It's because it was a tickbox that threw me.
Is it possible to make it more pin looking or something to make it clear what it does?

RodneyBaker:
I see this PR as merge ready although would like to see the following verified (as I haven't yet tested them myself):

- Existing macros with no pinned parameters behave exactly as before.
- Pin one parameter → Pinned tab appears → editing it changes the actual underlying FX parameter.
- Pin several parameters from different macro nodes → labels/grouping are correct.
- Unpin parameters, especially the last pinned parameter → Pinned tab disappears cleanly without changing to a wrong tab or crashing.
- Save scene → close OpenToonz → reopen → pins and values survive.
- Duplicate/copy or otherwise clone the macro → exposed parameters remain attached correctly.

Finally, we should test nested macros as this PR explicitly contains recursion for them.

If someone can vouch for these testing successfully that will be appreciated.

Issue #3314, opened 2020-05-27
Browser navigation improvments 


ghost:
This request  is similar to the behavior that we could find in the windows explorer for browsing files:

1. Would it be possible to be able to navigate our files with the keyboard. I mean by pressing a letter you can browse the files that refer to the letter.

2. Is it possible to keep the blue color of the last selected file maintained when you return to the "browser" room after changing rooms.

3. When using the flipbooks preview in addition (right click> view) allow to use the arrows to scroll through the different content files

This would help to find what we want more quickly and to better visualize things as well.

DarrenTAnims:
I'd agree that the OT browser should work like windows explorer. On top of the above:

- Double clicking to move into a folder, not a single click. The number of times I've double clicked and moved down 2 folders is just irritating. I'm sure it's a feature to save time, but it doesn't.
- Double clicking to open each file type. I know that has been done for scenes. Hopefully, it'll be done for the other files too.
- Backspace delete to move up a folder
- Cursor keys to move up and down the list.
- Tab to move from the folder list to the file list.

Or we could just use the standard browser and save time trying to reinvent the wheel. I'm sure this was revolutionary in its time, but you can just show a standard browser window so easily.

beeheemooth:
Really! What could be a “browser” without a search engine? ^__^

openanim:
I agree ,maybe single click or double click could be an option, it seems to me that there is already a similar option with cells in the preferences. A search bar would be great !

turtletooth:
I got double clicking working for folders and viewing files.

openanim:
On vector level :

In the browser, a sequence of numbers with 1 digit does not seem to work well with OT when the sequence changes to 2 digits for example:
1,2,3 ... 9 ... 10 or_1, _2, _3 ..._9 ..._10 (scenario 1)

but works in this way 01, 02, 03..09..10 or_01 ...._ 10 when all numbers are 2 digits. I mean, _10 PLI file will be after _09 file, (scenario 2).

I think OT in the 1st scenario should display the continuation as in the second in the browser. This is, I think, the expected behavior and basic file naming.

Zenahr:
Could we have an option to have OT use the default file explorer instead?

PR #7063, opened 2026-08-12
file browser enhancements

Ahl76:
image

related to #3314

This PR brings several improvements to the File Browser to make it easier to find your work, spot your productions, and move around project files with less friction. You can leave them unused if they do not fit a particular workflow.

I decided not to split this into several PRs, because of the dependencies between the parts and how many PRs that would have made. If you would rather I break it into smaller PRs for review, I can do that.

This is a fairly large proposal and it touches sensitive areas that will need careful checking, especially rendering and thumbnail display. Some of the icons already come from existing OpenToonz work; I reworked them quickly, mainly to get the functional side in place for review. They can be refined later if needed. I hope this doesn't cause any side effects.

**Proposal**

An optional Advanced Display mode for the File Browser, with the same thumbnail improvements in the Scene Browser.

With Advanced Display off, behaviour stays close to today. With it on, you get larger thumbnails and a size control, thumbnail background options, search, type filters, favorites, per-file thumbnail background when needed, and play FPS / loop for multi-frame previews. Thumbnails can adapt better to each file so levels and renders are easier to read (similar in spirit to the Flipbook and the various viewers). Right-click the advanced toolbar to show or hide individual controls (via gui / show / hide.) Selection in the File Browser is kept when you switch rooms or come back during the same session.

Project folders get letter shortcuts when Advanced Display is on. The letter system is not the most intuitive; it could be refined later if the feature proves useful.

A File Info panel can sit beside the browser with file details and a larger preview. It can be hidden from the toolbar or from a right-click on the panel.

Rescale is available from the context menu for rescalable raster files. It follows ideas already used in OpenToonz, and covers cases where you need to enlarge or shrink several drawings in one go.

For movie and video files, browser thumbnails still come from the clip, but only one frame is decoded for the preview, so long renders show up faster. 

Preferences

Open Folders in File Browser with Double-Click (optional).

**Notes**

Advanced Display is optional.

**Limitations**

Large thumbnails can be slower on heavy folders.

Edit Added :

Updates

- The File Browser and Scene Cast remember tree and File Info splitter widths. Default tree width matches Scene Cast on first launch and when a room is reset.

- The File Info side-panel thumbnail follows the Current Frame slider. The previous image stays until the next one is ready.

- HD thumbnails stay sharp at large size, including vectors and high-DPI. Play from the thumbnail keeps the same quality as the still image.

- The existing Info command opens a dockable File Info window, one per file as before. The File Browser side panel is unchanged.

- Folders can be pinned to the tree : right-click a folder to pin or unpin it. Pinned folders appear under Favorites at the top of the tree.


RodneyBaker:
You had me at: 

> With it on, you get larger thumbnails

Thanks @Ahl76 for leaning into this.
As you state this will take time to look into so your patience is appreciated.

I'm tagging this as a v2.0 Candidate but that does not mean it has to wait.
The optional nature of the change particularly makes it easy to fit as transitional which would make it ideal for the penultimate release (presumably v1.9).  

Adding this to the next release will allow users more than ample time to settle in to the changes, test, give feedback and solidify these and any other  file browser enhancements necessary.

PR #7066, opened 2026-08-13
Add optional Camera Capture sound

RodneyBaker:
Implements:  https://github.com/opentoonz/opentoonz/discussions/6480

Adds an option for sound after a successful Camera Capture in the Camera Capture window.

Uses the existing camera_snap.wav sound supplied by the similar feature in Stopmotion Control.

Functional testing:

Open Camera Capture and confirm Play Sound on Capture appears and defaults off.
Enable option. Take a normal capture:  one shutter sound should occur after successful capture.
Enable the interval timer:  each successfully captured frame should produce one sound.
Disable the option: captures should be silent.
Cancel/fail an overwrite or other capture:  no success sound should be produced.
Close and reopen Camera Capture: the sound setting should persist.

Tested successfully in opentoonz-dev.
ChatGPT 5.6 used for research, testing and processing of code.

### DIFF
--- a/toonz/sources/toonz/penciltestpopup.cpp
+++ b/toonz/sources/toonz/penciltestpopup.cpp
@@ -85,6 +85,7 @@
 #include <QKeyEvent>
 #include <QCommonStyle>
 #include <QTimer>
+#include <QSound>
 #include <QIntValidator>
 #include <QRegularExpressionValidator>
 
@@ -127,6 +128,7 @@ TEnv::RectVar CamCapSubCameraRect("CamCapSubCameraRect", TRect());
 TEnv::IntVar CamCapDoAutoDpi("CamCapDoAutoDpi", 1);
 TEnv::DoubleVar CamCapCustomDpi("CamCapDpiForNewLevel", 120.0);
 TEnv::StringVar CamCapFileType("CamCapFileType", "jpg");
+TEnv::IntVar CamCapPlayCaptureSound("CamCapPlayCaptureSound", 0);
 
 namespace {
 
@@ -1624,6 +1626,8 @@ PencilTestPopup::PencilTestPopup()
 
   m_saveOnCaptureCB =
       new QCheckBox(tr("Save images as they are captured"), this);
+  QCheckBox* playCaptureSoundCB =
+      new QCheckBox(tr("Play Sound on Capture"), this);
 
   QGroupBox* imageFrame        = new QGroupBox(tr("Image adjust"), this);
   m_colorTypeCombo             = new QComboBox(this);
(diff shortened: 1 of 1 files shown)


RodneyBaker:
I've discovered a problem with Camera Capture related to the interval timer that I'll post a fix for as a separate PR.  It's not clear when it broke.

shun-iwasawa:
I've confirmed that the feature itself is working correctly.
However, I feel it is inappropriate for the checkbox to be located inside the “File” group box.
How about moving this option to the menu that appears when you click the gear icon at the top of the popup?

image

RodneyBaker:
I've no problem with moving the option to that location.
As the option is a bit hidden would it be reasonablle to have the sound on by default?

I think I see another potential issue.
Moving the sound control solely into this gear menu means the user visible option will effectively be Windows only under the current UI logic, whereas our existing checkbox was available wherever this version of Camera Capture is built.  I'll investigate that.

Added:  It seems part of the gearbox option may currently be Windows only which might need to be addressed also.  We don't want the sound option to be available only for certain cameras or operating systems based on what is currently selected.  I'm pretty sure we can make it appear for any camera selected.

RodneyBaker:
Shun,
I've moved Play Sound on Capture to the gear menu as suggested.

I have also attempted to address the issue where the gear button itself is only being created on Windows.  Because `createOptionsMenu()` is already outside the Windows specific code I've attempted to properly place the option for sound in that area.

The gear button itself has now been made available on all platforms while keeping Video Capture Filter Settings and the use of appropriate camera specific drivers Windows only. This keeps the sound option cross platform while allow other appropriate options reachable on macOS/Linux.

This does broaden the change, so if preferred the cross platform gear menu change can be handled separately.

shun-iwasawa:
I forgot that the gear icon isn’t displayed in the macOS version. My apologies!
To implement this option, I think it would be fine to display the gear icon in the macOS version as well. (In that case, the menu would contain only this command.)

As for the default state of this option, personally, I’d prefer it to be turned off. The desks used for pencil test workflows are usually located directly next to the desks of animators at animation studios, and I imagine they’d want to avoid disturbing their colleagues’ concentration with the shutter sound.
On the other hand, I think there are many situations where the shutter sound is actually beneficial—like in workshops where children are creating stop-motion animation.

RodneyBaker:
Thanks Shun,

As a narrow follow up here's where I believe we are with this update:
- Windows keeps the full gear menu and MacOS gets the gear with only the sound toggle
- Linux remains unchanged
- The sound default stays 0  (Off)

shun-iwasawa:
LGTM Thank you @RodneyBaker for adding this feature!

PR #7067, opened 2026-08-13
Fix Camera Capture interval and onion skin controls

RodneyBaker:
Fixes the Camera Capture Interval Timer and Onion Skin opacity controls remaining disabled after their sections are enabled.

While researching and testing #7066 I discovered a regression introduced in #4250, when the separate checkboxes were replaced with checkable QGroupBox controls. The old explicit setDisabled(true) calls remained, preventing the group boxes from re-enabling their child fields.

This removes those two obsolete disables so the group boxes control their child fields as intended.


RodneyBaker:
Proceeding with this fix.
I'm not sure why we didn't discover it sooner.

PR #7068, opened 2026-08-13
Fix fast strokes becoming angular

konero:
A longstanding issue has been fast drawn curves like arcs, circles etc. become angular, circles can end up looking hexagonal. Not everyone experiences this and it hasn't been established why yet.

This should fix the issue in some part by resolving two areas:
- Tangent recalculation now invalidates and regenerates the preceding segment instead of retaining geometry generated with a provisional endpoint tanget: This fixes the issue with basic brushes (non-MyPaint)
- Tablet source timestamps are now preserved and mapped to `TToolTimer` (prevent coalesced event dispatch timing from producing random artificial velocity spikes found during profiling: Fixes the issue for MyPaint brushes. This seem to have been removed during inclusion of Assistants tool.

Anyone who experienced the issue, please give it a test. It may still produce issues on WinInk, my test target was WinTab devices (Wacom) first.

**Example of the issue (left):**

image

### DIFF
--- a/toonz/sources/include/tools/tool.h
+++ b/toonz/sources/include/tools/tool.h
@@ -9,6 +9,7 @@
 #include "toonz/imagepainter.h"
 #include "toonz/tapplication.h"
 #include "tools/cursors.h"
+#include "tools/tooltimer.h"
 
 // TnzCore includes
 #include "tcommon.h"
@@ -101,6 +102,8 @@ public:
   Qt::MouseButton m_button;
   QPointF m_mousePos;  // mouse position obtained with QMouseEvent::pos() or
                        // QTabletEvent::pos()
+  TTimerTicks m_time;  // Source event time mapped to TToolTimer ticks,
+                       // preserving tablet input timing when events are batched.
   bool m_isTablet;
   bool m_isHighFrequent;
 
@@ -110,6 +113,7 @@ public:
       , m_modifiersMask(NO_KEY)
       , m_buttons(Qt::NoButton)
       , m_button(Qt::NoButton)
+      , m_time(0)
       , m_isTablet(false)
       , m_isHighFrequent(false) {}
 
@@ -121,6 +125,7 @@ public:
   Qt::MouseButtons buttons() const { return m_buttons; }
   Qt::MouseButton button() const { return m_button; }
   QPointF mousePos() const { return m_mousePos; }
+  TTimerTicks time() const { return m_time; }
   bool isTablet() const { return m_isTablet; }
   bool isHighFrequent() const { return m_isHighFrequent; }
 
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -368,7 +368,7 @@ bool FullColorBrushTool::preLeftButtonDown() {
 void FullColorBrushTool::handleMouseEvent(MouseEventType type,
                                           const TPointD &pos,
                                           const TMouseEvent &e) {
-  TTimerTicks t = TToolTimer::ticks();
+  TTimerTicks t = e.time() > 0 ? e.time() : TToolTimer::ticks();
   bool alt      = e.getModifiersMask() & TMouseEvent::ALT_KEY;
   bool shift    = e.getModifiersMask() & TMouseEvent::SHIFT_KEY;
   bool control  = e.getModifiersMask() & TMouseEvent::CTRL_KEY;
--- a/toonz/sources/tnztools/modifiers/modifiertangents.cpp
+++ b/toonz/sources/tnztools/modifiers/modifiertangents.cpp
@@ -97,14 +97,18 @@ TModifierTangents::modifyTrack(
   int start = track.size() - track.pointsAdded;
   if (start < 0) start = 0;
   
-  // update tangents
+  // --- update tangents
   int tangentStart = start - 1;
   if (tangentStart < 0) tangentStart = 0;
   intr->tangents.resize(tangentStart);
   for(int i = tangentStart; i < track.size(); ++i)
     intr->tangents.push_back(calcTangent(track, i));
 
-  // update subTrack
+  // --- update subTrack
+  // Re-emit the preceding point because its newly calculated tangent also
+  // changes the segment leading into it.
+  if (start > 1) --start;
+  
   subTrack.truncate(start);
   for(int i = start; i < track.size(); ++i)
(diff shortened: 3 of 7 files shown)


RodneyBaker:
I'm almost exclusively focusing on the code first rather than actually drawing for reasons I'll explore in the next post.

Regarding the two issues mentioned.

Spatial/tangent correction 
`TModifierTangents` recalculates the tangent of the previous input point when a new point arrives **but until this PR it regenerated the output track only from the new point onward**.

This PR (#7068) backs the regeneration point up by one so the segment affected by that newly calculated tangent is regenerated too. (Repairing an off by one invalidation/regeneration bug)

Temporal/input timing correction
As of this PR (#7068), `TMouseEvent ` now carries an event time. Tablet events preserve their Qt source timestamp, map it into OpenToonz's nanosecond `TToolTimer` domain, and Full Color/Toonz Raster brushes feed that preserved time into `TInputManager` instead of sampling `TToolTimer::ticks() ` only when the event reaches the brush.

These two fixes address different causes that can produce (similar looking) bad strokes.

Tracing some history leading to these changes:
Timing
old FullColorBrush
→ MyPaint owns an elapsed timer
→ #4852 Assistants refactor
→ common `TInputManager`
→ timestamp sampled at OpenToonz dispatch
→ #7068
→ common `TInputManager` remains, but receives actual tablet source timing.

That is an Assistants era architectural regression.
This PR accepts the previous Assistants refactor (#4852) and seeks to work with actual tablet source time.

Tangents
original Assistant tangent modifier
→ correctly backs regeneration up one point
→ 2023 Assistants for ToonzVectorBrush refactor
→ tangent recalculation backs up, output regeneration doesn't
→ #7068
→ output regeneration backs up again.

This review:
ChatGPT 5.6 used in history tracing (consider errors can be geneated in any trace or in my interpretation of that trace and should not be considered reliable without verification/references) 

Try/Fix assessment:  This PR appears to be a solid direction forward to resolve the stated problem of fixing fast tablet strokes.  However, see next post...

RodneyBaker:
Framing this as a pipeline problem as opposed to a stroke smoothing problem may help in digging deeper to full cross platform fix.

Looking at cross platform handling:
On Windows, OpenToonz currently forwards distinct tablet moves continuously and merely marks some as “high frequency”; on the non Windows path the process explicitly gates tablet moves with a 20 ms timer (≈50 Hz). 

That gating is enough to polygonize a fast circle before any interpolation algorithm even sees the data and this matches old Linux reports where the same MyPaint brushes were reported to be smoother in other programs.

The important part (?)
This PR (#7068) is valuable because it fixes two correctness problems in that chain: the stale preceding spline segment and incorrect event timing. However, this important and legitimate fix cannot repair positional information that OpenToonz never receives or deliberately throws away.

**The strongest remaining problem then: Even with these two fixes OpenToonz is still throwing away input.**

This... for your consideration.

Added:  I'm investigating something Opentoonz already has in place that can potentially handle this problem:  `if (e.isHighFrequent()) return;`

RodneyBaker:
I hesitate to mention this because I get better performance drawing in Opentoonz than in Tahoma2D (that does not appear to be the case for everyone).

T2D does attempt to deal with the high frequency issue by commenting out code.

```
a.setAttribute(Qt::AA_CompressHighFrequencyEvents);
// a.setAttribute(Qt::AA_CompressTabletEvents);
```
I don't see this as a solution so much as an additional hint.

RodneyBaker:
I have not yet looked at the specific MyPaint exclusive aspects of these equations beyond this notice.  From libmypaint:

```
/**
 * mypaint_brush_stroke_to:
 * @dtime: Time since last motion event, in seconds.
 *
 * Should be called once for each motion event.
 */
```

Ahl76:
Testing this PR, I can confirm that the fix works fine on my old Wacom Cintiq Companion 2. The MyPaint brushes that showed this issue no longer do, with a few rare exceptions.

The result is different on my Huion. On my end, I notice an issue with stroke consistency and quality (slight variations can be seen throughout) in addition to the remaining angular/polygonal issue. The NB build gives me better results with both standard and MyPaint brushes. I have attached a few tests below. These are all tests related to my Huion Kamvas. The first image in each case was made using the OT Nightly Build, and the second on the current PR's artifact.

image

image

image

image

image
image

konero:
@Ahl76 Thanks for testing. Targeting non-Wacom devices is difficult for me since I don't have any on hand. I did some reading and seems the difference might come from how each tablet driver and input API reports or batches event timestamps, WinTab and WinInk can produce different timestamp patterns, so it might be the cause. It's a blind fix for me though because I can't actually test with an actual device, but I tried to handle unavailable and repeated timestamps.

Hopefully it's better. Although there is still a visual issue where the stroke might look fatter or thinner in different segments even with pressure disabled, that actually has nothing to do with this fix (it's present in the stable/nightly) and that's caused from a different place altogether I believe, unless this fix on certain devices is making the issue more obvious.

To double-check, make sure to zoom to 100% when testing, fractional values will make strokes look broken.

RodneyBaker:
The following is primarily Windows Ink vs Wintab related.

Because strokes appearing to be heavily influenced by the drivers and tablets I submit the following breakdown related to settings that have proved useful for some users.

image

As these combinations can be confusing, for testing purposes, I would first try tablet settings first as it that appears to work for most users.
  
Note:  This chart covers Windows 11 usage where there is no longer a direct means of turning Windows Ink off in Windows itself.  In Windows 10 there is yet another option to consider and my recommendation would be to turn Windows Ink off in Win 10 settings.  If that works:  done.  If not, explore the additional Windows Ink settings of the specific tablet and of Opentoonz.

Ahl76:
@konero 
It works better with the latest commit. There is no longer a glaring difference between the MyPaint brushes. The results are pretty close on my end.

But the difference is noticeable on native brushes, whether Toonz Raster, Toonz Vector, or standard Raster, although it's better than before.  

Disabling Windows Ink from my Huion settings or OT preferences didn't help.

Not sure that help a lot but here are a few screenshots with the native brushes, in the following order: Vector, Toonz Raster, and standard Raster

image

image

image

image

image
image

RodneyBaker:
@konero 
I've closed most of my posts as being outdated/noise as they are chasing theories more helping.  While tablet drivers and the like can play a significant factor in this it's good if we can remove those from the equation during tests.  From all indications, you are on the right path toward a fix with this PR.

I started https://github.com/OpenAnimationLibrary/opentoonz-dev/pull/127 to see if running diagnostic tests outside of the theorecticals might reveal anything of use.

PR #7072, opened 2026-08-15
Fix: Recognize Tahoma2D project files

RodneyBaker:
### Problem
This is designated a fix but addresses an external problem introduced in Tahoma2D.
OpenToonz scenes stored in a Tahoma2D project can fail to open even when the .tnz scene itself is otherwise compatible or even the same files.  This unnecessaryly creates incompatibility with the Opentoonz .tnz file extension and confusion on the part of users.

Background
Tahoma2D uses the fixed project descriptor name tahomaproject.xml, while OpenToonz currently recognizes only its historical *_otprj.xml, *_prj63ml.xml, *_prj6.xml, and *_prj.xml naming conventions.

This fixed proejct descriptor change introduced an incompatiblity requiring users to manually rename the tahomaproject.xml with the proper suffix before otherwise compatible projects can be opened in OpenToonz.

### Changes
`getProjectFile()` now explicitly recognizes tahomaproject.xml when no standard OpenToonz project descriptor is found.
`TProject::isAProjectPath() ` accepts the exact tahomaproject.xml filename so the discovered project is allowed through the normal loading path.

Existing OpenToonz project naming and discovery remain unchanged.

This PR is intentionally limited to project discovery compatibility. Specific .tnz scene additions (in Tahoma2D or otherwise) are separate from this fix.

A forward looking consideration might be for programs to maintain Opentoonz compatibility by  generating both the unique file and the original form recognized by Opentoonz.  This duplicate intially might be trivial as the files are currently the same except for the filename.  Should the files diverge in content in the future, maintaining both files could supply program specific data not recognized by other programs and be used to assist with targeted conversion.

Note:  There will always be outliers concerning compatibility needing to be addressed and once identified those can also be addressed.  Recommend reviewing those incompatiblities at their source before changing any code.  If unable to resolve the problem at the source other remedies can be explored. 

(Only change from from original PR is squashed commits)

### DIFF
--- a/toonz/sources/toonzlib/tproject.cpp
+++ b/toonz/sources/toonzlib/tproject.cpp
@@ -38,6 +38,7 @@ using namespace std;
 const std::wstring prjSuffix[4] = {L"_otprj", L"_prj63ml", L"_prj6", L"_prj"};
 const std::wstring xmlExt       = L".xml";
 const int prjSuffixCount        = 4;
+const std::wstring tahomaProjectFileName = L"tahomaproject.xml";
 
 //===================================================================
 /*! Default inputs folder: is used to save all scanned immage.*/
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@manongjohn I'd be honored if you reviewed and merged this PR.
Will make adjustments as necessary.

RodneyBaker:
Pressing forward with this fix due to its importance.
Any and all suggestions for improvement will be appreciated.

PR #7073, opened 2026-08-16
(WIP) Allow Command Bar auto-orientation

RodneyBaker:
Adds horizontal/vertical Command Bar orientation with automatic switching based on the docking target. Orientation persists with room layouts, and newly created floating Command Bars open at a useful half screen length instead of inheriting the previous popup size.

The Xsheet Toolbar is unchanged.

Tested in OpenAnimationLibrary/opentoonz-dev#84.
ChatGPT 5.6 was used to research and extensively refine this addition.

### DIFF
--- a/toonz/sources/toonz/commandbar.cpp
+++ b/toonz/sources/toonz/commandbar.cpp
@@ -1,4 +1,4 @@
-﻿
+
 
 #include "commandbar.h"
 
@@ -7,6 +7,7 @@
 #include "menubarcommandids.h"
 #include "tsystem.h"
 #include "commandbarpopup.h"
+#include "pane.h"
 
 // TnzQt includes
 #include "toonzqt/menubarcommand.h"
@@ -23,6 +24,26 @@
 #include <QtDebug>
 #include <QMenuBar>
 #include <QContextMenuEvent>
+#include <QActionGroup>
+#include <QLayout>
+#include <QGuiApplication>
+#include <QScreen>
+#include <QShowEvent>
+#include <QResizeEvent>
+#include <QStyleOptionToolBar>
+#include <QTimer>
+#include <QToolButton>
+
+namespace {
+constexpr int kCommandBarFloatingThickness       = 36;
+constexpr int kCommandBarDockedThickness         = 26;
+constexpr int kCommandBarFloatingFrameDelta      = 10;
+constexpr int kCommandBarTitleBarThickness       = 18;
+constexpr int kCommandBarHorizontalGripWidth     = 20;
+constexpr int kCommandBarVerticalExtensionHeight = 16;
+constexpr int kCommandBarCompactButtonSize       = 20;
+constexpr int kCommandBarCompactGripSize         = 4;
+}
 
 //=============================================================================
 // Toolbar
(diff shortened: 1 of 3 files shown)


Ahl76:
Very appreciable and works well! I noticed a minor issue that doesn't prevent the command bar from docking successfully, but when it's in a floating state and switches from its original orientation to another (during the auto-switch), the bar shrinks and becomes very small.

RodneyBaker:
> Very appreciable and works well! I noticed a minor issue that doesn't prevent the command bar from docking successfully, but when it's in a floating state and switches from its original orientation to another (during the auto-switch), the bar shrinks and becomes very small.

Thanks for testing @Ahl76 
There are some hardcoded numbers in there that are surely responsible for what you are observing.  Those numbers dictate the size of the icons.
I'm also concerned with how these look on Hi Def monitors

There's also a magic number assigned to Max Size that needs to be refactored out.

I'll look into this.

Edit:  Ah, in relooking it seems you aren't referring to the icon size but the length/height of the bar itself.  I thought I had addressed that but apparently not.

RodneyBaker:
@Ahl76 I believe the latest update should address the problem.
Please let me know if you see anything that still needs attention.

Ahl76:
Tested and everything works fine on my end now, nice !

RodneyBaker:
Thanks @Ahl76 
In hindsight this probably should be globally accessible for all window/panels to call instead of scoped just to the command bar.  As such we can call this a limited scope proof of concept.  Your Level Strip update could benefit from auto-orientation as could the Xsheet/Timeline reorientation swap.  But...

Because we'd have to account for some special cases beyond simple icons with more complex windows it seems proper to keep this auto-orientation limited to the Command Bar. At the point where we want to add the capability to other windows I'd say we whould investigate elevating the scope of auto-orientation as well as take a deeper look at how all windows/panels operate in Opentoonz.   

II haven't kept up with all the features and options available to Opentoonz windows but want to document that this feature is currently not available for other windows to call.

RodneyBaker:
Titling this as WIP as there is still some more work to do.
There is a case where a command bar can be very small and we need to allow for that case to only display the arrows that expand the command bar menu rather than display only part of the first icon on the command bar. 

As such I'm testing the introduction of a 'Compact' toggle that automatically resizes to that one icon size with the arrow. In this way custom menus can be place at locations with minimal space but still easily pop out custom command bar menus.

Note:  Because each command bar can contain its own custom shortcuts/icons multiple command bars can be created each with its own set of custom commands.  It not yet clear how we might distiguish each command bar from the others short of always displaying the first icon.  Currently the compact view displays the arrow.

RodneyBaker:
Github outage yesterday likely led to failure of Appveyor reporting.
Closing and reopening to restart that process.

doangelshavewings:
@Ahl76 whats ur discord?

Ahl76:
@RodneyBaker 

I really like the idea, it will definitely help optimize space on small screens.

However, I encountered a bug that froze the UI when I tested compact mode to dock it into one of the slots (green arrow on the screenshot). 

image

The icon somehow ended up on the opposite side, and the UI broke and froze making it impossible to undock any panel.

image

I managed to adjust the layout using the pane splitters (undocking was still broken, even though "Lock Room Panes" was disabled). At that point, another instance of the compact command bar appeared in the top-left corner of the black area (visible in the previous image). Moving it created visual artifacts as it dragged across existing icons.

image

Additionally, moving the Xsheet also moves the mini command bar, and vice versa. Given their respective positions, they shouldn't affect each other. i guess, because of this behavior, the Xsheet panel can end up completely off-screen.

image

Would it also be possible to have the command bar retain its state based on orientation? Currently, it always stays in compact mode. In my opinion, it would be beneficial for the command bar to remember its size/state so it can preserve it when switching between vertical, horizontal, and compact orientations.

@doangelshavewings 

What do you mean ?  I don't have my own Discord server for OpenToonz. I must still be a member of a few, but I'm neither present nor active on them.

RodneyBaker:
Thanks for checking this PR out @Ahl76 

> Would it also be possible to have the command bar retain its state based on orientation? Currently, it always stays in compact mode. In my opinion, it would be beneficial for the command bar to remember its size/state so it can preserve it when switching between vertical, horizontal, and compact orientations.

It certainly should be possible and that's what I hoped to get locked into place; horizontal when in a horizontal space and vertical when in vertical.  As you can see it hasn't quite got there.
Compact mode was added specifically to address oddities when the command bar was docked into a really small space... and particularly for when it was oriented wrong for that tiny space.  I almost immediately regretted adding it because of the new problems it revealed.

I'm adding (WIP) to the title and tagging as v2.0 as the need for rotated command bar isn't high priority.  That will give us additional time not to rush into implementing something less than ideal.

Because @JParks123 stated his desire for vertical command bar I may create a new PR just to add the core swap in orientation without the auto docking orientation etc.

Thanks for the feedback.

As for @doangelshavewings's ping, he's mostly just wanting to get together with you at some time and talk about development with AI.

PR #7075, opened 2026-08-20
Add Animate Tool arrow key nudging

RodneyBaker:
## Summary
- Move the selected Animate Tool object one logical pixel with each arrow key
- Use Shift + arrow for a ten-pixel nudge
- Reuse existing position locks, keyframe creation, undo handling, and object refresh behavior

## Credits
Originally implemented by @doangelshavewings in closed PR https://github.com/opentoonz/opentoonz/pull/7034 (Discussion #6550).

This re-opens the focused change as a standalone PR for review.

## Testing
- Compiled `edittool.cpp` (as noted in the original PR)
- Change is small and reuses the existing `DragPositionTool` path

### DIFF
--- a/toonz/sources/tnztools/edittool.cpp
+++ b/toonz/sources/tnztools/edittool.cpp
@@ -1000,6 +1000,48 @@ void EditTool::leftButtonUp(const TPointD &pos, const TMouseEvent &e) {
 }
 
 //-----------------------------------------------------------------------------
+
+bool EditTool::keyDown(QKeyEvent *event) {
+  if (!event || !doesApply() || m_dragTool || getSpline() ||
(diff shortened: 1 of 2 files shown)


RodneyBaker:
Note that this PR does not fully implement all aspects of  https://github.com/opentoonz/opentoonz/discussions/6550.

If the focus is still in the xsheet with the Animate Tool on the user will move to the adjacent column as normal.  Only when the canvas is the focus will the arrow keys nudge the position of the canvas level on canvas.

This allow the canvas to take focus whereas it did not allow that before.
Before the user would be moved to adjacent columns regardless because the xsheet/timeline always had focus.

RodneyBaker:
Posting intent to merge once all checks pass.

PR #7076, opened 2026-08-20
Add Segment Eraser and synchronized Savebox controls

doangelshavewings:
## Summary

- add Segment mode to the Toonz Raster Eraser, including Selective and Frame Range operation
- use the existing Savebox fill preference as the shared persisted setting for Fill and Segment Eraser controls
- preserve exact saveboxes through undo and redo, and skip undo entries for no-op erases
- traverse scene Frame Ranges by Xsheet row while skipping blank, mixed-level, and repeated held drawings
- improve Savebox overlays and keep autoclose caches separate for limited and full-raster filling

## Upstream credit

Ported and adapted from:
- https://github.com/tahoma2d/tahoma2d/pull/518 by @jeremybullock
- https://github.com/tahoma2d/tahoma2d/pull/738 and its follow-up fixes by @manongjohn

The commits include Co-authored-by trailers for both original developers.

## Validation

- compiled every changed translation unit with the generated CMake commands and warnings treated as errors, excluding existing macOS OpenGL deprecation warnings
- regenerated and compiled the Preferences Qt meta-object source
- passed focused clear-mode, no-op, mixed-style junction, border, boundary-equivalence, and recolor-regression tests
- verified git diff --check

The full local target link remains blocked by the bundled SuperLU archive, which has no arm64 slice.

## Upstream submission

This carries OpenAnimationLibrary/opentoonz-dev#67 on a clean branch based on opentoonz:master, with only the two Segment Eraser/savebox commits.

### DIFF
--- a/toonz/sources/include/tools/tooloptions.h
+++ b/toonz/sources/include/tools/tooloptions.h
@@ -593,9 +593,10 @@ protected slots:
 class EraserToolOptionsBox final : public ToolOptionsBox {
   Q_OBJECT
 
-  ToolOptionCheckbox *m_pencilMode, *m_invertMode, *m_multiFrameMode;
+  ToolOptionCheckbox *m_pencilMode, *m_invertMode, *m_multiFrameMode,
+      *m_eraseOnlySavebox;
   ToolOptionCombo *m_toolType, *m_colorMode;
-  QLabel *m_hardnessLabel;
+  QLabel *m_hardnessLabel, *m_colorModeLabel;
   ToolOptionSlider *m_hardnessField;
 
 public:
--- a/toonz/sources/include/toonz/autoclose.h
+++ b/toonz/sources/include/toonz/autoclose.h
@@ -47,6 +47,10 @@ struct AutocloseSettings {
 };
 
 class DVAPI TAutocloser {
+  static std::string segmentCacheKey(const std::string &id) {
+    return Preferences::instance()->getFillOnlySavebox() ? id + "s" : id;
+  }
+
 public:
   typedef std::pair<TPoint, TPoint> Segment;
 
@@ -69,13 +73,13 @@ public:
 
   // Cache management functions
   static bool hasSegmentCache(const std::string &id) {
+    std::string cacheKey = segmentCacheKey(id);
     std::lock_guard<std::mutex> lock(m_mutex);
-    return m_cache.find(id) != m_cache.end();
+    return m_cache.find(cacheKey) != m_cache.end();
   }
 
   static const std::vector<Segment> &getSegmentCache(const std::string &id) {
-    std::string cacheKey =
-        Preferences::instance()->getFillOnlySavebox() ? id + "s" : id;
+    std::string cacheKey = segmentCacheKey(id);
     std::lock_guard<std::mutex> lock(m_mutex);
     auto it = m_cache.find(cacheKey);
     if (it != m_cache.end()) {
@@ -87,8 +91,7 @@ public:
 
   static void setSegmentCache(const std::string &id,
                               const std::vector<Segment> &segments) {
-    std::string cacheKey =
-        Preferences::instance()->getFillOnlySavebox() ? id + "s" : id;
+    std::string cacheKey = segmentCacheKey(id);
     std::lock_guard<std::mutex> lock(m_mutex);
 
     constexpr size_t MAX_CACHE_SIZE = 50;  // maximum number of images in cache
--- a/toonz/sources/include/toonz/fill.h
+++ b/toonz/sources/include/toonz/fill.h
@@ -101,7 +101,7 @@ DVAPI void inkFill(const TRasterCM32P &r, const TPoint &p, int ink,
 
 bool DVAPI inkSegment(const TRasterCM32P &r, const TPoint &p, int ink,
                       float growFactor, bool isSelective,
-                      TTileSaverCM32 *saver = 0);
+                      TTileSaverCM32 *saver = 0, bool clearInk = false);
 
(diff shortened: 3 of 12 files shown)


doangelshavewings:
Tested and works

RodneyBaker:
First look suggests this PR is ready to move ahead and makes the right changes to porting from T2D to Opentoonz.

General Information:
The Fill tool and Segment Eraser don't maintain independent Savebox settings. 
Both use OpenToonz's existing persisted FillOnlySavebox preference.
That is acceptable and arguably preferable for this port.
It might not be ideal in the future so may need to be revisted.

The UI exposes Savebox directly in both tools while retaining one underlying behavioral preference. The added signal keeps Fill and Eraser controls synchronized. 
Modifying the Savebox in Fill Tool will modify it in the Eraser Tool... for that frame.
The Eraser Tool does not yet have the capability to modify the Savebox.
This shortfall should be addressed in future PR.

The user has to switch back and forth to Selection Tool to modify the Savebox size for every frame.  This can be improved upon at later date.  This is true for usage of the Savebox with both Eraser and Fill Tools.

Note to users:  Savebox position is stored on per frame basis.  
There are two options in Preferences that can effect this.
Both are in Preferences > Tools
Toggling the Savebox option off in toolbar will toggle it off in Prefererces and vice versa.
There is no equivalent for the Minimize Savebox after Edit option which is found only in Preferences.

Posting intent to merge although will allow time for others to test and share feedback.

RodneyBaker:
I believe there may be one thing needing to be addressed before merger.
Scene Frame Range appears to incorrectly remove all repeated frame IDs.

```
std::set<TFrameId> seenFrames;

...

TFrameId fid = cell.getFrameId();
if (!seenFrames.insert(fid).second) continue;
```

That skips every later occurrence of a level frame in the selected row range. 
It does not merely skip consecutive held cells.

Consider this exposure sequence:

Row 1: A1    ← first Segment Eraser stroke
Row 2: A2
Row 3: A1    ← second Segment Eraser stroke

The constructed target list becomes:

A1, A2

The second A1 is discarded. Interpolation then assigns:

A1 → progress 0.0, first stroke
A2 → progress 1.0, second stroke

The endpoint stroke that the user drew on the final A1 exposure is therefore applied to A2, not to A1.

This is not a difference of whether a reused drawing should be modified once or twice. The global deduplication changes which drawing receives the endpoint operation.

Tahoma2D’s scene-frame-range work was specifically intended to follow exposed Xsheet order, while its later Shift fix carried the endpoint forward from the current scene row. This port correctly carries the Shift row forward, but the global set changes the exposure-order semantics.

Recommended:
```
TFrameId previousFid;
bool previousWasTarget = false;

for (int row = firstRow; row <= lastRow; ++row) {
  TXshCell cell = xsheet->getCell(row, m_currCell.first);

  if (cell.isEmpty() || cell.getSimpleLevel() != sl.getPointer()) {
    previousWasTarget = false;
    continue;
  }

  TFrameId fid = cell.getFrameId();
  if (previousWasTarget && fid == previousFid) continue;

  TToonzImageP image = (TToonzImageP)cell.getImage(true);
  if (!image) {
    previousWasTarget = false;
    continue;
  }

  targets.push_back({fid, image});
  previousFid       = fid;
  previousWasTarget = true;
}
```
That should address the issue.

-----------------
Other thoughts / concerns
-----------------
Since we are back in the code anyway there is another change that could be refined.
Before this PR, the exported declaration was the six parameter form:

`bool DVAPI inkSegment(..., TTileSaverCM32 *saver = 0);`

The PR changes it to a seven parameter function with a defaulted final Boolean:

```
bool DVAPI inkSegment(..., TTileSaverCM32 *saver = 0,
                      bool clearInk = false);
```

The Actual Concern
The existing source code remains compatible BUT an already compiled consumer of the old DVAPI symbol would still reference the old six parameter C++ symbol. The default argument does not preserve that binary symbol.

As such, a direct ABI preserving arrangement might be:

```
bool DVAPI inkSegment(const TRasterCM32P &r, const TPoint &p, int ink,
                      float growFactor, bool isSelective,
                      TTileSaverCM32 *saver = 0);


bool DVAPI inkSegment(const TRasterCM32P &r, const TPoint &p, int ink,
                      float growFactor, bool isSelective,
                      TTileSaverCM32 *saver, bool clearInk);
```

With that change the original overload forwards clearInk == false.

This secondary concern should be addressed but arguably is not essential to this PR.

RodneyBaker:
Posting intent to merge with final tests still pending.
ChaptGPT generated merge assessment is attached.

[PR_7076_Merge_Assessment.pdf](https://github.com/user-attachments/files/31323718/PR_7076_Merge_Assessment.pdf)
(Standard disclaimer about ai usage applies)

Future work (from merge assessment document):

> Remaining non-blocking notes
> - The existing six-argument exported inkSegment() symbol could be retained as an ABI-hardening overload. This is optional and need not delay merge.
> - Size, Hardness, and Pencil controls remain visible in Segment mode although Segment does not use them. UI cleanup can follow later.
> - Direct Savebox resizing from Fill/Eraser remains future usability work.
> - Translation updates can follow normal localization work.

RodneyBaker:
Adminstrative note:  Will plan to merge normally which will maintain commit co-authors.
In the future we may prefer to squash commits to reduce noise and manually verify co-authors in the merge description.  Feedback on best practice is encouraged.

RodneyBaker:
Thank you @doangelshavewings for the port and refinement to work done by @motionmonster and @manongjohn.
Thanks to @JParks123 for continually moving Opentoonz forward.
LGTM

Furture:  Monitor Undo operations for hardening opportunity.

Issue #7070, opened 2026-08-15
Crash on save on recently updated system


unlessgames:
### What's the bug?

After updating the system (Linux with opentoonz installed from https://archlinux.org/packages/extra/x86_64/opentoonz/) OpenToonz can no longer save and crashes consistently.

### Steps to Reproduce

1. Create new project and a new scene
2. Draw something on first frame
3. Try save
4. Popup appears with "Failed to save the following resources: A.tlv"
5. Press Ok
6. Try to save again
7. Program crashes with crash report popup showing

### What did you expect to happen?

Saving worked before system update and/or previous opentoonz version.

### Screenshots, Videos or Crash Logs

```
OpenToonz 1.8  (Build Jul  3 2026)

Report Date: 2026-08-15 13:30:46
Crash Reason: (SIGABRT) Usually caused by an abort() or assert()

Build ABI: x86_64-little_endian-lp64
Operating System: EndeavourOS
OS Kernel: 7.1.3-arch1-2
CPU Threads: 8

==== Backtrace ====
  1> /usr/lib/libc.so.6(+0x9981c) [0x7f7cbb29981c]
  2> /usr/lib/libc.so.6(gsignal+0x20) [0x7f7cbb23e150]
  3> /usr/lib/libc.so.6(abort+0x26) [0x7f7cbb22567d]
  4> /usr/lib/libc.so.6(+0x266fc) [0x7f7cbb2266fc]
  5> /usr/lib/libc.so.6(__fprintf_chk+0) [0x7f7cbb3278d0]
  6> /usr/lib/libc.so.6(+0x127254) [0x7f7cbb327254]
  7> /usr/lib/libc.so.6(__fread_chk+0x185) [0x7f7cbb327b15]
  8> /bin/../lib/opentoonz/libimage.so(_ZNK15TImageReaderTzl14getImageInfo10Ev+0x6d0) [0x7f7cc1d749c0]
  9> /bin/../lib/opentoonz/libtnzcore.so(_ZN12TLevelReader12getImageInfoE8TFrameId+0x1ad) [0x7f7cbeb76dcd]
 10> /bin/../lib/opentoonz/libtoonzlib.so(_ZN15TXshSimpleLevel15saveSimpleLevelERK9TFilePathb+0xa7f) [0x7f7cbf4f41af]
[...]

PR #7078, opened 2026-08-20
fix(tzl): reject invalid frame buffer sizes in legacy .tlv loaders

Chantology:
> This change was created by Claude Sonnet 5 and requires proper code review by a maintainer before merging.

**Problem**
- Old `.tlv` level files (format versions 10, 11, 13) store a "how many bytes is this frame" number in the file itself.
- The loader trusts that number completely and copies that many bytes into a fixed-size buffer, with no real check — only a debug assert that does nothing in a normal build.
- If that number is wrong (corrupt file, truncated/failed save, etc.), the copy overflows the buffer and crashes the app.
- This is what's happening in #7070: a failed save leaves a broken `.tlv` behind, and the next save attempt reads that broken file and crashes.
- The current file format (version 14) was already fixed for this exact issue at some point — the older formats were just never updated to match.

**Solution**
- Added the same size check the current format already uses to the four places still missing it: `load10()`, `load11()`, `load13()`, `getImageInfo10()`.
- If the file's stated size doesn't make sense, it's now rejected with a normal, catchable error instead of overflowing memory.
- Tested with a deliberately broken `.tlv` file: old code crashes (confirmed segfault), new code fails cleanly with an error message.
- Tested with a normal, correctly-sized file to make sure valid levels still load exactly as before.

### DIFF
--- a/toonz/sources/image/tzl/tiio_tzl.cpp
+++ b/toonz/sources/image/tzl/tiio_tzl.cpp
@@ -1638,14 +1638,16 @@ m_lrp->m_frameIndex = m_frameIndex;*/
     reverse((char *)&ydpi, sizeof(double));
 #endif
 
+    delete[] imgBuff;
+    imgBuff = 0;
+
     imgBuffSize = m_lx * m_ly * sizeof(TPixelCM32);
     assert(actualBuffSize <= imgBuffSize);
+    if (actualBuffSize <= 0 || actualBuffSize > imgBuffSize)
+      throw TException("Loading tlv: buffer size error");
 
-    delete[] imgBuff;
     imgBuff = new UCHAR[imgBuffSize];
-    // int ret =
     fread(imgBuff, actualBuffSize, 1, chan);
-    // assert(ret==1);
   }
 
   Header *header = (Header *)imgBuff;
@@ -1762,11 +1764,11 @@ m_lrp->m_frameIndex = m_frameIndex;*/
 
   imgBuffSize = m_lx * m_ly * sizeof(TPixelCM32);
   assert(actualBuffSize <= imgBuffSize);
+  if (actualBuffSize <= 0 || actualBuffSize > imgBuffSize)
+    throw TException("Loading tlv: buffer size error");
 
   imgBuff = new UCHAR[imgBuffSize];
-  // int ret =
   fread(imgBuff, actualBuffSize, 1, chan);
-  // assert(ret==1);
 
   Header *header = (Header *)imgBuff;
 
@@ -1796,8 +1798,6 @@ m_lrp->m_frameIndex = m_frameIndex;*/
     }
     ras->unlock();
   }
-
-// codec.compress(ras, 1, &imgBuff, actualBuffSize);
 #endif
 
   TRect savebox(TPoint(sbx0, sby0), TDimension(sblx, sbly));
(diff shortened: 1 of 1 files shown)


RodneyBaker:
@Chantology I'm not quite sure this PR's description is meant to be human readable.
We may need to establish a first pass protocol to deal with that and aide in setting up the eventual human aided review.

Chantology:
@RodneyBaker I updated it. Actually this was my human made fault. I included the wrong one. 

But there is a way to get proper descriptions for real humans directly by creating specifications. I need to try it soon. Wanted to have my PRs done at first and this new Linux issue triggered me

RodneyBaker:
To intentaionally misquote you:
>  is there a way to get proper descriptions for real humans directly by creating specifications?

There certainly must be and it will be appropriately ironic if we arrive at that specification through use of ai tools.

We don't yet have that specification as this whole vibe coding things is quite new.
Do we charge forward and chart our own specification or consult best practices?
I vote for both.

Perhaps more importantly, I'll try to look at the content of this PR soon.  :)

RodneyBaker:
Posting intent to merge.
The specific tests mentioned in PR description are appreciated.

Future consideration
Identify additional tests users of legacy processes can run for validation of backward compatibility.

Chantology:
Two notes before this merges, both from re-reading the diff rather than new work.

**The diff also removes dead code.** Alongside the size checks it deletes a number of commented-out leftovers next to the lines being touched (`// int ret =`, `// assert(ret==1);`, `// codec.compress(…)`, an old `// TToonzImageP ti;` block). That's about 20 of the 26 deleted lines, and my description didn't mention it. Nothing functional, but you should know the diff is not purely the guards. Happy to split it out if you'd rather review the fix on its own.

**Two clarifications to the description:**

* It says the old loaders were guarded only by a debug assert. True for `load10()`, `load11()` and `getImageInfo10()`, but `load13()` had *no* guard at all on its icon path and a commented-out assert on its main path — so that one was unprotected in debug builds too.
* It says versions 10, 11 and 13. The dispatch in `load()` ends with `default: image = load10();`, so a file with an unrecognised or garbage version field also lands in `load10()`. The fix reaches a bit further than the description claims.

**Separately** while checking this I found the same bug class in the *current* format, which this PR does not touch. `load14()` validates the icon buffer against `iconLx * iconLx * sizeof(TPixelCM32)`, but reads into `TRasterCM32P(iconLx, iconLy)`, allocated as exactly `lx * ly * pixelSize`. Icon dimensions preserve the level's aspect ratio, so for any landscape level the guard admits more than the buffer holds — an 80x45 icon accepts 25600 bytes into 14400. It's a one-word fix (`iconLx` → `iconLy`); I'll open it as its own PR rather than widening this one.

RodneyBaker:
No need to split this PR for the dead code removal.  
I believe its fine to remove the specific dead code you've identified and as extra documentation I'll attach a PDF that describes that removal (with a note about some of the 'dead code' actually being moved to another location rather than be removed.

Dead code is a subject of interest from several angles to include the historical record.
That is largely captured in git history itself but it helps to identify it more visibly as well in our descriptions and comments.  Even the experimental work of the past such as some removed via this PR that can provide useful insights.  For instance, for a moment at least a developer was briefly exploring how to best work with DPI.

[OpenToonz_PR_7078_Dead_Code_Removal.pdf](https://github.com/user-attachments/files/31353669/OpenToonz_PR_7078_Dead_Code_Removal.pdf)

Posting intent to merge after a few more tests.

RodneyBaker:
(and for your continuing patience)

PR #7080, opened 2026-08-20
Add Darker theme

doangelshavewings:
## Summary

- add the Darker theme's generated QSS and four theme-specific timeline icons
- provide a variable-driven LESS theme that inherits OpenToonz's shared layouts
- use semantic palette aliases and derived values instead of a second selector/QSS layer

## Architecture

`darker-theme.less` is 237 lines with 160 meaningful variable assignments and zero theme-level selector blocks. The generated QSS now follows the same 3,378-line shared selector topology as the existing themes.

Where the historical QSS used component-specific differences that the shared variable model cannot express, this revision keeps the shared OpenToonz behavior instead of adding hard-coded selector overrides.

## Provenance

The original Darker design and QSS are by @JParks123 (Soda-Popz JP). The conversion was AI-assisted. Following maintainer feedback, the initial direct-QSS compatibility layer was removed and the source was reworked and validated as a semantic-variable theme.

## Verification

- compiled Darker twice with `lessc 4.5.1` and confirmed deterministic output
- confirmed the checked-in QSS matches the LESS output, apart from trimming two inherited comment-line spaces
- recompiled Default, Default-Green, Blue, Dark, Clay, Neutral, and Light byte-identically
- verified all 160 assignments are meaningful, with no unused or removable inherited overrides
- resolved all 56 generated asset URLs and XML-parsed all four bundled SVG icons
- visually smoke-tested the theme in OpenToonz 1.8 on Windows
- ran syntax, secret/PII, and `git diff --check` scans successfully


JParks123:
Thanks for adding this @doangelshavewings 

Screenshot (677)

RodneyBaker:
I am inclined to merge this just so @JParks123 will stop spamming PRs with code that isn't LESS.  @konero would eventually forgive me.

@doangelshavewings I must assume you don't have experience with LESS and are using files @JParks123 supplied to have AI create the LESS file.  

Regarding the content of this PR:
There are comments in the LESS file itself that reveal the likely path of how the LESS file was created and an effort was made to use variables which was a strong start.  However, once the variables were created the variables were never used.  Perhaps more strangely, the darker-theme.less inherits almsot 200 variables from the default-theme.less file, overrides them but then bypasses those variables and never uses them.

```
default-theme.less
        │
        │ imports inherited structure
        ▼
darker-theme.less
   ├── 199 variable overrides        ← good LESS architecture
   │
   └── ~974 lines direct QSS         ← largely bypasses that architecture
```

Because the darker-theme.less frequently repeats colors represented by the default-theme.less, the theme effectively has two sources of truth.

There are 510 literal color occurrences in the included LESS file representing 128 distinct values. Some of the supposedly thematic colors are repeated extraordinarily often.

So the problem isn't that the darker-theme.less file doesn't use mixins etc. but that it establishes a (mostly proper) semantic variable system that can use it and then bypasses that system throughout most of its own code.

Let's explore further.

In LESS an instance of:  background-color: #6378a6;
Would be expected to reference something like:  @hl-color: #6378a6;
Perhaps like this:  background-color: @hl-color
The hightlight color @hl-color is assigned a hexadecimal color value and then various background-color instances reference the variable.

There are 67 unnecessary copies of #333333, 55 copies of #242424, 53 copies of #424c61, and 51 copies of #6378a6.

If we examine one of those values we can see they are assigned to variables:
@hl-color: #6378a6;
@normal-label-color: #6378a6;
@rooms-tab-bg-color-hover: #6378a6;
@button-bg-color-hover: #6378a6;
@checkbox-bg-color-checked-hover: #6378a6;
@input-selection-bg-color: #6378a6;

The problem isn't merely that the file could introduce variables for these repeated colors. The variables already exist but aren't used.

In short, this LESS file uses essentially none of the LESS capability.

Please don't feel the need to close this PR.
It may serve as a documentation example for learning purposes of LESS and how use of ai doesn't always work.

If this was just a first pass at creating the LESS then... press on.  Press on.

konero:
As @RodneyBaker explains, much of the stuff is unused, and still bakes QSS directly into the LESS file. The LESS file would be about 100 lines tops if you'd just use the references from Default and override them. Look at the other themes for an example.

doangelshavewings:
Thanks for the detailed review, @RodneyBaker. You were right about the two sources of truth in the first pass.

The 199 variables did affect the imported layouts through LESS lazy evaluation, but the appended selector layer then bypassed that architecture. I should not have described that hybrid as a proper LESS port.

I reworked the theme locally and pushed one cohesive update:

- LESS: 1,251 -> 237 lines
- inherited/no-op overrides removed: 46
- meaningful variable assignments: 160
- theme-level selector blocks: 188 -> 0
- direct QSS declarations: 337 -> 0
- literal colors: 510 occurrences / 128 values -> 68 / 58
- generated QSS: 4,154 -> 3,378 lines

The revised theme now inherits the shared selector structure and uses semantic aliases/derived values. Where the shared model cannot express an exact historical component difference, I kept the shared behavior rather than restoring selector overrides.

I also recompiled all seven existing themes byte-identically, verified all assets, and smoke-tested the result in OpenToonz. The original design/QSS is by @JParks123, and the conversion was AI-assisted. Thank you for pointing the work toward the correct architecture.

RodneyBaker:
@doangelshavewings  I was writing this as you posted your update. Disregard as necessary.

-----------------------

In the above I overstate that the variables are never used.
Many of the inherited variables from default-theme.less are in fact used.
It's how variables are or are not used that is under consideration.

I'm attaching an attempt to move the provided file more into LESS territory than hard coded QSS. I've intentionally used darker theme variable names and added some explanatory comments.  I am not suggesting this LESS produces the desired QSS theme or comforms to ideal LESS standards.  For best results it helps to start with LESS.  @JParks123 can determine if the LESS creates the theme he wants.

As LESS files cannot be uploaded I've attached it as as a .zip.
[darker-theme-refactored.zip](https://github.com/user-attachments/files/31270736/darker-theme-refactored.zip)

RodneyBaker:
@konero I have not yet looked at the latest update deeply (I just skimned through) and am inclined to merge this PR even should there still be room for improvement.  Not to put you on the spot but this is your opporunity to shine.  Please consider making suggestions where necessary to refine -the current LESS presented in this PR- so it can be the absolute best it can be when merged.  Identifying any critical reasons not to merge will be particularly appreciated.

@doangelshavewings Thanks for the update.  I'll give the update a closer look and see if anything stands out.  From what I've seen thus far the LESS is much improved.

JParks123:
there are a few intentional errors in the color choices done on purpose for stylistic reasons @RodneyBaker

RodneyBaker:
One minor thing stands in the way of merger.

The code suggests buttons using:  @button-text-color-pressed: @dark-color-2;

The inherited Default-theme definition makes the pressed background:

@button-bg-color-pressed: @dark-color-2;

Both consequently compile to #424c61. The generated QSS confirms the collision in several shared button-family rules:

QPushButton:pressed {
  background-color: #424c61;
  color: #424c61;
}


#CustomPanelButton:pressed {
  background-color: #424c61;
  color: #424c61;
}


QComboBox:pressed {
  background-color: #424c61;
  color: #424c61;
}

The question to the author of the theme (@JParks123).
Is this disappearance of text when button's pressed intentional?

If intentional then this PR can be merged.
If not then we need to make the adjustment.

Edit:  Asked and answered.  Please see @JParks123 response above.
Users that don't like that choice can copy/paste the above into Additional Style Sheet and change the color.

RodneyBaker:
Thank you @doangelshavewings for helping get @JParks123's Darker theme across the finish line.  LGTM

PR #6991, opened 2026-07-31
Right-click Wheel Shape on the color wheel

Ahl76:
**Proposal**

Add a right-click menu on the color wheel to choose Wheel Shape (Classic / Round / Hexagonal).

This completes the split of the earlier color wheel work (#6981 / #6989), by keeping wheel shape selection available directly on the wheel.

Depends on #6989

**Notes**

This feature was developed with Cursor’s AI models.

PR #7081, opened 2026-08-21
Style editor advanced color selector

Ahl76:
### Proposal

This PR presents a new approach that replaces my previous proposals regarding the Style Editor wheel (#6981, #6989, #6990 and #6991). I am closing those old PRs in favor of this one. It builds upon the bug fixes, logic, and user feedback from those previous drafts.

This PR enriches the Color tab in the Style Editor and provides artists with quick assistance when picking colors. It introduces an optional Advanced mode as well as several color assistance tools inspired by workflows from Clip Studio Paint, Cacani, and previous Advanced Color Selector concepts. OpenToonz's Classic wheel remains the default behavior. All new additions remain entirely optional as well.

image

I used the same location as the sliders to place the features of the Color Features Bar. It comes as a compact toolbar displaying icons only, with the option to show the active tab name via the context menu. Activating an icon reveals a dedicated color space. Deactivating the active icon allows you to return to the classic slider space.

All actions performed with these new tools apply to the active style. They do not add any unnecessary colors to the scene palette, unless the artist explicitly chooses to collect or save a color.

### FEATURES

**Classic / Advanced Modes :** Toggleable via a dedicated icon on the picker or from the right-click menu. Classic mode remains unchanged. Advanced mode introduces a wheel with square or triangle SV selectors, as well as a Rectangle mode.

**Picker Components :** Display options for CP (color picker), Alpha, HSV, RGB, HEX, Wheel/Rectangle, and VAR (colors variations). Each element can be hidden or shown from the right-click menu. 

**Picker markers :** The current-color indicator on the wheel and chromatic spaces (Classic triangle, Advanced wheel, square, and Rectangle) is drawn larger than before, to make it easier to spot while picking.

**VAR :** value variations of the current color. A click applies the chip at once. The strip sits above the picker, with the CP / HSV / RGB controls.

**Color Collector :** A custom grid for storing color swatches. Saved between sessions (user settings). Ctrl+click on several feature grids can send a color to the Collector.

Color History: Displays recently used colors in the Color tab for quick reapplication. Saved between sessions (user settings).

**Color Harmonies :** Harmony types (complementary, analogous, tetrad) with chips in the features panel. On the wheel, visual guides appear in Classic and in Advanced Wheel mode. In Rectangle mode, guides also appear on the vertical hue slider.

**Color Shades :** Generates automated rows for values, temperature, saturation, The last row is a customizable ramp: set the left and right colors, and the cells in between interpolate between them.

Neighboring Colors : Offers a 2D matrix centered on the current color, configurable along H/S/V or R/G/B axes with an adjustable range.

**Intermediate Colors :** A four-corner blending grid allowing smooth gradients to be generated between multiple points.
The four corners of Intermediate Colors are editable: Alt-click sets a corner from the current color, Alt-right-click clears it. Right-click can reset one corner or all four.

Neighboring Colors and Intermediate Colors have Compact grid (default) and Large grid, from the right-click menu.

**Color Mixer :** A compact digital mixing space handling color mixing modes (RYB, RGB, Soft, and Finger), brush sizes, optional mix-with-background, and undo/redo. Note that this is not optimal yet.

Interface Customization : Right-click on the picker controls Advanced options, the Color Features Bar, and which small icons appear on the picker (Classic/Advanced, Wheel/Rectangle, VAR, section toggles). Which slider sections are shown (CP, HSV, Alpha, RGB, Hex) is toggled from those picker icons, or as before from the menu button at the bottom of the Style Editor panel.

**Question 1** : Picker Orientation

Currently, Advanced Wheel and Rectangle modes use the Classic mode layout, with hue on the left and white on the right.

Some users might prefer a different orientation or layout preferences, so feedback is welcome on whether OpenToonz should maintain consistency with Classic mode or offer a reversed orientation for Advanced mode.

**Question 2** : Classic mode naming in the code

The wheel name in Classic mode was changed to Color picker since the introduction of another wheel (which takes over the name) and rectangle.The former Wheel section is now Color Picker, so the name does not clash with Advanced Wheel / Rectangle. Feedback is welcome on keeping Color Picker, or restoring Wheel for that section if Advanced Wheel is enough of a distinction.

### Limitations

In the Color Mixer, it does not yet work totally as desired regarding rendering and pressure support. The behavior is fully functional as-is for testing purposes, and can always be refined over time.

### Notes

This update incorporates the geometry and HiDPI display fixes developed in previous wheel PRs.

@rodneybaker @jparks23, could you please check if this better aligns with your requests?

Added : 

**Common**

A color from the Color Features can be kept without applying it : sent to a color set, or added as a new style via right-click

**Interface**

On first launch, the extra Color page controls and the Color Features Bar stay hidden.

**Color Collector**

Ten named color sets persist between sessions.

**Color Harmonies**

Harmonies can hold their colors while Apply still updates the current style.

**Color Mixer**

The mixer canvas is kept between sessions.

### DIFF
--- a/toonz/sources/include/toonzqt/styleeditor.h
+++ b/toonz/sources/include/toonzqt/styleeditor.h
@@ -531,6 +531,8 @@ class PlainColorPage final : public StyleEditorPage {
   QWidget *m_historyGrid       = nullptr;
   QToolButton *m_harmonyBtn    = nullptr;
   QWidget *m_harmonyPane       = nullptr;
+  QToolButton *m_shadesBtn     = nullptr;
+  QWidget *m_shadesPane        = nullptr;
   QToolButton *m_wheelKindBtn;
   QToolButton *m_rectKindBtn;
   QToolButton *m_advancedModeBtn;
@@ -545,6 +547,7 @@ class PlainColorPage final : public StyleEditorPage {
   QWidget *m_variationStrip       = nullptr;
   int m_visibleParts;
   void updateControls();
+  void syncFeaturePage();
 
   // QGridLayout *m_mainLayout;
   QFrame *m_slidersContainer;
--- a/toonz/sources/include/toonzqt/styleeditor.h
+++ b/toonz/sources/include/toonzqt/styleeditor.h
@@ -601,6 +601,7 @@ signals:
   void svShapeClicked();
   void pickerKindClicked(int kind);
   void pickerContextMenuRequested(const QPoint &globalPos);
+  void addPaletteStyleRequested(const ColorModel &);
 
 protected slots:
   void onWheelChanged(const ColorModel &color, bool isDragging);
@@ -1117,6 +1118,7 @@ protected slots:
 
   // chiamato quando viene modificato uno slider o la color wheel
   void onColorChanged(const ColorModel &, bool isDragging);
+  void onAddPaletteStyleRequested(const ColorModel &);
 
   void selectStyle(const TColorStyle &style);
 
--- a/toonz/sources/include/toonzqt/styleeditor.h
+++ b/toonz/sources/include/toonzqt/styleeditor.h
@@ -38,6 +38,7 @@
 #include <QRectF>
 #include <QSettings>
 #include <QSplitter>
+#include <QStackedWidget>
 #include <QRadioButton>
 #include <QButtonGroup>
 
@@ -182,6 +183,7 @@ private:
   void drawHueRing();
   void drawSatValueTriangle();
   void drawCurrentColorMark();
+  void drawHarmonyMarks();
   void drawHueRingBaton(int hue, float innerDist, float outerDist,
                         const QPointF &center);
   void computeAdvancedSVTriangle();
@@ -521,6 +523,14 @@ class PlainColorPage final : public StyleEditorPage {
   QWidget *m_pickerChrome;
   QWidget *m_sectionBar;
   QToolButton *m_svShapeBtn;
+  QWidget *m_featureBar        = nullptr;
+  QStackedWidget *m_featureStack = nullptr;
+  QToolButton *m_collectorBtn  = nullptr;
+  QWidget *m_collectorGrid     = nullptr;
+  QToolButton *m_historyBtn    = nullptr;
+  QWidget *m_historyGrid       = nullptr;
+  QToolButton *m_harmonyBtn    = nullptr;
+  QWidget *m_harmonyPane       = nullptr;
   QToolButton *m_wheelKindBtn;
   QToolButton *m_rectKindBtn;
   QToolButton *m_advancedModeBtn;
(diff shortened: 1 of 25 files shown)


JParks123:
I will test for sure!  😊
will check back to see if you get it past the finish line.

JParks123:
so far the only problem I see is there is no function to keep things as they were for the original Style Editor color wheel. does it always have to have the big buttons there and the stuff next to the color sliders? all aspects of this upgrade should be completely optional.  I do like what you are doing though maybe it won't be a problem but it might be that the top buttons and slider option controls need to be hidden in the hamburger menu as an optional thing.

RodneyBaker:
@Ahl76 I have only given the build from this PR a cursory look but wanted to give you my initial impression because I had one primary concern going into this PR:  How disruptive will this change be to users of Opentoonz given their familiarity with previous state of the Style Editor.  I am happy to report that superficially I don't see any reason for concern.

It doesn't read like @JParks123 found the icon to get the current hex color wheel but it's there.

With all the options available it will take awhile to work through and test them all but so far so good.  Nice work.

P.S.  Color History?  Seriously.  I think you've been reading my mind.

Added:  The file count is deceptively high due to 22 SVG icons.  Only three files contain the code change.

RodneyBaker:
@Ahl76 
By the time you read this you might have already made this a moot point.

The diff is hard to review commit by commit for a number of reasons.
The first six commits build a "Wheel Shape: Classic / Round / Hexagonal" design that the seventh commit supersedes with Classic / Advanced. 

Reading the history walks us through the abandoned design.
Only the squashed end state matters.

RodneyBaker:
@Ahl76 
I wouldn't add this now as this PR is already large and complex but by way of feedback I submit the following for (mostly) future work consideration:

Given this new Style Editor's complexity adding Style Editor presets would be useful.  
Most of the code and data for presets is already present.  Presets would serve in making any specific Style Editor configuration optional, customizable and saveable and retrievable.
This would also address @JParks123 interest in having the classic Style Editor view as that preset could be shipped with Opentoonz. 

I'm attaching a PDF explores how implementing Style Editor presets might be achieved.
There are a few things that could be done now to adding presets later much easier.

There are also a few things to consider that harden the current proposal even without exploring presets.

[opentoonz-pr7081-style-editor-presets.pdf](https://github.com/user-attachments/files/31332985/opentoonz-pr7081-style-editor-presets.pdf)

I acknowedge the color mixer being Work In Progress at this stage.
That's a very interesting addition.

Bottom line:  This updated Style Editor will be a great addition to Opentoonz but the complexity and variety of options may be overwhelming for users.  The good news:  There will be plenty of opportunity for people to share video walkthroughs of their favorite settings and workflow.

Ahl76:
@JParks123
Everything is optional via right-click and the different toggle options. You're right though, I thought I had set the classic Style Editor (with all new features hidden) as the default behavior, but it wasn't the case. I'll update it so the initial launch always starts with the standard, familiar view. Thanks for the feedback!

@RodneyBaker
Thank you for the detailed feedback! I haven't fully looked into how presets should be handled for the Style Editor yet, but I'll read through your PDF and see how we could implement your proposal, following your suggestions.

edited intro after news commits

PR #7082, opened 2026-08-22
Check clang-format on changed PR source files

RodneyBaker:
# Check clang-format only on changed PR source files

This PR adds a clang-format CI check for C/C++ files changed by a pull request.

OpenToonz already defines its formatting policy in `toonz/sources/.clang-format`, based on Google style with project specific override.  Its existing beautification scripts operate on files changed relative to the base branch. This PR keeps that structure intact and brings the CI behavior into line with it.

Workflow:

* runs only when relevant C/C++ source files, `.clang-format`, or the workflow itself changes;
* checks only files changed by the pull request;
* ignores deleted files;
* uses the existing `toonz/sources/.clang-format` through clang-format's normal `-style=file` lookup;
* verifies formatting with `--dry-run --Werror` rather than modifying contributor branches;
* grants only `contents: read` permission.

The intent is to enforce the current formatting standard on code being introduced or modified without requiring unrelated legacy source files to be reformatted or causing pull requests to fail because of formatting outside their scope.

This supersedes the implementation approach explored in #5615 while preserving the underlying concern raised there: linting and formatting checks should remain aligned with the actual scope of a pull request.  There is the question remaining however about how best to lint legacy code.

### DIFF
--- a/.github/workflows/workflow_clang_format.yml
+++ b/.github/workflows/workflow_clang_format.yml
@@ -1,0 +1,63 @@
+name: clang-format Check
+
+on:
+  pull_request:
+    paths:
+      - "toonz/sources/**/*.c"
+      - "toonz/sources/**/*.cc"
+      - "toonz/sources/**/*.cpp"
+      - "toonz/sources/**/*.cxx"
+      - "toonz/sources/**/*.h"
+      - "toonz/sources/**/*.hh"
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Posting intent to merge on trial basis.
Follow up and refinement is anticipated.

PR #7085, opened 2026-08-24
Add Trail stamp cycle modes and improve Raster Trail quality

doangelshavewings:
## Summary

Add Off, Forward, Backward and Repeat Trail Cycle modes to the Vector Brush for vector and raster Trail styles. Each committed gesture selects one source-frame offset; assistant and symmetry replicas share that selection. Strokes retain their offset and step for subsequent rendering and saving.

Forward, Backward and Repeat share a cursor. Repeat holds one source frame throughout the stroke. Canceled gestures and inactive styles do not advance or replace the cursor. Changing the Trail resource, palette, style slot or frame count starts a new cursor on the next committed gesture. The control is enabled for multi-frame Trail styles when Frame Range is off, and the cycle shortcut remains configurable.

Frame Range selections are committed before refreshing the Vector Brush toolbar, so selecting a range mode no longer resets it to Off. The refresh still updates Trail Cycle availability after the selection.

## Raster Trail rendering

- Size and space stamps from artwork bounds instead of transparent source padding. **This intentionally changes existing projects that use padded Raster Trail sources.** No legacy-sizing switch is included.
- Measure original alpha bounds before reduction. All animation frames share the union of those bounds, preserving their relative size and movement, with a transparent gutter.
- Keep source detail up to 2048 pixels per side, subject to a conservative 128 MiB pixel budget per loaded style. Load in two passes while retaining one original decoded frame at a time.
- Cover stamp extents in bounding boxes and tile culling without expanding the box in proportion to the entire stroke length.
- Prepare power-of-two textures for the active context and displayed stamp size, applying the actual color function to the pixels.
- Distinguish opacity, filters and concrete color-function types in cache keys. Retain source identity so reloading cannot reuse stale pixels. Unsupported signatures bypass caching.
- Prepare textures outside the cache mutex and evict least-recently-used entries at 32 entries or 64 MiB, conservatively counting retained source and prepared pixels.

Texture uploads still occur during drawing; persistent GPU caching and mipmaps are not part of this change.

## Persistence and compatibility

Nondefault offset/step values extend the existing PLI outline-options tag. Original payloads remain readable; partial optional extensions use defaults, trailing data is ignored, and truncated original fields are rejected. Values outside the signed-magnitude encoding range are rejected while constructing outline tags. The miter-field width calculation uses the encoded values directly.

Older readers can skip the extension, but older applications do not implement its behavior and lose the cycle fields when saving. A released-application downgrade round trip remains unverified. `TStroke::OutlineOptions` has a changed binary layout, so dependent modules/plugins must be rebuilt against matching headers and libraries.

## Validation

- Full Debug OpenToonz build on macOS arm64 with Qt 5.15.19 and bundled TIFF/SuperLU rebuilt for arm64.
- Five headless CTest suites pass with the updated toolbar: color functions, gesture state, raster preparation/cache, actual PLI reader/writer, and brush options. The offscreen OpenGL rendering suite also passed during rendering validation.
- The brush-options regression sends Qt key events through the actual Vector Brush toolbar. It covers every Range mode, return to Off, and selection after a property refresh with Default, Darker and no stylesheet. It fails against the previous toolbar library and passes with the fix.
- Passing UndefinedBehaviorSanitizer runs for the color, state, texture, PLI and rendering test/helper code and directly compiled PLI/color sources; the linked core library is not instrumented.
- Rendering checks cover PNG, vector and rasterized-vector sources; Forward/Backward/Repeat sequences; wrapped offsets; tiled/full pixel equality; opacity; and tiny artwork on a 16K-wide padded canvas at thumbnail, normal and enlarged scales.
- Cache checks cover source replacement, unsupported functions, entry/byte eviction and concurrent insertion. Raising the cache budget makes the byte-limit regression fail; restoring it makes the test pass.
- PLI checks cover round trips, partial/trailing tags, malformed payloads, format limits and copy/split/transform metadata retention.
- Changed-line clang-format 14 checks and `git diff --check` pass. Linux and macOS CI run all five headless suites, including the toolbar regression; updated CI results are pending.

Still requiring verification: interactive toolbar enablement, real viewer/preview/column-icon parity, undo/redo across scene switches, a released older application's open/save cycle, and Windows/Linux rendering. Desktop automation is unavailable due to a client/server mismatch. AddressSanitizer cannot initialize with the installed Apple runtime. The full application build emits existing legacy/OpenGL deprecation warnings. The toolbar library emits existing path-initialization diagnostics before test setup; regression-test source compiles with warnings treated as errors.

See [Trail Cycle behavior and validation](https://github.com/doangelshavewings/opentoonz/blob/feature/trail-stamp-cycle/doc/trail_cycle_validation.md) for limits and reproducible test commands.

## AI assistance

OpenAI Codex assisted with the Frame Range investigation, fix,
regression testing, and documentation. That follow-up used GPT-6;
the exact model snapshot is unavailable.

Task summary: investigate why Vector Brush Range selections reset
to Off, correct the toolbar update order, and test selection through
the actual toolbar under Default, Darker, and no stylesheet.

Model and prompt details for earlier contributions have not been
verified. AI-assisted checks do not constitute human review.

## Review notes

- Existing-scene rendering: artwork-based Raster Trail sizing changes
  the appearance of existing scenes using padded sources. No legacy
  sizing mode is provided. This requires an explicit compatibility
  decision before merge.
- Scope: Trail Cycle controls and Raster Trail rendering changes are
  combined here. Their separation should be assessed against the
  upstream one-topic-per-PR guidance.
- Release classification: this remains a feature PR despite including
  the Frame Range fix. Merge eligibility depends on the release phase.
- Policy references: OpenToonz Discussions #6332 and #6937. No compatibility exception or maintainer approval is implied.

### DIFF
--- a/stuff/profiles/layouts/shortcuts/defopentoonz.ini
+++ b/stuff/profiles/layouts/shortcuts/defopentoonz.ini
@@ -53,6 +53,7 @@ A_ToolOption_SkeletonMode:Inverse Kinematics=
 A_ToolOption_Smooth=
 A_ToolOption_Snap=
 A_ToolOption_SnapSensitivity=
+A_ToolOption_TrailCycle=
 A_ToolOption_Type:Freehand=
 A_ToolOption_Type:Normal=
 A_ToolOption_Type:Polyline=
--- a/stuff/profiles/layouts/shortcuts/otadobe.ini
+++ b/stuff/profiles/layouts/shortcuts/otadobe.ini
@@ -53,6 +53,7 @@ A_ToolOption_SkeletonMode:Inverse Kinematics=
 A_ToolOption_Smooth=
 A_ToolOption_Snap=
 A_ToolOption_SnapSensitivity=
+A_ToolOption_TrailCycle=
 A_ToolOption_Type:Freehand=
 A_ToolOption_Type:Normal=
 A_ToolOption_Type:Polyline=
--- a/stuff/profiles/layouts/shortcuts/otharmony.ini
+++ b/stuff/profiles/layouts/shortcuts/otharmony.ini
@@ -53,6 +53,7 @@ A_ToolOption_SkeletonMode:Inverse Kinematics=
 A_ToolOption_Smooth=
 A_ToolOption_Snap=
 A_ToolOption_SnapSensitivity=
+A_ToolOption_TrailCycle=
 A_ToolOption_Type:Freehand=
 A_ToolOption_Type:Normal=
 A_ToolOption_Type:Polyline=
--- a/stuff/profiles/layouts/shortcuts/otretas.ini
+++ b/stuff/profiles/layouts/shortcuts/otretas.ini
@@ -53,6 +53,7 @@ A_ToolOption_SkeletonMode:Inverse Kinematics=
 A_ToolOption_Smooth=
 A_ToolOption_Snap=
 A_ToolOption_SnapSensitivity=
+A_ToolOption_TrailCycle=
 A_ToolOption_Type:Freehand=
 A_ToolOption_Type:Normal=
 A_ToolOption_Type:Polyline=
--- a/toonz/sources/common/tvectorimage/tstroke.cpp
+++ b/toonz/sources/common/tvectorimage/tstroke.cpp
@@ -3776,13 +3776,19 @@ TStroke::OutlineOptions::OutlineOptions()
     : m_capStyle(ROUND_CAP)
     , m_joinStyle(ROUND_JOIN)
     , m_miterLower(0.0)
-    , m_miterUpper(4.0) {}
+    , m_miterUpper(4.0)
+    , m_patternFrameOffset(0)
+    , m_patternFrameStep(1) {}
 
 //-----------------------------------------------------------------------------
 
 TStroke::OutlineOptions::OutlineOptions(UCHAR capStyle, UCHAR joinStyle,
-                                        double lower, double upper)
+                                        double lower, double upper,
+                                        int patternFrameOffset,
+                                        int patternFrameStep)
     : m_capStyle(capStyle)
     , m_joinStyle(joinStyle)
     , m_miterLower(lower)
-    , m_miterUpper(upper) {}
+    , m_miterUpper(upper)
+    , m_patternFrameOffset(patternFrameOffset)
+    , m_patternFrameStep(patternFrameStep) {}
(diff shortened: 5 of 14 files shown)


RodneyBaker:
@doangelshavewings Ignore the clang-format errors as they have nothing to do with your PR.
Thanks

doangelshavewings:
Confirmed working.
Use a Trail style whose source .pli has clearly different frames—e.g. draw 1, 2, 3 on three frames—so you can see exactly which frame each stamp uses.
1. Open a Vector level, choose the Vector Brush, and select that Trail style in the Style Editor.
2. In Tool Options, set Trail Cycle: to each mode and make separate click-stamps on the canvas

Mode | Expected 3-frame result
-- | --
Off | 1, 1, 1, 1…
Forward | 1, 2, 3, 1…
Backward | 3, 2, 1, 3…
Draw a longer stroke.- Forward should place source frames in forward order along the stroke.
- Backward should place them in reverse order.
- Changing the mode resets the next stamp to the appropriate start frame.

Enable an assistant/replicator, then make one stamp.- All copies created by that one gesture should show the same source frame.
- The next separate gesture should advance only once.

Verify persistence.- Make several Forward/Backward stamps.
- Save the level/scene, close it, and reopen it.
- The existing stamps must keep their frames and directions.
- Undo/redo must restore the same visuals. (The next brush stamp does not rewind after Undo; that is expected.)

RodneyBaker:
@doangelshavewings 
Attached is a (mostly) automated review of current status for this PR.
I share it here for informational purposes and as general feedback.

Aside from what is contained in the PDF I note the comments in the code are very verbose but this can be expected given that much of the update simply translates from older (Italian) comments.  This is of interest enough to mention as there is sufficient duplication to warrant consideration moving forward particularly where the comments might be merged into the code itself by way of naming conventions etc.  There are some specifically valid comments that allude to how toonz code works in ways that would not otherwise be obvious.  These are interesting and relevant to consider for future work and code improvement outside the scope of this PR.

[OpenToonz_PR_7085_Development_Review.pdf](https://github.com/user-attachments/files/31520413/OpenToonz_PR_7085_Development_Review.pdf)

This cursory review does not consider actual use of the Trail Stamp user's experience and focuses only on the current code.  I do not suggest it states any conclusions except the most important one which is that this PR is not yet ready to be merged.  At this stage in draft state you already know this.

As always, thanks for all you do.

JParks123:
I guess rodney pulled the plug on this project for now.

PR #7086, opened 2026-08-24
Check clang-format only on changed lines

RodneyBaker:


RodneyBaker:
clang-format should only be checking changed lines not all files.
Merging to get clang-format check properly online.

PR #7088, opened 2026-08-26
Load .cube 3D LUT support

RodneyBaker:
Adds 3D `.cube` LUT support to OpenToonz’s existing display color calibration. 

The loader supports common IRIDAS/Resolve domain headers, validates LUT data before activation, preserves existing `.3dl` behavior, and safely refreshes the OpenGL texture. 

LUTs affect viewer calibration only and do not alter rendered output. 
Unsupported 1D and combined shaper LUTs are rejected.

ChatGPT 5.6 used for research, testing and processing.

Future work:
[ ] Add shortcut key capability to toggle LUT on/off
[ ] Add Schematic node (or other appropriate process) to allow burnin of LUT to output images to support non linear color workflows - the average user will expect it
[ ] Add Opentoonz Library LUT directory and have LUT loader use that location to list LUTs
[ ] Supply license friendly LUTs collection
[ ] Enable image to LUT creation

Note:  LUT processing should use proper Color Science to ensure correct and consistent color space processing

### DIFF
--- a/toonz/sources/include/toonzqt/lutcalibrator.h
+++ b/toonz/sources/include/toonzqt/lutcalibrator.h
@@ -80,8 +80,8 @@ class DVAPI LutManager  // singleton
   LutManager();
 
   struct Lut {
-    int meshSize = 0;
-    float* data = nullptr;
+    int meshSize       = 0;
+    float* data        = nullptr;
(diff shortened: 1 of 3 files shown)


RodneyBaker:
This is also a test of github's stacked PRs.

PR #7093, opened 2026-08-27
Improve crash report diagnostics

RodneyBaker:
This PR improves the prior Crash Report in the following ways:

* Report exact OpenToonz version and build date separately.
* Identify Windows 10/11 from native kernel/not potentially outdated  Qt description.
* Report Qt build and runtime versions.
* Preserve Unicode paths and identifying paths containing non-ASCII characters.
* List detected Wintab and Wacom tablet modules.
* Correctly identify the OpenGL/driver information.
* Preserve the Windows exception context in generated minidumps.
* Record loaded module paths through Unicode-safe Windows APIs.
* Recommend builds older than 30 days be retested with current release or nightly build.
* Open unresolved crash listing/favor review and confirmation over duplicate reports.

These changes improve diagnostic accuracy without changing existing OpenToonz workflows, scene files, or rendering.

Developed with ChatGPT 5.6

### DIFF
--- a/toonz/sources/toonz/crashhandler.cpp
+++ b/toonz/sources/toonz/crashhandler.cpp
@@ -606,7 +606,7 @@ void CrashHandler::copyClipboard() {
 //-----------------------------------------------------------------------------
 
 void CrashHandler::openWebpage() {
-  QDesktopServices::openUrl(QUrl("https://github.com/opentoonz/opentoonz/issues"));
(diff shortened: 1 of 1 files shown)


RodneyBaker:
There is much work to do with regard to crash reporting but this update should help to get more accurate information into a crash report.

The ultimate goal is for the crash report 'feature' to be seldomly if ever used.

Future:
A user should never lose data and in the rare case of data loss, recovery should be as swift and painless as possible. 

At the point where crashes are extremely rare, Opentoonz crash reporter can properly shift focus to a role as development tool (as opposed to ad hoc means of user engagement during crisis).

PR #7095, opened 2026-08-28
Add Save as Custom Vector Brush command

RodneyBaker:
Summary
- Adds Save as Custom Vector Brush to the Viewer context menu for selected vector strokes.
- Saves the selection as a one frame PLI in library/vector brushes.
- Holding Shift while opening menu provides Save Vector Selection As for saving elsewhere.
- Existing PLI files can be overwritten or appended. 
- Append preserves existing frames and adds the selection after the highest frame number.
- Validates temporary PLI before replacing destination and refreshes brush library after saving.

A soft warning appears for custom brushes larger than 32 KiB. Vector Brush and Vector Image Pattern styles are rejected to prevent nested brush dependencies.

Validation
clang-format 14 check
git diff --check
Linux, Windows, and macOS CI passed on OpenAnimationLibrary/opentoonz-dev#100
ChatGPT 5.6 was used extensively in research, testing and processing

### DIFF
--- a/toonz/sources/include/tools/strokeselection.h
+++ b/toonz/sources/include/tools/strokeselection.h
@@ -86,6 +86,8 @@ public:
   void copy();
   void cut();
   void paste();
+  void saveAsCustomVectorBrush();
+  void saveVectorSelectionAs();
 
   void removeEndpoints();
   void sortWithPaletteOrder();
@@ -95,6 +97,8 @@ public:
   void selectAll();
 
 private:
+  void saveSelectedVectors(bool chooseDestination);
+
   TVectorImageP m_vi;          //!< Selected vector image.
   IndexesContainer m_indexes;  //!< Selected stroke indexes in m_vi.
 
--- a/toonz/sources/include/toonz/customvectorbrushexporter.h
+++ b/toonz/sources/include/toonz/customvectorbrushexporter.h
@@ -1,0 +1,61 @@
+#pragma once
+
+#ifndef CUSTOM_VECTOR_BRUSH_EXPORTER_H
+#define CUSTOM_VECTOR_BRUSH_EXPORTER_H
+
+#include "tfilepath.h"
+#include "tvectorimage.h"
+
+#include <vector>
+
+#undef DVAPI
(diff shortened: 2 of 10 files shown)


RodneyBaker:
This PR introduces a new process that Appends selected vectors to a PLI file.
If a PLI file exists the user is given the choice to overwrite the PLI file or to append to the file.
In the case of creating brushes replacement will tend to be preferred (for now).

A constraint exists where there is no pointer to a poster frame of a sequence.
As such, the first frame of a sequence is used for a variety of processes to include icon preview assignment.  This leads to a case where a PLI file serving as a brush only recognizes the first frame of the PLI.  The other frames are never used or referred to.  

To select a different 'brush' from a set of PLI frames (without a dedicated pointer) the current process would be to select the first frame and then the desired frame to be used as a brush and 'Reverse' the order.  Once saved the selected frame is now the first frame and is recognized as that 'poster frame' in that context.

As an image sequence may require a sequence to be in a specific order the hack/workaround to reassigning the first frame is not a viable solution.

All this to suggest there is much Future work to do.
The challenge will be to do this without breaking the PLI (and TLV) files.
The goal, in my estimation, is to introduce the concept and process of being able to set a poster frame.  For demo purposes it could be reasonable to use the MOV format which does support poster frames.

Appending drawings establishes that a PLI can serve as more than a single brush asset. The next step is to separate the first drawing in a level from the drawing chosen to represent or supply that level. A backward compatible representative drawing mechanism could initially support thumbnails and previews, then later allow vector brushes to select among multiple drawings without reordering the level.  This also relates to processes such as Trail Brushes where a pointer would allow the user to set a starting frame from any given sequence.

RodneyBaker:
Pressing forward with this to resolve user facing challenge of creating custom vector brushes.
The process intentionally directs the brush in a way to avoid creating PLI files not ideal for use as vector brushes.  The process mostly automates the manual process of saving PLI files to vector brushes directory but via selection rather than entire PLI vector file.

This process attempts a 'fix' for the creation of custom vector brushes by bypassing the manual process used before which is still available and not modified.

Issue #1846, opened 2018-03-06
Poor dialog and floating window hierarchy


artisteacher:
It's great no longer having the render dialog display on top of all open programs. However, when rendering directly from Output Settings, the render dialog displays underneath.
![screen shot 2018-03-06 at 8 10 02 am](https://user-images.githubusercontent.com/21159116/37034121-393723be-2116-11e8-80ee-12964b1a0af7.png)

There are occasionally other windows that unexpectedly display underneath something else when they are opened. It would be great if newly opened windows/dialogs always displayed on top of everything else - until the focus has changed.

artisteacher:
Still happens on MacOS. I agree that the render dialog as it’s own window doesn’t have a great location either way but on top is more expected. 

 I really like how in Blender, the render progress bar (with an X to cancel) displays at the top of the screen & the progress bar also displays on the app icon in the dock.

RodneyBaker:
Will tag as OS:Mac pending confirmation on other systems.

I hope to stare at the whole rendering flow more in the future.
I'd love to see the Farm Room become more of an all purpose Render room because many users likely see 'Farm' and say "I don't need no Farm' and they never explore that room again.
If it was a Render Room that also just happened to be able to render to network as well as local machine (which it can already) then users might be more inclined to explore it.  Then the Render Room itself could be plus'd up in useful ways.

And, because we can render behind the scenes with the Task/batch rendering, I hope someday we can gain an option to render to/through the batch renderer via the normal output settings dialog.  The benefit being we could launch our render in the background and immediately get back to work on the next thing.

artisteacher:
Render dialog still displays beneath the output settings.

artisteacher:
On a related note, the camera capture window displays on top of the preferences window, even when preferences is opened 2nd.

RodneyBaker:
@artisteacher  Checking in to verify this issue is still valid and persisting.
It would be good to confirm if you are experiencing the issue in Tahoma2D as well.

artisteacher:
Yes, it's definitely ongoing in both programs. Window focus is a little unpredictable in general. 

The render progress bar displays below the output settings window. Clicking on the render progress bar doesn't change the stacking order.

Some windows - ie. Output Settings, Flipbook or etc. always display above the program window. They might change stacking order with most other floating windows, but they never end up behind everything.

[...]

PR #7097, opened 2026-08-29
Fix macOS Preferences window hierarchy

RodneyBaker:
## Summary

Removes the macOS only `Qt::Tool` override from Preferences. 
Preferences remains a `QDialog` parented to the OpenToonz main window, preserving the normal application window hierarchy.

Addresses the Preferences case reported in #1846 without changing shared dialog behavior.

ChatGPT 5.6 was used to research and proccess this fix.


RodneyBaker:
Pressing forward with this simple change to make sure MacOS windows stack correctly.
@artisteacher Always appreciate your time to validate whenever available.  Thanks!

PR #7098, opened 2026-08-29
Add responsive level strip thumbnails option

Ahl76:
**The problem**

Level Strip thumbnails stay at the fixed size from Preferences, even when the panel is much wider or taller. On a large strip you end up with small previews and a lot of empty space. Making them bigger only through Preferences is clumsy if you just want to use the room you already have.

**Porposal**

responsive thumbnail

This PR adds a responsive Thumbnails option in the Level Strip panel settings.It's an optional way to make Level Strip thumbnails use the free space in the panel, so you can see larger previews without changing the Preferences size every time. You can leave it off if it does not fit how you work.

When it is on, thumbnails grow to use the free space in the panel (across the strip in vertical layout, along the strip height in horizontal layout), while keeping the same proportions as the Preferen

ces thumbnail size. When it is off, behaviour is unchanged: Preferences size only.

In practice this feels like a quick zoom on the strip, handy when you want to check details without leaving the Level Strip. It should also work well with the frame marks support I plan to propose next. It can open up new ways of working with the strip, depending on how you arrange your room.

The setting is saved with the panel layout, like the other Level Strip options.

While you resize the panel, previews stay usable; once you stop, they refresh at a sharper size. Works for raster, Toonz Raster, vector and mesh levels.

**Notes**

Off by default. Existing layouts keep the classic fixed size until you turn it on.

**Limitations**

Adjacent panels may change size or position when the strip grows.

### DIFF
--- a/toonz/sources/include/toonzqt/icongenerator.h
+++ b/toonz/sources/include/toonzqt/icongenerator.h
@@ -113,6 +113,8 @@ public:
                   bool onDemand = false);
   QPixmap getSizedIcon(TXshLevel *sl, const TFrameId &fid, std::string newId,
                        TDimension dim = TDimension(0, 0));
+  QPixmap getResponsiveIcon(TXshLevel *sl, const TFrameId &fid,
+                            const TDimension &dim);
   void invalidate(TXshLevel *sl, const TFrameId &fid,
                   bool onlyFilmStrip = false);
   void remove(TXshLevel *sl, const TFrameId &fid, bool onlyFilmStrip = false);
--- a/toonz/sources/include/toonzqt/icongenerator.h
+++ b/toonz/sources/include/toonzqt/icongenerator.h
@@ -101,7 +101,7 @@ public:
 
   TDimension getIconSize() const;
 
-  TOfflineGL *getOfflineGLContext();
+  TOfflineGL *getOfflineGLContext(const TDimension &minSize = TDimension());
 
   // icons from splines
   QPixmap getIcon(TStageObjectSpline *spline);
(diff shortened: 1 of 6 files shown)


RodneyBaker:
@Ahl76 
I don't mention it in the attached but when merged this option for responsive thumbnails should be the default.  Seriously, save a tree and just add it now.  This PR fixes a serious visual problem that users have worked with for 10 years and, if they prefer that, the option can be turned off.  :)

Attached PDF submitted for your consideration:

[OpenToonz_PR_7098_Blocking_Concerns_Review.pdf](https://github.com/user-attachments/files/31604069/OpenToonz_PR_7098_Blocking_Concerns_Review.pdf)

RodneyBaker:
@Ahl76 
Performatively, this change is working great.
Once code concerns are addressed (or otherwise documented) this can move forward and be merged.

Ahl76:
The five points from the PDF review are in this commit, with you as co-author, as this reuses what you did. The preference is now on by default.

RodneyBaker:
> The preference is now on by default.

Well, in that case... posting intent to merge.
The improvement over previous Level Strip visualization is obvious.

The residual concerns of this PR are maintenance/performance territory rather than reasons to hold this PR, particularly the added complexity in IconGenerator and long term behavior with unusually large Level Strip dimensions (multi-monitor?). Those are appropriate follow up observations and not merge blockers.

As the datetime stamp of this PR makes it look like new work and not previously refined at OT-Dev there will a slight delay for others willing and able to test. (Someone mark the time)

RodneyBaker:
Moving forward on this -fix-.
Thank you @Ahl76 
LGTM

PR #7099, opened 2026-08-29
Fix crash when changing frame with arrows

Ahl76:
I first hit this while testing #7098. It is not caused by that option: the same crash happens on the nightly build, without it. This is the separate fix.

Vector: drop the in-progress stroke when the frame changes. Do not use a missing image.

Toonz Raster and Raster: finish the interrupted stroke on the frame where it started, with its undo. Toonz Raster no longer creates a cell on the leftover mouse release; it matches Raster. Do not use a missing image.

Vector drops the stroke because after a frame change the current image is gone or is another drawing. Raster can finish on the start frame because those pixels live there. Switching tools already commits via finishTracks(); changing frame with the arrows is not the same case for Vector.

### DIFF
--- a/toonz/sources/tnztools/fullcolorbrushtool.cpp
+++ b/toonz/sources/tnztools/fullcolorbrushtool.cpp
@@ -34,6 +34,7 @@
 #include "toonz/imagestyles.h"
 #include "toonz/preferences.h"
 #include "toonz/toonzfolders.h"
+#include "toonz/txshsimplelevel.h"
 
 // TnzCore includes
 #include "tgl.h"
@@ -142,7 +143,8 @@ FullColorBrushTool::FullColorBrushTool(std::string name)
     , m_notifier(0)
     , m_presetsLoaded(false)
     , m_firstTime(true)
-    , m_started(false) {
+    , m_started(false)
+    , m_strokeFrameId() {
   bind(TTool::RasterImage | TTool::EmptyTarget);
   m_thickness.setNonLinearSlider();
 
@@ -272,6 +274,7 @@ void FullColorBrushTool::updateWorkAndBackupRasters(const TRect &rect) {
   if (!ri) return;
 
   TRasterP ras = ri->getRaster();
+  if (!ras || !m_workRaster || !m_backUpRas) return;
 
   const int denominator = 8;
   TRect enlargedRect    = rect + m_lastRect;
@@ -317,6 +320,7 @@ bool FullColorBrushTool::askRead(const TRect &rect) { return askWrite(rect); }
 
 bool FullColorBrushTool::askWrite(const TRect &rect) {
   if (rect.isEmpty()) return true;
+  if (!m_tileSaver || !m_workRaster) return false;
   m_strokeRect += rect;
   m_strokeSegmentRect += rect;
   updateWorkAndBackupRasters(rect);
@@ -368,6 +372,13 @@ bool FullColorBrushTool::preLeftButtonDown() {
 void FullColorBrushTool::handleMouseEvent(MouseEventType type,
                                           const TPointD &pos,
                                           const TMouseEvent &e) {
+  if (m_skipStrokeUntilDown) {
+    if (type == ME_DOWN)
+      m_skipStrokeUntilDown = false;
+    else if (type != ME_MOVE)
+      return;
+  }
+
   TTimerTicks t = TToolTimer::ticks();
   bool alt      = e.getModifiersMask() & TMouseEvent::ALT_KEY;
   bool shift    = e.getModifiersMask() & TMouseEvent::SHIFT_KEY;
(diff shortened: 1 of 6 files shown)


RodneyBaker:
Thanks @Ahl76 

The primary merge blocker with the proposed fix is in `FullColorBrushTool::commitStroke()`.

If `sl->getFrame(frameId, true)` does not return the stroke’s original image, the fallback

`if (!ri) ri = (TRasterImageP)getImage(true);`

can obtain the newly current frame’s image. 

The code nevertheless constructs `FullColorBrushUndo` using `frameId`, which still identifies the frame where the stroke began.

This mixes pixels from one frame with the undo identity of another. 

If the frames have different dimensions, `ras->extract(m_strokeRect)` may be out of bounds; if their dimensions match, the undo can capture unrelated pixels and later restore them into the stroke’s frame.

`ToonzRasterBrushTool::commitPainting()` does not allow its current-image fallback to feed an undo because undo registration also requires sl. 

Recommend:  Make the full-color path equally safe. 
For this PR, removing the fallback and skipping undo creation when the original frame image cannot be retrieved is the safest minimal correction.

Assessment:
A stronger (future?) implementation could retain or otherwise reliably recover the originating image, but the current fallback must not be used with the original frameId.

Submitted for your consideration.

RodneyBaker:
The following is considered minor but well worth addressing:

m_skipStrokeUntilDown is never cleared in onActivate/onDeactivate. 

It self-heals on the next ME_DOWN, but resetting it is cheap hygiene.

The skip check sits above the alt/shift/ctrl keyEvent sync in handleMouseEvent, so modifier state can drift while ME_DRAG/ME_UP are dropped. 

ME_MOVE resyncs it, so impact is low as it is moving the guard below the key sync is cleaner.

Worth stating explicitly in the PR description why vector discards the stroke while raster commits it. 

A user who nudges an arrow key mid stroke loses work in one tool and not the other, and onDeactivate already uses finishTracks() (commit) for the analogous tool-switch case.

RodneyBaker:
**Recommended tests:**

- [x] Using Vector brush, hold the button and press an arrow key, keep dragging, release
- [x] Check for a stray stroke on the new frame and that undo doesn't remove a pre-existing frame
- [x] Check Vector frame range mode with Ctrl-continue; verify the red preview survives
- [x] Check Toonz Raster on empty destination frame; confirm no cell is created on leftover release
- [x] Undo immediately after the interrupted stroke in each tool, verifying the stroke lands on the frame where the stroke started

Ahl76:
@RodneyBaker

I followed your recommendations:

- Raster undo does not fall back to the current image
- Skip undo when the start-frame image is missing
- Reset the leftover-stroke skip on activate/deactivate
- Apply that skip after the modifier keys

RodneyBaker:
The important changes look correct:
- [x] Full Color / Raster: the unsafe fallback to getImage(true) is gone from commitStroke().
- [x] Undo is now created only when the original frameId can be resolved through the simple level and yields a valid raster.  **This addresses the main merge blocker**
- [x] Toonz Raster: commitPainting() similarly prefers sl->getFrame(frameId, true) and requires both ras and sl before constructing the undo. The current frame image cannot accidentally be paired with the old frameId.
- [x] Leftover mouse events: m_skipStrokeUntilDown is reset in activate/deactivate and its check now comes after modifier synchronization, exactly as requested.
- [x] Vector: an image change resets the input manager and clears the in progress tracks, matching the documented decision to discard an interrupted Vector stroke rather than attempt to commit it.
- [x] The PR description explicitly explains why Vector discards while raster tools commit to the originating frame.

Additionally:  The tile saver/tile set ownership appears intentionally transferred to the undo when one is created and explicitly cleaned up otherwise. Null-image/null-raster checks have also been added around several previously optimistic accesses.

Assessment:  Nice work!

Recommendation:  Take this PR out of Draft and let's press toward merger.

Ahl76:
The last push corrects two Vector points that were still wrong after the last commit:

- The frame range preview stays on the viewer when the cursor leaves
- Leftover mouse input after a dropped stroke does not write on the new frame

RodneyBaker:
@Ahl76 
My 'final pass' exposed the same class of bug as before but one level higher: the raster tools remember the starting frame, but not the starting level.

Discussion
When the frame changes, TApp::onFrameSwitched() calls updateXshLevel(). In scene editing, updateXshLevel() obtains the level from the newly selected cell and updates m_currentLevel.

 Thus an arrow key frame change can change the current level when adjacent cells contain different levels.

Full Color is vulnerable: commitStroke() correctly uses the saved m_strokeFrameId, but obtains sl from app->getCurrentLevel() at commit time. 

If the arrow moves from Level A to Level B, it therefore looks for A's frame ID inside B and constructs FullColorBrushUndo against B. 

The undo object explicitly binds its tile set to the supplied level/frame, so this is not merely a notification issue.

Toonz Raster has the same issue: m_painting.frameId is retained, but commitPainting() resolves sl from the current level after the switch. 

The Painting state currently stores frameId but no originating level.

Concrete failure case
- [ ] Put Raster Level A frame 1 and Raster Level B frame 1 in adjacent rows of the same column. 
- [ ] Begin a stroke on A, keep the mouse down, arrow to B, then release. 

Because both levels contain frame ID 1, the code can obtain B.1's raster, combine it with the tile set produced while painting A.1, and register the undo against B. Undo can then leave the stroke on A while modifying B. Different raster dimensions could potentially bring back an invalid extract() case.

Minimal correction (?)
Capture a TXshSimpleLevelP at stroke start alongside the frame ID. 
commitStroke() / commitPainting() should retrieve the originating image from that stored level and construct the undo against it. 
Then call notifyImageChanged(frameId, originatingLevel) rather than relying on the current level; that overload specifically accepts an explicit TXshSimpleLevel *.

I'm not quite following all of this but hopefully you can make sense of it.

Ahl76:
@RodneyBaker

Can you give another look ? This should address it : 

Raster and Toonz Raster now keep the starting level with the frame ID, so commit, undo, and notifyImageChanged stay on that level after the arrow.

PR #7102, opened 2026-08-30
(Proposal) Set Brush as default tool for new scenes (Do not Merge)

RodneyBaker:
Select the Brush tool instead of the Hand tool when a new scene is created.

Note: This is 2 of 2 related PRs changing the default from the Hand Tool to the Brush Tool.
See also:  #7101

Changes the default tool selected when OpenToonz starts from the Hand Tool to the Brush Tool.
This reduces the step between launching OpenToonz and beginning to draw while otherwise leaving tool behavior unchanged.

Because this changes a longstanding default, feedback is welcome particularly from anyone who believes the Hand Tool, or another tool, should remain the default at startup.

Original author: **AHL (@Ahl76)**.

Authored with assistance from Cursor
(Processed with assistance from ChatGPT 5.6)
Based on OpenAnimationLibrary/opentoonz-dev#70.

### DIFF
--- a/toonz/sources/toonz/iocommand.cpp
+++ b/toonz/sources/toonz/iocommand.cpp
@@ -1378,7 +1378,7 @@ void IoCmd::newScene() {
   ToolHandle *toolH = TApp::instance()->getCurrentTool();
   if (toolH && toolH->getTool()) toolH->getTool()->reset();
 
-  CommandManager::instance()->execute("T_Hand");
+  CommandManager::instance()->execute("T_Brush");
 
   CommandManager::instance()->enable(MI_SaveSubxsheetAs, false);
 


KaseyFarron:
I prefer it to be Hand, otherwise stray dots will appear in my scanned (from camera) drawings, will take ages to change muscle memory... Hand seems like a safe starting point until you know what you're doing in the scene. 🤷‍♀️

RodneyBaker:
Thanks for the feedback @KaseyFarron 

> otherwise stray dots will appear in my scanned (from camera) drawings

👀 We probably should investigate and err... fix that.

RodneyBaker:
If this change isn't implemented then, in my estimation, that rules out implementing #7101 as I'm pretty sure that would be problematic for @KaseyFarron's use case.

Some alternatives (moving back to globally setting the default for both startup and creation of scenes) might be a Preference on Tools tab to set the default tool.  Preferrably for both startup and scene creastion to keep things simple.**  There are surely other alternatives as well but almost all start to move toward unnecessary complexity.

**The thought behind separating the two events is that if opening a scene the likelihood of using the Hand Tool increases.  If creating a new scene the likelihood of drawing increases.  If simply starting the program a goal or artists and animators is more often than not simply to start drawing immediately.  Originally having the startup change the default forces the decision onto both workflows.

Personally, I prefer Opentoonz to start and the user immediately be ready to draw.
I do understand that due to their role and workflow many users of Opentoonz may never draw in Opentoonz... ever.

There's a solution in here somewhere.

Abel032:
What about setting the **default tool for startup** to the **tool when app quitting**.

RodneyBaker:
> What about setting the **default tool for startup** to the **tool when app quitting**.

I like this idea @Abel032.
If you'll forgive the unnecessary complexity my mind wanders toward:
The default might change only after a specific count of 'same tool used during shutdown' is reached.  In this way Opentoonz would 'learn' the workflow preferences of its users but not immediately change when there is no pattern of preference.

The simplicity of YOUR suggestion (not my expansion on it) is very appealing to me.

Added:
There is some (potentially) parallel development for settings changes here:  #6977

All,
I'm going to investigage what success might look like with @Abel032's suggestion as that is simple and direct.  There are some outliers to account for such as making sure we account for crashes etc. as well as future considerations for saving other settings across sessions (not to mention transferring settings to new installations) but I want to implement the minimal viable product the idea suggests to address @Ahl76' original intent which was to have the Brush Tool be the default upon startup.  Abel's suggestion can allow for this... although... the user would want to make sure they aren't on a undesired Tool for startup when they close down the application.  It seems likely to me that Tool might more often than not be something other than the Brush Tool which would defeat the intent of the proposed change.

cc:  @Ahl76

RodneyBaker:
RE:  

> What about setting the default tool for startup to the tool when app quitting.

While I love this approach I'm backing off of it as it doesn't quite address @Ahl76's original intent.  The approach can too easily set ANY Tool to be the startup Tool which defeats the stated purpose:  Start Opentoonz with the Brush Tool active.

@Abel032's idea is more than valid and should be investigated for development.
Specifically I'd recommend it be considered with respect to a means of saving and setting preferences across installations (and portable applications).  As there is more than just Tool settings to consider this might be expanded to an option to remember last settings so a user can pick up exactly where they left off from session to session and to transfer and store settings for reuse or sharing.

Where this places my attention:  I will investigate what success might look like for adding a simple option to Preferences > Tools where either a default Tool during startup can be set or an option to recall last used Tool at Startup can persist.

Thanks for your continuing ideas and feedback.

RodneyBaker:
@Ahl76, @KaseyFarron, @Abel032 I've opened new PR thttps://github.com/opentoonz/opentoonz/pull/7116 to address the need to maintain current default Tool activated at startup and new scene creation.

That PR adds a Preferences > Tools > Default Startup and New Scene Tool dropdown.  While these two events can be set the same in Preferences they can be set differently in preferences.ini.  (Ex:  Startup can use the Brush Tool and creation of a New Scene can use the Hand Tool)

image

PR #7104, opened 2026-09-01
Convert NAAimage to TLV: Added BG Color Option

shun-iwasawa:
This PR adds a new checkbox to ConvertDialog that is enabled only when converting to TLV with the `“Painted tlv from non-AA source”` mode.

New option: `Only white or transparent pixels can be identified as the background`

Under the conventional specification, the color identified as the background pixel (StyleId = 0) was “the brightest color in the image, where the value of each channel is 230 or higher.”

In most cases, the color that met this condition was pure white, so it wasn’t a particular issue. However, especially in scenes with animation BG (背景動画), when converting images where the entire frame is filled with color, a problem occurs where parts of the image are unintentionally converted to transparent (i.e., the background style).

Therefore, we have added an option to limit the colors identified as the background to pure white or transparent.

### DIFF
--- a/toonz/sources/include/toonz/Naa2TlvConverter.h
+++ b/toonz/sources/include/toonz/Naa2TlvConverter.h
@@ -108,6 +108,7 @@ public:
   double m_inkThickness;
   TPalette *m_palette;
   bool m_valid;
+  bool m_allowOnlyWhiteBG;
 
   Naa2TlvConverter();
   ~Naa2TlvConverter();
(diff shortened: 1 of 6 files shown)


RodneyBaker:
Thanks @shun-iwasawa 

The label and tooltip say "white or transparent," but the code only compares against `TPixel::White`.  That appears intended as `setSourceImage()` runs every source pixel through `overPixOnWhite()` so fully transparent pixels have already become pure white by the time `m_colors` is built. 

The consequence is that semi-transparent pixels get composited to off-white and are therefore excluded.  This appears to be the intention of the PR but as an implicit dependency that isn't commented anywhere this is a note to identify the treatment of semi-tranparent pixels.  

In other words and if I understand correctly, with this new option on, any semi-transparent pixels in the source images will not be treated as background upon conversion.

RodneyBaker:
Much appreciated.

I see no reason to delay merger.

PR #7109, opened 2026-09-04
Add stuff/doc/index.html 

RodneyBaker:
Adds a simple index.html linking all current html and pdf documents already present in stuff/doc directory.  

Purpose
The index.html file adds a missing index file to stuff/docs directory that lists the available docs.  It also serves as a landing page for FX help links to point to where a dedicated document does not yet exist.

_This index.html file will be updated.  It's initial placement is to encourage development and maintenance of doc files distributed with Opentoonz._

A link to online Opentoonz manual is also provided as that is the official source of Opentoonz documentation.


RodneyBaker:
Merging as this is a stuff/doc file

konero:
I'm not sure this was a good idea, the docs dir just contains FX manuals which are already linked on the FX page, the index file code is also minified so it can't be read cleanly. HTML is also not good for docs, you'd want something that auto-generates the index.

Unless I missed something, what's wrong with the online manual for help?

RodneyBaker:
@konero I've edited the PR description to provide a little more detail on purpose.
I've posted a more lengthy explanation in the Opentoonz discord that touches upon more background and rationale as that venue seems more appropriate for general discussion.

konero:
The solution here was to update the online manual for many reasons, Google indexing, AI answers, etc., a customizable help command is not very intuitive and can't expect a user to want to make their own index.

Source should not be minified since we can't see what's in the code properly, can't maintain it.

RodneyBaker:
> The solution here was to update the online manual for many reasons, Google indexing, AI answers, etc., a customizable help command is not very intuitive and can't expect a user to want to make their own index.
> 
> Source should not be minified since we can't see what's in the code properly, can't maintain it.

Fully understood and agree with your assessment of what the ultimate goal is for us to plan and execute.

What (if anything) do you propose we do with the contents of stuff/docs (pdf and html) and the links that some FX have to those documents? 

This PR is a first step in getting to your solution but as of this moment only addresses the contents of the local stuff docs directory.  A logical next step is to make sure all FX help links point to where we want them to go so your answer to this question is all the more important.  I have concerns about where those links will point in cases where online documentation is not available for one reason or another.  Most FX (but not all) currently point to files in the stuff/docs directory.

RodneyBaker:
It may not be entirely obvious why we currently have local stuff/doc content.
The following should help in that regard.

The content at stuff/doc is not the same common user interfacing documentation provided to users in the online manual.  Content in stuff/doc is provided for distribution with Opentoonz by the developers and is technical and administrative in nature.  The focus of stuff/doc content is on locally stored documentation shipped with the software.  

Most documents in stuff/doc are focused on effects although in this first index.html file I neglected the licensing files in stuff/doc.  In addition to technical details, these documents often contain licensing and attributation not found in or prescribed for inclusion in the online manual.   These local documents therefore ensure licensing and attribution requirements are met where required in the local distribution.

Todo:  Unminify html and include references to the licenses documentation located in stuff/doc/LICENSE.  (Licensing docs are currently referenced in index.html under Supporting Files but should have their own category)

Future:
As a separate PR, repair broken FX Help links to local documentation.  Consider benefit of directing the Help link to the index.html** file rather than the individual documents.  Create documenation for missing entries.

**or online Opentoonz manual if if appropriate

PR #7113, opened 2026-09-06
Fix scientific notation in expression tokenizer

### LINKED ISSUE #7112: Expression tokenizer splits scientific notation with a single unsigned exponent digit
# Expression tokenizer splits scientific notation with a single unsigned exponent digit

## Summary

`TSyntax::Tokenizer::setBuffer()` splits `1.0e2` into a Number token `1.0` and an Ident token `e2`. Equivalent spellings `1.0e+2` and `1.0e02` correctly produce one Number token. The same problem occurs for uppercase `E` and exponent zero.

This is a source-level reproduction against upstream commit `868d795e7d80687af7afad7866ffc4e1997dca30`.

## Affected code and cause

https://github.com/opentoonz/opentoonz/blob/868d795e7d80687af7afad7866ffc4e1997dca30/toonz/sources/common/expressions/ttokenizer.cpp#L72-L96

The exponent condition always requires `s[i + 2]` to be a digit, even when `s[i + 1]` is already the first exponent digit. Thus a one-digit unsigned exponent at the end of a literal is rejected, while a signed exponent or an unsigned exponent with two digits is accepted.

Additionally, exponent scanning is nested inside the decimal-point branch, so `1e2` and `1e+2` also split incorrectly.

## Local reproduction

Save the following as `/tmp/opentoonz-exponent-repro.cpp`:

```cpp
#include "ttokenizer.h"
#include <iostream>
int main() {
  int failures = 0;
  for (const char *input : {"1.0e2", "1.0E2", "1.0e0", "1.0e+2", "1.0e-2", "1.0e02", "1e2", "1e+2", "100", "0.01"}) {
    TSyntax::Tokenizer tokenizer(input);
    bool ok = tokenizer.getTokenCount() == 2 &&
              tokenizer.getToken(0).getType() == TSyntax::Token::Number &&
              tokenizer.getToken(0).getText() == input;
    std::cout << (ok ? "PASS " : "FAIL ") << input << " ->";
    for (int i = 0; i < tokenizer.getTokenCount() - 1; ++i) {
      auto token = tokenizer.getToken(i);
      std::cout << " [" << (token.getType() == TSyntax::Token::Number ? "Number" : token.getType() == TSyntax::Token::Ident ? "Ident" : "Punct") << ": " << token.getText() << "]";
    }
    std::cout << '\n';
    failures += !ok;
  }
  std::cout << failures << " / 10 cases failed\n";
  return failures ? 1 : 0;
}
```

From the OpenToonz repository root, on Linux with Qt5Core development files installed:

```sh
c++ -std=c++17 -DLINUX -fPIC -I toonz/sources/include \
  $(pkg-config --cflags Qt5Core) /tmp/opentoonz-exponent-repro.cpp \
  toonz/sources/common/expressions/ttokenizer.cpp \
  $(pkg-config --libs Qt5Core) -o /tmp/opentoonz-exponent-repro
/tmp/opentoonz-exponent-repro
```

Expected: each input produces exactly one Number token followed by Eos, and the program exits 0.

Actual:

```text
FAIL 1.0e2 -> [Number: 1.0] [Ident: e2]
FAIL 1.0E2 -> [Number: 1.0] [Ident: E2]
FAIL 1.0e0 -> [Number: 1.0] [Ident: e0]
PASS 1.0e+2 -> [Number: 1.0e+2]
PASS 1.0e-2 -> [Number: 1.0e-2]
PASS 1.0e02 -> [Number: 1.0e02]
FAIL 1e2 -> [Number: 1] [Ident: e2]
FAIL 1e+2 -> [Number: 1] [Ident: e] [Punct: +] [Number: 2]
PASS 100 -> [Number: 100]
PASS 0.01 -> [Number: 0.01]
5 / 10 cases failed
```

The program exits 1, indicating the failed expectations.

## Impact and verification scope

This tokenizer is consumed by the expression parser. `NumberPattern::createNode()` converts the Number token to a numeric value, so the intended numeric literal is already broken before expression evaluation. The expression field also uses this parser for syntax highlighting.

Confirmed here: incorrect token streams from the unchanged production tokenizer. The full evaluator and GUI behavior have not been tested; this report does not claim a specific displayed value or GUI error.

A workaround for these inputs is to include a decimal point and an explicit exponent sign, e.g. `1.0e+2`.

## Suggested correction

Scan the exponent after the optional fractional part, independently of whether a decimal point occurred. After `e`/`E`, skip an optional sign and require at least one digit. Add regression coverage for the cases above and malformed exponents such as `1.0e`, `1.0e+`, and `1.0e-`.

No fix has been applied or validated as part of this report.

## Environment

- Linux
- GCC 16.0.1 20260319 (experimental), C++17
- Qt5Core 5.15.13
- Upstream master at `868d795e7d80687af7afad7866ffc4e1997dca30`
- No full application build or GPU required for this reproduction

killerdevildog:
Fixes #7112.

Expression literals such as `1.0e2` and `1e+2` were split into multiple tokens, while `1.0e+2` worked. Scan the exponent independently of the decimal point, allow an optional sign, and require a digit before consuming the exponent. A bare decimal point cannot serve as the mantissa.

This PR changes only `toonz/sources/common/expressions/ttokenizer.cpp`.

Validation: the original 10-case reproduction now passes; a local 34-case regression harness covering numeric values, incomplete exponents, token positions, and surrounding expression tokens passes. The same harness detects failures against the original source. AddressSanitizer, UndefinedBehaviorSanitizer, and LeakSanitizer checks pass. Test harnesses remain local and are not included in this PR. Formatting checked with clang-format 18; CI uses clang-format 14. The full application and GUI were not tested.

### DIFF
--- a/toonz/sources/common/expressions/ttokenizer.cpp
+++ b/toonz/sources/common/expressions/ttokenizer.cpp
@@ -81,15 +81,17 @@ void Tokenizer::setBuffer(std::string buffer) {
         token.append(1, s[i++]);
 
         while (isascii(s[i]) && isdigit(s[i])) token.append(1, s[i++]);
+      }
 
-        if ((s[i] == 'e' || s[i] == 'E') &&
-            (((isascii(s[i + 1]) && isdigit(s[i + 1])) ||
-             s[i + 1] == '-' || s[i + 1] == '+') && isascii(s[i + 2]) &&
-                 isdigit(s[i + 2]))) {
-          token.append(1, s[i++]);
+      if (token != "." && (s[i] == 'e' || s[i] == 'E')) {
+        // Only consume the exponent if its optional sign is followed by a
+        // digit.
+        int exponent = i + 1;
+        if (s[exponent] == '-' || s[exponent] == '+') ++exponent;
 
-          if (s[i] == '-' || s[i] == '+') token.append(1, s[i++]);
-
+        if (isascii(s[exponent]) && isdigit(s[exponent])) {
+          token.append(s + i, exponent - i);
+          i = exponent;
           while (isascii(s[i]) && isdigit(s[i])) token.append(1, s[i++]);
         }
       }


RodneyBaker:
Thanks @killerdevildog 

This appears to be correct, minimal, well-scoped and safe to merge. 

I believe use of a bool instead of the "." comparison would be preferred although I'm not sure that is a merge blocker.

RodneyBaker:
I'm moving toward merging but still haven't explicitly tested for the following:

- [x] Malformed exponents — 1.0e, 1.0e+, 1.0e-, 1e. These must not swallow the e into the number and must not crash. Correct output is a number followed by an identifier.

- [x] Exponent followed by real math — 1e2+3 should be [1e2][+][3] = 103, not [1][e2][+][3].

- [x] Exponent next to a real variable — something like 1e2*x or an expression using a defined parameter name, to confirm the greedier number scanning doesn't start eating identifiers.

- [x] The lone-dot change — .5 and a bare . on its own. This is the one behavioral change beyond the exponent fix, and it's the most likely place to have broken something that previously worked.

- [x] Values, not just tokens — confirm 1.0e2 actually evaluates to 100.0, since correct tokenization and correct NumberPattern::createNode() conversion are two different things.

killerdevildog:
Thanks @RodneyBaker — I tested your checklist locally, and all the checks passed, including tokenization and evaluated values. `.5` and a bare `.` also retain their previous behavior.

RodneyBaker:
Thanks @killerdevildog 

Nightlies haven't all been building so this may nor may not appear in the nightlies soon but if not it will eventually catch up.

Thanks for the fix.

PR #2355, opened 2018-11-01
(WIP) Fix Vector Flipping on Using Tape Tool

shun-iwasawa:
This will fix #2347 
This is WIP, as I think this fix is need to be evaluated regarding the following behavior:
When applying the tape tool with "Join Vectors" and "Smooth" options being on, if the target stroke is a part of some filled region, the result sometimes causes unexpected change of styles used for filling (i.e. the fill color sometimes changes). 

I'm not sure if it is acceptable for users, since the current OT already have such behavior when modifying vector strokes.

PR #7114, opened 2026-09-07
Preserve Tape Stroke Direction

### LINKED ISSUE #2347: Using the "Tape Tool" changes the direction of the vector.
### Issue Summary

Using the Tape Tool changes the direction of the vector, which leads to incorrect auto-inbetweening.

ezgif-4-8223b249b260

### Steps to Reproduce

1. Create a vector layer with more than two frames (for example, 12)
2. Draw (from left to right) in the first frame two non-intersecting lines as in the video.
3. Copy and paste them into the last frame (12)
4. Select all the frames in the Level Strip and perform the "Inbetweens" operation (the interpolation doesn't matter)
5. Go to the first frame, select the Tape Tool (Normal, Endpoint to Line, Join Vectors, Smooth)
6. Connect the end of one of the lines with the other line (do not connect the ends, so as not to form one solid vector)
7. Select all the frames in the Level Strip and perform the "Inbetweens" operation again. As a result of applying the Tape Tool, the second vector changed direction.

### Expected Results

The direction of the vector should not be changed, since the "Tape Tool" was used for editing.
### Actual Results

The direction of the vector changes if the "Tape Tool" has been applied to the beginning of the vector.
### System Information

RodneyBaker:
Fixes #2347

Preserve vector orientation when Tape extends a stroke at its beginning, with Smooth on or off, and adjust fill edge mapping for that orientation. 

Includes a precautionary confirmation before joining or extending strokes in filled drawings. 
Cancel is the default, and an optional remembered choice is stored silently in preferences.ini without a Preferences window entry. Rectangular gestures ask once before making changes.

Adapted from Shun Iwasawa's original fix in opentoonz/opentoonz#2355 (94e1643c8aa044438f65735b224346739a49f400).

Implementation and review assisted by ChatGPT (Codex).

### DIFF
--- a/toonz/sources/common/tvectorimage/tvectorimage.cpp
+++ b/toonz/sources/common/tvectorimage/tvectorimage.cpp
@@ -2001,7 +2001,7 @@ assert(m_strokes[strokeIndex-wSize+1]->m_edgeList.empty());*/
 //-----------------------------------------------------------------------------
 
 static void computeEdgeList(TStroke *newS, const std::list<TEdge *> &edgeList1,
-                            bool join1AtBegin,
+                            bool join1AtBegin, bool stroke1isFlipped,
                             const std::list<TEdge *> &edgeList2,
                             bool join2AtBegin, std::list<TEdge *> &edgeList) {
   std::list<TEdge *>::const_iterator it;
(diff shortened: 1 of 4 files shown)


shun-iwasawa:
I have confirmed that it works as proposed.
One concern, perhaps, is that once a behavior is selected using “Remember my choice,” there is no way to change it later via the GUI.
I think we can release it as is, and if users raise any issues, we can consider adding an option to Preferences.

LGTM Thank you for implementing this!

PR #7091, opened 2026-08-27
Add shortcut command to toggle active LUT

RodneyBaker:
## Summary

- Add a **Toggle Active LUT** command under the View shortcut category
- Keep the shortcut unassigned by default and add no menu item
- Synchronize the color calibration Preference and refresh LUT dependent views immediately

Use
Assign a key under **Preferences → Shortcuts → View → Toggle Active LUT**.

Initial validation via tests from [OT-Dev PR #104](https://github.com/OpenAnimationLibrary/opentoonz-dev/pull/104) using ChatGPT 5.6.

PR #7115, opened 2026-09-07
Add 3D LUT Bake FX

RodneyBaker:
Adds an **Image Adjust > 3D LUT Bake** FX node that applies a selected `.cube` or supported Lustre `.3dl` LUT directly to rendered image colors.

For an overall look, place the node after the completed composition. 
The effect preserves alpha and supports 8-bit, 16-bit, and floating-point rasters.
LUT parsing and trilinear sampling are shared with the existing Preferences display LUT feature.

The node provides:

* **LUT File / Browse…** — Select a LUT by entering its path or using the filtered file browser. Invalid selections display an inline error and retain the previous path. An empty path bypasses the effect.
* **Copy Preferences LUT** — Copies the LUT path configured for the current monitor, including when display calibration is disabled. This is a one-time convenience action; later Preferences changes do not update the FX, and the LUT file is not embedded in the scene.
* **Fx Intensity** — Controls the contribution of the effect from 0 to 1 and supports animation.
* **Help** — Opens `stuff/doc/3DLUTBakeFx.pdf`, which documents the controls, accepted files, color handling, and troubleshooting.

The implementation addresses the architectural and error-handling concerns raised during review:

* Refactors `TFilePathParam` to derive from `TStringParam`, with file-filter metadata, and adds a reusable `FilePathParamField` containing a text field and Browse button.
* Recognizes file-path parameters before the broader string type while preserving undo/redo, preview notifications, scene serialization, and cloning.
* Keeps LUT validation and the Preferences copy action in a separate LUT-specific field.
* Caches successful and failed LUT loads across FX clones. The shared MovieRenderer error path now reports the first computation failure, stops remaining rendering, and completes cleanup. This renderer correction also applies to computation failures from other FX.

Supported files include 3D `.cube` tables and Lustre `.3dl` files with a `3DMESH` header and supported uniform, full-range input grids. Flame-format `.3dl` files and unsupported shaper data are rejected with explanatory errors.

This FX changes rendered pixels; Preferences display calibration adjusts their appearance on the current monitor. A creative LUT can coexist with monitor calibration. Baking a monitor-specific correction does not guarantee matching appearance on other displays. Applying the same LUT in both places applies its transformation twice to the displayed result.

The FX evaluates the LUT in nonlinear RGB, with surrounding linear-color conversions handled by the FX framework. Transformed RGB is clipped to 0–1, including floating-point output.

AI-assisted implementation prepared with Codex.


RodneyBaker:
Example use of multiple 3DLUTBakeFX to see the effect as applied to the same image:

image

toonz/sources/toonzqt/lutcalibrator.cpp
```diff
@@ -748,100 +485,26 @@ QString& LutManager::getMonitorName() const {
...
-    loaded = parse3dl(stream, parsed, error);
-  else
-    error = QObject::tr("Supported file types are .3dl and .cube.");
-
-  file.close();
-  if (!loaded)
+  Lut3D loaded;
+  if (!loaded.load(fp, &error))
     return execWarning(
         QObject::tr("Failed to Load 3D LUT File.\n%1").arg(error));
```
shun-iwasawa, line 492: Sorry, I haven't actually tested this myself yet, so I hope I'm just imagining things, but:
For example, if you start rendering 100 frames while having entered the path to a nonexistent LUT file, wouldn't that result in a warning pop-up appearing 100 times?

toonz/sources/toonzqt/paramfield.cpp
```diff
@@ -1608,6 +1614,59 @@ StringParamField::StringParamField(QWidget *parent, QString name,
 
 //-----------------------------------------------------------------------------
 
+void StringParamField::enableLutFileControls() {
```
shun-iwasawa, line 1617: I’m a bit opposed to adding a function to this class solely for the limited purpose of specifying a LUT file path.
For example, how about creating a parameter class (named something like `TFilePathParam`) that inherits from `TStringParam` to hold the file path?
The idea is that the field for `TFilePathParam` (named something like `FilePathParamField`) would include a button to open a file browser.

shun-iwasawa (commented):
The LUT feature in Preferences was designed to perform color correction on monitors where the ViewLUT cannot be configured in hardware.
(For this reason, the LUT switches depending on the current monitor.)

I understand that this effect is intended not for color correction, but rather to transform colors as a visual effect (for example, turning the screen sepia-toned), but is that understanding correct?
If so, it seems to me that the “Use Preferences LUT” button wouldn’t be used very often.

toonz/sources/toonzqt/paramfield.cpp
RodneyBaker: Agreed that is the better approach and think I've addressed the concern.

RodneyBaker:
Thanks for your feedback @shun-iwasawa.  

>  it seems to me that the “Use Preferences LUT” button wouldn’t be used very often.

In testing the main workflow use of the same LUT as used in Preferences appeared essential. 
The primary use case:  
- The user working with a display LUT wishes to share that LUT influenced image with others not using LUTs.  
- The user adds a 3D LUT Bake FX node and needs to locate the LUT being used for display.
Without the option to use that same LUT, the user has to navigate and find the LUT and risks selecting the wrong LUT.  A simple "Use Current LUT" form of selection resolves this.

Additionally: A shortcut capability to quickly toggle the display LUT set in Preferences on/off was added in https://github.com/opentoonz/opentoonz/pull/7091

toonz/sources/toonzqt/lutcalibrator.cpp
RodneyBaker: See additional comments for why this should no longer be an issue.

RodneyBaker:
Likely next steps:

* [x] Refactor `TFilePathParam` to derive from `TStringParam`, with file-filter metadata.
* [x] Add `FilePathParamField` with a text field and Browse button.
* [x] Recognize file-path parameters before the broader string type in the field factory.
* [x] Preserve undo/redo, preview notifications, scene serialization, and parameter cloning.
* [x] Keep LUT validation and the Preferences copy action in the separate `LutFileParamField`.

This addresses the architectural concern while providing a reusable file selection mechanism for other FX, without expanding into a larger file management framework.

-----------------------------------

-----------------------------------

Separately, there is a mismatch between Preview and rendered output when combining display calibration with the LUT Bake FX.   

The Bake FX intentionally writes the LUT transformation into the image data, whereas display calibration changes its displayed appearance. The effect evaluates the LUT in nonlinear RGB, with the surrounding color space conversions handled by the FX framework. 

We need to verify that this behaves consistently in Preview and final rendering, with linear rendering both enabled and disabled.

Future work:
Investigate where the LUT and color space conversions are applied in each path, including whether the same LUT is being applied twice.  It seems overly complex to isolate the canvas only as a space to turn off the Preference (display) LUT when 3DLT Bake FX is introduced.

shun-iwasawa:
Thank you for sharing the potential use cases.

>The user working with a display LUT wishes to share that LUT influenced image with others not using LUTs.

However, I believe it might be difficult to achieve this with the 3D LUT Bake FX.

To use an analogy, the display LUTs settings per monitor in Preferences are like a pair of glasses adjusted to your vision.
If you want to show someone an image you see through your glasses and hand them the glasses and the image,
that person won’t be able to see the same image because their vision is different (i.e., they’re using a different monitor) or they’re already wearing glasses (i.e., they have a different LUT applied).
I believe that sharing the image you see on your monitor requires properly calibrating each monitor (using different display LUTs or other settings tailored to their characteristics) so that the data can be viewed in the same color space.

On the other hand, I think your 3D LUT Bake FX is like a pair of sunglasses.
Sunglasses add beautiful colors to the scenery regardless of the wearer’s vision.
You can even wear them over your regular glasses (though that looks a little strange!).
In other words, a ViewLUT for accurate color reproduction and a LUT used as a color conversion effect can coexist.

RodneyBaker:
@shun-iwasawa 
I like your analogy.  And you are correct about each monitor being equally calibrated in order to properly view an image (LUT or not).

I'm not suggesting images processed with  this 3DLUT FX carry with them any guarantee of color precision.  The FX just remaps colors from one value to another based on the LUT used.

Regarding the glasses:
Few who use this 3DLUT FX will have (truly) calibrated monitors.
Few who use this 3DLUT FX will take the time to calibrate their monitors even if they could.
A few will use your cablibration tools to calibrate their monitors.
Those that use calibrated displays likely won't use this 3DLUT FX at all as it is unnecessary and not advised.  They should use Linear Color instead and share their LUT file with their peers.

This 3DLUT Base FX is all about the sunglasses.
By using the same LUT as used in Preferences or elsewhere the result will be the transfer of color as dictacted by the chosen LUT.  

The following is a draft of the PDF for stuff/doc although it's changed a little since created:

[3DLUTbakFX.pdf](https://github.com/user-attachments/files/32087674/3DLUTbakFX.pdf)

RodneyBaker:
This should hopefully address the concerns raised:

**1. Repeated warnings during rendering**

Addressed through failed-load caching and the renderer correction in [[59d18c4](https://github.com/opentoonz/opentoonz/commit/59d18c4b23349fcc6c690572c9211d3ea7992333)](https://github.com/opentoonz/opentoonz/commit/59d18c4b23349fcc6c690572c9211d3ea7992333).

Successful and failed LUT loads are cached across FX clones using the file revision. 
If a LUT cannot be loaded during rendering, MovieRenderer records the first computation failure, stops the remaining render, clears pending saves, and completes cleanup. 
This prevents a separate failure report for every frame.

The renderer correction also applies to computation failures from other FX.
Focused tests cover concurrent failures and a single failure notification.

**2. Reusable file-path parameter and Browse field**

Implemented as suggested in [[3a53b5b](https://github.com/opentoonz/opentoonz/commit/3a53b5bc37affd61c19833950e3db7eb648cdd0e)](https://github.com/opentoonz/opentoonz/commit/3a53b5bc37affd61c19833950e3db7eb648cdd0e). The earlier architectural checklist is now complete:

Other FX can now use the file selection control without inheriting LUT specific behavior.

**3. Preferences LUT and monitor calibration**

A creative LUT effect can coexist with a calibrated display.
Linear rendering, display calibration, and baking a creative color transformation serve different purposes.

- The button is now named **Copy Preferences LUT**. 
It copies the current monitor’s configured LUT path once.
It does not embed the file or follow later Preferences or monitor changes.

- The tooltip and updated Help PDF explain monitor calibration relates to the display.
- Baking monitor specific correction does not guarantee matching appearance on other monitors. 
- Applying the same LUT in both places applies its transformation twice to the displayed result; this does not mean display calibration should necessarily be disabled when using a separate creative LUT.

The FX evaluates the LUT in nonlinear RGB, with surrounding linear color conversions handled by the FX framework. Preview/output comparisons still need equivalent render and viewing settings.

**4. File support, documentation, and checks**

The [[updated guide and validation changes](https://github.com/opentoonz/opentoonz/commit/e40f6b4fff325d2c62d7bea7e7261a42b357ba78)](https://github.com/opentoonz/opentoonz/commit/e40f6b4fff325d2c62d7bea7e7261a42b357ba78) document supported 3D `.cube` and Lustre `.3dl` files. Flame-format `.3dl` files and unsupported shaper grids are explicitly rejected. FX Settings Help opens `stuff/doc/3DLUTBakeFx.pdf`.

Windows, Linux, macOS, clang-format, and AppVeyor checks all passed for `e40f6b4`. 
Focused local tests also passed for parameters, UI behavior, LUT loading, caching, and renderer error handling.

PR #7116, opened 2026-09-07
Set Default Startup and New Scene Tool

RodneyBaker:
Replaces opentoonz/opentoonz#7101 and opentoonz/opentoonz#7102 with an opt-in tool preference.

- Adds **Preferences > Tools > Default Startup and New Scene Tool**.
- Keeps the Hand Tool as the default.
- Applies the selected tool at application startup and when creating a new scene.
- Stores stable tool command IDs and falls back to Hand if a manually entered ID is invalid.
- Documents the advanced `preferences.ini` override in the preference tooltip:
  - `defaultStartupTool=T_Hand`
  - `defaultNewSceneTool=T_Brush`

While the Preferences control writes both values together; the two INI keys may be set independently for workflows that need different startup and new-scene tools.

Co-authored-by: Ahl76 <145405932+Ahl76@users.noreply.github.com>

There is future work to do with regard to preferences.ini and its undocumented options.
This is important given current development policy regarding maintaining historical and keeping changes that might interfere with classical useage optional.

### DIFF
--- a/toonz/sources/include/toonz/preferences.h
+++ b/toonz/sources/include/toonz/preferences.h
@@ -375,6 +375,12 @@ public:
   TPixel32 getAnimateToolColor() const {
     return getColorValue(animateToolColor);
   }
+  QString getDefaultStartupTool() const {
+    return getStringValue(defaultStartupTool);
+  }
+  QString getDefaultNewSceneTool() const {
+    return getStringValue(defaultNewSceneTool);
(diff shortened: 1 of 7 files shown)


RodneyBaker:
Proceeding with this optional compromise that allows users to maintain current defaults or set the Tool of choice for use after Startup and creation of a new Scene.

PR #7118, opened 2026-09-07
Add Palette option to Reduce Colors

RodneyBaker:
A new suite of options allows the user to maintain artistic control while reducing the color style count.

This feature supports exact color merging, similar color merging and palette reduction to a target count.  It preserves tone and opacity data, supports optional unused style cleanup and consecutive renumbering, and provides Undo/Redo of the process.

Problem addressed
Converstion to TLV format can produce large number of duplicate and similar color styles.
This is usually capped at 4096 styles as that is the limit for palette style entries.
However the image may be able to use very few color styles but there is no effective way to reduce the style count without detrimental results.

This PR is stacked on “Add Palette Convert submenu for lines and areas.”
ChatGPT 6.0 was used in development, testing and video documentation recording automated virtual usage testing.

### DIFF
--- a/toonz/sources/common/tvrender/tpalette.cpp
+++ b/toonz/sources/common/tvrender/tpalette.cpp
@@ -365,6 +365,42 @@ void TPalette::setStyle(int styleId, const TPixelRGBM32 &color) {
 
 //-------------------------------------------------------------------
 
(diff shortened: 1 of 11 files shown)


RodneyBaker:
https://github.com/user-attachments/assets/bf45640a-4b81-4aae-8526-0ad4b5b9c176

PR #7119, opened 2026-09-07
Add Button to Save MyPaint Brushes with Modified Settings

RodneyBaker:
Problem:  Users cannot save/restore brushes with the modifications they have made.

This change adds Save As… button to MyPaint brush Settings so modified brush settings can be saved as a new reusable .myb brush without altering the original.

The default location for saving new brushes is a custom directory in the Library's MyPaint brush directory.  This helps prevent accidental overwriting of source brushes and allows private sets of brushes to be collected together.  An error also tells users to choose a different brush name or location if they try to replace the source brush.

Chat GPT 6.0 was used in creation and processing.


RodneyBaker:
Proceeding with this second part of Mypaint Revert and Save As additions.

PR #7120, opened 2026-09-07
Add Revert Button to MyPaint Brush Settings

RodneyBaker:
Problem:  Settings adjusted for MyPaint brushes must be reset/reverted one at a time.

Here we add a MyPaint only Revert button that restores the current Settings tab changes to the source brush values and disables itself once nothing remains to revert.

ChatGPT 6.0 was used to process this addition.


RodneyBaker:
Proceeding with this quality of life addition.

PR #7121, opened 2026-09-09
Fix enabling color calibration without a LUT

RodneyBaker:
Fixes a crash when enabling LUT color calibration without selecting a LUT file. 

Empty paths are no longer mistaken for a loaded LUT, so calibration remains inactive until valid LUT data is available.


RodneyBaker:
Proceeding with this fix.

PR #7123, opened 2026-09-11
Use theme colors for numeric roller controls

doangelshavewings:
The arrow-and-line controls beneath numeric FX fields were always painted black,
making them difficult to see in dark themes. Draw both arrows and the connecting
line with the widget's WindowText palette color. In Darker they become silver
(#d3d3d3); Light and Neutral retain black controls. Disabled controls use the
theme's disabled text color.

Validation: compiled the changed source with Qt 5.15.19, checked rendered enabled
and disabled controls against the palette in all eight bundled themes, and
confirmed the same check fails against the original black-painting implementation.
Visually inspected Darker and Light renders. Changed lines pass clang-format 14
and git diff --check. Full OpenToonz application testing on Windows is still pending.

Feedback from Soda-Popz JP following #7080. AI-assisted implementation and
validation using Codex.

### DIFF
--- a/toonz/sources/toonzqt/intfield.cpp
+++ b/toonz/sources/toonzqt/intfield.cpp
@@ -64,13 +64,13 @@ void RollerField::getRange(double &minValue, double &maxValue) {
 
 void RollerField::paintEvent(QPaintEvent *e) {
   QPainter p(this);
+  const QColor color = palette().color(QPalette::WindowText);
 
   int w = width();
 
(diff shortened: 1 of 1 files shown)


RodneyBaker:
Initial checks suggest all is good.
Posting intent to merge.

RodneyBaker:
Thank you @doangelshavewings for the adjustment.
LGTM

PR #7124, opened 2026-09-11
Fix restored room layout sizing

RodneyBaker:
Redistributes the deferred room layout after restoration so saved panel sizes fit the current room dimensions, avoiding gaps or clipping after window-size, resolution, or DPI changes.

A revert is unnecessary: https://github.com/opentoonz/opentoonz/pull/6975 deferred timing and viewer-state persistence remain correct; only the final layout fit was missing.


RodneyBaker:
Proceeding with this fix although I don't think nightlies are currently publishing.
@shun-iwasawa will likely need to address that as I don't see anything I can do from this angle.

PR #7125, opened 2026-09-11
Fix Darker theme pressed text contrast

RodneyBaker:
The Darker theme used the same color for pressed control text and its background, causing text in tool option fields and combo boxes to disappear.

This changes the pressed text color to the theme’s standard light text color in the LESS source and updates the generated QSS accordingly. No layouts or other themes are changed.

ChatGPT 6.0 was used in the processing of this adjustment.


RodneyBaker:

PR #6901, opened 2026-05-23
Validate PLT/TZP color metadata

bgyss:
## Summary

- Handles missing optional `TOONZCOLORNAMES` data in PLT/TZP metadata.
- Caps parsed effect parameter counts before copying into fixed local storage.
- Replaces the remaining touched `strcpy`/`strncpy` helpers with exact-length copies.
- Adds `tzp_color_names_regression` for missing metadata and excessive parameter counts.

## Validation

- Original split source commits: `00d08dbe3`, `17c8d7ac7`
- Regression target: `tzp_color_names_regression`
- Previous full-stack validation included the regression test, `image` target build, and focused Semgrep re-scan.

PR #7129, opened 2026-09-14
Fix missing-tag crash in PLT/TZP color metadata

RodneyBaker:
## Summary

Prevent a crash when loading PLT/TZP files that do not contain the optional `TOONZCOLORNAMES` TIFF tag.

Initialize the metadata pointer and pass an empty string to the color-name decoder when the tag is absent.

This focused fix was extracted from #6901. 
Credit to @bgyss for the original implementation.

## Scope

This change only addresses missing color-name metadata. Broader metadata validation and parser hardening are left for a separate PR.

## Testing

* Load a PLT/TZP file containing color-name metadata.
* Load a PLT/TZP file without the `TOONZCOLORNAMES` tag.
* Confirm both files load without crashing.


RodneyBaker:
Proceding with this fix.
Thanks @bgyss!

PR #7130, opened 2026-09-14
Preserve restored panel positions

RodneyBaker:
Saved room panel positions were properly being restored at startup and then immediately overwritten by an unnecessary layout redistribution.

This change preserves the restored panel geometry so customized room layouts remain consistent between sessions.

ChatGPT 6.0 was used to confirm, research and process the fix.


RodneyBaker:
I believe there is more to be addressed with panel position restoration after restart but we'll get those addressed as we can.

PR #7131, opened 2026-09-16
Fix Scene Initialization on Loading XDTS

shun-iwasawa:
This PR adds a line that initializes the scene using the current project settings when an XDTS file is loaded using the Load Scene command.

This is because XDTS files currently store only timing information, whereas properties such as camera size, output format, and others are missing.


RodneyBaker:
LGTM

PR #7135, opened 2026-09-18
Project Settings : Add "Allow No Separator Before Frame Number" Option

shun-iwasawa:
image

This PR will add a new option, `Allow No Separator Before Frame Number`, to Project Settings.

When this option is enabled, files that previously required a period (`.`) or underscore (`_`) as a separator between the level name and frame number—such as `A0001.png`—will now be recognized as sequential images (i.e., a single raster level).

This is useful, for example, when you want to import sequential images exported from Clip Studio Paint without renaming them.

For files recognized as sequential images without a separator, a hash symbol (`#`) will appear in the position where the separator would normally be when viewed in the File Browser or Level Settings.

`A0001.png, A0002.png... → A#.png`

Additionally, the “no separator” option becomes available in the output settings only when this option is turned on.


Abel032:
If it's possible to recognize these files as sequence images without separator, maybe we should revert this: #5768.

cc: @konero

RodneyBaker:
> If it's possible to recognize these files as sequence images without separator, maybe we should revert this: #5768.

My quick take is that your PR is still useful particularly because this new option is only on when the user opts in to it via Preferences.

RodneyBaker:
@shun-iwasawa 
The following areas were flagged during my initial review of this PR.
All of which are offset by this change being an optional Preference.

- withName() and withFrame() round tripping
- isDefault()
- getSepChar() consumers where # can reach real filenames
- Unused enum additions

----------------------------------

In `withName()`:  `A#.png.withName("B")` becomes `B.png`.

Should this remain `B#.png`?
Will renaming an output or level path in this way lose its sequence identity?

----------------------------------

in `withFrame()`:

Starting from A.png with an actual frame whose separator marker is # can produce A#0001.png whereas it should produce A0001.png.

This will effect Export Level and other callers in ways not currently addressed by this PR.
  
----------------------------------
----------------------------------
Regarding the check:  `m_noSeparatorFormatAllowed == true`

As the constructor's default is false it seems this should be false or `!m_noSeparatorFormatAllowed`. As written ordinary default projects are unnecessarily treated as non-default which will serialize a file path properties block.

----------------------------------

The internal # marker can leak into real board filenames.

----------------------------------
The new FrameFormat values appear incomplete and/or potentially unnecessary.

The three values are inserted in the middle of the exported enum, shifting existing numeric values, but getCurrentFormat() never returns them. Separator handling is already performed through the # marker. Unless genuinely required these can be removed, or implemented without renumbering existing entries. Additionally, `NOS_SEP_CUSTOM_PAD` appears to be a naming typo.

All this for whatever it's worth...

PR #7136, opened 2026-09-18
Fix Crash on New Level

shun-iwasawa:
This PR fixes crash on create a new Toonz Raster / Toonz Vector level under a project in which the `+palettes` alias is not set.

In such case, now OT will simply load the default palette from the studio palette folder (`$TOONZSTUDIOPALETTE\Global Palettes\Default Palettes`)


RodneyBaker:
Thanks @shun-iwasawa for this important fix.
LGTM

